PR #2, opened 2018-09-07
Style Proposal

beerriot:
This is more of a proposal with example than an actual pull request, but accepting it as a PR is also fine.

In reading through the code, it seems that style is undefined. Tabs in some places, spaces in others. Long lines, braces on the same line or the next line, extra newlines, trailing space, etc. I find that diffs are easiest to handle if a style is defined, which editors can follow. So, I'd like to propose we choose a style before diffs begin flying.

This PR proposes adopting Google's [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint) as style definition. The diff attached is what it took to clear all complaints from `ViewsManager.cpp` and `ViewsManager.hpp`.

Before:

[`Total errors found: 997`](https://gist.github.com/beerriot/f43543fef4e90aa343d81892936bdb6d)

After:

```
$ ~/deps/styleguide/cpplint/cpplint.py --headers=hpp ./bftengine/src/bftengine/ViewsManager.{cpp,hpp}
Done processing ./bftengine/src/bftengine/ViewsManager.cpp
Done processing ./bftengine/src/bftengine/ViewsManager.hpp
Total errors found: 0
```

Most of these errors (90%) are simple formatting (long lines, brace placement, comment placement). This is the bulk of what I want to propose standardizing, to help with diffs.

The remaining items are other C++ hygeine: explicit casting, constant naming, non-const references and out parameters. I've gone ahead and included those changes, just to see how extensive they may be.

I don't have any particular preference for specifics, but cpplint seemed popular enough to throw in as a concrete proposal.

What do people think?

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -363,7 +363,7 @@ namespace bftEngine
 		{
 			Assert(isCurrentPrimary() && currentViewIsActive());
 
-			if (primaryLastUsedSeqNum + 1 > lastStableSeqNum + workWindowSize) return;
+			if (primaryLastUsedSeqNum + 1 > lastStableSeqNum + kWorkWindowSize) return;
 
 			if (primaryLastUsedSeqNum + 1 > lastExecutedSeqNum + maxConcurrentAgreementsByPrimary) return; // TODO(GG): should also be checked by the non-primary replicas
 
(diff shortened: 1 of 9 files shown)


guyg8:
It looks like a good idea to me.
To make it easier to check the style updates, I suggest that we always change spaces in dedicated commits.

ittaiab:
Yes!

PR #4, opened 2018-09-22
Fix compiler warnings

beerriot:
Building concord-bft generated a number of warnings around:

 * extra semicolons
 * signed/unsigned comparison
 * unused functions/variables
 * unmatched enumeration values in switch statements

This change cleans them all up, so that no warnings are reported with either Clang 3.8.0 or GNU 5.4.0.

Since the changes were fairly extensive to CommFactory.{cpp,hpp}, I also cleaned up the cpplint complaints for those files (the last three commits here).

I also accidentally committed my addition to .gitignore for emacs temporary files. Sorry to include it in unrelated changes, but it's a good and simple addition.


guyg8:
@beerriot: 
Overall it looks good. Thanks!
Comments/Questions:
The new functions in CommFactory.cpp (create_config_impl) should be part of the bftEngine namespace (at least according to CommFactory.hpp).
Why the template function “create_config(Args... args)” was moved to CommFactory.cpp ? did you check that the instances of this template function are actually created (and the linker is able to find them …)?

guyg8:
@salieri11:
Do we really need all the “template games” in CommFactory.cpp/hpp ? ( This code may be more readable without templates)

salieri11:
@guy I think I got rid of them in the integration commits

guyg7:
@salieri11:
Great. Let's try to keep the interfaces as simple as possible.
I suggest to also delete these functions from this repository.

beerriot:
re: config-impl functions -- I didn't do much checking here, because the functions are not used anywhere. As it seems @guyg7 and @salieri11 agree, I will remove them and push an update here. Thanks for reviewing!

PR #5, opened 2018-10-08
Various fixes and refactoring

salieri11:
* removed config create methods - no usage for now
* renamed public header
* styling fix
* increase client max msg size to 64k
* fixing compilation warnings
* exposed threshsigh include folder as global build variable


salieri11:
This should be merged to vmware/concord-bft

PR #8, opened 2018-10-08
Style and refactoring

salieri11:
This PR contains multiple style fixes, minor code refactoring and compile warning fixes


beerriot:
This PR also includes the new `UPDATE_CALLBACK_FN statusCallback` feature.

beerriot:
We have reviewed these changes offline, and agreed that the resulting
state of the code is what we want. I've just checked, and the state of
the code in this PR matches the state of the code we agreed on offline
(i.e. there are no diffs). So, +1 from that angle.

However, this PR does not follow the contribution
guidelines. Particularly, the details about the format of a commit
message. So, -1 from that angle. I'd really recommend squashing all of
the empty merge commits here, and then re-writing the commit message,
and using that message as the description of this PR. "Integratton
with Athena" doesn't mean anything to our external collaborators, but
what would is something about the new status callback, some details
about what warnings were fixed, and the reasons for refactorying.

salieri11:
This PR will be closed and the new one will be opened for clean commits history

Issue #1, opened 2018-09-02
Logging utility refactor


salieri11:
Currently logging in the library is inconsistent. The desired solution is to implement a wrapper that will abstract dependencies from the library code.
1. Implement logging interface similar to standard interfaces ( Log4CPlus is preferrable one with support of << operator ).
2. Implement simple console logger to be a part of this library.

PR #11, opened 2018-10-24
Fix #10, implement #1

### LINKED ISSUE #10: UDP vs. interesting network topologies
We were recently debugging some strange behavior in an app running on concord-bft. When we started the application, we would see some initial read-only requests be handled just fine. But, when we started sending write requests, they would sometimes time out, and we would see errors in the log like:

```
 22:41:39.878 WARNING: Node 1 received invalid message from Node 0 (type==105)
```

We saw many different "type" values, and each replica index would appear as the receiver. Curiously, it was always Node 0 that was logged as the sender of the invalid message.

We tracked this down to a problem with our network, which confused the UDP communication module.

The problem with the network is that each replica was running in a docker container on a different host, and was attaching to the `bridge` network. The UDP communication module uses the socket details to determine the source of the message, and these socket details were giving the IP and port of the bridge network (172.17.*) instead of the host network (10.*). Unfortunately, if the IP:port pair is unknown to the UDP module, instead of returning an error, [it returns "0"](https://github.com/vmware/concord-bft/blob/6d2771fd4a68872c2e0a9c95522a9b5938ef768d/bftengine/src/communication/PlainUDPCommunication.cpp#L301). For a variety of reasons (signature matching, expectations of primary for a view, etc.), a message from some other node will be considered an invalid message if it's thought to come from Node 0.

The fix for our deployment was to switch our docker containers to using the `host` network, so that the socket source matches the configured source.

To help others find and fix this problem in their deployments, it would be nice to see a change in the UDP communication module, to log when message appears to come from an unknown source (including the source).

salieri11:
This PR contains the following items:

bug fixed:
when using UDP, instead of returning 0 if message is received from unknown sender, we just ignore this message and produce warning to logs

features:
implemented logging wrapper, that is supposed to be an interface and internally implements either 
log4cpp (if set in CMAKE) or naive console logger. The previous logger is not removed yet and the 
transition to the new one should be continuous effort. w

misc:
added bug and feature request templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,17 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,17 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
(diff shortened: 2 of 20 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
+---
+name: Bug report
```
beerriot: Oh, one more comment. This bug report template and the feature request template below it are not mentioned anywhere in the PR description. Was this an accidental addition?

PR #12, opened 2018-10-25
Disabling padding in all structs inheriting from MessageBase

shellygr:
- Wrapping each relevant struct definition with #pragma pack(push,1) ... #pragma pack(pop)
- Adding static_assertions for checking the size of each msg struct (I checked the effect without the pragma setting - it did not pass, as expected)


bftengine/src/bftengine/ClientReplyMsg.hpp
```diff
@@ -23,6 +23,7 @@ namespace bftEngine
 			static_assert(sizeof(ClientReplyMsgHeader::msgType) == sizeof(MessageBase::Header), "");
 			static_assert(sizeof(ClientReplyMsgHeader::reqSeqNum) == sizeof(ReqId), "");
 			static_assert(sizeof(ClientReplyMsgHeader::currentPrimaryId) == sizeof(ReplicaId), "");
+			static_assert(sizeof(ClientReplyMsgHeader) == 16, "ClientRequestMsgHeader is 16B");
```
beerriot: Looks like a copy-paste typo: the message should say "Reply", not "Request.

guyg8:
The messages format should not depend on the compiler and the architecture. This is the first step in doing so.

guyg8:
In our case, the overhead of accessing unaligned messages should be negligible

PR #13, opened 2018-10-25
zeroing buffer

shellygr:


guyg8:
We have seen that because of the padding some messages contain uninitialized regions. This may result in leakage of confidential information. This update should help to avoid sending uninitialized values.

PR #15, opened 2018-10-28
fix #14 - incorrect UDP Stop process implementation

### LINKED ISSUE #14: Incorrect UDP Stop implementation
**Describe the bug**
When PlainUdpCommunication::Stop is called, sometimes the call doesn't return. The reason is that isRunning is not used correctly and when the socket is closed, the receive thread continues to wait for messages.
Moreover, there is no need to call shutdown() on DGRAM socket. close() is enough.

**To Reproduce**
Run attached test few times - eventually one of the calls will not complete

**Expected behavior**
To shut UDP comm correctly and to return immediately

salieri11:
This PR fixes bug #14 - when Stop is called for the UDP module, the "continue" that is called if there is socket error skips sRunning assessment in the recvThreadRoutine method - and this method will not return thus preventing UDP receive thread to join.

The fix removes local sRunning variable and evaluates the "running" member directly. This member implemented as atomic<boolean> for better thread safety. 

Along with this fix, the shutdown call is fixed for windows.

### DIFF
--- a/bftengine/src/communication/PlainUDPCommunication.cpp
+++ b/bftengine/src/communication/PlainUDPCommunication.cpp
@@ -20,6 +20,7 @@
 #include "CommDefs.hpp"
 #include "Threading.h"
 #include "Logging.hpp"
+#include <atomic>
 
 #define Assert(cond, txtMsg) assert(cond && (txtMsg))
 
@@ -64,9 +65,6 @@ class PlainUDPCommunication::PlainUdpImpl {
   /* The list of all nodes we're communicating with. */
   std::unordered_map<NodeNum, NodeInfo> endpoints;
 
-  /** Flag to indicate whether the current communication layer still runs. */
-  bool running;
-
   /** Prevent multiple Start() invocations, i.e., multiple recvThread. */
   Mutex runningLock;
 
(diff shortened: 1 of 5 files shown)


bftengine/src/communication/PlainUDPCommunication.cpp
```diff
@@ -337,8 +347,7 @@ class PlainUDPCommunication::PlainUdpImpl {
 
   void
   recvThreadRoutine() {
-    bool sRunning = true;       /** Indicates whether we're still running */
-
+    running = true;
```
beerriot: This is set in `Start`, after the thread is created. I think that setting it here actually introduces a race condition:

1. Start starts this thread.
2. Start sets running = true.
3. Stop is called.
4. Stop sets running = false.
5. This thread finally starts executing, and sets running = true.
6. Stop is left waiting to join this thread, but nothing will set running to false.

Who knows why someone would call Start and then immediately call Stop in a way that would produce the above scheduling. But, I think it can be fixed anyway, by removing this `running = true` within the thread, and moving the `running = true` in `Start` to before the thread creation (the call to `startRecvThread`).

PR #18, opened 2018-10-29
Update relic build instructions for newer versions of relic

andrewjstone:
Newer versions of relic use `WSIZE` instead of `WORD` for word size.

Also, fixed trailing whitespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Community
 
 [Concord-BFT Slack](https://concordbft.slack.com/).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,6 +97,9 @@ Build and install [RELIC](https://github.com/relic-toolkit/relic)
     make
     sudo make install
 
+Note that on newer versions of cmake, `-DWORD` must be replaced with `-DWSIZE` in the above cmake command.
```
beerriot: > newer versions of cmake

I think this is for newer versions of *relic*, not cmake. https://github.com/relic-toolkit/relic/commit/bf6d0e37614d18afe52cfe1b5c6bbde18df1041e

andrewjstone: Ah, ok. Should I just go ahead and change the command in the instruction rather than adding a comment then? Or should I modify the comment to reflect it's about relic rather than cmake?

beerriot: You know, that's a good point. Anyone that pulls now is going to get the new Relic. Updating the actual command does seem like the right choice, in that case.

PR #27, opened 2018-11-08
fix TCP connectivity and  add support for status info callback

salieri11:
This PR fixes the TCP connectivity between replicas and clients and adds support for status info callback to the TCP module.
The design is that all replicas should always listen for incoming connections and the clients, with IDs higher then maxServerId, connect to them. Between the replicas themselves, replicas with higher IDs are supposed to connect to replicas with lower IDs - and the bug was that the last replica (maxServerId) was missed. Moreover, the highest replica was not listening.
It also fixes the incorrect ip and port information being reported via the status callback by UDP module.

### DIFF
--- a/bftengine/src/communication/PlainTcpCommunication.cpp
+++ b/bftengine/src/communication/PlainTcpCommunication.cpp
@@ -166,9 +166,10 @@ class AsyncTcpConnection :
   }
 
   void close_socket() {
-    LOG_TRACE(_logger, "enter, node " << _selfId << ", dest: " << _destId
-                                      << ", connected: " << connected << ", closed: "
-                                      << _closed);
+    LOG_TRACE(_logger, "enter, node " << _selfId
+              << ", dest: " << _destId
+              << ", connected: " << connected
+              << ", closed: " << _closed);
 
     try {
       boost::system::error_code ignored_ec;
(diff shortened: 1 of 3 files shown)


bftengine/src/communication/PlainTcpCommunication.cpp
```diff
@@ -159,20 +177,20 @@ class AsyncTcpConnection :
...
-                                   << ", connected: " << connected
-                                   << ", ex: " << e.what());
+                                            << ", connected: " << connected
+                                            << ", ex: " << e.what());
     }
 
     LOG_TRACE(_logger, "exit, node " << _selfId << ", dest: " << _destId
-                            << ", connected: " << connected << ", closed: "
-                            << _closed);
+                                     << ", connected: " << connected << ", closed: "
```
beerriot: This re-indentation doesn't seem inherently better than what was here before. I could see wanting to line everything up "connected: connected" over "closed:closed", but this sort of reindentation for reindentations sake is often a source of unnecessary merge conflicts later.

bftengine/src/communication/PlainTcpCommunication.cpp
```diff
@@ -687,25 +759,28 @@ class PlainTCPCommunication::PlainTcpImpl {
...
       _listenPort{listenPort},
       _listenIp{listenIp},
       _bufferLength{bufferLength},
       _maxServerId{maxServerId},
-      _nodes{nodes} {
-
-    if (_selfId < _maxServerId) {
+      _statusCallback{statusCallback} {
+    // all replicas are in listen mode
+    if (_selfId <= _maxServerId) {
```
beerriot: Was this the multi-client bug you fixed? Whether it was or not, could you please add some description to the PR about what the bug was, and how you fixed it? "Fixed a bug" doesn't tell the next maintainer much.

salieri11: I changed the title and the description to me more precise

beerriot: Looks good. Thanks for the update!

bftengine/src/communication/PlainTcpCommunication.cpp
beerriot: I like the update made in 083f284. The restructuring to pair label with value is good. Re-approved. 👍

PR #29, opened 2018-11-12
fix compilation warnings (signed/unsigned comparison)

salieri11:
This PR fixes signed/unsigned comparison warnings


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -1306,7 +1306,7 @@ bool BCStateTran::onMessage(const FetchBlocksMsg* m,
       break;
     }
     // if we still have chunks in block
-    else if (nextChunk+1 <= numOfChunksInNextBlock) {
+    else if (static_cast<uint16_t>(nextChunk+1) <= numOfChunksInNextBlock) {
```
beerriot: This one and the one below are caused by integer promotion. You could just do `nextChunk + 1u` instead. Not a super big deal, but less noise.

PR #51, opened 2019-02-15
Add Googletest as a dependency in a submodule

andrewjstone:
I added gtest as a submodule so we can start to write unit tests.
CMakeLists.txt was updated to pull it in to the build. A very basic
example test was added in bftengine/tests/datastore. The README was also
updated to reflect that submodules need to be initialized.

Additionally, vim temporary files are now ignored by adding them .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ cmake-build-debug/
 .vscode/
 README.html
 venv/
-
+*.swp
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "deps/gtest"]
+	path = deps/gtest
+	url = git@github.com:google/googletest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,3 +111,6 @@ add_subdirectory(threshsign)
 add_subdirectory(bftengine)
 add_subdirectory(tools)
 add_subdirectory(bftengine/tests)
+
+# Dependencies as submodules
+add_subdirectory(deps/gtest)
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -4,4 +4,4 @@ cmake-build-debug/
 .vscode/
 README.html
 venv/
-
+*.swp
```
upshaw-alex: Does everyone find this change acceptable and are there any conventions or style guides we are adhering to that include principals for `.gitignore`-ing editor temporary files? I would certainly find this change convenient, as I also use Vim, though I have heard some people object to including editor-specific temporary files in `.gitignore` files on the grounds of editor-agnosticism.

andrewjstone: IMO, if there is *any* file that may be created on your computer that should not be checked in, it's fair game to add it to gitignore. It doesn't hurt anybody, and saves us from accidentally committing files that shouldn't be there. 

Also, .vscode was already excluded, along with emacs backup files (*~)

andrewjstone:
> good idea, @andrewjstone ! One thing: how can we make sure that this dependancy is optional? We would like to give freedom of choice to the users, and to let them to decide about as much dependencies as possible

This would require some explicit changes to the build to not initialize all submodules. It would also require not building tests. Basically, we'd have to change the build and add a test only target where we pull needed dependencies and build. If you are building and running tests, the dependency is not optional.

Making dependencies optional can be valuable for some things, but it also adds build complexity. Let's discuss more offline about why we want to make dependencies optional.

andrewjstone:
I have a new PR coming in the future which pulls in google test optionally, only when tests are being built. I should have it out in a day or 2.

PR #52, opened 2019-02-24
Add support for loading client/replica configuration from a file to run simpleTest on remote machines.

yuliasherman:
Support a new cli parameter -cf that allows specifying configuration file instead of using hardcoded parameters. The file is in YAML-style format; proprietary parser is written for it.


bftengine/tests/simpleTest/config_file_parser_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+  cin >> tmp;
+  if (tmp == ".\n")
+    file = tmp;
+  concordlogger::Logger logger =
+      concordlogger::Logger::getLogger("simpletest.test");
+  ConfigFileParser parser(logger, file);
+  if (!parser.Parse())
+    return 1;
+  cout << "\n";
+  parser.Count("replicas_config");
```
beerriot: It would be nice if this test were a little stronger. It seems like it only verifies that parsing config.txt doesn't make the program crash. Can it also verify that the result of the parsing is what we expect?

bftengine/tests/simpleTest/test_comm_config.cpp
```diff
@@ -0,0 +1,175 @@
...
+    LOG_INFO(logger_, "setUpConfiguredNodes() id: " << id <<
+                      ", k: " << k - 1 <<
+                      ", port:" << (uint16_t)(std::stoi(ip_port_pair[1])));
+    if (ip_port_pair.size() != 2) {
+      LOG_FATAL(logger_, "Wrong number of parameters configured for "
+                          "replica/client ip/port pair: " <<
+                          ip_port_pair.size());
+      exit(-1);
+    }
+    if (i == id) {
```
beerriot: I might recommend giving `id` a more descriptive name, like `local_id`, or something else to indicate that it's the index of the principal running this code. Not a huge deal, but I found it took me a bit to figure out what its purpose was.

…oh, this is a holdover from the old config functions. Not necessary to fix for this review, but might be a helpful cleanup in the future.

bftengine/tests/simpleTest/CMakeLists.txt
```diff
@@ -1,26 +1,39 @@
 set(simpleTest_client_source_files
     client.cpp
-    config.cpp
+        test_comm_config.cpp
```
beerriot: nit: please follow the existing indentation style

bftengine/tests/simpleTest/test_comm_config.cpp
```diff
@@ -0,0 +1,175 @@
...
+                       ", numOfReplicas=" << num_of_replicas);
+    exit(-1);
+  }
+  vector<string> replicas = config_file_parser.GetValues(REPLICAS_CONFIG);
+  vector<string> clients  = config_file_parser.GetValues(CLIENTS_CONFIG);
+  std::unordered_map <NodeNum, NodeInfo> nodes;
+  int k = 0;
+  for (int i = 0; i < (num_of_replicas + num_of_clients); i++) {
+    vector<string>& current_vector = replicas;
+    if ((k == i) && (i >= num_of_replicas)) { // Ensure this happens only once.
```
beerriot: Could you add a little more detail about *why* this needs to happen only once, please? If I understand correctly, it's because `k` is used to map the principle ID, `i`, to the replica/client ID.

Would it be cleaner to just say `k = (i < num_replicas) ? i : i - num_replicas`, instead of this stateful variable? That would also get rid of the `k++` / `k - 1` juggle below.

bftengine/tests/simpleTest/CMakeLists.txt
yuliasherman: NP, missed it.

bftengine/tests/simpleTest/test_comm_config.cpp
yuliasherman: Done.

bftengine/tests/simpleTest/test_comm_config.cpp
yuliasherman: I'll change the comment, np.
Regarding the code: I still need to switch between replica and client vectors, so "if" should remain. Besides that, after a switch to clients vector, k should be reset to 0 and it will not be a function of "i", so I prefer to leave this code as is.

bftengine/tests/simpleTest/config_file_parser_test.cpp
yuliasherman: Done.

bftengine/tests/simpleTest/config_file_parser_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+  const string expected_replica1 = "127.0.0.1:3410";
+  const string expected_replica2 = "127.0.0.1:3420";
+  const string expected_replica3 = "127.0.0.1:3430";
+  const string expected_replica4 = "127.0.0.1:3440";
+  const string expected_client   = "127.0.0.1:4444";
+  const string values_to_split = "10.23.43.1:1234:1238";
+  const string expected_split_values[] = { "10.23.43.1", "1234", "1238"};
+  const string values_to_split_delimiter = ":";
+
+  cin >> given_config_file;
```
beerriot: I don't think this is actually working. It seems to be trying to open a local sample_config.txt no matter what.

```
~/builds/concord-bft/debug/bftengine/tests/simpleTest$ ./config
Enter configuration file name with a full path:
/home/bfink/vmwathena/concord-bft/bftengine/tests/simpleTest/sample_config.txt
FATAL 2019-02-26 10:56:22.520 (simpletest.test) Failed to open file: sample_config.txt
```

If I copy `sample_config.txt` into the local directory, it successfully opens that file no matter what path I type in.

Ultimately, my question is, should this be a parameter that is typed in? Given the new validation logic, the test is tightly coupled to the example file. Maybe that path should just be static?

beerriot:
Also it would be really nice if the `testReplicasAndClients.sh` script were converted to use this new `-cf` parameter, as a demonstration.

bftengine/tests/simpleTest/config_file_parser_test.cpp
yuliasherman: > @yuliasherman, VMware has approved your signed contributor license agreement.

As agreed with Guy, we leave our *.sh scripts without changes because the intention is to move to python scripts. testReplicasAndClients.sh does not support configuration on remote computers anyway. Our python scripts should be also modified to start supporting launching replicas on remote nodes. This should be done as a different task.

yuliasherman: You are right, I forgot to test this case - when a given file is different from the default. That's exactly a reason why I want to leave this option supported - it verifies that a given file is used and not a default one.

bftengine/tests/simpleTest/CMakeLists.txt
```diff
@@ -30,22 +43,28 @@ endif()
...
 
+target_include_directories(configTest
+                           PRIVATE
+                           ${concord_bft_tools_SOURCE_DIR}
+                           ${bftengine_SOURCE_DIR}/include)
+
 add_custom_target(copy_simple_test_scripts ALL COMMENT "Copying scripts of simpleTest")
 add_custom_command(TARGET copy_simple_test_scripts
     COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:simpleTest_client>/scripts
     COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/scripts $<TARGET_FILE_DIR:simpleTest_client>/scripts)
```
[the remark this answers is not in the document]
yuliasherman: I've moved sample_config.txt file to the scripts directory and now it will be copied as well.

PR #54, opened 2019-02-28
Format all code using clang-format

andrewjstone:
All code was formatted using .clang-format file. SimpleTest passes. 

This code is split into two logical commits. The first includes the support code for doing the formatting and the documentation changes. The second includes the actual formatted code.


scripts/format-code.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env sh
+
+BFT="concord-bft"
+
+if [ -z "$1" ]; then
+  >&2 echo "ERROR: Path to ${BFT} directory required"
```
beerriot: I wonder if we could be smarter here, and just try "." or ".." if no path is given. It might also be nice to add the usual example style to the end, e.g. `Usage: format-code.sh <concord-bft path>`.

bftengine/src/bftengine/ViewChangeSafetyLogic.cpp
```diff
@@ -1,11 +1,14 @@
-//Concord
+// Concord
 //
-//Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
 //
-//This product is licensed to you under the Apache 2.0 license (the "License").  You may not use this product except in compliance with the Apache 2.0 License. 
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
```
beerriot: Tagging this instance, though it occurs in many/most of these files. When I run `format-code.sh` locally, this is re-wrapped, because it's 82 characters wide here:

```
-// You may not use this product except in compliance with the Apache 2.0 License.
+// You may not use this product except in compliance with the Apache 2.0
+// License.
```

scripts/format-code.sh
andrewjstone: I think I prefer the Usage example. We don't want people to accidentally format everything and get spooked. Then again, we really do just want to run this from the top-level dir. I'll add usage.

bftengine/src/bftengine/ViewChangeSafetyLogic.cpp
andrewjstone: This is a quirk in how clang-format treats very long comments. For lines where the original sentence (82 chars) was on one line it wrapped it. For one very long line it wrapped it but left that lest part at 82 chars. Re-running it fixed those remaining lines. Re-running a third time produced no changes, so I think we are good now.

I rebased and pushed up the changes.

scripts/format-code.sh
andrewjstone: Fixed.

PR #58, opened 2019-03-18
Clang-format: Add support code for auto code formatting

andrewjstone:
This is a port of work originally done by @robem for our closed source
stuff.

Add clang-format config file

The config file introduced in this change will be the base for all future
clang-format runs. It was generated with the following command.

  clang-format -style=Google -dump-config > .clang-format

Three modifications were made to the generated config. We don't sort
includes, and we don't binpack parameters or arguments.

Add a wrapper script for running clang-format

We are interested in C/C++ related files only and hence we need to filter all
files under concord-bft. This change introduces a script that runs
clang-format with the proper arguments for you and is written with CI/CD in
mind. If you provide `--is-required` to the script then it will return 0 if no
code format change is needed and 1 otherwise.

A cmake script was added for the script, but not integrated into the
top-level build. We can integrate it later when we have a more flexible
top-level build with multiple targets.

Also add some documentation about installing clang-format.


andrewjstone:
It appears that #54 is not ever going to get merged. I'd like to merge this in, which only adds the script and config file so we can do cleanup piecemeal. Additionally, it allows developers to use the config file to setup their editors so that new code will be formatted correctly.

guyg7:
Is it possible to change the format of a single file (rather than an entire directory)?

andrewjstone:
> Is it possible to change the format of a single file (rather than an entire directory)?

The script does not do that, but clang-format can be used directly like so:
`clang-format -style=file -fallback-style=none -i bftengine/src/bftengine/ReplicaImp.cpp`

andrewjstone:
I added support so that format-code.sh can take an individual file in addition to directories. I verified it by running it like so: `scripts/format-code.sh bftengine/tests/simpleKVBCTests/simpleKVBCTests.cpp` and `scripts/format-code.sh bftengine`.

PR #59, opened 2019-03-20
Add initial bcstatetransfer test

andrewjstone:
This code introduces the first unit test for bcstatetransfer along with
gtest fixtures for writing the rest of the tests. The goal is to test
both fetching and sending sides in separate tests and ensure that fake
state can be properly transferred in block and chunk form correctly.

This test is the first part of that and only ensures that the
AskForCheckpointSummariesMsg is sent to all other replicas when state
transfer is started. The test will be enhanced to reply with checkpoint
summaries from faked out replicas and then to ensure state transfer
completes, as messages get sent and timers fire.

gtest is introduced as an optional dependency and only fetched and built
when tests are run.

I also removed some duplication in the BCStateTran code itself, and
shortened some very long functions by breaking some code out into its
own functions. I also added an underscore to some member variables that
were missing them, in accordance with both our style policy and style
exising in the file already. Lastly I removed some dead code and
extraneous whitespace. Behavior should not have changed.


bftengine/tests/bcstatetransfer/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(bcstatetransfer_tests
```
beerriot: If you also add `add_test(bcstatetransfer_tests bcstatetransfer_tests)`, then this will get run by a `make test`.

bftengine/tests/bcstatetransfer/test_app_state.hpp
```diff
@@ -0,0 +1,89 @@
...
+#ifndef BFTENGINE_TESTS_BCSTATETRANSFER_TEST_APP_STATE_HPP_
+#define BFTENGINE_TESTS_BCSTATETRANSFER_TEST_APP_STATE_HPP_
+
+#include <cassert>
+#include <cstring>
+#include <unordered_map>
+#include "SimpleBCStateTransfer.hpp"
+
+// This should be the same as test config
+const uint32_t kMaxBlockSize = 1024;
```
yuliasherman: Should kMaxBlockSize be visible outside SimpleBlockchainStateTransfer?

bftengine/tests/bcstatetransfer/test_app_state.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+class TestAppState : public IAppState {
+
+  public:
+    bool hasBlock(uint64_t blockId) override {
+      auto it = blocks_.find(blockId);
+      return it != blocks_.end();
+    }
+
+    bool getBlock(uint64_t blockId, char* outBlock, uint32_t* outBlockSize) override {
```
yuliasherman: Why not pass parameters by reference?

bftengine/tests/bcstatetransfer/test_app_state.hpp
guyg7: Yes, it is a parameter to SimpleBlockchainStateTransfer.
(see SimpleBlockchainStateTransfer::Config)

bftengine/tests/bcstatetransfer/test_app_state.hpp
andrewjstone: This is an implementation of IAppState, and that is the function signature. Note the `override` keyword.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2067,7 +2022,46 @@ uint16_t BCStateTran::selectSourceReplica() {
   return *i;
 }
 
+void BCStateTran::SetAllReplicasAsPreferred() {
+  set<uint16_t> tmp = replicas_;
+  tmp.erase(myId_);
+  preferredReplicas_ = tmp;
```
salieri11: Why can't you just do preferredReplicas_.erase(myId_)?

bftengine/tests/bcstatetransfer/CMakeLists.txt
andrewjstone: thanks!

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -1,7 +1,6 @@
 // Concord
 //
 // Copyright (c) 2018 VMware, Inc. All Rights Reserved.
-//
```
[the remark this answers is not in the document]
andrewjstone: yep. Will fix it up.

andrewjstone:
This is being replaced by two separate PRs. Thanks for the reviews!

PR #61, opened 2019-03-25
Refactor Simple KVBC tests CLI parameters

salieri11:
This PR introduces new CLI parameter, -n, for the Simple KVBC test. The parameter expects the name of communication configuration file, which enables the test to be ran on different machines, as described in #52 .
Moreover, the Simple KVBC client and replica code has been refactored to use common test structures, similar to the SImple test.


salieri11:
Closed, was open from wrong branch.

PR #63, opened 2019-03-25
Add initial bcstatetransfer test

andrewjstone:
This code introduces the first unit test for bcstatetransfer along with
gtest fixtures for writing the rest of the tests. The goal is to test
both fetching and sending sides in separate tests and ensure that fake
state can be properly transferred in block and chunk form correctly.

This test is the first part of that and only ensures that the
AskForCheckpointSummariesMsg is sent to all other replicas when state
transfer is started. The test will be enhanced to reply with checkpoint
summaries from faked out replicas and then to ensure state transfer
completes, as messages get sent and timers fire.

gtest is introduced as an optional dependency and only fetched and built
when tests are run.

For testing purposes I had to make FetchingState and its corresponding
getter public.


andrewjstone:
#59 is being split into 2 separate PRs:

 1. BCStateTransfer autotmated testing
 2. Refactoring of BCStateTransfer code

This is part 1.

PR #71, opened 2019-04-02
Set default client iterations to 2800 and add license header

salieri11:


bftengine/tests/config/test_parameters.hpp
```diff
@@ -1,12 +1,21 @@
+// Concord
 //
-// Created by Igor Golikov on 2019-03-24.
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
```
andrewjstone: Nit. This should Probably be 2019 (based on the comment being replaced)

PR #73, opened 2019-04-03
Track aggregation timestamp in component values

andrewjstone:
This will allow us to display the time when the metrics are from for
each component once we add serialization and a server.

Note that we use the system clock here because we are going to want a
human readable time. It's possible that this isn't monotonic, and so the
test can possibly fail. That'd be pretty damn unlucky though. If we
think it's a problem I can just remove the testing, as it's not strictly
necessary.

### DIFF
--- a/util/include/Metrics.hpp
+++ b/util/include/Metrics.hpp
@@ -19,6 +19,7 @@
 #include <vector>
 #include <mutex>
 #include <memory>
+#include <chrono>
 
 namespace concordMetrics {
 
@@ -98,6 +99,8 @@ class Counter {
 
 class Values {
  private:
+  std::chrono::system_clock::time_point last_aggregated_;
+
   std::vector<Gauge> gauges_;
   std::vector<Status> statuses_;
   std::vector<Counter> counters_;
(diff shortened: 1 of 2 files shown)


util/include/Metrics.hpp
```diff
@@ -162,13 +165,21 @@ class Component {
...
   // names will not properly exist in the aggregator, since only values get
   // updated at runtime for performance reasons.
-  void Register() { aggregator_->RegisterComponent(*this); }
+  void Register() {
+    values_.last_aggregated_ = std::chrono::system_clock::now();
+    aggregator_->RegisterComponent(*this);
+  }
 
   void UpdateAggregator() {
+    values_.last_aggregated_ = std::chrono::system_clock::now();
```
salieri11: I would have an internal method to have monotonic wall time, e.g. by having 2 internal variables that are set when the program starts (base wall time and base monotonic time) and then just have currentWallTime = baseWallTime + (currentMonotonicTime - baseMonotonicTime). Does it sounds reasonable? In this way we can have readable monotonic time.

andrewjstone: Thanks for the review Igor!

I'm not sure that implementation makes sense for our use case. This timestamp is primarily, ( possibly only ever ), going to be used by humans to make sense of the stats they are looking at in real time. The value as proposed will tend to drift over time from the wall clock time, and so it may be confusing for someone looking at the latest time and seeing that it's in the future.

However, if some system is going to be retrieving and recording these responses, monotonic time does make sense. But really what it such a system will want is just to know that one aggregation occurred after another. That also helps with testing, as we can wait for an update that way without checking the values completely. 

How about if we add just a uint64 counter that we increment each time along with the timestamp and always return both as a pair?

djvoth:
Time zones and daylight time shifts are a big can of worms.  If the need for currentWallTime is reasonably short (seconds/minutes) followed by a process exit, then we’ll probably get away with this straight forward approach.  But if processes run for hours/day/more and BaseWallTime is not updated properly (and if functions using currentWallTime expect monotonically increasing values) there will be serious problems a couple times per year…

Duane

andrewjstone:
After some discussion I'm closing this for the time being as it's not immediately needed. We'll come back to timestamps later if and when we need them.

PR #74, opened 2019-04-12
Add a UDP python client

andrewjstone:
A basic python client was created mainly for system testing.

It uses [trio](https://github.com/python-trio/trio) for networking and
concurrency.

Client requests and replies are implemented in msgs.py and match those
found in ClientMsgs.hpp. Tests for message serialization/deserialization
exist in test_msgs.py. Test can be run with `python3 test_msgs.py`

The client implements a synchronous api similar to simpleTest. The retry
strategy is very naive and retries every configured retry interval. The
code is documented inline describing the strategy.

Tests for the client utilize the simpleTest server. There are 4 main
tests:
 * Ensure a timeout (trio.TooSlowError) occurs when the servers are not
   up.
 * Ensure that a written value can be read when the servers are up
 * Ensure that when servers are brought up after the client request is
   sent a retry will occur and the request will succeed.
 * Ensure that we send to the primary on writes after it is learned

Tests can be run by
    cd concord-bft/build
    cmake ..
    ctest -R pyclient_tests --verbose

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build/*
 *.swp
 tags
 compile_commands.json
+*__pycache__
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,8 @@ function(add_submodule path)
     endif()
 endfunction()
 
+include(CTest)
+
 #
 # Subdirectories
 #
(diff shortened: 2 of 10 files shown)


yuliasherman:
I'm not familiar with python enough, but LGFM.

andrewjstone:
Hi @guyg8 

Thanks for the review! I pushed up changes relating to your feedback. I rebased the first commit with fixes for 2,3, and 4. The second commit contains the sequence number change and is large enough, and self contained enough to be it's own logical commit with an explanatory commit message.

PR #75, opened 2019-04-17
Collect metrics inside ReplicaImp

andrewjstone:
ReplicaImp now contains a metrics Component and registers some metrics.
The metrics get updated in both ReplicImp.cpp and
ControllerWithSimpleHistory.cpp.

Because the Component requires an aggregator, but an aggregator is not
yet instantiated for the bftengine as a whole, we create a dummy
aggregator, but never actually update it. To avoid having to change the
Replica constructor we also create a method that allows setting a
different aggregator and forward that onto the component in the Metrics
code so that it can register the component with the new Aggregator.

Since the metrics themselves are references into the component they must
be class members that are part of initialization. They will therefore
always be updated and part of ReplicaImp. The plan is to make the
feature optional by not always allowing the aggregated data to be
retrieved. In other words the to be built metric server will not always
be enabled. The overhead of tracking these metrics is very low, so this
seems like a reasonable tradeoff. Otherwise, the implementation of the
Metrics library itself may need to be changed.

A future commit will contain an optional UDP metric server which also contains the
centralized instantiated Aggregator. It will also contain code that
calls metricsComponent_.UpdateAggregator() periodically on some timer
tick.

### DIFF
--- a/bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
+++ b/bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
@@ -25,18 +25,28 @@ namespace bftEngine
 		static const float maxUpdateInTimeToStartSlowPath = 0.20F;
 
 
-		ControllerWithSimpleHistory::ControllerWithSimpleHistory(uint16_t C, uint16_t F, ReplicaId replicaId, ViewNum initialView, SeqNum initialSeq)
-			: onlyOptimisticFast(C == 0),
-			c(C), f(F), numOfReplicas(3 * F + 2 * C + 1), myId(replicaId), recentActivity(1, nullptr)
-		{
-			currentFirstPath = ControllerWithSimpleHistory_debugInitialFirstPath;
-			currentView = initialView;
-			isPrimary = ((currentView % numOfReplicas) == myId);
+		ControllerWithSimpleHistory::ControllerWithSimpleHistory(
+                    uint16_t C,
+                    uint16_t F,
+                    ReplicaId replicaId,
+                    ViewNum initialView,
+                    SeqNum initialSeq,
+                    concordMetrics::Component &metrics) :
+                  onlyOptimisticFast(C == 0),
+                  c(C), f(F), numOfReplicas(3 * F + 2 * C + 1), myId(replicaId),
+                  recentActivity(1, nullptr),
+                  currentFirstPath{ControllerWithSimpleHistory_debugInitialFirstPath},
+                  currentView{initialView},
+                  isPrimary{((currentView % numOfReplicas) == myId)},
+                  currentTimeToStartSlowPathMilli{defaultTimeToStartSlowPathMilli},
+                  metrics_first_path_{metrics.RegisterStatus(
+                      "firstCommitPath",
+                      CommitPathToStr(currentFirstPath))}
+                {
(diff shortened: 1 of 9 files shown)


guyg8:
> Yay, metrics! Now that `scripts/format-code.sh` exists, should we run it on this change before merging?

We want to use format-code.sh. However, we have to carefully check all the updates. We intend to do so at a later stage

bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
```diff
@@ -143,10 +153,14 @@ namespace bftEngine
...
 
-					if (cyclesWithFullCooperation >= (factorForFastOptimisticPath * EvaluationPeriod))
+					if (cyclesWithFullCooperation >= (factorForFastOptimisticPath * EvaluationPeriod)) {
 						currentFirstPath = CommitPath::OPTIMISTIC_FAST;
-					else if (!onlyOptimisticFast && (cyclesWithFullCooperation + cyclesWithPartialCooperation >= (factorForFastPath * EvaluationPeriod)))
+                                                metrics_first_path_.Get().Set(CommitPathToStr(currentFirstPath));
+                                        }
+					else if (!onlyOptimisticFast && (cyclesWithFullCooperation + cyclesWithPartialCooperation >= (factorForFastPath * EvaluationPeriod))) {
 						currentFirstPath = CommitPath::FAST_WITH_THRESHOLD;
+                                                metrics_first_path_.Get().Set(CommitPathToStr(currentFirstPath));
```
guyg8: The “Controllers” are objects may be passed as arguments to the replicas. We currently assume that their member functions do not have any implicit side effect. 
Is it possible to do the work outside the Controller object?

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -236,6 +234,27 @@ namespace bftEngine
...
+                        typedef concordMetrics::Component::Handle<
+                          concordMetrics::Counter> CounterHandle;
+
+                        GaugeHandle metric_view_;
+                        GaugeHandle metric_last_stable_seq_num__;
+                        GaugeHandle metric_last_executed_seq_num_;
+                        GaugeHandle metric_last_agreed_view_;
+
+                        CounterHandle metric_slow_path_count_;
+                        CounterHandle metric_received_msgs_;
```
yuliasherman: We probably need to split metric_received_msgs_ to different message kinds (i.e. reads/writes).

bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
andrewjstone: Ok. I fixed this, and kept the overhead low, by returning a bool indicating when the firstCommitPath is changed on return from `onNewSeqNumberExecution`. When that happens the replica retrieves the firstCommitPath from the executor and sets the metric.

bftengine/src/bftengine/ReplicaImp.hpp
andrewjstone: I changed this to include a separate counter for each external message.

andrewjstone:
Thanks for the Reviews Guy and Yulia! I made fixes corresponding to your requests.

PR #76, opened 2019-04-18
Metrics: Add JSON serialization support

andrewjstone:
Metrics: Add JSON serialization support

A simple hand written JSON serializer was implemented to serialize an
Aggregator and its corresponding components and metrics. There is no
deserializer as deserialization is expected to be done elsewhere as part
of the metric server clients. Since serialization is so simple here, we
decided we didn't need to pull in any dependencies and could hand roll
it.

Tests can be run via:
    ctest -R metric --verbose

The JSON output generated by the test as pretty printed through json_pp
looks like this:

         {
            "Components" : [
               {
                  "Gauges" : {
                     "total_peers" : 4,
                     "connected_peers" : 3
                  },
                  "Statuses" : {
                     "state" : "primary",
                     "commit_path" : "SLOW"
                  },
                  "Counters" : {
                     "messages_received" : 1,
                     "messages_sent" : 0
                  },
                  "Name" : "replica"
               },
               {
                  "Gauges" : {
                     "blocks-remaining" : 4
                  },
                  "Statuses" : {
                     "state" : "sending-blocks"
                  },
                  "Name" : "state-transfer",
                  "Counters" : {}
               }
            ]
         }

### DIFF
--- a/util/src/Metrics.cpp
+++ b/util/src/Metrics.cpp
@@ -104,4 +104,84 @@ Counter Aggregator::GetCounter(const string& component_name,
                    component.values_.counters_);
 }
 
+// Generate a JSON string of all aggregated components. To save space we don't
+// add any newline characters.
+std::string Aggregator::ToJson() {
+  ostringstream oss;
+  std::lock_guard<std::mutex> lock(lock_);
+
+  // Add the object opening
+  oss << "{\"Components\":[";
+
+  // Add all the components
+  for (auto it = components_.begin(); it != components_.end(); ++it) {
+    // Add a comma between every component
+    if (it != components_.begin()) {
+      oss << ",";
+    }
+    oss << it->second.ToJson();
+  }
+
+  // Add the object end
+  oss << "]}";
+
+  return oss.str();
+}
+
+// Generate a JSON string of the component. To save space we don't add any
+// newline characters.
+std::string Component::ToJson() {
+  ostringstream oss;
+
+  // Add the object opening and component name
+  oss << "{\"Name\":\"" << name_ << "\",";
+
+  // Add any gauges
+  oss << "\"Gauges\":{";
+
+  for (int i = 0; i < names_.gauge_names_.size(); i++) {
+    if (i != 0) {
+      oss << ",";
+    }
+    oss << "\"" << names_.gauge_names_[i] << "\":"
+      << values_.gauges_[i].Get() << "";
+  }
+
+  // End gauges
+  oss << "},";
+
+  // Add any status
+  oss << "\"Statuses\":{";
+
+  for (int i = 0; i < names_.status_names_.size(); i++) {
+    if (i != 0) {
+      oss << ",";
+    }
+    oss << "\"" << names_.status_names_[i] << "\":"
+      << "\"" << values_.statuses_[i].Get() << "\"";
+  }
+
+  // End status
+  oss << "},";
+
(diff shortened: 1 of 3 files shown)


util/test/metric_test.cpp
```diff
@@ -69,4 +71,32 @@ TEST(MetricsTest, Aggregator) {
...
+
+  Component c2("state-transfer", aggregator);
+  c2.RegisterGauge("blocks-remaining", 4);
+  c2.RegisterStatus("state", "sending-blocks");
+  c2.Register();
+
+  ostringstream oss;
+  oss << "echo '" << aggregator->ToJson() << "' | json_pp" << endl;
+
+  ASSERT_EQ(0, system(oss.str().c_str()));
```
beerriot: It took me a second to understand what was being tested here. Maybe add a tiny comment right above this, noting that json_pp will exit non-zero if it can't parse the input?

util/src/Metrics.cpp
```diff
@@ -104,4 +104,84 @@ Counter Aggregator::GetCounter(const string& component_name,
...
+
+  // Add any gauges
+  oss << "\"Gauges\":[";
+
+  for (int i = 0; i < names_.gauge_names_.size(); i++) {
+    if (i != 0) {
+      oss << ",";
+    }
+    oss << "{\"" << names_.gauge_names_[i] << "\":"
+      << values_.gauges_[i].Get() << "}";
```
beerriot: What's the reasoning behind a list of one-element maps, instead of just a map? That is, why `"Gauges":[{"alpha": 1},{"beta": 2},{"gamma":3}]` instead of `"Gauges":{"alpha":1,"beta":2,"gamma":3}`? Is it because the names might conflict? Or is this a format used by some other common tool?

util/src/Metrics.cpp
```diff
@@ -104,4 +104,84 @@ Counter Aggregator::GetCounter(const string& component_name,
                    component.values_.counters_);
 }
 
+// Generate a JSON string of all aggregated components. To save space we don't
+// add any newline characters.
```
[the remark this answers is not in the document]
andrewjstone: I was documenting it so that we don't add spaces in the string that we send over the wire. For example, if we pretty printed there would be spaces and newlines which would take up space in our 64k budget.

util/src/Metrics.cpp
```diff
@@ -104,4 +104,84 @@ Counter Aggregator::GetCounter(const string& component_name,
...
+std::string Aggregator::ToJson() {
+  ostringstream oss;
+  std::lock_guard<std::mutex> lock(lock_);
+
+  // Add the object opening
+  oss << "{\"Components\":[";
+
+  // Add all the components
+  for (auto it = components_.begin(); it != components_.end(); ++it) {
+    // Add a comma after every component
```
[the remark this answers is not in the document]
andrewjstone: I like it. I also did not like the wording.

util/test/metric_test.cpp
andrewjstone: As Bryan and I discussed offline, I'm going to change this test to eval python3 so that we don't rely on an extra program like json_pp.

util/src/Metrics.cpp
andrewjstone: No reason. Will change.

andrewjstone:
Updated to reflect Bryan's comments. TFTR!

PR #77, opened 2019-04-19
Add a UDP based metrics server

andrewjstone:
The server responds with aggregated metrics in JSON format. A test
runs a server with no metrics and ensures it responds correctly for
correct requests, and with an error response for incorrect requests.

This server is not integrated into concord-bft yet. The aggregator that
is part of the server will be used as the aggregator for the concord-bft
library. A future commit will integrate the server as an optional
feature.

This change also fixes the cmake logic to properly use logging as an
include only library.

### DIFF
--- a/util/src/MetricsServer.cpp
+++ b/util/src/MetricsServer.cpp
@@ -1,0 +1,144 @@
+// Concord
+//
+// Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the
+// LICENSE file.
+
+#include <string.h>
+#include <unistd.h>
+#include <iostream>
+
+#include "MetricsServer.hpp"
+
+// The MetricServer only handles a single type of request, that requests a JSON
+// metrics blob. The message format is just a single 8-bit 0 on the wire. We can
+// always change the protocol later if needed.
+const uint8_t kRequest = 0;
+
+// The MetricsServer only returns a single type of response. The response is
+// just an 8-bit 1 on the wire followed by a JSON string containing all metrics
+// for all components. Since we are using UDP, the entire message will always be
+// included, so no need to worry about framing. We can always change the
+// protocol if we decide to enhance the Metric server later on or move to a
+// different transport.
+const uint8_t kReply = 1;
+const uint8_t kError = 2;
+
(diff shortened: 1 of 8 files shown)


andrewjstone:
This code is ready to be reviewed and will merge cleanly. Tests pass.

andrewjstone:
Thanks for the review Bryan. We discussed using the regular protocol, but the rest of the team wanted to keep this separate for security and optionality reasons. And just a cleaner separation. 

I 100% agree with you about JSON over UDP being a weird choice given the size limits of UDP. However, it's the most expedient and will probably serve us well in our testing for quite a while. Ideally we'd end up moving this interface to HTTP 1.1 which would allow us to hit it with curl and from a browser. But bringing in such a heavy dependency like a web server was not in the cards right now. I have plans for how we may want to structure the code in the future to do such things outside of concord-bft, thus keeping the core code tight with limited deps but allowing us full reign for production usage and flexibility elsewhere. Lots of discussion still remains to be had there though.

util/src/MetricsServer.cpp
```diff
@@ -0,0 +1,144 @@
...
+// for all components. Since we are using UDP, the entire message will always be
+// included, so no need to worry about framing. We can always change the
+// protocol if we decide to enhance the Metric server later on or move to a
+// different transport.
+const uint8_t kReply = 1;
+const uint8_t kError = 2;
+
+namespace concordMetrics {
+
+void Server::Start() {
```
salieri11: Here I can suggest to check if the server is already running

util/src/MetricsServer.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+  running_lock_.lock();
+  running_ = true;
+  running_lock_.unlock();
+
+  auto recvThread = std::thread(&Server::RecvLoop, &*this);
+  std::swap(thread_, recvThread);
+}
+
+void Server::Stop() {
```
salieri11: I would check the running_ here as well, although less critical since the return value of close() is not used

util/src/MetricsServer.cpp
andrewjstone: If the server is already running the bind will fail with `FATAL 2019-04-29 08:48:14.746 (metrics-server) Error binding UDP socket: IP=0, Port=4376, errno=Address already in use`

util/src/MetricsServer.cpp
andrewjstone: The `thread_.join` call will fail if there is no running thread like so: `libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: thread::join failed: Invalid argument`

A bigger problem is calling join twice which may deadlock. But `running_` will already be set to false so this check won't help. In general this thread is supposed to run for the lifetime of the program and only be stopped on shutdown. It's relatively unlikely for a bug like this to pop up.

PR #78, opened 2019-04-23
Bugfix: Fix overflow of counter in SimpleClient

andrewjstone:
When the SimpleClient counter in a request sequence number hits its
maximum value of 0x3FFFFF, the time part of the sequence number
increases by 1 millisecond. However, the counter part was not getting
reset to 0, but instead was incremented also. If this scenario occurred the
an assert would fire and exit the program. The code was fixed to match
the invariant in the assert.

### DIFF
--- a/bftengine/include/bftengine/SimpleClient.hpp
+++ b/bftengine/include/bftengine/SimpleClient.hpp
@@ -48,14 +48,22 @@ namespace bftEngine
 		virtual int sendRequestToReadLatestSeqNum(uint64_t timeoutMilli, uint64_t& outLatestReqSeqNum) = 0;		
 	};
 
-	class SeqNumberGeneratorForClientRequests // users are allowed to generate their own sequence numbers (they do not have to use this class)
-	{
-	public:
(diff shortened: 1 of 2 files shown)


guyg8:
Thanks for fixing this bug.

I suggest we print a warning (to the log) when lastCountOfUniqueFetchID == 0x3FFFFF (we don't expect lastCountOfUniqueFetchID == 0x3FFFFF in normal situations).

We should also explain that SeqNumberGeneratorForClientRequests was mainly designed for testing and simple client applications. 
I assume that some client applications will prefer to use their own approach. Examples:
(1) a simple counter + store the last counter in a persistent storage
(2) an approach that utilizes the functions SimpleClient::sendRequestToResetSeqNum(), SimpleClient::sendRequestToReadLatestSeqNum(..) [These functions are not yet supported. But if necessary, we will support them.]

andrewjstone:
Thanks for the review Guy. I formatted the SeqNumberGeneratorForClientRequests class and added in a more explanatory comment in line with your review request. I also added a warning as suggested.

guyg8:
Thanks!

PR #79, opened 2019-04-25
 Add a python metrics client for testing

andrewjstone:
A simple UDP python client was created for testing. It uses trio so that
we can run concurrent requests during testing and it interroperates with
the bft client.

Additionally, client types were moved into config.py and test_client was
modified to reset its working directory so as not to break tests that
run after it.

## Important: Only review the second commit. This builds upon the work in #77, which must be merged first.


andrewjstone:
I closed for now because I want to make some changes to the retry and timeout strategies.

andrewjstone:
Timeouts and retries moved to the test code rather than metrics_client code. Ready for review now.

andrewjstone:
Guess I don't have permission to re-open this (sigh). #80 opened instead.

PR #81, opened 2019-04-26
Run the UDP Metrics Server for SimpleKVBC Replicas

andrewjstone:
The Replica abstract base class (along with ReplicaInternal and
ReplicaImp) now has an additional method, `SetAggregator` that allows
establishing the metrics aggregator to use for a given replica. This
aggregator is a member of the MetricsServer class which runs in main for
each SimpleKVBC replica. To allow configuring the SimpleKVBC replica,
the aggregator was added as a parameter to the constructor.

As part of this change the CMakeLists.txt for simpleKVBC was cleaned up
so that it no longer has to include a bunch of relative include
directories for bftEngine, threshsign, and logging. It now just uses
`target_link_libraries` and gets them all for free. This is the modern
cmake approach.

Additionally, a `using namespace SimpleKVBC` directive was removed from
simpleKVBCTests.h as using directives should not be in header files.
This using directive was moved to the cpp file instead.

### DIFF
--- a/bftengine/src/bftengine/BFTEngine.cpp
+++ b/bftengine/src/bftengine/BFTEngine.cpp
@@ -25,6 +25,8 @@ namespace bftEngine
 
 			virtual void stop() override;
 
+                        virtual void SetAggregator(std::shared_ptr<concordMetrics::Aggregator> a) override;
+
 			ReplicaImp* rep;
 		};
 
@@ -55,6 +57,12 @@ namespace bftEngine
 		{
 			return rep->stop();
 		}
+
+                void ReplicaInternal::SetAggregator(std::shared_ptr<concordMetrics::Aggregator> a)
+                {
+                        return rep->SetAggregator(a);
+                }
+
 	}
 }
 
--- a/bftengine/CMakeLists.txt
+++ b/bftengine/CMakeLists.txt
@@ -97,6 +97,7 @@ if(${USE_LOG4CPP})
     target_link_libraries(corebft PUBLIC log4cplus)
 endif()
 
+target_include_directories(corebft PUBLIC include/)
 target_include_directories(corebft PUBLIC include/bftengine)
 target_include_directories(corebft PUBLIC include/communication)
 target_include_directories(corebft PUBLIC include/bcstatetransfer)
--- a/bftengine/include/bftengine/Replica.hpp
+++ b/bftengine/include/bftengine/Replica.hpp
@@ -1,27 +1,37 @@
-//Concord
+// Concord
 //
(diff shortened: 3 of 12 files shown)


bftengine/src/bftengine/BFTEngine.cpp
```diff
@@ -1,93 +1,44 @@
...
+// file.
 
 #include "Replica.hpp"
 #include "ReplicaImp.hpp"
 
-namespace bftEngine
-{
-	namespace impl
-	{
-		struct ReplicaInternal : public Replica
```
guyg8: This trivial class/wrapper was deliberately created to separate the implementation from the external interface. This class simplifies the code. It is used to encapsulate some implementations details (for example, sequence number are represented by different types). It also shows that we don’t recommend to downcast Replica to ReplicaImp. At some point, it may also help to split ReplicaImp into several classes/modules.

bftengine/src/bftengine/PrimitiveTypes.hpp
```diff
@@ -26,6 +26,7 @@ namespace bftEngine
 	{
 
 
+                // TODO (AJS) - Why are SeqNum and ViewNum int64_t and not uint64?
 		typedef int64_t SeqNum;
```
guyg8: Internally we want to use SeqNum in signed expressions – e.g.,  s1-s2>0. It is possible to use uint64_t (to earn a bit), but we will have to check all expressions with SeqNum.

andrewjstone: Got it. Thanks.

bftengine/src/bftengine/BFTEngine.cpp
andrewjstone: I understand that may have been the intention. However, practically speaking, this separation makes it difficult to add new functionality to the ReplicaImp and expose it. In order to add a system wide aggregator, I need a way to add that aggregator to the ReplicaImp. I can add another parameter to `Replica::createNewReplica`, but that forces all users to care about the aggregator, even though we are only using it inside tests right now. 

Can you please suggest a mechanism for adding the aggregator functionality to the ReplicaImp that abides by your implementation choices?

andrewjstone: Nevermind, I suppose I can just add the method to ReplicaInternal. That will work fine.

andrewjstone:
I updated the code based on Guy's review. TFTR!

andrewjstone:
Rebased so that only the relevant commit shows up, since the others were merged.

PR #82, opened 2019-05-01
Fix incorrect connectivity status update

salieri11:
When the node is not being identified as replica node and not the client node, the optional status info callback is not called.


andrewjstone:
LGTM. Thanks!

PR #83, opened 2019-05-01
General Cleanup and bugfixes for testing

andrewjstone:
This change consists of 3 small commits and each is easier to review independent of the other.

The first commit namespaces the python client code for use in system tests.
We are going to be using this code outside of this directory for system
tests. It makes sense to not use names like 'Config' and 'Replica' that
might be common when used with a module like `client` or `config`
that is also common. Namespace the modules with `bft_` prefixes.

The second commit here fixes a bug in SimpleKVBCTests and cleans it up a bit. See the commit message for more detail.

The third commit fixes an interoperability bug between python and C++ by packing message structs.


bftengine/tests/simpleKVBCTests/TesterReplica/main.cpp
```diff
@@ -70,12 +62,9 @@ int main(int argc, char **argv) {
 	rp.replicaId = UINT16_MAX;
 
 	// allows to attach debugger
-	if(rp.debug)
```
salieri11: Why did you remove the shutdown cleanup code? I remember that when I used the library with TCP it may not release the socket and then the consecutive run will fail when binding to the port.

bftengine/tests/simpleKVBCTests/simpleKVBCTests.cpp
```diff
@@ -114,7 +110,7 @@ namespace BasicRandomTests
 		struct SimpleConditionalWriteHeader
 		{
 			SimpleRequestHeader h; // TODO: this is ugly ....
-			BlockId readVerion;
```
yuliasherman: We do a duplicate job...

bftengine/tests/simpleKVBCTests/simpleKVBCTests.cpp
```diff
@@ -967,7 +925,7 @@ namespace BasicRandomTests
 //				assert(0); 
 //			}				
 
-			Internal::CHECK(equiv, "actual reply != expected reply");
+			CHECK(equiv, "actual reply != expected reply");
```
yuliasherman: I've got rid of CHECK macro in my simpleKVBTests.

bftengine/tests/simpleKVBCTests/TesterReplica/main.cpp
```diff
@@ -48,14 +48,6 @@ ReplicaParams rp;
...
 		concordlogger::Logger::getLogger("skvbctest.replica");
 
-void signalHandler( int signum ) {
-	if(r)
-		r->stop();
-
-	LOG_INFO(replicaLogger, "replica " << rp.replicaId << " stopped");
-	exit(0);
-}
-
```
yuliasherman: I think we do support a graceful replica shutdown, so I think it's important to allow it shutting down nicely. 
I did see errors during replica stop, but not recursive ones. In my code, I just removed signal(SIGKILL, signalHandler), as a non-catchable.

bftengine/tests/simpleKVBCTests/TesterReplica/main.cpp
andrewjstone: The call to r->stop was throwing an exception inside the signal handler. It looked like the asserts in the tests caused memory to be freed during unwind that then were used during stop. Guy said we didn't support graceful shutdown right now and that he didn't think we needed the signal handler in the test so I removed the code rather than digging into that specific error at this time. We still need to look into it, but since we are going to be making some changes to interfaces, it may be easier to do that then. 

I didn't realize the part about the socket. That's likely due to TIME_WAIT state. We can likely get rid of that issue by just setting SO_REUSEADDR on the socket.

bftengine/tests/simpleKVBCTests/simpleKVBCTests.cpp
andrewjstone: Yep. We'll discuss tomorrow as mentioned offline.

andrewjstone:
> I wanted to make all the structures that pass between modules packed.
> Why do you pack only one of them?

They all get packed. The macro closes later with `#pragma pack(pop)`. I wish I knew you already fixed that. It would have saved me some time debugging. More reason for dedup discussion.

bftengine/tests/simpleKVBCTests/TesterReplica/main.cpp
salieri11: OK, if it works - I am fine

salieri11:
Approved

PR #85, opened 2019-05-05
Fix incorrect connection attempt timeout for TLS module

salieri11:
This commit adresses incorrect behaviour of the connection timeout management handling in the TLS module. When the outgoing connection is created and async_connect is started,
the deadline timer is started in parallel and has race condition with the connect_complete callback. We noticed on the slow machines that there may be a case when time ticks before the connection is opened, but connection is already open when we try to reset the socket for another attempt.

The fix is to start the timer after failed connection attempt and to delay the next connection attempt. Moreover, there were few places in code where the handlers where binded to 'this' raw pointer instead of using 'shared_from_this()'. This is fixed as well.


bftengine/include/communication/CommDefs.hpp
```diff
@@ -217,7 +218,7 @@ class TlsTCPCommunication : public ICommunication {
   virtual ~TlsTCPCommunication();
  private:
   class TlsTcpImpl;
-  TlsTcpImpl *_ptrImpl = nullptr;
+  std::shared_ptr<TlsTcpImpl> _ptrImpl = nullptr;
```
toly-kournik: std::unique_ptr<TlsTcpImpl> probably better suits semantically

bftengine/src/communication/TlsTCPCommunication.cpp
```diff
@@ -1297,11 +1228,11 @@ class TlsTCPCommunication::TlsTcpImpl {
             _cipherSuite);
 
     conn->connect(peerIp, peerPort);
-    LOG_DEBUG(_logger, "connect called for node " << _selfId);
+    LOG_INFO(_logger, "connect called for node " << _selfId << ", dest: " << nodeId);
   }
 
  public:
-  static TlsTcpImpl *create(NodeNum selfNodeId,
+  static std::shared_ptr<TlsTcpImpl> create(NodeNum selfNodeId,
```
toly-kournik: same as above std::unique_ptr

bftengine/src/communication/TlsTCPCommunication.cpp
```diff
@@ -1310,28 +1241,55 @@ class TlsTCPCommunication::TlsTcpImpl {
                             string certRootFolder,
                             string cipherSuite,
                             UPDATE_CONNECTIVITY_FN statusCallback) {
-    return new TlsTcpImpl(selfNodeId,
+    return std::shared_ptr<TlsTcpImpl>(new TlsTcpImpl(selfNodeId,
```
toly-kournik: same as above std::unique_ptr

PR #87, opened 2019-05-07
Add a sequence number to MetricsServer

andrewjstone:
We want to be able to ignore stale replies during tests, so change the
protocol used by the MetricServer to include a sequence number.

Also change the bft_metrics_client to only operate against a single
replica. This allows easier tracking of sequence numbers and ensures we
don't have to buffer replies in any way.

All tests were updated to reflect the new code. Tests pass.

    andrewstone@ubuntu:~/concord-bft/build$ ctest -R '(pyclient|metric)'
    Test project /home/andrewstone/concord-bft/build
        Start  9: pyclient_tests
    1/3 Test  #9: pyclient_tests ...................   Passed    1.46 sec
        Start 10: metric_tests
    2/3 Test #10: metric_tests .....................   Passed    0.01 sec
        Start 11: metric_server_tests
    3/3 Test #11: metric_server_tests ..............   Passed    0.18 sec

    100% tests passed, 0 tests failed out of 3

    Total Test time (real) =   1.66 sec

### DIFF
--- a/util/pyclient/bft_metrics_client.py
+++ b/util/pyclient/bft_metrics_client.py
@@ -13,16 +13,18 @@
 # This code requires python 3.5 or later
 import trio
 import json
+import struct
 
 from bft_config import Replica
 
-MAX_MSG_SIZE = 64*1024; # 64k
+REQUEST_TYPE = 0
+REPLY_TYPE = 1
+ERROR_TYPE = 2
 
-def req():
-    """Return a get request to the metrics server"""
-    req = bytearray()
-    req.append(0)
-    return req
+HEADER_FMT = "<BQ"
+HEADER_SIZE = struct.calcsize(HEADER_FMT)
+
+MAX_MSG_SIZE = 64*1024; # 64k
 
 class MetricsClient:
     def __enter__(self):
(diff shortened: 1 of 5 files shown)


util/pyclient/bft_metrics_client.py
```diff
@@ -33,22 +35,30 @@ def __exit__(self, *args):
...
         trio.fail_after as necessary`.
         """
-        await self.sock.sendto(req(), self.replicas[replica])
-        reply, _ = await self.sock.recvfrom(MAX_MSG_SIZE)
-        assert 1 == reply[0]
-        return json.loads(reply[1:])
+        destination = (self.replica.ip,self.replica.port)
+        await self.sock.sendto(self._req(), destination)
+        while True:
+            reply, _ = await self.sock.recvfrom(MAX_MSG_SIZE)
```
salieri11: Do we want to have timeout here?

andrewjstone: Good catch. But, no, I purposefully didn't put one here. I originally did have a timeout in a prior commit, but I removed it when writing some test code, because it hindered how I wanted to do retries. I wanted retries in the test code (via trio.move_on_after) but not in the metrics client, so I put the timeouts in the test code, and this call in the block that does the trio timeouts. It's a new way of thinking about timeouts, even for me. They aren't attached per operation, but to the surrounding block. It does seem to make some code cleaner, and you don't have to calculate time remaining for overall timeouts which is nice. We may even end up removing them from the udp client to let the caller configure the timeouts in the future, but for now I saw no need to make things more confusing there, especially since retries are more complex.

salieri11: got it.

PR #89, opened 2019-05-09
 Add automated system test framework in python

andrewjstone:
Add an automated system test framework in the test directory, along with
the first automated system test, `test_state_transfer`. Please see
`test/README.md` for more details about the design of the framework and
how to use it.

This test infrastructure relies on retrieving metrics from replicas
running concord-bft with an application specific state machine. This
commit also ensures that the aggregator in the metric server gets
updated periodically by adding a new timer to ReplicaImp. This ensures
the tests get valid metrics when using bft_metrics_client.

Additionally, the number of client proxies was increased in the
TesterReplica so that we could use more concurrent clients without
requiring a configuration file. A bug was also fixed such that the
configured number was used instead of the replicaParams version.

`statusReportTimerMillisec` was also shortened significantly in the
SimpleKVBCTests code so that nodes can discover each other more quickly
after a failure or partition.

This test builds upon #87 , and only the second commit should be reviewed. It should be merged after #87.


test/bft_tester.py
```diff
@@ -0,0 +1,252 @@
...
+        self.alpha = [i for i in range(65, 91)]
+        self.alphanum = [i for i in range(48, 58)]
+        self.alphanum.extend(self.alpha)
+        self.keys = self._create_keys()
+        os.chdir(self.testdir)
+        self._generate_crypto_keys()
+        self.clients = {}
+        self.metrics = None
+
+    def _generate_crypto_keys(self):
```
salieri11: I would add TLS certificates creation here as well.

salieri11:
LGTM! One comment and +1 to merge

test/bft_tester.py
andrewjstone: We only support UDP transport at the moment for these tests. When we add support for TLS I'll add in the key generation.

guyg8:
Please resolve the conflicts with the master branch

andrewjstone:
I rebased the code in this PR and pushed the updated code. Due to timing differences, I was able to frequently see crashes of SimpleTest replicas during pyclient/test_client test runs. This was due to a bug where metrics aggregation wasn't setup properly in ReplicaImp. I fixed the bug and also removed the signalHandler as done in SimpleKVBCTests TesterReplica, since it doesn't work correctly and causes the same "terminate called recursively" error to fill the screen for page after page when an uncaught exception occurs. Those changes are the second and third commits in this PR respectively. Tests pass repeatedly now.

bftengine/tests/simpleKVBCTests/TesterReplica/main.cpp
```diff
@@ -153,8 +157,10 @@ int main(int argc, char **argv) {
 	c.replicaId = rp.replicaId;
 	c.fVal = replicaConfig.fVal;
 	c.cVal = replicaConfig.cVal;
-	c.numOfClientProxies = rp.numOfClients;
-	c.statusReportTimerMillisec = 20 * 1000;
+	c.numOfClientProxies = replicaConfig.numOfClientProxies;
+        // Allow triggering of things like state transfer to occur faster in
+        // tests.
+	c.statusReportTimerMillisec = 1000;
```
guyg8: Sometimes we want to check that the replicas can continue working without sending frequent Status messages (it is mainly important when we use many replicas). I suggest using a small value only for tests that really need it.

andrewjstone: I changed the code to take a CLI parameter for this and added it as another commit.

PR #91, opened 2019-05-12
Add optional performance test

salieri11:
This PR adds optional performance measurements to the Python version of SimpleTest.
When enabled, each client measures raw latencies for each request. When the test finishes, histogram files are created for each client in the output folder. Moreover, the test prints the average duration and the weighted average of throughput during the test.
Use -p parameter to enable performance measurement when using Python version of SimpleTest


andrewjstone:
LGTM! Thanks.

PR #92, opened 2019-05-15
Multiple threading refactorings

toly-kournik:
Three unrelated commits:
1. Use c++11 thread library support in SimpleThreadPool and SimpleAutoResetEvent.
Move the above classes to util subdirectory and change the namspace accordingly.
Remove obsolete Threading.h (2 copies)
Add multithreading unittests.

2. Portable way of compiling relic-related libraries. Trivial: CMakeLists.txt

3. Cleanup: Remove unused code from Crypto.hpp, tabs -> spaces, indentation issues. Trivial.


bftengine/src/communication/PlainUDPCommunication.cpp
```diff
@@ -340,24 +336,10 @@ class PlainUDPCommunication::PlainUdpImpl {
   void
   stopRecvThread() {
     //LOG_DEBUG("Stopping the receiving thread..");
-    threadJoin(recvThreadRef);
+    recvThreadRef->join();
```
salieri11: recvThreadRef can be nullptr or the wrapped thread can be non joinable, if the Start() method was not called. I suggest to check for the pointer validity and thread being joinable.

PR #98, opened 2019-05-19
Persistency and crash recovery: interfaces, debug, and related functionality

guyg8:
- Add interface (pure abstract class) PersistentStorage – this class should be used to save/load data to/from the persistent storage. Its implementation should access the persistent storage by using the MetadataStorage interface.

- Add methods that save/load messages to/from buffers.

- Add methods that return the maximum size of messages.

- Add class DebugPersistentStorage: this class is an implementation of PersistentStorage. It was designed to help debugging the code that uses PersistentStorage.

- Add a new constant maxNumOfRequestsInBatch  (in SysConsts.hpp) that represents the maximum number of requests in a batch (in a single Pre-Prepare message). It is needed to implement part of the crash recovery in ReplicaImp.

- Add a new constant maxSizeOfCombinedsignature (in SysConsts.hpp) that represents the maximum size of a combined signature. This is needed to determine the maximum size of several messages with signatures.

- In ViewsManager: add methods that return the ViewChange and NewView messages that have been used to enter the current active view.

### DIFF
--- a/bftengine/src/bftengine/Bitmap.hpp
+++ b/bftengine/src/bftengine/Bitmap.hpp
@@ -1,0 +1,146 @@
+// Concord
+//
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include "PrimitiveTypes.hpp"
+#include "assertUtils.hpp"
+
+#include <vector>
+#include <cstring>
+
+namespace bftEngine {
+namespace impl {
+class Bitmap {
+ public:
+  Bitmap() : numBits_(0), p_(nullptr) {}
+
+  Bitmap(uint32_t numBits) : numBits_(numBits), p_(nullptr) {
+    if (numBits_ > 0) {
+      p_ = (unsigned char*)std::malloc(realSize());
+      zeroAll();
+    }
+  }
+
+  Bitmap(const Bitmap& other) : numBits_(other.numBits_), p_(nullptr) {
+    if (numBits_ == 0) {
+      p_ = nullptr;
+    } else {
+      p_ = (unsigned char*)std::malloc(realSize());
+      std::memcpy(p_, other.p_, realSize());
+    }
+  }
+
+  ~Bitmap() {
+    if (numBits_ > 0) {
+      Assert(p_ != nullptr);
+      std::free(p_);
+    }
+  }
+
+  Bitmap& operator=(const Bitmap& other) {
+    if (numBits_ > 0) {
+      Assert(p_ != nullptr);
+      std::free(p_);
+    }
+
+    numBits_ = other.numBits_;
+    if (numBits_ == 0) {
+      p_ = nullptr;
+    } else {
+      p_ = (unsigned char*)std::malloc(realSize());
+      std::memcpy(p_, other.p_, realSize());
+    }
+
+    return *this;
+  }
+
+  void zeroAll() {
+    if (p_ == nullptr) return;
+    const uint32_t s = realSize();
+    Assert(s > 0);
+    memset((void*)p_, 0, s);
+  }
+
+  uint32_t numOfBits() const { return numBits_; }
+
+  bool get(uint32_t i) const {
+    Assert(i < numBits_);
+    const uint32_t byteIndex = i / 8;
+    const unsigned char byteMask = (1 << (i % 8));
+    return ((p_[byteIndex] & byteMask) != 0);
+  }
+
+  void set(uint32_t i) {
+    Assert(i < numBits_);
+    const uint32_t byteIndex = i / 8;
+    const unsigned char byteMask = (1 << (i % 8));
+    p_[byteIndex] = p_[byteIndex] | byteMask;
+  }
+
+  void reset(uint32_t i) {
+    Assert(i < numBits_);
+    const uint32_t byteIndex = i / 8;
+    const unsigned char byteMask = ((unsigned char)0xFF) & ~(1 << (i % 8));
+    p_[byteIndex] = p_[byteIndex] & byteMask;
+  }
+
+  void writeToBuffer(char* buffer,
+                     uint32_t bufferLength,
+                     uint32_t* actualSize) const {
+    const uint32_t sizeNeeded = sizeNeededInBuffer();
+    Assert(bufferLength >= sizeNeeded);
+    uint32_t* pNumOfBits = (uint32_t*)buffer;
+    char* pBitmap = buffer + sizeof(uint32_t);
+    *pNumOfBits = numBits_;
+    memcpy(pBitmap, p_, realSize());
+    if (actualSize) *actualSize = sizeNeeded;
+  }
+
+  uint32_t sizeNeededInBuffer() const {
+    return (sizeof(uint32_t) + realSize(numBits_));
+  }
+
+  static Bitmap* createBitmapFromBuffer(char* buffer,
+                                        uint32_t bufferLength,
+                                        uint32_t* actualSize) {
+    if (actualSize) *actualSize = 0;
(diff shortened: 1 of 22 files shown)


andrewjstone:
Who reviewed this?

guyg8:
All pull requests should be peer reviewed.
This PR was reviewed by @salieri11

yuliasherman:
It feels like windows manipulation functionality does not really relate to PersistentStorage class. Consider separating it to another one.

bftengine/src/bftengine/DebugPersistentStorage.cpp
```diff
@@ -0,0 +1,531 @@
...
+      checkWindow{0, nullptr}
+
+{}
+
+uint8_t DebugPersistentStorage::beginWriteTran() {
+  return ++numOfNestedTransactions;
+}
+
+uint8_t DebugPersistentStorage::endWriteTran() {
+  Assert(numOfNestedTransactions != 0);
```
yuliasherman: Looks like condition is the opposite to one is needed.

bftengine/src/bftengine/PersistentStorage.hpp
```diff
@@ -0,0 +1,204 @@
...
+  virtual uint8_t endWriteTran() = 0;
+
+  // return true IFF write-only transactions are running now
+  virtual bool isInWriteTran() const = 0;
+
+  //////////////////////////////////////////////////////////////////////////
+  // Update methods (should only be used in write-only transactions)
+  //////////////////////////////////////////////////////////////////////////
+
+  virtual void setReplicaConfig(ReplicaConfig config) = 0;
```
yuliasherman: Is there any reason to pass this structure by value?

bftengine/src/bftengine/Bitmap.hpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+    if (actualSize) *actualSize = sizeNeeded;
+    return b;
+  }
+
+  static uint32_t maxSizeNeededToStoreInBuffer(uint32_t maxNumOfBits) {
+    return (sizeof(uint32_t) + realSize(maxNumOfBits));
+  }
+
+ protected:
```
andrewjstone: IMO these should default to private. Attempting to maintain invariants across inherited classes on protected fields when both the base and subclass can modify them is error prone in my experience.

bftengine/src/bftengine/MessageBase.cpp
```diff
@@ -119,5 +119,76 @@ namespace bftEngine
 			msgSize_ = size;
 		}
 
+		MessageBase* MessageBase::cloneObjAndMsg() const
```
andrewjstone: I know this is not possible in this PR but we should rework the code so we aren't passing around raw pointers for messages.

bftengine/src/bftengine/ViewsManager.cpp
```diff
@@ -272,6 +272,42 @@ NewViewMsg* ViewsManager::getMyNewViewMsgForCurrentView() {
   return r;
 }
 
+vector<ViewChangeMsg*> ViewsManager::getViewChangeMsgsForCurrentView()
```
andrewjstone: Should probably return a vector of unique_ptrs so the caller doesn't have to free them.

bftengine/src/bftengine/ViewsManager.cpp
```diff
@@ -272,6 +272,42 @@ NewViewMsg* ViewsManager::getMyNewViewMsgForCurrentView() {
...
+{
+	Assert(stat == Stat::IN_VIEW);
+
+	const uint16_t SMAJOR = (2 * F + 2 * C + 1);
```
andrewjstone: Nit: Quorum is clearer than SMAJOR. What is the S here?

bftengine/src/bftengine/PersistentStorage.hpp
```diff
@@ -0,0 +1,204 @@
...
+// crashed or restarted, this interface is used to restore the concord-bft 
+// state from the persistent storage. 
+// PersistentStorage is designed to only store data elements that are essential 
+// to reconstruct the internal in-memory data structures of concord-bft.
+// For simplicity and efficiency, some of the internal data elements are not 
+// written to the persistent storage (for example, messages with partial 
+// threshold signatures are not stored in persistent storage; if needed such 
+// messages are re-created and retransmitted by the replicas).
+
+class PersistentStorage {
```
andrewjstone: This pure virtual class has 44 methods. That means that every implementor has re-implement all of them to provide persistency. That seems to be a lot of effort if we want to port our persistency to different storage systems. Is there a way to change the structure of this code so that there is more abstraction and less specificity in terms of methods, perhaps by using data structs and/or templates?

For instance, the transaction methods make sense for a storage layer, but the type of data being stored is not abstracted here. I haven't gone through a full analysis, but it feels like all the getters and setters can perhaps be replaced with a single method each that take a type and return an opaque blob that can be decoded by a higher level. That decoding can be implemented once in a concrete type.

bftengine/src/bftengine/DebugPersistentStorage.cpp
yuliasherman: Sorry, it's a wrong place.

guyg7:
> It feels like windows manipulation functionality does not really relate to PersistentStorage class. Consider separating it to another one.

PersistentStorage represents the persistency layer from the point the view of the main replica code (ReplicaImp). If it helps (e.g., if it simplifies the implementation of PersistentStorage), we should consider implementing the window functions as a separate class. However, in order to simplify the code in ReplicaImp, I still prefer to use a single interface for the persistency layer (PersistentStorage or something similar).

bftengine/src/bftengine/PersistentStorage.hpp
guyg7: For ReplicaConfig , it doesn't really matter.
However, I agree that it will be better to update:
 virtual void setReplicaConfig(const ReplicaConfig& config) = 0;

bftengine/src/bftengine/Bitmap.hpp
```diff
@@ -0,0 +1,146 @@
...
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include "PrimitiveTypes.hpp"
+#include "assertUtils.hpp"
+
+#include <vector>
```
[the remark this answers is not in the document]
guyg7: Correct. Thanks.

bftengine/src/bftengine/Bitmap.hpp
```diff
@@ -0,0 +1,146 @@
...
+    *pNumOfBits = numBits_;
+    memcpy(pBitmap, p_, realSize());
+    if (actualSize) *actualSize = sizeNeeded;
+  }
+
+  uint32_t sizeNeededInBuffer() const {
+    return (sizeof(uint32_t) + realSize(numBits_));
+  }
+
+  static Bitmap* createBitmapFromBuffer(char* buffer,
```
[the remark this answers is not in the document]
guyg7: We should always understand the lifetime of our dynamic objects.
We are considering to start using smart pointers (at least for message objects). The main work will be in other classes/modules of the BFT engine (it is trivial to add smart pointers to classes like Bitmap and MessageBase).
There's no point in adding smart pointers to the simple classes before we update the (relatively tricky) code that uses these classes.

bftengine/src/bftengine/PersistentStorage.hpp
guyg7: The implementation of PersistentStorage should be based on the MetadataStorage interface. If we want to use a new storage system X, we just need to implement MetadataStorage for X.

The code that uses PersistentStorage  is relatively subtle and complex. In order to simplify it, we preferred to create a single interface with specific methods (PersistentStorage  should contain the methods that are required by the main code of “persistency and crash recovery”).

andrewjstone:
Ok. Thanks for cla

> > It feels like windows manipulation functionality does not really relate to PersistentStorage class. Consider separating it to another one.
> 
> PersistentStorage represents the persistency layer from the point the view of the main replica code (ReplicaImp). If it helps (e.g., if it simplifies the implementation of PersistentStorage), we should consider implementing the window functions as a separate class. However, in order to simplify the code in ReplicaImp, I still prefer to use a single interface for the persistency layer (PersistentStorage or something similar).

Ok. Thanks for clarifying Guy. That makes sense.

PR #100, opened 2019-05-23
initialize PeerConnectivityStatus structs

beerriot:
Running under valgrind, I see warnings about use of uninitialized
variables that point to the peerPort field of a PeerConnectivityStatus
struct. This field was not set in every place the struct was created
in the communication modules. This change forces default
initialization, to make sure this memory is initialized.

@salieri11 I'm actually not sure this is the correct solution, but it makes valgrind happy. Should peerPort and peerIp be set in all of these places instead? (See VB-931)


salieri11:
Thanks for catching this! LGTM!

beerriot:
Further investigation showed that even when we're setting peerPort in PeerConnectivityStatus, we're sometimes using a value (`_port`) that is not yet initialized. The additional commit in this PR addresses that.

PR #101, opened 2019-05-31
Add support for testing concurrent operations

andrewjstone:
A new component to the python testing framework, SkvbcTracker, has been
added that is capable of tracking requests and replies for messages sent
to a cluster of SimpleKVBCTests TesterReplica nodes. Tests can use this
component to record requests and replies and then verify that all
returned replies can be placed in a total order that makes sense. In
other words all the operations are linearizable. The tool itself makes
no promises about finding all possible violations, although it will not
provide any false positives.

The SkvbcTracker lives in skvbc_linearizability.py, and significant
documentation for it lives in the doc comments, particularly the doc
comment for the SkvbcTracker class. Reviewers should read and
understand that comment prior to proceeding with review. Tests in
test_skvbc_linearizability show how to use the tracker and should
provide more clarity.

Additionally, many specific exceptions were created to indicate
linearizability violations. These live in skvbc_exceptions.py. Other
exceptions for skvbc testing may be added there in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/concord-bft/101)

### DIFF
--- a/test/bft_test_exceptions.py
+++ b/test/bft_test_exceptions.py
@@ -14,14 +14,29 @@ class Error(Exception):
     """Base class for exceptions in this module."""
     pass
 
+##
+## Exceptions for bft_tester
+##
 class AlreadyRunningError(Error):
     def __init__(self, replica):
         self.replica = replica
+        j
+    def __repr__(self):
+        return (f'{self.__class__.__name__}:\n'
+           f'  replica={self.replica}\n')
 
 class AlreadyStoppedError(Error):
     def __init__(self, replica):
         self.replica = replica
 
+    def __repr__(self):
+        return (f'{self.__class__.__name__}:\n'
+           f'  replica={self.replica}\n')
+
 class BadReplyError(Error):
     def __init__(self):
         pass
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}\n'
+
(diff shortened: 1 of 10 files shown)


andrewjstone:
I rebased on master and added a few bugfixes and enhancements to the testing framework. There are 3 separate small commits with individual commit messages. It probably isn't necessary to look at them individually when doing the review, as they just enhance what was already there.

The bugs in the linearizability checker were found as part writing a concurrent system test and running it against a healthy cluster with 3 different configurations. That test is coming in a separate PR.

test/bft_test_exceptions.py
```diff
@@ -14,14 +14,29 @@ class Error(Exception):
     """Base class for exceptions in this module."""
     pass
 
+##
+## Exceptions for bft_tester
+##
 class AlreadyRunningError(Error):
     def __init__(self, replica):
         self.replica = replica
+        j
```
beerriot: Is this `j` a typo?

andrewjstone: Yes indeed! Good catch. That would have been a fun bug during a test run if that exception was actually raised.

andrewjstone: I pushed up another small commit to fix this an another bug.

PR #102, opened 2019-06-02
Logging re-factoring

toly-kournik:
1. replace threshold library's logger by a common logger defined in logging:
                loginfo << ...  => LOG_INFO(...)
        2. move all logging building logics to logging/CMakeLists.txt
        3. simplify internal logger implementation
        4. fix printf warnings (-Wformat2)
        5. default log4cplus initialization for internal use
        6. portable way for finding relic-related 3rd-party libraries.


logging/FindLog4cplus.cmake
```diff
@@ -0,0 +1,84 @@
+
+#--------------------------------------------------------------------------------
+# Copyright (c) 2012-2013, Lars Baehren <lbaehren@gmail.com>
```
guyg8: Please make sure we are allowed to use this file

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -34,11 +34,12 @@ using std::chrono::milliseconds;
...
 using std::chrono::system_clock;
 
-#define Assert(expr) {                                             \
-    if((expr) != true) {                                                    \
-        STLogger.error("'%s' is NOT true (in function '%s' in %s:%d)\n", \
-            #expr, __FUNCTION__, __FILE__, __LINE__); assert(false);                \
-    }                                                                       \
+#define Assert(expr) {                                                          \
+    if((expr) != true) {                                                        \
+        LOG_ERROR_F(STLogger, "'%s' is NOT true (in function '%s' in %s:%d)\n", \
```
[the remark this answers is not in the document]
toly-kournik: The PR was about replacing the Log.h used in threshsign library by a common concord-bft logger (which also supports log4cplus).
All the logging macros were remained as is.
I agree that *_F macros are less convenient, less c++ and probably sub-optimal.
We can gradually replace them like we do with formatting.

logging/FindLog4cplus.cmake
toly-kournik: I'm replacing this file by cmake code written entirely by me :-)

bftengine/src/bcstatetransfer/BCStateTran.cpp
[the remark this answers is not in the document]
salieri11: These macros were left for backward compatibility with some BFT engine code. I am +1 to remove them

logging/FindLog4cplus.cmake
toly-kournik: done

ibalajiarun:
So, how does one compile threshsign with the new logger now?  I am getting the following error when I try to compile threshsign alone

```
CMake Error at CMakeLists.txt:98 (add_executable):
  Error evaluating generator expression:

    $<TARGET_OBJECTS:logging_dev>

  Objects of target "logging_dev" referenced but no such target exists.
Call Stack (most recent call first):
  test/CMakeLists.txt:14 (add_relic_executable)
```

PR #103, opened 2019-06-03
Fix status callback in TLS comm module

salieri11:
This PR fixes inconsistency of viewing peer status info when using TLS comm module.
When the upper level application provides the callback for connectivity status update, sometimes the IP and port are not set yet, e.g. for incoming connections. In this case the view will be different between calls.
The fix is to create the map of nodes at the time when the comm module is initialised and when messages are sent/received or there is an error, to update the status only.


andrewjstone:
LGTM!

PR #104, opened 2019-06-03
Add a msg for getting blocks from SimpleKVBCTests

andrewjstone:
In order to simplify concurrency testing, we need a mechanism for
retrieving block data from SimpleKVBCTests TesterReplica clusters. This
is used to fill in information about blocks that were created by
requests that did not receive responses in the test. This strategy
eliminates the need to do an exponential cost search to find possible
linearizable histories of written blocks that satisfy reads and other
writes, by just filling in the missing block directly.

A test was added to show that the command works.

It implements VB-966.

It will be used in conjunction with #101 to write concurrent system
tests that contend for keys.

### DIFF
--- a/bftengine/tests/simpleKVBCTests/simpleKVBCTests.cpp
+++ b/bftengine/tests/simpleKVBCTests/simpleKVBCTests.cpp
@@ -97,7 +97,9 @@ namespace BasicRandomTests
 
 		struct SimpleRequestHeader
 		{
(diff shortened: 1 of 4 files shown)


guyg8:
Can we merge it before #101 ?

andrewjstone:
> Can we merge it before #101 ?

Yes. It's independent.

PR #107, opened 2019-06-10
Add a viewChangeTimeout CLI param to Skvbc Replica

andrewjstone:
Add a `-v` CLI parameter that allows changing the viewChangeTimeout for
a SimpleKVBCTests TesterReplica. Setting this timeout implicitly enables
view change in the replicas.

Also, add an error message that occurs when a block can't be retrieved
from roStorage.

This is independent of any other change and can be merged directly. Tests pass.

### DIFF
--- a/bftengine/tests/simpleKVBCTests/TesterReplica/main.cpp
+++ b/bftengine/tests/simpleKVBCTests/TesterReplica/main.cpp
@@ -60,6 +60,8 @@ int main(int argc, char **argv) {
   logConfig.configure();
 #endif
 	rp.replicaId = UINT16_MAX;
+        rp.viewChangeEnabled = false;
+        rp.viewChangeTimeout = 45*1000;
 
 	// allows to attach debugger
 	if(rp.debug) {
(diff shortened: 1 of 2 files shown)


salieri11:
Lgtm.

guyg8:
In order to support small view change timeouts, it is required to change several other timeouts (e.g., the slow path timeout, and the status timeouts). For simplicity, in the current version, I suggest that the initial view change timeout will be >= 30 seconds.

andrewjstone:
That length of timeout is not suitable to fast automated testing, where you want to induce multiple view changes. 

I will create a separate PR to make the other timeouts configurable or dependent on top level timeouts.

ittaiab:
Lets indeed put the other timeouts to be configurable or dependent on top level timeouts.

Can we also have some place where we document *all* the timeout parameters and discuss how they are supposed to work.

andrewjstone:
> Lets indeed put the other timeouts to be configurable or dependent on top level timeouts.
> 
> Can we also have some place where we document _all_ the timeout parameters and discuss how they are supposed to work.

Any configurable parameters should have doc comments in the config structs at a minimum. Other than that though, we should also document all of them in a section on timeouts in the (coming soon) developer's guide.

PR #109, opened 2019-06-22
Concurrent test wreak havoc

andrewjstone:
This is based on #101, #107 and #106. Those *must* be merged in first. Only the last commit in this PR is relevant.

This is the beginning of a concurrent test that introduces crash faults.
It is the most basic possible version in that it only crashes the
primary once. Periodic crashes and other faults will be introduced in
future versions.

With this test I can induce endless view changes approximately 1/4 of the time I run it. I doubt that this is an actual bug in the view change logic itself, and more likely a result of internal timers that need to be reduced as a result of the overall view change timeout change used in the test. I reset the view change timeout from 45 seconds to 3 seconds in this test. I only include the configuration that causes problems, by commenting out the rest in the `interesting_configs` function. I am debugging the root cause and will issue a patch when I figure it out.

### DIFF
--- a/test/test_skvbc_chaos.py
+++ b/test/test_skvbc_chaos.py
@@ -14,6 +14,7 @@ import unittest
 import trio
 import os.path
 import random
+import time
 
 import bft_tester
 import skvbc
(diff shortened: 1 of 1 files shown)


guyg8:
Please resolve the conflicts here.
See also a related  comment in #107

ittaiab:
please resolve conflicts

andrewjstone:
Rebased and pushed.

PR #110, opened 2019-06-28
WIP: Format IncomingMsgsStorage.(cpp|hpp)

andrewjstone:
**This PR is to be merged in after persistency work, so that merge conflicts can be resolved later**

This is in preparation for another commit which makes this file take
unique_ptrs instead of raw pointers. I formatted the code to make the
changes easier, and this will allow an easier diff for the actual coming
changes.

The coming commit is the first step in using smart pointers in the
replica code path.

### DIFF
--- a/bftengine/src/bftengine/IncomingMsgsStorage.cpp
+++ b/bftengine/src/bftengine/IncomingMsgsStorage.cpp
@@ -1,146 +1,148 @@
-//Concord
+// Concord
 //
-//Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
 //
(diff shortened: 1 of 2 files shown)


andrewjstone:
Status changed to WIP, so that it can be rebased after persistency work lands.

andrewjstone:
Thanks for the review @beerriot!. I moved all the comments above their respective code lines and reformatted as you suggested. I have noticed this in the past, and it does indeed make things look a lot better.

salieri11:
+1 to merge

robem:
👍 for the change!
Did you use `clang-format` or is it hand-formatted?

andrewjstone:
> 👍 for the change!
> Did you use `clang-format` or is it hand-formatted?

Both. I first moved all the comments above the relevant code, than ran clang-format with scripts/format-code.sh. Finally I gave it a once over to make sure it looked clean, then ran tests.

PR #111, opened 2019-06-28
Use unique_ptrs in InternalMsgsStorage

andrewjstone:
InternalMsgsStorage is used for transferring internal and external
messages across threads. This makes the queues use unique_ptrs instead
of raw pointers so that we don't have to keep track of ownership and
lifetimes cross thread. All callers and users were changed to reflect
the new types.

Note that this change causes no changes in the safety of the existing
code. It is being used an introduction to using unique_ptrs to indicate
transfer of ownership across function and data structure boundaries.
Future work will remove the need for manual deletes as the use of smart
pointers is continued.

This was stripped out of a much larger change that attempted to convert
all messages to being contained in unique_ptrs. That change was huge and
effectively unreviewable, so a piecemeal attempt is being made instead.

### DIFF
--- a/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
+++ b/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
@@ -434,14 +434,14 @@ namespace bftEngine
 						verifier->release(accWithVer);
 
 						// send internal message with the results
-						InternalMessage* iMsg = ExternalFunc::createInterCombinedSigFailed(context, expectedSeqNumber, expectedView, replicasWithBadSigs);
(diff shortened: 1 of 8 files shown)


guyg8:
Hi Andrew,
I suggest to wait with these pull requests.
IMO, we should first merge the work on persistence (these PRs have many conflicts with  our work on persistency and crash recovery).

--Guy

andrewjstone:
Status changed to WIP, so that it can be rebased after persistency work lands.

andrewjstone:
Rebased on latest master as conflicting persistence changes are in. Conflicts were minimal. Please take another look.

PR #119, opened 2019-07-10
Add initial version of design document.

andrewjstone:
This document is intended to cover the overall structure and
implementation of the code as it currently exists. This initial commit
covers just the public interfaces to the bftengine. Future commits will
cover more details such as state transfer, bftengine / protocol
implementation, timeouts, persistence, etc...

It's possible that we'll want to turn this from a single document into a
developer's guide with different "chapters". For now, let's keep it
simple though.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -1,0 +1,118 @@
+# Overview
+
+Concord-bft is an implementation of the [SBFT
+Protocol](https://arxiv.org/abs/1804.01626). SBFT implements State Machine
+Replication (SMR) in the byzantine model. Concord-bft is a generic protocol,
+with pluggable interfaces for communication, storage, state transfer, and
+abstract state machine execution. This document covers the implementation
+of concord-bft. When the implementation differs from the paper, it will be
+mentioned explicitly.
+
+It is important to note that this document covers the code as it exists at the
+time of writing. When the code changes substantially this document should be
+updated.
+
+# Public Interfaces
+
+Concord-bft is a library meant to be used by application code. The external user
+API for the core engine is described by a [set of structs and abstract
+interfaces](../bftengine/include/bftengine). The file names are pretty self
+explanatory. Important structures will be mentioned explicitly below.
+
+All user interfaces live in the `bftEngine` namespace.
+
+# Clients
+A default client,
+[SimpleClient](../bftengine/include/bftengine/SimpleClient.hpp), is included
+with concord-bft. It is just one manner of implementing a client, and is
+implemented in C++ utilizing the `ICommunication` and `IReceiver` interfaces
+below. Additional clients can be implemented by utilizing the client messages
+described in the next subsection. In fact, we have a [client in
+Python](../util/pyclient/bft_client.py) that we use for testing that does just
+that.
+
+## Client Messages
+Client request and reply messages are prefixed with a `ClientRequestMsgHeader` or
+`ClientReplyMsgHeader` respectively. Each header, like all other messages in
+concord-bft is implemented as a packed C struct, meaning that there are no padding
+bytes between struct fields. This is accomplished via enclosing the structure
+definitions between `#pragma pack(push,1)` and `#pragma pack(pop)` lines.
+Packing ensures that the size of the structure is the sum of the size of the
+fields, and allows them to be sent over the wire and interpreted
+deterministically, given both machines have the same endianess.
+
+## Communication
+
+The `ICommunication` interface is used by the bftengine code to send
+asynchronous messages across the network. It is implemented for different
+transport layers such as UDP, TCP, and TLS in
+[bftengine/src/communication](../bftengine/src/communication).
+
+The `setReceiver` method is called by the bftengine code to let implementers of
+the `ICommunication` interface know where received messages should be handled.
+Handlers for received messages implement the `IReceiver` interface. The two sole
+implementers of `IReceiver` are the [ReplicaImp](../bftengine/src/bftengine/ReplicaImp.hpp)
+and the [SimpleClient](../bftengine/src/bftengine/SimpleClient.cpp)
+
+## State transfer
+
+State transfer refers to the mechanism via which out of date replicas can
+retrieve already committed data efficiently and bring themselves in sync with
+the rest of the cluster. State transfer is done outside of the ordering and
+replication protocol performed by SBFT, and can often be achieved by data transfer
+between only two replicas.
+
+As state transfer is outside of the byzantine agreement protocol, it must be
+built as a function of the instantiation of the specific state machine
+implementation. Concord-bft treats state machine operations and data to be
+replicated/executed upon commit as opaque. Since the data being stored by the
+application is opaque, and can be arbitrarily large, concord-bft cannot provide
+a universally efficient mechanism for transferring such state, and relies on
+state machine specific implementations for optimal distribution.
+
+For applications to implement state transfer, and have concord-bft properly be
+able to initiate state transfer when needed, and also know when there is more
+state that needs to be transferred, there must be hooks in both the application
+code and concord-bft code that allow interaction between the application and
+bftengine in a generic manner. This is the purpose of the state transfer
+interfaces in
+[IStateTransfer.hpp](../bftengine/include/bftengine/IStateTransfer.hpp).
+
+The `IStateTransfer` interface is implemented by the application developer for a given
+state transfer/checkpoint mechanism. It is called by the main replica thread of
+the core bftengine implementation. The concord-bft library contains two
+implementations of `IStateTransfer`:
+[SimpleStateTransfer](../bftengine/src/simplestatetransfer) and
+[BcStateTransfer](../bftengine/src/bcstatetransfer). These will be described
+further in the detailed design section for state transfer.
+
+The `IReplicaForStateTransfer` interface allows the state transfer
+implementations to call functions in the bftengine replica.
+
+## Metadata Storage
+
+The bftengine has internal metadata that must be persisted to disk. The
+[MetadataStorage](../bftengine/include/bftengine/MetadataStorage.hpp) interface
+should be implemented by the application developer for a given storage system
+such as RocksDB.
+
+## Replica
+
+The [Replica](../bftengine/include/bftengine/Replica.hpp) interface is the API
+used to instantiate a new replica given specific implementations of the other
+public interfaces.
+
+## RequestsHandler
+
+The [RequestsHandler](../bftengine/include/bftengine/Replica.hpp) interface
+contains a single callback function to be implemented by the application and
+called by the bftengine when an operation at a given sequence number has
+committed. This is the hook that allows arbitrary SMR implemented in concord-bft
+to result in execution of application specific operations at all replicas in a
+total order.
+
+# ReplicaConfig
+ReplicaConfig contains most configurable attributes of concord-bft and should be
+created by the application and passed into `Replica::createNewReplica(...)`.
+Note that other configurable attributes are set as part of the communication
+interface.


docs/design.md
```diff
@@ -0,0 +1,128 @@
...
+should be implemented by the application developer for a given storage system
+such as RocksDB.
+
+## Replica
+
+The [Replica](../bftengine/include/bftengine/Replica.hpp) interface is the API
+used to instantiate a new replica given specific implementations of the other
+public interfaces.
+
+## RequestsHandler
```
upshaw-alex: Is this intended to be the primary documentation of `RequestsHandler` and how to implement it, or is there other documentation? If this is the primary documentation of `RequestsHandler`, should we mention any requirements we have for its implementation here? (For example, IIRC `RequestsHandler` needs to be deterministic given the same request and preceding state on which there is consensus; meaning it can't be dependent on replica-specific state on which there is no consensus, such as the replica's system time).

docs/design.md
```diff
@@ -0,0 +1,128 @@
...
+
+The [RequestsHandler](../bftengine/include/bftengine/Replica.hpp) interface
+contains a single callback function to be implemented by the application and
+called by the bftengine when an operation at a given sequence number has
+committed. This is the hook that allows arbitrary SMR implemented in concord-bft
+to result in execution of application specific operations at all replicas in a
+total order.
+
+# ReplicaConfig
+ReplicaConfig contains all configurable attributes of concord-bft and should be
```
upshaw-alex: Don't we still also have the separate communication configuration, and, if so, should we weaken this statement about "all configurable attributes"?

upshaw-alex:
Thank you very much for working on enhancement of the documentation; I feel like it is a lacking area at the moment.

docs/design.md
```diff
@@ -0,0 +1,128 @@
+# Overview
+
+Concord-bft is an implementation of the [SBFT
+Protocol](https://arxiv.org/abs/1804.01626). SBFT implements State Machine
+Replication (SMR) in the byzantine model. It is a generic protocol, with
```
ittaiab:
```suggestion
Replication (SMR) in the byzantine model. Concord-bft is a generic protocol, with
```

docs/design.md
```diff
@@ -0,0 +1,128 @@
...
+pluggable interfaces for communication, storage, and state transfer. This
```
ittaiab: should we also mention the generic interface for an abstract state machine?

docs/design.md
```diff
@@ -0,0 +1,128 @@
...
+implement and understand.
+
+The drawbacks are:
+ * Endianness is not taken into consideration, so a struct sent from a big
+   endian machine will be incorrectly interpreted on a little endian machine,
+   and vice versa.
+ * Messages field access may be slow, as it may not be on an aligned boundary.
+ * Messages are fixed. New fields cannot be added to messages and remain
+   backwards compatible.
+
```
ittaiab: Perhaps this document should just describe the system without providing too much opinion. We can move the advantages and drawbacks to a different document that discusses design choices and roadmap.

docs/design.md
```diff
@@ -0,0 +1,128 @@
...
+## State transfer
+
+State transfer refers to the mechanism via which out of date replicas can
+retrieve already committed data efficiently and bring themselves in sync with
+the rest of the cluster. State transfer is done outside of the ordering and
+replication protocol performed by SBFT, and can often be achieved by data transfer
+between only two replicas.
+
+As state transfer is outside of the core SMR protocol, it may be implemented in
```
ittaiab: I am not sure saying its outside the core SMR is correct. Maybe say outside the Byzantine Agreement protocol?  Also I think we need to say that ST must be implemented as a function of the instantiation of the specific state machine implementation (the term application may be confusing in this context). This is related to my comment above about the discussion of Concord implementing the BASE approach of having an abstract state machine interface.

docs/design.md
```diff
@@ -0,0 +1,128 @@
...
+explanatory. Important structures will be mentioned explicitly below.
+
+All user interfaces live in the `bftEngine` namespace.
+
+# Clients
+A default client,
+[SimpleClient](../bftengine/include/bftengine/SimpleClient.hpp), is included
+with concord-bft. It is just one manner of implementing a client, and is
+implemented in C++ utilizing the `ICommunication` and `IReceiver` interfaces
+below. Additional clients can be implemented by utilizing the ClientMessages
```
beerriot: I don't see the string ClientMessage in the code base. Should this be two words?

docs/design.md
```diff
@@ -0,0 +1,128 @@
+# Overview
+
+Concord-bft is an implementation of the [SBFT
+Protocol](https://arxiv.org/abs/1804.01626). SBFT implements State Machine
+Replication (SMR) in the byzantine model. It is a generic protocol, with
+pluggable interfaces for communication, storage, and state transfer. This
+document covers the implementation of concord-bft. When the implementation
+differs from the paper, it will be mentioned explicitly.
```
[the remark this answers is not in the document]
andrewjstone: All documents about code are to some degree aspirational, and they asymptotically approach correctness.  Documentation can always get out of date, and we should always have the goal of keeping it in sync.

However, saying that "we should" do something in the docs doesn't really makes sense IMO. We should just do it. If it gets out of date we should fix it. This is an implied contract with the reader. This comment does bring up an important question though: "What can we do to prevent the docs from getting out of sync?"

We probably should have a checklist for pull requests of things to look for when updating the code and docs. This would also be useful to help with determining reviewers. Ittai and I specifically discussed that we want core protocol changes to be run by researchers, and having a checklist would also help with this. Ideally, we could make the checklist part of the PR signoff process. The checklist must be completed before any PR can be merged. I think GitHub might be able to help us with this functionality. If not we can just enforce it via convention in code reviews, by having the author put it in the PR description.

docs/design.md
andrewjstone: This is just an introduction section to the interfaces. We may choose to document it more later in a specific section.

docs/design.md
andrewjstone: Good point!

docs/design.md
andrewjstone: Good point!

docs/design.md
andrewjstone: Yes. Good catch!

andrewjstone:
@upshaw-alex @ittaiab @beerriot Thanks for the review! I updated the doc to reflect your comments. I pushed the changes as a second commit so that you could easily diff the changes.

@ittaiab **Please do not merge this in immediately**. Once re-reviewed, let me know it is approved and I will squash it to a single commit, so that we can have a clean history. Then it may be merged.

docs/design.md
andrewjstone: Updated as you described. Thanks!

robem:
Thanks for the write-up!

ittaiab:
Thank you Andrew! seems like a good initial version

andrewjstone:
@ittaiab Rebased and pushed. Ready for merge! Thanks for the review.

docs/design.md
```diff
@@ -0,0 +1,118 @@
+# Overview
+
+Concord-bft is an implementation of the [SBFT
```
tfidali: Concord-bft (byzantine fault tolerance)

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+Protocol](https://arxiv.org/abs/1804.01626). SBFT implements State Machine
+Replication (SMR) in the byzantine model. Concord-bft is a generic protocol,
+with pluggable interfaces for communication, storage, state transfer, and
+abstract state machine execution. This document covers the implementation
+of concord-bft. When the implementation differs from the paper, it will be
```
tfidali: from the SBFT Protocol paper,

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+mentioned explicitly.
+
+It is important to note that this document covers the code as it exists at the
```
tfidali: NOTE: This document covers the code as it exists at the time of writing. When the code changes substantially, this document will be updated to reflect the changes.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+time of writing. When the code changes substantially this document should be
+updated.
+
+# Public Interfaces
+
+Concord-bft is a library meant to be used by application code. The external user
```
tfidali: Concord-bft is a library used by the application code. The core engine API is described by a set of structs and abstract interfaces.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+API for the core engine is described by a [set of structs and abstract
+interfaces](../bftengine/include/bftengine). The file names are pretty self
```
tfidali: The file names are self-explanatory.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+explanatory. Important structures will be mentioned explicitly below.
```
tfidali: Important structures are mentioned below explicitly.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+
+All user interfaces live in the `bftEngine` namespace.
```
tfidali: All the user interface APIs are available in the 'bftEngine' namespace.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+
+# Clients
+A default client,
+[SimpleClient](../bftengine/include/bftengine/SimpleClient.hpp), is included
+with concord-bft. It is just one manner of implementing a client, and is
```
tfidali: This is one way of implementing a client. The client is implemented in C++ utilizing the ICommunication and IReceiver interfaces below.

tfidali: A default client, SimpleClient, is included in concord-bft.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+implemented in C++ utilizing the `ICommunication` and `IReceiver` interfaces
+below. Additional clients can be implemented by utilizing the client messages
```
tfidali: You can implement additional clients by utilizing the client messages described in the 'Client Messages' subsection. There is a client in Python used for testing client messages.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+described in the next subsection. In fact, we have a [client in
+Python](../util/pyclient/bft_client.py) that we use for testing that does just
+that.
+
+## Client Messages
+Client request and reply messages are prefixed with a `ClientRequestMsgHeader` or
+`ClientReplyMsgHeader` respectively. Each header, like all other messages in
```
tfidali: Each header in concord-bft has a packed C struct without padding bytes between the struct fields.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+concord-bft is implemented as a packed C struct, meaning that there are no padding
+bytes between struct fields. This is accomplished via enclosing the structure
```
tfidali: This padding encloses the structure definitions between #pragma pack(push,1) and #pragma pack(pop) lines.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+definitions between `#pragma pack(push,1)` and `#pragma pack(pop)` lines.
+Packing ensures that the size of the structure is the sum of the size of the
```
tfidali: Packing ensures that the size of the structure is the sum of the size of the fields. Packing also allows the information to be sent over the wire and interpreted deterministically, given both machines have the same endianess.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+fields, and allows them to be sent over the wire and interpreted
+deterministically, given both machines have the same endianess.
+
+## Communication
+
+The `ICommunication` interface is used by the bftengine code to send
```
tfidali: The bftengine code uses the 'ICommunication interface' to send asynchronous messages across the network.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+asynchronous messages across the network. It is implemented for different
+transport layers such as UDP, TCP, and TLS in
```
tfidali: Different transport layers such as UDP, TCP, and TLS in bftengine/src/communication use this interface.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+[bftengine/src/communication](../bftengine/src/communication).
+
+The `setReceiver` method is called by the bftengine code to let implementers of
+the `ICommunication` interface know where received messages should be handled.
+Handlers for received messages implement the `IReceiver` interface. The two sole
```
tfidali: Handlers of the received messages use the IReceiver interface.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+implementers of `IReceiver` are the [ReplicaImp](../bftengine/src/bftengine/ReplicaImp.hpp)
+and the [SimpleClient](../bftengine/src/bftengine/SimpleClient.cpp)
```
tfidali: The two sole implementers of the IReceiver interface are the ReplicaImp and the SimpleClient.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+
+## State transfer
+
+State transfer refers to the mechanism via which out of date replicas can
```
tfidali: State transfer allows out of date replicas to retrieve already committed data efficiently and sync with the rest of the cluster.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+retrieve already committed data efficiently and bring themselves in sync with
+the rest of the cluster. State transfer is done outside of the ordering and
```
tfidali: SBFT performs the state transfer process outside of the ordering and replication protocol. State transfer is achieved by data transfer between only two replicas.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+replication protocol performed by SBFT, and can often be achieved by data transfer
+between only two replicas.
+
+As state transfer is outside of the byzantine agreement protocol, it must be
```
tfidali: As state transfer is outside of the byzantine agreement protocol, it must be built as an instantiation function of the specific state machine implementation.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+built as a function of the instantiation of the specific state machine
+implementation. Concord-bft treats state machine operations and data to be
+replicated/executed upon commit as opaque. Since the data being stored by the
```
tfidali: This opaque data can be arbitrarily large. Concord-bft cannot provide a universally efficient mechanism for transferring such state and relies on state machine-specific implementations for optimal distribution.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+application is opaque, and can be arbitrarily large, concord-bft cannot provide
+a universally efficient mechanism for transferring such state, and relies on
+state machine specific implementations for optimal distribution.
+
+For applications to implement state transfer, and have concord-bft properly be
```
tfidali: For applications to implement state transfer, have concord-bft initiate any needed state transfer, and know when more state needs to be transferred, there must be hooks in both the application code and concord-bft code that allow generic interaction between the application and bftengine.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+able to initiate state transfer when needed, and also know when there is more
+state that needs to be transferred, there must be hooks in both the application
+code and concord-bft code that allow interaction between the application and
+bftengine in a generic manner. This is the purpose of the state transfer
+interfaces in
+[IStateTransfer.hpp](../bftengine/include/bftengine/IStateTransfer.hpp).
+
+The `IStateTransfer` interface is implemented by the application developer for a given
```
tfidali: Application developers use the IStateTransfer interface for a given state transfer/checkpoint mechanism.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+state transfer/checkpoint mechanism. It is called by the main replica thread of
```
tfidali: The interface is called by the main core bftengine replica thread.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+the core bftengine implementation. The concord-bft library contains two
+implementations of `IStateTransfer`:
+[SimpleStateTransfer](../bftengine/src/simplestatetransfer) and
+[BcStateTransfer](../bftengine/src/bcstatetransfer). These will be described
```
tfidali: These implementations will be described in the detailed state transfer design section.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+further in the detailed design section for state transfer.
+
+The `IReplicaForStateTransfer` interface allows the state transfer
+implementations to call functions in the bftengine replica.
+
+## Metadata Storage
+
+The bftengine has internal metadata that must be persisted to disk. The
+[MetadataStorage](../bftengine/include/bftengine/MetadataStorage.hpp) interface
```
tfidali: Application developers use the MetadataStorage interface for a storage system such as RocksDB.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+should be implemented by the application developer for a given storage system
+such as RocksDB.
+
+## Replica
+
+The [Replica](../bftengine/include/bftengine/Replica.hpp) interface is the API
```
tfidali: The Replica interface API instantiates a new replica given the specific implementations of other public interfaces.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+used to instantiate a new replica given specific implementations of the other
+public interfaces.
+
+## RequestsHandler
+
+The [RequestsHandler](../bftengine/include/bftengine/Replica.hpp) interface
```
tfidali: The RequestsHandler interface contains a single callback function to be implemented by the application. The bftengine uses the RequestsHandler interface when a given sequence number is committed.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+contains a single callback function to be implemented by the application and
+called by the bftengine when an operation at a given sequence number has
+committed. This is the hook that allows arbitrary SMR implemented in concord-bft
```
tfidali: This hook allows the arbitrary SMR implemented in concord-bft to execute the application-specific operations for all the replicas.

docs/design.md
```diff
@@ -0,0 +1,118 @@
...
+to result in execution of application specific operations at all replicas in a
+total order.
+
+# ReplicaConfig
+ReplicaConfig contains most configurable attributes of concord-bft and should be
```
tfidali: ReplicaConfig contains the most configurable attributes of concord-bft. This interface should be created by the application and passed into Replica::createNewReplica(...). 
NOTE: Other configurable attributes are set as part of the communication interface.

PR #122, opened 2019-07-14
Configurable message size

robem:
In order to access the replica configuration values from anywhere inside
concord-bft we can either pass around the ReplicaConfig structure, introduce a
global variable or create a singleton. The former isn't easy to do because it
would touch a lot of function signatures and hence would cause a lot of merge
conflicts for other developers. A singleton is pretty much a global variable
but with the guarantee of a consistent view. At the moment, the properties are
read-only but this is no blocker for future setters.
The main motivation is to make concord-bft properties configurable at start-up
time. Re-compiling every time we want to change a constant is simply not
practical. With the proposed approach, you can add a new config value to the
ReplicaConfig structure, add a getter in the singleton and use it wherever its
needed without passing around the ReplicaConfig structure.

- - -

**Review hint**

1st commit: Introduction of the singleton w/o changing existing logic
2nd commit: Replace constants with configurable values

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfigSingleton.hpp
+++ b/bftengine/include/bftengine/ReplicaConfigSingleton.hpp
@@ -1,0 +1,54 @@
+//Concord
+//
+//Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+//
+//This product is licensed to you under the Apache 2.0 license (the "License").  You may not use this product except in compliance with the Apache 2.0 License. 
+//
+//This product may include a number of subcomponents with separate copyright notices and license terms. Your use of these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE file.
+
+#pragma once
+
+#include <stdint.h>
+#include <set>
+#include <string>
+
+#include "ReplicaConfig.hpp"
+
+class IThresholdSigner;
+class IThresholdVerifier;
+
+namespace bftEngine {
+
+class ReplicaConfigSingleton {
+    public:
+	// Note, for initialization, this function is not thread-safe. The
+	// caller needs to make sure that only one thread at a time calls this
+	// function when a non-nullptr `config` parameter is supplied.
+	static ReplicaConfigSingleton& GetInstance(const ReplicaConfig* config=nullptr);
+
+	uint16_t GetFVal() const;
+	uint16_t GetCVal() const;
+	uint16_t GetReplicaId() const;
+	uint16_t GetNumOfClientProxies() const;
+	uint16_t GetStatusReportTimerMillisec() const;
+	uint16_t GetConcurrencyLevel() const;
+	bool GetAutoViewChangeEnabled() const;
+	uint16_t GetViewChangeTimerMillisec() const;
+	std::set<std::pair<uint16_t, std::string>> GetPublicKeysOfReplicas() const;
+	std::string GetReplicaPrivateKey() const;
+	IThresholdSigner const* GetThresholdSignerForExecution() const;
+	IThresholdVerifier const* GetThresholdVerifierForExecution() const;
+	IThresholdSigner const* GetThresholdSignerForSlowPathCommit() const;
+	IThresholdVerifier const* GetThresholdVerifierForSlowPathCommit() const;
+	IThresholdSigner const* GetThresholdSignerForCommit() const;
+	IThresholdVerifier const* GetThresholdVerifierForCommit() const;
+	IThresholdSigner const* GetThresholdSignerForOptimisticCommit() const;
+	IThresholdVerifier const* GetThresholdVerifierForOptimisticCommit() const;
+
+    private:
+	ReplicaConfigSingleton();
+	ReplicaConfig config_;
+	bool is_initialized_;
+};
+
+}  // namespace bftEngine
--- a/bftengine/include/bftengine/ReplicaConfigSingleton.hpp
+++ b/bftengine/include/bftengine/ReplicaConfigSingleton.hpp
@@ -44,6 +44,10 @@ class ReplicaConfigSingleton {
 	IThresholdVerifier const* GetThresholdVerifierForCommit() const;
 	IThresholdSigner const* GetThresholdSignerForOptimisticCommit() const;
 	IThresholdVerifier const* GetThresholdVerifierForOptimisticCommit() const;
+	uint32_t GetMaxExternalMessageSize() const;
+	uint32_t GetMaxReplyMessageSize() const;
+	uint32_t GetMaxNumOfReservedPages() const;
+	uint32_t GetSizeOfReservedPage() const;
 
     private:
 	ReplicaConfigSingleton();
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -144,7 +144,7 @@ BCStateTran::BCStateTran(
   :
   pedanticChecks_{ config.pedanticChecks },
   as_{ stateApi },
-  psd_{ createDataStore(persistentDataStore, kSizeOfReservedPage) },
+  psd_{ createDataStore(persistentDataStore, config.sizeOfReservedPage) },
   replicas_{ generateSetOfReplicas((3*config.fVal) + (2*config .cVal) + 1) },
   myId_{ config.myReplicaId },
   fVal_{ config.fVal },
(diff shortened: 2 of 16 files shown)


salieri11:
Great idea, @robem ! I would like to raise one issue with this proposal. We might want to have some cases, in future, where replicas may have different configuration (e.g. different timeouts etc). I suggest to do the change you propose in the following way: 1) redo the ReplicaConfig singleton and 2) create a wrapper, say ConfigurationManager with static getConfiguration method. In the wrapper you can decide to return any type of config struct, singleton or not. The wrapper will be exposed over the code. The wrapper itself maybe singleton instead of static class, although IMO singleton is creational pattern and using it as a global variable is somehow behavioural pattern, but I don't have any strong position here.

bftengine/include/bftengine/ReplicaConfigSingleton.hpp
```diff
@@ -0,0 +1,51 @@
...
+class IThresholdSigner;
+class IThresholdVerifier;
+
+namespace bftEngine {
+
+class ReplicaConfigSingleton {
+    public:
+	static ReplicaConfigSingleton& GetInstance(const ReplicaConfig* config=nullptr);
+
+	uint16_t GetFVal();
```
andrewjstone: All these getters should be const methods. Additionally, the methods that return pointers should return const* to prevent global mutation.

bftengine/src/bftengine/ClientReplyMsg.cpp
```diff
@@ -9,14 +9,15 @@
...
+#include "ReplicaConfigSingleton.hpp"
 
 namespace bftEngine
 {
 	namespace impl
 	{
 
 		ClientReplyMsg::ClientReplyMsg(ReplicaId primaryId, ReqId reqSeqNum, ReplicaId replicaId)
-			: MessageBase(replicaId, MsgCode::Reply, maxExternalMessageSize) 
+			: MessageBase(replicaId, MsgCode::Reply, bftEngine::ReplicaConfigSingleton::GetInstance().GetMaxExternalMessageSize())
```
andrewjstone: I don't believe you need the `bftEngine` namespace qualifier, since it's an enclosing namespace. I'm fine leaving it, but removing it would make these lines shorter. Singleton indicates that it's a global type anyway.

bftengine/src/bftengine/ReplicaConfigSingleton.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+class IThresholdSigner;
+class IThresholdVerifier;
+
+namespace bftEngine
+{
+
+ReplicaConfigSingleton::ReplicaConfigSingleton() : is_initialized_(false) {}
+
+ReplicaConfigSingleton& ReplicaConfigSingleton::GetInstance(const ReplicaConfig* config)
```
beerriot: Could you please add a comment, either here or in the header file, noting that this function is not thread-safe, so it is the caller's job to ensure that only a single thread calls this with a non-null pointer at a time?

robem:
> Great idea, @robem ! I would like to raise one issue with this proposal. We might want to have some cases, in future, where replicas may have different configuration (e.g. different timeouts etc). I suggest to do the change you propose in the following way: 1) redo the ReplicaConfig singleton and 2) create a wrapper, say ConfigurationManager with static getConfiguration method. In the wrapper you can decide to return any type of config struct, singleton or not. The wrapper will be exposed over the code. The wrapper itself maybe singleton instead of static class, although IMO singleton is creational pattern and using it as a global variable is somehow behavioural pattern, but I don't have any strong position here.

@salieri11 Thanks for your review! I thought about multiple replicas and a singleton as well but I assume we would run each replica (its SMR stack plus the db adapter) in a separate process anyways and hence a singleton should still work. But if we want to support multiple configurations per replica at runtime then the ConfigurationManager makes sense to me. However, AFAIK, we don't do multiple configs at runtime yet and hence I would defer the ConfigurationManager.

bftengine/src/bftengine/ReplicaConfigSingleton.cpp
robem: Thanks for the review!

bftengine/src/bftengine/ClientReplyMsg.cpp
robem: Thanks for the review!

salieri11:
I would not defer it right now. Besides being a part of any product, this library is used extensively for the research. Making core structs singleton in the library decreases its generic nature. The decision about providing this or that type of configuration should be outside the library or at least wrapped. It will pay off later IMO. Moreover, it implies that now there 2 semantically similar replica config structs - I believe we should not go there. This can be accepted as a temporary solution only if it blocks you but please mark it explicitly.

andrewjstone:
> I would not defer it right now. Besides being a part of any product, this library is used extensively for the research. Making core structs singleton in the library decreases its generic nature. The decision about providing this or that type of configuration should be outside the library or at least wrapped. It will pay off later IMO. Moreover, it implies that now there 2 semantically similar replica config structs - I believe we should not go there. This can be accepted as a temporary solution only if it blocks you but please mark it explicitly.

I'm really not following the concern here. A configuration type (ReplicaConfig) should be shared among all replicas. The values of that type can be different per replica. However, since each replica runs in a separate process there are no conflicts. Once a configuration is set to start a replica it is static for the lifetime of the replica process. The singleton just provides accessors to data that is already static. This is not any less generic than what we currently have, it just makes the configuration available inside the BFT library. The configuration is still chose and set by the code above by passing in a ReplicaConfig.

I'm not saying we should defer work on a configuration manager, because I'm not sure we should implement one at all. What explicit value does it provide?

salieri11:
Singleton is not a pattern for replacing global variables and having more then 2 replicas in the process is very realistic.

salieri11:
Discussed offline - agreed that this is temporary solution. +1 to merge

bftengine/src/bftengine/ReplicaConfigSingleton.cpp
robem: I ran `make test` and the simpleTest as Andrew pointed out below.

upshaw-alex:
> LGTM! Thanks Rob.
> 
> If you didn't do it already, please make sure our basic tests pass. Run simpleTest as described in the README and `cd build && make test`

I am sure you already know this, it would be more reliable if we had some form of CI that could run automatically and reject merge requests that break any tests we have rather than relying on people to manually run the tests.

upshaw-alex:
For whatever it is worth, I might throw out that the solution I was considering was to add the size parameters of interest to the `BaseCommConfig` struct (which I chose over the `ReplicaConfig` struct because these parameters seemed somehow communication-related) and then pass that struct around through the codebase to wherever these limits are needed.

I think something along those lines might be a better long-term solution because it would avoid the problem of creating a singleton which is kind of global in nature that Igor seems to object to. In the process of this work I did not finish before Rob M. was able to open this PR, I was modifying _many_ function and constructor signatures because I was finding the limits were needed in places many layers of functions below where the communication configuration was accessible, and this work was evidently taking longer than what Rob M. did even without accounting for the merge conflicts it would be creating (the descriptiveness of the merge conflicts could be something I honestly neglected to consider in hindsight).

I would be somewhat curious to hear if anyone has thoughts on what the long-term solution replacing this temporary solution will be. Does anyone have an idea differing from adding the limits to a config struct then passing that everywhere it is needed? In the process of working on the implementation I was attempting, I found a lot of things were directly or indirectly dependent on the size limits of interest. I am wondering if we shouldn't consider seeing if we can refactor anything to reduce the number of things dependent on these communication size limits. I am not entirely convinced that unnecessary dependencies on the size limits won't create other problems.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -558,21 +559,21 @@ void BCStateTran::saveReservedPage(uint32_t reservedPageId,
...
   Assert(reservedPageId < numberOfReservedPages_);
-  Assert(copyLength <= kSizeOfReservedPage);
+  Assert(copyLength <= sizeOfReservedPage_);
 
   psd_->setPendingResPage(reservedPageId, inReservedPage, copyLength);
 }
 
 void BCStateTran::zeroReservedPage(uint32_t reservedPageId) {
-    LOG_INFO(STLogger, "BCStateTran::zeroReservedPage - reservedPageId="
+    LOG_DEBUG(STLogger, "BCStateTran::zeroReservedPage - reservedPageId="
```
upshaw-alex: Was it your intention to commit the demotion of this logging statement from `INFO` to `DEBUG`, and, if so, can I ask why? This change does not seem immediately related to your configurability fix, so I wanted to validate it is intentional and understand the context for it.

robem: Thanks for your review! Yes, this is intentional, I was playing around with the number of reserved pages. Logging this for every page in INFO delays start-up time significantly.

upshaw-alex: Thank you for the clarification.

ittaiab:
merged based on support as temporary solution

PR #123, opened 2019-07-15
Conflicting write validation

teoparvanov:
The idea is to validate that a write request containing a conflict does not modify the blockchain state.

### DIFF
--- a/test/test_skvbc.py
+++ b/test/test_skvbc.py
@@ -40,6 +40,9 @@ def start_replica_cmd(builddir, replica_id):
 
 class SkvbcTest(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


test/test_skvbc.py
```diff
@@ -132,6 +133,64 @@ def test_get_block_data(self):
             kv2 = p.parse_reply(data)
             self.assertDictEqual(kv2, dict(kv))
 
+    def test_conflicting_write(self):
+        """
+        The goal is to validate that a conflicting write request does not modify the blockchain state.
```
andrewjstone: Please keep lines to 80 chars.

PR #129, opened 2019-07-23
enable -Werror for compilation

beerriot:
While the build is clean, let's turn on warnings-as-errors, to help
keep it that way.


beerriot:
Whoops - don't merge this yet, it breaks when `-DUSE_LOG4CPP=true` is specified.

beerriot:
Closer. `threshsign/bin/test/TestThresholdSerialization` is segfaulting when attempting to log.

beerriot:
Ready! Code builds, and `make test` passes both with and without `-DUSE_LOG4CPP=true`.

PR #130, opened 2019-07-23
VB-1169 Make bft-engine replica working with real persistency code.

yuliasherman:
This PR contains bug fixes for persistency functionality found during testing with file-based storage.


bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -317,112 +315,104 @@ void PersistentStorageImp::initDescriptorOfLastExecution() {
...
   actualSize = seqNumData.serializeFullCommitProofMsg(movablePtr);
   itemId = BEGINNING_OF_SEQ_NUM_WINDOW + FULL_COMMIT_PROOF_MSG + shift;
   Assert(itemId < BEGINNING_OF_CHECK_WINDOW);
-  metadataStorage_->writeInTransaction(itemId, buf, actualSize);
+  metadataStorage_->writeInTransaction(itemId, buf.get(), actualSize);
   Assert(actualSize != 0);
 
-  movablePtr = buf;
+  movablePtr = buf.get();
   actualSize = seqNumData.serializePrepareFullMsg(movablePtr);
```
andrewjstone: I just realized something with regards to these `movablePtr` use cases. The only reason we need this pattern at all is because the functions like `serializePrepareFullMsg` take a reference to a pointer. Therefore they change the pointer passed in, rather than just copying it and moving the copy in the parameter. We can get rid of all these movablePtrs if we take `char* buf` instead of `char*& buf` as a parameter. We are literally fighting ourselves because of references to pointers that we don't actually want. In the future, we should probably be suspicious of any code that takes a `*&` parameter. It's not really a necessary thing to do in most c++ code.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -317,112 +315,104 @@ void PersistentStorageImp::initDescriptorOfLastExecution() {
...
+  metadataStorage_->writeInTransaction(itemId, buf.get(), actualSize);
 }
 
-void PersistentStorageImp::serializeAndSaveCheckWindow(const SharedPtrCheckWindow &checkWindow) {
-  writeBeginningOfActiveWindow(BEGINNING_OF_CHECK_WINDOW, checkWindow.get()->getBeginningOfActiveWindow());
-  for (uint32_t i = checkWindowFirst_; i < checkWindowLast_; ++i)
-    setCheckDataElement(i, checkWindow);
+void PersistentStorageImp::saveDefaultsInCheckWindow() {
+  writeBeginningOfActiveWindow(BEGINNING_OF_CHECK_WINDOW, checkWindowBeginning_);
+  const CheckData checkData;
```
andrewjstone: Same as above. Do we need to pass this parameter in, or can it be local to to setCheckDataElement?

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -676,12 +676,8 @@ bool PersistentStorageImp::hasDescriptorOfLastNewView() {
 }
 
 bool PersistentStorageImp::hasDescriptorOfLastExecution() {
-  if (!hasDescriptorOfLastExecution_) {
-    DescriptorOfLastExecution defaultDesc;
-    DescriptorOfLastExecution storedDesc = getDescriptorOfLastExecution();
-    if (!storedDesc.equals(defaultDesc))
-      hasDescriptorOfLastExecution_ = true;
-  }
+  if (descriptorOfLastExecution_.equals(DescriptorOfLastExecution{0, Bitmap()}))
```
andrewjstone: This is a very nice shortening. Thanks!

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -905,18 +897,20 @@ void PersistentStorageImp::verifyLastNewViewMsgs(const DescriptorOfLastNewView &
   }
 }
 
-void PersistentStorageImp::verifySetDescriptorOfLastNewView(const DescriptorOfLastNewView &desc) const {
```
andrewjstone: Why are this and other verify methods no longer const ?

bftengine/src/bftengine/PrimitiveTypes.cpp
```diff
@@ -14,7 +15,7 @@ std::string CommitPathToStr(CommitPath path) {
     case CommitPath::SLOW:
       return "SLOW";
     default:
-      return "";
+      throw std::runtime_error("Unsupported CommitPath specified.");
```
andrewjstone: Thanks for this fix!

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -180,26 +180,17 @@ namespace bftEngine
...
-			}	else {
-				// set expected
-				Digest tmpDigest;
-				Digest::calcCombination(commitDigest, m->viewNumber(), m->seqNumber(), tmpDigest);
-				commitMsgsCollector->initExpected(m->seqNumber(), m->viewNumber(), tmpDigest);
-
-				// add msg
-				bool r = commitMsgsCollector->initMsgWithPartialSignature(m, m->senderId());
-				Assert(r);
+			  commitMsgsCollector->setExpected(m->seqNumber(), m->viewNumber(), tmpDigest);
```
andrewjstone: Awesome code shrinkage. Thanks for this fix.

bftengine/src/bftengine/SerializableActiveWindow.cpp
```diff
@@ -47,7 +47,7 @@ bool SerializableActiveWindow<INPUT_PARAMS>::equals(const SerializableActiveWind
 }
 
 template<TEMPLATE_PARAMS>
-void SerializableActiveWindow<INPUT_PARAMS>::deserializeElement(const uint16_t &index, char *buf, const size_t &bufLen,
+void SerializableActiveWindow<INPUT_PARAMS>::deserializeElement(const SeqNum &index, char *buf, const size_t &bufLen,
```
andrewjstone: Switching these types makes the code clearer. Thanks!

bftengine/src/communication/PlainUDPCommunication.cpp
```diff
@@ -177,13 +175,13 @@ class PlainUDPCommunication::PlainUdpImpl {
...
     std::lock_guard<std::mutex> guard(runningLock);
 
-    bufferForIncomingMessages = (char *) std::malloc(maxMsgSize);
-
     if (running == true) {
       LOG_DEBUG(_logger, "Cannot Start(): already running!");
       return -1;
     }
 
+    bufferForIncomingMessages = (char *) std::malloc(maxMsgSize);
```
andrewjstone: Sweet. Thanks for the fix.

bftengine/tests/simpleStorage/FileStorage.cpp
```diff
@@ -140,11 +142,23 @@ void FileStorage::verifyFileMetadataSetup() const {
 void FileStorage::verifyOperation(uint16_t objectId, uint32_t dataLen, const char *buffer) const {
   verifyFileMetadataSetup();
   MetadataObjectInfo *objectInfo = objectsMetadata_->getObjectInfo(objectId);
-  if (!objectInfo || objectId > objectsMetadata_->getObjectsNum() - 1 ||
-      !dataLen || dataLen > objectInfo->maxSize || !buffer) {
-    LOG_FATAL(logger_, "FileStorage::verifyOperation " << WRONG_PARAMETER);
-    throw runtime_error(WRONG_PARAMETER);
+  if (objectInfo && objectId <= (objectsMetadata_->getObjectsNum() - 1) &&
```
andrewjstone: Nice expansion. It will make tracking down errors easier. I really appreciate this.

bftengine/tests/simpleStorage/MetadataStorageTypes.hpp
```diff
@@ -46,8 +47,8 @@ struct MetadataObjectInfo {
 
 // Object data to be stored for every write operation in transaction.
 struct RequestInfo {
```
andrewjstone: It'd probably be easier to just use a vector rather than unique_ptr + len here.

I'm also not sure if you should take this by rvalue reference `&&buf` rather than buf. It's a tricky c++ area. It may be easier, and clearer to just pass the unique_ptr by value also. That shows that ownership is being given to the RequestInfo struct. Passing by reference, and looking at the constructor signature doesn't imply that's what's happening.

bftengine/tests/simpleStorage/FileStorage.cpp
```diff
@@ -202,9 +216,12 @@ void FileStorage::writeInTransaction(uint16_t objectId, char *data, uint32_t dat
     LOG_ERROR(logger_, "FileStorage::writeInTransaction " << WRONG_FLOW);
     throw runtime_error(WRONG_FLOW);
   }
-  char *buf = new char[dataLength];
-  memcpy(buf, data, dataLength);
-  transaction_->insert(pair<uint16_t, RequestInfo> (objectId, RequestInfo(buf, dataLength)));
+  std::unique_ptr<char> buf(new char[dataLength]);
```
andrewjstone: In general, anytime we are creating a unique_ptr to wrap a char*, we should probably just use a vector instead. We can still memcpy into the vector by using it's `data()` member function.

It may be worth it in the future to go through the code and make these changes. I also think most current existing raw char* can be converted to vectors and make the code shorter, clearer, and safer.

bftengine/tests/simpleTest/replica.cpp
```diff
@@ -378,7 +381,15 @@ int main(int argc, char **argv) {
...
-  replica = Replica::createNewReplica(&replicaConfig, &simpleAppState, st, comm, nullptr);
+  replicaConfig.statusReportTimerMillisec = 3000;
+  replicaConfig.viewChangeTimerMillisec = 20000;
+  ostringstream dbFile;
+  dbFile << "metadataStorageTest_" << replicaConfig.replicaId << ".txt";
+  remove(dbFile.str().c_str());
+  MetadataStorage *fileStorage = new FileStorage(replicaLogger, dbFile.str());
+
+  replica = Replica::createNewReplica(&replicaConfig, &simpleAppState, st, comm, fileStorage);
+  //replica = Replica::createNewReplica(&replicaConfig, &simpleAppState, st, comm, nullptr);
```
andrewjstone: Let's delete this line if we aren't using it.

bftengine/src/bftengine/BFTEngine.cpp
```diff
@@ -114,40 +116,42 @@ Replica *Replica::createNewReplica(ReplicaConfig *replicaConfig,
...
-  } else if (metadataStorage != nullptr) {
-    // TODO(GG):
-    // - use metadataStorage to create an instance of PersistentStorage
-    // - if needed, write false to isNewReplica
-    Assert(false);
+  if (debugPersistentStorageEnabled)
+    if (metadataStorage == nullptr)
+      persistentStoragePtr.reset(new impl::DebugPersistentStorage(replicaConfig->fVal, replicaConfig->cVal));
+
+  if (metadataStorage != nullptr) {
```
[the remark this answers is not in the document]
yuliasherman: I agree that this part is not very clear, I'll add an explanation to the code.
There are 3 legal cases:
1. metadataStorage equals to null and debugPersistentStorageEnabled is true - we create DebugPersistentStorage.
2. metadataStorage is not null and debugPersistentStorageEnabled is true - still, debug mode; file-based storage is passed in metadataStorage.
3. metadataStorage is not null and debugPersistentStorageEnabled is false - real storage is passed in metadataStorage.
So, the code covers those cases and the logic is a bit different than you expected.

bftengine/src/bftengine/BFTEngine.cpp
```diff
@@ -114,40 +116,42 @@ Replica *Replica::createNewReplica(ReplicaConfig *replicaConfig,
...
+    persistentStoragePtr.reset(new impl::PersistentStorageImp(replicaConfig->fVal, replicaConfig->cVal));
+    shared_ptr<MetadataStorage> metadataStoragePtr(metadataStorage);
+    auto objectDescriptors =
+        ((PersistentStorageImp *) persistentStoragePtr.get())->getDefaultMetadataObjectDescriptors(numOfObjects);
```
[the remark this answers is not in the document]
yuliasherman: I thought about this before I wrote this code. Don't know which case is worse. DefaultMetadataObjectDescriptors is something very specific for PersistentStorageImp and I don't like to put it in the base class. Regarding your second proposal - DebugPersistentStorage derives from PersistentStorage, so I cannot create PersistentStorageImp directly as a raw pointer. And DefaultMetadataObjectDescriptors is not relevant to it, too.

bftengine/src/bftengine/BFTEngine.cpp
```diff
@@ -114,40 +116,42 @@ Replica *Replica::createNewReplica(ReplicaConfig *replicaConfig,
...
+    isNewStorage = metadataStoragePtr->initMaxSizeOfObjects(objectDescriptors.get(), numOfObjects);
+    ((PersistentStorageImp *) persistentStoragePtr.get())->init(metadataStoragePtr);
```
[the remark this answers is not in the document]
yuliasherman: Similar issue. init function that receives metadataStorage is not relevant for the common interface. But if you insist, I can put it there.

bftengine/src/bftengine/BFTEngine.cpp
```diff
@@ -114,40 +116,42 @@ Replica *Replica::createNewReplica(ReplicaConfig *replicaConfig,
...
-    // - use metadataStorage to create an instance of PersistentStorage
-    // - if needed, write false to isNewReplica
-    Assert(false);
+  if (debugPersistentStorageEnabled)
+    if (metadataStorage == nullptr)
+      persistentStoragePtr.reset(new impl::DebugPersistentStorage(replicaConfig->fVal, replicaConfig->cVal));
+
+  if (metadataStorage != nullptr) {
+    persistentStoragePtr.reset(new impl::PersistentStorageImp(replicaConfig->fVal, replicaConfig->cVal));
+    shared_ptr<MetadataStorage> metadataStoragePtr(metadataStorage);
```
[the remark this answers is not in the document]
yuliasherman: Experienced problems in the past, but now it looks OK. Changed.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -22,25 +24,22 @@ namespace impl {
...
+  configSerializer_.reset(new ReplicaConfigSerializer(nullptr)); // ReplicaConfig placeholder
 }
 
 void PersistentStorageImp::retrieveWindowsMetadata() {
   seqNumWindowBeginning_ = readBeginningOfActiveWindow(BEGINNING_OF_SEQ_NUM_WINDOW);
   checkWindowBeginning_ = readBeginningOfActiveWindow(BEGINNING_OF_CHECK_WINDOW);
 }
 
-bool PersistentStorageImp::init(MetadataStorage *&metadataStorage) {
+bool PersistentStorageImp::init(const shared_ptr<MetadataStorage> &metadataStorage) {
```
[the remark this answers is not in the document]
yuliasherman: Yeah.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -154,86 +154,87 @@ void PersistentStorageImp::setVersion() const {
...
-  delete[] outBuf;
+  metadataStorage_->writeInTransaction(VERSION_PARAMETER, outBuf.get(), outBufSize);
 }
 
 void PersistentStorageImp::setFetchingStateInternal(const bool &state) {
   metadataStorage_->writeInTransaction(FETCHING_STATE, (char *) &state, sizeof(state));
+  fetchingState_ = state;
 }
 
 void PersistentStorageImp::setFetchingState(const bool state) {
```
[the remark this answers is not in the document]
yuliasherman: Fixed code wide.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -154,86 +154,87 @@ void PersistentStorageImp::setVersion() const {
...
+  strictLowerBoundOfSeqNums_ = seqNum;
 }
 
 void PersistentStorageImp::setStrictLowerBoundOfSeqNums(const SeqNum seqNum) {
-  Assert(!nonExecSetIsAllowed());
+  Assert(nonExecSetIsAllowed());
   setStrictLowerBoundOfSeqNumsInternal(seqNum);
 }
 
 void PersistentStorageImp::setLastViewTransferredSeqNumbersInternal(const ViewNum &view) {
```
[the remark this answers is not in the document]
yuliasherman: All those Asserts are irrelevant for initialization functions.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -317,112 +315,104 @@ void PersistentStorageImp::initDescriptorOfLastExecution() {
...
 
-void PersistentStorageImp::serializeAndSaveSeqNumWindow(const SharedPtrSeqNumWindow &seqNumWindow) {
-  writeBeginningOfActiveWindow(BEGINNING_OF_SEQ_NUM_WINDOW, seqNumWindow.get()->getBeginningOfActiveWindow());
-  for (uint32_t i = seqNumWindowFirst_; i < seqNumWindowLast_; ++i)
-    setSeqNumDataElement(i, seqNumWindow);
+void PersistentStorageImp::saveDefaultsInSeqNumWindow() {
+  writeBeginningOfActiveWindow(BEGINNING_OF_SEQ_NUM_WINDOW, seqNumWindowBeginning_);
+  const SeqNumData seqNumData;
+  for (uint32_t i = 0; i < seqWinSize; ++i)
+    setSeqNumDataElement(i, seqNumData);
```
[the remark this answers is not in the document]
yuliasherman: The idea behind this is sometimes to pass real data to setSeqNumDataElement and in other cases (like this) to reset it do the default.

bftengine/src/bftengine/PersistentStorageImp.cpp
yuliasherman: That's the way how ifstream works. It moves the pointer to after serialized data position. I have to support two separate cases: serialization of singe parameters and of whole elements containing a bunch of parameters. In the second case, this functionality is required.

bftengine/src/bftengine/PersistentStorageImp.cpp
yuliasherman: Explained before.

bftengine/src/bftengine/PersistentStorageImp.cpp
yuliasherman: hasDescriptorOfLastExecution changes data member.

bftengine/tests/simpleTest/replica.cpp
yuliasherman: Igor is improving this code now. Don't like to touch it to minimize merging issues.

bftengine/tests/simpleStorage/FileStorage.cpp
yuliasherman: I'm not sure that vector's overhead worth that.

yuliasherman: Read a bit regarding this on the internet. Should be considered to be done as a separate task.

bftengine/tests/simpleStorage/MetadataStorageTypes.hpp
yuliasherman: Changed it.

bftengine/src/bftengine/PersistentStorageImp.cpp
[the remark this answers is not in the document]
andrewjstone: Ah, ok. Thanks!

bftengine/src/bftengine/BFTEngine.cpp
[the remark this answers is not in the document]
andrewjstone: I guess the one (big) downside to doing this is that its less exception safe, as if an exception was thrown on getDefaultMetadataObjectDescriptors, the memory would not be freed.

I'm fine leaving it as is. I just wanted to point out what I meant.

bftengine/src/bftengine/BFTEngine.cpp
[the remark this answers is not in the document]
andrewjstone: No, it's fine as is. Thanks.

yuliasherman:
> I really don't think we should be removing all the const qualifiers before the parameters. It reduces the clarity of the code, because now you have to read the function to see that the passed in value doesn't change. Yes, it's a copy of a value, but it still adds information.

As for me, passing parameter by value explains the intention. CLion agrees with me.

bftengine/tests/simpleStorage/FileStorage.cpp
andrewjstone: 👍

andrewjstone:
> > I really don't think we should be removing all the const qualifiers before the parameters. It reduces the clarity of the code, because now you have to read the function to see that the passed in value doesn't change. Yes, it's a copy of a value, but it still adds information.
> 
> As for me, passing parameter by value explains the intention. CLion agrees with me.

Ok.

bftengine/src/bftengine/BFTEngine.cpp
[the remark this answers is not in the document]
yuliasherman: Got it now.

bftengine/src/bftengine/PersistentStorageImp.cpp
andrewjstone: Ok. I really think there's gotta be a cleaner way than aliasing pointers, but let's defer this for now.

yuliasherman:
Thanks for the review! Don’t you like to merge it?

andrewjstone:
> Thanks for the review! Don’t you like to merge it? 
Toly is also assigned as a reviewer. I don't think it's strictly necessary this time, but I was going to give him some time to review. 

I'll merge it.

Issue #131, opened 2019-07-23
-Werror results in compilation error on macos


no2chem:
**Describe the bug**
When building on macos (10.14.15), compilation of corebft fails due to assertions in cryptopp. I haven't tested if the behavior occurs on other platforms, but this appears related to clang and #129

update: this seems to be directly related to #129 and clang, BUILD_TESTING isn't required

```
In file included from /concord-bft/bftengine/src/bftengine/PrePrepareMsg.cpp:11:
In file included from /concord-bft/bftengine/src/bftengine/Crypto.hpp:13:
In file included from /usr/local/include/cryptopp/dll.h:17:
In file included from /usr/local/include/cryptopp/aes.h:9:
In file included from /usr/local/include/cryptopp/rijndael.h:11:
/usr/local/include/cryptopp/seckey.h:180:32: error: extra ';' inside a class [-Werror,-Wextra-semi]
        CRYPTOPP_COMPILE_ASSERT(Q > 0);
                                      ^
/usr/local/include/cryptopp/seckey.h:181:37: error: extra ';' inside a class [-Werror,-Wextra-semi]
        CRYPTOPP_COMPILE_ASSERT(N % Q == 0);
                                           ^
/usr/local/include/cryptopp/seckey.h:182:37: error: extra ';' inside a class [-Werror,-Wextra-semi]
        CRYPTOPP_COMPILE_ASSERT(M % Q == 0);
```

**To Reproduce**
Configure the project on macos and build.
```
cmake ..
make
```

**Expected behavior**
The project should build.

no2chem:
As an update, it seems BUILD_TESTING isn't required. It might have to do with the really old version of cryptopp suggested by the readme: CRYPTOPP_5_6_5

salieri11:
@no2chem we don't support MacOS as for now, I prefer to tackle MacOS support as a bigger PR, where everything should be tested - ability to run tests etc. I think that there are few dependencies that should be installed differently on MacOS, I suggest to defer this issue

no2chem:
I don't think this is a macos issue. This is just an issue with building with clang with -Wall.

upshaw-alex:
Can you confirm this is not a MacOS issue by reproducing this on a Linux VM if you do not think it is a MacOS issue?

no2chem:
[...]

PR #133, opened 2019-07-24
Fix formatting errors which result in clang warnings

### LINKED ISSUE #132: Some warnings cause the build to fail after -Werror
**Describe the bug**
It seems that after -Werror was enabled in #129 - several warnings which occur when using clang to build now cause the build to fail. One is the cryptopp dependency, as documented in issue #131 - but there are several format type warnings which clang generates that gcc does not catch as well, which cause the build to fail:

```
/concord-bft/bftengine/src/bftengine/DebugStatistics.cpp:89:53: warning: format specifies type 'long' but the argument has type 'int64_t'
      (aka 'long long') [-Wformat]
                        fprintf(stdout, "lastExecutedSeqNumber = %ld\t", d.lastExecutedSequenceNumber);
                                                                 ~~~     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```
/concord-bft/bftengine/src/bftengine/ReplicaImp.cpp:615:19: warning: format specifies type 'long' but the argument has type 'bftEngine::impl::Time'
      (aka 'unsigned long long') [-Wformat]
                                           i, currTime, timeOfPartProof);
                                              ^~~~~~~~
```

**To Reproduce**
Build the project using clang (for example, on macos).

**Expected behavior**
The project should build successfully under clang.

no2chem:
This PR addresses formatting warnings emitted by clang, as documented in #132 which cause the build to break when compiling using clang after #129 enabled -Werror. In particular, the wrong formatting type (long instead of long long) were used in these formatting strings. This patch fixes the formatting strings.

In the future, it might be a good idea to have CI test builds under both clang and gcc.

Fixes #132

### DIFF
--- a/bftengine/src/bftengine/DebugStatistics.cpp
+++ b/bftengine/src/bftengine/DebugStatistics.cpp
@@ -86,7 +86,7 @@ namespace bftEngine
 			fprintf(stdout, "numberOfReceivedSTMessages = %zd\t", d.numberOfReceivedSTMessages);
 			fprintf(stdout, "numberOfReceivedStatusMessages = %zd\t", d.numberOfReceivedStatusMessages);
 			fprintf(stdout, "numberOfReceivedCommitMessages = %zd\t", d.numberOfReceivedCommitMessages);
-			fprintf(stdout, "lastExecutedSeqNumber = %ld\t", d.lastExecutedSequenceNumber);
+			fprintf(stdout, "lastExecutedSeqNumber = %lld\t", d.lastExecutedSequenceNumber);
 			fprintf(stdout, "\n");
 
 			if (d.completedReadOnlyRequests > 0 || d.completedReadWriteRequests > 0)
--- a/bftengine/src/bftengine/DebugStatistics.cpp
+++ b/bftengine/src/bftengine/DebugStatistics.cpp
@@ -86,7 +86,7 @@ namespace bftEngine
 			fprintf(stdout, "numberOfReceivedSTMessages = %zd\t", d.numberOfReceivedSTMessages);
 			fprintf(stdout, "numberOfReceivedStatusMessages = %zd\t", d.numberOfReceivedStatusMessages);
 			fprintf(stdout, "numberOfReceivedCommitMessages = %zd\t", d.numberOfReceivedCommitMessages);
-			fprintf(stdout, "lastExecutedSeqNumber = %lld\t", d.lastExecutedSequenceNumber);
+			fprintf(stdout, "lastExecutedSeqNumber = %" PRId64 "\t", d.lastExecutedSequenceNumber);
 			fprintf(stdout, "\n");
 
 			if (d.completedReadOnlyRequests > 0 || d.completedReadWriteRequests > 0)
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2015,7 +2015,7 @@ namespace bftEngine
 			}
 
 			LOG_INFO_F(GL, "**************** In onNewView curView=%" PRId64 " (num of PPs=%ld, first safe seq=%" PRId64 ","
-                       " last safe seq=%" PRId64 ", lastStableSeqNum=%lld, lastExecutedSeqNum=%" PRId64 ","
+                       " last safe seq=%" PRId64 ", lastStableSeqNum=%" PRId64 ", lastExecutedSeqNum=%" PRId64 ","
                        "stableLowerBoundWhenEnteredToView= %" PRId64 ")",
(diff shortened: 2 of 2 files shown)


no2chem:
Hmm, after looking into this, it seems that this was changed in #128. Perhaps this is treated as a different type in linux and macos?

no2chem:
Digging into this further, it seems that the type is currently int64_t, but was changed recently

https://github.com/vmware/concord-bft/blob/38ed076ddcb11ed3abbc71d0f5e083c252a72d81/bftengine/src/bftengine/DebugStatistics.hpp#L61

@beerriot perhaps it as a long in a previous revision?

andrewjstone:
Hi @no2chem.  Thanks for the patch. We currently only have tested building on Ubuntu 18.04. It's not that other platforms won't work, it's just that we haven't really had the bandwidth to build/test on other platforms. We also default to building using clang in our cmake files.

It does seem like there is a difference between the the types as interpreted by clang on MacOS and Linux, as you point out, as this change essentially reverts #128.

Ideally we'd have builds work cross platform, but It may be best to just use Linux for the time being. If you can get it working on both platforms that's great, but we'd also want CI for both. And yes, I realize we don't even have CI for linux yet.

salieri11:
Agree with @andrewjstone I commented about it in the #131

no2chem:
@andrewjstone @salieri11 I don't think this is actually a cross platform issue. It seems that the type _is_ really a int64_t:


https://github.com/vmware/concord-bft/blob/38ed076ddcb11ed3abbc71d0f5e083c252a72d81/bftengine/src/bftengine/DebugStatistics.hpp#L61

In fact, it's not clear to me why it doesn't error with clang on Linux. I'll try on Linux in a bit.

bftengine/src/bftengine/DebugStatistics.cpp
```diff
@@ -86,7 +86,7 @@ namespace bftEngine
 			fprintf(stdout, "numberOfReceivedSTMessages = %zd\t", d.numberOfReceivedSTMessages);
 			fprintf(stdout, "numberOfReceivedStatusMessages = %zd\t", d.numberOfReceivedStatusMessages);
 			fprintf(stdout, "numberOfReceivedCommitMessages = %zd\t", d.numberOfReceivedCommitMessages);
-			fprintf(stdout, "lastExecutedSeqNumber = %ld\t", d.lastExecutedSequenceNumber);
+			fprintf(stdout, "lastExecutedSeqNumber = %lld\t", d.lastExecutedSequenceNumber);
```
beerriot:
```suggestion
			fprintf(stdout, "lastExecutedSeqNumber = %" PRId64 "\t", d.lastExecutedSequenceNumber);
```

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -611,7 +611,7 @@ namespace bftEngine
 				if (absDifference(currTime, timeOfPartProof) < controller->timeToStartSlowPathMilli() * 1000)
 					break; // don't try the next seq numbers
 
-				LOG_INFO_F(GL, "Primary initiates slow path for seqNum=%" PRId64 " (currTime=%ld timeOfPartProof=%ld",
+				LOG_INFO_F(GL, "Primary initiates slow path for seqNum=%" PRId64 " (currTime=%llu timeOfPartProof=%llu",
```
beerriot:
```suggestion
				LOG_INFO_F(GL, "Primary initiates slow path for seqNum=%" PRId64 " (currTime=%" PRIu64 " timeOfPartProof=%" PRIu64,
```

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2015,7 +2015,7 @@ namespace bftEngine
 			}
 
 			LOG_INFO_F(GL, "**************** In onNewView curView=%" PRId64 " (num of PPs=%ld, first safe seq=%" PRId64 ","
-                       " last safe seq=%" PRId64 ", lastStableSeqNum=%ld, lastExecutedSeqNum=%" PRId64 ","
+                       " last safe seq=%" PRId64 ", lastStableSeqNum=%lld, lastExecutedSeqNum=%" PRId64 ","
```
beerriot:
```suggestion
                       " last safe seq=%" PRId64 ", lastStableSeqNum=%" PRId64 ", lastExecutedSeqNum=%" PRId64 ","
```

no2chem:
Thanks @beerriot - I haven't used those macros before, but I think they do the trick!

andrewjstone:
> Thanks @no2chem. LGTM. Can you please squash these to a single commit? Then I'll merge the fix in.

Nevermind, Bryan pointed out the squash and merge button to me :)

PR #140, opened 2019-07-25
Fix the build by temporary removing -Werror and PRId64

toly-kournik:


andrewjstone:
I can see how turning off warnings fixes the build, but this breaks the fix that Bryan added to remove a warning on clang: 

```
This reverts the fix Bryan made earlier as well: home/andrewstone/concord-bft/bftengine/src/bftengine/DebugStatistics.cpp:90:54: warning: format specifies type ‘long long’ but the argument has type ‘int64_t’ (aka ‘long’) [-Wformat]
                       fprintf(stdout, “lastExecutedSeqNumber = %lld\t”, d.lastExecutedSequenceNumber);
                                                                ~~~~     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                %ld
```
I don't think we should conflate the PRId64 with turning off -Werror. The former is aesthetic, while the latter fixes the build on all platforms.

andrewjstone:
Ok, so if we merge this as is, we'll fix things but add back a fixed warning. Igor and I talked about it. Let's merge to fix the build immediately.  Then we'll convert the log lines to C++ style fix all the warnings.

PR #144, opened 2019-07-25
Change DebugStatistics to use logging instead of printf

andrewjstone:
Builds successfully on Linux with both Clang and GCC.

DebugStatistics was also changed such that it uses C++ style logging
rather than printf. Timestamps were removed from logging, since the
logger provides them for us.

DebugStatistics output looks like the following:

```
INFO 2019-07-25 13:50:14.696 (concord) STAT:
    ReadOnlyThroughput = 21.49
    WriteThroughput = 128.79
    ReceivedMessages = 3157
    SentMessages = 3161
    NumberOfReceivedSTMessages = 0
    NumberOfReceivedStatusMessages = 3
    NumberOfReceivedCommitMessages = 0
    LastExecutedSeqNumber = 2200
```

### DIFF
--- a/bftengine/src/bftengine/DebugStatistics.cpp
+++ b/bftengine/src/bftengine/DebugStatistics.cpp
@@ -1,205 +1,156 @@
-//Concord
+// Concord
 //
-//Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
 //
-//This product is licensed to you under the Apache 2.0 license (the "License").  You may not use this product except in compliance with the Apache 2.0 License. 
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
 //
-//This product may include a number of subcomponents with separate copyright notices and license terms. Your use of these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE file.
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
 
 #include "DebugStatistics.hpp"
 
-
 #ifdef DEBUG_STATISTICS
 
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/DebugStatistics.cpp
```diff
@@ -1,205 +1,160 @@
...
+    readThruput = (d.completedReadOnlyRequests / durationSec);
+
+  if (d.completedReadWriteRequests > 0)
+    writeThruput = (d.completedReadWriteRequests / durationSec);
+
+  // We use INFO logging instead of DEBUG logging since there is already a
+  // separate switch to turn DebugStatistics on and off. It's likely we want
+  // this data to appear when we do not want full debug logging.
+  LOG_INFO(GL, " STAT:" << endl <<
+      "    ReadOnlyThroughput = " << fixed << setprecision(2)
```
beerriot:
```
concord-bft/bftengine/src/bftengine/DebugStatistics.cpp:63:47: error: use of undeclared identifier 'setprecision'
      "    ReadOnlyThroughput = " << fixed << setprecision(2) 
```

I think you need to `#include <iomanip>`.

andrewjstone: WTF. I built this using both clang and gcc on Ubuntu 18.04 and neither complained. I'm really confused here.

andrewjstone: I can certainly add `#include <iomanip>`, but I'd like to understand why my system doesn't seem to need it.

andrewjstone: Ok, I just went ahead and added the header. I have no idea why it wasn't needed on my system. I also removed the windows related includes, and stdio.h since they aren't needed.

salieri11:
+1 for this clean printing of the status message! I opened #148 to clean the thread local storage leftovers

andrewjstone:
Rebased on latest master. I also changed the commit message since -Werror was enabled in #162.

andrewjstone:
Thanks for the review @salieri11

PR #135, opened 2019-07-24
Introduce simple Travis-CI build

no2chem:
This PR introduces a patch that introduces a simple Travis-CI build.

It currently builds on the Ubuntu 18.04 (bionic) infrastructure, using the distribution provided clang and gcc, and also macos with xcode10.1 (Mojave).

In currently does not run tests or clang-format. It should in the future (or I could amend this PR to).

This shouldn't be merged yet, as the build currently fails. 

The GCC build fails with an error about unused variables, documented in #136. This PR shouldn't be merged until the fix, #137, is merged.

The clang build succeeds.

The macos build fails with an error about format specifiers, documented in #132. This PR shouldn't be merged until that issue is fixed. This shows that #132 seems to actually be a cross platform issue. Merging #133 should fix this build.

PR #145, opened 2019-07-25
Fix skvbc tests

andrewjstone:
Runnng `make test` was causing the skvbc_tests to fail. Running in
verbose mode, an assert was being triggered because persistence wasn't
setup properly.

```
15: skvbc_replica: /home/andrewstone/concord-bft/bftengine/src/bftengine/BFTEngine.cpp:123: static bftEngine::Replica *bftEngine::Replica::createNewReplica(bftEngine::ReplicaConfig *, bftEngine::RequestsHandler *, bftEngine::IStateTransfer *, bftEngine::ICommunication *, bftEngine::MetadataStorage *): Assertion `false' failed.
```

PR #130 was merged in without running the automated tests which caused
this failure. The fix was to use the FileStorage class as backing for
MetadataStorage when calling the SimpleKVBC::ReplicaImp constructor. We
also could have turned on debugPersistence and avoided the assertion,
but it's better to be running our tests with actual persistence. This
way we can start testing the persistence itself from the testing
framework.

### DIFF
--- a/bftengine/tests/simpleKVBC/src/CMakeLists.txt
+++ b/bftengine/tests/simpleKVBC/src/CMakeLists.txt
@@ -12,8 +12,11 @@ set(simpleKVBC_sources
 )
 
 add_library(simpleKVBC ${simpleKVBC_sources})
-target_link_libraries(simpleKVBC PUBLIC corebft threshsign util)
+target_link_libraries(simpleKVBC PUBLIC corebft threshsign util
+    simple_file_storage_lib)
 
 target_include_directories(simpleKVBC PUBLIC .)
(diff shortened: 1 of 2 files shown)


andrewjstone:
@no2chem This should fix the problems with `make test` seen in #135

andrewjstone:
I'm merging this in without further review, because it fixes the tests and allows us proceed with CI. It basically just copies from Yulia's changes to simpletest.

yuliasherman:
That's because of disabling debugPersistentStorageEnabled flag. Sorry, about that. In fact, this flag should be enabled during the tests. Igor is working on corrresponding changes.

PR #146, opened 2019-07-27
Add a Timers class to the util library

andrewjstone:
`Timers` maintains a vector of timers, where each is associated with a
callback.  The `Timers::evaluate` method finds all expired timers and
runs their corresponding callback.

Timers can be either `ONESHOT` or `RECURRING`. A ONESHOT timer will be
removed from the collection after it fires once. A RECURRING timer will be
rescheduled to expire at the current time + the duration passed to it's
constructor. Durations can be modified by calling the `Timers::reset`
method with the handle returned from calling `Timers::add`.

The Timer and Timers classes use the std::chrono::steady_clock, and
utilize std::chrono types for safety, to reduce the occurrence of issues
like those in #141.

A detailed test showing how to use the API is also part of this commit.

Future commits will use the Timers class instead of the
SimpleTimerScheduler to provide a typesafe interface for callbacks,
and allow the removal of timer methods from the InternalReplicaAPI. We
can also remove the *TimerHandlerFunc methods from ReplicaImp, since
recurring timers will be rescheduled automatically.

### DIFF
--- a/util/include/Timers.hpp
+++ b/util/include/Timers.hpp
@@ -1,0 +1,144 @@
+// Concord
+//
+// Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <functional>
+#include <stdint.h>
+#include <chrono>
+#include <vector>
+#include <algorithm>
+
+namespace concordUtil {
+
+// A collection of timers backed by a vector.
+class Timers {
+ public:
+  class Handle {
+    Handle() {}
+    Handle(uint64_t id) : id_(id) {}
+
+   private:
+    uint64_t id_;
+    friend class Timers;
+  };
+
+  class Timer {
+   public:
+    enum Type {
+      ONESHOT,
+      RECURRING,
+    };
+
+   private:
+    Timer(std::chrono::milliseconds d, Type t, std::function<void()> cb) {
+      Timer(d, t, cb, std::chrono::steady_clock::now());
+    }
+
+    Timer(std::chrono::milliseconds d, Type t, std::function<void()> cb, std::chrono::steady_clock::time_point now)
+        : duration_(d), expires_at_(now + d), type_(t), callback_(cb) {}
+
+    bool expired() const { return expired(std::chrono::steady_clock::now()); }
+
+    bool expired(std::chrono::steady_clock::time_point now) const { return now >= expires_at_; }
+
+    bool recurring() const { return type_ == Type::RECURRING; }
+
+    void run_callback() { callback_(); }
+
+    void reset(std::chrono::steady_clock::time_point now) { expires_at_ = now + duration_; }
+
+    void reset(std::chrono::steady_clock::time_point now, std::chrono::milliseconds d) {
+      duration_ = d;
+      expires_at_ = now + duration_;
+    }
+
+    std::chrono::milliseconds duration_;
+    std::chrono::steady_clock::time_point expires_at_;
+    Type type_;
+    uint64_t id_;
+    std::function<void()> callback_;
+
+    friend class Timers;
+  };
+
+ public:
+  Timers() : id_counter_(0) {}
+
+  Handle add(std::chrono::milliseconds d, Timer::Type t, std::function<void()> cb) {
+    return add(d, t, cb, std::chrono::steady_clock::now());
+  }
+
+  Handle add(std::chrono::milliseconds d,
+             Timer::Type t,
+             std::function<void()> cb,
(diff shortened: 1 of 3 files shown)


util/include/Timers.hpp
```diff
@@ -0,0 +1,127 @@
...
+#include <chrono>
+#include <boost/heap/binomial_heap.hpp>
+
+namespace concordUtil {
+
+class TimerHeap;
+
+class Timer {
+ public:
+  class greater {
```
robem: Lower-case class name? Does it have to be a class with an overloaded operator? Why don't you define a static function in the timer class?
I feel like I'm missing something ...

util/test/timers_tests.cpp
```diff
@@ -0,0 +1,98 @@
...
+  int recurring_counter = 0;
+  auto heap = TimerHeap();
+
+  // Create a oneshot and a recurring timer and add them to the heap. Each one
+  // will fire a callback when it expires that increments a counter.
+  heap.emplace(Timer(duration, Timer::ONESHOT, [&oneshot_counter]() { ++oneshot_counter; }));
+  auto recurring_handle = heap.emplace(
+      Timer(duration, Timer::RECURRING, [&recurring_counter]() { ++recurring_counter; }));
+
+  steady_clock::time_point now = steady_clock::now();
```
beerriot: If it takes 100ms to get from line 33 to here, things are going very poorly, but would it be better to move this line above the two `Timer` instantiations anyway? They each call `steady_clock::now` as well, and reordering would make sure that the `now` passed to `evaluate` on line 40 would not have advanced.

util/include/Timers.hpp
andrewjstone: Removed, as it's no longer needed.

util/include/Timers.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+  typedef boost::heap::binomial_heap<Timer, boost::heap::compare<Timer::greater>> Heap;
+
+  enum type {
+    ONESHOT,
+    RECURRING,
+  };
+
+  Timer(std::chrono::milliseconds d, type t, std::function<void()> cb)
+      : duration_(d), expires_at_(std::chrono::steady_clock::now() + d), type_(t), callback_(cb) {}
```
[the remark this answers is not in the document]
andrewjstone: Not an issue anymore, as we only loop once.

util/test/timers_tests.cpp
andrewjstone: Nice catch! I changed this in a different manner. There is now a Timer constructor that takes the now as a parameter, like the other functions. This allows tests to be deterministic.

andrewjstone:
I updated this code to remove reliance on boost. Since we have a small number of timers, using a vector makes more sense anyways.

util/include/Timers.hpp
```diff
@@ -0,0 +1,127 @@
...
+ private:
+  void run_callback() { callback_(); }
+
+  Heap::handle_type handle() const { return handle_; }
+
+  void reset(std::chrono::steady_clock::time_point now) { expires_at_ = now + duration_; }
+
+  void reset(std::chrono::steady_clock::time_point now, std::chrono::milliseconds d) {
+    duration_ = d;
+    expires_at_ = now + duration_;
```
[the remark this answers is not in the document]
andrewjstone: I find the prior version easier to read, so I left it.

util/include/Timers.hpp
```diff
@@ -0,0 +1,144 @@
+// Concord
+//
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
```
upshaw-alex: Are you getting this header file from somewhere existing or is it new? I notice you are using a 2018 copyright here rather than a 2019 copyright.

andrewjstone: typo: fixed.

util/include/Timers.hpp
```diff
@@ -0,0 +1,144 @@
...
+  void reset(Handle handle, std::chrono::milliseconds d) { reset(handle, d, std::chrono::steady_clock::now()); }
+
+  void reset(Handle handle, std::chrono::milliseconds d, std::chrono::steady_clock::time_point now) {
+    auto it = find(handle);
+    it->reset(now, d);
+  }
+
+  void cancel(Handle handle) {
+    auto it = find(handle);
+    timers_.erase(it);
```
teoparvanov: I guess we should also decrement total_timers, in order to maintain the invariant. Actually, this counter seems redundant with timers_.size(), of constant complexity.

andrewjstone: The total_timers is used to give a unique id to each timer. Therefore it must monotonically grow. I suppose the name is a bit unclear. I could have called it `count`, but that didn't seem descriptive enough. I suppose `id_counter_` could work.

teoparvanov: Yep, id_counter_ sounds great!

andrewjstone: Ok. I changed it. I also added a test to show the monotonic behavior, and one to show what happens when a removed timer is attempted to be accessed via handle. Force pushed.

teoparvanov: LGTM, thanks!

andrewjstone:
Last test run failed due to format-check. I fixed formatting, rebased and pushed.

PR #153, opened 2019-07-30
Default debugPersistentStorageEnabled to true

### LINKED ISSUE #152: Use DebugPersistentStorage instead of FileStorage for SimpleKVBC TesterReplica
**Describe the bug**
In order to fix the skvbc tests, I added file persistence for bft metadata to the SimpleKVBC TesterReplica. However, the application state itself is not persisted. This is confusing, and can potentially lead to weird errors. The file persistence should be removed and instead, the in memory, debug persistence should be used.

Thanks to @guyg8 for catching this silly mistake on my part.

**To Reproduce**
N/A

**Expected behavior**
We should use Debug persistence flag.

andrewjstone:
Our tests don't all work with persistent storage. As the storage is not
yet configurable, re-enable debugPersistentStorageEnabled = true, so
that tests pass.

Also, remove related FileStorage code from simpleKVBC, as it makes no
sense to have persistent metadata with in memory application state.

Fixes #152

### DIFF
--- a/bftengine/src/bftengine/SysConsts.hpp
+++ b/bftengine/src/bftengine/SysConsts.hpp
@@ -137,4 +137,4 @@ constexpr uint32_t MaxSizeOfPrivateKey = 1024; // TODO(GG): should be checked
 constexpr uint32_t MaxSizeOfPublicKey = 1024; // TODO(GG): should be checked
 
 
-constexpr bool debugPersistentStorageEnabled = false;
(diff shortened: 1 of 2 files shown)


salieri11:
Lets defer it until my changes will be merged, this variable doesnt exist anymore in the SysConsts file

andrewjstone:
Closed in favor of #161

PR #162, opened 2019-08-10
Add rocksdb related storage code

andrewjstone:
All rocksdb code serves as an example and is completely optional.

The storage code is spit into two layers in separate directories:

 * `rocksdb` - contains rocksdb client code and MetadataStorage
 implementation
 * `blockchain` contains higher level code that uses the database
 interfaces to organize blocks of data such that they can be stored
 persistently.

The rocksdb layer is the absolute minimum storage code and can be used
independently of the blockchain layer. The blockchain layer is a thin
abstraction on top of the database interfaces that creates a useful API
for accessing persistent blockchain information. This code serves as an
example and is also useful for building a test replica similar to the
simpleKVBCTests TesterReplica that will allow us to create test clusters
full of persistent replicas.

The majority of this code was not written by me. It was ported from a
closed source repository. I only made minor tweaks to remove circular
dependencies, change namespaces, and split the code into multiple
libraries. Credit for most of the code goes to @guyg8 @robem @beerriot
@yuliasherman. Apologies if I forgot anyone.

Some types were extracted into our util library so that they could be
shared by both the blockchain and rocksdb libraries. It's anticipated
they will be useful elsewhere.

This change also includes tests for storage code and sliver that can be
run in the standard manner with `make test` and `ctest -R <TEST>`.

Please see the README in the storage directory for more information.

# Recommendations for Reviewers
I selected reviewers who are familiar with this code already. Although it is a large pull request, the vast majority of it is unchanged from the closed source version. I would focus on the namespaces, the CMake code and any noticeable changes to interfaces. One specific change I made to break a circular dependency between the blockchain and rocksdb libraries is worth noting. Rather than use the `KeyManipulator` class directly in the rocksdb code, we now pass in a function to the constructor of `RocksDBMetadataStorage` for generating keys from `ObjectId`s.

@yuliasherman I used your unmerged branch to update the `RocksDBMetadataStorage` implementation to match the `MetadataStorage` interface changes made for persistence. Please take a look at that code and make sure everything looks alright.

In general, if you can build the code and run the tests, everything is likely fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,9 @@ option(BUILD_COMM_TCP_PLAIN "Enable TCP communication" FALSE)
 # Default BUILD_COMM_TCP_TLS to FALSE
 option(BUILD_COMM_TCP_TLS "Enable TCP TLS communication" FALSE)
 
+# This requires the rocksdb dependencies to be installed, so defaults to FALSE
+option(BUILD_ROCKSDB_STORAGE "Enable building of RocksDB storage library" FALSE)
+
 set(COMM_MODULES 0)
 if(BUILD_COMM_TCP_PLAIN)
     math(EXPR COMM_MODULES "${COMM_MODULES}+1")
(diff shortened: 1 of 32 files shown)


andrewjstone:
Closing to retarget against feature/storage. Toly and I will be refining the interfaces together on that branch.

PR #170, opened 2019-08-13
Add an in-memory db client to storage

andrewjstone:
An in-memory client was added as a new low level storage library. It
implements the db client interfaces as an alternative to the rocksdb
implementation. This allows the blockchain code to be built on top of
the in-memory implementation if rocksdb is not being used.

The cmake files were changed to decouple the rocksdb build from the
blockchain build. The README was also updated to reflect these and the
prior changes.


andrewjstone:
Closing, as the layout of the storage library has changed significantly with #173. I'll re-add the important parts in a new PR.

PR #174, opened 2019-08-15
Add an in-memory db client to storage library

andrewjstone:
An in-memory client was added as a new low level storage library. It
implements the db client interfaces as an alternative to the rocksdb
implementation. This allows the blockchain code and db_metdata_storageto
be built on top of the in-memory implementation if rocksdb is not being
used.

The README was also updated to reflect these changes.


andrewjstone:
> LGTM!
> Should we add a test or make the existing one to work with inmem client?

Thanks for the review Toly. I'm going to use this code in the existing skvbc tests instead of what's there currently. That should cover it, but we can always change the existing one to work with inmem as well.

PR #175, opened 2019-08-15
Use storage instead of test code in simpleKVBC

andrewjstone:
This PR is a deduplication effort for in memory storage code used by the
simpleKVBC test library. The test code consists of old copies of the
code in the storage library. The storage library contains the canonical
versions and is useful outside tests as well.

Of note is that we have replaced the Slice datatype, which aliases
raw pointers, with the Sliver data type that takes ownership of the data
in a shared_ptr. This requires copies in some places where they did not
exist before. However, this change is an intermediate step towards a
full refactor of SimpleKVBC, which is in progress. That will remove some
of these extra copies, as the signatures of some functions that take
slivers have changed to taking the raw pointers.

In order to make copying of buffers more ergonomic, a static `copy`
method has been added to the Sliver class, along with a test.

The code changes have been validated via existing simpleKVBC tests. The
scripts for manual testing have been run, along with the python testing
framework that uses the SimpleKVBCTests TesterReplica. All tests pass.


bftengine/tests/simpleKVBC/src/ClientImp.cpp
```diff
@@ -56,37 +57,30 @@ namespace SimpleKVBC {
...
 
-		outReply = Slice(p, replySize);
+		outReply = Sliver(p, replySize);
 
 		return Status::OK();
 	}
 
 
-	Status ClientImp::release(Slice& slice)
+	Status ClientImp::release(Sliver& _slice)
```
beerriot: I suppose `release` is part of an interface that we have to implement? No way to just get rid of this function? If that's the case, I think a comment here to remind people that Sliver does its own memory management would be a good idea.

bftengine/tests/simpleKVBC/src/ReplicaImp.h
```diff
@@ -33,19 +43,20 @@ namespace SimpleKVBC {
...
 	public:
 		// IReplica methods
-																	 
+                
 		virtual Status start() override;
 		virtual Status stop() override;
 		virtual RepStatus getReplicaStatus() const override;
 		virtual bool isRunning() const override;
-																				 
+                virtual void monitor() const override {}
```
beerriot: nit: alignment … oh, it's a tabs-vs-spaces issue. Not a blocker.

bftengine/tests/simpleKVBC/src/ReplicaImp.h
```diff
@@ -17,11 +17,21 @@
...
 
 #include "KVBCInterfaces.h"
-#include "BlockchainDBAdapter.h"
+#include "blockchain/db_interfaces.h"
+#include "blockchain/db_adapter.h"
 #include "SimpleBCStateTransfer.hpp"
 #include "Replica.hpp"
 #include "Metrics.hpp"
 
+using concord::storage::blockchain::ILocalKeyValueStorageReadOnly;
```
beerriot: `using` is strongly discouraged in header files, since they get injected into all files that include them. I guess this small test utility doesn't make a huge difference.

yuliasherman: +1 against "using".

bftengine/tests/simpleKVBC/src/ClientImp.cpp
andrewjstone: Good catch. I ended up removing it from the interface, as it is no longer necessary.

bftengine/tests/simpleKVBC/src/ReplicaImp.h
andrewjstone: I formatted the file.

bftengine/tests/simpleKVBC/src/ReplicaImp.h
andrewjstone: I removed them all.

andrewjstone:
Closing in favor of #180

PR #177, opened 2019-08-16
ALLOW_0_FAULT_TOLERANCE

jaybny:
enable f_val to be set to 0, this is first for testing purposes, but also a possible feature to continue running with a single node.

(3 * fVal + 2 * cVal + 1) equals 1 , when fVal and cVal are both 0.


bftengine/src/bftengine/SimpleClient.cpp
```diff
@@ -170,7 +170,11 @@ namespace bftEngine
 			clientSendsRequestToAllReplicasPeriodThresh{p.clientSendsRequestToAllReplicasPeriodThresh},
 			clientPeriodicResetThresh{p.clientPeriodicResetThresh}
 		{
-				Assert(_fVal >= 1);
+#ifdef ALLOW_0_FAULT_TOLERANCE
+                Assert(_fVal >= 0);
+#else
+                Assert(_fVal >= 1);
```
upshaw-alex: Is it your intention to change the indentation level of line 173 (old)/176 (new)? I am aware this easy to do unintentionally, as I notice this uses tab characters rather than spaces for indents (which is, IMHO, something we should change, but I perceive that revision as unrelated to your current proposed changes, so I would suggest complying with existing formatting until we make such a broader revision).

upshaw-alex:
I have noticed that, for some of the other compile definitions for `ifdef`/`ifndef`, we have some code in our CMake files to enable setting the flags via command line input to CMake, rather than requiring modification to the source code. I notice you add a new compile definition for `ifdef` checking here, but do not add a way to enable by, say, passing `-DALLOW_0_FAULT_TOLERANCE=TRUE` to CMake. Does anyone else (including both Jay and any other reviewers who show up here) have an opinion on whether we should add CMake functionality to turn on and off the new parameter via command line to CMake?

bftengine/src/bftengine/SimpleClient.cpp
jaybny: yes. had clang-format disabled on my ide. will update. thanks!

jaybny:
> add CMake functionality to turn on and off the new parameter via command line to CMake?
depends on git strategy for the public-source repo, and if 0F is actually useful. would hold off for now.

salieri11:
I don't see any usage for 1 node scenario. Even in the tests - what type of functionality can be tested with 1 node?

lazaridiscom:
> I don't see any usage for 1 node scenario. Even in the tests - what type of functionality can be tested with 1 node?

Testing, playing around, reference speed, ...

salieri11:
The whole point of the system is Byzantine Fault Tolerance. For edge cases of testing only execution layer IMO its not desirable to support such types of configurations.

lazaridiscom:
then just close this PR.

salieri11:
Make sense. Thanks

PR #179, opened 2019-08-20
Reduce log level for select log messages.

### LINKED ISSUE #165: Reduce log level in core BFT engine
Change the INFO level to DEBUG in most BFT internal log messages. The current INFO level produces huge logs and are not suitable for production.

upshaw-alex:
This PR demotes a number of log statements in Concord-BFT from `INFO` to `DEBUG` log level. This log demotion is helpful because Concord-BFT has a number of log statements, at the `INFO` level before this commit, in common, frequently executed, and/or non-error code paths that do not provide much information for a normally operating system but which do clutter up the logs and increase their growth rate. The log messages demoted by this commit were deemed not to provide enough new information in the common case to justify the `INFO` log level. Additionally, this commit removes the definition of `DEBUG_STATISTICS` by default (which controls whether Concord-BFT will automatically periodically output debug statistics), and makes the enhancements needed to re-enable `DEBUG_STATISTICS` by providing `-DDEBUG_STATISTICS=TRUE` as a command line option to `cmake` when building Concord-BFT.

With these changes, all tests accounted for with `make test` still pass for me on my development VM.

One thing this PR does not do which I hope to address in a near-future PR is revise `concord-bft/bftengine/tests/simpleTest/README.md` to reflect what the output of simpleTest is expected to look like after these log demotions. I am holding off on this revision because I am having difficulty running `simpleTest` on my machine as I am trying to use it at the moment. I do not think my difficulties are caused by my changes, however, as I have the same issues on master.

Fixes #165.

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -80,6 +80,10 @@ struct ReplicaConfig {
   // StateTransfer
   uint32_t maxNumOfReservedPages = 2048;
   uint32_t sizeOfReservedPage = 4096;
+
+  // If set to true, this replica will periodically log debug statistics such as
+  // throughput and number of messages sent.
+  bool debugStatisticsEnabled = false;
 };
 
 }
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -193,9 +193,9 @@ void ReplicaImp::sendRaw(char *m, NodeIdType dest, uint16_t type, MsgSize size) 
     return;
   }
 
-#ifdef DEBUG_STATISTICS
-  DebugStatistics::onSendExMessage(type);
-#endif
+  if (debugStatisticsEnabled) {
+    DebugStatistics::onSendExMessage(type);
+  }
 
   errorCode = communication->sendAsyncMessage(dest, m, size);
 
@@ -317,7 +317,7 @@ void ReplicaImp::onMessage(ClientRequestMsg *m) {
       }
     }
   } else if (seqNumberOfLastReply == reqSeqNum) {
-    LOG_INFO_F(GL, "ClientRequestMsg has already been executed - retransmit reply to client");
+    LOG_DEBUG_F(GL, "ClientRequestMsg has already been executed - retransmit reply to client");
 
     ClientReplyMsg *repMsg = clientsManager->allocateMsgWithLatestReply(clientId, currentPrimary());
 
@@ -412,7 +412,7 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
 
   Assert(pp->numberOfRequests() > 0);
 
-  LOG_INFO_F(GL, "Sending PrePrepareMsg (seqNumber=%"
+  LOG_DEBUG_F(GL, "Sending PrePrepareMsg (seqNumber=%"
       PRId64
       ", requests=%d, size=%d",
              pp->seqNumber(), (int) pp->numberOfRequests(), (int) requestsQueueOfPrimary.size());
@@ -475,7 +475,7 @@ void ReplicaImp::onMessage(PrePrepareMsg *msg) {
   metric_received_pre_prepares_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
 
-  LOG_INFO_F(GL, "Node %d received PrePrepareMsg from node %d for seqNumber %"
+  LOG_DEBUG_F(GL, "Node %d received PrePrepareMsg from node %d for seqNumber %"
       PRId64
       " (size=%d)",
              (int) myReplicaId, (int) msg->senderId(), msgSeqNum, (int) msg->size());
@@ -484,12 +484,12 @@ void ReplicaImp::onMessage(PrePrepareMsg *msg) {
     Assert(!msg->isNull()); // we should never send (and never accept) null PrePrepare message
 
     if (viewsManager->addPotentiallyMissingPP(msg, lastStableSeqNum)) {
-      LOG_INFO_F(GL, "Node %d adds PrePrepareMsg for seqNumber %"
+      LOG_DEBUG_F(GL, "Node %d adds PrePrepareMsg for seqNumber %"
           PRId64
           " to viewsManager", (int) myReplicaId, msgSeqNum);
       tryToEnterView();
     } else {
-      LOG_INFO_F(GL, "Node %d does not add PrePrepareMsg for seqNumber %"
+      LOG_DEBUG_F(GL, "Node %d does not add PrePrepareMsg for seqNumber %"
           PRId64
           " to viewsManager", (int) myReplicaId, msgSeqNum);
     }
@@ -757,7 +757,7 @@ void ReplicaImp::sendPreparePartial(SeqNumInfo &seqNumInfo) {
 
     Assert(pp != nullptr);
 
-    LOG_INFO_F(GL, "Sending PreparePartialMsg for seqNumber %"
+    LOG_DEBUG_F(GL, "Sending PreparePartialMsg for seqNumber %"
         PRId64
         "", pp->seqNumber());
 
@@ -786,7 +786,7 @@ void ReplicaImp::sendCommitPartial(const SeqNum s) {
   if (seqNumInfo.committedOrHasCommitPartialFromReplica(myReplicaId))
     return; // not needed
 
-  LOG_INFO_F(GL, "Sending CommitPartialMsg for seqNumber %"
+  LOG_DEBUG_F(GL, "Sending CommitPartialMsg for seqNumber %"
       PRId64
       "", s);
 
@@ -808,7 +808,7 @@ void ReplicaImp::onMessage(PartialCommitProofMsg *msg) {
   Assert(repsInfo->isIdOfPeerReplica(msgSender));
   Assert(repsInfo->isCollectorForPartialProofs(msgView, msgSeqNum));
 
-  LOG_INFO_F(GL, "Node %d received PartialCommitProofMsg (size=%d) from node %d for seqNumber %"
+  LOG_DEBUG_F(GL, "Node %d received PartialCommitProofMsg (size=%d) from node %d for seqNumber %"
       PRId64
       "",
              (int) myReplicaId, (int) msg->size(), (int) msgSender, msgSeqNum);
@@ -844,7 +844,7 @@ void ReplicaImp::onInternalMsg(FullCommitProofMsg *msg) {
 
 void ReplicaImp::onMessage(FullCommitProofMsg *msg) {
   metric_received_full_commit_proofs_.Get().Inc();
-  LOG_INFO_F(GL, "Node %d received FullCommitProofMsg message for seqNumber %d",
+  LOG_DEBUG_F(GL, "Node %d received FullCommitProofMsg message for seqNumber %d",
              (int) myReplicaId, (int) msg->seqNumber());
 
   const SeqNum msgSeqNum = msg->seqNumber();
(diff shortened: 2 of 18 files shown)


bftengine/src/bftengine/InternalReplicaApi.hpp
```diff
@@ -80,10 +80,11 @@ namespace bftEngine
 			virtual void onStatusReportTimer(Time cTime, Timer& timer) = 0;
 			virtual void onSlowPathTimer(Time cTime, Timer& timer) = 0;
 			virtual void onInfoRequestTimer(Time cTime, Timer& timer) = 0;
-			virtual void onDebugStatTimer(Time cTime, Timer& timer) = 0;
 			virtual void onMetricsTimer(Time cTime, Timer& timer) = 0;
 
-
+         #ifdef DEBUG_STATISTICS
+			virtual void onDebugStatTimer(Time cTime, Timer& timer) = 0;
+         #endif /* DEBUG_STATISTICS defined */
```
beerriot: I'm not wild about the idea of this function only being defined if a macro is set at compile time. In my experience, this leads rather quickly to the project only compiling when the flag is set in the default manner. Also, I think it's likely that I'll want to enable/disable statistics output on a deployed cluster, without recompiling. Can we make this a config option instead?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -442,7 +445,7 @@ namespace bftEngine
 
 			Assert(pp->numberOfRequests() > 0);
 
-			LOG_INFO_F(GL, "Sending PrePrepareMsg (seqNumber=%" PRId64 ", requests=%d, size=%d",
+			LOG_DEBUG_F(GL, "Sending PrePrepareMsg (seqNumber=%" PRId64 ", requests=%d, size=%d",
```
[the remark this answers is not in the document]
upshaw-alex: A GitHub review comment does not seem like the best place to keep track of a "note for later" to me, though perhaps I use them differently than you. Should an issue be created in GitHub to track this? Should we make a note of this in some other channel?

bftengine/src/bftengine/InternalReplicaApi.hpp
upshaw-alex: As an aside, I made this function dependent on a compile time macro after I found Concord-BFT didn't compile with `DEBUG_STATISTICS` turned off (i.e. not what was previously in the default manner). I can try to make this a config option.

To be completely clear, when you say you want to enable/disable statistics output on a deployed cluster without recompiling, restarting the cluster is still okay, right? I can make this a launch-time parameter, but making it dynamically reconfigurable for a running cluster is not something I feel we have the infrastructure to easily do yet in Concord-BFT.

beerriot: Yes, restarting is fine. I just might not be able to copy a new executable to the environment I'm working with.

upshaw-alex:
> It would help with evaluation if there were sample logs attached for the same test before and after the change.

I will work on getting these estimates. When you say, "the same test", did you have any particular test in mind? As I previously mentioned, I am having difficulty running `simpleTest` currently, even on `master`. I have attempted to reach out to the Concord-BFT people about this in another channel, but I have not received any responses from them on the matter yet.

beerriot:
> I will work on getting these estimates. When you say, "the same test", did you have any particular test in mind?

No, it's fine if it's manual, even. It would just be nice to have an example handy of how this changes log contents in practice.

bftengine/src/bftengine/ReplicaImp.cpp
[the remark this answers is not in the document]
upshaw-alex: Please correct me if you use GitHub review comments in a way differently than I do, but they do not seem like the most natural place to keep a "note for later" to me. Should we be making a GitHub issue for this in order to track this need?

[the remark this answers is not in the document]
beerriot: You're right. File #182

upshaw-alex:
I have pushed a new commit to make toggling debug statistics a `ReplicaConfig` parameter rather than a compile-time macro based on Bryan's advice.

bftengine/src/bftengine/InternalReplicaApi.hpp
upshaw-alex: In my most recent commit, I haveremove the `DEBUG_STATISTICS` compile-time macro and added a `debugStatisticsEnabled` `ReplicaConfig` configuration parameter as you have suggested. Please let me know if this addresses your concerns.

beerriot: Looks great, thank you!

beerriot:
I'll back off on the request for sample logs before and after. I grepped through the code on your branch to find the INFO-and-higher logs still in place. They look reasonable. I'm tempted to move the message about receiving a ClientRequestMsg in ReplicaImp back to INFO, but I'll hold off and see whether I really miss it.

What do the other concord-bft devs think of the new levels? @salieri11 @yuliasherman @toly-kournik @andrewjstone

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -337,7 +337,8 @@ namespace bftEngine
 
 						send(m, currentPrimary()); // TODO(GG): add a mechanism that retransmits (otherwise we may start unnecessary view-change )
 
-						LOG_INFO_F(GL, "Sending ClientRequestMsg to current primary");
+						LOG_DEBUG_F(GL,
```
andrewjstone: This one is pretty rare, except in cases of a faulty primary. It may be useful to leave it at INFO for now.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1364,7 +1365,7 @@ namespace bftEngine
 					CheckpointMsg* x = new CheckpointMsg(msgSenderId, msgSeqNum, msgDigest, msgIsStable);
 					tableOfStableCheckpoints[msgSenderId] = x;
 
-					LOG_INFO_F(GL, "Node %d added stable Checkpoint message to tableOfStableCheckpoints (message from node %d for seqNumber %" PRId64 ")", (int)myReplicaId, (int)msgSenderId, msgSeqNum);
+					LOG_DEBUG_F(GL, "Node %d added stable Checkpoint message to tableOfStableCheckpoints (message from node %d for seqNumber %" PRId64 ")", (int)myReplicaId, (int)msgSenderId, msgSeqNum);
```
andrewjstone: I think we should leave this as INFO level. It's a very important message, and infrequent.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1407,7 +1408,7 @@ namespace bftEngine
 
 			if (askForStateTransfer)
 			{
-				LOG_INFO_F(GL, "call to startCollectingState()");
```
andrewjstone: This is probably useful for debugging during testing, but we can always turn on debug mode. I don't think it's that frequent, so it may be beneficial to leave it at INFO, in case we ship a bug where state transfer isn't turning on for some reason.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2217,7 +2201,7 @@ namespace bftEngine
 		{
 			Assert(newStateCheckpoint % checkpointWindowSize == 0);
 
-			LOG_INFO_F(GL, "onTransferringCompleteImp with newStateCheckpoint==%" PRId64 "", newStateCheckpoint);
+			LOG_DEBUG_F(GL, "onTransferringCompleteImp with newStateCheckpoint==%" PRId64 "", newStateCheckpoint);
```
andrewjstone: If we leave the message about starting state transfer at INFO, we'll likely want to leave this at INFO as well.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2307,7 +2292,7 @@ namespace bftEngine
 
 			if (askAnotherStateTransfer)
 			{
-				LOG_INFO_F(GL, "call to startCollectingState()");
```
andrewjstone: Same as above.

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -80,6 +80,10 @@ struct ReplicaConfig {
   // StateTransfer
   uint32_t maxNumOfReservedPages = 2048;
   uint32_t sizeOfReservedPage = 4096;
+
+  // If set to true, this replica will periodically log debug statistics such as
+  // throughput and number of messages sent.
+  bool debugStatisticsEnabled = false;
```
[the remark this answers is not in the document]
upshaw-alex: I am aware that I am changing the default behavior here. It was my intention to make the debug statistics disabled by default, as I perceived them as something we used in development and debugging, but not something that we would actually want in production (though perhaps I am making assumptions I shouldn't be based on the fact these are called the "debug" statistics). It was my thinking that we should probably make intended production behavior the default rather than debug behavior as the default. I cannot tell whether your comment is just asking me to confirm that disabling this by default was my intention or whether you intend to imply the maintainers of Concord-BFT need debug statistics to be enabled by default.

If you want me to leave this enabled by default, I can work with that. I have a production use case from a consumer of this library where we probably want debug statistics disabled, but I can make that consumer elect the non-default value here if you want to make enabled default.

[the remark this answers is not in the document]
andrewjstone: I wasn't sure if you did it on purpose. Since you did this on purpose, I personally don't use it, I  don't believe it's heavily used elsewhere, and we have a customer who wants it off, I think we can leave it off. It's easy enough to configure for tests.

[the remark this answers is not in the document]
upshaw-alex: I am going to err on the side of leaving the statistics disabled by default if no one more explicitly objects to this.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -277,7 +277,7 @@ namespace bftEngine
 			const bool readOnly = m->isReadOnly();
 			const ReqId reqSeqNum = m->requestSeqNum();
 
-			LOG_INFO_F(GL, "Node %d received ClientRequestMsg (clientId=%d reqSeqNum=%" PRIu64 ", readOnly=%d) from Node %d", myReplicaId, clientId, reqSeqNum, readOnly ? 1 : 0, senderId);
+			LOG_DEBUG_F(GL, "Node %d received ClientRequestMsg (clientId=%d reqSeqNum=%" PRIu64 ", readOnly=%d) from Node %d", myReplicaId, clientId, reqSeqNum, readOnly ? 1 : 0, senderId);
```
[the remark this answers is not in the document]
upshaw-alex: As you and Bryan both have reservations about this line, I am thinking I might revert it to `INFO`. I believe the log reductions I have should still reduce log growth by a factor of several times even if they don't slow it to a crawl in non-error cases. I think we can revisit demoting this statement if we find log growth is still unacceptable.

bftengine/src/bftengine/ReplicaImp.cpp
upshaw-alex: I will do so.

bftengine/src/bftengine/ReplicaImp.cpp
upshaw-alex: Thank you for highlighting the importance of this specific message; I did not realize it myself. I will revert this to `INFO`.

bftengine/src/bftengine/ReplicaImp.cpp
upshaw-alex: I will return in to `INFO`.

bftengine/src/bftengine/ReplicaImp.cpp
upshaw-alex: I will do so.

bftengine/src/bftengine/ReplicaImp.cpp
upshaw-alex: Will do.

upshaw-alex:
I have pushed a commit making revisions in light of @andrewjstone 's feedback.

upshaw-alex:
I have force-pushed a rebase off of master to this branch to remove the merge conflicts.

upshaw-alex:
I have force pushed a commit that squashes the previous three I had on this branch to one in the interest of making the Git history prettier.

upshaw-alex:
I have validate `make test` still passes all tests on my development VM on the current state of this branch.

PR #180, opened 2019-08-20
Use storage instead of test code in simpleKVBC

andrewjstone:
This is a replacement PR for #175. The feature/storage branch was rebased and so I needed to open another PR with cherry-picked commits. I left the fixes for the prior review there. It should be good to merge.


bftengine/tests/simpleKVBC/include/KVBCInterfaces.h
```diff
@@ -13,218 +13,169 @@
...
-            Ready                = 0,
-            Starting,
-            Running,
-            Stopping,
-        };
-
-        // returns the current status of the replica
-        virtual RepStatus getReplicaStatus() const = 0;
+using concordUtils::Status;
+using concordUtils::Sliver;
```
beerriot: I guess I missed these two the first time. There should not be `using` directives in header files.

andrewjstone: I'll fix that in a future commit. Thanks!

PR #181, opened 2019-08-20
VB-1507 Correct concord-bft internal configuration parameters to meet…

yuliasherman:
The product configures concord-bft parameters in a way that contradicts internal configuration. Adapt internal parameters to support a large number of reserved pages and big reply messages. 
Add logs to reflect the required configuration.

### DIFF
--- a/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
+++ b/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
@@ -91,9 +91,16 @@ struct Config {
 
   bool pedanticChecks = false;
 
-  uint32_t maxBlockSize = 10 * 1024 * 1024; // 10MB
-  uint32_t maxChunkSize = 16384; // 128 - for debugging/testing
+#if defined USE_COMM_PLAIN_TCP || defined USE_COMM_TLS_TCP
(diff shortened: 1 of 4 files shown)


andrewjstone:
These changes appear to make state transfer take longer to complete during our testing. They were causing our state transfer test to fail because of this. Its likely the large page sizes cause more allocation and copying, which then cause our test to timeout. This can be rectified by increasing the timeout waiting for state transfer to complete.

```
andrewstone@ubuntu:~/concord-bft/build$ git diff ../test/bft_tester.py
diff --git a/test/bft_tester.py b/test/bft_tester.py
index c484cb4..497f6f5 100644
--- a/test/bft_tester.py
+++ b/test/bft_tester.py
@@ -239,7 +239,7 @@ class BftTester:
     async def wait_for_state_transfer_to_stop(self,
                                               up_to_date_node,
                                               stale_node):
-        with trio.fail_after(30): # seconds
+        with trio.fail_after(60): # seconds
             # Get the lastExecutedSeqNumber from a started node
             key = ['replica', 'Gauges', 'lastExecutedSeqNum']
             last_exec_seq_num = await self.metrics.get(up_to_date_node, *key)
```

yuliasherman:
Thanks, you are right, as always.

PR #186, opened 2019-08-26
VB-1244 Serialize/deserialize boolean parameters as uint8_t.

yuliasherman:
- Make boolean parameters serialization platform-independent.
- Remove unused function from Serializable class and rename its logger to more readable.


toly-kournik:
lgtm

PR #191, opened 2019-08-28
Refactor SimpleKVBCTests and SimpleKVBC

andrewjstone:
Significant refactoring of the SimpleKVBC code and tests was done in a
private repository. This PR ports over the majority of those changes,
modifying headers and types where appropriate.

For tests, we serialize block metadata the same way we do with the rest
of our serialization code, by writing the types directly as a char* buf.

Additionally, this commit:
    * Adds a log message indicating whether storage is New
    * ReplicaConfig duplication was removed from tests. Tests now use only
      the bftEngine::ReplicaConfig.
    * Clearer warning logs were added to BcStateTran.cpp
    * Removed debug logs from blockchain_db_adapter.cpp, since they log
      every operation and cause tests to timeout with debug logging turned on.
    * Fixed the ordering of request types in skvbc.py since they were
      reorded in the modified message code ported over.
    * Reordered the message struct fields in simpleKVBTestsBuilder to
      match the originals, to prevent having to change the order in the
      python test definitions.
    * Added a `!=` operator to Sliver.

All tests pass using both the MemoryDB and RocksDB storage backends. The
tests are currently hardcoded to use the MemoryDB backends. A future
commit will create two separate binaries so that we can run all relevant
tests for both in memory and persistent codebases.


bftengine/tests/simpleKVBC/include/block_metadata.hpp
```diff
@@ -0,0 +1,37 @@
...
+        key_(new uint8_t[1]{kBlockMetadataKey}, 1) {}
+
+  concordUtils::Sliver Key() const { return key_; }
+
+  uint64_t Get(concordUtils::Sliver &key);
+
+  concordUtils::Sliver Serialize(uint64_t bft_sequence_num);
+};
+
+}  // namespace SimpleKVBC
```
yuliasherman: This is a simplified concord_metadata_storage. I think we need to port it from concord, too, to avoid code duplication.

storage/src/blockchain_db_adapter.cpp
```diff
@@ -188,8 +188,6 @@ char KeyManipulator::extractTypeFromKey(Key _key) {
 BlockId KeyManipulator::extractBlockIdFromKey(Key _key) {
   size_t offset = _key.length() - sizeof(BlockId);
   BlockId id = *(BlockId *)(_key.data() + offset);
-
-  LOG_DEBUG(logger_, "Got block ID " << id << " from key " << _key  << ", offset " << offset);
```
yuliasherman: I would not remove those logs, they help to debug. Don't we have an additional log level, lower than DEBUG? VERBOSE or TRACK?

andrewjstone:
Thanks for the review Yulia!

PR #193, opened 2019-08-29
Let SimpleKVBCTests TesterReplica use RocksDB

andrewjstone:
Configuration of the TesterReplica was moved out into a TestSetup class.
A new command line argument `-p` was added that allows the replica to
run with peristent storage backed by RocksDB. If the -p flag is not
given, in-memory storage will be used. Additionally, if the -p flag is
given but the binary was not compiled with RocksDB support it will exit
with an error message.

This change allows new system tests to be written to test our persistent
storage code.


andrewjstone:
Thanks for the review @beerriot

PR #195, opened 2019-08-29
Fix incorrect assert

### LINKED ISSUE #194: Bug in https://github.com/vmware/concord-bft/blob/dc77051f87e4c32cc7418dc4e44530d2cbc68230/bftengine/src/bftengine/PersistentStorageImp.cpp#L875
The following assert is incorrect: https://github.com/vmware/concord-bft/blob/dc77051f87e4c32cc7418dc4e44530d2cbc68230/bftengine/src/bftengine/PersistentStorageImp.cpp#L875

You probably want to write "Assert(hasDescriptorOfLastExitFromView() || desc.view == 0)”

andrewjstone:
The prior assert would cause a crash when the view change transitioned
from view > 0 when a request was not yet executed. This scenario can
occur when a client is sending requests but not enough replicas are up
to achieve consensus. In this case they keep moving through the views,
and this assert fires.

Fix to use the correct assert, which instead of checking if there was a
descriptor for a request executing, checks that there is a descriptor
for exiting the last view. This descriptor is always saved before view
transition in ReplicaImp.cpp:1918.

Fixes #194

### DIFF
--- a/bftengine/src/bftengine/PersistentStorageImp.cpp
+++ b/bftengine/src/bftengine/PersistentStorageImp.cpp
@@ -872,7 +872,7 @@ void PersistentStorageImp::verifySetDescriptorOfLastExitFromView(const Descripto
   Assert(setIsAllowed());
   Assert(desc.view >= 0);
   // Here we assume that the first view is always 0 (even if we load the initial state from the disk).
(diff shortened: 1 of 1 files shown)


andrewjstone:
cc @guyg8

Issue #182, opened 2019-08-20
Add batching behavior to stats


beerriot:
**Is your feature request related to a problem? Please describe.**

When diagnosing performance of my cluster, one metric I'd like to have access to is how many requests the primary replica is putting into each consensus batch. That information is currently in the "Sending PrePrepareMsg" log ("requests="). It would be nice to have it somewhere more accessible, especially since that log is scheduled to be lowered from INFO to DEBUG in #179 .

**Describe the solution you'd like**

There is now a stats accumulator. It would be great if a batching metric were added there.

PR #198, opened 2019-08-30
Fix issue #182

blp:
This pair of commits fixes https://github.com/vmware/concord-bft/issues/182.

The first commit is not strictly related, but without it, simpleTest never really logs any stats at all because the timers never fire when I run it as "./simpleTest.py -l DEBUG -i 5000 -nc -bft n=6,r=6,f=1,c=1,cl=2".

The second commit really fixes the issue. The most difficult part of it was the indentation, which I still don't understand. DebugStatistics.cpp has a mix of tabs on some lines and spaces on others and I don't see how they actually line up properly with any tab width I tried. I tried to imitate what was there but all the possibilities looked wrong one way or another.

I tested this with simpleTest modified to always log debug statistics, like this:

```
diff --git a/bftengine/tests/simpleTest/simple_test_replica.hpp b/bftengine/tests/simpleTest/simple_test_replica.hpp
index 953968a58dd0..16b2f5c8da56 100644
--- a/bftengine/tests/simpleTest/simple_test_replica.hpp
+++ b/bftengine/tests/simpleTest/simple_test_replica.hpp
@@ -211,6 +211,7 @@ class SimpleTestReplica {
     replicaConfig.concurrencyLevel = 1;
     replicaConfig.debugPersistentStorageEnabled = rp.persistencyMode == PersistencyMode::InMemory ||
         rp.persistencyMode == PersistencyMode::File;
+    replicaConfig.debugStatisticsEnabled = true;
 
     // This is the state machine that the replica will drive.
     SimpleAppState *simpleAppState = new SimpleAppState(rp.numOfClients, rp.numOfReplicas);
```

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -210,12 +210,14 @@ void ReplicaImp::sendRaw(char *m, NodeIdType dest, uint16_t type, MsgSize size) 
 IncomingMsg ReplicaImp::recvMsg() {
   while (true) {
     auto msg = incomingMsgsStorage.pop(timersResolution);
-    if (msg.tag != IncomingMsg::INVALID) {
-      return msg;
-    }
+
     // TODO(GG): make sure that we don't check timers too often
     // (i.e. much faster than timersResolution)
     timersScheduler.evaluate();
+
+    if (msg.tag != IncomingMsg::INVALID) {
+      return msg;
+    }
   }
 }
 
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1,6 +1,6 @@
 //Concord
 //
-//Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+//Copyright (c) 2018, 2019 VMware, Inc. All Rights Reserved.
 //
 //This product is licensed to you under the Apache 2.0 license (the "License").  You may not use this product except in compliance with the Apache 2.0 License. 
 //
@@ -414,6 +414,9 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
 
   Assert(pp->numberOfRequests() > 0);
 
+  if (debugStatisticsEnabled) {
+    DebugStatistics::onSendPrePrepareMessage(pp->numberOfRequests(), requestsQueueOfPrimary.size());
+  }
   LOG_DEBUG_F(GL, "Sending PrePrepareMsg (seqNumber=%"
       PRId64
       ", requests=%d, size=%d",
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -210,12 +210,14 @@ void ReplicaImp::sendRaw(char *m, NodeIdType dest, uint16_t type, MsgSize size)
...
   while (true) {
     auto msg = incomingMsgsStorage.pop(timersResolution);
-    if (msg.tag != IncomingMsg::INVALID) {
-      return msg;
-    }
+
     // TODO(GG): make sure that we don't check timers too often
     // (i.e. much faster than timersResolution)
     timersScheduler.evaluate();
+
```
andrewjstone: This change makes it so that timers are evaluated on every message which we don't want. While the code isn't necessarily clear as to what `IncomingMsg::Invalid` means, we only want to check when the message is Invalid. Essentially that return code means that the thread local queue is empty and it happens when a condVar timer expires and their are no more messages to pop. You can see that this behavior follows the original intent of the code when modified to use smart pointers: https://github.com/vmware/concord-bft/commit/8583ded17c29e627244f199cbfa5f58737b4f9b5#diff-48324f7e49aa046d298836b4ccc2a318L230-L239

Please revert this change and also add a message relating to what I just said for bonus points. I'm glad you actually did this because it points out a place where the intent isn't clear. Perhaps we should also change `IncomingMsg::Invalid` to `IncomingMsg::QueueEmpty` or something.

blp: OK. I'll do that. Can you help me understand one thing first, though?  I added that particular change because in practice I saw little or no logging without it.  The commit message related to this change was:

    Prevent timer starvation in bftengine main loop under load.
    
    Before this commit, ReplicaImp::recvMsg() tries to obtain an incoming
    message and, if that fails, attempts to fire any pending timers.  If
    messages are arriving faster than the replica can process them, this means
    that timers never fire.  I actually observed this starvation in simpleTest
    running on my laptop.
    
    This commit fixes the problem by ensuring that timers are evaluated every
    time through the main loop rather than just when no messages are available.

Is there some other change that should have been made instead to avoid the timer starvation? Or is this a sign that I invoked the tests wrong, etc.?

andrewjstone: I take it back Ben. I missed the commit message by not looking at it directly and only looking at the GitHub comment. I didn't realize how intentional it was. That was my fault.  I think this is a good mechanism to avoid the timer starvation problem. Please disregard my prior comment.

andrewjstone: Theoretically we could use some sort of counter in addition to prevent timer starvation, and not have to check on every message, but I don't see a ton of harm here either.

PR #201, opened 2019-08-30
Add a regression test for #194.

andrewjstone:
This test currently fails. The fix for this issue should be checked in
before this test. It can be used to verify the fix.

The test can be run with the following command:

```
cd build && cmake -DBUILD_ROCKSDB_STORAGE=TRUE .. && make && ctest -R skvbc_persistence --verbose
```


yuliasherman:
#194? Did you mean #193?

PR #206, opened 2019-09-03
Serialization improvements

toly-kournik:
- unified serializable API functions for recursive serialization/deserialization;
- added support for std c++ containers;
- automatic class registration for deserialization, no need to provide create() and registerClass() methods;
- polymorphic getName()
- headers only
- unittests for serialization.


util/include/Serializable.h
```diff
@@ -37,72 +40,199 @@ typedef std::unique_ptr<char[], std::default_delete<char[]>> UniquePtrToChar;
...
+
+protected:
+  /** *****************************************************************************************************************
+   *  std::container
+   *  - container size
+   *  - every element of the container (recursively)
+   */
+  template<typename T, typename std::enable_if<is_std_container<T>::value, T>::type* = nullptr>
+  static void serialize_impl(std::ostream& outStream, const T& container, int) {
+   typename T::size_type size = container.size();
```
andrewjstone: When we serialize sizes, should we try to make them platform independent, by serializing as a uint64_t rather than a size_t ?

util/include/Serializable.h
```diff
@@ -37,72 +40,199 @@ typedef std::unique_ptr<char[], std::default_delete<char[]>> UniquePtrToChar;
...
+  /**
+   * convenience serialization function
+   */
+  virtual void serialize(std::ostream& outStream) const final{ serialize(outStream, *this); }
+
+protected:
+  /**
+   * the class version
+   */
+  virtual const std::string getVersion()  const = 0;
```
andrewjstone: Why a string for version instead of a uint16_t or other int?

util/include/Serializable.h
```diff
@@ -37,72 +40,199 @@ typedef std::unique_ptr<char[], std::default_delete<char[]>> UniquePtrToChar;
...
-
-  Serializable();
+public:
+  Serializable() = default;
   virtual ~Serializable() = default;
+  /**
+   * serialization API
+   */
+  template<typename T>
+  static void serialize(std::ostream& outStream, const T& t) { serialize_impl(outStream, t, int{}); }
```
andrewjstone: Pardon my ignorance on template type defs. What is the `int{}` for here?

util/include/Serializable.h
```diff
@@ -37,72 +40,199 @@ typedef std::unique_ptr<char[], std::default_delete<char[]>> UniquePtrToChar;
...
+    const std::size_t sz = str.size();
+    serialize_impl(outStream, sz, int{});
+    std::string::size_type size = str.size();
+    serialize(outStream, str.data(), size);
+  }
+  template<typename T, typename std::enable_if<std::is_same<T, std::string>::value>::type* = nullptr>
+  static void deserialize_impl(std::istream& inStream, T& t, int) {
+    std::size_t sz;
+    deserialize_impl(inStream, sz, int{});
+    char* str =  new char[sz];
```
andrewjstone: Is it possible to avoid the copy and read directly into the string?

util/include/Serializable.h
```diff
@@ -37,72 +40,199 @@ typedef std::unique_ptr<char[], std::default_delete<char[]>> UniquePtrToChar;
...
-    objectFactory_[className] = objectPtr;
+/** *******************************************************************************************************************
+ *  Class for automatic Serializable object registration
+ */
+template<typename T>
+class SerializableFactory: public virtual Serializable {
+public:
+  SerializableFactory(){(void)registered_;}
+  static bool registerT() {
+    registry()[typeid(T).name()] = []() -> Serializable* { return new T;};
```
andrewjstone: Is a `typeid` guaranteed to stay the same across compilations and across compiler versions?

toly-kournik: added name demangling

util/include/Serializable.h
toly-kournik: Here int is a third argument with no meaning, just for a signature

util/include/Serializable.h
toly-kournik: it's a size_type of a container.

util/include/Serializable.h
toly-kournik: It's how it was designed by Yulia, I don't see any problem with it.
It may be even more informative.

util/include/Serializable.h
toly-kournik: I don't know a good way of doing it.

bftengine/src/bftengine/ReplicaConfigSerializer.hpp
```diff
@@ -22,9 +22,10 @@ namespace impl {
 // class for its serialization/deserialization functionality.
 // Any ReplicaConfig changes require synchronization with this class and an
 // update of ReplicaConfigSerializer::classVersion_.
-class ReplicaConfigSerializer: public concord::serialize::Serializable {
+class ReplicaConfigSerializer: public concord::serialize::SerializableFactory<ReplicaConfigSerializer>  {
  public:
   explicit ReplicaConfigSerializer(ReplicaConfig *config);
+  ReplicaConfigSerializer();
```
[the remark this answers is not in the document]
toly-kournik: fixing in upcoming commit

util/include/Serializable.h
```diff
@@ -37,72 +41,199 @@ typedef std::unique_ptr<char[], std::default_delete<char[]>> UniquePtrToChar;
...
+   for (auto& it : container)
+     serialize(outStream, it);
+  }
+  /**
+   *  std::vector
+   *  we always deserialize containers to vector first and then copy to the designated container if needed
+   */
+  template<typename T, typename std::enable_if<is_vector<std::vector<T>>::value, T>::type* = nullptr>
+  static void deserialize_impl(std::istream& inStream, std::vector<T>& vec, int) {
+    LOG_TRACE(logger(), "");
```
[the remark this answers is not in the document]
toly-kournik: log4cplus will print the pretty function with template parameters.
for debugging

util/include/Serializable.h
andrewjstone: cool.

util/include/Serializable.h
andrewjstone: Thanks!

util/include/Serializable.h
andrewjstone: Me neither :)

util/include/demangle.hpp
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include <cxxabi.h>
+#include <memory>
+
+struct demangler {
+  static std::string demangle(const char* name) {
+      int status = -4;
+      std::unique_ptr<char, void(*)(void*)> res {abi::__cxa_demangle(name, NULL, NULL, &status), std::free};
+      return (status==0) ? res.get() : name;
```
andrewjstone: It seems wrong to just accept the mangled name if demangling fails. Shouldn't we throw an exception instead?

util/include/Serializable.h
andrewjstone: I get that. But if this data is ever transferred to a machine with a different architecture, the size_type will be different, and it may be read back incorrectly.

util/include/Serializable.h
andrewjstone: Hmm, I thought name demangling was likely good enough, but this comment is concerning. If we can't guarantee the same type name we can have very hard to find bugs.
 
> Returns an implementation defined null-terminated character string containing the name of the type. No guarantees are given; in particular, the returned string can be identical for several types and change between invocations of the same program.

https://en.cppreference.com/w/cpp/types/type_info/name

toly-kournik: demangled is just a qualified class name: [namespace::]*ClassName

util/include/demangle.hpp
toly-kournik: to tell the truth I'm not sure what's the possible scenario of such a failure when we provide compiler-specified typeid

andrewjstone: As discussed, we'll put an assert since this should never fail.

util/include/Serializable.h
andrewjstone: Yep. Agreed. I don't see how this can be a problem in practice. Let's stick with it for now, especially since we'll have an assert in case demangling fails, which we'd see immediately in tests.

PR #210, opened 2019-09-04
Add RocksDB to the travis build matrix

andrewjstone:
There are now 8 builds jobs. I tried to DRY up the config but yaml anchors didn't appear to be working.

Please look at individual commits. There were a few bug fixes found by the new builds, especially release builds with RocksDB storage enabled.

We should likely add caching and turn on verbose logging in future changes to these builds.


bftengine/tests/simpleStorage/FileStorage.cpp
```diff
@@ -133,7 +133,10 @@ bool FileStorage::initMaxSizeOfObjects(ObjectDesc *metadataObjectsArray, uint32_
   const size_t maxFileSize = fileMetadataSize + maxObjectsSize;
   LOG_INFO(logger_, "FileStorage::initMaxSizeOfObjects Maximum size of objects is "
       << maxObjectsSize << ", file size is " << maxFileSize);
-  ftruncate(fileno(dataStream_), maxFileSize);
+  if (ftruncate(fileno(dataStream_), maxFileSize)) {
+    LOG_ERROR(logger_, "Failed to truncate file: " << fileName_);
+    assert(true);
```
beerriot: I know there have been a few different assertion libraries used around here. Maybe I'm thinking of the wrong one, but isn't `assert(true)` a no-op?

andrewjstone: Whoops! Great catch! Yes, this should be `assert(false)`.

beerriot:
Re-:+1:.

PR #213, opened 2019-09-07
Add memorydb transactions to storage library

andrewjstone:
Support for strict serializable transactions was added to the in memory
db. The transactions work by taking an exclusive lock over the in-memory
map holding the keys and values when the transaction is created. Writes
(Put and Remove) are put into a shadow map inside the transaction and
only applied on transaction commit. Reads inside the transaction first
look in the shadow map to see if the transaction modified the given
key and return the value in the shadow map if so. Otherwise they go to
the underlying map via the memorydb client.

Because the transactions take an exclusive lock for simplicity,
transactions should be short lived. We don't anticipate much concurrency
or complex transactions in the near future so this should be fine.

The storage tests were also modified to test the memorydb as well as
rocksdb. The multiIO tests verify the transactions work as expected.

### DIFF
--- a/storage/test/metadataStorage_test.cpp
+++ b/storage/test/metadataStorage_test.cpp
@@ -1,15 +1,19 @@
 // Copyright 2019 VMware, all rights reserved
 /**
- * Test public functions for RocksDBMetadataStorage class.
+ * Test public functions for DBMetadataStorage class.
  */
-
-#define USE_ROCKSDB 1
 
 #include "Logger.hpp"
 #include "hash_defs.h"
 #include "gtest/gtest.h"
+
+#ifdef USE_ROCKSDB
 #include "rocksdb/key_comparator.h"
 #include "rocksdb/client.h"
+#else
+#include "memorydb/client.h"
+#endif
+
 #include "storage/db_metadata_storage.h"
 #include "blockchain/db_types.h"
 #include "blockchain/db_adapter.h"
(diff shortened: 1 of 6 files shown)


storage/test/multiIO_test.cpp
```diff
@@ -157,9 +168,16 @@ TEST(multiIO_test, no_commit_during_exception)
 
 int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
+
+#ifdef USE_ROCKSDB
   const string dbPath = "./rocksdb_test";
   dbClient = new Client(dbPath, new KeyComparator(new KeyManipulator()));
   dbClient->init();
```
beerriot: Total nit, but why not put this call outside of the ifdefs, just in case memorydb::Client::init has a non-empty body in the future?

storage/test/metadataStorage_test.cpp
```diff
@@ -100,8 +108,14 @@ int main(int argc, char **argv) {
   KeyComparator* kk = new KeyComparator(new KeyManipulator);
   const string dbPath = "./metadataStorage_test_db";
   remove(dbPath.c_str());
-  Client *dbClient = new Client(dbPath, kk);
+
+#ifdef USE_ROCKSDB
+  auto* dbClient = new concord::storage::rocksdb::Client(dbPath, kk);
   dbClient->init();
```
beerriot: Same nit as below, for future-proofing.

andrewjstone:
We aren't going to pull in the in-memory transactions as we don't need them. However, we will pull in the testing related changes in another PR.

PR #217, opened 2019-09-10
Asserts should be preceded by a log and print a call stack

yuliasherman:
Print readable call stack and log message in case of assert.
Additional Assert APIs were added - AssertOR and AssertAND, that support asserts with two conditions, e.g Assert(a>b || c==d) could be replaced by AssertOR(a>b, c==d) and value of each condition will be printed in the log.

Example of a call stack printed as a log:
[bt] bftEngine::impl::PartialProofsSet::addSelfMsgAndPPDigest(bftEngine::impl::
    PartialCommitProofMsg*, bftEngine::impl::Digest&)+0x988
[bt] bftEngine::impl::ReplicaImp::sendPartialProof(bftEngine::impl::SeqNumInfo&)+0x333
[bt] bftEngine::impl::ReplicaImp::tryToSendPrePrepareMsg(bool)+0xeed
[bt] bftEngine::impl::ReplicaImp::onMessage(bftEngine::impl::ClientRequestMsg*)+0x333
[bt] void bftEngine::impl::ReplicaImp::metaMessageHandler<bftEngine::impl::ClientRequestMsg> . 
  (bftEngine::impl::MessageBase*)+0x62
[bt] void bftEngine::impl::ReplicaImp::metaMessageHandler_IgnoreWhenCollectingState
    <bftEngine::impl::ClientRequestMsg>(bftEngine::impl::MessageBase*)+0x77
[bt] bftEngine::impl::ReplicaImp::processMessages()+0x667

Closes: VB-1568


andrewjstone:
Is it possible to get the line numbers, like in boost::stacktrace?

```
Expression 'i < N' is false in function 'T& boost::array<T, N>::operator[](boost::array<T, N>::size_type) [with T = int; long unsigned int N = 5ul; boost::array<T, N>::reference = int&; boost::array<T, N>::size_type = long unsigned int]': out of range.
Backtrace:
 0# boost::assertion_failed_msg(char const*, char const*, char const*, char const*, long) at ../example/assert_handler.cpp:39
 1# boost::array<int, 5ul>::operator[](unsigned long) at ../../../boost/array.hpp:124
 2# bar(int) at ../example/assert_handler.cpp:17
 3# foo(int) at ../example/assert_handler.cpp:25
 4# bar(int) at ../example/assert_handler.cpp:17
 5# foo(int) at ../example/assert_handler.cpp:25
 6# main at ../example/assert_handler.cpp:54
 7# 0x00007F991FD69F45 in /lib/x86_64-linux-gnu/libc.so.6
 8# 0x0000000000401139
```

yuliasherman:
It requires the use of addr2line functions that for some reason does not work for me. A function name is probably good enough to understand the flow. Regarding different kinds of assertions - yes this requires a cleanup effort.

yuliasherman:
> > It requires the use of addr2line functions that for some reason does not work for me. A function name is probably good enough to understand the flow. Regarding different kinds of assertions - yes this requires a cleanup effort.
> 
> Ok, we can fix this later. We may also want to just consider using boost or some other library for this stuff in the future. I know the past rationale, but we keep re-inventing the wheel. Something to discuss at least.

It looks like addr2line has been broken: https://bugs.launchpad.net/ubuntu/+source/binutils/+bug/1759248. 
I considered using boost stacktrace, but it requires -rdynamic linkage flag to be used. Not sure we want to.

PR #221, opened 2019-09-11
Restore default C++ compiler to clang++ when compiling.

blp:
Commit 53c6b149360f ("Enhance build via direct use of cmake (#56)") made
clang++ the default C++ compiler on Unix-like systems, but commit
b8bb68ac500c ("1. removed 3rd-party cmake FindLog4cplus for possible...")
reverted the setting.  I think that the latter was a mistake or an
oversight, because the commit message didn't mention the change and left
the comment that said that clang++ was the default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.2)
 
 # Default to building with clang on Linux,OSX,BSD etc...
 # This must come before the project definition or it won't be picked up.
+if(NOT ENV{CXX} AND UNIX)
+    set(ENV{CXX} clang++)
+endif()
+
 project(concord-bft VERSION 0.1.0.0 LANGUAGES CXX)
 
 #


andrewjstone:
> > Thanks Ben. There was a discussion on this, but this appears to be the correct change since it matches the comment. We also now have CI building both, so we shouldn't run into problems.
> 
> I guess the intention, then, is to make sure that both g++ and clang++ build OK? I was not sure.

Yep. I still think Clang is the correct default, but it doesn't really matter. We should at least remove the comment though, if we aren't going to make this change.

blp:
> > > Thanks Ben. There was a discussion on this, but this appears to be the correct change since it matches the comment. We also now have CI building both, so we shouldn't run into problems.
> > 
> > 
> > I guess the intention, then, is to make sure that both g++ and clang++ build OK? I was not sure.
> 
> Yep. I still think Clang is the correct default, but it doesn't really matter. We should at least remove the comment though, if we aren't going to make this change.

OK, either way. If you prefer deleting the comment then let me know and I'll prepare a commit to do that instead.

(I haven't used a Github pull request based workflow much before, so I goofed and deleted the branch and apparently closed the pull request prematurely.)

andrewjstone:
> > > > Thanks Ben. There was a discussion on this, but this appears to be the correct change since it matches the comment. We also now have CI building both, so we shouldn't run into problems.
> > > 
> > > 
> > > I guess the intention, then, is to make sure that both g++ and clang++ build OK? I was not sure.
> > 
> > 
> > Yep. I still think Clang is the correct default, but it doesn't really matter. We should at least remove the comment though, if we aren't going to make this change.
> 
> OK, either way. If you prefer deleting the comment then let me know and I'll prepare a commit to do that instead.
> 
> (I haven't used a Github pull request based workflow much before, so I goofed and deleted the branch and apparently closed the pull request prematurely.)

You know what, let's just delete the comment so we don't ruffle any feathers based on past discussions.

andrewjstone:
I appreciate all the help finding and cleaning this stuff up Ben.

Issue #212, opened 2019-09-05
pyclient_test failing periodically


andrewjstone:
**Describe the bug**
pyclient_tests is failing intermittently. Master build currently failed because of it.
**To Reproduce**
Run the test.
**Expected behavior**
Test passes.

andrewjstone:
This is likely the same issue seen [here](https://github.com/vmware/concord-bft/pull/135#issuecomment-515773172)

andrewjstone:
I ran `pyclient_tests` 2166 times by itself with no failures. This makes sense, since the test waits for shutdown and the seen error was a bind error. However the test failed with gcc not clang, so maybe that makes a difference. Will try with a gcc build.

andrewjstone:
If you actually look at the failure [here](https://github.com/vmware/concord-bft/pull/135#issuecomment-515773172) you'll notice the output makes it look like another tests is running simultaneously while bound to the port. That's the real issue, but I can't see how it can happen when pyclient_tests runs first and in its own VM.

andrewjstone:
After approximately 4k runs using a gcc created builld, I haven't seen the failure. I'm going to try running CI tests with verbose again since we've significantly trimmed the logs since the issue was originally seen.

andrewjstone:
Looks like we have a reproduction with logs here: https://travis-ci.com/vmware/concord-bft/jobs/235340703

andrewjstone:
I believe this is fixed. We haven't seen it in quite a long time. A bunch of fixes including #229 have been made.

PR #222, opened 2019-09-11
Run pyclient_tests in verbose mode for CI

andrewjstone:
I'm having trouble reproducing failure of this test, but it fails
periodically in CI (See #212). Since it only takes about 2 seconds to run, run it
in verbose mode so we can see the stack trace. We may want to turn on
verbose mode for all tests in the future, but we need to ensure that
doesn't slow down runs significantly or exceed the 4MB log limit. We ran
into that problem prior.


andrewjstone:
Merging as a trivial change to CI to help debug #212

PR #223, opened 2019-09-12
Reduce copying and Sliver construction/destruction in comparator

beerriot:
Profiling showed that a read/write-heavy workload spent significant
amount of time in the rocksdb key comparator, and that most of that
time was spent copying buffers, and constructing and destructing
Sliver objects. This MR adds versions of the functions required by the
key comparator that operate directly on pointers, instead of expecting
a Sliver wrapping. This removes copying and extraneous Sliver
construction in this path, which showed performance improvements in
read/write-heavy tests.

### DIFF
--- a/storage/include/blockchain/db_adapter.h
+++ b/storage/include/blockchain/db_adapter.h
@@ -20,15 +20,19 @@ namespace blockchain {
 class KeyManipulator: public IDBClient::IKeyManipulator{
  public:
   KeyManipulator():logger_(concordlogger::Log::getLogger("concord.storage.blockchain.KeyManipulator")){}
-  virtual int   composedKeyComparison(const Sliver&, const Sliver& ) override;
+  virtual int   composedKeyComparison(const uint8_t* _a_data, size_t _a_length, const uint8_t* _b_data, size_t _b_length) override;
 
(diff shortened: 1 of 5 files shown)


beerriot:
One more force-push to fix the gcc [Tool](https://en.wikipedia.org/wiki/Forty_Six_%26_2)chain mistake:

```
/home/travis/build/vmware/concord-bft/storage/src/blockchain_db_adapter.cpp:46:2: error: extra ';' [-Werror=pedantic]
 };
```

andrewjstone:
> One more force-push to fix the gcc [Tool](https://en.wikipedia.org/wiki/Forty_Six_%26_2)chain mistake:
> 
> ```
> /home/travis/build/vmware/concord-bft/storage/src/blockchain_db_adapter.cpp:46:2: error: extra ';' [-Werror=pedantic]
>  };
> ```

I like how the error is winking at you

andrewjstone:
This needs to be rebased. However, we may want to consider using string_view rather than separate pointer and length arguments. Slivers can return string_views in #228. I anticipate we can do the same elsewhere in the code and it may be a good start here, although I'm fine to merge this as is and fix it later if we choose. It also will conflict with #228, so we can also merge this in first and make the decision when #228 goes in.

beerriot:
The rebase was not difficult, so I've gone ahead and done it.

PR #227, opened 2019-09-13
Take const Sliver& where appropriate in storage

andrewjstone:
Slivers contain a std::shared_ptr<uint8_t> that points to the underlying
data buffer. While taking them by value doesn't incur any copies of the
data, it does incur an expensive atomic ref count bump. In most of our
storage methods we bump this refcount only for the parameter copy, and
then go ahead and destruct it, incurring another cost. This change makes
it so that we eliminate those operations by only taking Slivers by const
reference when we can.

Note that this doesn't actually require changing any of the call sites,
just the signatures :)


robem:
Good stuff!

PR #228, opened 2019-09-14
Enhance Sliver for safety and performance

andrewjstone:
This change consists of 4 commits to be reviewed independently.

The first updates CryptoPP to the latest version: 8.2.0. This is required because of a change to the standard library in C++ 17 that includes `std::byte` which conflicts with the global `byte` type in version 5.6.5 of CryptoPP.

The second refactors and enhances Sliver significantly. There is a detailed commit message that should be reviewed.

The third is just some code formatting.

The fourth adds a method for returning a [string_view](https://en.cppreference.com/w/cpp/string/basic_string_view) from Sliver.

### DIFF
--- a/.travis/preinstall.sh
+++ b/.travis/preinstall.sh
@@ -25,10 +25,7 @@ sudo make install
 cd /tmp
 git clone https://github.com/weidai11/cryptopp.git
 cd cryptopp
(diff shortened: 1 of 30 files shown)


beerriot:
So much cleaner, too. 👏 LGTM!

PR #229, opened 2019-09-16
Don't assert when sendto fails in UDP communication

andrewjstone:
sendto can fail for various reasons including a message too large and
the send buffer being full. We shouldn't kill the process when this
occurs.

This was seen in a CI test run:
 https://travis-ci.com/vmware/concord-bft/jobs/235340703#L2443

This is part of a fix when debugging #212, although it doesn't appear to
be the direct issue from the log linked above, since the test that
failed came after the test containing the assert.

### DIFF
--- a/bftengine/src/communication/PlainUDPCommunication.cpp
+++ b/bftengine/src/communication/PlainUDPCommunication.cpp
@@ -287,14 +287,11 @@ class PlainUDPCommunication::PlainUdpImpl {
     if (error < 0) {
       /** -1 return value means underlying socket error. */
       string err = strerror(errno);
-      LOG_DEBUG(_logger, "Error while sending: " << strerror(errno));
-      Assert(false, "Failure occurred while sending!");
-      /** Fail-fast. */
+      LOG_INFO(_logger, "Error while sending: " << strerror(errno));
     } else if (error < (int) messageLength) {
       /** Mesage was partially sent. Unclear why this would happen, perhaps
         * due to oversized messages (?). */
-      LOG_DEBUG(_logger, "Sent %d out of %d bytes!");
-      Assert(false, "Send error occurred!");    /** Fail-fast. */
+      LOG_INFO(_logger, "Sent %d out of %d bytes!");
     }
 
     if (error == (ssize_t) messageLength) {


bftengine/src/communication/PlainUDPCommunication.cpp
```diff
@@ -287,14 +287,11 @@ class PlainUDPCommunication::PlainUdpImpl {
...
-      LOG_DEBUG(_logger, "Error while sending: " << strerror(errno));
-      Assert(false, "Failure occurred while sending!");
-      /** Fail-fast. */
+      LOG_INFO(_logger, "Error while sending: " << strerror(errno));
     } else if (error < (int) messageLength) {
       /** Mesage was partially sent. Unclear why this would happen, perhaps
         * due to oversized messages (?). */
-      LOG_DEBUG(_logger, "Sent %d out of %d bytes!");
-      Assert(false, "Send error occurred!");    /** Fail-fast. */
+      LOG_INFO(_logger, "Sent %d out of %d bytes!");
```
upshaw-alex: With your new changes, in cases where a communication error occurs, it looks like the behavior of this UDP communication will just be to log a statement mentioning the failure, but (correct me if I am wrong) it does not look like anything will happen that could inform layers above this one in Concord-BFT of the failure. Looking below these changes, it looks like `sendAsyncMessage` will still return `0` (I assume `sendAsyncMessage`'s returned `int` is supposed to be some sort of error code, though it is not completely clear just looking at the surrounding failures, as, from its caller's perspective, `sendAsyncMessage`'s existing behaviors only appear to include returning `0` or crashing the program). Should `sendAsyncMessage` be returning something other than `0` if the send failed? Should it be throwing an exception if the send failed?

On a related note, have you done any testing of cases where this send fails to see if layers above handle the failure reasonably?

salieri11: The caller of sendAsyncMessage doesn't check the result as for the moment. This removal is natural since in any BFT system the sender can't assume anything about the message's fate, that's the reason the return status is not checked.

andrewjstone: yup.

upshaw-alex: Thank you for the clarification; that makes sense; it occurs to me there shouldn't be too much distinction for correctness purposes between the case where we fail to send a message at all and the case where it gets lost by the network before reaching its destination.

We might want to recognize the difference when trying to debug or troubleshoot connectivity issues, though I guess that is why we have the log messages, and I guess it doesn't make sense to provide better programmatic error reporting if we aren't adding any logic to respond to it.

PR #230, opened 2019-09-19
Disable Nagle's algorithm for the TLS module

salieri11:
Set TCP_NODELAY to TRUE on both sender and receiver side to reduce latency

### DIFF
--- a/bftengine/src/communication/TlsTCPCommunication.cpp
+++ b/bftengine/src/communication/TlsTCPCommunication.cpp
@@ -243,6 +243,13 @@ class AsyncTlsConnection : public
     return it->second.isReplica;
   }
 
+  void set_no_delay() {
+    asio::ip::tcp::no_delay option;
+    get_socket().set_option(boost::asio::ip::tcp::no_delay(true));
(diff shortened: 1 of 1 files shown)


bftengine/src/communication/TlsTCPCommunication.cpp
```diff
@@ -243,6 +243,13 @@ class AsyncTlsConnection : public
     return it->second.isReplica;
   }
 
+  void set_no_delay() {
+    asio::ip::tcp::no_delay option;
+    get_socket().set_option(boost::asio::ip::tcp::no_delay(true));
+    get_socket().get_option(option);
```
andrewjstone: Is there a reason to think that the option wouldn't get set with set_option, such that we have to check it?

andrewjstone: I'm fine merging this if it's strictly required that nagle is supported to use. What I mean is that we should actually crash if it fails to set. Seems like that's the intent here so +1.

andrewjstone:
LGTM!

Issue #232, opened 2019-09-23
Implement TLS support for python test framework


salieri11:
Currently python test client (the ```bft_client.py```) supports only UDP communication to the concord replicas.
Need to implement TLS support, either with Trio or natively using Python SSL library.
Important points:
 * we use certificate pinning and we trust only our certificates. The solution should support custom verification callback
* currently, the python client should connect to all replicas (in other words, it should be the "Client" in the TCP world)

jaybny:
am currently implementing this feature

andrewjstone:
This has been implemented.

PR #233, opened 2019-09-23
WIP: Add simple log4cpp config file, copy TLS certificates to test folders

salieri11:
This PR adds 2 small improvements for running tests:
 * Simple log4cpp config file to silence "noisy" loggers when working in DEBUG mode. To enable log4cpp, build with ```-DUSE_LOG4CPP``` and the configuration file will be copied to all test folders
 * TLS certificates are copied to all manual test folders, and if compiled with ```BUILD_COMM_TCP_TLS``` - manual tests will work with TLS as well.

** IMPORTANT ** Tests that are written with the python test framework will not work with TLS until the test client will not implement TLS support (see #232)

### DIFF
--- a/.travis/preinstall.sh
+++ b/.travis/preinstall.sh
@@ -46,5 +46,17 @@ if [ -n "$USE_ROCKSDB" ]; then
     sudo make install-shared
 fi
 
+if [-n "$USE_LOG4CPP" ]; then
+  #Install log4cpp
+  sudo apt-get install autoconf automake
+  cd $TRAVIS_BUILD_DIR
+  git clone https://github.com/log4cplus/log4cplus.git
+  cd log4cplus
+  git checkout REL_1_2_1
+  ./configure CXXFLAGS="--std=c++11"
+  make
+  sudo make install
+fi
+
 # trio is need for tests
 python3 -m pip install --upgrade trio
--- a/.travis/preinstall.sh
+++ b/.travis/preinstall.sh
@@ -46,17 +46,5 @@ if [ -n "$USE_ROCKSDB" ]; then
     sudo make install-shared
 fi
 
-if [-n "$USE_LOG4CPP" ]; then
-  #Install log4cpp
-  sudo apt-get install autoconf automake
-  cd $TRAVIS_BUILD_DIR
-  git clone https://github.com/log4cplus/log4cplus.git
-  cd log4cplus
-  git checkout REL_1_2_1
-  ./configure CXXFLAGS="--std=c++11"
-  make
-  sudo make install
-fi
-
 # trio is need for tests
 python3 -m pip install --upgrade trio
--- a/bftengine/CMakeLists.txt
+++ b/bftengine/CMakeLists.txt
@@ -106,3 +106,30 @@ target_include_directories(corebft PUBLIC include/metadatastorage)
 target_link_libraries(corebft PUBLIC threshsign)
 target_link_libraries(corebft PUBLIC Threads::Threads)
 target_link_libraries(corebft PUBLIC util)
+
+# copy certificate creation script to all tests
+add_custom_target(copy_tls_certificates ALL COMMENT "Copying TLS certificates")
+add_custom_command(TARGET copy_tls_certificates
+        COMMAND ${CMAKE_COMMAND} -E copy_directory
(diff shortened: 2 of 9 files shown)


andrewjstone:
CI is currently failing with the new CMake recipes. While these logging changes are immediately useful we don't currently build with LOG4CPLUS at all in CI. Adding the usage of LOG4CPLUS and not using it would bring our jobs up from 8 - 16. We may want to make it so that LOG4CPLUS is always used in CI to keep our jobs at 8. This brings up the question of why we even maintain our own logger though. The tradeoff is that every optional dependency exponentially increases our possible build targets. Mandatory deps do not.

I really think we should consider removing our homegrown logging implementation and creating a generic API. We can then just back that by LOG4CPLUS. If we decide to switch loggers later, we still can by keeping to our API. This would allow us to use things like named loggers in all cases and not have to conditionalize builds and code. Since a logger is always needed, I don't think it's unreasonable to mandate the dependency. Our homegrown logger is never going to be as full-featured as LOG4CPLUS or other production loggers and it makes debugging and testing harder when we have to consider both.

.travis/preinstall.sh
```diff
@@ -46,5 +46,17 @@ if [ -n "$USE_ROCKSDB" ]; then
     sudo make install-shared
 fi
 
+if [-n "$USE_LOG4CPP" ]; then
```
andrewjstone: This won't work. The variable will never be set because it does not come from CMAKE. These environment variables come from .travis.yml.

See my comment regarding that we don't actually build with LOG4CPLUS at all in CI.

salieri11: THANKS! I will remove it from the build.

salieri11:
> CI is currently failing with the new CMake recipes. While these logging changes are immediately useful we don't currently build with LOG4CPLUS at all in CI. Adding the usage of LOG4CPLUS and not using it would bring our jobs up from 8 - 16. We may want to make it so that LOG4CPLUS is always used in CI to keep our jobs at 8. This brings up the question of why we even maintain our own logger though. The tradeoff is that every optional dependency exponentially increases our possible build targets. Mandatory deps do not.
> 
> I really think we should consider removing our homegrown logging implementation and creating a generic API. We can then just back that by LOG4CPLUS. If we decide to switch loggers later, we still can by keeping to our API. This would allow us to use things like named loggers in all cases and not have to conditionalize builds and code. Since a logger is always needed, I don't think it's unreasonable to mandate the dependency. Our homegrown logger is never going to be as full-featured as LOG4CPLUS or other production loggers and it makes debugging and testing harder when we have to consider both.

Well our logger is just very basic and in fact it is already the wrapper for any other logging system. I think the API is pretty good, in fact it was copied from log4cpp.

andrewjstone:
> > CI is currently failing with the new CMake recipes. While these logging changes are immediately useful we don't currently build with LOG4CPLUS at all in CI. Adding the usage of LOG4CPLUS and not using it would bring our jobs up from 8 - 16. We may want to make it so that LOG4CPLUS is always used in CI to keep our jobs at 8. This brings up the question of why we even maintain our own logger though. The tradeoff is that every optional dependency exponentially increases our possible build targets. Mandatory deps do not.
> > I really think we should consider removing our homegrown logging implementation and creating a generic API. We can then just back that by LOG4CPLUS. If we decide to switch loggers later, we still can by keeping to our API. This would allow us to use things like named loggers in all cases and not have to conditionalize builds and code. Since a logger is always needed, I don't think it's unreasonable to mandate the dependency. Our homegrown logger is never going to be as full-featured as LOG4CPLUS or other production loggers and it makes debugging and testing harder when we have to consider both.
> 
> Well our logger is just very basic and in fact it is already the wrapper for any other logging system. I think the API is pretty good, in fact it was copied from log4cpp.

It is a bit too basic I think. For instance there is no way to configure the log level at run time. In fact, you can't configure it without changing a constant and recompiling. Because of the use of globals, fixing this is non-trivial. You also can't turn on and off logging for different named components, which is a requirement in production. These 2 things make it essentially useless and will drive everyone to use LOG4CPLUS anyway. We can certainly spend time fixing them and implementing a full featured logger, but why?

teoparvanov:
Closing this PR as it has become stale.

PR #234, opened 2019-09-23
Log a msg when Checkpoint summaries do not match

andrewjstone:
This is useful to detect non-deterministic state machines and malicious
nodes. This is the minimum possible change so we can use it immediately.
More robust detection and alerting strategies can be implemented in the
future.

### DIFF
--- a/bftengine/src/bcstatetransfer/Messages.hpp
+++ b/bftengine/src/bcstatetransfer/Messages.hpp
@@ -18,6 +18,7 @@
 
 #include "STDigest.hpp"
 #include "IStateTransfer.hpp"
+#include "Logger.hpp"
 
 namespace bftEngine {
 namespace SimpleBlockchainStateTransfer {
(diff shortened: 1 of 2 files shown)


bftengine/src/bcstatetransfer/Messages.hpp
```diff
@@ -64,11 +65,28 @@ struct CheckpointSummaryMsg : public BCStateTranBaseMsg {
...
-                         const CheckpointSummaryMsg* b) {
+  static bool equivalent(const CheckpointSummaryMsg* a, const CheckpointSummaryMsg* b) {
     return (memcmp(a, b, sizeof(CheckpointSummaryMsg)) == 0);
   }
 
+  static bool equivalent(const CheckpointSummaryMsg* a, uint16_t a_id,
+                         const CheckpointSummaryMsg* b, uint16_t b_id) {
+    if (memcmp(a, b, sizeof(CheckpointSummaryMsg)) != 0) {
+      auto logger = concordlogger::Log::getLogger("state-transfer");
+      LOG_INFO(logger, "Mismatched Checkpoints for checkpointNum=" << a->checkpointNum  << std::endl
```
robem: This will help debugging a lot but I think the log level should be either warning or error.

andrewjstone: I'll change it to Warning

andrewjstone:
I changed the log level to LOG_WARN and force pushed the change.

PR #235, opened 2019-09-24
Add metrics to BCStateTran

andrewjstone:
This commit adds and initializes and updates metrics for BCStateTran.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -11,35 +11,27 @@
 // terms and conditions of the subcomponent's license, as noted in the LICENSE
 // file.
 
-#include <cassert>
 #include <algorithm>
 #include <chrono>
-#include <iostream>
 #include <list>
 #include <set>
 #include <string>
+#include <sstream>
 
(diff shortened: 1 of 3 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -153,13 +145,77 @@ BCStateTran::BCStateTran(
     maxNumOfChunksInVBlock_{calcMaxNumOfChunksInBlock(maxItemSize_, config.maxBlockSize, config.maxChunkSize, true)},
     maxNumOfStoredCheckpoints_{0},
     numberOfReservedPages_{0},
-    randomGen_{randomDevice_()} {
+    randomGen_{randomDevice_()},
+    metrics_component_{concordMetrics::Component("bc_state_transfer",
+                                                  std::make_shared<concordMetrics::Aggregator>())},
+
+    // We must make sure that we actually initialize all these metrics in the
+    // correct order!
```
beerriot: I'd suggest replacing "correct order" with "same order as defined in the header file", to make the intention clearer.

andrewjstone: Fixed and rebased.

PR #243, opened 2019-10-14
Fix release build

no2chem:
This PR fixes several unused variable errors when concord-bft is built in release mode.

This is due to the use of variables which are only used in assert() statements.
To turn off the warning, I've added ``__attribute__((unused))`` to these variables to prevent the errors from being emitted, which appears to be the preferred way to suppress this issue without C++17's ``[[maybe_unused]]`` attribute.

In the future, CI should build both for debug and release.

### DIFF
--- a/bftengine/src/communication/TlsTCPCommunication.cpp
+++ b/bftengine/src/communication/TlsTCPCommunication.cpp
@@ -716,7 +716,7 @@ class AsyncTlsConnection : public
                            bool first) {
     // if first is true - we came from partial reading, no timer was started
     if(!first) {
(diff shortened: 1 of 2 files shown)


no2chem:
@andrewjstone ah, yeah, forgot about -NDEBUG. Good to know, might use that to get the release build in the concord repository working, but it's always good to have the vanilla release build work as well.

andrewjstone:
> it's always good to have the vanilla release build work as well.

Yes indeed. This is a useful patch. Our users will not get confused or deterred this way! Much appreciated.

PR #245, opened 2019-10-15
Allow hostnames/fqdns instead of IP addresses for PlainTCP and TLSTCP

beerriot:
The main point of this PR is the first commit, bb0343a, which allows a hostname, instead of an IP, to be passed in the communication config for the TCP and TLS modules. This was not added to the UDP module, because it would have required quite a bit more change there. From the commit message:

    If a hostname is used for local principals, that hostname must be
    resolvable at startup time. If it is not, the node will fail to start,
    after throwing an exception while trying to set up its listening
    sockets.
    
    If a hostname for any remote principal is not resolvable at startup
    time, it is treated as any other connection failure and retried later.

The second commit, e85e3d1, is just running clang-format over the communication code.

### DIFF
--- a/bftengine/include/communication/CommDefs.hpp
+++ b/bftengine/include/communication/CommDefs.hpp
@@ -18,6 +18,7 @@
 #include <ws2def.h>
 #else
 #include <arpa/inet.h>
+#include <netdb.h>
 #include <sys/socket.h>
 #include <unistd.h>
 #endif
@@ -32,7 +33,7 @@
 typedef struct sockaddr_in Addr;
 
 struct NodeInfo {
-  std::string ip;
+  std::string host;
   std::uint16_t port;
   bool isReplica;
 };
@@ -50,7 +51,7 @@ enum CommType {
 
 struct BaseCommConfig {
   CommType commType;
-  std::string listenIp;
+  std::string listenHost;
   uint16_t listenPort;
   uint32_t bufferLength;
   NodeMap nodes;
@@ -58,14 +59,14 @@ struct BaseCommConfig {
   NodeNum selfId;
 
   BaseCommConfig(CommType type,
-                 std::string ip,
+                 std::string host,
                  uint16_t port,
                  uint32_t bufLength,
                  NodeMap _nodes,
                  NodeNum _selfId,
                  UPDATE_CONNECTIVITY_FN _statusCallback = nullptr) :
       commType{type},
-      listenIp{std::move(ip)},
+      listenHost{std::move(host)},
       listenPort{port},
       bufferLength{bufLength},
       nodes{std::move(_nodes)},
@@ -77,14 +78,14 @@ struct BaseCommConfig {
 };
 
 struct PlainUdpConfig : BaseCommConfig {
-  PlainUdpConfig(std::string ip,
+  PlainUdpConfig(std::string host,
                  uint16_t port,
                  uint32_t bufLength,
                  NodeMap _nodes,
                  NodeNum _selfId,
                  UPDATE_CONNECTIVITY_FN _statusCallback = nullptr) :
       BaseCommConfig(CommType::PlainUdp,
-                     std::move(ip),
+                     std::move(host),
                      port,
                      bufLength,
                      std::move(_nodes),
@@ -96,7 +97,7 @@ struct PlainUdpConfig : BaseCommConfig {
 struct PlainTcpConfig : BaseCommConfig {
   int32_t maxServerId;
 
-  PlainTcpConfig(std::string ip,
+  PlainTcpConfig(std::string host,
                  uint16_t port,
                  uint32_t bufLength,
                  NodeMap _nodes,
@@ -104,7 +105,7 @@ struct PlainTcpConfig : BaseCommConfig {
                  NodeNum _selfId,
                  UPDATE_CONNECTIVITY_FN _statusCallback = nullptr) :
       BaseCommConfig(CommType::PlainTcp,
-                     std::move(ip),
+                     std::move(host),
                      port,
                      bufLength,
                      std::move(_nodes),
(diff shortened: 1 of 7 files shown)


beerriot:
Rebased to fix the conflict with #243

bftengine/src/communication/PlainTcpCommunication.cpp
```diff
@@ -561,32 +561,46 @@ class AsyncTcpConnection :
...
-              << ", ip: " << ip
+              << ", host: " << host
               << ", port: " << port);
 
-    tcp::endpoint ep(address::from_string(ip), port);
-    LOG_DEBUG(_logger, "connecting from: " << _selfId
-              << " ,to: " << _destId
-              << ", timeout: " << _currentTimeout
-              << ", dest is replica: " << _destIsReplica);
+    asio::ip::tcp::resolver::query query(asio::ip::tcp::v4(), _host, std::to_string(_port));
```
andrewjstone: It seems pretty much all of these mechanisms are deprecated. I'm not sure how to do things in the recommended manner, but we should probably look into it. https://www.boost.org/doc/libs/1_71_0/doc/html/boost_asio/reference/ip__tcp/resolver.html

beerriot: These are not deprecated in the version of boost that README documents: https://www.boost.org/doc/libs/1_64_0/doc/html/boost_asio/reference/ip__tcp/resolver.html

Nor are they deprecated in the version of boost we actually use: https://www.boost.org/doc/libs/1_65_0/doc/html/boost_asio/reference/ip__tcp/resolver.html

🤷‍♂

bftengine/src/communication/PlainTcpCommunication.cpp
```diff
@@ -715,100 +629,94 @@ class PlainTCPCommunication::PlainTcpImpl {
...
+      if (!ec && results != asio::ip::tcp::resolver::iterator()) {
+        asio::ip::tcp::endpoint ep = *results;
+        LOG_INFO(_logger, "Resolved " << _listenHost << ":" << _listenPort << " to " << ep);
+        _pAcceptor = boost::make_unique<tcp::acceptor>(_service, ep);
+        start_accept(nodes);
+      } else {
+        LOG_WARN(_logger, "Unable to resolve listen host (" << _listenHost << ") for node " << _selfId << ": " << ec);
+        // This node is dead in the water if it can't resolve its own listen
+        // host/ip.
+        return -1;  // failed
```
beerriot: Self review: this is a copy-paste-o from an earlier version of `TcpTlsImpl::Start`. The comment at the start is wrong - that version of `resolve` does not throw an exception. Since this is in the constructor here, it also can't return -1.

I feel like reducing duplication between the plain TCP and TLS/TCP implementations would help prevent this sort of mistake in the future. That may require changing the build to allow building both at the same time.

bftengine/src/communication/PlainTcpCommunication.cpp
upshaw-alex: I guess the deprecation could become a problem if we upgrade Boost versions. Does anyone here have thoughts on whether upgrading Boost versions is something that is likely to happen in the future?

andrewjstone: Just because we are using an old version doesn't mean the interfaces aren't deprecated. We don't have any immediate plans to upgrade, but if we upgraded we'd likely upgrade to the latest version. If when we do that, the deprecation has taken effect and the methods are removed our code will break. We can push this off until then, when some poor soul will have to fix it, likely due to an unrelated need in a new version of boost.

andrewjstone: I'm fine merging this in FYI, as long as we all understand this. This is a *very* useful patch.

upshaw-alex: If it is decided to go through with merging this as is and we are concerned about the burden we are creating for some poor soul, I would recommend we immediately create an issue upon merging this to track revising the code to remove the use of deprecated things.

beerriot: I've done more investigation, and we have the standard deprecation problem. The function we're using, `resolve(query&)` is deprecated in a future version, but its replacement `resolve(protocol_type&, string_view, string_view)` is not yet available in the version we're using. So, if we don't want to implement a non-deprecated solution now, we also have to upgrade the boost version now.

The change, when it is needed, should be small: replace the `query` argument to the resolve function with the `ip::tcp::v4`, `_listenHost`, and `_listenPort` arguments that are currently used to construct that query object. What do y'all think about a comment in this code mentioning this, for your "poor soul's" reference?

andrewjstone: Thanks for investigating Bryan! Let's keep the version as is for now to minimize churn. Let me know when you think this PR is ready to merge.

beerriot: Fixed in bb0343a.

bftengine/src/communication/PlainTcpCommunication.cpp
beerriot: Fixed in bb0343a.

bftengine/include/communication/CommDefs.hpp
```diff
@@ -50,22 +51,22 @@ enum CommType {
...
+  std::string listenHost;
   uint16_t listenPort;
   uint32_t bufferLength;
   NodeMap nodes;
   UPDATE_CONNECTIVITY_FN statusCallback;
   NodeNum selfId;
 
   BaseCommConfig(CommType type,
-                 std::string ip,
+                 std::string host,
```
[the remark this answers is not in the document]
beerriot: Assuming you don't want to hold this PR up over this, I've filed an issue to track it: https://github.com/vmware/concord-bft/issues/253

beerriot:
@andrewjstone 
> Let me know when you think this PR is ready to merge.

Ready.

PR #246, opened 2019-10-15
Add use of AssertAND and AssertOR for a better error reporting

yuliasherman:
Write an error log message and a stack trace for asserts that have two conditions : (expr1 && expr2)  or (expr1 || expr2).

Closes VB-1826


yuliasherman:
MR against a wrong parent.

PR #252, opened 2019-10-22
Write State Transfer blocks and their keys atomically

yuliasherman:
State Transfer blocks should be written together with their keys using one RocksDB transaction like it is done for regular blocks.

Closes VB-1831


andrewjstone:
Nice find!

PR #254, opened 2019-10-23
State Transfer bugs fixes cont’d:

toly-kournik:
1. Restarting replica during state transfer in process
   - Before first state transfer checkpoint reached - don't load reserved pages
   - loadReplicaData() - fix index in seqNumWinArray

2. ClientsManager::allocateNewReplyMsgAndWriteToStorage()
   - allow equality in assert in favour of crash recovery

3. ReplicaImpl
  -  not to use setFetchingState() as no one eventually reads this value as it’s set in state transfer functionality.

4. Re-enable skvbc_persistence_tests

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1384,12 +1384,6 @@ void ReplicaImp::onMessage(CheckpointMsg *msg) {
   if (askForStateTransfer) {
     LOG_INFO_F(GL, "call to startCollectingState()");
 
-    if (ps_) {
-      ps_->beginWriteTran();
-      ps_->setFetchingState(true);
-      ps_->endWriteTran();
-    }
-
     stateTransfer->startCollectingState();
   } else if (msgSeqNum > lastStableSeqNum + kWorkWindowSize) {
     onReportAboutAdvancedReplica(msgSenderId, msgSeqNum);
@@ -2202,9 +2196,8 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
 
   if (ps_) {
     ps_->beginWriteTran();
-    ps_->setFetchingState(false);
   }
-
+    
   if (newStateCheckpoint <= lastExecutedSeqNum) {
     LOG_DEBUG_F(GL,
                "Executing onTransferringCompleteImp(newStateCheckpoint) where newStateCheckpoint <= lastExecutedSeqNum");
@@ -2278,12 +2271,6 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
 
   if (askAnotherStateTransfer) {
     LOG_INFO_F(GL, "call to startCollectingState()");
-
-    if (ps_) {
-      ps_->beginWriteTran();
-      ps_->setFetchingState(true);
-      ps_->endWriteTran();
-    }
 
     stateTransfer->startCollectingState();
   }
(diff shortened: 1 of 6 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1384,12 +1384,6 @@ void ReplicaImp::onMessage(CheckpointMsg *msg) {
   if (askForStateTransfer) {
     LOG_INFO_F(GL, "call to startCollectingState()");
 
-    if (ps_) {
```
andrewjstone: If we are no longer using `setFetchingState` we should remove it and `getFetchingState` completely in a future commit.

```
andrewstone@ubuntu:~/concord-bft$ grep getFetchingState bftengine/src/bftengine/*
bftengine/src/bftengine/DebugPersistentStorage.cpp:bool DebugPersistentStorage::getFetchingState() {
bftengine/src/bftengine/DebugPersistentStorage.hpp:  bool getFetchingState() override;
bftengine/src/bftengine/PersistentStorage.hpp:  virtual bool getFetchingState() = 0;
bftengine/src/bftengine/PersistentStorageImp.cpp:bool PersistentStorageImp::getFetchingState() {
bftengine/src/bftengine/PersistentStorageImp.hpp:  bool getFetchingState() override;
andrewstone@ubuntu:~/concord-bft$ grep setFetchingState bftengine/src/bftengine/*
bftengine/src/bftengine/DebugPersistentStorage.cpp:void DebugPersistentStorage::setFetchingState(bool state) {
bftengine/src/bftengine/DebugPersistentStorage.hpp:  void setFetchingState(bool state) override;
bftengine/src/bftengine/PersistentStorage.hpp:  virtual void setFetchingState(bool state) = 0;
bftengine/src/bftengine/PersistentStorageImp.cpp:  setFetchingStateInternal(fetchingState_);
bftengine/src/bftengine/PersistentStorageImp.cpp:void PersistentStorageImp::setFetchingStateInternal(uint8_t state) {
bftengine/src/bftengine/PersistentStorageImp.cpp:void PersistentStorageImp::setFetchingState(bool state) {
bftengine/src/bftengine/PersistentStorageImp.cpp:  setFetchingStateInternal(state);
bftengine/src/bftengine/PersistentStorageImp.hpp:  void setFetchingState(bool state) override;
bftengine/src/bftengine/PersistentStorageImp.hpp:  void setFetchingStateInternal(uint8_t state);
bftengine/src/bftengine/ReplicaImp.cpp:      ps_->setFetchingState(true);
bftengine/src/bftengine/ReplicaImp.cpp:    ps_->setFetchingState(false);
bftengine/src/bftengine/ReplicaImp.cpp:      ps_->setFetchingState(true);
```

bftengine/src/bftengine/ReplicaLoader.cpp
```diff
@@ -337,7 +337,7 @@ ReplicaLoader::ErrorCode loadReplicaData(shared_ptr<PersistentStorage> p, Loaded
       Verify(d.executedSeqNum > ld.lastStableSeqNum, InconsistentErr);
       Verify(d.executedSeqNum <= ld.lastStableSeqNum + kWorkWindowSize, InconsistentErr);
 
-      uint64_t idx = d.executedSeqNum - ld.lastStableSeqNum;
+      uint64_t idx = d.executedSeqNum - ld.lastStableSeqNum - 1;
```
andrewjstone: Why was this change made?

if `d.executedSeqNum == 0 && ld.lastStableSeqNum == 0` then `idx` will underflow become UINT64_MAX and the Assert below will fire.

andrewjstone: Nevermind the last part of this. I see from the surrounding code (not shown in the PR) that this is safe as `d.executedSeqNum` must be greater than `ld.lastStableSeqNum.`

I'm still curious as to why this change is needed though. I just want to understand. I don't doubt it is correct.

bftengine/src/bftengine/ReplicaImp.cpp
toly-kournik: I agree

bftengine/src/bftengine/ReplicaLoader.cpp
toly-kournik: It just what we saw during the debugging. @guyg8  confirmed.

PR #259, opened 2019-10-29
Additional tests for persistence & ST

teoparvanov:
This PR adds a system test that triggers state transfer in the 
persistence context, while crashing both the fetcher and the sender.

I have also factored-out some common test logic, cleaned-up the existing tests and added several asserts.

Attached, logs from OK runs, as well as from intermittent test failures:
[persistence_with_st_logs.zip](https://github.com/vmware/concord-bft/files/3789488/persistence_with_st_logs.zip)

### DIFF
--- a/test/test_skvbc_persistence.py
+++ b/test/test_skvbc_persistence.py
@@ -17,8 +17,8 @@ import random
 
 import bft_tester
 import skvbc
-import bft_client
-import bft_metrics_client
+
+from math import inf
 
 KEY_FILE_PREFIX = "replica_keys_"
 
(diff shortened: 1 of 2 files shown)


test/test_skvbc_persistence.py
```diff
@@ -338,7 +308,7 @@ def test_st_when_fetcher_crashes(self):
 
         # Wait for initial replicas to take 2 checkpoints (exhausting
         # the full window)
-        checkpoint_num = 2;
+        checkpoint_num = 2
```
andrewjstone: nice catch.

teoparvanov:
I think the problem is in the _PlainUdpImpl::sendAsyncMessage()_ function. We actually can't know the destination is down, so _BCStateTran_ keeps on doing _sendFetchBlockMsg()_, even though the source node is not available. 

One option would be to reduce the _config.sourceReplicaReplacementTimeoutMilli_ property, by default set to 10 min.
I tried with _config.sourceReplicaReplacementTimeoutMilli = 1000 ms_ (1sec) and the test now succeeds consistently.

bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
```diff
@@ -111,7 +111,7 @@ struct Config {
   uint32_t refreshTimerMilli = 300; // ms
   uint32_t checkpointSummariesRetransmissionTimeoutMilli = 2500; // ms
   uint32_t maxAcceptableMsgDelayMilli = 60000; // 1 minute
-  uint32_t sourceReplicaReplacementTimeoutMilli = 600000; // 10 minutes
+  uint32_t sourceReplicaReplacementTimeoutMilli = 1000; // 1 second
```
teoparvanov: This mitigates the problem that when replicas communicate via UDP, the fetcher cannot know the source replica is down. Changing the source replica frequently solves this problem and, IMHO does not introduce overhead.

PR #264, opened 2019-10-31
Fix intermittent persistence + ST test failure

teoparvanov:
Here we fix an intermittent failure of test_st_when_fetcher_and_sender_crash. The failure is due to the fetcher sending UDP messages to the ST sender node without ever knowing whether the sender node might be down. To remediate this, we reduce significantly the sourceReplicaReplacementTimeoutMilli timer (by default set to 10 min).

This way, sender replicas are rotated more frequently and we avoid the situation described above.

To confirm this change actually fixes the bug, I have run 10 consecutive times the SKVBC persistence test suite which would previously fail intermittently (every 3-4 runs):
``for i in `seq 10`; do ctest -R skvbc_persistence_tests; done``


teoparvanov:
Hi Andrew, I've rebased the branch and now it only adds a single commit.

Best regards,
Teo

andrewjstone:
> Hi Andrew, I've rebased the branch and now it only adds a single commit.
> 
> Best regards,
> Teo



> Hi Andrew, I've rebased the branch and now it only adds a single commit.
> 
> Best regards,
> Teo

Excellent. Thanks.

PR #266, opened 2019-11-01
Release the TransactionDB instead of the BaseDB

beerriot:
When the RocksDB client is created, if it's not in read-only mode, we
create a TransactionDB, and then grab the base DB pointer out of
it. Until now, we had always been releasing only the base DB. This
patch releases the TransactionDB instead, if it exists.


beerriot:
Whoops - missed a variable rename in my rebase from my test branch. Fixed in fc00336.

Issue #269, opened 2019-11-04
State transfer does not complete when crashing primary replica


teoparvanov:
**Describe the bug**
When crashing the primary in the presence of a stale replica, state transfer (reserved pages) does not complete.

**To Reproduce**
1) Start 3 nodes out of a 4 node cluster. Write a specific key, then enough data to the cluster to trigger several checkpoints.
2) Stop the primary (do NOT trigger view change)
3) Start the stale node, which should start fetching/state transfer
4) [EXPECTED] State transfer completes for transaction blocks, but loops infinitely for reserved pages
4) [OBSERVED] State transfer never completes

I suspect the following logic to be related:
https://github.com/vmware/concord-bft/blob/master/bftengine/src/bcstatetransfer/BCStateTran.cpp#L1901

PS: I have attached logs which demonstrate the issue. The following snippet seems to repeat a lot:
INFO |state-transfer|void bftEngine::SimpleBlockchainStateTransfer::impl::BCStateTran::processData()|state is GettingMissingResPages|
INFO |state-transfer|void bftEngine::SimpleBlockchainStateTransfer::impl::BCStateTran::processData()|nextRequiredBlock_=18446744073709551615 

**Expected behavior**
State transfer should be able to complete with only 3 up-to-date nodes, regardless of whether view change has completed.

**Screenshots**
N/A - attaching logs
[persistence_with_st_and_crashing_primary.zip](https://github.com/vmware/concord-bft/files/3804805/persistence_with_st_and_crashing_primary.zip)

PR #270, opened 2019-11-05
Test for persistence + ST + view change

teoparvanov:
This PR adds two tests combining persistence, state transfer and view change:

1) Crash primary once (without forcing view change), then trigger state transfer
This test is marked as "skipped" due to an issue it surfaces, tracked under https://github.com/vmware/concord-bft/issues/269

2) Repeatedly trigger view change while interrupting state transfer
In this test, we bring down the primary, trigger view change and repeatedly restart the stale node in the process.

At the end we ensure that state transfer has completed and the value initially written to the incomplete cluster can be
read via a quorum including the stale node.

Both tests run on a 7 node cluster (f=2), thus tolerating both the stale and primary replicas being down simultaneously.

PS: I have also simplified the logic in wait_for_fetching_state() which had become overly complex

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,5 @@ env:
     - RELEASE=-DCMAKE_BUILD_TYPE=RELEASE
     - RELEASE=-DCMAKE_BUILD_TYPE=RELEASE USE_ROCKSDB=-DBUILD_ROCKSDB_STORAGE=TRUE
 script:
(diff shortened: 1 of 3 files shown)


teoparvanov:
I am currently skipping test_st_when_primary_crashes() as it consistently passes locally, but fails in CI for some reason... I'm looking into it, but even in this state, the PR adds value because we can test Persistence + ST + view change automatically instead of manually.

teoparvanov:
After rerunning the RocksDB jobs several times, I now feel confident for merging this PR.

PR #276, opened 2019-11-18
KVBC code deduplication

toly-kournik:
Move simpleKVBC from tests to separate top-level kvbc folder to be used with concord.
Move several tests from bftengine/tests to tests top-level folder.


kvbc/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required (VERSION 3.2)
+project(libkvbc VERSION 0.1.0.0 LANGUAGES CXX)
+
+add_library(kvbc 	src/ClientImp.cpp
+					src/ReplicaImp.cpp
+  				    src/replica_state_sync_imp.cpp
+   				    src/block_metadata.cpp
+   				    )
```
teoparvanov: nit - the indentation seems broken here

tests/simpleTest/CMakeLists.txt
```diff
@@ -1,46 +1,69 @@
...
-        simple_test_client.hpp
-        )
+    client.cpp
+    ${concord_bft_tools_SOURCE_DIR}/KeyfileIOUtils.cpp
+    simple_test_client.hpp
+)
+        
+add_executable(	simpleTest_client 
+				$<TARGET_OBJECTS:logging_dev>
+    			${simpleTest_client_source_files}
```
teoparvanov: indentation seems broken

kvbc/CMakeLists.txt
toly-kournik: fixed

tests/simpleTest/CMakeLists.txt
toly-kournik: fixed

teoparvanov:
LGTM!

teoparvanov:
Toly, we just had a short sync with Andrew.

Actually, why not move the entire simpleKVBCTests to the top-level test/ folder, together with the Python tests? We could rename simpleKVBCTests to simpleKVBC, as it essentially contains "simple" implementations of a replica and a client.

Additionally, the Python tests rely on the TesterReplica, so it makes sense that simpleKVBCTests (or simpleKVBC) would be under the Python test folder.

Thanks!
Teo

beerriot:
Hi, @toly-kournik. Github has helpfully added links to the PR I just opened against your repo, the source of this pull request: https://github.com/toly-kournik/concord-bft/pull/1. It fixes a leak in the code that is moved in this pull request. I opened it there to make it more obvious what the change is. I will happily rebase it and resubmit it as a PR to this repo after this PR merges, if you prefer.

toly-kournik:
Teo, I think the location is good, simpleKVBCTests can be run without Python framework.
The name can be changed though. I also think that python tests should be moved to tests as well.

toly-kournik:
Bryian, please do it after my PR merges to master.
This is not a "real" leak as it happens only once during a process lifetime, isn't it?
But it's definitely adds noise to valgrind runs.

andrewjstone:
> Teo, I think the location is good, simpleKVBCTests can be run without Python framework.
> The name can be changed though. I also think that python tests should be moved to tests as well.

Makes sense. We can move everything under `/tests.`

teoparvanov:
Hi Toly, agreed. Let's go ahead with merging your current changes.

PR #283, opened 2019-11-26
Parameterized Python test suites

teoparvanov:
Currently the system tests run against statically encoded BFT network
configurations. To run the same test suites against BFT networks with
varying parameters (n, c, f), we need to make significant code changes.

This PR addresses the issue by parameterizing all existing system tests,
which consume the eligible BFT configurations from a central factory
called bft.interesting_configs(). This factory allows to produce and
filter BFT configurations according to key criteria, such as minimum
values for f and for c.

Running the existing system tests against a variety of BFT configurations
now becomes very easy - we can simply add entries in bft.interesting_configs().

Later on, we could even consider feeding those values from command line,
allowing us to run quick CI (n=4) and slower CI (n=4, n=7, n=11, ...).

### DIFF
--- a/tests/test_skvbc.py
+++ b/tests/test_skvbc.py
@@ -9,7 +9,7 @@
 # notices and license terms. Your use of these subcomponents is subject to the
 # terms and conditions of the subcomponent's license, as noted in the LICENSE
 # file.
-
+import random
 import unittest
 import trio
 import os.path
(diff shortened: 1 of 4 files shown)


tests/util/bft.py
```diff
@@ -212,13 +232,28 @@ def stop_replica(self, replica):
...
         if replica not in self.procs:
-            raise AlreadyStoppedError
+            raise AlreadyStoppedError(replica)
         p = self.procs[replica]
         p.kill()
         p.wait()
 
         del self.procs[replica]
 
+    def force_quorum_including_replica(self, replica_id, primary=0):
```
andrewjstone: Can you please add a doc comment here explaining what is going on?

tests/test_skvbc.py
```diff
@@ -85,49 +81,50 @@ def test_get_block_data(self):
...
+        for bft_config in bft.interesting_configs():
+            config = bft.TestConfig(n=bft_config['n'],
+                                    f=bft_config['f'],
+                                    c=bft_config['c'],
+                                    num_clients=bft_config['num_clients'],
+                                    key_file_prefix=KEY_FILE_PREFIX,
+                                    start_replica_cmd=start_replica_cmd)
+            with bft.BftTestNetwork(config) as bft_network:
+                await bft_network.init()
+                [bft_network.start_replica(i) for i in range(config.n)]
```
andrewjstone: Is this the same as `bft_network.start_all_replicas()` ?

tests/util/bft.py
teoparvanov: Fixed.

tests/test_skvbc.py
teoparvanov: Good catch - indeed!

PR #284, opened 2019-11-27
Test BFT engine fast path

teoparvanov:
With this PR we introduce the first explicit test suite for the BFT engine's fast commit path.

We first test the normal behavior (without failures), making sure all messages are processed by the fast path. Then we simulate a failure and check the transition to the slow commit path.

Finally, we add a test (disabled for now) which checks the fast path's resilience to up to c failing nodes (as suggested by the SBFT paper). The test can be re-enabled once this functionality becomes available.

### DIFF
--- a/tests/test_skvbc_fast_path.py
+++ b/tests/test_skvbc_fast_path.py
@@ -1,0 +1,169 @@
+# Concord
+#
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+#
+# This product is licensed to you under the Apache 2.0 license (the "License").
+# You may not use this product except in compliance with the Apache 2.0 License.
+#
+# This product may include a number of subcomponents with separate copyright
+# notices and license terms. Your use of these subcomponents is subject to the
+# terms and conditions of the subcomponent's license, as noted in the LICENSE
+# file.
+
+import os.path
+import random
+
+import trio
+
+import unittest
+
+from util import bft, skvbc
+
+KEY_FILE_PREFIX = "replica_keys_"
+
+
+def start_replica_cmd(builddir, replica_id):
+    """
+    Return a command that starts an skvbc replica when passed to
+    subprocess.Popen.
+
+    Note each arguments is an element in a list.
+    """
+    statusTimerMilli = "500"
+    path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
+    return [path,
+            "-k", KEY_FILE_PREFIX,
+            "-i", str(replica_id),
+            "-s", statusTimerMilli]
+
+
+class SkvbcFastPathTest(unittest.TestCase):
+
+    def setUp(self):
+        self.protocol = skvbc.SimpleKVBCProtocol()
+        # Whenever a replica goes down, all messages initially go via the slow path.
+        # However, when an "evaluation period" elapses (set at 64 sequence numbers),
+        # the system should return to the fast path.
+        self.evaluation_period_seq_num = 64
+
+    def test_fast_path_read_your_write(self):
+        """
+        This test aims to check that the fast commit path is prevalent
+        in the normal, synchronous case (no failed replicas, no network partitioning).
+
+        First we write a series of known K/V entries.
+        Then we check that, in the process, we have stayed on the fast path.
+
+        Finally we check if a known K/V has been executed.
+        """
+        trio.run(self._test_fast_path_read_your_write)
+
+    async def _test_fast_path_read_your_write(self):
+        for bft_config in bft.interesting_configs():
+            config = bft.TestConfig(n=bft_config['n'],
+                                    f=bft_config['f'],
+                                    c=bft_config['c'],
+                                    num_clients=bft_config['num_clients'],
+                                    key_file_prefix=KEY_FILE_PREFIX,
+                                    start_replica_cmd=start_replica_cmd)
+            with bft.BftTestNetwork(config) as bft_network:
+                await bft_network.init()
+                bft_network.start_all_replicas()
+
+                for _ in range(10):
+                    key, val = await bft_network.write_known_kv(self.protocol)
+
+                await bft_network.assert_fast_path_prevalent()
+
+                await bft_network.assert_kv_write_executed(self.protocol, key, val)
+
+    def test_fast_to_slow_path_transition(self):
+        """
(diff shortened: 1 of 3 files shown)


tests/test_skvbc_fast_path.py
```diff
@@ -0,0 +1,158 @@
...
+
+                for _ in range(10):
+                    key, val = await bft_network.write_known_kv(self.protocol)
+
+                await bft_network.assert_slow_path_prevalent(as_of_seq_num=10)
+
+                await bft_network.assert_kv_write_executed(self.protocol, key, val)
+
+    @unittest.skip("According to the SBFT spec, the fast path should tolerate up to c crashed nodes."
+                   "Enable the test once this resilience gets implemented.")
```
teoparvanov: As discussed offline with @salieri11, we have three commit path modes: 
- optimistic fast path (all 3f + 2c + 1 nodes are available)
- threshold fast path (between 2f + c + 1 and 3f + 2c + 1 nodes are available)
- slow path (no more than 2f + c + 1 nodes are available)

Whenever a replica crashes, we automatically degrade to slow path, but after a timeout expires, we return to threshold fast path (see ControllerWithSimpleHistory).

tests/util/bft.py
```diff
@@ -393,6 +393,46 @@ def force_quorum_including_replica(self, replica_id, primary=0):
                     nursery.start_soon(self._assert_state_transfer_not_started,
                                        r)
 
+    async def assert_fast_path_prevalent(self, as_of_seq_num):
+        metric_key = ['replica', 'Counters', 'slowPathCount']
+        nb_slow_paths = await self.metrics.get(0, *metric_key)
+
+        assert nb_slow_paths - as_of_seq_num <= 1, \
```
andrewjstone: I don't follow what you are trying to show here. Is it that there are more fast paths messages than slow path messages at a given seq number? A doc comment would probably help.

tests/util/bft.py
```diff
@@ -393,6 +393,46 @@ def force_quorum_including_replica(self, replica_id, primary=0):
...
+            f'Fast path is not prevalent for n={self.config.n}, f={self.config.f}, c={self.config.c}.'
+
+    async def assert_slow_path_prevalent(self, as_of_seq_num):
+        metric_key = ['replica', 'Gauges', 'lastExecutedSeqNum']
+        all_executed_sequences = await self.metrics.get(0, *metric_key)
+
+        metric_key = ['replica', 'Counters', 'slowPathCount']
+        nb_slow_paths = await self.metrics.get(0, *metric_key)
+
+        assert nb_slow_paths >= all_executed_sequences - as_of_seq_num, \
```
andrewjstone: I think I get what's going on here, but a doc comment for the method would be useful.

tests/util/bft.py
```diff
@@ -393,6 +393,46 @@ def force_quorum_including_replica(self, replica_id, primary=0):
...
+            f'Slow path is not prevalent for n={self.config.n}, f={self.config.f}, c={self.config.c}.'
+
+    async def write_known_kv(self, protocol):
```
andrewjstone: Passing the protocol and using it in the BFT tester breaks abstractions. The reasoning is that the protocol is tied specifically to the skvbc implementation, while the rest of the test network code is supposed to be usable with any state machine. The protocol itself is a concrete type and not an abstract interface here, as it's specifically tied to a state machine that allows transactions and reading of blocks.

This isn't the first time abstraction has been broken in this module (see `assert_successful_put_get` and `prime_for_state_transfer`) I'm fine leaving this here for this PR and fixing this in a future one.

However, the intention was that this module be generic way of running code and performing tests that are bft specific but agnostic to the state machine. We should have another module for specific functions like this that is part of skvbc. For places in this code where we need specific functionality such as "performing a write", we should either move that to a higher level module or create an [abstract type](https://docs.python.org/3/library/abc.html) when we truly want to share the functionality across state machines.

tests/test_skvbc_fast_path.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def setUp(self):
+        self.protocol = skvbc.SimpleKVBCProtocol()
+
+    def test_fast_path_read_your_write(self):
+        """
+        This test aims to check that the fast commit path is prevalent
+        in the normal, synchronous case (no failed replicas, no network partitioning).
+
+        First we write a series of known K/V entries.
```
andrewjstone: typo: s/staid/stayed/

tests/test_skvbc_fast_path.py
```diff
@@ -0,0 +1,158 @@
...
+                                    f=bft_config['f'],
+                                    c=bft_config['c'],
+                                    num_clients=bft_config['num_clients'],
+                                    key_file_prefix=KEY_FILE_PREFIX,
+                                    start_replica_cmd=start_replica_cmd)
+            with bft.BftTestNetwork(config) as bft_network:
+                await bft_network.init()
+                bft_network.start_all_replicas()
+
+                unstable_replicas = list(set(range(0, config.n)) - {0})
```
andrewjstone: This type of list(set(range)) code is used in many places. We should abstract it to a shared function. Something like `unstable_replicas = set_subtract(config.n, {0})`. I don't necessarily like that name, but I've just been thinking about how we could reduce this duplication, which is somewhat hard to read. It can also be done in a separate PR. I'm just pointing it out here while thinking of it.

On the other hand, maybe it's better that it's explicitly done here this way. Any python programmer should grasp it.

tests/util/bft.py
teoparvanov: Agreed, I also thought that bft.py is in charge of more than it should be. I'll track this in a separate tech debt task.

tests/test_skvbc_fast_path.py
teoparvanov: Indeed, it can be added to bft.py - will do it in a separate PR.

tests/util/bft.py
teoparvanov: Agreed - both the implementation and the doc need improvement.

tests/util/bft.py
teoparvanov: Agreed - both the implementation and the doc need improvement.

tests/test_skvbc_fast_path.py
teoparvanov: Thanks, fixed.

PR #289, opened 2019-12-02
Fixing intermittent failure of skvbc-persistence-tests

teoparvanov:
This PR improves the stability of the BFT system tests when persistence is enabled.

With this PR we significantly simplify the logic for determining when state transfer has completed.
Most notably:
- the is_state_transfer_complete() is dropped as following several refactorings, it became redundant with wait_for_state_transfer_to_stop() - both implement essentially the same logic
- we slightly generalize _fetch_or_finish_state_transfer_while_crashing() and reuse it for test_st_while_primary_crashes()

Testing done: ran test_st_while_primary_crashes() 20 consecutive times without failure (previously it would fail every 5-10 runs)

### DIFF
--- a/tests/util/bft.py
+++ b/tests/util/bft.py
@@ -358,13 +358,6 @@ class BftTestNetwork:
                         if n >= expected_seq_num:
                            return
 
-    async def is_state_transfer_complete(self, up_to_date_node, stale_node):
-            # Get the lastExecutedSeqNumber from a reference node
-            key = ['replica', 'Gauges', 'lastExecutedSeqNum']
-            last_exec_seq_num = await self.metrics.get(up_to_date_node, *key)
-            n = await self.metrics.get(stale_node, *key)
-            return n == last_exec_seq_num
-
     async def wait_for_replicas_to_checkpoint(self, replica_ids, checkpoint_num):
         """
         Wait for every replica in `replicas` to take a checkpoint.
--- a/tests/test_skvbc_persistence.py
+++ b/tests/test_skvbc_persistence.py
@@ -260,8 +260,9 @@ class SkvbcPersistenceTest(unittest.TestCase):
     async def _fetch_or_finish_state_transfer_while_crashing(self,
                                                              bft_network,
                                                              up_to_date_node,
-                                                             stale_node):
-       for _ in range(20):
+                                                             stale_node,
+                                                             nb_crashes=20):
+       for _ in range(nb_crashes):
            print(f'Restarting replica {stale_node}')
            bft_network.start_replica(stale_node)
            try:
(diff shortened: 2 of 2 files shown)


tests/util/bft.py
```diff
@@ -359,11 +359,15 @@ def force_quorum_including_replica(self, replica_id, primary=0):
                            return
 
     async def is_state_transfer_complete(self, up_to_date_node, stale_node):
-            # Get the lastExecutedSeqNumber from a reference node
-            key = ['replica', 'Gauges', 'lastExecutedSeqNum']
-            last_exec_seq_num = await self.metrics.get(up_to_date_node, *key)
-            n = await self.metrics.get(stale_node, *key)
-            return n == last_exec_seq_num
+            try:
+                with trio.fail_after(seconds=5):
```
andrewjstone: I don't think the timeout needs to be that large, because the intention is that we are just checking when the system is up. It should be more like 1s or less. Larger timeouts are usually used with loops when waiting for something to happen. We are actually catching this error, presumably because we may loop using the value. A long timeout will make the tests take too long.

teoparvanov: Good point, thanks! 
-> resolved

teoparvanov: Hi Andrew, I tried applying your comment, but it seems to make the tests unstable. Reverting to my first proposal, trading-off test speed vs stability.

Let me know your thoughts.

andrewjstone: I'm very curious as to why this is is the case. If this is true, then we should be using a wait_for function in those scenarios, not a check that is supposed to be immediate. The goal of this function was always to be able to do an immediate check. When it's used the system should be up and responding with metrics. I actually think we should make the change in the C++ code such that we don't start the replica until the metrics server is up and running with valid metrics.

teoparvanov: How about making the BftTestNetwork.start_replica() async and waiting for the metrics to be available before we declare a replica started?

andrewjstone: The replica starting already doesn't guarantee anything, so changing the C++ code doesn't make any sense. We already start the replica server after the UDP server, but the metrics may not be published yet. So I was wrong about that fix.

However, the metrics not being available are why we have wait_for functions in our tests. If we need to wait for certain metrics we should. This is specific to the test though. We don't know what we need to wait for generically, and therefore don't want to wait for any specific metrics for all tests. Especially since they are published by different modules and potentially at different rates. 

My suggestion is that if we are seeing this problem, we should rename the function to wait_for or use the already existing wait_for_state_transfer_to_stop instead.

teoparvanov: Thanks Andrew, I'll give it another go tomorrow.

teoparvanov: Change of strategy today :) I got rid of is_state_transfer_complete() in favor of wait_for_state_transfer_to_stop(). After a chain of refactorings the two methods had become redundant.

Fingers crossed :)

PR #291, opened 2019-12-03
Modify bftengine source tree

yuliasherman:
The current bftengine source tree is flat and contains too many files.
- move files defining messages to separate directory
- move some files to util directory


bftengine/CMakeLists.txt
```diff
@@ -104,6 +104,8 @@ target_include_directories(corebft PUBLIC include/communication)
 target_include_directories(corebft PUBLIC include/bcstatetransfer)
 target_include_directories(corebft PUBLIC include/simplestatetransfer)
 target_include_directories(corebft PUBLIC include/metadatastorage)
+target_include_directories(corebft PUBLIC utils/include)
```
andrewjstone: This is not the correct way to use the utils library in cmake. `target_link_libraries` will automatically include it. We already have that, so this line should not be needed.

yuliasherman: It made me troubles before, but now it's OK.

yuliasherman:
Hi Andrew,
Thank you for the review. The idea behind putting includes files above messages directory was the common usage of them in both - messages and other bftengine cpp files.

andrewjstone:
> Hi Andrew,
> Thank you for the review. The idea behind putting includes files above messages directory was the common usage of them in both - messages and other bftengine cpp files.

Why couldn't the bftengine cpp files look in the messages directory? If we want it to be more clear where it came to a user then we could then `#include "messages/blah.hpp"

andrewjstone:
Whoops accidentally closed. Also didn't want to really leave that comment as is. Let's discuss tomorrow.

PR #296, opened 2019-12-05
Use new Timers class in ReplicaImp

andrewjstone:
This class adds type safety and reduces the amount of work required to
add a new timer.

This change replaces the use of the old Timer and SimpleOperationsScheduler with
the new Timers class.

As the new Timers class uses `std::function` for callbacks, we can capture
`this`directly and no longer need to pass it as a void* parameter and
then cast it to InternalReplicaAPI. Timers also can now be scheduled to
recur and thus remove the need to start them on every callback. And
since these timers are only ever created and manipulated in ReplicaImp,
the above 2 changes allow us to remove all the xxxTimerHandlerFunc
wrapper functions in ReplicaImp, and call the handlers directly via a
closure. It also allows us to remove the callback functions from the
InternalReplicaAPI, since they are unused.

Since these timers are never referred to outside ReplicaImp, we also
remove the accessor functions from InternalReplicaAPI. If we need access
to these specific timers from other modules in the future we can create
a more concrete mechanism to wrap the timer manager and handles.

The Timer callback signature itself was also changed to take a Handle
when a timer fires so it can be reset or cancelled from within the
callback.

A future commit will remove the old TimeUtils code from util and replace
integer usages of time with std::chrono types.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.hpp
+++ b/bftengine/src/bftengine/ReplicaImp.hpp
@@ -40,6 +40,8 @@
 #include "ReplicaLoader.hpp"
 #include "Metrics.hpp"
 
+#include "Timers.hpp"
+
 #include <thread>
 
 namespace bftEngine {
@@ -98,9 +100,6 @@ class ReplicaImp : public InternalReplicaApi, public IReplicaForStateTransfer {
 
   // retransmissions manager (can be disabled)
   RetransmissionsManager* retransmissionsManager = nullptr;
-
-  // scheduler
-  SimpleOperationsScheduler timersScheduler;
 
   // controller
   ControllerBase* controller;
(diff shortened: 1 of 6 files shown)


util/include/Timers.hpp
```diff
@@ -40,11 +48,14 @@ class Timers {
     };
 
    private:
-    Timer(std::chrono::milliseconds d, Type t, std::function<void()> cb) {
+    Timer(std::chrono::milliseconds d, Type t, std::function<void(Handle)> cb) {
       Timer(d, t, cb, std::chrono::steady_clock::now());
     }
```
yuliasherman: Don't we need to handle stop/remove timers differently now?

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -457,30 +459,14 @@ class ReplicaImp : public InternalReplicaApi, public IReplicaForStateTransfer {
...
-  virtual void onDebugStatTimer(Time cTime, Timer& timer) override;
-  virtual void onMetricsTimer(Time cTime, Timer& timer) override;
+  virtual void onViewsChangeTimer(concordUtil::Timers::Handle);
+  virtual void onStateTranTimer(concordUtil::Timers::Handle);
+  virtual void onRetransmissionsTimer(concordUtil::Timers::Handle);
+  virtual void onStatusReportTimer(concordUtil::Timers::Handle);
+  virtual void onSlowPathTimer(concordUtil::Timers::Handle);
+  virtual void onInfoRequestTimer(concordUtil::Timers::Handle);
+  virtual void onDebugStatTimer(concordUtil::Timers::Handle);
+  virtual void onMetricsTimer(concordUtil::Timers::Handle);
```
dartdart26: Based on `ReplicaImp`'s interface and implementation, I think these can be made non-virtual. For example, an inheriting class is not able to change message handling and I would assume that is also the case for timer handling.

util/include/Timers.hpp
andrewjstone: We never stop or remove timers in the current code AFAIK. Recurring timers can be stopped and removed via `timers_.cancel(handle)`

bftengine/src/bftengine/ReplicaImp.hpp
andrewjstone: Good catch.

util/include/Timers.hpp
```diff
@@ -40,11 +48,14 @@ class Timers {
     };
 
    private:
-    Timer(std::chrono::milliseconds d, Type t, std::function<void()> cb) {
+    Timer(std::chrono::milliseconds d, Type t, std::function<void(Handle)> cb) {
```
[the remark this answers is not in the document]
andrewjstone: I modified the two parameters that weren't clear from context (type or initializers).

PR #301, opened 2019-12-05
Rename SimpleClient to SimpleClientImp - the class it really introduces and implements

yuliasherman:
+ Fix missing ";" after assert
+ Unify class data member names

### DIFF
--- a/bftengine/src/bftengine/SimpleClientImp.cpp
+++ b/bftengine/src/bftengine/SimpleClientImp.cpp
@@ -32,49 +32,45 @@ class SimpleClientImp : public SimpleClient, public IReceiver {
  public:
   SimpleClientImp(
       ICommunication* communication, uint16_t clientId, uint16_t fVal, uint16_t cVal, SimpleClientParams& p);
+  ~SimpleClientImp() override;
 
-  // SimpleClient methods
+  int sendRequest(bool isReadOnly,
+                  const char* request,
+                  uint32_t lengthOfRequest,
+                  uint64_t reqSeqNum,
+                  uint64_t timeoutMilli,
+                  uint32_t lengthOfReplyBuffer,
+                  char* replyBuffer,
+                  uint32_t& actualReplyLength) override;
 
-  virtual ~SimpleClientImp() override;
-
-  virtual int sendRequest(bool isReadOnly,
-                          const char* request,
-                          uint32_t lengthOfRequest,
-                          uint64_t reqSeqNum,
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/SimpleClientImp.cpp
```diff
@@ -426,39 +404,37 @@ class SeqNumberGeneratorForClientRequestsImp : public SeqNumberGeneratorForClien
...
+  Assert(_lastCountOfUniqueFetchID <= 0x3FFFFF);
+  r = r | ((uint64_t)_lastCountOfUniqueFetchID);
 
   return r;
 }
 
 }  // namespace impl
-}  // namespace bftEngine
 
-namespace bftEngine {
```
andrewjstone: I don't believe moving this into the impl namespace is correct. The caller should never use the impl namespace directly. l think that's why they were separated.

bftengine/src/bftengine/SimpleClientImp.cpp
```diff
@@ -32,49 +32,45 @@ class SimpleClientImp : public SimpleClient, public IReceiver {
...
  protected:
-  static const uint32_t maxLegalMsgSize = 64 * 1024;  // TODO(GG): ???
-  static const uint16_t timersResolutionMilli = 50;
+  void sendPendingRequest();
+  void onMessageFromReplica(MessageBase* msg);
+  void onRetransmission();
+  void reset();
+
+ protected:
+  static const uint32_t _maxLegalMsgSize = 64 * 1024;  // TODO(GG): ???
```
andrewjstone: The style guide mandates trailing underscores for member variables rather than leading. I'm fine leaving this for now as it unifies the style in this file. Just mentioning in case you want to clean it up in another pass.

bftengine/src/bftengine/SimpleClientImp.cpp
yuliasherman: Oh, no I did not move it to impl namespace, just removed duplication. Take a look at a full file.

bftengine/src/bftengine/SimpleClientImp.cpp
yuliasherman: Yeah, I had doubts about it.

bftengine/src/bftengine/SimpleClientImp.cpp
andrewjstone: Oh you are right! My bad.

PR #305, opened 2019-12-07
Replace int based time with std::chrono types

andrewjstone:
This removes the need for switching from milliseconds to microseconds
and vice versa manually and adds a level of type safety.

In all cases the use of absDifference appeared unnecessary, since we are
using monotonic times and we never read them from disk, so they can't go
backwards.

Future changes may want to change constants and member variables for
configured values to use chrono types as well rather than `xxxMilli`.


bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
```diff
@@ -87,23 +90,20 @@ bool ControllerWithSimpleHistory::onNewSeqNumberExecution(SeqNum n) {
...
-  if (now > s.prePrepareTime) duration = subtract(now, s.prePrepareTime);
-
-  const uint64_t MAX_DURATION_MICRO = 2 * 1000 * 1000;  // 2 sec
-  const uint64_t MIN_DURATION_MICRO = 100;              // 100 micro
+  const auto MAX_DURATION_MICRO = microseconds(2 * 1000 * 1000);
+  const auto MIN_DURATION_MICRO = microseconds(100);
 
   if (duration > MAX_DURATION_MICRO)
-    duration = MAX_DURATION_MICRO;
+    s.durationMicro_ = MAX_DURATION_MICRO;
```
teoparvanov: I think the idea here is to cap "duration" up to MAX_DURATION_MICRO.
Why are we setting s.durationMicro_ instead?

bftengine/src/bftengine/TimeUtils.hpp
```diff
@@ -0,0 +1,40 @@
+// Concord
+//
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
```
teoparvanov: typo: 2019

bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
andrewjstone: duration is a local that we only use past this point to set s.durationMicro_. see line 102. I just removed the setting of the temporary.

bftengine/src/bftengine/TimeUtils.hpp
andrewjstone: Good catch. I changed it to 2018-2019, since this was based on an older file.

teoparvanov:
Awesome, thanks

teoparvanov:
LGTM!

PR #306, opened 2019-12-08
Disable copy operations on RSASigner and RSAVerifier

dartdart26:
Current implementations of copy operations on RSASigner and RSAVerifier
will cause core dumps because of double free of the internal
implementation pointer. This PR does the following:
*  Disable copy operations on RSASigner and RSAVerifier
*  Allow move operations on RSASigner and RSAVerifier
*  Implement the pimpl idiom via a std::unique_ptr to avoid casts and
manual memory management

Copy operations might be added in the future if a need arises. That will
need further investigation on the safety of CryptoPP copy operations.

### DIFF
--- a/bftengine/src/bftengine/Crypto.cpp
+++ b/bftengine/src/bftengine/Crypto.cpp
@@ -135,9 +135,9 @@ DigestUtil::Context::~Context() {
   }
 }
 
-class RSASignerInternal {
+class RSASigner::Impl {
  public:
-  RSASignerInternal(BufferedTransformation& privateKey) : rand(sGlobalRandGen), priv(privateKey) {}
+  Impl(BufferedTransformation& privateKey) : rand(sGlobalRandGen), priv(privateKey) {}
 
   size_t signatureLength() const { return priv.SignatureLength(); }
 
@@ -160,9 +160,9 @@ class RSASignerInternal {
   RSASS<PKCS1v15, SHA256>::Signer priv;
 };
 
-class RSAVerifierInternal {
+class RSAVerifier::Impl {
  public:
-  RSAVerifierInternal(BufferedTransformation& publicKey) : pub(publicKey) {}
+  Impl(BufferedTransformation& publicKey) : pub(publicKey) {}
 
   size_t signatureLength() const { return pub.SignatureLength(); }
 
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/Crypto.cpp
```diff
@@ -177,50 +177,43 @@ class RSAVerifierInternal {
 
 RSASigner::RSASigner(const char* privateKey) {
   StringSource s(privateKey, true, new HexDecoder);
-  d = new RSASignerInternal(s);
+  impl.reset(new Impl(s));
```
teoparvanov: To complete the abstraction from raw pointers, can we use make_unique() here?

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -177,50 +177,43 @@ class RSAVerifierInternal {
...
-  RSASignerInternal* p = (RSASignerInternal*)d;
-  bool succ = p->sign(inBuffer, lengthOfInBuffer, outBuffer, lengthOfOutBuffer, lengthOfReturnedData);
-  return succ;
+  return impl->sign(inBuffer, lengthOfInBuffer, outBuffer, lengthOfOutBuffer, lengthOfReturnedData);
 }
 
 RSAVerifier::RSAVerifier(const char* publicKey) {
   StringSource s(publicKey, true, new HexDecoder);
-  d = new RSAVerifierInternal(s);
+  impl.reset(new Impl(s));
```
teoparvanov: same question here

bftengine/src/bftengine/Crypto.cpp
dartdart26: Good point, done.

bftengine/src/bftengine/Crypto.cpp
dartdart26: Good point, done.

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -177,50 +177,43 @@ class RSAVerifierInternal {
...
   StringSource s(privateKey, true, new HexDecoder);
-  d = new RSASignerInternal(s);
+  impl = std::make_unique<Impl>(s);
 }
 
-RSASigner::~RSASigner() {
-  RSASignerInternal* p = (RSASignerInternal*)d;
-  delete p;
-}
+RSASigner::RSASigner(RSASigner&&) = default;
```
salieri11: Do we want to it to be movable?

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -177,50 +177,43 @@ class RSAVerifierInternal {
...
   StringSource s(publicKey, true, new HexDecoder);
-  d = new RSAVerifierInternal(s);
+  impl = std::make_unique<Impl>(s);
 }
 
-RSAVerifier::~RSAVerifier() {
-  RSAVerifierInternal* p = (RSAVerifierInternal*)d;
-  delete p;
-}
+RSAVerifier::RSAVerifier(RSAVerifier&&) = default;
```
salieri11: Same question as above

bftengine/src/bftengine/Crypto.cpp
dartdart26: I would say yes. For example, we can use it in standard vectors and return it from functions (without expecting RVO).

bftengine/src/bftengine/Crypto.cpp
dartdart26: Please see above.

PR #307, opened 2019-12-09
Replica configuration code dedup and refactoring.

toly-kournik:
https://jira.eng.vmware.com/browse/VB-2168

Deduplicate configuration fields which exist in ReplicaConfig from ReplicaImp.
Refactor ReplicaConfigSingleton for optimization and proper initialization.

### DIFF
--- a/tests/simpleTest/CMakeLists.txt
+++ b/tests/simpleTest/CMakeLists.txt
@@ -33,7 +33,7 @@ add_executable(simpleTest_server
 
 target_include_directories(simpleTest_server
                            PRIVATE
-                           ${CMAKE_SOURCE_DIR}/tests/config
+                           ../config
                            ${bftengine_SOURCE_DIR}/tests/simpleStorage
                            ${concord_bft_tools_SOURCE_DIR}
                            ${bftengine_SOURCE_DIR}/src/bftengine
@@ -64,7 +64,7 @@ target_include_directories(persistency_test PUBLIC
         ${bftengine_SOURCE_DIR}/tests/simpleStorage
         ${bftengine_SOURCE_DIR}/include/communication
         ${concord_bft_tools_SOURCE_DIR}
-        ${CMAKE_SOURCE_DIR}/tests/config
+        ../config
         ${SIMPLE_STORAGE_DIR})
 
 target_link_libraries(persistency_test PUBLIC


bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -62,34 +60,31 @@ class StateTransferMsg;
...
 
   // pointers to message handlers
   const std::unordered_map<uint16_t, PtrToMetaMsgHandler> metaMsgHandlers;
 
   // communication
-  class MsgReceiver;  // forward declaration
-  IncomingMsgsStorage incomingMsgsStorage;
-  MsgReceiver* msgReceiver;
+  class MsgReceiver;                                // forward declaration
+  IncomingMsgsStorage incomingMsgsStorage = 20000;  // TODO(GG): use configuration
```
andrewjstone: This is a bit strange. You are assigning an integer to a class instance. I assume it's using that in the constructor somehow, but it's not apparent from that line. Maybe change to `incomingMsgsStorage(MAX_INCOMING_MSGS)`?

bftengine/tests/testSerialization/TestSerialization.cpp
```diff
@@ -148,7 +148,7 @@ const uint16_t msgsNum = 2 * fVal + 2 * cVal + 1;
 
 typedef pair<uint16_t, string> IdToKeyPair;
 
-ReplicaConfig config;
+ReplicaConfig *config;
```
andrewjstone: Why dynamically allocate? I also don't see a corresponding delete. Better to use some RAII, so we don't have leaks, no?

bftengine/tests/testSerialization/TestSerialization.cpp
```diff
@@ -425,45 +425,49 @@ void testSetSimpleParams(bool toSet) {
...
+  config->thresholdSignerForExecution = nullptr;
+  config->thresholdVerifierForExecution = new IThresholdVerifierDummy;
+  config->thresholdSignerForSlowPathCommit = new IThresholdSignerDummy;
+  config->thresholdVerifierForSlowPathCommit = new IThresholdVerifierDummy;
+  config->thresholdSignerForCommit = new IThresholdSignerDummy;
+  config->thresholdVerifierForCommit = new IThresholdVerifierDummy;
+  config->thresholdSignerForOptimisticCommit = new IThresholdSignerDummy;
+  config->thresholdVerifierForOptimisticCommit = new IThresholdVerifierDummy;
+  config->singletonFromThis();
+  config->singletonFromThis();
```
andrewjstone: Why do this twice?

toly-kournik: just to check that it works as expected (only once)

bftengine/src/bftengine/ReplicaImp.hpp
toly-kournik: yes, but this syntax doesn't work

bftengine/tests/testSerialization/TestSerialization.cpp
toly-kournik: because ReplicaConfigSingleton holds a pointer to ReplicaConfig

toly-kournik: I have changed an approach a little. Now ReplicaConfigSingleton doesn't hold a pointer to the original ReplicaConfig, but creates a copy, decoupling the life cycles of two objects

PR #310, opened 2019-12-10
Test BFT engine view change

teoparvanov:
With this PR we introduce several basic view change test scenarios:
- single view change when only the primary crashes
- single view change when f replicas (including the primary) crash
- multiple view changes on the slow path

Note: view change on the fast path cannot be tested easily by bringing down the primary.
This is because when the primary goes down, we always downgrade to the slow path for 64 sequences
and the fast path is restored only afterwards.

PS: automatic view change will be tested in a separate PR and it will allow to test
fast path & view change combined

### DIFF
--- a/tests/util/bft.py
+++ b/tests/util/bft.py
@@ -44,9 +44,9 @@ def interesting_configs(
         selected=lambda *config: True):
 
     bft_configs = [{'n': 4, 'f': 1, 'c': 0, 'num_clients': 4},
-                   {'n': 7, 'f': 2, 'c': 0, 'num_clients': 4},
                    {'n': 6, 'f': 1, 'c': 1, 'num_clients': 4},
-                   # {'n': 9, 'f': 2, 'c': 1, 'num_clients': 4},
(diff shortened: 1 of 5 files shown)


tests/util/bft.py
```diff
@@ -44,9 +44,9 @@ def interesting_configs(
         selected=lambda *config: True):
 
     bft_configs = [{'n': 4, 'f': 1, 'c': 0, 'num_clients': 4},
-                   {'n': 7, 'f': 2, 'c': 0, 'num_clients': 4},
                    {'n': 6, 'f': 1, 'c': 1, 'num_clients': 4},
-                   # {'n': 9, 'f': 2, 'c': 1, 'num_clients': 4},
+                   {'n': 7, 'f': 2, 'c': 0, 'num_clients': 4},
+                   # {'n': 9, 'f': 2, 'c': 1, 'num_clients': 4}
```
teoparvanov: Note: I haven't changed anything significant here - just sorted the entries by ascending _n_

PR #316, opened 2019-12-12
Add types useful for building a sparse merkle tree

andrewjstone:
Implementation of the tree will come piecemeal in future commits.

Tests for the base types are included.

### DIFF
--- a/storage/include/sparse_merkle/base_types.h
+++ b/storage/include/sparse_merkle/base_types.h
@@ -1,0 +1,126 @@
+// Concord
+//
+// Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the sub-component's license, as noted in the
+// LICENSE file.
+
+#include <cstdint>
+#include <vector>
+
+#include "assertUtils.hpp"
+
+namespace concord {
+namespace storage {
+namespace sparse_merkle {
+
+// A nibble is 4 bits, stored in the lower bits of a byte.
+class Nibble {
+ public:
+  Nibble(uint8_t byte) {
+    Assert((byte & 0xF0) == 0);
+    data_ = byte;
+  }
+
+  bool get_bit(const size_t bit) const {
+    Assert(bit < 4);
+    return (data_ >> bit) & 1;
+  }
+
+  bool operator==(const Nibble& other) const { return data_ == other.data_; }
+
+  // Promote the underlying uint8_t to a size_t so that it can be used as an
+  // index into arrays and vectors.
+  size_t to_index() const { return data_; }
+
+  // Return the underlying representation
+  uint8_t data() const { return data_; }
+
+ private:
+  // Only the lower 4 bits are used
+  uint8_t data_;
+};
+
+template <typename T>
+Nibble get_nibble(const size_t n, const T& buf) {
+  size_t index = n / 2;
+  uint8_t byte = buf[index];
+  if (n % 2 == 0) {
+    // Even value: Get the first nibble of the byte
+    byte = byte >> 4;
+  } else {
+    // Odd value: Get the second nibble of the byte
+    byte = byte & 0x0F;
+  }
+  return Nibble(byte);
+}
+
+// A Hash is a wrapper around a byte buffer containing a hash
+class Hash {
+ public:
+  // Change the following 2 constants if using a different hash algorithm
+  static constexpr size_t SIZE_IN_BITS = 256;
+  static constexpr const char* const HASH_ALGORITHM = "SHA3-256";
+
+  static constexpr size_t SIZE_IN_BYTES = SIZE_IN_BITS / 8;
+  static constexpr size_t MAX_NIBBLES = SIZE_IN_BITS / 16;
+
+  // This is the hash of the empty string ''. It's used as a placeholder value
+  // for merkle tree hash calculations.
+  //
+  // Change this if the hash algorithm changes
+  static constexpr std::array<uint8_t, SIZE_IN_BYTES> EMPTY = {
+      0xa7, 0xff, 0xc6, 0xf8, 0xbf, 0x1e, 0xd7, 0x66, 0x51, 0xc1, 0x47, 0x56, 0xa0, 0x61, 0xd6, 0x62,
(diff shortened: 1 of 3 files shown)


storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,129 @@
...
+// This is a path used to find internal nodes in a merkle tree.
+//
+// To find the location of where to put a leaf in a tree, the hash of the key is
+// walked nibble by nibble until an empty node is found.
+class NibblePath {
+ public:
+  NibblePath() : num_nibbles_(0) {}
+
+  // Return the length of the path_ in nibbles
+  size_t length() { return num_nibbles_; }
```
dartdart26: Consider making it const?

storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,129 @@
...
+  static const size_t SIZE_IN_BYTES = SIZE_IN_BITS / 8;
+  static const size_t MAX_NIBBLES = SIZE_IN_BITS / 16;
+
+  // This is the hash of the empty string ''. It's used as a placeholder value
+  // for merkle tree hash calculations.
+  //
+  // Change this if the hash algorithm changes
+  static constexpr std::array<uint8_t, SIZE_IN_BYTES> EMPTY = {
+      0xa7, 0xff, 0xc6, 0xf8, 0xbf, 0x1e, 0xd7, 0x66, 0x51, 0xc1, 0x47, 0x56, 0xa0, 0x61, 0xd6, 0x62,
+      0xf5, 0x80, 0xff, 0x4d, 0xe4, 0x3b, 0x49, 0xfa, 0x82, 0xd8, 0x0a, 0x4b, 0x80, 0xf8, 0x43, 0x4a};
```
teoparvanov: does it make sense to do a static_assert() here, ensuring this is exactly _h('')_?

storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,129 @@
...
+  // Only the lower 4 bits are used
+  uint8_t data_;
+};
+
+// A Hash is a wrapper around a byte buffer containing a hash
+class Hash {
+ public:
+  // Change the following 2 constants if using a different hash algorithm
+  static const size_t SIZE_IN_BITS = 256;
+  const char* HASH_ALGORITHM = "SHA3-256";
```
dartdart26: Can we either make it `static const char* const` or `static const std::string` so that the pointer is not present in every object?

storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,129 @@
...
+ private:
+  // Only the lower 4 bits are used
+  uint8_t data_;
+};
+
+// A Hash is a wrapper around a byte buffer containing a hash
+class Hash {
+ public:
+  // Change the following 2 constants if using a different hash algorithm
+  static const size_t SIZE_IN_BITS = 256;
```
dartdart26: Can we make it `constexpr` to propagate in calculations and usage in `std::array`?

storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,129 @@
...
+  const char* HASH_ALGORITHM = "SHA3-256";
+
+  static const size_t SIZE_IN_BYTES = SIZE_IN_BITS / 8;
```
dartdart26: Can we make it `constexpr`(same rationale as for `SIZE_IN_BITS`)?

storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,129 @@
...
+    if (num_nibbles_ % 2 == 0) {
+      path_.push_back(nibble.data() << 4);
+    } else {
+      path_.back() |= nibble.data();
+    }
+    num_nibbles_ += 1;
+  }
+
+  // Get the nth nibble.
+  Nibble get(size_t n) const {
```
dartdart26: Does it make sense to write a common function that gets nibbles from an array and use it here and in `Hash::get_nibble()` as the implementations are almost identical?

storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,129 @@
...
+class NibblePath {
+ public:
+  NibblePath() : num_nibbles_(0) {}
+
+  // Return the length of the path_ in nibbles
+  size_t length() { return num_nibbles_; }
+
+  // Append a nibble to the path_
+  void append(Nibble nibble) {
+    Assert(num_nibbles_ < Hash::MAX_NIBBLES - 1);
```
dartdart26: Could you please add information about the rationale behind this assert?

storage/include/sparse_merkle/base_types.h
andrewjstone: Part of the problem is I don't have the hash function available to do this yet :) I double checked it manually 3 times though, with 2 different implementations of sha3-256.

storage/include/sparse_merkle/base_types.h
andrewjstone: done.

storage/include/sparse_merkle/base_types.h
andrewjstone: done.

storage/include/sparse_merkle/base_types.h
andrewjstone: done.

storage/include/sparse_merkle/base_types.h
andrewjstone: done.

storage/include/sparse_merkle/base_types.h
andrewjstone: done.

storage/include/sparse_merkle/base_types.h
andrewjstone: done.

andrewjstone:
Rebased off master and squashed to force a new build, since the other one was hung.

storage/include/sparse_merkle/base_types.h
```diff
@@ -0,0 +1,126 @@
...
+  // Return the underlying representation
+  uint8_t data() const { return data_; }
+
+ private:
+  // Only the lower 4 bits are used
+  uint8_t data_;
+};
+
+template <typename T>
+Nibble get_nibble(const size_t n, T buf) {
```
dartdart26: Can we maybe make it take the buffer by `const T&` to avoid copying?

andrewjstone: Great catch!

andrewjstone: fixed!

PR #317, opened 2019-12-12
Test BFT engine automatic view change

teoparvanov:
Test BFT engine automatic view change

This PR introduces system tests for automatic primary rotation ("auto view change").
However, so far this functionality was disabled in a non-configurable manner, hence not available for automated testing.
So this PR has two parts:

**[C++]** Expose configuration parameters for enabling view change. The configuration parameters were simplified
- "basic" view change - configurable via viewChangeProtocolEnabled & viewChangeTimerMillisec (-v in the TesterReplica)
-  auto view change - configurable via autoPrimaryRotationEnabled & autoPrimaryRotationTimerMillisec (-a in the TesterReplica)

Note that autoPrimaryRotationEnabled implies viewChangeProtocolEnabled (auto view change being an "add-on" option to basic view change).

Note: I have removed the "constexpr" forceViewChangeProtocolEnabled & forceViewChangeProtocolDisabled, so that the view change protocol is configurable only via properties.

**[Python]** Add a new system test suite for auto view change:
- verify auto view change occurs "at rest" (no crashing replicas, no incoming requests)
- verify auto view change while the primary is down (no incoming requests)
- verify auto view change while processing requests on the fast commit path

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -45,11 +45,17 @@ struct ReplicaConfig {
   // 1 <= concurrencyLevel <= 30
   uint16_t concurrencyLevel = 0;
 
-  // autoViewChangeEnabled=true , if the automatic view change protocol is enabled
-  bool autoViewChangeEnabled = false;
+  // viewChangeProtocolEnabled=true , if the view change protocol is enabled at all
+  bool viewChangeProtocolEnabled = false;
 
   // a time interval in milliseconds. represents the timeout used by the  view change protocol (TODO: add more details)
   uint16_t viewChangeTimerMillisec = 0;
+
+  // autoPrimaryRotationEnabled=true , if the automatic primary rotation is enabled
+  bool autoPrimaryRotationEnabled = false;
+
+  // a time interval in milliseconds, represents the timeout for automatically replacing the primary
+  uint16_t autoPrimaryRotationTimerMillisec = 0;
 
   // public keys of all replicas. map from replica identifier to a public key
   std::set<std::pair<uint16_t, const std::string>> publicKeysOfReplicas;
@@ -113,8 +119,11 @@ class ReplicaConfigSingleton {
   uint16_t GetNumOfClientProxies() const { return config_->numOfClientProxies; }
   uint16_t GetStatusReportTimerMillisec() const { return config_->statusReportTimerMillisec; }
   uint16_t GetConcurrencyLevel() const { return config_->concurrencyLevel; }
-  bool GetAutoViewChangeEnabled() const { return config_->autoViewChangeEnabled; }
+  bool GetViewChangeProtocolEnabled() const { return config_->viewChangeProtocolEnabled; }
   uint16_t GetViewChangeTimerMillisec() const { return config_->viewChangeTimerMillisec; }
+  bool GetAutoPrimaryRotationEnabled() const { return config_->autoPrimaryRotationEnabled; }
+
+  uint16_t GetAutoPrimaryRotationTimerMillisec() const { return config_->autoPrimaryRotationTimerMillisec; }
   std::set<std::pair<uint16_t, const std::string>> GetPublicKeysOfReplicas() const {
     return config_->publicKeysOfReplicas;
   }
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2429,9 +2429,9 @@ void ReplicaImp::onViewsChangeTimer(Timers::Handle timer)  // TODO(GG): review/u
   //
   //////////////////////////////////////////////////////////////////////////////
 
-  if (autoPrimaryUpdateEnabled && currentViewIsActive()) {
-    const uint64_t timeout =
-        (isCurrentPrimary() ? (autoPrimaryUpdateMilli) : (autoPrimaryUpdateMilli + viewChangeTimeoutMilli));
+  if (autoPrimaryRotationEnabled && currentViewIsActive()) {
+    const uint64_t timeout = (isCurrentPrimary() ? (autoPrimaryRotationTimerMilli)
+                                                 : (autoPrimaryRotationTimerMilli + viewChangeTimeoutMilli));
 
     const uint64_t diffMilli = duration_cast<milliseconds>(currTime - timeOfLastViewEntrance).count();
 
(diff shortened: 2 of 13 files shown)


bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -44,12 +44,18 @@ struct ReplicaConfig {
...
-  // autoViewChangeEnabled=true , if the automatic view change protocol is enabled
-  bool autoViewChangeEnabled = false;
+  // viewChangeProtocolEnabled=true , if the view change protocol is enabled at all
+  bool viewChangeProtocolEnabled = false;
 
   // a time interval in milliseconds. represents the timeout used by the  view change protocol (TODO: add more details)
   uint16_t viewChangeTimerMillisec = 0;
 
+  // autoViewChangeEnabled=true , if the automatic view change is enabled
+  bool autoViewChangeEnabled = false;
```
andrewjstone: The name of this flag and feature doesn't really indicate what its doing. We can keep the name the same and add a comment, but it may be better to name it something else that shows that it is periodically rotating the leader. Maybe something like `PeriodicLeaderRotationEnabled` or  `AutoLeaderRotationEnabled`.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2419,9 +2419,9 @@ void ReplicaImp::onViewsChangeTimer(Timers::Handle timer)  // TODO(GG): review/u
   //
   //////////////////////////////////////////////////////////////////////////////
 
-  if (autoPrimaryUpdateEnabled && currentViewIsActive()) {
+  if (autoViewChangeEnabled && currentViewIsActive()) {
     const uint64_t timeout =
-        (isCurrentPrimary() ? (autoPrimaryUpdateMilli) : (autoPrimaryUpdateMilli + viewChangeTimeoutMilli));
+        (isCurrentPrimary() ? (autoViewChangeTimerMilli) : (autoViewChangeTimerMilli + viewChangeTimeoutMilli));
```
andrewjstone: Maybe `PrimaryRotationTimeoutMilli` ?

tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -75,11 +75,21 @@ std::unique_ptr<TestSetup> TestSetup::ParseArgs(int argc, char** argv) {
...
+          rp.viewChangeTimeout = (uint16_t)tempId;
           rp.viewChangeEnabled = true;
         }
       } break;
+      case 'a': {
+        strncpy(argTempBuffer, optarg, sizeof(argTempBuffer) - 1);
+        argTempBuffer[sizeof(argTempBuffer) - 1] = 0;
+        idStr = argTempBuffer;
+        int tempId = std::stoi(idStr);
+        if (tempId >= 0 && (uint16_t)tempId < UINT16_MAX) {
```
andrewjstone: I know this is the same logic as above, but now that I think about it, we should assert on this rather than silently ignoring the parameter. In the future we should also allow much larger ints. We should just use size_t for everything parameter related instead of trying to optimize and have to do all this custom logic. If we want to restrict values we can do that specifically rather than relying on the size of the int.

Note that this doesn't have to be changed here, but is something we should likely track.

tests/test_skvbc_auto_view_change.py
```diff
@@ -0,0 +1,153 @@
...
+                                    num_clients=bft_config['num_clients'],
+                                    key_file_prefix=KEY_FILE_PREFIX,
+                                    start_replica_cmd=start_replica_cmd)
+            with bft.BftTestNetwork(config) as bft_network:
+                await bft_network.init()
+                bft_network.start_all_replicas()
+
+                initial_primary = 0
+
+                # do nothing - just wait for an automatic view change
```
andrewjstone: Regardless of a 10 second wait for a 5 second timeout, we shouldn't be relying on view change to have taken place based on a sleep. We should instead do a wait_for after 5 seconds, retrying every 100ms or something until it passes, and maybe timeout at 10 or 20 seconds. This not only speeds up the test, but if we discover really slow machines timing out during the test, we can bump the wait without making all test runs wait that long.

It looks like `bft_network.get_view_number` already does a wait_for, but it doesn't fail after 10 seconds but moves on after. This is presumably why you needed the assert afterwords also. We should instead fail after 10 seconds. I'd rewrite that function or add a new one and explicitly name it `wait_for_view_number`. I'd also call it after a 5 second sleep to shorten test runs.

tests/test_skvbc_auto_view_change.py
```diff
@@ -0,0 +1,153 @@
...
+                                    start_replica_cmd=start_replica_cmd)
+            with bft.BftTestNetwork(config) as bft_network:
+                await bft_network.init()
+                bft_network.start_all_replicas()
+
+                initial_primary = 0
+                bft_network.stop_replica(initial_primary)
+
+                # do nothing - just wait for an automatic view change
+                await trio.sleep(seconds=10)
```
andrewjstone: same as above. Should probably be a 5 second sleep.

tests/test_skvbc_auto_view_change.py
```diff
@@ -0,0 +1,153 @@
...
+                await bft_network.init()
+                bft_network.start_all_replicas()
+                skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+                initial_primary = 0
+
+                for _ in range(150):
+                    key, val = await skvbc.write_known_kv()
+
+                view = await bft_network.get_view_number(
```
andrewjstone: Same as above.

tests/test_skvbc_auto_view_change.py
teoparvanov: Hi Andrew, thanks for the detailed review! I have decided to address this comment (along with other related improvements) in a separate PR https://github.com/vmware/concord-bft/pull/320

This way, I can unblock Toly's PR https://github.com/vmware/concord-bft/pull/318, and at the same time keep the focus of the "automatic view change test" PR.

tests/test_skvbc_auto_view_change.py
teoparvanov: see PR https://github.com/vmware/concord-bft/pull/320

tests/test_skvbc_auto_view_change.py
teoparvanov: see PR https://github.com/vmware/concord-bft/pull/320

tests/simpleKVBC/TesterReplica/setup.cpp
teoparvanov: I agree, but as you suggest, I prefer to tackle this separately.

tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -106,8 +116,10 @@ std::unique_ptr<TestSetup> TestSetup::ParseArgs(int argc, char** argv) {
   // This allows more concurrency and only affects known ids in the
   // communication classes.
   replicaConfig.numOfClientProxies = 100;
-  replicaConfig.autoViewChangeEnabled = rp.viewChangeEnabled;
+  replicaConfig.viewChangeProtocolEnabled = rp.viewChangeEnabled;
   replicaConfig.viewChangeTimerMillisec = rp.viewChangeTimeout;
+  replicaConfig.autoViewChangeEnabled = rp.autoViewChangeEnabled;
+  replicaConfig.autoViewChangeTimerMillisec = rp.autoViewChangeTimeout;
```
teoparvanov: We should also make sure those changes are replicated into the actual replica's configuration:
https://gitlab.eng.vmware.com/blockchain/vmwathena_blockchain/blob/master/concord/src/consensus/replica_imp.cpp#L214

https://gitlab.eng.vmware.com/blockchain/vmwathena_blockchain/blob/master/concord/src/consensus/replica_imp.cpp#L222

PR #318, opened 2019-12-12
Replica config refactoring

toly-kournik:
replacement for the original pull request


teoparvanov:
Hi Toly, I see the view change test is failing again. I suggest we sit down and debug this together, if it can wait till Monday...

CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ string(APPEND CXX_FLAGS " -Werror")
 #string(APPEND CXX_FLAGS
 #    " -fno-strict-overflow")
 # Prevents some buffer overflows: https://access.redhat.com/blogs/766093/posts/1976213
-string(APPEND CXX_FLAGS_DEBUG
+string(APPEND CXX_FLAGS_RELEASE
```
andrewjstone: This is incorrect. This will change the optimization level of all release builds, which we do not want. We want to continue building releases with the default of `-03`.

What is the actual problem forcing this change? (This rationale should be part of the commit message as shown in our [contributing guide](https://github.com/vmware/concord-bft/blob/master/CONTRIBUTING.md#formatting-commit-messages)). Is it causing builds to fail? Is it hindering debugging? Knowing what the issue is would help us come up with a proper fix.

If it's because you don't want `-O` (which appears to be an alias for -O2 in clang), because it is hindering debugging, we can do 1 of 2 things: use the `-Og` flag and keep this setting for DEBUG mode or just disable FORTIFY_SOURCE=2 in DEBUG mode and only use it in release mode. My preference would be the former, as it may make tests run faster. But if you really must turn off all optimizations, just keep this change but remove the `-O` from it.

toly-kournik: yes, it's because of debugging.
CXX_FLAGS_DEBUG as it was previously doesn't change the release behaviour

andrewjstone: Got it. Thanks for the explanation.

You are correct, FORTIFY_SOURCE was never on in RELEASE mode. I assume that was only because of fear of false positives. Since we have CI for RELEASE mode and Debian and Ubuntu have it turned on by default I'm not too worried about this.

Since I don't want to cause you or anyone else more problems with the debugger, especially given its unclear the behavior of -Og with clang, let's just not do anything to debug mode.

In short, keep this change without the -O:
```
string(APPEND CXX_FLAGS_RELEASE " -D_FORTIFY_SOURCE=2")
```

toly-kournik: I read that -D_FORTIFY_SOURCE=2 has no effect without -O, is it correct?
If yes, why use it?

toly-kournik:
@techpavan , It passes on my local machine, I'm lost...

CMakeLists.txt
andrewjstone: Release mode already has -O3 set by default. If you pass -O, you will change it to -O2, which lowers the optimization level.

toly-kournik: I mean in debug mode

andrewjstone: Sorry, to be more clear. You need at least some optimizations on for FORTIFY_SOURCE to work. We already have -O3 by default, so it should work fine.

toly-kournik: But I don't want optimisation in debug mode at all

andrewjstone: > I mean in debug mode

Oh, I was suggesting you keep your change that removes the FORTIFY_SOURCE from DEBUG mode. Just remove the -O when setting it for RELEASE.

toly-kournik: ah, ok

andrewjstone: > But I don't want optimisation in debug mode at all

Yes, optimizations are off by default in debug mode, but set to 03 in release mode. So if you just use the line I gave you it should do what you intended, without reducing our release optmizations by a level.

andrewjstone: :)

andrewjstone: BTW, rest of the changes look great, same as last time. 👍

toly-kournik: :-) but for some reason all the debug builds fail on view change test

andrewjstone: Very strange.. I guess take @teoparvanov  up on his suggestion and enjoy your weekend :)

teoparvanov: I believe the failures are due to the replicas becoming slightly slower due to debug mode (combined with time-sensitivity of the test itself). I suggest the following improvements here:
https://github.com/vmware/concord-bft/pull/320

teoparvanov: As expected, when combining PRs https://github.com/vmware/concord-bft/pull/318 and https://github.com/vmware/concord-bft/pull/320, CI succeeds:

- dumm PR: https://github.com/vmware/concord-bft/pull/321
- Travis CI build: https://travis-ci.com/vmware/concord-bft/builds/140952125

PR #319, opened 2019-12-12
Introduce MsgsCommunicator - WIP

yuliasherman:
- Take MsgReceiver and InternalStorage out from ReplicaImp
- Separate internal messages and their execution jobs from launching logic
- Fix format, typos, etc.


yuliasherman:
Needs to be separated into different commits.

PR #320, opened 2019-12-13
Making view change system tests more robust

teoparvanov:
The current view change test suite appears to be time sensitive, hence non-deterministic in some circumstances (slower replicas f.e.).

With this PR we introduce two main improvements:
- throughout, we use a batch of parallel random write requests to trigger view change (instead of a single request, waiting for a reply, which may timeout occasionally)
- the get_view_number() + assert "idiom" is replaced by the more general wait_for_view_number(), with a timeout increased up to 30 sec


teoparvanov:
Hi Andrew, thanks for the review! 

PS: About your remark regarding correctness, there's an item on the backlog to enhance all current test suites with linearizability checks.

PR #322, opened 2019-12-13
Add a utility for calculating SHA3-256 hashes.

andrewjstone:
This is needed for our merkle tree. It relies on openssl 1.1.1.

Relevant to the tests: https://vimeo.com/6973519#t=2m52s

### DIFF
--- a/util/test/sha3_256_tests.cpp
+++ b/util/test/sha3_256_tests.cpp
@@ -1,0 +1,74 @@
+// Concord
+//
+// Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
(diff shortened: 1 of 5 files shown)


util/test/sha3_256_tests.cpp
```diff
@@ -0,0 +1,73 @@
...
+      hash.at(i / 2) |= nibble;
+    }
+  }
+
+  return hash;
+}
+
+TEST(sha3_256_test, basic) {
+  auto sha3 = SHA3_256();
+  auto expected = string_to_array("a7ffc6f8bf1ed76651c14756a061d662f580ff4de43b49fa82d80a4b80f8434a");
```
beerriot: I think a small comment explaining how you computed the expected hashes in this test might be a good idea - to help others verify that these really are the SHA3-256 hashes of these inputs, and not just the output of this implementation.

andrewjstone: Good call!.  Fixed.

PR #323, opened 2019-12-14
Sparse merkle enhancements

andrewjstone:
This is two individual commits to support merkle trees. 
 * SHA3-256 enhancements 
* Base Type enhancements

### DIFF
--- a/storage/include/sparse_merkle/base_types.h
+++ b/storage/include/sparse_merkle/base_types.h
@@ -14,6 +14,7 @@
 #include <vector>
 
 #include "assertUtils.hpp"
+#include "sha3_256.h"
 
 namespace concord {
 namespace storage {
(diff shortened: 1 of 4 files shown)


storage/include/sparse_merkle/base_types.h
```diff
@@ -68,25 +69,56 @@ class Hash {
...
       0xa7, 0xff, 0xc6, 0xf8, 0xbf, 0x1e, 0xd7, 0x66, 0x51, 0xc1, 0x47, 0x56, 0xa0, 0x61, 0xd6, 0x62,
       0xf5, 0x80, 0xff, 0x4d, 0xe4, 0x3b, 0x49, 0xfa, 0x82, 0xd8, 0x0a, 0x4b, 0x80, 0xf8, 0x43, 0x4a};
 
+  Hash() {}
+
+  Hash(std::array<uint8_t, SIZE_IN_BYTES> buf) : buf_(buf) {}
+
+  const uint8_t* data() const { return buf_.data(); }
+  size_t size() const { return buf_.size(); }
+
```
yuliasherman: It's strange to see classes defined in the file called base_types.h :-)

storage/include/sparse_merkle/base_types.h
```diff
@@ -110,6 +145,25 @@ class NibblePath {
     num_nibbles_ += 1;
   }
 
+  // Pop the last nibble off the path
+  Nibble pop_back() {
```
yuliasherman: Just curious, what all this Nibble stuff is used for in merkle trees?

storage/include/sparse_merkle/base_types.h
andrewjstone: I agree. I just couldn't think of a better name. I'm happy to call it something else. They are the basic types needed to build a merkle tree other than the nodes of the trees and keys themselves.

storage/include/sparse_merkle/base_types.h
andrewjstone: In order to reduce the number of database accesses we batch internal nodes of the sparse merkle tree into a 4 level tree. A four level tree has 2^4 = 16 possible leaves, each uniquely identifiable by a nibble (4 bits).  Some of these leaves are pointers to other internal nodes. We can therefore walk from the root of the tree to the actual leaf being searched by taking the hash key, and pulling off a nibble at a time to retrieve the "next" internal node from the database. As we take each nibble from the beginning of the hash of the real key, we build up a `NibblePath` that (along with versions - ignore for now) uniquely identifies a node in the tree. We can walk back up the tree, by popping off the tail nibble off the path to get to the prior internal node.

Picture our tree looks like the following: 3 internal nodes with the leaf pointer precisely in the leaf of the 3rd internal node. Remember from prior discussions we don't need to store any deeper in the tree if there are no subtrees underneath. Root is just an internal node with an empty nibble path.

Root -> 3 -> F -> A (Leaf Pointer)

We hash a key = 3FA12B9...
We need to insert it by walking the tree. We get the first internal node after the root = 3, get the second at 3F, and see that 3FA, a leaf in 3F internal node is empty. We then decide to store the pointer to the leaf the there.

PR #325, opened 2019-12-16
Replica configuration: read-only replica aware configuration files

toly-kournik:
https://jira.eng.vmware.com/browse/VB-2228

Add the following info to the configuration files:
For all types of replicas:
	- amount of read-only replicas
	- whether the replica is read-only
For read-only replicas:
	- don't generate threshold signatures

GenerateConcordKeys now can optionally receive -r N paremeter to indicate the amount of read-only replicas in setup.

### DIFF
--- a/bftengine/src/bftengine/ReplicaLoader.cpp
+++ b/bftengine/src/bftengine/ReplicaLoader.cpp
@@ -83,7 +83,7 @@ ReplicaLoader::ErrorCode checkReplicaConfig(const LoadedReplicaData &ld) {
   Verify(c.concurrencyLevel <= maxLegalConcurrentAgreementsByPrimary, InconsistentErr);
 
   std::set<uint16_t> repIDs;
-  for (auto & v : c.publicKeysOfReplicas) {
+  for (auto &v : c.publicKeysOfReplicas) {
     VerifyAND(v.first >= 0, v.first < numOfReplicas, InconsistentErr);
     Verify(!v.second.empty(), InconsistentErr);  // TODO(GG): make sure that the key is valid
     repIDs.insert(v.first);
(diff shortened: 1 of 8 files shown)


bftengine/src/bftengine/ReplicaLoader.cpp
```diff
@@ -83,7 +83,7 @@ ReplicaLoader::ErrorCode checkReplicaConfig(const LoadedReplicaData &ld) {
   Verify(c.concurrencyLevel <= maxLegalConcurrentAgreementsByPrimary, InconsistentErr);
 
   std::set<uint16_t> repIDs;
-  for (std::pair<uint16_t, std::string> v : c.publicKeysOfReplicas) {
+  for (auto & v : c.publicKeysOfReplicas) {
```
andrewjstone: :) Shorter and more efficient!

tools/GenerateConcordKeys.cpp
```diff
@@ -322,20 +288,10 @@ int main(int argc, char** argv) {
...
-      outputFiles.push_back(std::ofstream(outputPrefix + std::to_string(i)));
-      if (!outputFiles.back().is_open()) {
-        std::cout << "Could not open output file " << outputPrefix << i << ".\n";
-        return -1;
-      }
-    }
-
     std::vector<std::pair<std::string, std::string>> rsaKeys;
-    for (uint16_t i = 0; i < n; ++i) {
+    for (uint16_t i = 0; i < n+ro; ++i) {
```
andrewjstone: This has the possibility to overlow, since we are adding to uint16_t values. In practice it's not an issue, but I think we should stop trying to optimize for configuration and just store everything as uint64_t in the future.

PR #326, opened 2019-12-17
Utilities for BFT network partitioning

teoparvanov:
Utilities for BFT network partitioning

With this PR we introduce the **bft_network_partitioning** module. It offers basic utilities and abstractions for simulating "adversaries" that are capable of interfering with network communication.
The base class for such adversaries is the **NetworkPartitioningAdversary**. It provides several utility methods for dropping packets based on the **iptables** utility, as well as an abstract _interfere()_ method.
Several sample implementations are provided:
- **PassiveAdversary** - basically the "no-op" adversary
- **PrimaryIsolatingAdversary** - identifies and isolates the primary replica's outgoing communication
- **PacketDroppingAdversary** - drops random packets between all replicas

This PR also provides two samples of how the partitioning module can be used:
1) Introduce the PrimaryIsolatingAdversary into view change tests (verify that view change is triggered due to the presence of such an adversary)
2) Introduce the PacketDroppingAdversary into linearizability tests (with 5% of packets dropped). Going above 5% significantly slows down the test, even though it still succeed.

PS: using _iptables_ requires root privileges, hence the "sudo" addition to the commands for starting system tests.
PPS: all BFT-related partitioning rules are defined in a custom _iptables_ rule chain called "bft-network-partitioning". This rule chain is created/cleaned-up as part of the adversary object's lifecycle (via context manager methods)

### DIFF
--- a/tests/test_skvbc_linearizability.py
+++ b/tests/test_skvbc_linearizability.py
@@ -18,6 +18,7 @@ import time
 
 from util import bft, skvbc_history_tracker
 from util import skvbc as kvbc
+from util import bft_network_partitioning as net
 
 KEY_FILE_PREFIX = "replica_keys_"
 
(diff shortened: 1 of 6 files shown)


tests/pythonSudo.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+# This shell script allows running the Python interpreter as root
+# NB: the user needs sudo NOPASSWD enabled
+sudo PYTHONPATH=$PYTHONPATH python3 "$@"
```
teoparvanov: This is needed in order to run tests that require root privileges from within an IDE, such as PyCharm.

tests/util/bft_network_partitioning.py
```diff
@@ -0,0 +1,116 @@
...
+    """ Adversary does nothing = synchronous network """
+
+    def interfere(self):
+        pass
+
+
+class PrimaryIsolatingAdversary(NetworkPartitioningAdversary):
+    """ Adversary that intercepts and drops all outgoing packets from the current primary """
+
+    async def interfere(self):
```
andrewjstone: Does overriding a def abstract method with async def make a difference?

tests/test_skvbc_linearizability.py
```diff
@@ -80,8 +81,8 @@ class SkvbcChaosTest(unittest.TestCase):
     def test_healthy(self):
         """
         Run a bunch of concurrrent requests in batches and verify
-        linearizability. The system is healthy and stable and no faults are
-        intentionally generated.
+        linearizability. No crashed nodes, but some network packets
```
andrewjstone: Why change the healthy test to drop packets? This test was intended to serve as a baseline that our system is linearizable with no faults. 

Since these tests are somewhat expensive, I'm fine changing it, but I'd also change the name of the test to test_with_drops or something.

tests/pythonSudo.sh
andrewjstone: I'm uncomfortable turning on sudo with NOPASSWD on my personal machine. Running `sudo make test` should work though, right?

teoparvanov: Both `sudo ctest` and `sudo make test` work of course, when running the tests from command line. When running from within PyCharm though, there's no easy way to run with sudo, except for using this shell script as the interpreter. NOPASSWD will only be needed in this case, as PyCharm doesn't prompt for root credentials.

tests/test_skvbc_linearizability.py
teoparvanov: +1, the intention is not very clear - I'll create a separate test case for this.

tests/util/bft_network_partitioning.py
teoparvanov: It doesn't seem to... Initially I had made the abstract method async, but then I thought - do we want this to be necassarily a coroutine? Probably not. Such as I see it, "async" is more of an implementation detail, so I moved it to the concrete classes (where applicable).

PR #331, opened 2019-12-19
SparseMerkle: Add BatchedInternalNode

andrewjstone:
The BatchedInternalNode is first major piece of the sparse merkle tree
datastructure. A BatchedInternalNode contains up to 4 levels of a merkle
tree, and is identified by an InternalNodeKey, used for database access.

Insertion into a BatchedInternalNode works now, such that keys can be
added to a single node. In some cases, such as matching prefixes of a
certain length, this can cause the node to split. In this case we inform
the caller of this and tell it to create new nodes and insert the old
stored leaf that triggered the split and the new child attempting to be
inserted. In other cases, insert can discover that the leaf attempting
to be inserted doesn't fit in this node and needs to be placed further
down the tree. There is a return value indicating this as well.

The logic to handle the DB access as well as creation of new nodes and
updates of hashes when required by the creation of a new node is
relegated to the sparse merkle tree logic itself which will come in
further commits.

Base types were modified in order to support the BatchedInternalNode.
Tests were also added to cover the variety of insertion scenarios.

This code heavily uses value types to reduce allocations and keep things
on the stack for performance reasons.

### DIFF
--- a/storage/include/sparse_merkle/base_types.h
+++ b/storage/include/sparse_merkle/base_types.h
@@ -1,6 +1,6 @@
 // Concord
 //
-// Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
 //
 // This product is licensed to you under the Apache 2.0 license (the "License").
 // You may not use this product except in compliance with the Apache 2.0 License.
@@ -9,6 +9,8 @@
 // notices and license terms. Your use of these subcomponents is subject to the
 // terms and conditions of the sub-component's license, as noted in the
 // LICENSE file.
+
+#pragma once
 
 #include <cstdint>
 #include <vector>
@@ -20,24 +22,33 @@ namespace concord {
 namespace storage {
 namespace sparse_merkle {
 
+constexpr size_t BYTE_SIZE_IN_BITS = 8;
+static_assert(BYTE_SIZE_IN_BITS == CHAR_BIT);
+
+// A type safe version wrapper
+struct Version {
+  Version() : value(0) {}
+  Version(uint64_t val) : value(val) {}
+  bool operator==(const Version& other) const { return value == other.value; }
+  uint64_t value;
+};
+
 // A nibble is 4 bits, stored in the lower bits of a byte.
 class Nibble {
  public:
+  static constexpr size_t SIZE_IN_BITS = 4;
+
   Nibble(uint8_t byte) {
     Assert((byte & 0xF0) == 0);
     data_ = byte;
   }
 
-  bool get_bit(const size_t bit) const {
-    Assert(bit < 4);
+  bool getBit(const size_t bit) const {
+    Assert(bit < SIZE_IN_BITS);
     return (data_ >> bit) & 1;
   }
 
   bool operator==(const Nibble& other) const { return data_ == other.data_; }
-
-  // Promote the underlying uint8_t to a size_t so that it can be used as an
-  // index into arrays and vectors.
-  size_t to_index() const { return data_; }
 
   // Return the underlying representation
   uint8_t data() const { return data_; }
@@ -48,12 +59,12 @@ class Nibble {
 };
 
 template <typename T>
-Nibble get_nibble(const size_t n, const T& buf) {
+Nibble getNibble(const size_t n, const T& buf) {
   size_t index = n / 2;
   uint8_t byte = buf[index];
   if (n % 2 == 0) {
     // Even value: Get the first nibble of the byte
-    byte = byte >> 4;
+    byte = byte >> Nibble::SIZE_IN_BITS;
   } else {
     // Odd value: Get the second nibble of the byte
     byte = byte & 0x0F;
@@ -68,7 +79,7 @@ class Hash {
   static constexpr size_t SIZE_IN_BITS = 256;
   static constexpr const char* const HASH_ALGORITHM = "SHA3-256";
 
-  static constexpr size_t SIZE_IN_BYTES = SIZE_IN_BITS / 8;
+  static constexpr size_t SIZE_IN_BYTES = SIZE_IN_BITS / BYTE_SIZE_IN_BITS;
   static constexpr size_t MAX_NIBBLES = SIZE_IN_BYTES * 2;
 
   // This is the hash of the empty string ''. It's used as a placeholder value
@@ -83,14 +94,51 @@ class Hash {
 
   Hash(std::array<uint8_t, SIZE_IN_BYTES> buf) : buf_(buf) {}
 
+  bool operator==(const Hash& other) const { return buf_ == other.buf_; }
+
   const uint8_t* data() const { return buf_.data(); }
   size_t size() const { return buf_.size(); }
 
-  Nibble get_nibble(const size_t n) const {
+  Nibble getNibble(const size_t n) const {
     Assert(!buf_.empty());
     Assert(n < MAX_NIBBLES);
-    return ::concord::storage::sparse_merkle::get_nibble(n, buf_);
+    return ::concord::storage::sparse_merkle::getNibble(n, buf_);
   }
+
+  // Count the number of contiguous bits in common these hashes have from the
+  // start (MSB) until a bit mismatch.
+  size_t prefix_bits_in_common(const Hash& other) const {
+    size_t count = 0;
+    for (size_t i = 0; i < SIZE_IN_BYTES; i++) {
+      if ((buf_[i] ^ other.buf_[i]) == 0) {
+        count += BYTE_SIZE_IN_BITS;
+      } else {
+        // Check bit by bit
+        for (size_t j = 7; j >= 0; j--) {
+          if (((buf_[i] >> j) & 1) == ((other.buf_[i] >> j) & 1)) {
+            ++count;
+          } else {
+            return count;
+          }
+        }
+      }
+    }
+    return count;
+  };
+
+  // Count the number of contiguous bits in common (starting from MSB) common
+  // these hashes have from depth*4 until a bit mismatch.
+  //
+  // TODO: This can be optimized to only start searching from depth
+  size_t prefix_bits_in_common(const Hash& other, size_t depth) const {
+    Assert(depth < Hash::MAX_NIBBLES);
+    auto total = prefix_bits_in_common(other);
+    auto bits_to_depth = depth * Nibble::SIZE_IN_BITS;
+    if (total > bits_to_depth) {
+      return total - bits_to_depth;
+    }
+    return 0;
+  };
 
  private:
   std::array<uint8_t, SIZE_IN_BYTES> buf_;
@@ -127,6 +175,8 @@ class NibblePath {
  public:
   NibblePath() : num_nibbles_(0) {}
 
+  bool operator==(const NibblePath& other) const { return num_nibbles_ == other.num_nibbles_ && path_ == other.path_; }
+
   // Return the length of the path_ in nibbles
   size_t length() const { return num_nibbles_; }
 
@@ -138,7 +188,7 @@ class NibblePath {
     // We don't want a NibblePath to ever be longer than a Hash
     Assert(num_nibbles_ < Hash::MAX_NIBBLES - 1);
     if (num_nibbles_ % 2 == 0) {
-      path_.push_back(nibble.data() << 4);
+      path_.push_back(nibble.data() << Nibble::SIZE_IN_BITS);
     } else {
       path_.back() |= nibble.data();
     }
@@ -146,7 +196,7 @@ class NibblePath {
   }
 
   // Pop the last nibble off the path
-  Nibble pop_back() {
+  Nibble popBack() {
     Assert(!empty());
     uint8_t data = 0;
     if (num_nibbles_ % 2 == 0) {
@@ -156,7 +206,7 @@ class NibblePath {
       path_.back() |= 0xF0;
     } else {
       // We have an incomplete byte at the end of path_. Remove the upper nibble.
-      data = path_.back() >> 4;
+      data = path_.back() >> Nibble::SIZE_IN_BITS;
       // Remove the last byte containing a single nibble.
       path_.pop_back();
     }
@@ -167,7 +217,7 @@ class NibblePath {
   // Get the nth nibble.
   Nibble get(size_t n) const {
     Assert(n < num_nibbles_);
-    return ::concord::storage::sparse_merkle::get_nibble(n, path_);
+    return ::concord::storage::sparse_merkle::getNibble(n, path_);
   }
 
  private:
--- a/storage/include/sparse_merkle/internal_node.h
+++ b/storage/include/sparse_merkle/internal_node.h
@@ -1,0 +1,287 @@
+// Concord
+//
+// Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the sub-component's license, as noted in the
+// LICENSE file.
+
+#pragma once
+
+#include <optional>
+#include <variant>
+#include <algorithm>
+
+#include "sliver.hpp"
+#include "sparse_merkle/keys.h"
+#include "sparse_merkle/base_types.h"
+
+namespace concord {
+namespace storage {
+namespace sparse_merkle {
+
+/*
+A partially full BatchedInternalNode may look like the following:
+
+```
+Level 4               ---------root-------
+                      |                  |
+                  LeafChild       InternalChild
+                                  /            \
+Level 2                     LeafChild    InternalChild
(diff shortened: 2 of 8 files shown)


storage/include/sparse_merkle/keys.h
```diff
@@ -0,0 +1,66 @@
...
+// Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the sub-component's license, as noted in the
+// LICENSE file.
+
```
dartdart26: [Minor] Possibly a missing include guard?

storage/include/sparse_merkle/keys.h
```diff
@@ -0,0 +1,66 @@
...
+};
+
+// The key for a leaf node of a sparse merkle tree.
+//
+// We identify this by the full key hash rather than NibblePath so that we can
+// do direct DB lookups rather than having to walk from the root of the sparse
+// merkle tree.
+class LeafKey {
+ public:
+  LeafKey(Hash key, Version version) : key_(key), version_(version) {}
```
dartdart26: Can we take the key by `const Hash&` to avoid copying?

storage/include/sparse_merkle/keys.h
```diff
@@ -0,0 +1,66 @@
...
+
+  bool operator==(const LeafKey& other) const { return key_ == other.key_ && version_ == other.version_; }
+
+  Nibble get_nibble(size_t n) {
+    Assert(n < Hash::MAX_NIBBLES);
+    return key_.get_nibble(n);
+  }
+
+  Version version() { return version_; }
```
dartdart26: Can we make it `const`?

storage/include/sparse_merkle/keys.h
```diff
@@ -0,0 +1,66 @@
...
+class InternalNodeKey {
+ public:
+  InternalNodeKey(Version version, NibblePath path) : version_(version), path_(path) {}
+
+  // Return the root of a sparse merkle tree at a given version.
+  static InternalNodeKey root(Version version) { return InternalNodeKey(version, NibblePath()); }
+
+  bool operator==(const InternalNodeKey& other) const { return version_ == other.version_ && path_ == other.path_; }
+
+  Version version() { return version_; }
```
dartdart26: Can we make it `const`?

dartdart26:
Big thanks for `internal_node_test.cpp`, great examples and very informative!

storage/include/sparse_merkle/keys.h
andrewjstone: Fixed.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,280 @@
...
+
+  // Return the version of the root node of this BatchedInternalNode
+  Version version() const {
+    auto& root = children_[0];
+    if (!root) return 0;
+    return std::get<InternalChild>(root.value()).version;
+  }
+
+  // Return the number of childrent that are not std::nullopt.
+  size_t num_children() const {
```
yuliasherman: Most of our code has mixed upper-low case function name conventions, like 'executeNextCommittedRequests'.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,280 @@
...
+  bool is_left_child(size_t index) const { return index % 2 != 0; }
+
+  // All internal and leaf nodes are stored in this array, starting from the
+  // root and proceeding level by level from left to right.
+  std::array<std::optional<Child>, MAX_CHILDREN> children_;
+};
+
+}  // namespace sparse_merkle
+}  // namespace storage
+}  // namespace concord
```
yuliasherman: This include file became quite big. You already have corresponding .cpp file, it will probably be better to move even small functions to it.

yuliasherman: BTW, why do we have .h and not .hpp files?

storage/src/sparse_merkle/internal_node.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  // We have reached the leaf of this BatchedInternalNode and it points to another
+  // BatchedInternalNode. Return the Version so the caller can construct an
+  // InternalNodeKey and retrieve the next batched node in the tree.
+  auto& stored_child = children_[index];
+  return BatchedInternalNode::InsertIntoExistingNode{std::get<InternalChild>(stored_child.value()).version};
+}
+
+std::optional<BatchedInternalNode::InsertResult> BatchedInternalNode::insert_into_empty_child(size_t index,
+                                                                                              LeafChild& child) {
```
yuliasherman: 'const LeafChild& child' ?

storage/src/sparse_merkle/internal_node.cpp
```diff
@@ -0,0 +1,171 @@
...
+    stored_child = child;
+    update_hashes(index, child.key.version());
+    return result;
+  }
+  return std::nullopt;
+}
+
+BatchedInternalNode::InsertResult BatchedInternalNode::split_node(size_t index, LeafChild& child, size_t depth) {
+  auto version = child.key.version();
+  auto stored_leaf_child = std::get<LeafChild>(children_.at(index).value());
```
yuliasherman: in overwrite() function you gave a hint: auto**&** stored_child. It was better than here.

storage/src/sparse_merkle/internal_node.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  // We've reached the leaf of this node. Tell the caller to create new
+  // BatchedInternalNodes and insert the previously stored leaf as well as the
+  // leaf it was attempting to insert.
+  //
+  // We purposefully don't call `update_hashes`, since we don't know the true
+  // value of the hash of the child until it gets inserted. The caller will fix
+  // this BatchedInternalNode appropriately when the hash is known.
+  return BatchedInternalNode::CreateNewBatchedInternalNodes{stored_leaf_child};
+}
```
yuliasherman: The function is too big.

storage/src/sparse_merkle/internal_node.cpp
```diff
@@ -0,0 +1,171 @@
...
+    internal_child.version = version;
+    if (parent_idx == 0) {
+      // We have reached the root of this BatchedInternalNode
+      return;
+    }
+    index = parent_idx;
+  }
+}
+
+BatchedInternalNode::InsertResult BatchedInternalNode::insert(LeafChild& child, size_t depth) {
```
dartdart26: Can we make the `child` parameter a const ref -`const LeafChild&`? That also propagates to `insert_into_empty_child()`, `overwrite()` and `split_node()`.

dartdart26: I missed @yuliasherman 's comment above on `insert_into_empty_child()`. This comment adds to it.

storage/include/sparse_merkle/internal_node.h
andrewjstone: All the storage code has .h files. I just kept to that convention.

storage/src/sparse_merkle/internal_node.cpp
andrewjstone: Two different things are happening in these functions. In overwrite, we need a reference so we can **overwrite** the location in `children_`. We don't want a reference here though, we want a copy. If we take a reference, then when we put a new `InternalChild` into `children_`, the reference will contain the new value, and we will have lost the old value that we need for use later in the function. Since the node is being split we need to save the old value so we can insert it later or return it to the caller to insert in a new `BatchedInternalNode`.

storage/include/sparse_merkle/base_types.h
```diff
@@ -92,6 +102,41 @@ class Hash {
...
+  size_t bits_in_common(const Hash& other) const {
+    size_t count = 0;
+    for (size_t i = 0; i < SIZE_IN_BYTES; i++) {
+      if ((buf_[i] ^ other.buf_[i]) == 0) {
+        count += 8;
+      } else {
+        // Check bit by bit
+        for (size_t j = 7; j >= 0; j--) {
+          if (((buf_[i] >> j) & 1) == ((other.buf_[i] >> j) & 1)) {
+            count += 1;
```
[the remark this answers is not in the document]
andrewjstone: fixed.

storage/include/sparse_merkle/base_types.h
```diff
@@ -92,6 +102,41 @@ class Hash {
     return ::concord::storage::sparse_merkle::get_nibble(n, buf_);
   }
 
+  // Count the number of contiguous bits in common these hashes have from the
+  // start until a bit mismatch.
+  size_t bits_in_common(const Hash& other) const {
+    size_t count = 0;
+    for (size_t i = 0; i < SIZE_IN_BYTES; i++) {
+      if ((buf_[i] ^ other.buf_[i]) == 0) {
+        count += 8;
```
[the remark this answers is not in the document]
andrewjstone: fixed.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,280 @@
...
+#include "sliver.hpp"
+#include "sparse_merkle/keys.h"
+#include "sparse_merkle/base_types.h"
+
+namespace concord {
+namespace storage {
+namespace sparse_merkle {
+
+// A child that is a leaf of the overall tree. It isn't necessarily at height 0
+// in a BatchedInternalNode, although there will not be any nodes below it.
```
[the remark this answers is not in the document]
andrewjstone: Thanks for pointing this out. Documenting these types are tricky so it's nice to get feedback on this sort of thing.

Please let me know if this helps.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,280 @@
...
+
+/*
+// A batched internal node. The node is a representation of a 4-level tree, such
+// that there are a total possibility of 16 leaves.
+//
+// Children of this tree are stored in an array. A child may represent an
+// internal node of the overall sparse merkle tree, a leaf, or a placeholder
+// node. If an InternalChild resides at height 0 it will point to another
+// BatchedInternalNode. A leaf will always have a pointer to a key containing a
+// the actual stored data. It may or may not reside at height 0.
```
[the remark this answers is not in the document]
andrewjstone: haha. Fixed.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,280 @@
...
+//
+// The following is a diagram of a complete 4-level tree, stolen shamelessly
+// from
```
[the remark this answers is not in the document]
andrewjstone: It was tongue in cheek, and a common phrase in the open source world. I changed it for professionalism as suggested though.

storage/include/sparse_merkle/internal_node.h
andrewjstone: Fixed. I left single liners there and one double liner that wouldn't make sense apart from those 3.

storage/src/sparse_merkle/internal_node.cpp
andrewjstone: Good catch! Fixed.

storage/include/sparse_merkle/internal_node.h
andrewjstone: Good catch. I was indeed not following the style guide. Fixed.

Now I get to point out when you all use camelCase variable names 😆

storage/src/sparse_merkle/internal_node.cpp
andrewjstone: I broke out the two large clauses at the bottom into their own functions. I think this does make it a bit clearer to see what's going on.

storage/src/sparse_merkle/internal_node.cpp
andrewjstone: Good catch. Fixed.

storage/include/sparse_merkle/base_types.h
```diff
@@ -92,6 +102,41 @@ class Hash {
     return ::concord::storage::sparse_merkle::get_nibble(n, buf_);
   }
 
+  // Count the number of contiguous bits in common these hashes have from the
+  // start until a bit mismatch.
```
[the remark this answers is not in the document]
andrewjstone: Thanks for the detailed read. Added.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,280 @@
...
+
+namespace concord {
+namespace storage {
+namespace sparse_merkle {
+
+// A child that is a leaf of the overall tree. It isn't necessarily at height 0
+// in a BatchedInternalNode, although there will not be any nodes below it.
+struct LeafChild {
+  bool operator==(const LeafChild& other) const { return hash == other.hash && key == other.key; }
+  // The hash of the value blob stored at `key_`
```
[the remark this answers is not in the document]
andrewjstone: Fixed.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,280 @@
...
+
+  // Walk the tree from `index` up to the root of this BatchedInternalNode to
+  // compute the root hash.
+  //
+  // Update the hashes and versions of nodes along the way.
+  void update_hashes(size_t index, Version version);
+
+  // Return the the height of the node at the given index.
+  size_t height(size_t index) {
+    if (index == 0) return 4;
```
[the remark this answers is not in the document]
andrewjstone: Indeed, it's related to the max number of children. I fixed it. Thanks for pointing this out. I struggled with the impl as  well. For clarity and performance, I did the straightforward thing. We can always change that later as needed.

storage/include/sparse_merkle/keys.h
```diff
@@ -0,0 +1,68 @@
...
+
+ private:
+  Version version_;
```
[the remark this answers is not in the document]
andrewjstone: Fixed.

storage/include/sparse_merkle/internal_node.h
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  // Return the index of the left child of a node at a given index.
+  size_t leftChildIndex(size_t index) const { return 2 * index + 1; }
+
+  // Return the index of the right child of a node at a given index.
+  size_t rightChildIndex(size_t index) const { return 2 * index + 2; }
+
+  // Is this node a left child?
+  bool isLeftChild(size_t index) const { return index % 2 != 0; }
```
dartdart26: [Minor]: Does it make sense to move these to the cpp file with internal linkage as they are private and don't use member variables?

storage/test/sparse_merkle/internal_node_test.cpp
```diff
@@ -0,0 +1,373 @@
...
+TEST(internal_node_tests, split_until_new_batch_node_needed) {
+  BatchedInternalNode node;
+
+  Hasher hasher;
+  const char* key1 = "artist";
+  const char* value1 = "REM";
+  const char* value2 = "Nas";
+  size_t depth = 0;
+
+  auto key1_hash = hasher.hash(key1, strlen(key1));
```
dartdart26: I've been debugging and, if not mistaken, the comment about the structure of the tree having all internal children being left ones seems incorrect, because some of the bits in the first nibble of `key1_hash` are 1 and that leads to `insertInternalChildren()` adding right children. Maybe we can change `key1_hash` to reflect the comment as:

```
  auto key1_hash = hasher.hash(key1, strlen(key1));
  // Copy the actual hash and set the first nibble to zeroes to make all
  // internal children left ones.
  std::array<uint8_t, Hash::SIZE_IN_BYTES> key1_hash_data;
  std::copy(key1_hash.data(), key1_hash.data() + key1_hash.size(), key1_hash_data.begin());
  key1_hash_data[0] &= 0x0F;
  key1_hash = Hash{key1_hash_data};
```

dartdart26:
Added a few minor comments. Looks really good, thanks!

storage/include/sparse_merkle/base_types.h
```diff
@@ -92,6 +102,41 @@ class Hash {
...
+  size_t bits_in_common(const Hash& other) const {
```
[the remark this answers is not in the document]
andrewjstone: How about `prefix_bits_in_common` ?

storage/include/sparse_merkle/internal_node.h
andrewjstone: Hmmm. Will consider it for a future commit.

storage/test/sparse_merkle/internal_node_test.cpp
andrewjstone: Great catch! That's a good point. I wasn't really thinking about that. I was more worried that we hit height 0 in this test. I Will change as you suggest, since this makes the diagram work and doesn't change the intent of the test.

storage/include/sparse_merkle/base_types.h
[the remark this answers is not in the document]
andrewjstone: I went with this.

andrewjstone:
I pushed all fixes related to comments and rebased.

PR #334, opened 2019-12-23
Add test for a blinking replica

f-squirrel:


f-squirrel:
@salieri11 , please review ;)

salieri11:
@teoparvanov The basic test for blinking replica is the one I am doing manually - replica should be restarted just when it  boots, synchronizes with the rest and then goes down again - without even reaching one checkpoint. We used this test for persistency and its great to have it as a part of the python suite. The one you suggest is more related to testing byzantine behaviour and of course should be implemented

tests/util/replica_blinker.py
```diff
@@ -0,0 +1,67 @@
...
+
+import argparse
+import random
+import signal
+import subprocess
+import sys
+import time
+
+
+BLINKING_INTERVAL = (1, 10)
```
teoparvanov: BLINKING_INTERVAL_SECONDS?

Actually, I'd imagine the blinking to happen much more frequently, even at a sub-second rate.

In any case, we must make sure to 'blink' at least several times while processing the 300 iterations of write requests.

You could verify this by checking the logs after running the test.

tests/util/bft.py
```diff
@@ -165,14 +175,28 @@ def stop_all_replicas(self):
 
         self.procs = {}
 
+    def start_replicas(self, replicas):
+        """
+        Start from list "replicas"
+        """
+        for r in replicas:
```
teoparvanov: If you want to be more "pythonistic", you could say:
[self.start_replica(r) for r in replicas]

tests/util/bft.py
```diff
@@ -165,14 +175,28 @@ def stop_all_replicas(self):
...
+            self.start_replica(r)
+
+    def stop_replicas(self, replicas):
+        """
+        Start from list "replicas"
+        """
+        for r in replicas:
```
teoparvanov: Same here

tests/util/blinking_replica.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+import subprocess
+
+
+class BlinkingReplica(object):
+    """docstring for BlinkingReplica"""
+
+    def __init__(self):
+        super(BlinkingReplica, self).__init__()
+        self.blinking_replica = None
```
teoparvanov: This should actually say 'blinker_process'

tests/util/blinking_replica.py
```diff
@@ -0,0 +1,51 @@
...
+    def start_blinking(self,start_replica_cmd):
+        self.start_replica_cmd = start_replica_cmd
+        self.blinking_replica = subprocess.Popen(self._cmd_line(), close_fds=True)
+        print("Started blinking: {}".format(self._cmd_line()))
+
+    def stop_blinking(self):
+        if self.blinking_replica:
+            self.blinking_replica.terminate()
+            if self.blinking_replica.wait() != 0:
+                raise Exception("Error occured while running blinking replica")
```
teoparvanov: ... while stopping the blinker process

tests/util/replica_blinker.py
```diff
@@ -0,0 +1,67 @@
...
+
+import argparse
+import random
+import signal
+import subprocess
+import sys
+import time
+
+
+BLINKING_INTERVAL_SECONDS = (1, 10)
```
teoparvanov: Are you sure a blinking interval of 10 seconds makes sense?

PR #339, opened 2019-12-26
Throw by value, catch by reference...

f-squirrel:


f-squirrel:
@salieri11 , @yuliasherman please review

PR #340, opened 2019-12-26
Add assignment operator for G1T and G2T

f-squirrel:
Since the two classes have non-trivial copy constructors they seem to
need user-defined assignment operators.
Used in concord-bft/threshsign/src/bls/relic/BlsBatchVerifier.h:274


f-squirrel:
@yuliasherman , please review

PR #341, opened 2019-12-30
Throw exception in Timers class in case timer not found

yuliasherman:
Return the previous behavior - don't ignore non-existing timers.


andrewjstone:
LGTM. Can you please restore the test as well?

yuliasherman:
Yes, sure, I forgot about it.

PR #345, opened 2020-01-06
Reduce boilerplate code in Python tests by using decorators

teoparvanov:
This PR introduces two new utility decorators which take care of two widely repeated tasks:
 * running async test methods with trio
 * creating a new BFT test network for each test

The @with_trio decorator removes the need to have two methods per test (test* and _test*).
On the other hand, @with_bft_network eliminates the copy/pasted code snippet for initializing
the BftTestNetwork in the beginning of each test.

Most of the changes in this PR have to do with deleting the obsolete boilerplate code.

PS: I have also updated tests/README.md with respect to the above changes.

### DIFF
--- a/tests/util/bft.py
+++ b/tests/util/bft.py
@@ -21,6 +21,8 @@ import random
 import subprocess
 from collections import namedtuple
 import tempfile
+from functools import wraps
+
 import trio
 
 sys.path.append(os.path.abspath("../util/pyclient"))
@@ -41,9 +43,12 @@ TestConfig = namedtuple('TestConfig', [
     'start_replica_cmd'
 ])
 
+KEY_FILE_PREFIX = "replica_keys_"
 
-def interesting_configs(
-        selected=lambda *config: True):
+
+def interesting_configs(selected=None):
+    if selected is None:
+        selected=lambda *config: True
 
     bft_configs = [{'n': 4, 'f': 1, 'c': 0, 'num_clients': 4},
                    {'n': 6, 'f': 1, 'c': 1, 'num_clients': 4},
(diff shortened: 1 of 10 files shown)


tests/util/bft.py
```diff
@@ -60,11 +65,45 @@ def interesting_configs(
...
+        return trio.run(async_fn, *args, **kwargs)
+
+    return trio_wrapper
+
+
+def with_bft_network(start_replica_cmd, selected_configs=None):
+    """
+    Runs the decorated async function for all selected BFT configs
+    """
+    def decorator(async_fn):
```
andrewjstone: I'm a bit confused about how the double nesting works here. It appears to me that when `@with_bft_network(start_replica_cmd)` is used as a decorator it should return `decorator`, which is a standard (non-async) function. Then `@with_trio` runs which will wrap `decorator` and return` trio_wrapper` as the test to run. But when `trio_wrapper` runs it will attempt to call `trio.run` on `decorator` which is not async. However, clearly that can't be the case or none of this would work. So can you please help clarify my understanding? of how things get unwrapped etc...

teoparvanov: Hey Andrew, thanks, great question! 

The `with_bft_network` function acts as a _decorator factory_. It produces a closure called `decorator` which has access to the `@with_bft_network` parameters. This closure is the actual decorator applied to the underlying async function (`decorator` takes `async_fn` as an argument). From here onwards, it's as if we applied `@decorator` to the underlying function - `wrapper` replaces the decorated function, preserving its identity & metadata using `@wraps`.

Thus, the output when applying `@with_bft_network` is an async function which can be further decorated by using `@with_trio`.

Phew :) I hope that clarifies things. Let me know if I'm missing something.

teoparvanov: PS: here's a much more detailed explanation, if you wish to dig deeper into the subject:
https://gist.github.com/Zearin/2f40b7b9cfc51132851a#passing-arguments-to-the-decorator

salieri11:
Just great job! Thanks, +1 to merge

tests/util/bft.py
andrewjstone: Wow. That took me a while to wrap my head around. No pun intended! 

That blog post was great. And it does match your explanation.

 I think the part I was missing was that when the function after the `@` has parentheses, it is evaluated before the `@` is applied to the decorated function. This makes sense, as the `@` needs a function object. It has no idea what to do with those parentheses :D

PR #348, opened 2020-01-12
Test "read-your-writes" in view change tests

Liorgal28:
@teoparvanov please review after required changes.


teoparvanov:
This looks much better, thanks Lior!

PS: can you also use the skvbc.read_your_writes() method in test_skvbc.py, because the same code is repeated quite a lot?

tests/util/skvbc.py
```diff
@@ -287,6 +287,30 @@ def _create_keys(self):
 
         return keys
 
+    async def _read_your_writes(self, bft_network, test_class):
```
teoparvanov: this is a public method, so it shouldn't have a leading "_"

tests/test_skvbc_auto_view_change.py
```diff
@@ -51,16 +51,20 @@ class SkvbcAutoViewChangeTest(unittest.TestCase):
...
         # do nothing - just wait for an automatic view change
-        await bft_network.wait_for_view_change(
+        await bft_network.wait_for_view(
             replica_id=random.choice(
                 bft_network.all_replicas(without={initial_primary})),
             expected=lambda v: v > initial_primary,
             err_msg="Make sure automatic view change has occurred."
         )
 
+        skvbc._read_your_writes(bft_network, self)
```
teoparvanov: You should remove "_" here.

tests/test_skvbc_auto_view_change.py
```diff
@@ -51,16 +51,20 @@ class SkvbcAutoViewChangeTest(unittest.TestCase):
...
         # do nothing - just wait for an automatic view change
-        await bft_network.wait_for_view_change(
+        await bft_network.wait_for_view(
             replica_id=random.choice(
                 bft_network.all_replicas(without={initial_primary})),
             expected=lambda v: v > initial_primary,
             err_msg="Make sure automatic view change has occurred."
         )
 
+        skvbc.read_your_writes(bft_network, self)
```
teoparvanov: By the way, no need to pass a _bft_network_ parameter here - the _SimpleKVBCProtocol_ already has one, accessible via _self.bft_network_.

PR #349, opened 2020-01-13
SparseMerkle: Add initial tree implementation

andrewjstone:
The first implementation of the tree was added, such that batches of
key-value pairs can be inserted via a call to `Update`. The call to
update is an iterative wrapper around calls to
`BatchedInternalNode::insert`, with one call for each KV pair in the
update.

A call to `Tree::update` returns an `UpdateBatch` with data that should
be written to the database. Besides internal and leaf nodes of the
latest updates that need writing, the `UpdateBatch` also contains
information about stale nodes such that we can prune those nodes later.
The database is never directly modified.  This allows for easier testing
and the opportunity to batch disperate writes into single transactions.
However, in order for `update` to operate, it must be able to read
existing key-value pairs from the database. This is performed via the
`IDBReader` interface. Furthermore a cache of these nodes is created and
maintained during each update call, since multiple BatchedInternalNodes
may be modified via multiple insert calls and we always want to operate
on the latest updates.

Each insert must find the correct BatchedInternalNode to insert into,
and then walk back up the tree to the root updating the hashes along
the way. This is performed via the use of a stack. Each
BatchedInternalNode is pushed onto the stack as we descend the tree and
then popped off when we reverse course and update toward the root. This
pop loop occurs when a BatchedInternalNode::insert call returns
`InsertComplete`.

Note that all KV pairs in an update call share the same version, and
therefore, each modified BatchedInternalNode or LeafNode that needs to
be written to the database will be keyed by the same version.

There are a bunch of test covering the update behavior, and test
implementation of IDBReader, called `TestDB`.

Additionally, a bunch of the underlying types were enhanced to add new
operators for comparison, etc...

Lastly, during testing two existing bugs were found:

 1. Nibble.popBack was improperly returning the first Nibble of full
 bytes because an `or` was being used instead of an `and`. The test for
 that function hid it, because the test data was an `F`, which just so
 happens to be the data returned by the bug. The existing test was
 change to use a `7` rather than an `F`. This shows the need for
 property based testing where we can test "every possible input" (via
 randomness) and get more input coverage. We should bring in RapidCheck
 for this.

 2. In `BatchedInternalNode::insertTwoLeafChildren`, we were inserting
 at the wrong index in the node because we were calling
 `child_key.getBit` on the number of prefix_bits_in_common, when we
 should have been starting from the next bit after that to find the
 insertion points, since the two children matched up to
 prefix_bits_in_common and differ in the next bit.

### DIFF
--- a/storage/include/sparse_merkle/base_types.h
+++ b/storage/include/sparse_merkle/base_types.h
@@ -57,7 +57,7 @@ class Nibble {
 
   // Get the bit of the Nibble starting from LSB.
   // Bits 0-3 are available.
-  bool getBit(const size_t bit) const {
+  bool getBit(size_t bit) const {
     Assert(bit < SIZE_IN_BITS);
     return (data_ >> bit) & 1;
   }
--- a/storage/include/sparse_merkle/base_types.h
+++ b/storage/include/sparse_merkle/base_types.h
@@ -26,11 +26,23 @@ constexpr size_t BYTE_SIZE_IN_BITS = 8;
 static_assert(BYTE_SIZE_IN_BITS == CHAR_BIT);
 
 // A type safe version wrapper
-struct Version {
-  Version() : value(0) {}
-  Version(uint64_t val) : value(val) {}
-  bool operator==(const Version& other) const { return value == other.value; }
-  uint64_t value;
+class Version {
+ public:
+  Version() {}
+  Version(uint64_t val) : value_(val) {}
+  bool operator==(const Version& other) const { return value_ == other.value_; }
+  bool operator!=(const Version& other) const { return value_ != other.value_; }
+  bool operator<(const Version& other) const { return value_ < other.value_; }
+  Version operator+(const Version& other) { return Version(value_ + other.value_); }
+  Version operator+(const int other) {
+    Assert(other > 0);
+    return Version(value_ + 1);
+  }
+  uint64_t value() { return value_; }
+  std::string toString() const { return std::to_string(value_); };
+
+ private:
+  uint64_t value_ = 0;
 };
 
 // A nibble is 4 bits, stored in the lower bits of a byte.
@@ -43,15 +55,26 @@ class Nibble {
     data_ = byte;
   }
 
+  // Get the bit of the Nibble starting from LSB.
+  // Bits 0-3 are available.
   bool getBit(const size_t bit) const {
     Assert(bit < SIZE_IN_BITS);
     return (data_ >> bit) & 1;
   }
 
   bool operator==(const Nibble& other) const { return data_ == other.data_; }
+  bool operator<(const Nibble& other) const { return data_ < other.data_; }
 
   // Return the underlying representation
   uint8_t data() const { return data_; }
+
+  // Return the nibble as its lowercase hex character.
+  char hexChar() const {
+    if (data_ >= 0 && data_ <= 9) {
+      return '0' + data_;
+    }
+    return 'a' + (data_ - 10);
+  }
 
  private:
   // Only the lower 4 bits are used
@@ -95,9 +118,20 @@ class Hash {
   Hash(std::array<uint8_t, SIZE_IN_BYTES> buf) : buf_(buf) {}
 
   bool operator==(const Hash& other) const { return buf_ == other.buf_; }
+  bool operator!=(const Hash& other) const { return buf_ != other.buf_; }
+  bool operator<(const Hash& other) const { return buf_ < other.buf_; }
 
   const uint8_t* data() const { return buf_.data(); }
   size_t size() const { return buf_.size(); }
+
+  // Return buf_ as lowercase hex string
+  std::string toString() const {
+    std::string output;
+    for (size_t i = 0; i < MAX_NIBBLES; i++) {
+      output.push_back(getNibble(i).hexChar());
+    }
+    return output;
+  }
 
   Nibble getNibble(const size_t n) const {
     Assert(!buf_.empty());
@@ -176,6 +210,21 @@ class NibblePath {
   NibblePath() : num_nibbles_(0) {}
 
   bool operator==(const NibblePath& other) const { return num_nibbles_ == other.num_nibbles_ && path_ == other.path_; }
+  bool operator<(const NibblePath& other) const {
+    size_t count = std::min(num_nibbles_, other.num_nibbles_);
+    for (size_t i = 0; i < count; i++) {
+      if (get(i) < other.get(i)) {
+        return true;
+      } else if (other.get(i) < get(i)) {
+        return false;
+      }
+    }
+    // Both match up to count nibbles. Is this one shorter ?
+    if (num_nibbles_ < other.num_nibbles_) {
+      return true;
+    }
+    return false;
+  }
 
   // Return the length of the path_ in nibbles
   size_t length() const { return num_nibbles_; }
@@ -203,7 +252,7 @@ class NibblePath {
       // We have a complete byte at the end of path_. Remove the lower nibble.
       data = path_.back() & 0x0F;
       // Clear the lower nibble.
-      path_.back() |= 0xF0;
+      path_.back() &= 0xF0;
     } else {
       // We have an incomplete byte at the end of path_. Remove the upper nibble.
       data = path_.back() >> Nibble::SIZE_IN_BITS;
@@ -218,6 +267,15 @@ class NibblePath {
   Nibble get(size_t n) const {
     Assert(n < num_nibbles_);
     return ::concord::storage::sparse_merkle::getNibble(n, path_);
+  }
+
+  // Return path_ as lowercase hex string
+  std::string toString() const {
+    std::string output;
+    for (size_t i = 0; i < num_nibbles_; i++) {
+      output.push_back(get(i).hexChar());
+    }
+    return output;
   }
 
  private:
--- a/storage/include/sparse_merkle/internal_node.h
+++ b/storage/include/sparse_merkle/internal_node.h
@@ -180,6 +180,12 @@ class BatchedInternalNode {
   // `depth` represents how many nibbles down the sparse merkle tree this BatchedInternalNode is.
   InsertResult insert(const LeafChild& child, size_t depth);
 
+  // Write an InternalChild of the BatchedInternalNode at level 0.
+  //
+  // This occurs when this InternalChild poiints to another BatchedInternalNode
+  // down the tree, and that BatchedInternalNode was just updated.
+  void write_internal_child_at_level_0(Nibble child_key, const InternalChild& child);
+
   // Return the root hash of this node.
   const Hash& hash() const { return getHash(0); }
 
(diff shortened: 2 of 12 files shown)


storage/include/sparse_merkle/base_types.h
```diff
@@ -43,16 +55,27 @@ class Nibble {
     data_ = byte;
   }
 
+  // Get the bit of the Nibble starting from LSB.
+  // Bits 0-3 are available.
   bool getBit(const size_t bit) const {
```
yuliasherman: "const size_t bit" - const does not make sense as 'bit' is passed by value

storage/include/sparse_merkle/base_types.h
```diff
@@ -43,16 +55,27 @@ class Nibble {
...
 
   // Return the underlying representation
   uint8_t data() const { return data_; }
 
+  // Return the nibble as its lowercase hex character.
+  char hexChar() const {
+    if (data_ >= 0 && data_ <= 9) {
+      return '0' + data_;
+    }
+    return 'a' + (data_ - 10);
```
yuliasherman: Do we want to verify that data_ does not exceed 0x0F?

storage/include/sparse_merkle/internal_node.h
```diff
@@ -180,6 +180,12 @@ class BatchedInternalNode {
   // `depth` represents how many nibbles down the sparse merkle tree this BatchedInternalNode is.
   InsertResult insert(const LeafChild& child, size_t depth);
 
+  // Write an InternalChild of the BatchedInternalNode at level 0.
+  //
+  // This occurs when this InternalChild poiints to another BatchedInternalNode
```
yuliasherman: poiints - typo

storage/include/sparse_merkle/tree.h
```diff
@@ -0,0 +1,98 @@
...
+    std::vector<std::pair<LeafKey, LeafNode>> leaf_nodes;
+  };
+
+  typedef std::stack<BatchedInternalNode, std::vector<BatchedInternalNode>> NodeStack;
+
+  explicit Tree(std::shared_ptr<IDBReader> db_reader) : db_reader_(db_reader) {
+    // Perform a single allocation of the maximum depth of the tree for the
+    // insert_stack_.
+    auto vec = std::vector<BatchedInternalNode>();
+    vec.reserve(Hash::MAX_NIBBLES);
```
yuliasherman: MAX_NIBBLES is constant, so you can specify it already in the vector constructor.

storage/src/sparse_merkle/tree.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+using namespace concordUtils;
+
+namespace concord::storage::sparse_merkle {
+
+// All data that is loaded from the DB, manipulated, and/or returned from a
+// call to `update`.
+//
+// A new instance of this structure is created during every update call.
+struct UpdateCache {
```
yuliasherman: I think it's better to have those structures to be defined in the include file.

storage/src/sparse_merkle/tree.cpp
```diff
@@ -0,0 +1,198 @@
...
+  UpdateCache(const BatchedInternalNode& root, const std::shared_ptr<IDBReader>& db_reader) : db_reader_(db_reader) {
+    internal_nodes_[NibblePath()] = root;
+  }
+
+  // Get a node if it's in the cache, otherwise get it from the DB.
+  BatchedInternalNode getInternalNode(const InternalNodeKey& key) {
+    auto it = internal_nodes_.find(key.path());
```
yuliasherman: In this MR you sometimes use struct members with trailing '_' and sometimes not :-)

storage/src/sparse_merkle/tree.cpp
```diff
@@ -0,0 +1,198 @@
...
+  for (int i = 0; i < nodes_to_create - 1; i++) {
+    // Add a bunch of empty intermediate nodes
+    state.nibble_path.append(child_to_insert.key.getNibble(state.depth));
+    state.stack.emplace(BatchedInternalNode());
+    state.depth += 1;
+  }
+  // Create the bottom most node and make it the current node
+  state.nibble_path.append(child_to_insert.key.getNibble(state.depth));
+  state.current_node = BatchedInternalNode();
+  state.depth += 1;
```
yuliasherman: ++ - kind reminder :-)

storage/src/sparse_merkle/tree.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+// A node split has occurred, due to a collision, triggering the need to create more
+// BatchedInternalNodes.
+void create_internal_nodes(const LeafChild& stored_child, const LeafChild& child_to_insert, InsertState& state) {
+  int nodes_to_create = child_to_insert.key.hash().prefix_bits_in_common(stored_child.key.hash(), state.depth) / 4;
+  for (int i = 0; i < nodes_to_create - 1; i++) {
+    // Add a bunch of empty intermediate nodes
+    state.nibble_path.append(child_to_insert.key.getNibble(state.depth));
+    state.stack.emplace(BatchedInternalNode());
+    state.depth += 1;
```
yuliasherman: ++ - kind reminder :-)

storage/src/sparse_merkle/tree.cpp
```diff
@@ -0,0 +1,198 @@
...
+  }
+}
+
+// Modify the insert state to walk further down the tree
+void descend(Nibble next_nibble, Version next_version, UpdateCache& cache, InsertState& state) {
+  state.nibble_path.append(next_nibble);
+  InternalNodeKey next_internal_key{next_version, state.nibble_path};
+  state.current_node = cache.getInternalNode(next_internal_key);
+  cacheStaleInternalNode(cache, state);
+  state.depth += 1;
```
yuliasherman: ++ - kind reminder :-)

storage/src/sparse_merkle/tree.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// A node split occurred due to a key collision.
+//
+// Create a bunch of new internal nodes and then insert the previously stored
+// child that collided with the child that was attempting to be inserted.
+//
+// On the next loop around the `child_to_insert` will be successfully inserted
+// inside `state.current_node`.
+void handleNodeSplit(BatchedInternalNode::CreateNewBatchedInternalNodes result,
```
yuliasherman: Any reason for passing this by value?

storage/src/sparse_merkle/tree.cpp
```diff
@@ -0,0 +1,198 @@
...
+                     const LeafChild& child_to_insert,
+                     InsertState& state) {
+  create_internal_nodes(result.stored_child, child_to_insert, state);
+
+  // Ignore results. We already know this insert will succeed.
+  state.current_node.insert(result.stored_child, state.depth);
+}
+
+// Insert a single leaf. This is called as part of `update`.
+void insert(UpdateCache& cache, const LeafChild& child, const Version version, Tree::NodeStack& stack) {
```
yuliasherman: const Version version - remove const

storage/include/sparse_merkle/base_types.h
andrewjstone: It's already a constraint on construction of the Nibble.

storage/include/sparse_merkle/internal_node.h
andrewjstone: Good eye! :)

storage/include/sparse_merkle/tree.h
andrewjstone: That doesn't work, since that will chance the `size` of the vector. We only want to change the `capacity`, but keep the size 0, because we don't want anything on the stack. We could do that and then call clear() I guess, but I believe that may be more expensive because it includes initialization.

storage/src/sparse_merkle/tree.cpp
andrewjstone: I originally had them in there, but it presented a few problems. For example, since they were private, internally linked (plain functions) in the CPP file could not use them. Making them public or putting them in the namespace, also effectively makes them public and part of the API. They are both purely implementation details so I put them in the cpp file. They are both subject to change with future optimzations.

storage/src/sparse_merkle/tree.cpp
andrewjstone: haha. old habit from Rust.

storage/src/sparse_merkle/tree.cpp
andrewjstone: Nope. Good catch!

storage/src/sparse_merkle/tree.cpp
andrewjstone: I'm also open to adding an "implementation class" in a future PR if necessary. That would allow direct testing of these helper functions.

storage/include/sparse_merkle/base_types.h
```diff
@@ -26,11 +26,23 @@ constexpr size_t BYTE_SIZE_IN_BITS = 8;
...
+  Version() {}
+  Version(uint64_t val) : value_(val) {}
+  bool operator==(const Version& other) const { return value_ == other.value_; }
+  bool operator!=(const Version& other) const { return value_ != other.value_; }
+  bool operator<(const Version& other) const { return value_ < other.value_; }
+  Version operator+(const Version& other) { return Version(value_ + other.value_); }
+  Version operator+(const int other) {
+    Assert(other > 0);
+    return Version(value_ + 1);
+  }
```
dartdart26: Can we either make them `const` or make them free functions? Additionally, the `int` overload only increments by one and not by `other` - I guess that is not the desired behaviour.

storage/include/sparse_merkle/base_types.h
```diff
@@ -26,11 +26,23 @@ constexpr size_t BYTE_SIZE_IN_BITS = 8;
...
+  uint64_t value() { return value_; }
```
dartdart26: Can we make it `const`?

storage/include/sparse_merkle/base_types.h
```diff
@@ -95,10 +118,21 @@ class Hash {
...
 
+  // Return buf_ as lowercase hex string
+  std::string toString() const {
+    std::string output;
+    for (size_t i = 0; i < MAX_NIBBLES; i++) {
+      output.push_back(getNibble(i).hexChar());
+    }
+    return output;
+  }
+
```
dartdart26: Can we share implementation with `NibblePath::toString()`?

storage/include/sparse_merkle/base_types.h
andrewjstone: woah! Nice find on the increment bug!

storage/include/sparse_merkle/base_types.h
andrewjstone: Maybe with some template chicanery. But I think it's simple enough that the duplication doesn't really matter here.

PR #353, opened 2020-01-15
Refactoring and update linearizability tests

Liorgal28:
1) Extract common utility functions from SkvbcChaosTest to SkvbcTracker.
2) Move Class Status from "test_skvbc_linearizability.py" to "skvbc_history_tracker" - status class helps us to track reads and writes so its main usage is from history tracker.

### DIFF
--- a/tests/test_skvbc_linearizability.py
+++ b/tests/test_skvbc_linearizability.py
@@ -63,9 +63,9 @@ class SkvbcChaosTest(unittest.TestCase):
         self.tracker = skvbc_history_tracker.SkvbcTracker(init_state, self.skvbc, self.bft_network)
         self.bft_network.start_all_replicas()
         async with trio.open_nursery() as nursery:
-            nursery.start_soon(self.tracker.run_concurrent_ops, num_ops)
+            nursery.start_soon(self.run_concurrent_ops, num_ops)
 
-        await self.tracker.verify_linearizability()
+        await self.tracker.fill_missing_blocks_and_verify()
 
     @with_trio
     @with_bft_network(start_replica_cmd)
@@ -88,9 +88,9 @@ class SkvbcChaosTest(unittest.TestCase):
             adversary.interfere()
 
             async with trio.open_nursery() as nursery:
-                nursery.start_soon(self.tracker.run_concurrent_ops, num_ops)
+                nursery.start_soon(self.run_concurrent_ops, num_ops)
 
-            await self.tracker.verify_linearizability()
+            await self.tracker.fill_missing_blocks_and_verify()
 
     @with_trio
     @with_bft_network(start_replica_cmd)
@@ -109,10 +109,25 @@ class SkvbcChaosTest(unittest.TestCase):
         self.tracker = skvbc_history_tracker.SkvbcTracker(init_state, self.skvbc, self.bft_network)
         self.bft_network.start_all_replicas()
         async with trio.open_nursery() as nursery:
-            nursery.start_soon(self.tracker.run_concurrent_ops, num_ops)
+            nursery.start_soon(self.run_concurrent_ops, num_ops)
             nursery.start_soon(self.crash_primary)
 
-        await self.tracker.verify_linearizability()
+        await self.tracker.fill_missing_blocks_and_verify()
+
+    async def run_concurrent_ops(self, num_ops):
+        max_concurrency = len(self.bft_network.clients) // 2
+        write_weight = .70
+        max_size = len(self.skvbc.keys) // 2
+        sent = 0
+        while sent < num_ops:
+            clients = self.bft_network.random_clients(max_concurrency)
+            async with trio.open_nursery() as nursery:
+                for client in clients:
+                    if random.random() < write_weight:
+                        nursery.start_soon(self.tracker.send_tracked_write, client, max_size)
+                    else:
+                        nursery.start_soon(self.tracker.send_tracked_read, client, max_size)
+            sent += len(clients)
 
     async def crash_primary(self):
         await trio.sleep(.5)
--- a/tests/test_skvbc_linearizability.py
+++ b/tests/test_skvbc_linearizability.py
@@ -97,9 +97,11 @@ class SkvbcChaosTest(unittest.TestCase):
         self.status = Status(bft_network.config)
         bft_network.start_all_replicas()
         async with trio.open_nursery() as nursery:
-            nursery.start_soon(self.run_concurrent_ops, num_ops)
+            nursery.start_soon(self.tracker.run_concurrent_ops, self, num_ops)
+            #nursery.start_soon(self.run_concurrent_ops, num_ops)
 
-        await self.verify()
+        #await self.verify()
+        await self.tracker.verify_linearizability(self)
 
     @with_trio
     @with_bft_network(start_replica_cmd)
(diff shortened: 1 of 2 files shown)


tests/util/skvbc_history_tracker.py
```diff
@@ -724,3 +729,110 @@ def _get_matching_request(self, rpy):
         causal_state = self.outstanding[(rpy.client_id, rpy.seq_num)]
         index = causal_state.req_index
         return (self.history[index], index)
+
+    async def verify_linearizability(self,test_case):
```
teoparvanov: I don't like the fact that we pass the test_case to the tracker. It shouldn't be aware of any test-specific stuff. Its job is to track and verify linearizability.

I suggest the following changes:
- create a new tracker instance for each test
- pass a bft_network instance to the tracker instance and use it internally where necessary (for creating a BFT client f.e.)
- refactor the debug statements (print ...), so that they don't depend on the specific test_case object

tests/test_skvbc_linearizability.py
```diff
@@ -21,6 +21,7 @@
 from util import skvbc as kvbc
 from util import skvbc_history_tracker
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
+from util.skvbc_history_tracker import SkvbcTracker as with_tracker
```
teoparvanov: this doesn't seem to be used anywhere - also the 'with_tracker' name seems weird...

tests/util/skvbc_history_tracker.py
```diff
@@ -724,3 +735,110 @@ def _get_matching_request(self, rpy):
...
+            print(f'Retrieved block {block_id}')
+        return blocks
+
+    async def get_last_block_id(self, client):
+        msg = kvbc.SimpleKVBCProtocol.get_last_block_req()
+        return kvbc.SimpleKVBCProtocol.parse_reply(await client.read(msg))
+
+    async def crash_primary(self):
+        await trio.sleep(.5)
+        self.bft_network.stop_replica(0)
```
teoparvanov: nothing to do with the tracker - this should return to test_linearizability.py

tests/util/skvbc_history_tracker.py
```diff
@@ -724,3 +735,110 @@ def _get_matching_request(self, rpy):
...
+            clients = self.bft_network.random_clients(max_concurrency)
+            async with trio.open_nursery() as nursery:
+                for client in clients:
+                    if random.random() < write_weight:
+                        nursery.start_soon(self.send_linearizability_write, client, max_size)
+                    else:
+                        nursery.start_soon(self.send_linearizability_read, client, max_size)
+            sent += len(clients)
+
+    async def send_linearizability_write(self, client, max_set_size):
```
teoparvanov: send_tracked_write() sounds better IMO

tests/util/skvbc_history_tracker.py
```diff
@@ -724,3 +735,110 @@ def _get_matching_request(self, rpy):
...
+        try:
+            serialized_reply = await client.write(msg, seq_num)
+            self.status.record_client_reply(client_id)
+            reply = self.skvbc.parse_reply(serialized_reply)
+            self.handle_write_reply(client_id, seq_num, reply)
+        except trio.TooSlowError:
+            self.status.record_client_timeout(client_id)
+            return
+
+    async def send_linearizability_read(self, client, max_set_size):
```
teoparvanov: send_tracked_read() ?

tests/util/skvbc_history_tracker.py
```diff
@@ -351,6 +356,12 @@ def __init__(self, initial_kvpairs={}):
         # These blocks were created by write requests that never got replies.
         self.filled_blocks = {}
 
+        self.skvbc = skvbc
+
+        self.bft_network = bft_network
+
+        self.status = status
```
teoparvanov: Why do we need this? It seems to be test-specific - please drop if possible

tests/util/skvbc_history_tracker.py
```diff
@@ -196,6 +201,39 @@ def __repr__(self):
            f'  kvpairs={self.kvpairs}\n'
            f'  req_index={self.req_index}\n')
 
+class Status:
+    """
+    Status about the running test.
+    This is useful for debugging if the test fails.
+
+    TODO: Should this live in the tracker?
```
teoparvanov: this comment is now obsolete...

teoparvanov: update comment please (do not mention tests)

tests/util/skvbc_history_tracker.py
```diff
@@ -196,6 +201,37 @@ def __repr__(self):
            f'  kvpairs={self.kvpairs}\n'
            f'  req_index={self.req_index}\n')
 
+class Status:
+    """
+    Status about the order of writes and reads.
+    This is useful for debugging and checking linearizability.
```
andrewjstone: This class isn't involved in checking linearizability. It's soley for debugging after a test fails.

tests/util/skvbc_history_tracker.py
```diff
@@ -724,3 +766,106 @@ def _get_matching_request(self, rpy):
...
+                    if i == retries - 1:
+                        raise
+            print(f'Retrieved block {block_id}')
+        return blocks
+
+    async def get_last_block_id(self, client):
+        msg = kvbc.SimpleKVBCProtocol.get_last_block_req()
+        return kvbc.SimpleKVBCProtocol.parse_reply(await client.read(msg))
+
+    async def run_concurrent_ops(self, num_ops):
```
andrewjstone: This is a single specific test case. I don't think it should be part of the tracker without some slight modification. For example, the percentage of reads and writes is hardcoded (.7) and the set size is limited (to try to cause conflicts), and also the max concurrency is limited. We may want to make these configurable in a future PR.

teoparvanov: +1, makes sense, let's return `run_concurrent_ops()` in the test module. We can keep it there until we figure it can be useful in other tests. Then we can generalize it and make it a utility, rather than a test method,

tests/util/skvbc_history_tracker.py
teoparvanov: Indeed, let's update the Status class documentation, as suggested by Andrew.

tests/util/skvbc_history_tracker.py
```diff
@@ -351,6 +387,12 @@ def __init__(self, initial_kvpairs={}):
         # These blocks were created by write requests that never got replies.
         self.filled_blocks = {}
 
+        self.skvbc = skvbc
+
+        self.bft_network = bft_network
+
+        self.status = Status(bft_network.config)
```
teoparvanov: To fix the CI failure, this should become:
```
if bft_network is not None:
    self.status = Status(bft_network.config)
```

tests/util/skvbc_history_tracker.py
```diff
@@ -724,3 +766,106 @@ def _get_matching_request(self, rpy):
         causal_state = self.outstanding[(rpy.client_id, rpy.seq_num)]
         index = causal_state.req_index
         return (self.history[index], index)
+
+    async def verify_linearizability(self):
```
teoparvanov: fetch_missing_blocks_and_verify()

tests/util/skvbc_history_tracker.py
```diff
@@ -196,6 +201,37 @@ def __repr__(self):
            f'  kvpairs={self.kvpairs}\n'
            f'  req_index={self.req_index}\n')
 
+class Status:
+    """
+    Status about the order of writes and reads.
+    This is useful for debugging when test is fall.
```
teoparvanov: "... for debugging linearizability check failures."

PR #357, opened 2020-01-17
Await "read-your-writes" checks in view change tests

teoparvanov:
The skvbc.read_your_writes() calls were not awaited, hence a co-routine object would get created, but never executed.
Note: Those checks are important as they validate that we can properly interact with the SKVBC following various view change scenarios.

### DIFF
--- a/kvbc/src/ReplicaImp.cpp
+++ b/kvbc/src/ReplicaImp.cpp
@@ -334,7 +334,15 @@ void ReplicaImp::insertBlockInternal(BlockId blockId, Sliver block) {
 }
 
 Sliver ReplicaImp::getBlockInternal(BlockId blockId) const {
-  assert(blockId <= m_lastBlock);
+  if (blockId > m_lastBlock) {
+    LOG_ERROR(logger,
+              "Block " << blockId
+                       << " not processed yet. "
+                          "Last executed block is: "
+                       << m_lastBlock);
+    return Sliver();
+  }
+
   Sliver retVal;
 
   bool found;
--- a/kvbc/src/ReplicaImp.cpp
+++ b/kvbc/src/ReplicaImp.cpp
@@ -334,15 +334,7 @@ void ReplicaImp::insertBlockInternal(BlockId blockId, Sliver block) {
 }
 
 Sliver ReplicaImp::getBlockInternal(BlockId blockId) const {
-  if (blockId > m_lastBlock) {
-    LOG_ERROR(logger,
-              "Block " << blockId
-                       << " not processed yet. "
-                          "Last executed block is: "
-                       << m_lastBlock);
-    return Sliver();
-  }
-
+  assert(blockId <= m_lastBlock);
   Sliver retVal;
 
   bool found;
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1908,8 +1908,6 @@ void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNew
              viewsManager->stableLowerBoundWhenEnteredToView());
 
   Assert(viewsManager->viewIsActive(curView));
-  metric_view_.Get().Set(curView);
-
   Assert(lastStableSeqNum >= viewsManager->stableLowerBoundWhenEnteredToView());
   Assert(lastExecutedSeqNum >= lastStableSeqNum);  // we moved to the new state, only after synchronizing the state
 
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1908,6 +1908,8 @@ void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNew
              viewsManager->stableLowerBoundWhenEnteredToView());
 
   Assert(viewsManager->viewIsActive(curView));
+  metric_view_.Get().Set(curView);
+
   Assert(lastStableSeqNum >= viewsManager->stableLowerBoundWhenEnteredToView());
   Assert(lastExecutedSeqNum >= lastStableSeqNum);  // we moved to the new state, only after synchronizing the state
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -19,7 +19,7 @@ add_test(NAME skvbc_slow_path_tests COMMAND sh -c
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_test(NAME skvbc_view_change_tests COMMAND sudo sh -c
-        "python3 -m unittest test_skvbc_view_change"
+        "python3 -m unittest test_skvbc_view_change 2>&1 > /dev/null"
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_test(NAME skvbc_auto_view_change_tests COMMAND sh -c
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -19,7 +19,7 @@ add_test(NAME skvbc_slow_path_tests COMMAND sh -c
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_test(NAME skvbc_view_change_tests COMMAND sudo sh -c
-        "python3 -m unittest test_skvbc_view_change 2>&1 > /dev/null"
+        "python3 -m unittest test_skvbc_view_change"
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_test(NAME skvbc_auto_view_change_tests COMMAND sh -c
--- a/tests/test_skvbc_auto_view_change.py
+++ b/tests/test_skvbc_auto_view_change.py
@@ -63,7 +63,7 @@ class SkvbcAutoViewChangeTest(unittest.TestCase):
             err_msg="Make sure automatic view change has occurred."
         )
 
-        skvbc.read_your_writes(self)
+        await skvbc.read_your_writes(self)
 
     @with_trio
     @with_bft_network(start_replica_cmd)
@@ -91,7 +91,7 @@ class SkvbcAutoViewChangeTest(unittest.TestCase):
             err_msg="Make sure automatic view change has occurred."
         )
 
-        skvbc.read_your_writes(self)
+        await skvbc.read_your_writes(self)
 
     @with_trio
     @with_bft_network(start_replica_cmd)
@@ -122,4 +122,4 @@ class SkvbcAutoViewChangeTest(unittest.TestCase):
         await skvbc.assert_kv_write_executed(key, val)
         await bft_network.assert_fast_path_prevalent()
 
-        skvbc.read_your_writes(self)
+        await skvbc.read_your_writes(self)
--- a/tests/test_skvbc_view_change.py
+++ b/tests/test_skvbc_view_change.py
@@ -49,6 +49,7 @@ class SkvbcViewChangeTest(unittest.TestCase):
         2) Make sure the initial view is preserved during those writes.
         3) Stop the primary replica and send a batch of write requests.
         4) Verify the BFT network eventually transitions to the next view.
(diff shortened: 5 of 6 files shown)


teoparvanov:
I just noticed there's a C++ assertion failure, which might just explain the actual CI test failures:

_kvbc_replica: /home/travis/build/vmware/concord-bft/kvbc/src/**ReplicaImp.cpp:337**: concordUtils::Sliver concord::kvbc::ReplicaImp::getBlockInternal(concordUtils::BlockId) const: Assertion `**blockId <= m_lastBlock**' failed._

andrewjstone:
> I just noticed there's a C++ assertion failure, which might just explain the actual CI test failures:
> 
> _kvbc_replica: /home/travis/build/vmware/concord-bft/kvbc/src/**ReplicaImp.cpp:337**: concordUtils::Sliver concord::kvbc::ReplicaImp::getBlockInternal(concordUtils::BlockId) const: Assertion `**blockId <= m_lastBlock**' failed._

This test appears to be failing because of a race condition in which view change
for all nodes does not finish, before we call `read_your_writes`.

First let me try to explain why I think the segault is occurring in the single node case. I can trigger this segfault locally on a single node when I log (without redirecting to `/dev/null`), and the test still passes. I can cause the test to fail similar to CI (with logging going to `/dev/null`) when 2 or 3 nodes segfault.

What I believe is happening to cause the segfault is that quorum is achieved for the write, and
the client returns that the last written block is block 1, since it was indeed generated.
But since view change hasn't completed on node 2 (the node that asserted), node 2 didn't
handle the write, since writes get ignored during view change. Then a read request
for block 1 comes in to node2 which is still in view change. Since *read* requests are handled during view change, node 2 tries to get the block data and the assert fires. The read actually causing the assert to fire is [this line](https://github.com/vmware/concord-bft/blob/master/tests/util/skvbc.py#L311). I can avoid the asserts altogether by commenting that read out.

You can also see from the log that view change hasn't finished for node 2, since
`tryToEnterView` is called, but not `onNewView`:

```
30: INFO 2020-01-17 11:27:31.368 (skvbctest.replica)ConditionalWrite message handled; writesCounter=1 currBlock=1
30: INFO 2020-01-17 11:27:31.270 (concord) Node 2 received ReplicaStatusMsg from node 3
30: INFO 2020-01-17 11:27:31.271 (concord) Node 2 received ViewChangeMsg (generatedReplicaId=0, newView=3, lastStable=0, numberOfElements=0).
30: INFO 2020-01-17 11:27:31.271 (concord) ViewChangeMsg add=1
30: INFO 2020-01-17 11:27:31.271 (concord) Primary asks to leave view (primary Id=0 , view=0)
30: INFO 2020-01-17 11:27:31.271 (concord) **************** In MoveToHigherView (curView=0, nextView=1, wasInPrevViewNumber=1)
30: INFO 2020-01-17 11:27:31.285 (concord) Sending view change message: new view=1, wasInPrevViewNumber=1, new primary=1, lastExecutedSeqNum=0, lastStableSeqNum=0
30: INFO 2020-01-17 11:27:31.287 (concord) maxKnownCorrectView=0, maxKnownAgreedView=0
30: INFO 2020-01-17 11:27:31.287 (concord) Node 2 received ViewChangeMsg (generatedReplicaId=4, newView=3, lastStable=0, numberOfElements=0).
30: INFO 2020-01-17 11:27:31.287 (concord) ViewChangeMsg add=1
30: INFO 2020-01-17 11:27:31.287 (concord) maxKnownCorrectView=1, maxKnownAgreedView=0
30: INFO 2020-01-17 11:27:31.287 (concord) Node 2 received ViewChangeMsg (generatedReplicaId=1, newView=3, lastStable=0, numberOfElements=0).
30: INFO 2020-01-17 11:27:31.288 (concord) ViewChangeMsg add=1
30: INFO 2020-01-17 11:27:31.288 (concord) Primary asks to leave view (primary Id=1 , view=1)
30: INFO 2020-01-17 11:27:31.288 (concord) **************** In MoveToHigherView (curView=1, nextView=2, wasInPrevViewNumber=0)
30: INFO 2020-01-17 11:27:31.288 (concord) Sending view change message: new view=2, wasInPrevViewNumber=0, new primary=2, lastExecutedSeqNum=0, lastStableSeqNum=0
30: INFO 2020-01-17 11:27:31.289 (concord) maxKnownCorrectView=3, maxKnownAgreedView=0
30: INFO 2020-01-17 11:27:31.289 (concord) **************** In MoveToHigherView (curView=2, nextView=3, wasInPrevViewNumber=0)
30: INFO 2020-01-17 11:27:31.289 (concord) Sending view change message: new view=3, wasInPrevViewNumber=0, new primary=3, lastExecutedSeqNum=0, lastStableSeqNum=0
30: INFO 2020-01-17 11:27:31.291 (concord) maxKnownCorrectView=3, maxKnownAgreedView=0
30: INFO 2020-01-17 11:27:31.291 (concord) Node 2 received ViewChangeMsg (generatedReplicaId=3, newView=3, lastStable=0, numberOfElements=0).
30: INFO 2020-01-17 11:27:31.291 (concord) ViewChangeMsg add=1
30: INFO 2020-01-17 11:27:31.291 (concord) maxKnownCorrectView=3, maxKnownAgreedView=3
30: INFO 2020-01-17 11:27:31.291 (concord) **************** Calling to viewsManager->tryToEnterView(curView=3, lastStableSeqNum=0, lastExecutedSeqNum=0)
30: INFO 2020-01-17 11:27:31.291 (concord) Node 2 received ViewChangeMsg (generatedReplicaId=5, newView=3, lastStable=0, numberOfElements=0).
30: INFO 2020-01-17 11:27:31.291 (concord) ViewChangeMsg add=1
30: INFO 2020-01-17 11:27:31.291 (concord) maxKnownCorrectView=3, maxKnownAgreedView=3
30: INFO 2020-01-17 11:27:31.291 (concord) **************** Calling to viewsManager->tryToEnterView(curView=3, lastStableSeqNum=0, lastExecutedSeqNum=0)
30: INFO 2020-01-17 11:27:31.291 (concord) Node 2 received ViewChangeMsg (generatedReplicaId=6, newView=3, lastStable=0, numberOfElements=0).
30: INFO 2020-01-17 11:27:31.291 (concord) ViewChangeMsg add=1
30: INFO 2020-01-17 11:27:31.291 (concord) maxKnownCorrectView=3, maxKnownAgreedView=3
30: INFO 2020-01-17 11:27:31.291 (concord) **************** Calling to viewsManager->tryToEnterView(curView=3, lastStableSeqNum=0, lastExecutedSeqNum=0)
30: INFO 2020-01-17 11:27:31.291 (concord) Node 2 received ReplicaStatusMsg from node 4
30: INFO 2020-01-17 11:27:31.291 (concord) Node 2 received ReplicaStatusMsg from node 5
30: INFO 2020-01-17 11:27:31.291 (concord) Node 2 received ReplicaStatusMsg from node 6
30: INFO 2020-01-17 11:27:31.333 (concord) Node 2 received ReplicaStatusMsg from node 0
30: INFO 2020-01-17 11:27:31.333 (concord) Node 2 received StartSlowCommitMsg for seqNumber 1
30: INFO 2020-01-17 11:27:31.351 (concord) Node 2 received ReplicaStatusMsg from node 1
30: INFO 2020-01-17 11:27:31.369 (concord) Node 2 received ClientRequestMsg (clientId=7 reqSeqNum=6624019224169807872, readOnly=1) from Node 7
30: INFO 2020-01-17 11:27:31.369 (skvbctest.replica)Execute GET_BLOCK_DATA command: type=4, BlockId=1
30: ERROR 2020-01-17 11:27:31.369 (skvbc.replicaImp)blockId = 1, m_lastBlock = 0
30: ERROR 2020-01-17 11:27:31.369 (concord) Assert: expression 'blockId <= m_lastBlock' is false in function getBlockInternal (/home/andrewstone/concord-bft/kvbc/src/ReplicaImp.cpp 338)
```

I think at a minimum we should remove this assert, and instead return an error
to the caller or drop the request. Malicious clients can easily trigger this
assert, and it can accidentally happen for the reasons described here. Reads
should still work, since a quorum of nodes must agree for the read to be
successful.

Additionally, it appears our check for when view change completes is not
actually correct. We are checking that the `lastAgreedView` is what we expect, but on a
given node, we may not have actually entered that view yet. This is because [we set the lastAgreedView before actually entering the view](https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L1746-L1747)!

We need to add new metric and check that [viewsManager->viewIsActive(curView)](https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L1910) is true.

My guess for why the test fails on a timeout on [this line](https://github.com/vmware/concord-bft/blob/master/tests/util/skvbc.py#L302) is because view change hasn't completed for a quorum of nodes and so [writes just get dropped](https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L212-L216). I also am assuming here that the read assertion happens in a different configuration or subtest that passes and the test only fails for one configuration/test where the write is blocked.

Unfortunately, since logging is redirected to `/dev/null` we can't tell for sure if this is the case.
With logging on it may change the timing enough that the test no longer fails. 

So my suggestions stand:
 * Fix how we wait for view change to complete with a new metric
 * Remove the assert in kvbc

Hopefully those two changes will fix the test and remove a dangerous assert for production.

Phew, that was a wall of text. @guyg7 does this analysis make sense?

guyg7:
Regarding “Additionally, it appears our check for when view change completes is not
actually correct. “: You are right. It is not sufficient to check  lastAgreedView . You should make sure that the current view change is active.

Regarding “My guess for why the test fails on a timeout on this line is because view change hasn't completed for a quorum of nodes and so writes just get dropped. “: I am not sure about this. The view-change protocol is an internal module that should not affect the behavior of the state machine (from the client’s point of view).

kvbc/src/ReplicaImp.cpp
```diff
@@ -334,7 +334,15 @@ void ReplicaImp::insertBlockInternal(BlockId blockId, Sliver block) {
 }
 
 Sliver ReplicaImp::getBlockInternal(BlockId blockId) const {
-  assert(blockId <= m_lastBlock);
+  if (blockId > m_lastBlock) {
+    LOG_ERROR(logger,
```
guyg7: This should never happen in the old desgin of the KVBC. I suggest reconsidering the issue

teoparvanov:
Hi @guyg7, thanks for looking into this. I have the following two questions:
1) While view change is ongoing is it expected that messages get dropped? This is what seems to happen during those failures and why the client is waiting indefinitely for a quorum of replies (2f+c+1).
2) In the context of the test, how can we determine a view is active? Should we check for at least 2f+c+1 nodes for which the last agreed view is active?

kvbc/src/ReplicaImp.cpp
teoparvanov: Do you mean we should keep the assert, instead of returning an empty sliver?

teoparvanov:
Here is a log from the failure in the case of n=7, f=2, c=0...
I've disabled all tests but **test_single_vc_only_primary_down()**.

It appears that the BFT network is attempting a second view change, which might explain why we lose liveness. I would speculate the reason for this is the write from "read-your-writes" doesn't execute in time, which would trigger the view change.

Should we maybe increase the view change timer (currently set to 3 seconds)?

[view_change_test_log_n7_f2_c0.log](https://github.com/vmware/concord-bft/files/4082447/view_change_test_log_n7_f2_c0.log)

kvbc/src/ReplicaImp.cpp
guyg7: No. I only suggest thinking about it again. According to the original design, this should never happen (unless the code contains a bug).
However, I may be missing something (I am not aware of the updates made in recent months).

guyg7:
I suspect the bug here is not really related to view-change. According the last lines in Andrew’s log, node/replica 2 receives a read-only request to read block 1. However, block 1 does not exist in replica 2 (notice that block 1 may already exist in the other replicas). In such a case, I expect replica 2 to return “block does not exists” (this is not an error). This should not be a problem to the client, because the other replicas should send the required data from block 1.

teoparvanov:
Closing this PR - I'd prefer to create a new one for the actual fixes, as this one got polluted with a lot of WIP/CI testing commits.

PR #358, opened 2020-01-17
Add support for a block info initialization value in BlockchainView

dartdart26:
BlockchainView's handling of BlockInfo is lacking in terms of a missing
option to pass arbitrary data when constructing BlockInfo objects. For
example, the user might want to pass pointers to storage handles in
order to load indices and data. This PR remedies that by adding a
BlockInfoInit type to BlockchainView and calling the BlockInfo
constructor with the passed initialization value. Backward compatibility
is preserved by making BlockInfoInit optional. Tests are added to show
the new changes work.

Additionally, this PR adds more documentation about the requirements on
the BlockInfo type. Tests are added too.

### DIFF
--- a/kvbc/include/blockchain_view.h
+++ b/kvbc/include/blockchain_view.h
@@ -41,16 +41,19 @@ using BlockIdDifferenceType = std::make_signed<concordUtils::BlockId>::type;
 // The maximum possible number of blocks this view is able to hold.
 inline constexpr auto MAX_BLOCKCHAIN_VIEW_SIZE = std::numeric_limits<detail::BlockIdDifferenceType>::max();
 
+// Represents lack of block info initialization data.
+struct NoBlockInfoInit {};
+
 // A random access blockchain iterator that supports caching block info.
-template <typename BlockInfo, bool loadDataOnAccess>
-class BlockchainIterator : public boost::iterator_facade<BlockchainIterator<BlockInfo, loadDataOnAccess>,
+template <typename BlockInfo, typename BlockInfoInit, bool loadDataOnAccess>
+class BlockchainIterator : public boost::iterator_facade<BlockchainIterator<BlockInfo, BlockInfoInit, loadDataOnAccess>,
                                                          const BlockInfo,  // The value type this iterator points to.
                                                          boost::random_access_traversal_tag,
                                                          const BlockInfo&,  // Reference to the value type.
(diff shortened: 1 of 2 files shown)


kvbc/include/blockchain_view.h
```diff
@@ -74,12 +87,13 @@ class BlockchainIterator : public boost::iterator_facade<BlockchainIterator<Bloc
...
-    auto blockInfo = BlockInfo{id};
-    blockInfo.loadIndices();
-    if constexpr (loadDataOnAccess) {
-      blockInfo.loadData();
+    if constexpr (std::is_same_v<BlockInfoInit, BlockInfoNullInit>) {
+      auto blockInfo = BlockInfo{id};
+      loadAndCacheInfo(blockInfo);
+    } else {
+      auto blockInfo = BlockInfo{id, blockInfoInit_};
+      loadAndCacheInfo(blockInfo);
```
teoparvanov: I guess `loadAndCacheInfo(blockInfo)` can go outside of the "if" statement

kvbc/include/blockchain_view.h
```diff
@@ -74,12 +87,13 @@ class BlockchainIterator : public boost::iterator_facade<BlockchainIterator<Bloc
     }
 
     // load data from storage and, if successful, add to cache
-    auto blockInfo = BlockInfo{id};
-    blockInfo.loadIndices();
-    if constexpr (loadDataOnAccess) {
-      blockInfo.loadData();
+    if constexpr (std::is_same_v<BlockInfoInit, BlockInfoNullInit>) {
```
teoparvanov: BlockInfoNoInit ?

dartdart26: Good point, doing as you suggest.

kvbc/include/blockchain_view.h
dartdart26: Yes, it can. However, the intention is that we want to either impose `BlockInfo{id}` or `BlockInfo{id, blockInfoInit_}` constructors on `BlockInfo` and constructing `blockInfo` outside of the if-statement will be problematic.

dartdart26:
Pushed changes based on comments.

kvbc/include/blockchain_view.h
dartdart26: Actually, `NoBlockInfoInit`?

PR #359, opened 2020-01-20
Fixing the "read-your-writes" checks in view change tests

teoparvanov:
Fixing the "read-your-writes" checks in view change tests

The key takeaway from this effort is that our current view change mechanism is actually quite slow (in the order of 10-s of seconds).
This is why we need to increase the view change timeout, to avoid a second view change kicking-in because the first one did not complete in time.

From our discussions with Guy, it appears the "sanitary minimum" for the view change timeout is around 10 seconds (it should be noted that he was testing with 20 seconds). We may also need to increase this value to make the test more stable. Additionally, to help view change complete in a single iteration, we also bump up the network communication buffer size (a small buffer may result in lost messages for networks bigger than 4 nodes).

The rest of the changes presented here mainly have to do with making the `wait_for_view()` implementation more realistic, waiting for the view change to actually become active, instead of just agreed.

This PR introduces the following changes:
- [SBFT] Introduce a new SBFT metric for tracking the current active view of a given replica
- [SBFT/TesterReplica] Increase the buffer size for network communication when testing
- [Tests] Await the skvbc.read_your_writes() calls
- [Tests] Modify wait_for_view() so that it waits for the new view to become active (in addition to have been "agreed upon")
- [Tests] Make sure the next expected primary is not stopped (otherwise two view changes in a row occur)

### DIFF
--- a/tests/util/bft.py
+++ b/tests/util/bft.py
@@ -289,60 +289,87 @@ class BftTestNetwork:
         if expected is None:
             expected = lambda _: True
 
+        matching_view = None
+        nb_replicas_in_matching_view = 0
         try:
-            # Step 1: Wait for an agreed view that matches the "expected" predicate
-            last_agreed_view = None
-            replicas_in_expected_view = 0
-            with trio.fail_after(seconds=30):
-                while True:
-                    try:
-                        with trio.move_on_after(seconds=1):
-                            key = ['replica', 'Gauges', 'lastAgreedView']
-                            view = await self.metrics.get(replica_id, *key)
-                            if expected(view):
-                                last_agreed_view = view
-                                break
-                    except KeyError:
-                        # metrics not yet available, continue looping
-                        continue
+            matching_view = await self._wait_for_matching_agreed_view(replica_id, expected)
+            print(f'Matching view #{matching_view} has been agreed among replicas.')
 
-            # Step 2: Wait for enough (n-f) replicas to activate the last agreed view
-            with trio.fail_after(seconds=30):
-                while True:
-                    view = 0
-                    replicas_in_expected_view = 0
+            nb_replicas_in_matching_view = await self._wait_for_active_view(matching_view)
+            print(f'View #{matching_view} has been activated by '
+                  f'{nb_replicas_in_matching_view} >= n-f = {self.config.n - self.config.f}')
 
-                    async def count_view_participants(r, expected_view):
-                        nonlocal view
-                        nonlocal replicas_in_expected_view
-
-                        key = ['replica', 'Gauges', 'currentActiveView']
-
-                        with trio.move_on_after(seconds=5):
-                            while True:
-                                with trio.move_on_after(seconds=1):
-                                    try:
-                                        view = await self.metrics.get(r, *key)
-                                        if view == expected_view:
-                                            replicas_in_expected_view += 1
-                                    except KeyError:
-                                        # metrics not yet available, continue looping
-                                        continue
-                                    else:
-                                        break
-
-                    async with trio.open_nursery() as nursery:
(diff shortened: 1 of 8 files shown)


teoparvanov:
I have addressed the main comments by @guyg7 - specifically, increasing the view change timeout to 10 sec, and bumping up the network communication buffer size.

One of the tests is still unstable - `test_multiple_vc_slow_path()`. I suggest to disable it for now and investigate this separately, once we enable log4cpp in Travis CI, where the failures can be observed.

The remaining 3 test cases are stable - I have rerun the entire CI job several times, without seeing any failures.

@andrewjstone, could you please review for merging into master?

Thanks!

tests/util/bft.py
```diff
@@ -290,19 +290,59 @@ def get_live_replicas(self):
             expected = lambda _: True
 
         try:
+            # Step 1: Wait for an agreed view that matches the "expected" predicate
```
andrewjstone: I get what you are trying to do here with the use of closures, but the nesting makes it hard to see what the loops contain. Can you please break this up into non-nested named functions?

andrewjstone:
Thanks for the good work @teoparvanov. Thanks very much for the help @guyg7. It is much appreciated!

This Generally looks correct, but I want to make sure I am properly understanding the code paths by cleaning it a bit into smaller functions. Thanks!

tests/util/bft.py
```diff
@@ -289,60 +289,87 @@ def get_live_replicas(self):
...
+                    with trio.move_on_after(seconds=1):
+                        try:
+                            replica_view = await self.metrics.get(r, *key)
+                            if replica_view == expected_view:
+                                nb_replicas_in_view += 1
+                        except KeyError:
+                            # metrics not yet available, continue looping
+                            continue
+                        else:
+                            break
```
andrewjstone: I believe you meant to do this, since there's another loop in `_wait_for_active_view` that retries, but I want to make sure. This will break and not retry if `replica_view != expected_view`. However, other replicas may still try for 5 seconds if they are unavailable... We could take that time to retry in this loop. I don't think there's a tradeoff here, since even if all replicas are available and get the non matching view they would just retry in this loop vs the `wait_for_active_view` loop.

Either way, it's not a big deal and the code is correct. I think my suggestion may be an optimization in future commits though.

tests/util/bft.py
```diff
@@ -289,60 +289,87 @@ def get_live_replicas(self):
...
-                        with trio.move_on_after(seconds=1):
-                            key = ['replica', 'Gauges', 'lastAgreedView']
-                            view = await self.metrics.get(replica_id, *key)
-                            if expected(view):
-                                last_agreed_view = view
-                                break
-                    except KeyError:
-                        # metrics not yet available, continue looping
-                        continue
+            matching_view = await self._wait_for_matching_agreed_view(replica_id, expected)
```
andrewjstone: Do we even need this check, since we check the expected view for n-f replicas in `_wait_for_active_view`. I guess the difference here is the predicate.

tests/util/bft.py
teoparvanov: Thanks Andrew, we can indeed address this as a separate optimization.

tests/util/bft.py
teoparvanov: Well, actually the predicate might say something like `lambda v: v > 0`.
So we don't know exactly which view number we are waiting for to be activated.
So, I've decided to first wait for a view that matches the predicated to be agreed. Only then we wait for this view to be activated. 

It's a sort of two-step process which helps preserve determinism and... one's sanity when facing view change :)

PS: this all started sounding very Lovecraftian, so I'll probably stop here :)))

PR #360, opened 2020-01-20
Add linearizability checks to fast path tests

Liorgal28:
add verify_linearizability decorator to the existing tests to track and verify linearizability.

### DIFF
--- a/tests/test_skvbc_fast_path.py
+++ b/tests/test_skvbc_fast_path.py
@@ -13,7 +13,7 @@
 import os.path
 import random
 import unittest
-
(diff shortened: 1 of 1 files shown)


tests/test_skvbc_fast_path.py
```diff
@@ -13,7 +13,7 @@
 import os.path
 import random
 import unittest
-
+from util.skvbc_history_tracker import verify_linearizability
```
teoparvanov: !?

teoparvanov: won't you also decorate the test methods with @check_linearizability?

PR #361, opened 2020-01-20
Add linearizability checks to fast path tests

Liorgal28:
1)add verify_linearizability decorator for each test.
2)replace regular KV writing to tracked KV writing.
3)check KV writing successful execution at the end of the decorator using tracker verify.

### DIFF
--- a/tests/test_skvbc_fast_path.py
+++ b/tests/test_skvbc_fast_path.py
@@ -79,8 +79,11 @@ class SkvbcFastPathTest(unittest.TestCase):
         Finally the decorator verifies the KV execution.
         """
         bft_network.start_all_replicas()
-        for _ in range(10):
-            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.5)
+
+        write_weight = 1
+        numops = 10
+
+        await tracker.run_concurrent_ops(num_ops=numops, write_weight=write_weight)
 
         await bft_network.assert_fast_path_prevalent()
 
@@ -88,10 +91,9 @@ class SkvbcFastPathTest(unittest.TestCase):
         bft_network.stop_replica(
             replica=random.choice(unstable_replicas))
 
-        for _ in range(10):
-            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.5)
+        await tracker.run_concurrent_ops(num_ops=numops, write_weight=write_weight)
 
-        await bft_network.assert_slow_path_prevalent(as_of_seq_num=10)
+        await bft_network.assert_slow_path_prevalent(as_of_seq_num=numops+1)
 
     @with_trio
     @with_bft_network(start_replica_cmd,
@@ -114,15 +116,15 @@ class SkvbcFastPathTest(unittest.TestCase):
         for _ in range(bft_network.config.c):
             replica_to_stop = random.choice(unstable_replicas)
             bft_network.stop_replica(replica_to_stop)
-
+        write_weight = 0.5
         # make sure we first downgrade to the slow path...
-        for _ in range(self.evaluation_period_seq_num):
-            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.5)
+
+        writings = await tracker.run_concurrent_ops(num_ops=self.evaluation_period_seq_num, write_weight=write_weight)
         await bft_network.assert_slow_path_prevalent()
 
         # ...but eventually (after the evaluation period), the fast path is restored!
-        for _ in range(self.evaluation_period_seq_num + 1,
-                       self.evaluation_period_seq_num * 2):
-            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.5)
+        num_ops = self.evaluation_period_seq_num / write_weight
+        await tracker.run_concurrent_ops(num_ops=num_ops, write_weight=write_weight)
+
         await bft_network.assert_fast_path_prevalent(
             nb_slow_paths_so_far=self.evaluation_period_seq_num)
--- a/tests/test_skvbc_fast_path.py
+++ b/tests/test_skvbc_fast_path.py
@@ -55,10 +55,8 @@ class SkvbcFastPathTest(unittest.TestCase):
         Finally the decorator verifies the KV execution.
         """
         bft_network.start_all_replicas()
-        max_size = 1
         for _ in range(10):
-            client=bft_network.random_client()
-            await tracker.send_tracked_write(client, max_size)
+            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.7)
 
         await bft_network.assert_fast_path_prevalent()
 
@@ -81,10 +79,8 @@ class SkvbcFastPathTest(unittest.TestCase):
         Finally the decorator verifies the KV execution.
         """
         bft_network.start_all_replicas()
-        max_size = 1
         for _ in range(10):
-            client = bft_network.random_client()
-            await tracker.send_tracked_write(client, max_size)
+            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.5)
 
         await bft_network.assert_fast_path_prevalent()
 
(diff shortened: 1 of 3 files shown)


tests/test_skvbc_fast_path.py
```diff
@@ -78,13 +78,13 @@ def setUp(self):
...
+        Finally the decorator verifies the KV execution.
         """
         bft_network.start_all_replicas()
-        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
-
+        max_size = 1
         for _ in range(10):
-            await skvbc.write_known_kv()
+            client = bft_network.random_client()
+            await tracker.send_tracked_write(client, max_size)
```
teoparvanov: To make things more interesting from the linearizability perspective, let's also do a `await tracker.send_tracked_read()` after each `await tracker.send_tracked_write()`. I suggest to do this for every test.

teoparvanov: I was wondering, would it make sense to generalize `run_concurrent_ops()` from `SkvbcChaosTest`, so that we use it for the fast path tests?

Let's pull  `run_concurrent_ops()` it into SkvbcTracker by parameterizing it
and then using it in the fast path tests (instead of sending just 10 sequential writes).

tests/test_skvbc_fast_path.py
```diff
@@ -93,45 +89,40 @@ def setUp(self):
             replica=random.choice(unstable_replicas))
 
         for _ in range(10):
-            key, val = await skvbc.write_known_kv()
+            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.5)
```
teoparvanov: Sorry, I don't understand here... Why not simply do:

```python
            key, val = await skvbc.write_known_kv()
            await tracker.run_concurrent_ops(num_ops=10, write_weight=0.5)
```

tests/util/skvbc_history_tracker.py
```diff
@@ -869,4 +869,18 @@ def readset(self, min_size, max_size):
     def writeset(self, max_size):
         writeset_keys = self.skvbc.random_keys(random.randint(0, max_size))
         writeset_values = self.skvbc.random_values(len(writeset_keys))
-        return list(zip(writeset_keys, writeset_values))
\ No newline at end of file
+        return list(zip(writeset_keys, writeset_values))
+
+    async def run_concurrent_ops(self, num_ops, write_weight=.70):
+        max_concurrency = min(num_ops, len(self.bft_network.clients) // 2)
```
teoparvanov: If you address my comment above, this should simply become as before:
```python
max_concurrency = len(self.bft_network.clients) // 2
```

tests/test_skvbc_fast_path.py
```diff
@@ -93,45 +89,40 @@ def setUp(self):
...
-
         unstable_replicas = bft_network.all_replicas(without={0})
         for _ in range(bft_network.config.c):
             replica_to_stop = random.choice(unstable_replicas)
             bft_network.stop_replica(replica_to_stop)
 
         # make sure we first downgrade to the slow path...
         for _ in range(self.evaluation_period_seq_num):
-            await skvbc.write_known_kv()
+            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.5)
```
teoparvanov: Same here, why not simply say:
```python
            await tracker.run_concurrent_ops(num_ops=self.evaluation_period_seq_num, write_weight=0.5)
```

or something in that spirit...

tests/test_skvbc_fast_path.py
```diff
@@ -43,33 +43,31 @@ def setUp(self):
...
 
-        Finally we check if a known K/V has been executed.
+        Finally the decorator verifies the KV execution.
         """
         bft_network.start_all_replicas()
-        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
-
         for _ in range(10):
-            key, val = await skvbc.write_known_kv()
+            await tracker.run_concurrent_ops(num_ops=1, write_weight=0.7)
```
teoparvanov: why not remove the for loop here too ?

tests/test_skvbc_fast_path.py
```diff
@@ -78,60 +76,55 @@ def setUp(self):
...
 
-        Finally we check if a known K/V has been executed.
+        Finally the decorator verifies the KV execution.
         """
         bft_network.start_all_replicas()
-        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
 
-        for _ in range(10):
-            await skvbc.write_known_kv()
+        write_weight = 1
```
teoparvanov: how about:
numops = 20
write_weight = 0.5 ?

tests/test_skvbc_fast_path.py
```diff
@@ -78,60 +76,55 @@ def setUp(self):
...
             replica=random.choice(unstable_replicas))
 
-        for _ in range(10):
-            key, val = await skvbc.write_known_kv()
-
-        await bft_network.assert_slow_path_prevalent(as_of_seq_num=10)
+        await tracker.run_concurrent_ops(num_ops=numops, write_weight=write_weight)
 
-        await skvbc.assert_kv_write_executed(key, val)
+        await bft_network.assert_slow_path_prevalent(as_of_seq_num=numops+1)
```
teoparvanov: it can be more general here: `numops * write_weight + 1`

tests/test_skvbc_fast_path.py
Liorgal28: Just inattention because this test is the simplest and always passed so I forgot to handle it

tests/test_skvbc_fast_path.py
```diff
@@ -78,60 +81,56 @@ def setUp(self):
...
         """
         bft_network.start_all_replicas()
-        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
 
-        for _ in range(10):
-            await skvbc.write_known_kv()
+        write_weight = 0.5
+        numops = 20
+
+        read_count, write_count = await tracker.run_concurrent_ops(num_ops=numops, write_weight=write_weight)
```
teoparvanov: `_, fast_path_writes = ...`

teoparvanov: also, the line is too long, please re-format

tests/test_skvbc_fast_path.py
```diff
@@ -78,60 +81,56 @@ def setUp(self):
...
         for _ in range(bft_network.config.c):
             replica_to_stop = random.choice(unstable_replicas)
             bft_network.stop_replica(replica_to_stop)
-
+        write_weight = 0.5
         # make sure we first downgrade to the slow path...
-        for _ in range(self.evaluation_period_seq_num):
-            await skvbc.write_known_kv()
+
+        _, write_count = await tracker.run_concurrent_ops(num_ops=self.evaluation_period_seq_num-1, write_weight=1)
```
teoparvanov: write_count -> slow_path_writes

teoparvanov: and also - line too long :)

tests/test_skvbc_fast_path.py
```diff
@@ -43,33 +46,33 @@ def setUp(self):
...
     @with_bft_network(start_replica_cmd)
-    async def test_fast_path_read_your_write(self, bft_network):
+    @verify_linearizability
+    async def test_fast_path_read_your_write(self, bft_network, tracker):
         """
         This test aims to check that the fast commit path is prevalent
         in the normal, synchronous case (no failed replicas, no network partitioning).
 
-        First we write a series of known K/V entries.
+        First we write a series of K/V entries and tracked them using the tracker from the decorator.
```
teoparvanov: "tracked them" -> "track them"

tests/test_skvbc_fast_path.py
```diff
@@ -78,60 +81,56 @@ def setUp(self):
...
+    @verify_linearizability
+    async def test_fast_path_resilience_to_crashes(self, bft_network, tracker):
         """
         In this test we check the fast path's resilience when up to "c" nodes fail.
 
         As a first step, we bring down no more than c replicas,
         triggering initially the slow path.
 
-        Then we write a series of known K/V entries, making sure
+        Then we write a series of K/V entries and tracked them using the tracker from the decorator, making sure
```
teoparvanov: "tracked them" -> "track them"

PR #362, opened 2020-01-21
SparseMerkle: Add BatchedInternalNode::remove

andrewjstone:
Support for removing a key from a `BatchedInternalNode` was added.
Removal returns one of four variants depending upon the structure of the
BatchedInternalNode:

 1. `RemoveComplete` if this leaf was deleted but other children prevent
 the BatchedInternalNode from being removed
 2. `NotFound` if the key does not exist in the BatchedInternalNode and
 there is not an InternalChild implying it lives in another
 BatchedInternalNode further down the tree.
 3. `RemoveBatchedInternalNode` if there were only 0 or 1 LeafChildren
 remaining such that the caller should remove the node and move the
 LeafChild into its parent.
 4. `Descend` if there is an InternalChild at depth 4 in this node where
 the key would have been found. This implies the key lives further down
 the tree and the caller should retrieve the next BatchedInternalNode.

Some helper methods were also added.

A comprehensive set of tests shows how the remove operates on the
BatchedInternalNode.

### DIFF
--- a/storage/include/sparse_merkle/internal_node.h
+++ b/storage/include/sparse_merkle/internal_node.h
@@ -1,5 +1,4 @@
-// Concord
-//
+// Concord //
 // Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
 //
 // This product is licensed to you under the Apache 2.0 license (the "License").
@@ -30,7 +29,7 @@ A partially full BatchedInternalNode may look like the following:
 ```
 Level 4               ---------root-------
                       |                  |
-                  LeafChild       InternalChild
+Level 3           LeafChild       InternalChild
                                   /            \
 Level 2                     LeafChild    InternalChild
                                                   \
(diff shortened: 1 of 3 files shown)


storage/include/sparse_merkle/internal_node.h
```diff
@@ -1,5 +1,4 @@
-// Concord
-//
+// Concord //
```
dartdart26: [Minor]: Maybe a missing line?

storage/test/sparse_merkle/internal_node_test.cpp
```diff
@@ -16,6 +16,19 @@
...
+// bit 1, the second bit, etc...
+Hash flipMSBit(size_t bit, const Hash& hash) {
+  Assert(bit < 8);
+  // Flip the first bit so that key1_hash becomes a sibling of key2_hash
+  std::array<uint8_t, Hash::SIZE_IN_BYTES> flipped;
+  std::copy(hash.data(), hash.data() + hash.size(), flipped.begin());
+  flipped[0] ^= (1 << (7 - bit));
+  return Hash(flipped);
+}
+
```
dartdart26: [Minor]: Maybe this can be named differently as it is not always the MS bit. Something like`flipByte0Bit()`?

andrewjstone:
I agree with all the requested changes. I'll add them in the next commit.

PR #366, opened 2020-01-23
Logging improvements for debugging

teoparvanov:
- adding a timestamp for each log statement
- logging the replica ID when processing incoming messages
- fixed the log level of "NUM OF ELEMENTS IN BLOCK = " (ERROR -> INFO)

Sample log messsage:
```
[Node 2] [139645684168448] INFO |concord|void bftEngine::impl::ReplicaImp::onMessage(bftEngine::impl::ClientRequestMsg *)|Node 2 received ClientRequestMsg (clientId=4 reqSeqNum=6626155588176838656, readOnly=0) from Node 4|
```

### DIFF
--- a/bftengine/src/bftengine/IncomingMsgsStorageImp.cpp
+++ b/bftengine/src/bftengine/IncomingMsgsStorageImp.cpp
@@ -17,14 +17,20 @@
 #include "Logger.hpp"
 #include <future>
 
+#ifdef USE_LOG4CPP
+#include <log4cplus/mdc.h>
+#endif
+
 using std::queue;
 using namespace std::chrono;
 
 namespace bftEngine::impl {
 
 IncomingMsgsStorageImp::IncomingMsgsStorageImp(std::shared_ptr<MsgHandlersRegistrator>& msgHandlersPtr,
-                                               std::chrono::milliseconds msgWaitTimeout)
+                                               std::chrono::milliseconds msgWaitTimeout,
+                                               uint16_t replicaId)
     : IncomingMsgsStorage(), msgHandlers_(msgHandlersPtr), msgWaitTimeout_(msgWaitTimeout) {
+  replicaId_ = replicaId;
   ptrProtectedQueueForExternalMessages_ = new queue<std::unique_ptr<MessageBase>>();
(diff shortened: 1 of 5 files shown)


bftengine/src/bftengine/IncomingMsgsStorageImp.cpp
```diff
@@ -42,7 +48,14 @@ IncomingMsgsStorageImp::~IncomingMsgsStorageImp() {
 void IncomingMsgsStorageImp::start() {
   if (!dispatcherThread_.joinable()) {
     std::future<void> futureObj = signalStarted_.get_future();
-    dispatcherThread_ = std::thread([=] { dispatchMessages(signalStarted_); });
+    dispatcherThread_ = std::thread([=] {
+#ifdef USE_LOG4CPP
```
andrewjstone: I = dislike the use of conditional code for logging here. If we start doing this, it will litter the code everywhere. We've already removed it in a bunch of places prior to this. 

While I'm fine with it for now, as it definitely helps debugging in the short term, we should really talk about cleaning up logging altogether.  My suggestion is just to force the use of LOG4CPP. We bend over backwards to prevent having a dependency for logging, but it essentially means we have no straightforward way to even configure our log levels for tests. 

If we decide to maintain an abstraction, we should make it much more advanced, provide an interface, and allow the user of the library to pass in the implementation of the logger themselves. 
We should really consider a

teoparvanov: Hi Andrew, thanks for bringing this up! I agree the current situation is not great, and suggest the following:

1) **[Short term]** as a first step, we should always use LOG4CPP for CI & local testing. This way we won't be stuck because our logging isn't flexible enough.

2) **[Long term]** In terms of general logging approach, I don't think that forcing the use of LOG4CPP is a good idea. Concord BFT is a library, so it should provide a thin interface where different logging frameworks can be plugged. It seems like a bigger and more complex effort than step 1.

Both of those items are added to our backlog. The first one can be executed as part of Concord-BFT testing. The second one seems like typical Concord-BFT tech debt, which should be prioritized with respect to the needs of Concord BFT library clients.

Best regards,
Teo

PR #368, opened 2020-01-24
Add linearizabilty checks into persistence tests

Liorgal28:
Add linearizabilty checks into persistence tests, add tracking option using verify_linearizabilty decorator and with this tracker we track every write and read

1)add verify_linearizability decorator for each test.
2)replace regular KV writing to tracked KV writing.
3)check KV writing successful execution at the end of the decorator using tracker verify.
4)create a linearizability tracked known kv writing function.
5)create a linearizability tracked known kv reading function.
4)create a linearizability tracked prime_for_state_transfer function.

### DIFF
--- a/tests/test_skvbc_persistence.py
+++ b/tests/test_skvbc_persistence.py
@@ -163,7 +163,7 @@ class SkvbcPersistenceTest(unittest.TestCase):
         # Perform a put/get transaction pair to ensure we can read newly
         # written data after state transfer.
 
-        # await skvbc.read_your_writes(self)
+        await tracker.tracked_read_your_writes(self)
 
     @with_trio
     @with_bft_network(start_replica_cmd)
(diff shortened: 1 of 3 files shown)


tests/util/skvbc.py
```diff
@@ -20,8 +20,9 @@
 
 WriteReply = namedtuple('WriteReply', ['success', 'last_block_id'])
 
+
 class SimpleKVBCProtocol:
-    KV_LEN = 21 ## SimpleKVBC requies fixed size keys and values right now
+    KV_LEN = 21  ## SimpleKVBC requies fixed size keys and values right now
```
teoparvanov: I see a bunch of pure formatting/spacing changes like this one. Can you please clean-up the PR by removing all of them, as they create unnecessary noise?

Thanks!

tests/test_skvbc_persistence.py
```diff
@@ -211,42 +206,44 @@ class SkvbcPersistenceTest(unittest.TestCase):
...
+                # We never made it to fetching state. Are we done?
+                try:
+                    await bft_network.wait_for_state_transfer_to_stop(
+                        up_to_date_node, stale_node)
+                except trio.TooSlowError:
+                    self.fail("State transfer did not complete, " +
+                              "but we are not fetching either!")
+            finally:
+                print(f'Stopping replica {stale_node}')
+                bft_network.stop_replica(stale_node)
```
teoparvanov: Here it's unclear what has changed. Is this formatting or have you added new code?

tests/util/skvbc_history_tracker.py
```diff
@@ -901,4 +901,91 @@ def writeset(self, max_size):
                     await self.send_tracked_read(client, max_size)
             except:
                 pass
-            await trio.sleep(.1)
\ No newline at end of file
+            await trio.sleep(.1)
+
+    async def write_and_track_known_kv(self, kv, client):
```
teoparvanov: I really like those tracked read/write utilities, but there's some duplication already with the skvbc.py module. 

I suggest to keep it like this for now. Later on we might consider injecting an optional SkvbcTracker into the SKVBC Protocol and extracting the common logic.

I'll add a backlog entry for this.

Liorgal28: Yes I know, I did kind of identical functions but with using the tracking capability and still preserved the existing functions because I didn't want to delete without getting permission.

tests/util/skvbc_history_tracker.py
```diff
@@ -901,4 +901,91 @@ def writeset(self, max_size):
...
+        # Perform an unconditional KV put.
+        # Ensure keys aren't identical
+        kv = [(keys[0], self.skvbc.random_value()),
+              (keys[1], self.skvbc.random_value())]
+
+        #reply = await client.write(self.write_req([], kv, 0))
+        #reply = self.parse_reply(reply)
+        reply = await self.write_and_track_known_kv(kv,client)
+        test_class.assertTrue(reply.success)
+        test_class.assertEqual(last_block + 1, reply.last_block_id)
```
teoparvanov: Let's remove the dependency on test_class by saying:
```python
assert reply.success
assert last_block + 1 == reply.last_block_id
```

PR #372, opened 2020-01-28
Cleanup sparse merkle tree insert code

andrewjstone:
Please see individual commits for descriptions.

### DIFF
--- a/storage/include/sparse_merkle/base_types.h
+++ b/storage/include/sparse_merkle/base_types.h
@@ -244,6 +244,17 @@ class NibblePath {
     num_nibbles_ += 1;
   }
 
+  // Return the last nibble off the path
+  Nibble back() {
+    Assert(!empty());
+    if (num_nibbles_ % 2 == 0) {
+      // We have a complete byte at the end of path_. Remove the lower nibble.
+      return path_.back() & 0x0F;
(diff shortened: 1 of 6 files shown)


storage/include/sparse_merkle/base_types.h
```diff
@@ -244,6 +244,17 @@ class NibblePath {
     num_nibbles_ += 1;
   }
 
+  // Return the last nibble off the path
+  Nibble back() {
```
dartdart26: Can we make it `const`?

storage/include/sparse_merkle/internal_node.h
```diff
@@ -202,7 +203,10 @@ class BatchedInternalNode {
 
   // A remove has failed, because the the leaf is further down the tree.
   // Instruct the caller to try the next node.
-  struct Descend {};
+  struct Descend {
+    // The version of the next BatchedInternalNode further down the tree.
+    Version version;
```
dartdart26: [Minor]: Maybe we can name the version so that it is a bit clearer, e.g. `next_batched_node_version` or similar.

andrewjstone:
Thanks @dartdart26 ! I'll make the changes in the next commit.

PR #374, opened 2020-01-28
Linearizability of many concurrent operations after view change

teoparvanov:
In addition of a simple "read-your-write" check after view change has completed in various scenarios, we also want to make sure linearizability is preserved.
We check this with a load of 100 concurrent operations (random reads/writes) which are then checked by the @verify_linearizability decorator.


andrewjstone:
I'm not sure how much value this has without any induced failures. It doesn't hurt and I'll happily approve it, but really what we want is to run view change during concurrent operations and verify those operations later. Some ops should run before view change, some after. The tracker and checker are sophisticated enough to handle lost messages. That is when they become especially useful.

PR #375, opened 2020-01-28
Fix Sliver's constructor that moves a std::string

dartdart26:
Sliver's constructor that is supposed to move a std::string into the
sliver has the following signature: Sliver(const std::string&&). This
PR fixes the mentioned constructor by taking a non-const std::string&&,
i.e. Sliver(std::string&&). This change enables proper move semantics.

Moreover, add a move constructor to StringBuf so that an unnecessary
move operation is avoided when using aggregate initialization.

Add a simple test to show that the move constructor works as expected.

Fix a memory leak in the simple_copy test in sliver_test.cpp to make
valgrind happy.

### DIFF
--- a/util/test/sliver_test.cpp
+++ b/util/test/sliver_test.cpp
@@ -9,7 +9,10 @@
 
 #include "sliver.hpp"
 #include "gtest/gtest.h"
+
 #include <iostream>
+#include <string>
+
 #include <string.h>
 
 using namespace std;
@@ -66,6 +69,8 @@ TEST(sliver_test, simple_copy) {
   auto actual = Sliver::copy(expected, test_size);
   ASSERT_TRUE(is_match(expected, test_size, actual));
   ASSERT_NE(expected, actual.data());
+
+  delete[] expected;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


dartdart26:
Added a check in the test to verify that the string is actually being moved.

andrewjstone:
> Added a check in the test to verify that the string is actually being moved.

Awesome. I was going to recommend checking by looking at the moved from `str`, but realized that the state of the moved from object is technically undefined, so just approved. This way is better :) You are a mind reader.

util/test/sliver_test.cpp
```diff
@@ -171,6 +176,30 @@ TEST(sliver_test, copying) {
...
+  const auto str_data_ptr = str.c_str();
+  const auto sliver2 = Sliver{std::move(str)};
+
+  // Make sure that both slivers contain the same data.
+  ASSERT_TRUE(is_match(expected, test_size, sliver1));
+  ASSERT_TRUE(is_match(expected, test_size, sliver2));
+  ASSERT_TRUE(sliver1 == sliver2);
+
+  // Make sure that the string has actually been moved without copying.
+  ASSERT_EQ(str_data_ptr, sliver2.data());
```
teoparvanov: Very cool, thanks!

PR #377, opened 2020-01-29
Rename system testing framewok to "Apollo"

teoparvanov:
This PR includes the following changes:
* renaming the 'tests' folder to 'apollo'
* updating README.md's where appropriate


toly-kournik:
Why Apollo? I think we have to name more after female greek myths characters.
What about Calypso or Penelope?

teoparvanov:
@toly-kournik, we decided together. I believe you were there during the meeting when I presented the testing framework update?

Also, Apollo is Hermes's brother - so it also has a symbolic meaning of alignment with the product :-)

toly-kournik:
I haven't participated the meeting. I would never approve this. Do you know how many siblings (Zeus children) Hermes has?

teoparvanov:
'Never' is a strong word I believe ... :)

toly-kournik:
Apollo: Hermes we have a problem!

teoparvanov:
Actually, not only is Hermes Apollo's little brother, but he also apparently stole Apollo's sheep, right after he was born.

because of ... reasons?

https://www.theoi.com/Olympios/Hermes.html

andrewjstone:
Hmm. I thought we were only going to call the python testing framework Apollo, not all our tests. This change puts things like SimpleTest, which is totally unrelated to the framework under the Apollo name. I would prefer keeping the top-level tests directory and just renaming the appropriate python code and any C++ related helper classes as needed.

andrewjstone:
> Hmm. I thought we were only going to call the python testing framework Apollo, not all our tests. This change puts things like SimpleTest, which is totally unrelated to the framework under the Apollo name. I would prefer keeping the top-level tests directory and just renaming the appropriate python code and any C++ related helper classes as needed.

I guess since our python tests lived in the top level, this makes some sense. I would prefer it if apollo was separated out into its own directory though. It's logically separate from a lot of the stuff under tests and should be physically separated IMO.

teoparvanov:
Hi Andrew, thanks for commenting! When you mention it, maybe it makes sense to keep the 'tests' folder as is, and separate-out the Python framework under 'tests/apollo'. I'll post a separate PR for this, so that we can compare and decide.

PR #380, opened 2020-01-30
Introducing Apollo - Concord BFT's system testing framework

teoparvanov:
This PR includes the following changes:
* moving Python tests and utilities under tests/apollo
* updating README.md's where appropriate


teoparvanov:
@andrewjstone, how does this look to you? I believe it's indeed cleaner than the approach I had suggested initially.

PR #383, opened 2020-01-31
Don't remove root BatchedInternalNode with a child

andrewjstone:
The previous behavior of BatchedInternalNodeRemove always returned
`RemoveBatchedInternalNode` when there was one remaining LeafChild, as
it was capable of being promoted to the parent node. However, when the
LeafChild lives in the root BatchedInternalNode, it cannot be moved up
any further. In this case we should return `RemoveComplete`, which this
commit does.

Additionally, do some minor cleanup related to a review of a prior
commit done.

### DIFF
--- a/storage/test/sparse_merkle/internal_node_test.cpp
+++ b/storage/test/sparse_merkle/internal_node_test.cpp
@@ -429,6 +429,9 @@ TEST(remove_tests, remove_a_single_leaf) {
 /*
 // Add 2 LeafChildren to a BatchedInternalNode and then remove one of them.
 //
+// Since this is a remove from the second node, the BatchedInternalNode can be
+// removed when a peer moves up to the root BatchedInternalNode.
+//
 // The logical tree inside the BatchedInternalNode looks like the following
 // before remove is called:
 //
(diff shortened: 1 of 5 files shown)


storage/test/sparse_merkle/internal_node_test.cpp
```diff
@@ -480,6 +485,64 @@ TEST(remove_tests, remove_a_leaf_with_a_peer) {
...
+  // There should be one root and two leaves
+  ASSERT_EQ(3, node.numChildren());
+  ASSERT_EQ(1, node.numInternalChildren());
+  ASSERT_EQ(2, node.numLeafChildren());
+
+  // Deleting key1 should return RemoveComplete, since the peer cannot be promoted above the root node.
+  auto result = node.remove(key1_hash, depth, Version(3));
+  ASSERT_TRUE(std::holds_alternative<BatchedInternalNode::RemoveComplete>(result));
+  ASSERT_EQ(node.hash(), hasher.parent(PLACEHOLDER_HASH, value2_hash));
+}
```
dartdart26: Maybe we can verify the children count after the `remove` call, something like:
```
  // There should be one root and one leaf
  ASSERT_EQ(2, node.numChildren());
  ASSERT_EQ(1, node.numInternalChildren());
  ASSERT_EQ(1, node.numLeafChildren());
```

andrewjstone: Great suggestion. Also, thank you for taking the time to understand these tests :)

andrewjstone: Fixed.

PR #384, opened 2020-01-31
Abstract away blockchain storage details

dartdart26:
In order to prepare for changing the storage engine to using merkle
trees, abstract away blockchain storage details (block headers, block
entries, block keys and their respective implementations). When we
introduce merkle trees, we can more easily switch to the new
implementation.

Separate KeyManipulator to DBKeyComparator and DBKeyManipulator as
these two notions can be used independently.

Introduce the v1DirectKeyValue namespace to aid in the future switch to
merkle trees. All existing implementations reside in this namespace.

In addition, introduce layered key types and the v2MerkleTree namespace
for the future merkle tree storage implementation.

Lastly, cleanup code in terms of const, includes, using directives and
static/non-static members.

### DIFF
--- a/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
+++ b/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
@@ -14,6 +14,7 @@
 #ifndef BFTENGINE_SRC_BCSTATETRANSFER_SIMPLEBCSTATETRANSFER_HPP_
 #define BFTENGINE_SRC_BCSTATETRANSFER_SIMPLEBCSTATETRANSFER_HPP_
 
+#include <cstdint>
 #include <set>
 #include <memory>
 
@@ -40,7 +41,7 @@ namespace SimpleBlockchainStateTransfer {
 // The application/storage layer is responsible to store the digests in the
 // blocks.
 // Blocks are numbered. The first block should be block number 1.
-constexpr static uint32_t BLOCK_DIGEST_SIZE = 32;
+inline constexpr std::uint32_t BLOCK_DIGEST_SIZE = 32;
 
 // represnts a digest
 #pragma pack(push, 1)
--- a/kvbc/src/ReplicaImp.cpp
+++ b/kvbc/src/ReplicaImp.cpp
@@ -19,9 +19,8 @@
 #include "hex_tools.h"
 #include "replica_state_sync.h"
 #include "sliver.hpp"
-#include "SimpleBCStateTransfer.hpp"
-#include "blockchain/db_types.h"
 #include "blockchain/db_interfaces.h"
+#include "blockchain/block.h"
 #include "storage/db_metadata_storage.h"
 
 using bftEngine::ICommunication;
@@ -33,12 +32,11 @@ using concord::storage::IDBClient;
 using concord::storage::DBMetadataStorage;
 
 using concord::storage::blockchain::DBAdapter;
-using concord::storage::blockchain::BlockEntry;
-using concord::storage::blockchain::BlockHeader;
 using concord::storage::blockchain::BlockId;
 using concord::storage::blockchain::ILocalKeyValueStorageReadOnly;
 using concord::storage::blockchain::ILocalKeyValueStorageReadOnlyIterator;
-using concord::storage::blockchain::KeyManipulator;
+using concord::storage::blockchain::DBKeyManipulator;
+namespace block = concord::storage::blockchain::block;
 
 using concordUtils::SetOfKeyValuePairs;
 using concordUtils::Value;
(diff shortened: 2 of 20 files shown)


storage/include/blockchain/block_helper.h
```diff
@@ -0,0 +1,44 @@
...
+#include <cstdint>
+#include <utility>
+
+namespace concord {
+namespace storage {
+namespace blockchain {
+inline namespace v1_direct_key_value {
+
+// A helper class used to abstract away block storage details.
+struct BlockHelper {
```
andrewjstone: I'd prefer if we could make this class name a bit more concrete. It's just personal taste, but I don't really like names such as `*Helper` and `*Wrapper`. I don't feel they add much value, and just lead to more typing. 

I just did a quick search and it looks like we don't actually have a type called `Block`, so we could just use that as the name. Then we can also shorten the names of these static functions to things like: `Block::create` and `Block::getData`. `getParentDigest` is a nice name still :)

Alternatively, since these are all static functions we could just not wrap them in a type but instead tie them to the namespace directly.

dartdart26: I agree, using a namespace is better, doing that. I am also using the names you suggest - I think they are simple and descriptive enough.

storage/include/blockchain/db_types.h
```diff
@@ -14,26 +14,26 @@
...
 namespace storage {
 namespace blockchain {
 
+inline constexpr std::uint32_t BLOCK_DIGEST_SIZE = 32;
+
+inline namespace v1_direct_key_value {
 struct BlockHeader {
-  uint32_t numberOfElements;
-  uint32_t parentDigestLength;
-  int8_t parentDigest[bftEngine::SimpleBlockchainStateTransfer::BLOCK_DIGEST_SIZE];
```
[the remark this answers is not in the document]
dartdart26: See above comment.

bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
```diff
@@ -40,7 +42,7 @@ namespace SimpleBlockchainStateTransfer {
 // The application/storage layer is responsible to store the digests in the
 // blocks.
 // Blocks are numbered. The first block should be block number 1.
-constexpr static uint32_t BLOCK_DIGEST_SIZE = 32;
+inline constexpr auto BLOCK_DIGEST_SIZE = concord::storage::blockchain::BLOCK_DIGEST_SIZE;
```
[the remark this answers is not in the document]
dartdart26: That's a good point and I agree. I was thinking of something like a trait template that users can specialise  and specify the constant that way. Something like `std::hash()`. However, I was looking at `bftengine`, `kvbc` and `storage` directories and I am not sure yet where to put the current specialisation with 32. It seems like it should be in `kvbc`, but need to think it through.

In the meantime, I have done further refactoring and will push that first.

dartdart26:
Pushed improvements to interfaces, naming and code structure.

bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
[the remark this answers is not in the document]
dartdart26: Actually, I am reverting that part as it is the state transfer code that computes block hashes and determines the size of 32 bytes. In doing so, the dependency stays from state transfer to a specific implementation in block.h/cpp .

I am thinking this can be resolved at a later stage.

dartdart26:
Reverted the `BLOCK_DIGEST_SIZE` change - definition stays in state transfer code as it is the one that computes block hashes and, therefore, determines the size of 32 bytes.

kvbc/src/ReplicaImp.cpp
```diff
@@ -145,7 +143,7 @@ Status ReplicaImp::getBlockData(BlockId blockId, SetOfKeyValuePairs &outBlockDat
     return Status::NotFound("todo");
   }
 
-  outBlockData = fetchBlockData(block);
+  outBlockData = block::getData(block);
```
andrewjstone: Excellent :)

kvbc/src/ReplicaImp.cpp
```diff
@@ -674,9 +579,9 @@ bool ReplicaImp::BlockchainAppState::getPrevDigestFromBlock(uint64_t blockId, St
     exit(1);
   }
 
-  const BlockHeader *bh = reinterpret_cast<const BlockHeader *>(result.data());
+  auto parentDigest = block::getParentDigest(result);
```
andrewjstone: Eventually maybe we can start using proper C++ types like the SHA3 hash in util. Then we won't need memcpy, etc... Definitely do not change that in this PR though. Just thinking out loud.

andrewjstone: Also, nice cleanup :)

storage/include/blockchain/block.h
```diff
@@ -0,0 +1,69 @@
...
+                            const void *parentDigest);
+
+// Returns the block data in the form of a set of key/value pairs.
+concordUtils::SetOfKeyValuePairs getData(const concordUtils::Sliver &block);
+
+// Returns the parent digest of size BLOCK_DIGEST_SIZE bytes.
+const void *getParentDigest(const concordUtils::Sliver &block);
+
+// Block structure is an implementation detail. External users should not rely on it.
+namespace detail {
```
andrewjstone: nice!

dartdart26:
> Very cool, thanks :) Sorry I wasn't able to look into this earlier.
> 
> PS: just for my understanding, how is the switch between the v1 and v2 namespaces going to work?
> Are we going to use directives so that, depending on a build option, one or the other becomes "inline"?

Yes, that is an option we can use.

dartdart26:
Thank you @andrewjstone and @teoparvanov for the reviews!

PR #394, opened 2020-02-10
Support for object store

salieri11:
This PR introduce 2 major add ons to the library: Application State implementation for Read Only replica and Objecst store support, which can be used to replicate the data from the RO replica.
The Object Store implementation uses S3 protocol.

Details:
The support is implemented using libs3 library. The library provides basic S3 protocol features and uses libcurl to issue HTTP requests. This new client,ObjectStoreClient which implements IDBClient interface and can be easily changed to local RockDB client (e.g. for tests).

The RoRApplicationState class implements IAppState interface, which was extended to support wait() function. This function may be useful when any of the IAppState IO operations return false (e.g. when the remote Object Store is unavailable) and the State Transfer module (e.g. of the RO replica) needs to wait before pushing new blocks. The call will be blocked until the underlying storage (e.g. S3) will become available.

Tests:
There are gtest suites for both modules, which in CI will run using local RocksDB as an object store.

E2E test:
There is a new script for manual run, readOnlyReplicaTest.sh, which start 4 regular and 1 RO replica. Further development will include automatic tests using our Apollo framework.

CMAKE flags:
There are 2 new cmake flags: BUILD_OBJECTSTORE_TEST when set, the real s3 Gtest suite will be build (assumes working S3 instance, the test needs to have parameters changed such as machine IP, access key and secret key)
USE_OBJECTSTORE_FOR_ROR_TEST when set, the ror_app_state Gtest suite will use real object store and the default (unset) mode will use local RocksDB (CI is running in this mode)

### DIFF
--- a/kvbc/src/ReplicaImp.cpp
+++ b/kvbc/src/ReplicaImp.cpp
@@ -8,7 +8,6 @@
 #include <string.h>
 #include <sys/param.h>
 #include <unistd.h>
-
 #include "ReplicaImp.h"
 #include <inttypes.h>
 #include <cassert>
@@ -37,6 +36,7 @@ using concord::storage::blockchain::ILocalKeyValueStorageReadOnly;
 using concord::storage::blockchain::ILocalKeyValueStorageReadOnlyIterator;
 using concord::storage::blockchain::DBKeyManipulator;
 namespace block = concord::storage::blockchain::block;
+namespace st = bftEngine::SimpleBlockchainStateTransfer;
 
 using concordUtils::SetOfKeyValuePairs;
 using concordUtils::Value;
@@ -83,13 +83,15 @@ void ReplicaImp::createReplicaAndSyncState() {
       &m_replicaConfig, m_cmdHandler, m_stateTransfer, m_ptrComm, m_metadataStorage);
   if (!isNewStorage && !m_stateTransfer->isCollectingState()) {
     uint64_t removedBlocksNum = replicaStateSync_->execute(
-        logger, *m_bcDbAdapter, m_appState->m_lastReachableBlock, m_replicaPtr->getLastExecutedSequenceNum());
+        logger, *m_bcDbAdapter, m_appState->getLastReachableBlockNum(), m_replicaPtr->getLastExecutedSequenceNum());
     m_lastBlock -= removedBlocksNum;
-    m_appState->m_lastReachableBlock -= removedBlocksNum;
+    auto t = std::dynamic_pointer_cast<BlockchainAppState>(m_appState);
+    assert(t);
+    t->m_lastReachableBlock -= removedBlocksNum;
     LOG_INFO(logger,
              "createReplicaAndSyncState: removedBlocksNum = "
                  << removedBlocksNum << ", new m_lastBlock = " << m_lastBlock
-                 << ", new m_lastReachableBlock = " << m_appState->m_lastReachableBlock);
+                 << ", new m_lastReachableBlock = " << t->m_lastReachableBlock);
   }
 }
 
@@ -187,6 +189,16 @@ Status ReplicaImp::addBlock(const SetOfKeyValuePairs &updates, BlockId &outBlock
 
 void ReplicaImp::set_command_handler(ICommandsHandler *handler) { m_cmdHandler = handler; }
 
+void ReplicaImp::set_app_state(std::shared_ptr<bftEngine::SimpleBlockchainStateTransfer::IAppState> appState) {
+  if(!appState)
+    m_appState = std::make_shared<BlockchainAppState>(shared_from_this());
+  else
+    m_appState = appState;
+    
+  m_stateTransfer = bftEngine::SimpleBlockchainStateTransfer::create(
+      state_transfer_config_, m_appState.get(), m_bcDbAdapter->getDb(), aggregator_);
+}
+
 ReplicaImp::ReplicaImp(ICommunication *comm,
                        bftEngine::ReplicaConfig &replicaConfig,
                        DBAdapter *dbAdapter,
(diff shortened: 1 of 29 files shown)


tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -42,7 +41,7 @@ std::unique_ptr<TestSetup> TestSetup::ParseArgs(int argc, char** argv) {
   string idStr;
 
   int o = 0;
-  while ((o = getopt(argc, argv, "r:i:k:n:s:v:a:p")) != EOF) {
+  while ((o = getopt(argc, argv, "r:i:k:n:s:v:a:po:")) != EOF) {
```
teoparvanov: shouldn't this be a:p:o ?

storage/include/object_store/object_store_client.hpp
```diff
@@ -0,0 +1,82 @@
...
+  virtual IDBClient::IDBClientIterator* getIterator() const override;
+  virtual Status freeIterator(IDBClientIterator* _iter) const override;
+  virtual ITransaction* beginTransaction() override;
+
+  void wait_for_storage();
+  Status object_exists(const Sliver& key);
+  Status test_bucket();
+
+ private:
+  std::shared_ptr<IDBClient> pImpl_ = nullptr;
```
teoparvanov: why not unique_ptr ?

storage/include/object_store/object_store_client.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include "storage/db_interface.h"
+#include <chrono>
+
+namespace concord {
+namespace storage {
+
+struct S3_StoreConfig {
+  std::string bucketName;  // assuminh pre configured, no need to create
+  std::string url;         // from the customer
+  std::string protocol;    // currently tested with HTTP
```
teoparvanov: nit - the comment refers to an implementation detail, so maybe it shouldn't be in the header

kvbc/src/ReplicaImp.cpp
```diff
@@ -83,13 +83,15 @@ void ReplicaImp::createReplicaAndSyncState() {
       &m_replicaConfig, m_cmdHandler, m_stateTransfer, m_ptrComm, m_metadataStorage);
   if (!isNewStorage && !m_stateTransfer->isCollectingState()) {
     uint64_t removedBlocksNum = replicaStateSync_->execute(
-        logger, *m_bcDbAdapter, m_appState->m_lastReachableBlock, m_replicaPtr->getLastExecutedSequenceNum());
+        logger, *m_bcDbAdapter, m_appState->getLastReachableBlockNum(), m_replicaPtr->getLastExecutedSequenceNum());
     m_lastBlock -= removedBlocksNum;
-    m_appState->m_lastReachableBlock -= removedBlocksNum;
+    auto t = std::dynamic_pointer_cast<BlockchainAppState>(m_appState);
```
teoparvanov: what if the m_appState is a `RoRAppState` ?

kvbc/src/ReplicaImp.cpp
```diff
@@ -232,7 +238,8 @@ ReplicaImp::~ReplicaImp() {
 
 Status ReplicaImp::addBlockInternal(const SetOfKeyValuePairs &updates, BlockId &outBlockId) {
   m_lastBlock++;
-  m_appState->m_lastReachableBlock++;
+  auto pt = downcast_appstate();
```
teoparvanov: If you have an RoRAppState, this downcast would fail, right?

kvbc/src/ReplicaImp.cpp
salieri11: This code is executed only on tjhe non-ro replica, see line 62 - only for non-ro replica this function will be called

kvbc/src/ReplicaImp.cpp
salieri11: This function is called only with non-ro replica, but you are right -  I will think of how to make bit more clear and safe.

storage/include/object_store/object_store_client.hpp
salieri11: No special reason. May be converted

tests/simpleKVBC/TesterReplica/setup.cpp
salieri11: as per get getopt() specs, the option should have : after it only if it accepts value and here "p" is boolean option

storage/include/object_store/object_store_client.hpp
salieri11: yep. will fix

storage/src/object_store/ecs_s3_client.hpp
```diff
@@ -0,0 +1,412 @@
...
+  void set_bucket_name(string name) { config_.bucketName = name; }
+
+  /**
+   * @brief Initializing underlying libs3. The S3_initialize function must be
+   * called exactly once and only from 1 thread
+   *
+   * @param readOnly
+   */
+  virtual void init(bool readOnly) override {
+    std::lock_guard<std::mutex> g(initLock_);
```
andrewjstone: If this can be created and initialized from multiple threads, shouldn't some of the internal state also be protected? When would multiple threads try to initialize an ObjectStoreClient but not use it? Or is this because the rest of the functionality is already thread safe in the libS3 library?

storage/src/object_store/ecs_s3_client.hpp
```diff
@@ -0,0 +1,412 @@
...
+
+  Status object_exists(const Sliver& key) {
+    using namespace std::placeholders;
+    std::function<Status(const Sliver&)> f = std::bind(&EcsS3ClientImpl::object_exists_internal, this, _1);
+    Status res = Status::OK();
+    do_with_retry(res, f, key);
+    return res;
+  }
+
+  void wait_for_storage() {
```
andrewjstone: Just a note for the architectural structure of the rest of the feature: The caller of this function should live in a separate thread from the main thread (replica thread). Otherwise we won't be able to get metrics out of that thread or inspect it in any way, since this will block in endless retry.

storage/src/object_store/ecs_s3_client.hpp
```diff
@@ -0,0 +1,412 @@
...
+                                     << ", msg: " << rData.errorMessage);
+      if (rData.status == S3Status::S3StatusHttpErrorNotFound || rData.status == S3Status::S3StatusErrorNoSuchBucket ||
+          rData.status == S3Status::S3StatusErrorNoSuchKey)
+        return Status::NotFound("Status: " + to_string(rData.status) + "msg: " + rData.errorMessage);
+
+      return Status::GeneralError("Status: " + to_string(rData.status) + "msg: " + rData.errorMessage);
+    }
+  }
+
+  Status test_bucket_internal() {
```
andrewjstone: This should all live in a separate test wrapper class.

storage/src/object_store/ecs_s3_client.hpp
```diff
@@ -0,0 +1,412 @@
...
+
+    void Write(const char* _data, int& dataSize) {
+      CheckAndResize(dataSize);
+      memcpy(data + readLength, _data, dataSize);
+      readLength += dataSize;
+    }
+
+    ~GetObjectResponseData() { delete[] data; }
+
+    char* data = nullptr;
```
andrewjstone: Can we please use vectors for data storage instead of raw pointers to make this exception safe and prevent leaks, etc...

storage/src/object_store/ecs_s3_client.hpp
```diff
@@ -0,0 +1,412 @@
...
+
+  /**
+   * @brief Represents intermidiate data for get operation and is passed to
+   * the lambda callback.
+   *
+   */
+  struct GetObjectResponseData : public ResponseData {
+    GetObjectResponseData(size_t linitialLength) : data(new char[linitialLength]), dataLength(linitialLength) {}
+
+    void CheckAndResize(int& nextReadLength) {
```
andrewjstone: If we use vectors, the resize will be done automatically and this will not be needed.

storage/include/object_store/object_store_client.hpp
```diff
@@ -0,0 +1,80 @@
...
+namespace concord {
+namespace storage {
+
+struct S3_StoreConfig {
+  std::string bucketName;     // assuminh pre configured, no need to create
+  std::string url;            // from the customer
+  std::string protocol;       // currently tested with HTTP
+  std::string secretKey;      // from the customer
+  std::string accessKey;      // from the customer
+  std::uint32_t maxWaitTime;  // in milliseconds
```
andrewjstone: Ideally this would be a `std::chrono::duration` for type safety.

kvbc/src/ReplicaImp.cpp
andrewjstone: As mentioned in a prior comment: This downcast would not be needed if there was a separate ReplicaImp for the read only replica.

kvbc/src/ror_app_state.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+using namespace concordUtils;
+using namespace bftEngine::SimpleBlockchainStateTransfer;
+using namespace concord::storage;
+
+class RoRAppStateImpl : public IAppState {
+ public:
+  RoRAppStateImpl(std::shared_ptr<concord::storage::blockchain::DBKeyManipulator> keyManipulator,
+                  std::shared_ptr<concord::storage::IDBClient> localClient,
+                  std::shared_ptr<concord::storage::IDBClient> objectStoreClient)
```
andrewjstone: If you follow the suggestion in #406, you would pass in a std::shared_ptr<concord::storage::IObjectStoreClient> here.

salieri11:
This work has been moved to https://github.com/vmware/concord-bft/pull/444, closing this PR

PR #395, opened 2020-02-11
Add basic support for correlation-id in concord-bft

yontyon:
# Description
We add correlation id for the concord-bft client message such that a client is able to pass correlation ID to the bft engine.
## Detailed Description
Most of concord-bft flows handle a batch of size > 0 of requests. 
Thus, adding correlation ID is done only where the flow handle a single client request (i.e., upon receiving messages and upon executing them).

# More changes
* Formatting the code with clang-format-7
* Adding clang-format-7 to travis prebuild script
* Fixing a little mistake in conan recipe for rocksdb
* Removing some ifdef statement (when using log4cplus)

### DIFF
--- a/.conan/rocksdb_pkg/conanfile.py
+++ b/.conan/rocksdb_pkg/conanfile.py
@@ -35,10 +35,11 @@ class RocksdbConan(ConanFile):
         self.options["zstd"].shared = True
 
     def source(self):
-        download("https://github.com/facebook/rocksdb/archive/v5.7.3.tar.gz", "v5.7.3.tar.gz")
-        self.run("tar -xzf v5.7.3.tar.gz")
-        shutil.move("rocksdb-5.7.3", self.name)
-        os.unlink("v5.7.3.tar.gz")
+        tools.replace_in_file("CMakeLists.txt", 'project(rocksdb)',
+          '''project(rocksdb)
+          include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+          conan_basic_setup()
+            ''')
 
     def build(self):
         autotools = AutoToolsBuildEnvironment(self)
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -128,6 +128,7 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
   const ReqId reqSeqNum = m->requestSeqNum();
   const uint8_t flags = m->flags();
 
+  MDC_CID_PUT(GL, m->getCid());
   LOG_DEBUG_F(GL,
               "Node %d received ClientRequestMsg (clientId=%d reqSeqNum=%" PRIu64 ", flags=%d) from Node %d",
               config_.replicaId,
@@ -233,6 +234,7 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
   ClientRequestMsg *first = requestsQueueOfPrimary.front();
   while (first != nullptr &&
          !clientsManager->noPendingAndRequestCanBecomePending(first->clientProxyId(), first->requestSeqNum())) {
+    MDC_CID_PUT(GL, first->getCid());
     delete first;
     requestsQueueOfPrimary.pop();
     first = (!requestsQueueOfPrimary.empty() ? requestsQueueOfPrimary.front() : nullptr);
@@ -286,6 +288,7 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
 
   ClientRequestMsg *nextRequest = requestsQueueOfPrimary.front();
   while (nextRequest != nullptr && nextRequest->size() <= pp->remainingSizeForRequests()) {
+    MDC_CID_PUT(GL, nextRequest->getCid());
     if (clientsManager->noPendingAndRequestCanBecomePending(nextRequest->clientProxyId(),
                                                             nextRequest->requestSeqNum())) {
       pp->addRequest(nextRequest->body(), nextRequest->size());
@@ -3010,6 +3013,7 @@ void ReplicaImp::executeRequestsInPrePrepareMsg(PrePrepareMsg *ppMsg, bool recov
     if (!recoverFromErrorInRequestsExecution) {
       while (reqIter.getAndGoToNext(requestBody)) {
         ClientRequestMsg req((ClientRequestMsgHeader *)requestBody);
+        MDC_CID_PUT(GL, req.getCid());
         NodeIdType clientId = req.clientProxyId();
 
         const bool validClient = isValidClient(clientId);
@@ -3052,9 +3056,12 @@ void ReplicaImp::executeRequestsInPrePrepareMsg(PrePrepareMsg *ppMsg, bool recov
     while (reqIter.getAndGoToNext(requestBody)) {
       size_t tmp = reqIdx;
       reqIdx++;
-      if (!requestSet.get(tmp)) continue;
+      if (!requestSet.get(tmp)) {
+        continue;
+      }
 
       ClientRequestMsg req((ClientRequestMsgHeader *)requestBody);
+      MDC_CID_PUT(GL, req.getCid());
       NodeIdType clientId = req.clientProxyId();
 
       uint32_t actualReplyLength = 0;
--- a/bftengine/src/bftengine/SimpleClientImp.cpp
+++ b/bftengine/src/bftengine/SimpleClientImp.cpp
@@ -217,7 +217,7 @@ int SimpleClientImp::sendRequest(uint8_t flags,
   if (isPreProcessRequired)
     reqMsg = new preprocessor::ClientPreProcessRequestMsg(clientId_, reqSeqNum, lengthOfRequest, request, msgCid);
   else
-    reqMsg = new ClientRequestMsg(clientId_, isReadOnly, reqSeqNum, lengthOfRequest, request, msgCid);
+    reqMsg = new ClientRequestMsg(clientId_, flags, reqSeqNum, lengthOfRequest, request, msgCid);
   pendingRequest_ = reqMsg;
 
   sendPendingRequest();
--- a/bftengine/src/bftengine/SimpleClientImp.cpp
+++ b/bftengine/src/bftengine/SimpleClientImp.cpp
@@ -44,7 +44,8 @@ class SimpleClientImp : public SimpleClient, public IReceiver {
                   uint64_t timeoutMilli,
                   uint32_t lengthOfReplyBuffer,
                   char* replyBuffer,
-                  uint32_t& actualReplyLength) override;
+                  uint32_t& actualReplyLength,
+                  std::string cid = "") override;
 
   int sendRequestToResetSeqNum() override;
   int sendRequestToReadLatestSeqNum(uint64_t timeoutMilli, uint64_t& outLatestReqSeqNum) override;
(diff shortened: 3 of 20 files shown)


.conan/rocksdb_pkg/conanfile.py
```diff
@@ -35,10 +35,11 @@ def configure(self):
         self.options["zstd"].shared = True
 
     def source(self):
-        download("https://github.com/facebook/rocksdb/archive/v5.7.3.tar.gz", "v5.7.3.tar.gz")
-        self.run("tar -xzf v5.7.3.tar.gz")
-        shutil.move("rocksdb-5.7.3", self.name)
-        os.unlink("v5.7.3.tar.gz")
+        tools.replace_in_file("CMakeLists.txt", 'project(rocksdb)',
+                              '''project(rocksdb)
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
```
andrewjstone: Can you please fix the indentation here?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -155,10 +157,12 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
     oss << invalidClient << ", sentFromReplicaToNonPrimary=" << sentFromReplicaToNonPrimary;
     onReportAboutInvalidMessage(m, oss.str().c_str());
     delete m;
+    MDC_CID_REMOVE(GL);
     return;
   }
 
   if (readOnly) {
+    MDC_CID_REMOVE(GL);
```
andrewjstone: Why is this line before the delete call?

bftengine/src/bftengine/SimpleClientImp.cpp
```diff
@@ -187,15 +188,16 @@ int SimpleClientImp::sendRequest(uint8_t flags,
...
   bool isReadOnly = flags & READ_ONLY_REQ;
   bool isPreProcessRequired = flags & PRE_PROCESS_REQ;
-
+  std::string msgCid = cid.empty() ? std::to_string(reqSeqNum) : cid;
   // TODO(GG): check params ...
-  LOG_DEBUG(GL,
-            "Client " << clientId_ << " - sends request " << reqSeqNum << " (isRO=" << isReadOnly
-                      << ", isPreProcess=" << isPreProcessRequired << " , request size=" << lengthOfRequest
-                      << ", retransmissionMilli=" << limitOfExpectedOperationTime_.upperLimit() << " ) ");
+  LOG_INFO(GL,
```
andrewjstone: Do we really want to log every single client request? I think we demoted this on purpose.

bftengine/src/bftengine/SimpleClientImp.cpp
```diff
@@ -213,9 +215,9 @@ int SimpleClientImp::sendRequest(uint8_t flags,
 
   ClientRequestMsg* reqMsg;
   if (isPreProcessRequired)
-    reqMsg = new preprocessor::ClientPreProcessRequestMsg(clientId_, reqSeqNum, lengthOfRequest, request);
+    reqMsg = new preprocessor::ClientPreProcessRequestMsg(clientId_, reqSeqNum, lengthOfRequest, request, msgCid);
   else
-    reqMsg = new ClientRequestMsg(clientId_, flags, reqSeqNum, lengthOfRequest, request);
```
andrewjstone: This looks like a reversion of a prior change made for pre-execution.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -32,12 +32,17 @@ uint32_t getRequestSizeTemp(const char* request)  // TODO(GG): change - TBD
...
+                                   uint8_t flags,
+                                   uint64_t reqSeqNum,
+                                   uint32_t requestLength,
+                                   const char* request,
+                                   const std::string& cid)
+    : MessageBase(sender, MsgCode::ClientRequest, (sizeof(ClientRequestMsgHeader) + requestLength + cid.size() + 1)) {
   setParams(sender, reqSeqNum, requestLength, flags);
+  msgBody()->cid_length = cid.size() + 1;
   memcpy(body() + sizeof(ClientRequestMsgHeader), request, requestLength);
+  strcpy(body() + sizeof(ClientRequestMsgHeader) + requestLength, cid.c_str());
```
andrewjstone: Please use `memcpy` instead. At a minimum, use `strncpy` if we are sure the string doesn't actually contain null characters. Raw `strcpy` is a vector for buffer overflow attacks.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -75,4 +80,8 @@ void ClientRequestMsg::setParams(NodeIdType sender, ReqId reqSeqNum, uint32_t re
   setParams(reqSeqNum, requestLength, flags);
 }
 
+std::string ClientRequestMsg::getCid() {
```
andrewjstone: Are we sure that a CID is always null terminated?
This seems like a vector for attack, since the request message is sent by a client and processed in the replicas, and we can't guarantee this. I'd use a different constructor that includes the size of the CID. If the string is supposed to be well-formed, I'd also check that the last byte is indeed null, etc...

kvbc/include/KVBCInterfaces.h
```diff
@@ -65,7 +65,8 @@ class IClient {
                                     std::chrono::milliseconds timeout,
                                     uint32_t replySize,
                                     char* outReply,
-                                    uint32_t* outActualReplySize) = 0;
+                                    uint32_t* outActualReplySize,
+                                    std::string cid = "") = 0;
```
andrewjstone: Maybe take an rvalue-ref and move the string? I'm not sure if this matters, since the cid is likely small. If it's going to be copied into a message anyway, though, it should be taking by string ref.

logging/include/Logging.hpp
```diff
@@ -61,15 +74,27 @@ class Logger {
...
+  void putMdc(const std::string& key, const std::string& val) { mdc_.emplace(key, val); }
+  void removeMdc(const std::string& key) { mdc_.erase(key); }
 };
 
 class Log {
  public:
   static Logger getLogger(std::string name) { return Logger(name); }
 };
-
+class MDC {
```
andrewjstone: Can you please add comments about what and MDC is, what it does and how it is used.

In the long run, we'll likely want to make it such that we can scope the MDC using RAII, rather than having put and remove macros all over the place. We'd have to get the scope right though for that to work, though.

util/pyclient/bft_msgs.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, message):
 # Little Endian format with no padding
 # We don't include the msg type here, since we have to read it first to
 # understand what message is incoming.
-REQUEST_HEADER_FMT = "<HBQL"
+REQUEST_HEADER_FMT = "<HBQLL" # + str(CID_SIZE + 1) + 's'
```
andrewjstone: I don't understand this comment. We are only describing the header here. And the CID comes after the message anyway, so this seems to be incorrect. I suggest just removing it. It's clear from the code changes what's going on.

util/pyclient/test_msgs.py
```diff
@@ -32,8 +33,9 @@ def test_unpack_request(self):
         client_id = 4
         req_seq_num = 1
         read_only = True
-        packed = bft_msgs.pack_request(client_id, req_seq_num, read_only, msg)
-        (header, unpacked_msg) = bft_msgs.unpack_request(packed)
+        cid = str(req_seq_num)
+        packed = bft_msgs.pack_request(client_id, req_seq_num, read_only, cid, msg)
+        (header, unpacked_msg) = bft_msgs.unpack_request(packed, len(cid))
```
andrewjstone: Can you add an `assertEqual` to check that the cid is what you expect please?

logging/test/logger_test.cpp
```diff
@@ -0,0 +1,18 @@
...
+// Created by ybuchnik on 2/6/20.
+//
+#include "gtest/gtest.h"
+#include "Logger.hpp"
+using namespace concordlogger;
+TEST(logger_mdc_test, mdc_test) {
+  auto logger = Log::getLogger("test");
+  MDC::put(logger, "cid", "123456");
+  MDC::put(logger, "1", "2");
+  LOG_INFO(logger, "It works :)");
```
andrewjstone: This test isn't really automated, as it requires a user to look at the output. Can you automate it to check that the context is output as you expect? You can do this any number of ways. If you log to a file you can grep it, etc... Or you can ensure your log output goes to a string you can search.

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: This is because the `executeReadOnlyRequest` method also inserts the CID on its beginning, so we wouldn't want to add it twice.
The reason why `executeReadOnlyRequest` also inserts the CID is for the for when we invoke this method from another part of the code (currently, there isn't such case though).

bftengine/src/bftengine/SimpleClientImp.cpp
yontyon: Fixed

.conan/rocksdb_pkg/conanfile.py
yontyon: Fixed

bftengine/src/bftengine/SimpleClientImp.cpp
yontyon: Fixed

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
yontyon: FIxed

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
yontyon: OK.
I've changed the constructor.
Yet, it still has a fundamental security problem:
To build the CID from the raw data, the method relays on the CID size that the *client* provided. 
Thus, a client may supply a bigger length for the CID and this way to leak data to the server logs.
The immediate solution is to define a maximum length for the CID, and then always allocate this size.
Do you have a better idea?

kvbc/include/KVBCInterfaces.h
yontyon: FIxed

util/pyclient/bft_msgs.py
yontyon: Fixed

util/pyclient/test_msgs.py
yontyon: Fixed

logging/test/logger_test.cpp
yontyon: I removed this test.
I think that for logging, we can see that it works in the logs, so we don't need such a simple test.

logging/include/Logging.hpp
yontyon: Done.
I changed the framework to support RAII.

bftengine/src/bftengine/SimpleClientImp.cpp
andrewjstone: This still looks wrong. It should take `flags` and not `isReadOnly`.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
andrewjstone: I agree, but these are two vastly different security issues. The one I was talking about allows buffer overflows that could allow fulll Remote Code Execution vulnerabilities. The one you are talking about is a denial of service attack where the user can crash the server by allocating too much memory. Both are important.

In many other places in the code we do define a max size as you suggest. We don't have to preallocate though. We can just check that the cid size is within a max bound and then allocate the actual size. Using memcpy with the given size won't overflow any buffers and will prevent that attack, while the max size will prevent the DOS you mention. 

We can add the max size code here or in a followup PR.

logging/include/Logger.hpp
```diff
@@ -25,5 +25,28 @@
...
+  std::string key_;
+
+ public:
+  MDC(concordlogger::Logger& logger, const std::string& key, const std::string& value);
+  ~MDC();
+};
+}  // namespace concordlogger
+
+/*
+ * These macros meant to append a temporary key-value pairs to the log messages.
```
andrewjstone: grammar: `are meant`

grammar: remove the `a`

bftengine/src/bftengine/SimpleClientImp.cpp
yontyon: Fixed

PR #396, opened 2020-02-11
Apollo tests for ReadOnlyReplica

toly-kournik:
Apollo tests for ReadOnlyReplica
Added an option to explicitly provide configuration for bft_network in Apollo.
Piggybacking on this commit to fix ReplicaLoader assertion.

### DIFF
--- a/tests/apollo/test_skvbc_ro_replica.py
+++ b/tests/apollo/test_skvbc_ro_replica.py
@@ -49,8 +49,7 @@ class SkvbcReadOnlyReplicaTest(unittest.TestCase):
 
 ####################################################################################################################### 
     @with_trio
-    @with_bft_network(start_replica_cmd=start_replica_cmd,
-                      explicit_config=[{'n': 4, 'f': 1, 'c': 0, 'num_clients': 1, 'num_ro_replicas': 1}])
+    @with_bft_network(start_replica_cmd=start_replica_cmd, num_ro_replicas=1)
     @verify_linearizability
     async def test_ro_replica_start_with_delay(self, bft_network, tracker):
         """
@@ -60,7 +59,6 @@ class SkvbcReadOnlyReplicaTest(unittest.TestCase):
         Wait for State Transfer in ReadOnlyReplica to complete.
         """
         [bft_network.start_replica(i) for i in range(0, bft_network.config.n)]
-        
         with trio.fail_after(60):  # seconds
             async with trio.open_nursery() as nursery:
                 nursery.start_soon(tracker.send_indefinite_tracked_ops)
@@ -76,21 +74,20 @@ class SkvbcReadOnlyReplicaTest(unittest.TestCase):
                             nursery.cancel_scope.cancel()
         
         # start the read-only replica
-        bft_network.start_replica(4)
+        ro_replica_id = bft_network.config.n
+        bft_network.start_replica(ro_replica_id)
         with trio.fail_after(60):  # seconds
                 while True:
                     with trio.move_on_after(.5):  # seconds
                         key = ['replica', 'Gauges', 'lastExecutedSeqNum']
-                        replica_id = 4
-                        lastExecutedSeqNum = await bft_network.metrics.get(replica_id, *key)
+                        lastExecutedSeqNum = await bft_network.metrics.get(ro_replica_id, *key)
                         # success!
                         if lastExecutedSeqNum >= 150:
-                            print("Replica 4: lastExecutedSeqNum:" + str(lastExecutedSeqNum))
+                            print("Replica " + str(ro_replica_id) + ": lastExecutedSeqNum:" + str(lastExecutedSeqNum))
                             break
 ####################################################################################################################### 
     @with_trio
-    @with_bft_network(start_replica_cmd=start_replica_cmd,
-                      explicit_config=[{'n': 4, 'f': 1, 'c': 0, 'num_clients': 1, 'num_ro_replicas': 1}])
+    @with_bft_network(start_replica_cmd=start_replica_cmd, num_ro_replicas=1)
     @verify_linearizability
     async def test_ro_replica_start_simultaneously (self, bft_network, tracker):
         """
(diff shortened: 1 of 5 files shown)


tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -0,0 +1,119 @@
...
+                        replica_id = 4
+                        lastExecutedSeqNum = await bft_network.metrics.get(replica_id, *key)
+                        # success!
+                        if lastExecutedSeqNum >= 150:
+                            print("Replica 4: lastExecutedSeqNum:" + str(lastExecutedSeqNum))
+                            break
+####################################################################################################################### 
+    @with_trio
+    @with_bft_network(start_replica_cmd=start_replica_cmd,
+                      explicit_config=[{'n': 4, 'f': 1, 'c': 0, 'num_clients': 1, 'num_ro_replicas': 1}])
```
teoparvanov: Hi Toly, why does this only work for 4 replicas?

andrewjstone: I'm against this change, for the simple reason that all tests should pass in all valid configurations. This removes the power from the decorator and may hide bugs.

toly-kournik: @teoparvanov We are planning to extend the tests.
@andrewjstone The existing configuration (interesting_configs) is not flexible enough.
For instance without adding the explicit_config I had to change the existing interesting_configs and several test cased (lambdas choosing configs) to support another configuration parameter.
I'm a python amateur, so please suggest another way.
Besides I don't get the 'all tests should pass in all valid configurations'. Lambdas are used to chose valid configurations from the existing ones. I gave another way to provide a valid configuration. Also it feels strange to change the infra (bft.py) every time we need to add configuration.

teoparvanov: @toly-kournik, let's first agree that your current approach is even less flexible than the existing configuration (interesting_configs). So we need to find a better solution.

I would suggest to keep `bft.interesting_configs()` unchanged and add a `num_ro_replicas` (by default = 0) parameter to the `@with_bft_network` decorator. Then you pass it to the `TestConfig` and it works as expected.

By doing the above, we achieve the following:
* support a RO replica as an "add-on" to the BFT network
* can run RO replica against a BFT network of arbitrary size (N = 3f + 2c +1)

toly-kournik: makes sense

tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -0,0 +1,116 @@
...
+    @with_bft_network(start_replica_cmd=start_replica_cmd, num_ro_replicas=1)
+    @verify_linearizability
+    async def test_ro_replica_start_with_delay(self, bft_network, tracker):
+        """
+        Start up N of N regular replicas.
+        Send client commands until checkpoint 1 is reached.
+        Start read-only replica.
+        Wait for State Transfer in ReadOnlyReplica to complete.
+        """
+        [bft_network.start_replica(i) for i in range(0, bft_network.config.n)]
```
teoparvanov: you could simply say: `bft_network.start_all_replicas()`

tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -0,0 +1,116 @@
...
+        with trio.fail_after(60):  # seconds
+                while True:
+                    with trio.move_on_after(.5):  # seconds
+                        key = ['replica', 'Gauges', 'lastExecutedSeqNum']
+                        lastExecutedSeqNum = await bft_network.metrics.get(ro_replica_id, *key)
+                        # success!
+                        if lastExecutedSeqNum >= 150:
+                            print("Replica " + str(ro_replica_id) + ": lastExecutedSeqNum:" + str(lastExecutedSeqNum))
+                            break
+#######################################################################################################################
```
teoparvanov: nit - I don't think the "###" line separator is necessary...

tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -0,0 +1,116 @@
...
+                while True:
+                    with trio.move_on_after(.5):  # seconds
+                        key = ['replica', 'Gauges', 'lastStableSeqNum']
+                        replica_id = 0
+                        lastStableSeqNum = await bft_network.metrics.get(replica_id, *key)
+                        
+                        if lastStableSeqNum >= 150:
+                            #enough requests
+                            print("Consensus: lastStableSeqNum:" + str(lastStableSeqNum))  
+                            nursery.cancel_scope.cancel()
```
teoparvanov: have you considered reusing the `skvbc.fill_and_wait_for_checkpoint()` method?

tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -0,0 +1,116 @@
...
+        bft_network.start_replica(ro_replica_id)
+        with trio.fail_after(60):  # seconds
+                while True:
+                    with trio.move_on_after(.5):  # seconds
+                        key = ['replica', 'Gauges', 'lastExecutedSeqNum']
+                        lastExecutedSeqNum = await bft_network.metrics.get(ro_replica_id, *key)
+                        # success!
+                        if lastExecutedSeqNum >= 150:
+                            print("Replica " + str(ro_replica_id) + ": lastExecutedSeqNum:" + str(lastExecutedSeqNum))
+                            break
```
teoparvanov: extract in a separate method - f.e. `wait_for_ro_replica_to_sync()`

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: done

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: Do you prefer another character?

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: didn't know it existed... we do

toly-kournik: As agreed with @teoparvanov I'll do it after implementing tracked_fill_and_wait_for_checkpoint()

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: Same as in the above comment since the function is not exactly the same in two cases - nursery.cancel_scope.cancel() should be addressed

PR #398, opened 2020-02-11
BCStateTran: Evict VBlock from cache correctly

andrewjstone:
There's an off by one error that causes virtual blocks to never be
evicted from the cache. This results in the assertion at
BCStateTran:1510 firing during load tests.

This commit fixes the bug and now LRU evicts when the max number of
vblocks exist in the cache.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -1500,7 +1500,7 @@ void BCStateTran::setVBlockInCache(const DescOfVBlockForResPages &desc, char *vB
 
   Assert(p == cacheOfVirtualBlockForResPages.end());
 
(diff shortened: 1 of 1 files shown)


andrewjstone:
> PS: is there an Apollo/unit test we could write to reproduce this & test the fix?

Yes, but it may be somewhat expensive to run. I think reproduction would work with a single stale node and creation of 28 checkpoints. Easy enough to write, but not high priority. I'll give it a whirl shortly if not interrupted :)

teoparvanov:
> 
> 
> > PS: is there an Apollo/unit test we could write to reproduce this & test the fix?
> 
> Yes, but it may be somewhat expensive to run. I think reproduction would work with a single stale node and creation of 28 checkpoints. Easy enough to write, but not high priority. I'll give it a whirl shortly if not interrupted :)

Food for thought for our long-running tests... :-)

PR #387, opened 2020-02-02
Use log4cp always on travis

Liorgal28:
Insert Log4cplus as default logging in Travis.

PR #399, opened 2020-02-12
Fix logger cmake

yontyon:


teoparvanov:
@yontyon, why are you removing those two lines?
They were introduced as part of another PR for enabling log4cpp in CI by @Liorgal28 :
https://github.com/vmware/concord-bft/pull/387

Pls help me understand.

PR #404, opened 2020-02-21
SparseMerkle: Complete key removal functionality

andrewjstone:
This commit builds upon the previous work that enabled deletion of
individual keys inside BatchedInternalNodes. This enhances the update
functionality, such that keys can be removed as well as added and
overwritten during a `Tree::update` call.

A remove operation is the inverse of an addition, and consequently, must
reverse consequences of addition like splitting nodes. For example,
given a tree with depth 2, that looks like the following, where there
are 2 keys in the bottom BatchedInternalNode, labeled with it's nibble
path, "b":

      Root
       |
      "b"

If the hashes of the two keys are "bax" and "bex", and "bax" is
removed, then the "b" node will also be removed, and "bex" will be
**promoted** to live in the root node. Furthermore, "bex" will move up the
4-level tree inside the root BatchedInternalNode such that it resides at
the highest level in that internal tree where it has a corresponding
peer node. If "bex" is the only child in the root node, then it will
live at depth 1, right under the root of the BatchedInternalNode tree.

Note that if the tree was deeper, it is possible that "bex" would be
promoted all the way to the root still. An example of this is if the two
keys were "bex" and "bey". They would then generate a
BatchedInternalNode at depth 3 named "be", as in the tree below:

     Root
      |
     "b"
      |
     "be"

Removing one of them would cause the other to move upwards until it
reached the root.

The purpose of these examples is to display the primary invariant of a
remove operation on a tree:

    After the removal of a key on a tree, the tree should look the same
    as if the key was never inserted, without regards to versions.

In other words we should maintain the shape of the tree from before a
key is added, after that key is removed.

It's important to realize, that this promotion/compaction, while
computationally more complex, is actually an optimization. Without it,
new versions of BatchedInternalNodes further down the tree would need to
be written to disk.  Additionally, proofs would require more reads from
disks given a deeper tree, but the same number of keys. We always want
to keep the BatchedInternalNodes as densely populated as possible to
reduce disk IOPS.

A key component of the algorithm is the `Walker` class, that manages
loading and caching BatchedInternalNodes as part of the tree traversal.
It also manages the tracking of stale BatchedInternalNodes.
It is used for both insertion and removal of keys. The architectural
separation is such that the Walker is supposed to walk the tree and then
the Tree code itself will mutate the current BatchedInternalNode. This
isn't 100% true, but that's how you can think about it. Additionally,
the walker tries to abstract the cache from the insert and remove
algorithms. This isn't 100% true either, as stale Leafs are still
directly marked from the Tree code.

Many tests were added that construct trees of certain shapes and remove
keys to trigger different code paths. These tests use the following
types of assertions upon the UpdateBatches returned from update
operations to help ensure correctness:
 * The correct nodes and keys are updated
 * The updated nodes are properly linked to form a subtree
 * The correct stale nodes are returned
 * The nodes contain the correct keys
 * The nodes have the proper internal shape

Additionally, other functionality was added as part of this PR:
 * `toString()` methods were added for many types to aid in debugging
 * Types such as `UpdateCache` and `Walker` were moved into their own
 files
 * Base types were enhanced with `<<` operators to obviate the need for
 string allocation in future logging calls.
 * gitignore was enhanced to ignore apollo test files

### DIFF
--- a/storage/include/sparse_merkle/internal_node.h
+++ b/storage/include/sparse_merkle/internal_node.h
@@ -50,6 +50,9 @@ BatchedInternalNode and their corresponding header comments.
 // in a BatchedInternalNode, although there will not be any children below it.
 struct LeafChild {
   static constexpr auto SIZE_IN_BYTES = Hash::SIZE_IN_BYTES + LeafKey::SIZE_IN_BYTES;
+  std::string toString() const {
+    return std::string("LeafChild {hash=") + hash.toString() + ", key = " + key.toString() + "}";
+  }
   bool operator==(const LeafChild& other) const { return hash == other.hash && key == other.key; }
   // The hash of the value blob stored at `key`
   Hash hash;
@@ -60,6 +63,9 @@ struct LeafChild {
 // will refer to another BatchedInternalNode if it resides at height 0.
 struct InternalChild {
   static constexpr auto SIZE_IN_BYTES = Hash::SIZE_IN_BYTES + Version::SIZE_IN_BYTES;
+  std::string toString() const {
+    return std::string("InternalChild {hash=") + hash.toString() + ", version= " + version.toString() + "}";
+  }
   bool operator==(const InternalChild& other) const { return hash == other.hash && version == other.version; }
   Hash hash;
   Version version;
@@ -136,7 +142,7 @@ typedef std::variant<LeafChild, InternalChild> Child;
 class BatchedInternalNode {
  public:
   static constexpr size_t MAX_CHILDREN = 31;
-  using ChildrenContainer = std::array<std::optional<Child>, MAX_CHILDREN>;
+  using Children = std::array<std::optional<Child>, MAX_CHILDREN>;
 
   // The leaf was inserted into this node successfully.
   //
@@ -196,12 +202,14 @@ class BatchedInternalNode {
   // In some cases where this value is returned there was one remaining child
   // that was a peer of the removed LeafChild and is able to be promoted to live
   // in the parent BatchedInternalNode. Return it so that the caller can move it
-  // to that node. Note that if this is the root most BatchedInternalNode, the caller will
-  // have to create a new node with this LeafChild. This does not have any
-  // overhead, as the caller must write updates to disk regardless of whether
-  // the node is a new node or a modified node.
+  // to that node.
   struct RemoveBatchedInternalNode {
+    RemoveBatchedInternalNode(Version removed_version) : removed_version(removed_version) {}
+    RemoveBatchedInternalNode(std::optional<LeafChild> promoted, Version removed_version)
+        : promoted(promoted), removed_version(removed_version) {}
+
     std::optional<LeafChild> promoted;
+    Version removed_version;
   };
 
   // A remove has failed, because the the leaf is further down the tree.
@@ -217,14 +225,20 @@ class BatchedInternalNode {
   BatchedInternalNode() = default;
 
   // Construct from the passed children container.
-  BatchedInternalNode(const ChildrenContainer& children) : children_{children} {}
+  BatchedInternalNode(const Children& children) : children_{children} {}
 
   // Insert a LeafChild into this internal node. Return a type indicating success or
   // failure that contains the necessary data for the caller (the sparse merkle
   // tree implementation) to do the right thing.
   //
   // `depth` represents how many nibbles down the sparse merkle tree this BatchedInternalNode is.
-  InsertResult insert(const LeafChild& child, size_t depth);
+  //
+  // We pass `current_version`, because we may want to insert an existing LeafChild
+  // that was promoted as a result of removal of BatchedInternalNodes below this
+  // one in the tree. The version of the LeafChild in this case is going to be
+  // an older version than the current version. However, we want the internal
+  // nodes, and the new root to have the current version.
+  InsertResult insert(const LeafChild& child, size_t depth, Version current_version);
 
   // Remove a LeafChild from this internal node. Return a type indicating to the
   // caller whether the key was successfully removed, and whether the caller
@@ -239,7 +253,12 @@ class BatchedInternalNode {
   //
   // This occurs when this InternalChild points to another BatchedInternalNode
   // down the tree, and that BatchedInternalNode was just updated.
-  void write_internal_child_at_level_0(Nibble child_key, const InternalChild& child);
+  void linkChild(Nibble child_key, const InternalChild& child);
+
+  // Remove a pointer to another BatchedInternalNode.
+  //
+  // This is called when the BatchedInternalNode further down the tree was removed.
+  std::optional<LeafChild> unlinkChild(Nibble child_key, Version version, const std::optional<LeafChild>& promoted);
 
   // Return the root hash of this node.
   const Hash& hash() const { return getHash(0); }
@@ -257,10 +276,22 @@ class BatchedInternalNode {
   size_t numLeafChildren() const;
 
   // Return the internal children container.
-  const ChildrenContainer& children() const { return children_; }
+  const Children& children() const { return children_; }
 
   // Return true if this node is identical to the other one and false otherwise.
   bool operator==(const BatchedInternalNode& other) const { return children_ == other.children_; }
+
+  // Take a nibble representing the logical location a child at height 0 in a
+  // BatchedInternalNode and return the index of that child in `children_`.
+  //
+  // This should only be used outside of this class during testing.
+  size_t nibbleToIndex(Nibble nibble) const;
+
+  // Return true if there are no linked children (InternalChild at level 0) and no leaf children.
+  bool safeToRemove() const;
+
+  // Return a string representation useful for debugging
+  std::string toString() const;
 
  private:
   // A LeafChild collission has occurred. We need to create new InternalChild nodes so
@@ -297,7 +328,13 @@ class BatchedInternalNode {
   //
   // Return InsertComplete if the node was empty.
   // Otherwise return std::nullopt.
-  std::optional<InsertResult> insertIntoEmptyChild(size_t index, const LeafChild& child);
+  //
+  // We pass the version because we may want to insert an existing LeafChild
+  // that was promoted as a result of removal of BatchedInternalNodes below this
+  // one in the tree. The version of the LeafChild in this case is going to be
+  // an older version than the current version. However, we want the internal
+  // nodes, and the new root to have the current version.
+  std::optional<InsertResult> insertIntoEmptyChild(size_t index, const LeafChild& child, Version current_version);
 
   // Attempt to overwrite a matching key at index.
   //
@@ -323,10 +360,6 @@ class BatchedInternalNode {
   //
   // Update the hashes and versions of nodes along the way.
   void updateHashes(size_t index, Version version);
-
-  // Take a nibble representing the logical location a child at height 0 in a
-  // BatchedInternalNode and return the index of that child in `children_`.
-  size_t nibble_to_index(Nibble nibble);
 
   // Return the the height of the node at the given index.
   //
@@ -366,9 +399,12 @@ class BatchedInternalNode {
   // Is this node a left child?
   bool isLeftChild(size_t index) const { return index % 2 != 0; }
 
+  // The number of children at level 0 of the 4 level tree in this BatchedInternalNode
+  size_t numLevel0Children() const;
+
   // All internal and leaf nodes are stored in this array, starting from the
   // root and proceeding level by level from left to right.
-  ChildrenContainer children_;
+  Children children_;
 };
 
 }  // namespace sparse_merkle
--- a/storage/include/sparse_merkle/update_batch.h
+++ b/storage/include/sparse_merkle/update_batch.h
@@ -1,0 +1,45 @@
+
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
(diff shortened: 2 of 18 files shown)


andrewjstone:
This is no longer WIP. Please Review.

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -273,13 +406,658 @@ TEST(tree_tests, multiple_updates) {
...
+  ASSERT_EQ("", root_key.path().toString());
+  ASSERT_EQ(Version(2), root_key.version());
+  ASSERT_EQ(Version(2), remove_root_node.version());
+
+  // There are no leaf updates, since key3 was not updated. A pointer
+  // (LeafChild) was just moved to the root node.
+  ASSERT_EQ(0, batch.leaf_nodes.size());
+  ASSERT_EQ(1, remove_root_node.numLeafChildren());
+  ASSERT_TRUE(leafChildExists("key3", 1, remove_root_node));
+
```
dartdart26: Should we check `batch.stale.leaf_keys` for `key9` here?

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -273,13 +406,658 @@ TEST(tree_tests, multiple_updates) {
...
+  // The hash of the old root should differ from the hash of the new root
+  ASSERT_NE(root_node.hash(), remove_root_node.hash());
+}
+
+// Add 3 nodes such that they all overlap in the first nibble, and two overlap in
+// the second nibble. Thus 1 LeafChild resides at depth 1, and the others reside at depth 2
+// after insert. Removing one of the nodes at depth 2 should cause a promotion
+// to depth 1 such that two LeafChildren live at depth 1.
```
dartdart26: I am trying to visualize the tree in this test for my understanding. Could you please explain what `depth` means in this context?

andrewjstone: As we discussed offline, depth is the location of the BatchedInternalNode in the tree. So for a key with hash "bac..." root would be at depth 0, b at 1, a at 2, and c at 3.

dartdart26: Thanks!

storage/include/sparse_merkle/internal_node.h
```diff
@@ -136,7 +142,7 @@ typedef std::variant<LeafChild, InternalChild> Child;
 class BatchedInternalNode {
  public:
   static constexpr size_t MAX_CHILDREN = 31;
-  using ChildrenContainer = std::array<std::optional<Child>, MAX_CHILDREN>;
+  using Children = std::array<std::optional<Child>, MAX_CHILDREN>;
```
dartdart26: I think that is a good simplification.

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -31,19 +36,130 @@ void db_put(const shared_ptr<TestDB>& db, const Tree::UpdateBatch& batch) {
   }
 }
 
-template <class T>
-void assert_version(uint64_t version, const T& keys) {
+template <typename T>
+::testing::AssertionResult keyVersionMatches(uint64_t version, const T& keys) {
```
dartdart26: I like the idea of using `AssertionResult`!

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -273,13 +406,658 @@ TEST(tree_tests, multiple_updates) {
...
+  std::shared_ptr<TestDB> db(new TestDB);
+  Tree tree(db);
+  SetOfKeyValuePairs updates;
+  updates.emplace(Sliver("key3"), Sliver("val1"));
+  updates.emplace(Sliver("key9"), Sliver("val2"));
+  auto batch = tree.update(updates);
+
+  // There should be 2 internal nodes, with no leaf children in the root node.
+  ASSERT_EQ(2, batch.internal_nodes.size());
+  auto& root_node = batch.internal_nodes[0].second;
```
dartdart26: Can we rely on internal node ordering in the batch? If yes, maybe we can document it. Or maybe we can encourage users to use the root getters in `Tree`. I realise this is test code and my comment is mostly about the API.

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -273,13 +406,658 @@ TEST(tree_tests, multiple_updates) {
...
+// should cause the other leaf child to go to "b", but leave "ba" alone. Thus
+// after the delete, the tree should look like the following:
+//
+//                Root
+//                 |
+//                "b"
+//                 |
+//                "ba"
+//
+TEST(tree_tests, remove_left_node_and_promote_leaving_right_node_alone) {
```
dartdart26: This test provides a brilliant example, thank you!

andrewjstone: :)

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -273,13 +406,658 @@ TEST(tree_tests, multiple_updates) {
...
+  // "" and "b" should be updated.
+  ASSERT_EQ(2, batch.internal_nodes.size());
+  auto& [root_key, root_node] = batch.internal_nodes[0];
+  auto& [b_key, b_node] = batch.internal_nodes[1];
+  ASSERT_EQ("", root_key.path().toString());
+  ASSERT_EQ("b", b_key.path().toString());
+  ASSERT_EQ(0, root_node.numLeafChildren());
+  ASSERT_EQ(MAX_INTERNAL_PATH, root_node.numInternalChildren());
+  ASSERT_EQ(0, b_node.numLeafChildren());
+  ASSERT_EQ(MAX_INTERNAL_PATH, b_node.numInternalChildren());
```
dartdart26: Could you please explain the `MAX_INTERNAL_PATH` asserts?

storage/test/sparse_merkle/tree_test.cpp
andrewjstone: We actually can rely on the ordering at this moment, but I'm not sure we should. We may end up changing this to be a set instead of a vector, although a vector probably has better perf characteristics (if it matters.). It's mainly just internal knowledge used in testing. I prefer to leave it undocumented for now. Of course, if we change it, then we'll need to rework the test.

storage/test/sparse_merkle/tree_test.cpp
andrewjstone: Thanks! I was getting test failures that always showed the failure coming from the called function that used `ASSERT_XXX` macros and so I couldn't actually tell where in the caller the test was failing, since those called functions were called multiple times in a single test. This lets us actually know which part of the test is failing. Good on gtest for providing a solution.

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -273,13 +406,658 @@ TEST(tree_tests, multiple_updates) {
...
+  // There is a new version of the root only.
+  ASSERT_EQ(1, batch.internal_nodes.size());
+  auto& [root_key2, root_node2] = batch.internal_nodes[0];
+  ASSERT_EQ("", root_key2.path().toString());
+
+  // There is only one leaf and the root
+  ASSERT_EQ(1, root_node2.numLeafChildren());
+  ASSERT_EQ(1, root_node2.numInternalChildren());
+
+  // There are 3 old internal keys (root, "b") and one stale leaf
```
dartdart26: And also "ba"?

storage/test/sparse_merkle/tree_test.cpp
dartdart26: I am going to use that in the future.

storage/test/sparse_merkle/tree_test.cpp
```diff
@@ -216,34 +337,42 @@ TEST(tree_tests, multiple_updates) {
...
   ASSERT_EQ(1, batch.stale.leaf_keys.size());
 
-  // The root node was updated on the second update. But the next
-  // BatchedInternalNode down wasn't, since the udpate inserted into the root
-  // node.
-  ASSERT_EQ(Version(2), batch.stale.internal_keys[0].version());
-  ASSERT_EQ(Version(1), batch.stale.internal_keys[1].version());
+  // The root node was updated on the insert of key1, but the next
+  // BatchedInternalNode down wasn't, since key1 was inserted into the root
+  // node.ys
```
dartdart26: Maybe a typo?

storage/include/sparse_merkle/internal_node.h
```diff
@@ -217,14 +225,14 @@ class BatchedInternalNode {
...
-  BatchedInternalNode(const ChildrenContainer& children) : children_{children} {}
+  BatchedInternalNode(const Children& children) : children_{children} {}
 
   // Insert a LeafChild into this internal node. Return a type indicating success or
   // failure that contains the necessary data for the caller (the sparse merkle
   // tree implementation) to do the right thing.
   //
   // `depth` represents how many nibbles down the sparse merkle tree this BatchedInternalNode is.
-  InsertResult insert(const LeafChild& child, size_t depth);
+  InsertResult insert(const LeafChild& child, size_t depth, Version current_version);
```
dartdart26: Could you please document the `current_version` parameter as opposed to `child.key.version()`?

dartdart26: I saw it in `insertIntoEmptyChild()`.

storage/include/sparse_merkle/update_batch.h
```diff
@@ -0,0 +1,45 @@
...
+// to the database.
+struct UpdateBatch {
+  StaleNodeIndexes stale;
+  std::vector<std::pair<InternalNodeKey, BatchedInternalNode>> internal_nodes;
+  std::vector<std::pair<LeafKey, LeafNode>> leaf_nodes;
+};
+
+}  // namespace sparse_merkle
+}  // namespace storage
+}  // namespace concord
```
dartdart26: [Minor]: Missing new line.

storage/include/sparse_merkle/update_cache.h
```diff
@@ -0,0 +1,80 @@
...
+  // These nodes are mutable and are all being updated to a single version.
+  // Therefore we key them by their NibblePath alone, without a version.
+  std::map<NibblePath, BatchedInternalNode> internal_nodes_;
+};
+
+}  // namespace detail
+
+}  // namespace sparse_merkle
+}  // namespace storage
+}  // namespace concord
```
dartdart26: [Minor]: Missing new line.

storage/include/sparse_merkle/walker.h
```diff
@@ -0,0 +1,81 @@
...
+  NodeStack stack_;
+  BatchedInternalNode current_node_;
+  NibblePath nibble_path_;
+};
+
+}  // namespace detail
+
+}  // namespace sparse_merkle
+}  // namespace storage
+}  // namespace concord
```
dartdart26: [Minor]: Missing new line.

storage/src/sparse_merkle/base_types.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+std::ostream& operator<<(std::ostream& os, const NibblePath& path) {
+  for (size_t i = 0; i < path.length(); i++) {
+    os << path.get(i).hexChar();
+  }
+  return os;
+}
+
+}  // namespace concord::storage::sparse_merkle
```
dartdart26: [Minor]: Missing new line.

storage/test/sparse_merkle/tree_test.cpp
dartdart26: Agreed.

storage/test/sparse_merkle/tree_test.cpp
andrewjstone: It's to check that in the case of a single InternalChild at the Leaf (depth 4) of the BatchedInternalNode, pointing to another BatchedInternalNode, that there is a path from the root of the BatchedInternalNode to that leaf. It's actually a property of the BatchedInternalNode and not the tree, and as such doesn't really belong in this test. I think it's also confusing and as it doesn't add much I'm going to remove it.

storage/test/sparse_merkle/tree_test.cpp
andrewjstone: Done. I also made it so that we check that proper internal node is removed as well.

storage/test/sparse_merkle/tree_test.cpp
andrewjstone: fixed.

storage/test/sparse_merkle/tree_test.cpp
andrewjstone: fixed.

storage/include/sparse_merkle/internal_node.h
andrewjstone: Good idea! Done.

storage/src/sparse_merkle/base_types.cpp
andrewjstone: fixed.

storage/include/sparse_merkle/walker.h
andrewjstone: fixed.

storage/include/sparse_merkle/update_batch.h
andrewjstone: fixed.

andrewjstone:
Thanks for the review @dartdart26 !

PR #405, opened 2020-02-21
Fix usage of invalid RocksDB iterators

dartdart26:
Remove calls to rocksdb::Iterator::Next()/Prev()/key()/value() when the
iterator is invalid. Documentation in iterator.h specifies that above
methods require a valid iterator. Assumption is that if the iterator is
invalid, behavior would be undefined.

### DIFF
--- a/storage/src/rocksdb_client.cpp
+++ b/storage/src/rocksdb_client.cpp
@@ -401,10 +401,15 @@ KeyValuePair ClientIterator::seekAtLeast(const Sliver &_searchKey) {
  * @return The previous key value pair.
  */
 KeyValuePair ClientIterator::previous() {
-  m_iter->Prev();
-
   if (!m_iter->Valid()) {
-    LOG_ERROR(logger, "Iterator out of bounds");
+    LOG_ERROR(logger, "Iterator is not valid");
+    m_status = Status::GeneralError("Iterator is not valid");
+    return KeyValuePair();
+  }
+  m_iter->Prev();
+  if (!m_iter->Valid()) {
+    LOG_WARN(logger, "No previous key");
+    m_status = Status::GeneralError("No previous key");
     return KeyValuePair();
   }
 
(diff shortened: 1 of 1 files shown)


storage/src/rocksdb_client.cpp
```diff
@@ -401,10 +401,13 @@ KeyValuePair ClientIterator::seekAtLeast(const Sliver &_searchKey) {
...
 KeyValuePair ClientIterator::previous() {
+  if (!m_iter->Valid()) {
+    LOG_ERROR(logger, "Iterator is not valid");
+    return KeyValuePair();
+  }
   m_iter->Prev();
-
   if (!m_iter->Valid()) {
-    LOG_ERROR(logger, "Iterator out of bounds");
+    LOG_ERROR(logger, "No previous key");
```
andrewjstone: Nit: I think this should perhaps be a warning, since it's reasonable to ask previous on a valid iterator, unlike the case above.

storage/src/rocksdb_client.cpp
```diff
@@ -429,6 +432,11 @@ KeyValuePair ClientIterator::next() {
     LOG_DEBUG(logger, "Reading count = " << g_rocksdb_called_read);
   }
 
+  if (!m_iter->Valid()) {
+    LOG_ERROR(logger, "Iterator is not valid");
+    m_status = Status::GeneralError("Iterator is not valid");
```
andrewjstone: Why do we set the m_status here, but not in `previous` ?

storage/src/rocksdb_client.cpp
dartdart26: Agreed, changing as you suggest. Copy-paste...

storage/src/rocksdb_client.cpp
dartdart26: I noticed it too. I am not really sure, I suspect it was just missed. However, m_status is not being used anywhere. Therefore, I will also set it here.

dartdart26:
As per @andrewjstone 's comments:

- Change log level on error messages after m_iter->Next() and m_iter->Previous() to WARN.
- Set m_status in previous().

PR #414, opened 2020-02-27
[Apollo] Long-running tests Make some of existing decorators optional

Liorgal28:
Add the option to get bft_network from the calling function on @with_trio and @with_bft_network, then we just send it to the async_fn.

Tests that will run from other tests with already initialized bft_network will pass the creation of new bft_network and just run the test.
The same with with_trio decorator, if we already made a trio.run so we cant make it again so we are checking if it already made and if it is (the times when a test called from another test) we just call the requested test.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -75,7 +75,10 @@ def with_trio(async_fn):
     """ Decorator for running a coroutine (async_fn) with trio. """
     @wraps(async_fn)
     def trio_wrapper(*args, **kwargs):
-        return trio.run(async_fn, *args, **kwargs)
+        if "bft_network" in kwargs:
+            return async_fn(*args, **kwargs)
+        else:
+            return trio.run(async_fn, *args, **kwargs)
 
     return trio_wrapper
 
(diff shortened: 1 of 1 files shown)


tests/apollo/util/bft.py
```diff
@@ -75,7 +75,10 @@ def with_trio(async_fn):
     """ Decorator for running a coroutine (async_fn) with trio. """
     @wraps(async_fn)
     def trio_wrapper(*args, **kwargs):
-        return trio.run(async_fn, *args, **kwargs)
+        if "bft_network" in kwargs:
```
teoparvanov: Actually, we should have a separate parameter for disabling the @with_trio decorator (say, "already_in_trio=True/False")

teoparvanov:
> 
> 
> While we are at it, can you also add support for a "skip_linearizability" parameter, which disables the @check_linearizability decorator? You probably won't need it for the long-running BFT tests, but I will for running Apollo as part of product CI. So I'd appreciate your help with this :)

As discussed with @Liorgal28, we can implement this later, in a separate PR.

PR #415, opened 2020-02-28
Correlation ID support for batching

yontyon:
# Introduction
* This PR introduces our approach for correlation ID logging while handling a batch of requests.
* In addition, we also add correlation id support for the pre-execution requests
# Motivation
The correlation ID is an id that every client request has (In particular, if the bft-client hasn't got a correlation ID with a request it creates a correlation id in the format of REQUEST_SEQ_NUM-CLIENT_ID and adjusts it to the request).

Recall that correlation ID is meant for tracking the log messages that particular item produces while going through the system stack.

Yet, when handling a batch of requests we need a strategy for the correlation id logging - as each batch contains multiple requests and thus can be represented by multiple correlation ids. 
# Our Solution
In concord-bft we already have a batch identifier which is the batch sequence number.
Thus, we would like to reuse it as the batch correlation ID.
For that, when starting handling a batch we log the following key-value pairs: {cid, content: [cid list]}{sn, BATCH_SEQ_NUM}.
For example: 
```
[Node 3] [140674221184768] %{cid, content: 6639082829609172992;}{rid, 3}{sn, 334}% INFO |concord||void bftEngine::impl::ReplicaImp::onMessage(T*) [with T = bftEngine::impl::PrePrepareMsg]||
```
Once this mapping between the batch cids and a sequence number is created, we can log only the BATCH_SEQ_NUM as the batch correlation id.

When tracking a specific request, one has to filter the log messages by the cid field to find the relevant BATCH_SEQ_NUM, and then to track the BATCH_SEQ_NUM in the log messages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
     - RELEASE=-DCMAKE_BUILD_TYPE=RELEASE
     - RELEASE=-DCMAKE_BUILD_TYPE=RELEASE USE_ROCKSDB=-DBUILD_ROCKSDB_STORAGE=TRUE
 script:
-  - cd $TRAVIS_BUILD_DIR && mkdir build && cd build && conan install --build missing .. && cmake $CMAKE_CXX_FLAGS $DEBUG $RELEASE $USE_LOG4CPP $USE_ROCKSDB .. && make format-check && make -j $(getconf _NPROCESSORS_ONLN) && ctest --output-on-failure
+  - cd $TRAVIS_BUILD_DIR && mkdir build && cd build && conan install --build missing .. && cmake $CMAKE_CXX_FLAGS $DEBUG $RELEASE $USE_LOG4CPP  $USE_ROCKSDB .. && make format-check && make -j $(getconf _NPROCESSORS_ONLN) && ctest --output-on-failure
 
 cache:
   ccache: true
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1755,6 +1755,7 @@ void ReplicaImp::onMessage<NewViewMsg>(NewViewMsg *msg) {
     return;
   }
   metric_received_new_views_.Get().Inc();
+
   const ReplicaId senderId = msg->senderId();
 
   Assert(senderId != config_.replicaId);  // should be verified in ViewChangeMsg
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -308,7 +308,7 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
   }
   primaryLastUsedSeqNum++;
   {
-    concordlogger::MDC content(GL, "cid", "content: " + pp->getAllRequestsCorrelationIdAsString());
+    concordlogger::MDC content(GL, "cid", "content: " + pp->getBatchCorrelationIdAsString());
     LOG_INFO(GL, "");
   }
   SeqNumInfo &seqNumInfo = mainLog->get(primaryLastUsedSeqNum);
@@ -369,7 +369,7 @@ template <>
 void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
   metric_received_pre_prepares_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   LOG_DEBUG_F(GL,
               "Node %d received PrePrepareMsg from node %d for seqNumber %" PRId64 " (size=%d)",
               (int)config_.replicaId,
@@ -404,7 +404,7 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
 
     if (seqNumInfo.addMsg(msg)) {
       {
-        concordlogger::MDC content(GL, "cid", "content: " + msg->getAllRequestsCorrelationIdAsString());
+        concordlogger::MDC content(GL, "cid", "content: " + msg->getBatchCorrelationIdAsString());
         LOG_INFO(GL, "");
       }
       msgAdded = true;
@@ -563,7 +563,7 @@ template <>
 void ReplicaImp::onMessage<StartSlowCommitMsg>(StartSlowCommitMsg *msg) {
   metric_received_start_slow_commits_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   LOG_INFO_F(GL, "Node %d received StartSlowCommitMsg for seqNumber %" PRId64 "", config_.replicaId, msgSeqNum);
 
   if (relevantMsgForActiveView(msg)) {
@@ -694,7 +694,7 @@ void ReplicaImp::onMessage<PartialCommitProofMsg>(PartialCommitProofMsg *msg) {
   const SeqNum msgSeqNum = msg->seqNumber();
   const SeqNum msgView = msg->viewNumber();
   const NodeIdType msgSender = msg->senderId();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   Assert(repsInfo->isIdOfPeerReplica(msgSender));
   Assert(repsInfo->isCollectorForPartialProofs(msgView, msgSeqNum));
 
@@ -731,7 +731,7 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
               (int)msg->seqNumber());
 
   const SeqNum msgSeqNum = msg->seqNumber();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   if (relevantMsgForActiveView(msg)) {
     SeqNumInfo &seqNumInfo = mainLog->get(msgSeqNum);
     PartialProofsSet &pps = seqNumInfo.partialProofs();
@@ -778,7 +778,7 @@ void ReplicaImp::onMessage<PreparePartialMsg>(PreparePartialMsg *msg) {
   metric_received_prepare_partials_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
   const ReplicaId msgSender = msg->senderId();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   bool msgAdded = false;
 
   if (relevantMsgForActiveView(msg)) {
@@ -828,7 +828,7 @@ void ReplicaImp::onMessage<CommitPartialMsg>(CommitPartialMsg *msg) {
   metric_received_commit_partials_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
   const ReplicaId msgSender = msg->senderId();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   bool msgAdded = false;
 
   if (relevantMsgForActiveView(msg)) {
@@ -872,7 +872,7 @@ void ReplicaImp::onMessage<PrepareFullMsg>(PrepareFullMsg *msg) {
   metric_received_prepare_fulls_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
   const ReplicaId msgSender = msg->senderId();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   bool msgAdded = false;
 
   if (relevantMsgForActiveView(msg)) {
@@ -913,7 +913,7 @@ void ReplicaImp::onMessage<CommitFullMsg>(CommitFullMsg *msg) {
   metric_received_commit_fulls_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
   const ReplicaId msgSender = msg->senderId();
-  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
+  concordlogger::MDC seqNum(GL, SEQ_NUM_KEY, msgSeqNum);
   bool msgAdded = false;
 
   if (relevantMsgForActiveView(msg)) {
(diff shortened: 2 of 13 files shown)


yuliasherman:
> CLIETN

->CLIENT

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
+  while (!it.end() && req_num < index) {
+    it.gotoNext();
+    req_num++;
+  }
+  char* requestBody = nullptr;
+  it.getCurrent(requestBody);
+  ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
+  return req.getCid();
+}
+std::string PrePrepareMsg::getAllRequestsCorrelationIdAsString() {
```
yuliasherman: getAllRequestsCorrelationIdAsString -> getBatchCorrelationIdAsString

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
+std::string PrePrepareMsg::getMessageClientCorrelationId(int index) {
+  auto it = RequestsIterator(this);
+  int req_num = 0;
+  while (!it.end() && req_num < index) {
+    it.gotoNext();
+    req_num++;
+  }
+  char* requestBody = nullptr;
+  it.getCurrent(requestBody);
+  ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
```
yuliasherman: No need to create it as a separate variable; do it directly in return.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
   Assert(false);
   return false;
 }
+std::string PrePrepareMsg::getMessageClientCorrelationId(int index) {
```
yuliasherman: getMessageClientCorrelationId -> getClientCorrelationIdForMsg

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -307,7 +307,10 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
     DebugStatistics::onSendPrePrepareMessage(pp->numberOfRequests(), requestsQueueOfPrimary.size());
   }
   primaryLastUsedSeqNum++;
-
+  {
+    concordlogger::MDC content(GL, "cid", "content: " + pp->getAllRequestsCorrelationIdAsString());
+    LOG_INFO(GL, "");
```
dartdart26: Is that empty so that it just logs the content from above?

bftengine/src/bftengine/SimpleClientImp.cpp
```diff
@@ -192,7 +192,7 @@ int SimpleClientImp::sendRequest(uint8_t flags,
                                  std::string cid) {
   bool isReadOnly = flags & READ_ONLY_REQ;
   bool isPreProcessRequired = flags & PRE_PROCESS_REQ;
-  std::string msgCid = cid.empty() ? std::to_string(reqSeqNum) : cid;
+  std::string msgCid = cid.empty() ? std::to_string(reqSeqNum) + "-" + std::to_string(clientId_) : cid;
```
dartdart26: Can we make `msgCid` const?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
+    req_num++;
+  }
+  char* requestBody = nullptr;
+  it.getCurrent(requestBody);
+  ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
+  return req.getCid();
+}
+std::string PrePrepareMsg::getAllRequestsCorrelationIdAsString() {
+  std::stringstream ret;
+  ;
```
dartdart26: Maybe a typo?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
+    it.gotoNext();
+    req_num++;
+  }
+  char* requestBody = nullptr;
+  it.getCurrent(requestBody);
+  ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
+  return req.getCid();
+}
+std::string PrePrepareMsg::getAllRequestsCorrelationIdAsString() {
+  std::stringstream ret;
```
dartdart26: Maybe we can use the "+" operator for strings instead of `std::stringstream` in order to avoid copying the output in `std::stringstream::str()` .

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
+  ;
+  auto it = RequestsIterator(this);
+  char* requestBody = nullptr;
+  while (it.getAndGoToNext(requestBody)) {
+    ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
```
dartdart26: This object can be `const`. I guess it leads to making `ClientRequestMsg::getCid()` a const member function as it is a getter.

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
```diff
@@ -90,6 +90,9 @@ class PrePrepareMsg : public MessageBase {
 
   void updateView(ViewNum v, CommitPath firstPath = CommitPath::SLOW);
 
+  std::string getMessageClientCorrelationId(int index);
+  std::string getAllRequestsCorrelationIdAsString();
```
dartdart26: Can we make these functions `const`?

bftengine/src/preprocessor/messages/PreProcessRequestMsg.cpp
```diff
@@ -41,4 +54,8 @@ void PreProcessRequestMsg::setParams(NodeIdType senderId, uint16_t clientId, Req
       "senderId=" << senderId << " clientId=" << clientId << " reqSeqNum=" << reqSeqNum << " reqLength=" << reqLength);
 }
 
+std::string PreProcessRequestMsg::getCid() {
```
dartdart26: Can we make this function `const`?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
+  int req_num = 0;
+  while (!it.end() && req_num < index) {
+    it.gotoNext();
+    req_num++;
+  }
+  char* requestBody = nullptr;
+  it.getCurrent(requestBody);
+  ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
+  return req.getCid();
+}
```
dartdart26: Can we add an empty line here?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
+std::string PrePrepareMsg::getAllRequestsCorrelationIdAsString() {
+  std::stringstream ret;
+  ;
+  auto it = RequestsIterator(this);
+  char* requestBody = nullptr;
+  while (it.getAndGoToNext(requestBody)) {
+    ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
+    ret << req.getCid() << ";";
```
dartdart26: [Minor]: Would having a trailing `;` in the output be a problem?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -194,6 +196,29 @@ bool PrePrepareMsg::checkRequests() const {
...
 }
+std::string PrePrepareMsg::getMessageClientCorrelationId(int index) {
+  auto it = RequestsIterator(this);
+  int req_num = 0;
+  while (!it.end() && req_num < index) {
+    it.gotoNext();
+    req_num++;
+  }
+  char* requestBody = nullptr;
+  it.getCurrent(requestBody);
```
dartdart26: If `it.end() == true`, I think we will have a segfault. Since it is a public method, we can either document what the index should be as a precondition (and assert on it) or we should handle the end case in some other way - maybe an optional return or an exception.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -366,7 +369,7 @@ template <>
 void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
   metric_received_pre_prepares_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
-
+  concordlogger::MDC sn(GL, SEQ_NUM_KEY, msgSeqNum);
```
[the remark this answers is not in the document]
yontyon: Fixed

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: Fixed

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: Fixed

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: Fixed

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: Yes, and in the specific format of MDC such that it will be easy to parse it

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: fixed

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: fixed

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: FIxed

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
yontyon: Fixed

bftengine/src/preprocessor/messages/PreProcessRequestMsg.cpp
yontyon: done

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: done

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: I don't think so and I would prefer to leave it simple as we can...

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
yontyon: fixed

bftengine/src/bftengine/SimpleClientImp.cpp
yontyon: fixed

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
dartdart26: Great, makes sense.

.travis.yml
```diff
@@ -17,7 +17,7 @@ env:
     - RELEASE=-DCMAKE_BUILD_TYPE=RELEASE
     - RELEASE=-DCMAKE_BUILD_TYPE=RELEASE USE_ROCKSDB=-DBUILD_ROCKSDB_STORAGE=TRUE
 script:
-  - cd $TRAVIS_BUILD_DIR && mkdir build && cd build && conan install --build missing .. && cmake $CMAKE_CXX_FLAGS $DEBUG $RELEASE $USE_LOG4CPP $USE_ROCKSDB .. && make format-check && make -j $(getconf _NPROCESSORS_ONLN) && ctest --output-on-failure
+  - cd $TRAVIS_BUILD_DIR && mkdir build && cd build && conan install --build missing .. && cmake $CMAKE_CXX_FLAGS $DEBUG $RELEASE $USE_LOG4CPP  $USE_ROCKSDB .. && make format-check && make -j $(getconf _NPROCESSORS_ONLN) && ctest --output-on-failure
```
dartdart26: [Minor]: Do we need this change as part of this MR?

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -81,7 +82,7 @@ void ClientRequestMsg::setParams(NodeIdType sender, ReqId reqSeqNum, uint32_t re
   setParams(reqSeqNum, requestLength, flags);
 }
 
-std::string ClientRequestMsg::getCid() {
+const std::string ClientRequestMsg::getCid() const {
```
dartdart26: [Minor]: Not sure if we want to return a `const` string here.

bftengine/src/bftengine/messages/ClientRequestMsg.hpp
```diff
@@ -53,7 +53,7 @@ class ClientRequestMsg : public MessageBase {
   char* requestBuf() const { return body() + sizeof(ClientRequestMsgHeader); }
 
   void set(ReqId reqSeqNum, uint32_t requestLength, uint8_t flags);
-  std::string getCid();
+  const std::string getCid() const;
```
dartdart26: [Minor]: Same as above, return type can be non-const.

.travis.yml
yontyon: No. But I have another MR that touch this file anyway, so I will leave it as it is for now.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
yontyon: Isn't it a good idea such that no one will be able to change the return value?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -307,7 +307,10 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
     DebugStatistics::onSendPrePrepareMessage(pp->numberOfRequests(), requestsQueueOfPrimary.size());
   }
   primaryLastUsedSeqNum++;
-
+  {
+    concordlogger::MDC content(GL, "cid", "content: " + pp->getBatchCorrelationIdAsString());
+    LOG_INFO(GL, "");
```
upshaw-alex: @yontyon, Won't this statement get logged for every `PrePrepareMsg` sent? If so, I suspect logging this statement at the `INFO` level will be incredibly noisy, contribute to clutter in the logs that makes them more cumbersome to interpret as a human, and accelerate the growth of log file size.

bhavsi: Also @yontyon I am not sure if there is a need to output an INFO level message without any message body? Also, I am not sure if it's a standard to log without a message body.

yontyon: The purpose of this "empty" logline is to create a mapping between the batch correlation IDs and the batch sequence number.
I will change it to TRACE log info and add a simple message into the actual logline

PR #416, opened 2020-02-28
Simplify key searches by introducing seekAtMost()

dartdart26:
A typical operation on merkle tree keys is to search for a key that is
less than or equal to a given key. Current implementation uses
seekAtLeast() and additional logic to implement these kind of searches.
This PR aims to simplify code by introducing the seekAtMost() method
that can be used in the mentioned cases.


dartdart26:
Rebased due to CI failures related to Conan cache.

PR #418, opened 2020-02-28
Long run tests: call other tests inside timed loop

Liorgal28:
create an timed loop that sends requests.
The requests are test scenarios that using pre-provisioned bft_network.

### DIFF
--- a/tests/apollo/test_skvbc.py
+++ b/tests/apollo/test_skvbc.py
@@ -95,7 +95,8 @@ class SkvbcTest(unittest.TestCase):
         Ensure that we can put a block and use the GetBlockData API request to
         retrieve its KV pairs.
         """
-        bft_network.start_all_replicas()
+        if bft_network.pre_built_network is False:
+            bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
         client = bft_network.random_client()
         last_block = skvbc.parse_reply(await client.read(skvbc.get_last_block_req()))
--- a/tests/apollo/test_skvbc.py
+++ b/tests/apollo/test_skvbc.py
@@ -95,8 +95,7 @@ class SkvbcTest(unittest.TestCase):
         Ensure that we can put a block and use the GetBlockData API request to
         retrieve its KV pairs.
         """
-        if bft_network.pre_built_network is False:
-            bft_network.start_all_replicas()
+        bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
         client = bft_network.random_client()
         last_block = skvbc.parse_reply(await client.read(skvbc.get_last_block_req()))
--- a/tests/apollo/test_skvbc_long_running.py
+++ b/tests/apollo/test_skvbc_long_running.py
@@ -11,15 +11,10 @@
 # file.
 
 import os.path
-import random
 import unittest
-import logging
 import trio
 
-from util import bft_network_partitioning as net
-from util import skvbc as kvbc
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
-from util.skvbc_history_tracker import verify_linearizability
 from test_skvbc import SkvbcTest
 
 # Time consts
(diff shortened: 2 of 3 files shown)


tests/apollo/test_skvbc_long_running.py
```diff
@@ -50,7 +50,7 @@ class SkvbcLongRunningTest(unittest.TestCase):
     @with_bft_network(start_replica_cmd,
                       selected_configs=lambda n, f, c: n == 7)
     async def test_stability(self, bft_network):
-        bft_network.start_all_replicas()
         with trio.move_on_after(seconds=ONE_HOUR_IN_SECONDS):
             while True:
-                assert True
+                await SkvbcTest().test_conflicting_write(bft_network=bft_network, already_in_trio=True)
```
teoparvanov: This test is not very representative. Let's use test_get_block_data() instead.

tests/apollo/test_skvbc_long_running.py
```diff
@@ -50,7 +50,7 @@ class SkvbcLongRunningTest(unittest.TestCase):
...
+                bft_network.stop_all_replicas()
```
teoparvanov: Doing this invalidates the whole point of the long running test. The idea is that replicas run for a long time, without being stopped.

tests/apollo/test_skvbc_long_running.py
Liorgal28: picked that just because its the simplest, I'll change it

tests/apollo/test_skvbc_long_running.py
Liorgal28: ok I'll make the start_all_replicas optional as you suggested

Liorgal28:
The changes I've made after Teo review:
1) insert pre_built_network variable into the bft_network entity, this variable says if the bft_network is already running.

2) on each test we check if pre_built_network is False so we start the replicas else we just using the already running replicas.

3) on long_running_tests we start the replicas and just calling to test one by one without stopping the replicas between each call

tests/apollo/test_skvbc_long_running.py
```diff
@@ -51,6 +51,6 @@ class SkvbcLongRunningTest(unittest.TestCase):
                       selected_configs=lambda n, f, c: n == 7)
     async def test_stability(self, bft_network):
         bft_network.start_all_replicas()
-        with trio.move_on_after(seconds=ONE_HOUR_IN_SECONDS):
+        with trio.move_on_after(seconds=ONE_HOUR_IN_SECONDS/30):
             while True:
-                assert True
+                await SkvbcTest().test_get_block_data(bft_network=bft_network, already_in_trio=True)
```
teoparvanov: Let's add some waiting (say, 10 sec) between test executions.

teoparvanov:
PS: I see the failure is related to Conan. I disabled caching for Conan on Friday. Try rebasing and also ask Yoni about this.

tests/apollo/util/bft.py
```diff
@@ -93,6 +93,7 @@ def decorator(async_fn):
         async def wrapper(*args, **kwargs):
             if "bft_network" in kwargs:
                 bft_network = kwargs.pop("bft_network")
+                bft_network.pre_built_network = True
```
teoparvanov: let's call this "is_existing"

tests/apollo/test_skvbc.py
```diff
@@ -95,7 +95,8 @@ class SkvbcTest(unittest.TestCase):
         Ensure that we can put a block and use the GetBlockData API request to
         retrieve its KV pairs.
         """
-        bft_network.start_all_replicas()
+        if bft_network.pre_built_network is False:
```
teoparvanov: `if not bft_network.is_existing:`

tests/apollo/test_skvbc.py
```diff
@@ -95,7 +95,8 @@ class SkvbcTest(unittest.TestCase):
...
+            bft_network.start_all_replicas()
```
teoparvanov: Instead of this, let's modify the `start_all_replicas()` method itself. If it catches an `AlreadyRunningError`, rethrow it only if `is_existing = False`. This way we won't have to modify every single test that starts all replicas.

robem: Don't do `raise e` but `raise` only instead.

tests/apollo/util/bft.py
```diff
@@ -218,7 +220,13 @@ def start_replica_cmd(self, replica_id):
         return self.config.start_replica_cmd(self.builddir, replica_id)
 
     def start_all_replicas(self):
-        [self.start_replica(i) for i in range(0, self.config.n)]
+        try:
+            [self.start_replica(i) for i in range(0, self.config.n)]
+        except AlreadyRunningError as e:
```
teoparvanov: as per Rob's earlier comment, this can simply become:
```python
except AlreadyRunningError
```

PR #420, opened 2020-02-29
Test - send request from client while the primary is off.

oded282:
At the beginning, this test was suppose to check that a block is written even if the request was sent while the primary is off. 
I realized that this is not an expected behavior, so I changed it. 
It seems to me that if a a client send request while the primary is off, this request will be resend for a limited period of time (request timeout), if view change did not finished in this limited time, the block will not be written.
We can set the client's request timeout for a longer period of time and then we get the initial expected behavior.
I did not change the request timeout, therefore our test expects that no new block will be written.
Please, correct me if I'm wrong.
@teoparvanov @guyg8


teoparvanov:
PS: I see the CI failure is related to Conan. I disabled caching for Conan on Friday. Try rebasing and also ask Yoni about this.

tests/apollo/test_skvbc_view_change.py
```diff
@@ -36,8 +35,63 @@ def start_replica_cmd(builddir, replica_id):
             "-s", statusTimerMilli,
             "-v", viewChangeTimeoutMilli]
 
+
 class SkvbcViewChangeTest(unittest.TestCase):
 
+    @with_trio
+    @with_bft_network(start_replica_cmd)
+    async def test_request_block_written_primary_down(self, bft_network):
```
teoparvanov: block_not_written

oded282: Sure.

PR #422, opened 2020-03-02
Enable conan caching in Travis CI

teoparvanov:
Disabling the Conan cache resulted in a significant slow-down of the build process. In this PR we attempt to re-enable it.


teoparvanov:
dropping this, because caching isn't working properly.

PR #423, opened 2020-03-03
Skipping n=4 for Apollo tests

teoparvanov:
A recent combination of growing build times and increasing number of Apollo tests has led to CI jobs exceeding 50 min of execution. In such cases Travis just interrupts the job and results in CI failure, and blocked PRs/failed master.

Until we find a solution to those Travis limitations, I suggest to only run Apollo for the following BFT configurations:
 * n=6, f=1, c=1
 * n=7, f=2, c=0

Doing so provides sufficient coverage of the BFT parameters, while significantly reducing the overall duration of CI jobs.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -51,7 +51,7 @@ def interesting_configs(selected=None):
     if selected is None:
         selected=lambda *config: True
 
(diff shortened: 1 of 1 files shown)


Liorgal28:
Maybe we should consider some flag that says if we run through Travis or locally, so when someone runs locally it will still run the tests with this configuration? and if it passes locally we can pass on this configuration on Travis.

PR #424, opened 2020-03-03
Updating building methods

yontyon:
Simplifying the build process with and without Conan


.travis/preinstall.sh
```diff
@@ -14,32 +14,9 @@ fi
 # Install Conan package manager
```
toly-kournik: remove the whole file if not in use

teoparvanov:
Last time there were Conan-related changes, they broke the Concord product build. It took @dartdart26 and myself an entire day to fix those. 

I wouldn't be in favor of the current changes, unless they are tested to not impact the product build.

install.sh
```diff
@@ -1,15 +1,97 @@
...
-sudo apt-get install -y ccache cmake clang-format python3-pip python3-setuptools libgmp3-dev parallel
-
+sudo apt-get install -y ccache cmake clang clang-format-7 python3-pip python3-setuptools libgmp3-dev g++ parallel
 python3 -m pip install --upgrade wheel
-python3 -m pip install --upgrade conan
 python3 -m pip install --upgrade trio
 
-conan profile new default --detect
-conan profile update settings.compiler.libcxx=libstdc++11 default
+if [ "$1" = "USE_CONAN" ]; then
```
andrewjstone: This doesn't look correct. Shouldn't it be `==` instead of `=` ? 
Also, shouldn't the value be either `ON` or `OFF` based on configuration. I think it's better to use `-n` and pass an unset variable when we don't wan to use conan, as this allows all the booleanesque functionality of CMAKE in it's variables (ON, OFF, TRUE, FALSE, 0, 1, etc...)

How did you test this?

.travis.yml
```diff
@@ -11,14 +9,18 @@ env:
     # We purposefully leave asserts on by excluding -NDEBUG
     - CMAKE_CXX_FLAGS="-DCMAKE_CXX_FLAGS_RELEASE=-O3 -g"
     - USE_LOG4CPP=-DUSE_LOG4CPP=TRUE
+    - USE_CONAN=-DUSE_CONAN=ON
+    - BUILD_DEPS=-DBUILD_DEPS=ON
```
andrewjstone: Unfortunately, if we are going to maintain builds with and without conan, I think we should ensure that both builds actually work and add them to the matrix. This means more jobs and longer CI times.

install.sh
```diff
@@ -1,15 +1,97 @@
...
+  echo "Using Conan build..."
+  echo "Installing Conan..."
+  python3 -m pip install --upgrade conan
+  conan profile new default --detect
+  conan profile update settings.compiler.libcxx=libstdc++11 default
```
andrewjstone: Are we sure this actually uses the compiler configured in travis?

README.md
```diff
@@ -72,6 +84,8 @@ available:
 | `BUILD_COMM_TCP_TLS`   | TRUE \| FALSE | FALSE - UDP is used |
 | `USE_LOG4CPP`          | TRUE \| FALSE | FALSE |
 | `CONCORD_LOGGER_NAME`  | STRING |"concord" |
+| `USE_CONAN`            | OFF \| ON  | ON |
```
andrewjstone: Thanks for updating this table :)

install.sh
```diff
@@ -1,15 +1,97 @@
...
+
+echo "Installing Cryptopp"
+cd /tmp
+git clone https://github.com/weidai11/cryptopp.git
+cd cryptopp
+git checkout CRYPTOPP_8_2_0
+make
+sudo make install
+
+echo "Installing rocksdb"
```
andrewjstone: This is a minor comment, don't feel like you need to change anything for it in this PR. 

The conditional around installing rocksdb and log4cplus in the travis preinstall script existed to speed up builds not using them. Now I think we should always use log4cplus, but building rocksdb does take a while, so we can speed up our travis jobs by not building it when not configured. Maybe it doesn't matter due to caching. Not sure... Either way, we may want to keep an eye on build times to see if they increase drastically without the conditional.

kvbc/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.2)
 project(libkvbc VERSION 0.1.0.0 LANGUAGES CXX)
 
-find_package(Boost ${MIN_BOOST_VERSION})
+find_package(Boost ${MIN_BOOST_VERSION} REQUIRED)
```
andrewjstone: I don't think boost actually is required for KVBC. We might be able to just remove this line altogether.

conanfile.txt
```diff
@@ -48,5 +48,4 @@ boost:without_context=True
 [imports]
 bin, *.dll -> ./lib # Copies all dll files from packages bin folder to my "lib" folder
 lib, *.dylib* -> ./lib # Copies all dylib files from packages lib folder to my "lib" folder
-lib, *.so* -> ./lib # Copies all dylib files from packages lib folder to my "lib" folder
-lib, *.a* -> ./lib
+lib, *.so* -> ./lib # Copies all dylib files from packages lib folder to my "lib" folder
```
andrewjstone: Why are we no longer copying the archive files? I believe we do static linking in some places. Although, I guess if this works it doesn't matter.

.travis/preinstall.sh
yontyon: Done

install.sh
yontyon: Fixed

install.sh
yontyon: I think so.
WIth clang Conan we also uses libstdc++11 (see [https://docs.conan.io/en/latest/reference/config_files/settings.yml.html](https://docs.conan.io/en/latest/reference/config_files/settings.yml.html)).

install.sh
yontyon: FIxed

conanfile.txt
yontyon: The static libraries are found and taken from their original location. So we don't actually need to copy then to the lib folder (which is meant for the dynamic libraries only)

kvbc/CMakeLists.txt
yontyon: Done

.travis.yml
yontyon: I also don't think that we need to maintain two build methods.
I wanted to simplify our former build method to a case in which we have troubles with Conan (as in the last week) and of course I wouldn't want to make the build process longer.

yontyon:
@andrewjstone @teoparvanov 
The integration with concord works fine [see this MR](https://gitlab.eng.vmware.com/blockchain/vmwathena_blockchain/merge_requests/1380)

install.sh
andrewjstone: Yep. It does. I checked it in the travis logs for a successful build on Friday.

lazaridiscom:
Wouldn't it be preferable to 
* fix the conan problems, or
* choose another package-manager which is without problems?

Seeing MS positive open-source efforts, possibly `vcpkg` would be the way to go. Another solution would be to use `hunter`.

* https://github.com/Microsoft/vcpkg
* https://github.com/cpp-pm/hunter

CMakeLists.txt
```diff
@@ -103,6 +96,27 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     string(APPEND CMAKE_CXX_FLAGS " -fmax-errors=3")
 endif()
 
+if (USE_CONAN)
+    message("Using conan package manager")
+    if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
+        message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+        file(DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/master/conan.cmake"
```
andrewjstone: I'm not totally opposed to this, but I think it may be better to copy the cmake file and ship it with our code, as we do with other cmake files. The rationale is twofold:
 1. We won't get updates if the file already exists
 2. An update of the remote file can break our build if the user doesn't already have it.

README.md
```diff
@@ -72,9 +70,10 @@ available:
 | `BUILD_COMM_TCP_TLS`   | TRUE \| FALSE | FALSE - UDP is used |
 | `USE_LOG4CPP`          | TRUE \| FALSE | FALSE |
 | `CONCORD_LOGGER_NAME`  | STRING |"concord" |
+| `USE_CONAN`            | OFF\| ON | ON |
 
- Note: You can't set both `BUILD_COMM_TCP_PLAIN` and `BUILD_COMM_TCP_TLS` to TRUE.
- 
+ Note(1): You can't set both `BUILD_COMM_TCP_PLAIN` and `BUILD_COMM_TCP_TLS` to TRUE.
+ Note(2): In case of setting `USE_CONAN=OFF` cmake will search for packages in their native location (`/usr/local`).
```
andrewjstone: Nit: I wouldn't specify the install location here. It can change, and users can have all sorts of wacky configurations on their systems. They may set PREFIX= to indicate other places, and we don't want to have to update this README to reflect changes in future supported platforms.

andrewjstone:
> Wouldn't it be preferable to
> 
> * fix the conan problems, or
> * choose another package-manager which is without problems?
> 
> Seeing MS positive open-source efforts, possibly `vcpkg` would be the way to go. Another solution would be to use `hunter`.
> 
> * https://github.com/Microsoft/vcpkg
> * https://github.com/cpp-pm/hunter

You're absolutely right @lazaridiscom. A few of us had the same thought. We've discussed this and the actual problems with conan and realized they are such that any package manager would have these problems:
1. Temporary unavailability of the repository
2. Complexity around rewriting community packages when they break (i.e. learning the package manager in detail)

This PR has been subsequently modified to always use conan for installing dependencies, unless the user decides to install them manually. We won't provide a second automated method. Note that we also don't expect any users to do a manual install of dependencies, but it's how our current closed source product works at the moment, as that code was written before we started using conan and scripting install in the first place. We will change that in due time, but we still *may* leave a flag allowing disabling conan for user flexibility. The key here is that we no longer have two automated installation methods and will only support conan.

CMakeLists.txt
yontyon: done

README.md
yontyon: done

CMakeLists.txt
andrewjstone: Since we no longer need to download, and conan.cmake will always exist, can you please remove this conditional?

PR #425, opened 2020-03-03
Hardning pre-processing

yuliasherman:
Launch request without pre-processing in case the pre-processing consensus not reached.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -167,7 +167,7 @@ void PreProcessor::onMessage<ClientPreProcessRequestMsg>(ClientPreProcessRequest
   const ReqId &seqNumberOfLastReply = myReplica_.seqNumberOfLastReplyToClient(clientId);
   if (seqNumberOfLastReply < reqSeqNum) {
     if (myReplica_.isCurrentPrimary()) {
-      return handleClientPreProcessRequest(clientPreProcessReqMsg);
+      return handleClientPreProcessRequest(move(clientPreProcessReqMsg));
     } else {  // Not the current primary => pass it to the primary
       sendMsg(msg->body(), myReplica_.currentPrimary(), msg->type(), msg->size());
       LOG_INFO(GL, "Sending ClientPreProcessRequestMsg reqSeqNum=" << reqSeqNum << " to current primary");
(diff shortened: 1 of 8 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -137,18 +136,18 @@ template <>
...
+  const NodeIdType &clientId = msg->clientProxyId();
+  const ReqId &reqSeqNum = msg->requestSeqNum();
   LOG_INFO(GL,
            "Received ClientPreProcessRequestMsg reqSeqNum=" << reqSeqNum << " from clientId=" << clientId
                                                             << ", senderId=" << senderId);
-  if (!checkClientMsgCorrectness(clientPreProcessReqMsg, reqSeqNum)) {
+  if (!checkClientMsgCorrectness(msg, reqSeqNum)) {
     preProcessorMetrics_.requestIgnored.Get().Inc();
-    return;
+    return delete msg;
```
robem: Instead of using `delete`, can we make `msg` a smart pointer? This way we would be more resilient to future changes.

bftengine/src/preprocessor/RequestProcessingInfo.cpp
```diff
@@ -85,4 +85,9 @@ unique_ptr<MessageBase> RequestProcessingInfo::convertClientPreProcessToClientMs
   return clientPreProcessReqMsg_->convertToClientRequestMsg(resetPreProcessFlag);
 }
 
+void RequestProcessingInfo::freeClientMsg() {
+  delete clientPreProcessReqMsg_;
+  clientPreProcessReqMsg_ = nullptr;
```
robem: Similar to my comment above, can we make `clientPreProcessReqMsg_` a smart pointer instead of doing the free ourselves?

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: No. It was a smart pointer before but in order to be able to control this pointer time life cycle in a smart way without unneeded data copies, I need to have it raw. I've added an explaining comment in the code.

bftengine/src/preprocessor/RequestProcessingInfo.cpp
yuliasherman: As above.

PR #428, opened 2020-03-04
Long run tests: add Test cases into infinite loop

Liorgal28:
1)  **Make verify_linearizability decorator optional**- a tracker can be passed as input to test cases and then the tracking remains update between each test case.
after each test case, we still check that we don't have missing blocks.

2)  **Tests that not supported linearizability**- I've made those tests to support linearizability.

3)  **creates a method that goes through all the replicas and checks if they're running** - after each test case we check this because some tests are stop replicas and no starting them again.

4)  insert several test cases into the long run loop.

### DIFF
--- a/tests/apollo/test_skvbc_long_running.py
+++ b/tests/apollo/test_skvbc_long_running.py
@@ -13,9 +13,15 @@
 import os.path
 import unittest
 import trio
+import subprocess
 
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
+from util.skvbc_history_tracker import verify_linearizability
+
 from test_skvbc import SkvbcTest
+from test_skvbc_fast_path import SkvbcFastPathTest
+from test_skvbc_linearizability import SkvbcChaosTest
+from test_skvbc_view_change import SkvbcViewChangeTest
 
 # Time consts
 EIGHT_HOURS_IN_SECONDS = 8 * 60 * 60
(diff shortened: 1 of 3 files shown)


tests/apollo/test_skvbc_long_running.py
```diff
@@ -44,9 +50,39 @@ class SkvbcLongRunningTest(unittest.TestCase):
...
+    async def test_stability(self, bft_network, tracker):
         bft_network.start_all_replicas()
-        with trio.move_on_after(seconds=ONE_HOUR_IN_SECONDS):
+        with trio.move_on_after(seconds=4*ONE_HOUR_IN_SECONDS):
             while True:
-                await SkvbcTest().test_get_block_data(bft_network=bft_network, already_in_trio=True)
-                trio.sleep(seconds=10)
+                await SkvbcTest().test_get_block_data\
+                    (bft_network=bft_network, already_in_trio=True, tracker=tracker)
+                self.start_all_replicas(bft_network)
```
teoparvanov: I don't understand. Who stops the replicas, so that you need to start them?
Again, the purpose of the long run tests is the replicas are never stopped...

tests/apollo/test_skvbc_long_running.py
```diff
@@ -44,9 +50,39 @@ class SkvbcLongRunningTest(unittest.TestCase):
...
+        Start a replica if it isn't already started.
+        Otherwise raise an AlreadyStoppedError.
+        """
+        if replica_id in bft_network.procs:
+            return
+        cmd = bft_network.start_replica_cmd(replica_id)
+        bft_network.procs[replica_id] = subprocess.Popen(cmd, close_fds=True)
+
+    def start_all_replicas(self, bft_network):
+        [self.start_replica(bft_network, i) for i in range(0, bft_network.config.n)]
```
teoparvanov: why is this duplication necessary? You have the same methods in bft.py...

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -33,12 +33,17 @@ def verify_linearizability(async_fn):
...
-        bft_network = kwargs['bft_network']
-        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
-        init_state = skvbc.initial_state()
-        tracker = SkvbcTracker(init_state, skvbc, bft_network)
-        await async_fn(*args, **kwargs, tracker=tracker)
-        await tracker.fill_missing_blocks_and_verify()
+        if 'tracker' in kwargs:
+            tracker = kwargs.pop('tracker')
+            await async_fn(*args, **kwargs, tracker=tracker)
+            await tracker.fill_missing_blocks_and_verify()
```
teoparvanov: this means the verification is done on every test execution...

PR #429, opened 2020-03-05
Apollo support for an external BFT network

teoparvanov:
This PR prepares the existing Apollo tests to be run as part of the Concord product CI.

The most important change here is adding support for an externally provided BFT network. Instead of initializing the BftTestNetwork via its initializer method, we introduce two factory class-methods:
* new() - creates a new BFT network locally
* existing() - the BftTestNetwork object acts as a "proxy" for managing the externally provided BFT network.

In the existing() case we also provide the ability to inject externally prepared client instances.

The start_replica() and stop_replica() methods now both support running externally provided commands (useful for managing replicas that run as Docker containers, for example).

Some additional enhancements are required in order to run Apollo as part of product CI (which uses Pytest):
* apollo/util becomes a package by now having an initialization script \_\_init__.py
* added \_\_test__ = False to test scenarios relevant for product CI (so that PyTest does not directly invoke Apollo tests)

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -41,6 +41,7 @@ TestConfig = namedtuple('TestConfig', [
     'num_clients',
     'key_file_prefix',
     'start_replica_cmd',
+    'stop_replica_cmd',
     'num_ro_replicas'
 ])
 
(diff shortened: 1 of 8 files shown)


tests/apollo/util/bft.py
```diff
@@ -256,21 +295,37 @@ def start_replica(self, replica_id):
...
-        cmd = self.start_replica_cmd(replica_id)
-        self.procs[replica_id] = subprocess.Popen(cmd, close_fds=True)
 
-    def stop_replica(self, replica):
+        if self.is_existing:
+            subprocess.run(
+                self.start_replica_cmd(replica_id),
+                check=True
+            )
+            self.procs[replica_id] = self.replicas[replica_id]
```
andrewjstone: This is a bit confusing. It seems to imply that for existing replicas, the start_replica_cmd will automatically (and implicitly) fill in `self.replicas`.  I find this non-explicit behavior to be somewhat confusing. I'd prefer if instead we did the same thing we do for local procs and start the replica and explicitly get the value from start_replica_cmd. If that's not possible, because the replica is not actually the returned proc, which can certainly be possible since it's not local, I think we should instead have some wrapper function that gets run instead of calling `subprocess.run`. 

Take this suggestion with a grain of salt, as it may end up being more convoluted than what's going on, but at a minimum a comment would help. 

I bring all this up because I try to live by the mantra "Explicit is better than Implicit" when it comes to code.

teoparvanov: Thanks Andrew, I agree with you and will try to improve this. The tricky part is that in case of "external" replicas we don't have a direct representation of the process/container/whatever runtime. 
So in such cases self.procs[replica_id] is simply the Replica instance (effectively a stub representing the external runtime). 

I'll update the PR shortly.

andrewjstone: Thanks Teo! You rock!

PR #431, opened 2020-03-06
Increase number of clients to speed up tests

HristoStaykov:
Increase number of clients to speed up tests
by making more parallel requests.


HristoStaykov:
Tests fail, so further investigation is needed.

PR #432, opened 2020-03-06
Fix Preprocessor build

robem:
With a fresh dev environment I noticed that the build was failing due to the
preprocessor couldn't find openssl/evp.h.


yuliasherman:
Why Travis does not detect such issues?

PR #442, opened 2020-03-11
Fix unexpected pre-processing order

yuliasherman:
Fix the case when non-primary replicas complete pre-processing before the primary:

1. Store PreProcessRequest ing RequestProcessingInfo during RequestProcessingInfo constructing.
2. In case all non-primary replicas returned with the reply, verify that the primary has completed the pre-execution. If not, wait for it.
3. Retry primary pre-execution only after its hash has been calculated.

https://jira.eng.vmware.com/browse/BC-1785

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -291,15 +291,16 @@ void PreProcessor::finalizePreProcessing(NodeIdType clientId, SeqNum reqSeqNum) 
 
 uint16_t PreProcessor::numOfRequiredReplies() { return myReplica_.getReplicaConfig().fVal + 1; }
 
-void PreProcessor::registerClientPreProcessRequest(ClientPreProcessReqMsgUniquePtr clientReqMsg) {
+void PreProcessor::registerRequest(ClientPreProcessReqMsgUniquePtr clientReqMsg,
+                                   PreProcessRequestMsgSharedPtr preProcessRequestMsg) {
   const uint16_t &clientId = clientReqMsg->clientProxyId();
   const ReqId &requestSeqNum = clientReqMsg->requestSeqNum();
   {
     // Only one request is supported per client for now
     auto &clientEntry = ongoingRequests_[clientId];
     lock_guard<mutex> lock(clientEntry->mutex);
-    clientEntry->clientReqInfoPtr =
-        make_unique<RequestProcessingInfo>(numOfReplicas_, numOfRequiredReplies(), requestSeqNum, move(clientReqMsg));
+    clientEntry->clientReqInfoPtr = make_unique<RequestProcessingInfo>(
+        numOfReplicas_, numOfRequiredReplies(), requestSeqNum, move(clientReqMsg), preProcessRequestMsg);
   }
   LOG_DEBUG(GL, "clientId=" << clientId << " requestSeqNum=" << requestSeqNum << " registered");
 }
@@ -328,7 +329,7 @@ void PreProcessor::handleClientPreProcessRequest(ClientPreProcessReqMsgUniquePtr
                                                                                          clientReqMsg->requestLength(),
                                                                                          clientReqMsg->requestBuf(),
                                                                                          clientReqMsg->getCid());
-  registerClientPreProcessRequest(move(clientReqMsg));
+  registerRequest(move(clientReqMsg), preProcessRequestMsg);
   sendPreProcessRequestToAllReplicas(preProcessRequestMsg);
 
   // Pre-process the request and calculate a hash of the result
@@ -391,8 +392,7 @@ void PreProcessor::handlePreProcessedReqByPrimary(PreProcessRequestMsgSharedPtr 
   auto &clientEntry = ongoingRequests_[clientId];
   lock_guard<mutex> lock(clientEntry->mutex);
   if (clientEntry->clientReqInfoPtr)
-    clientEntry->clientReqInfoPtr->handlePrimaryPreProcessed(
-        preProcessReqMsg, getPreProcessResultBuffer(clientId), resultBufLen);
+    clientEntry->clientReqInfoPtr->handlePrimaryPreProcessed(getPreProcessResultBuffer(clientId), resultBufLen);
 }
 
 void PreProcessor::handlePreProcessedReqByNonPrimary(uint16_t clientId, ReqId reqSeqNum, uint32_t resBufLen) {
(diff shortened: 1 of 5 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -372,14 +383,14 @@ uint32_t PreProcessor::launchReqPreProcessing(uint16_t clientId, ReqId reqSeqNum
   return resultLen;
 }
 
+PreProcessingResult PreProcessor::getPreProcessingConsensusResult(uint16_t clientId) {
+  auto &clientEntry = ongoingRequests_[clientId];
+  lock_guard<mutex> lock(clientEntry->mutex);
+  return clientEntry->clientReqInfoPtr->getPreProcessingConsensusResult();
+}
```
teoparvanov: it's kind of confusing to have two methods called the same - getPreProcessingConsensusResult:
```
PreProcessingResult PreProcessor::getPreProcessingConsensusResult(uint16_t clientId) {
  auto &clientEntry = ongoingRequests_[clientId];
  return clientEntry->clientReqInfoPtr->getPreProcessingConsensusResult();
}
```

yuliasherman: The first one is just a proxy to another.

bftengine/src/preprocessor/tests/preprocessor_test.cpp
```diff
@@ -177,67 +178,97 @@ void setUpConfiguration() {
...
+  memset(buf, '5', bufLen);
+  for (auto i = 1; i <= numOfRequiredReplies; i++) {
+    if (i != numOfReplicas - 1) Assert(reqInfo.getPreProcessingConsensusResult() == CONTINUE);
+    memset(buf, '4', bufLen);
+    reqInfo.handlePreProcessReplyMsg(preProcessNonPrimary(i, repInfo));
+  }
+  Assert(reqInfo.getPreProcessingConsensusResult() == CONTINUE);
+  memset(buf, '4', bufLen);
+  reqInfo.handlePrimaryPreProcessed(buf, bufLen);
   Assert(reqInfo.getPreProcessingConsensusResult() == COMPLETE);
```
teoparvanov: given the test title "primaryReplicaDidNotCompletePreProcessingWhileNonPrimariesDid", does it make sense to have:
`reqInfo.getPreProcessingConsensusResult() == COMPLETE`

yuliasherman: That's all the idea. Until primary did not complete pre-execution we cannot complete request handling even with f+1 replies collected. We need to verify that the primary pre-execution result is the same that passed the pre-execution consensus, otherwise, we need to retry primary pre-execution.

bftengine/src/preprocessor/RequestProcessingInfo.cpp
```diff
@@ -65,20 +65,26 @@ PreProcessingResult RequestProcessingInfo::getPreProcessingConsensusResult() con
   if (maxNumOfEqualHashes >= numOfRequiredEqualReplies_) {
     if (itOfChosenHash->first == myPreProcessResultHash_)
       return COMPLETE;  // Pre-execution consensus reached
-    else {
-      // Primary replica calculated hash is different from hash that passed consensus => we don't have correct
-      // pre-processed results. The request could be cancelled, retried or executed without pre-processing.
+    else if (myPreProcessResultLen_ != 0) {
+      // Primary replica calculated hash is different from a hash that passed pre-execution consensus => we don't have
+      // correct pre-processed results. Let's launch a pre-processing retry.
```
teoparvanov: I cannot understand the logical mapping between the `if (myPreProcessResultLen_ != 0)` and the comment below...

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -68,7 +68,8 @@ class PreProcessor {
   template <typename T>
   void onMessage(T *msg);
 
-  void registerClientPreProcessRequest(ClientPreProcessReqMsgUniquePtr clientReqMsg);
+  void registerRequest(ClientPreProcessReqMsgUniquePtr clientReqMsg,
+                       PreProcessRequestMsgSharedPtr preProcessRequestMsg);
```
teoparvanov: Yulia, what's the difference between `clientReqMsg` and `preProcessRequestMsg` ?

yuliasherman: PreProcessRequestMsg is what we send to non-primary replicas to ask for pre-execution.

bftengine/src/preprocessor/tests/preprocessor_test.cpp
teoparvanov: It's probably just a naming problem from my perspective. Actually, what's the difference between COMPLETE and CONTINUE?

yuliasherman: Before the fix, I did not verify that the primary complete the pre-execution and this was treated as the primary generated a different pre-execution results hash. The function returned with RETRY_PRIMARY status and then crashed as PreProcessRequestMsg was not been saved yet in RequestProcessingInfo and was NULL. Complicated.

teoparvanov: I see. Which unit test checks this doesn't happen now?

bftengine/src/preprocessor/RequestProcessingInfo.cpp
yuliasherman: Once the primary pre-executed the request, it calculates its hash and saves it in myPreProcessResultHash_. 
myPreProcessResultLen_ stores myPreProcessResultHash_ length.

teoparvanov: So, it's either 0 or 1 ?

bftengine/src/preprocessor/tests/preprocessor_test.cpp
yuliasherman: primaryReplicaDidNotCompletePreProcessingWhileNonPrimariesDid

bftengine/src/preprocessor/RequestProcessingInfo.cpp
yuliasherman: In fact, it contains a size of myPreProcessResultHash_.

teoparvanov: and what and how many elements can be contained in `myPreProcessResultHash_`?

yuliasherman: What elements? RequestProcessingInfo object is created per request.

teoparvanov: Ahhh, never mind, myPreProcessResultHash_ is a single hash :) Got it now. We only retry if the myPreProcessResultHash_ is non-empty.

yuliasherman: Exactly!

PR #444, opened 2020-03-11
Ro replica object store integration

toly-kournik:
Introducing S3-compatible external object store support.
Co-authored-by: @salieri11 
Co-authored-by: @toly-kournik 
Key points:
 S3:
- concord::storage::s3::client with S3KeyGenerator
- handing off state transfer messages in BCStateTransfer if used with remote object store

General:
- kvbc::ReplicaImp implements IAppState
- removed ReplicaImp::StorageWrapperForIdleMode

### DIFF
--- a/.conan/libs3_pkg/conanfile.py
+++ b/.conan/libs3_pkg/conanfile.py
@@ -1,0 +1,16 @@
+from conans import ConanFile, CMake, tools
+from conans.tools import os_info, SystemPackageTool
+
+
+class LibS3Conan(ConanFile):
+    name = "libs3-dev"
+    version = "2.0-3"
+    def system_requirements(self):
+        pack_name = None
+        if os_info.linux_distro == "ubuntu":
+            pack_name = "libs3-dev"
+ 
+        if pack_name:
+            installer = SystemPackageTool()
+            installer.install(pack_name)
+
--- a/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
+++ b/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
@@ -69,7 +69,7 @@ class IAppState {
 
   // returns true IFF block blockId exists
   // (i.e. the block is stored in the application/storage layer).
-  virtual bool hasBlock(uint64_t blockId) = 0;
+  virtual bool hasBlock(uint64_t blockId) const = 0;
 
   // If block blockId exists, then its content is returned via the arguments
   // outBlock and outBlockSize. Returns true IFF block blockId exists.
@@ -83,25 +83,19 @@ class IAppState {
   // blockId   - the block number
   // block     - pointer to a buffer that contains the new block
   // blockSize - the size of the new block
-  virtual bool putBlock(uint64_t blockId, char *block, uint32_t blockSize) = 0;
+  virtual bool putBlock(const uint64_t blockId, const char *block, const uint32_t blockSize) = 0;
 
   // returns the maximal block number n such that all blocks 1 <= i <= n exist.
   // if block 1 does not exist, returns 0.
-  virtual uint64_t getLastReachableBlockNum() = 0;
-
+  virtual uint64_t getLastReachableBlockNum() const = 0;
   // returns the maximum block number that is currently stored in
   // the application/storage layer.
-  virtual uint64_t getLastBlockNum() = 0;
+  virtual uint64_t getLastBlockNum() const = 0;
 
   // When the state is updated by the application, getLastReachableBlockNum()
   // and getLastBlockNum() should always return the same block number.
   // When that state transfer module is updating the state, then these methods
   // may return different block numbers.
-
-  // This method MAY be called by ST module when putBlock returns false. The call will block until the underlying
-  // storage becomes available (storage non availabilty is the reason for putBlock to fail). Mainly relevant for network
-  // storages. In regular replica which uses local DB this function should not be called.
-  virtual void wait() = 0;
 };
 
 struct Config {
@@ -110,6 +104,7 @@ struct Config {
   uint16_t cVal = 0;
   uint16_t numReplicas = 0;
   bool pedanticChecks = false;
+  bool isReadOnly = false;
 
 #if defined USE_COMM_PLAIN_TCP || defined USE_COMM_TLS_TCP
   uint32_t maxChunkSize = 16384;
--- a/bftengine/src/bcstatetransfer/BCStateTran.hpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.hpp
@@ -11,8 +11,7 @@
 // terms and conditions of the subcomponent's license, as noted in the LICENSE
 // file.
 
-#ifndef BFTENGINE_SRC_BCSTATETRANSFER_BCSTATETRAN_HPP_
-#define BFTENGINE_SRC_BCSTATETRANSFER_BCSTATETRAN_HPP_
+#pragma once
 
 #include <set>
 #include <map>
@@ -37,10 +36,11 @@
 using std::set;
 using std::map;
 using std::string;
+using concordMetrics::StatusHandle;
+using concordMetrics::GaugeHandle;
+using concordMetrics::CounterHandle;
 
-namespace bftEngine {
-namespace SimpleBlockchainStateTransfer {
-namespace impl {
+namespace bftEngine::SimpleBlockchainStateTransfer::impl {
 
 class BCStateTran : public IStateTransfer {
  public:
@@ -82,7 +82,11 @@ class BCStateTran : public IStateTransfer {
   void handleStateTransferMessage(char* msg, uint32_t msgLen, uint16_t senderId) override;
 
  protected:
-  const bool pedanticChecks_;
+  std::function<void(char*, uint32_t, uint16_t)> messageHandler_;
+  // actual handling function. can be used in context of dedicated thread
+  void handleStateTransferMessageImp(char* msg, uint32_t msgLen, uint16_t senderId);
+  // handling from other context
+  void handoff(char* msg, uint32_t msgLen, uint16_t senderId);
 
   ///////////////////////////////////////////////////////////////////////////
   // Constants
@@ -103,21 +107,8 @@ class BCStateTran : public IStateTransfer {
   ///////////////////////////////////////////////////////////////////////////
   // Management and general data
   ///////////////////////////////////////////////////////////////////////////
-
+  const Config config_;
   const set<uint16_t> replicas_;
-  const uint16_t myId_;
-  const uint16_t fVal_;
-  const uint32_t maxBlockSize_;
-  const uint32_t maxChunkSize_;
-  const uint16_t maxNumberOfChunksInBatch_;
-  const uint32_t maxPendingDataFromSourceReplica_;
-
-  const uint32_t maxNumOfReservedPages_;
-  const uint32_t sizeOfReservedPage_;
-  const uint32_t refreshTimerMilli_;
-  const uint32_t checkpointSummariesRetransmissionTimeoutMilli_;
-  const uint32_t maxAcceptableMsgDelayMilli_;
-
   const uint32_t maxVBlockSize_;
   const uint32_t maxItemSize_;
   const uint32_t maxNumOfChunksInAppBlock_;
@@ -328,11 +319,6 @@ class BCStateTran : public IStateTransfer {
   std::chrono::seconds last_metrics_dump_time_;
   std::chrono::seconds metrics_dump_interval_in_sec_;
   concordMetrics::Component metrics_component_;
-
-  typedef concordMetrics::Component::Handle<concordMetrics::Gauge> GaugeHandle;
-  typedef concordMetrics::Component::Handle<concordMetrics::Status> StatusHandle;
-  typedef concordMetrics::Component::Handle<concordMetrics::Counter> CounterHandle;
-
   struct Metrics {
     StatusHandle fetching_state_;
     StatusHandle pedantic_checks_enabled_;
@@ -395,8 +381,4 @@ class BCStateTran : public IStateTransfer {
   mutable Metrics metrics_;
 };
 
-}  // namespace impl
-}  // namespace SimpleBlockchainStateTransfer
-}  // namespace bftEngine
-
-#endif  // BFTENGINE_SRC_BCSTATETRANSFER_BCSTATETRAN_HPP_
+}  // namespace bftEngine::SimpleBlockchainStateTransfer::impl
--- a/kvbc/include/ReplicaImp.h
+++ b/kvbc/include/ReplicaImp.h
@@ -25,52 +25,47 @@
 
 namespace concord::kvbc {
 
-class ReplicaInitException : public std::exception {
+class ReplicaImp : public IReplica,
+                   public ILocalKeyValueStorageReadOnly,
+                   public IBlocksAppender,
+                   public bftEngine::SimpleBlockchainStateTransfer::IAppState {
  public:
-  explicit ReplicaInitException(const std::string &what) : msg(what){};
-
-  virtual const char *what() const noexcept override { return msg.c_str(); }
-
- private:
-  std::string msg;
-};
-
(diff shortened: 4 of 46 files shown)


bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
```diff
@@ -92,11 +93,6 @@ class IAppState {
   // and getLastBlockNum() should always return the same block number.
   // When that state transfer module is updating the state, then these methods
   // may return different block numbers.
-
```
salieri11: No more wait() method?

kvbc/include/ReplicaImp.h
```diff
@@ -232,26 +241,24 @@ class ReplicaImp : public IReplica,
...
+    bool getBlock(uint64_t blockId, char *outBlock, uint32_t *outBlockSize) override;
+    bool getPrevDigestFromBlock(
         uint64_t blockId, bftEngine::SimpleBlockchainStateTransfer::StateTransferDigest *outPrevBlockDigest) override;
-    virtual bool putBlock(uint64_t blockId, char *block, uint32_t blockSize) override;
-    virtual uint64_t getLastReachableBlockNum() override;
-    virtual uint64_t getLastBlockNum() override;
-    virtual void wait() override;
+    bool putBlock(uint64_t blockId, char *block, uint32_t blockSize) override;
+    uint64_t getLastReachableBlockNum() override;
+    void setLastReachableBlockNum(uint64_t blockNum) override { m_lastReachableBlock = blockNum; }
```
salieri11: Is it accessible by one thread?

bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
toly-kournik: No, in order to keep replica responsive even when the object store is temporarily inaccessible, we handoff state transfer messages to another dedicated thread and it waits synchronously for the object store

toly-kournik: It works only in RO replica, which doesn't maintain its own state. So there's no thread safety problems

kvbc/include/ReplicaImp.h
toly-kournik: Didn't get the question.

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -35,10 +34,11 @@
 using std::set;
 using std::map;
 using std::string;
+using concordMetrics::StatusHandle;
+using concordMetrics::GaugeHandle;
+using concordMetrics::CounterHandle;
```
dartdart26: [Minor]: I think it may not be a good idea to put these in header files. Can we put them in the cpp file instead and use the full name in the header?

.conan/libs3_pkg/conanfile.py
```diff
@@ -0,0 +1,15 @@
...
+    name = "libs3-dev"
+    version = "2.0-3"
+    def system_requirements(self):
+        pack_name = None
+        if os_info.linux_distro == "ubuntu":
+            pack_name = "libs3-dev"
+ 
+        if pack_name:
+            installer = SystemPackageTool()
+            installer.install(pack_name)
```
dartdart26: [Minor]: Missing empty line.

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -35,10 +34,11 @@
...
 
-namespace bftEngine {
-namespace SimpleBlockchainStateTransfer {
-namespace impl {
+namespace bftEngine::SimpleBlockchainStateTransfer::impl {
```
dartdart26: Nice! :)

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -80,7 +80,11 @@ class BCStateTran : public IStateTransfer {
   void handleStateTransferMessage(char* msg, uint32_t msgLen, uint16_t senderId) override;
 
  protected:
-  const bool pedanticChecks_;
+  std::function<void(char*, uint32_t, uint16_t)> messageHandler_;
+  // actual handling function. can be used in context of dedicated thread
+  void handleStateTransferMessageImp(char* msg, uint32_t msgLen, uint16_t senderId);
+  // handling from other context
+  void handoff(char* msg, uint32_t msgLen, uint16_t senderId);
```
dartdart26: Can we make these functions accept a `const void*` buffer instead?

bftengine/src/bcstatetransfer/BCStateTran.hpp
toly-kournik: Is it kinda good practice?

bftengine/src/bcstatetransfer/BCStateTran.hpp
toly-kournik: Basically yes, but we'll need to replace all classes implementing `IStateTransfer::handleStateTransferMessage`
If you think it's essential, let's do it as a separate effort.

.conan/libs3_pkg/conanfile.py
toly-kournik: will add

bftengine/src/bcstatetransfer/BCStateTran.hpp
toly-kournik: so why not to replace all `using` declarations above?

dartdart26: I think we should, especially the `std` ones. I think we may do it in a separate PR.

bftengine/src/bcstatetransfer/BCStateTran.hpp
dartdart26: Sure, separate one is fine.

kvbc/include/ror_app_state.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include "blockchain/block.h"
+#include "blockchain/db_adapter.h"
+#include "kv_types.hpp"
+#include "sliver.hpp"
+#include "status.hpp"
+#include "storage/db_interface.h"
+#include "object_store/object_store_client.hpp"
+
+const std::string kLastBlockIdKey = "1";
+const std::string kLastReachableBlockIdKey = "2";
```
dartdart26: Can we declare these as either `static const...` or better yet `inline const...`?

kvbc/src/ror_app_state.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <cstring>
+#include <unordered_map>
+#include <vector>
+#include "hash_defs.h"
+
+using namespace concordUtils;
+using namespace bftEngine::SimpleBlockchainStateTransfer;
+using namespace concord::storage;
+
+class RoRAppStateImpl : public IAppState {
```
dartdart26: I don't think we need the implementation derive from `IAppState`. It is an implementation detail that is defined in the cpp file and it can just have the same (or similar) interface, without inheriting from `IAppState`.

kvbc/include/ror_app_state.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace concord {
+namespace consensus {
+
+class RoRAppState : public bftEngine::SimpleBlockchainStateTransfer::IAppState {
+ public:
+  ~RoRAppState();
```
dartdart26: Can we make it `=default`?

kvbc/src/ror_app_state.cpp
toly-kournik: I'm trying to remove the RoRAppstate at all right now

kvbc/src/ror_app_state.cpp
```diff
@@ -0,0 +1,160 @@
...
+  void setLastReachableBlockNum(uint64_t blockNum) override { lastReachableBlockId_ = blockNum; }
+
+  uint64_t getLastBlockNum() override { return lastBlockId_; }
+
+ private:
+  uint64_t lastBlockId_ = 0;
+  uint64_t lastReachableBlockId_ = 0;
+
+  Sliver lastBlockIdKey = Sliver::copy(kLastBlockIdKey.c_str(), kLastBlockIdKey.size());
+  Sliver lastReachableBlockIdKey = Sliver::copy(kLastReachableBlockIdKey.c_str(), kLastReachableBlockIdKey.size());
```
dartdart26: Can these be made `static`? Rationale is that they have the same value for all objects.

dartdart26: I guess it doesn't really matter much as there won't be many `RoRAppStateImpl` objects.

util/include/Handoff.hpp
```diff
@@ -0,0 +1,96 @@
+// Concord
+//
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
```
dartdart26: [Minor]: (c) 2020

kvbc/src/ReplicaImp.cpp
```diff
@@ -187,6 +187,16 @@ Status ReplicaImp::addBlock(const SetOfKeyValuePairs &updates, BlockId &outBlock
 
 void ReplicaImp::set_command_handler(ICommandsHandler *handler) { m_cmdHandler = handler; }
 
+void ReplicaImp::set_app_state(std::shared_ptr<bftEngine::SimpleBlockchainStateTransfer::IAppState> appState) {
```
dartdart26: Does it make sense to create another method for the case of a `nullptr` parameter being passed? Maybe something like `set_default_app_state()` or similar. If we do that, then `set_app_state()` might be just a simple setter.

storage/include/ecs/s3_client.hpp
```diff
@@ -0,0 +1,393 @@
...
+
+  /**
+   * @brief Represents intermidiate data for get operation and is passed to
+   * the lambda callback.
+   *
+   */
+  struct GetObjectResponseData : public ResponseData {
+    GetObjectResponseData(size_t linitialLength) : data(new char[linitialLength]), dataLength(linitialLength) {}
+
+    void CheckAndResize(int& nextReadLength) {
```
dartdart26: Can we remove the reference here as we are not changing the value?

storage/include/ecs/s3_client.hpp
```diff
@@ -0,0 +1,393 @@
...
+    S3Status status = S3Status::S3StatusOK;
+    string errorMessage;
+  };
+
+  /**
+   * @brief Represents intermidiate data for get operation and is passed to
+   * the lambda callback.
+   *
+   */
+  struct GetObjectResponseData : public ResponseData {
```
dartdart26: Can we use an `std::vector` or an `std::string` (utilising `reseve()`) for simplicity?

util/include/Handoff.hpp
```diff
@@ -0,0 +1,94 @@
...
+  class ThreadCanceledException : public std::runtime_error {
+   public:
+    ThreadCanceledException() : std::runtime_error("thread cancelled") {}
+    const char* what() const noexcept override { return std::runtime_error::what(); }
+  };
+
+  func_type pop() {
+    while (true) {  // for spurious wake ups
+      std::unique_lock<std::mutex> ul(queue_lock_);
+      queue_cond_.wait(ul, [this] { return !(task_queue_.empty() && !stopped_); });
```
dartdart26, line 69: [Minor]: Please correct me if I am wrong, but I think this comment is not entirely correct as spurious wake ups are handled in the predicate passed to `wait()`. It seems this loop allows us to exhaust the queue after `stopped_` has been set.

util/include/Handoff.hpp
```diff
@@ -0,0 +1,94 @@
...
+ */
+class Handoff {
+  typedef std::lock_guard<std::mutex> guard;
+  typedef std::function<void()> func_type;
+
+ public:
+  Handoff() {
+    thread_ = std::thread([this] {
+      try {
+        for (;;) pop()();
```
dartdart26, line 37: I think we should either specify `func_type` as `noexcept` or handle any exceptions, because `pop()()` calls user code that might throw.

tests/simpleKVBC/TesterReplica/setup.hpp
```diff
@@ -19,14 +19,25 @@
...
+class IDataKeyGenerator;
 
 class TestSetup {
  public:
   static std::unique_ptr<TestSetup> ParseArgs(int argc, char** argv);
 
+  std::tuple<std::shared_ptr<concord::storage::IDBClient>,
+             std::shared_ptr<concord::storage::IDBClient>,
+             IDataKeyGenerator*>
+  get_dbadapter_configuration();
```
dartdart26, line 39: [Minor]: I think a named struct would be clearer, especially with the same type (`IDBClient`) of the first two values.

storage/include/storage/db_types.h
```diff
@@ -22,7 +22,7 @@ inline namespace v1DirectKeyValue {
 namespace detail {
 
 enum class EDBKeyType : std::uint8_t {
-  E_DB_KEY_TYPE_FIRST = 1,
+  E_DB_KEY_TYPE_FIRST = 65,
```
dartdart26, line 25: Could you please explain the rationale for the value of `65`?

toly-kournik: it's in "visible" ascii range ;-)

storage/include/rocksdb/client.h
```diff
@@ -44,7 +44,7 @@ class ClientIterator : public concord::storage::IDBClient::IDBClientIterator {
   KeyValuePair next() override;
   KeyValuePair getCurrent() override;
   bool isEnd() override;
-  concordUtils::Status getStatus() override;
```
dartdart26, line 47: I think if we rely that the `concordUtils` namespace or some if its types are accessible in the current header, it hides bringing it up explicitly in some other header file.

storage/include/s3/client.hpp
```diff
@@ -0,0 +1,289 @@
...
+
+#include <libs3.h>
+#include <cstring>
+#include <thread>
+#include "Logger.hpp"
+#include "assertUtils.hpp"
+#include "storage/db_interface.h"
+
+using namespace std;
+using namespace concordUtils;
```
dartdart26, line 24: I think it might be better for users to not inject these namespaces in a header.

kvbc/include/ReplicaImp.h
```diff
@@ -118,65 +110,16 @@ class ReplicaImp : public IReplica, public ILocalKeyValueStorageReadOnly, public
     }
   };
 
-  // TODO(GG): do we want synchronization here ?
-  class StorageWrapperForIdleMode : public ILocalKeyValueStorageReadOnly {
```
dartdart26, line 122: Great, thanks for doing this!

kvbc/include/ReplicaImp.h
```diff
@@ -118,65 +110,16 @@ class ReplicaImp : public IReplica, public ILocalKeyValueStorageReadOnly, public
...
-
-    virtual Status getBlockData(BlockId blockId, concord::storage::SetOfKeyValuePairs &outBlockData) const override;
-
-    virtual Status mayHaveConflictBetween(const Sliver &key,
-                                          BlockId fromBlock,
-                                          BlockId toBlock,
-                                          bool &outRes) const override;
-  };
-
-  class BlockchainAppState : public bftEngine::SimpleBlockchainStateTransfer::IAppState {
```
dartdart26, line 142: Thanks for removing that!

dartdart26:
Thanks @toly-kournik and @salieri11 for the effort, great work! It is great that we can now save data to S3! And the code is clearer and simpler.

storage/include/s3/client.hpp
toly-kournik: done

util/include/Handoff.hpp
toly-kournik: correct. done

tests/simpleKVBC/TesterReplica/setup.hpp
toly-kournik: done

util/include/Handoff.hpp
toly-kournik: done

PR #445, opened 2020-03-11
Activate RocksDB globaly.

HristoStaykov:


teoparvanov:
Hi Andrew, thanks for pointing this out! We will add support to Apollo tests to switch between persistent/non-persistent replicas where applicable (for persistency and long-running tests we should always have "-p"). 

On the other hand, I still think CI should only run with persistent replicas.

tests/apollo/CMakeLists.txt
```diff
@@ -1,33 +1,39 @@
...
     add_test(NAME skvbc_persistence_tests COMMAND sh -c
             "python3 -m unittest test_skvbc_persistence 2>&1 > /dev/null"
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
             
-add_test(NAME skvbc_ro_replica_tests COMMAND sh -c
-        "python3 -m unittest test_skvbc_ro_replica 2>&1 > /dev/null"
-        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+    add_test(NAME skvbc_ro_replica_tests COMMAND sh -c
+            "python3 -m unittest test_skvbc_ro_replica 2>&1 > /dev/null"
+            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```
teoparvanov: please revert this - it's unrelated to the current PR

tests/apollo/CMakeLists.txt
```diff
@@ -1,33 +1,39 @@
+if (BUILD_ROCKSDB_STORAGE)
+    set(TESTS_ARGS = "-p")
+else()
+    set(TESTS_ARGS = "")
+endif()
+
 add_test(NAME skvbc_basic_tests COMMAND sh -c
-        "python3 -m unittest test_skvbc 2>&1 > /dev/null"
+        "TESTS_ARGS=${TESTS_ARGS} python3 -m unittest test_skvbc 2>&1 > /dev/null"
```
teoparvanov: I think it would be more elegant to inject BUILD_ROCKSDB_STORAGE here (and all environment variables, ideally)

tests/apollo/CMakeLists.txt
```diff
@@ -1,33 +1,39 @@
...
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_test(NAME skvbc_linearizability_tests COMMAND sudo sh -c
-        "python3 -m unittest test_skvbc_history_tracker test_skvbc_linearizability 2>&1 > /dev/null"
+        "TESTS_ARGS=${TESTS_ARGS} python3 -m unittest test_skvbc_history_tracker test_skvbc_linearizability 2>&1 > /dev/null"
```
teoparvanov: same here

tests/apollo/test_skvbc.py
```diff
@@ -27,12 +28,14 @@ def start_replica_cmd(builddir, replica_id):
...
     """
+    argsToAdd = os.environ.get('TESTS_ARGS', "")
     statusTimerMilli = "500"
     path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
     return [path,
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
-            "-s", statusTimerMilli]
+            "-s", statusTimerMilli,
+            argsToAdd]
```
teoparvanov: add "-p" here only if BUILD_ROCKSDB_STORAGE=TRUE/ON

HristoStaykov:
Still doing research how to inject all the options set in CMake.
The tests are failing because of a minor mistake (accidentally added "=" when setting the variable TESTS_ARGS).

teoparvanov:
I suggest to focus on passing BUILD_ROCKSDB_STORAGE. Passing the entire environment is not a priority right now - I just mentioned it, in case it could be done easily.

tests/apollo/CMakeLists.txt
HristoStaykov: Yes, a map of all environment variables would be better in the long run if more options appear, only the "start_replica_cmd" in the python code has to be modified.

tests/apollo/test_skvbc_linearizability.py
```diff
@@ -40,8 +41,8 @@ def start_replica_cmd(builddir, replica_id):
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
             "-s", statusTimerMilli,
-            "-v", viewChangeTimeoutMilli
-            ]
+            "-v", viewChangeTimeoutMilli,
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE') == "true" else ""]
```
teoparvanov: maybe don't check against a specific value here... Just verify if the variable is set at all, to avoid weird regressions in the future.

tests/apollo/test_skvbc_slow_path.py
```diff
@@ -34,7 +35,8 @@ def start_replica_cmd(builddir, replica_id):
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
             "-s", statusTimerMilli,
-            "-v", viewChangeTimeoutMilli]
+            "-v", viewChangeTimeoutMilli,
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE') == "true" else ""]
```
teoparvanov: same here and below

tests/apollo/test_skvbc_linearizability.py
teoparvanov: on second thought, let's check against TRUE or ON (should be case insensitive)

tests/apollo/test_skvbc.py
```diff
@@ -32,7 +33,10 @@ def start_replica_cmd(builddir, replica_id):
     return [path,
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
-            "-s", statusTimerMilli]
+            "-s", statusTimerMilli,
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower() == "true" \
+                 or os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower() == "on" \
```
teoparvanov: The following would be more elegant:
`if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower() in set(["true", "on"])`

tests/apollo/test_skvbc_auto_view_change.py
```diff
@@ -34,7 +35,10 @@ def start_replica_cmd(builddir, replica_id):
             "-i", str(replica_id),
             "-s", statusTimerMilli,
             "-v", viewChangeTimeoutMilli,
-            "-a", autoPrimaryRotationTimeoutMilli]
+            "-a", autoPrimaryRotationTimeoutMilli,
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower() == "true" \
+                 or os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower() == "on" \
```
teoparvanov: same here and below

andrewjstone:
Awesome work. Thanks!

PR #446, opened 2020-03-12
Remove the 'found' param in merkle getBlockById()

dartdart26:
Return Status::NotFound() instead of returning Status::OK() and setting
the 'found' output parameter to true. This makes the interface clearer
and easier to use.

### DIFF
--- a/storage/src/merkle_tree_db_adapter.cpp
+++ b/storage/src/merkle_tree_db_adapter.cpp
@@ -285,10 +285,9 @@ Sliver DBAdapter::createBlockNode(const SetOfKeyValuePairs &updates,
   // Make sure the digest is zero-initialized by using {} initialization.
   auto parentBlockDigest = StateTransferDigest{};
   if (blockId > 1) {
-    auto found = false;
     Sliver parentBlock;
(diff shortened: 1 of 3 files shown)


storage/src/merkle_tree_db_adapter.cpp
```diff
@@ -303,20 +302,12 @@ Sliver DBAdapter::createBlockNode(const SetOfKeyValuePairs &updates,
   return block::detail::createNode(node);
 }
 
-Status DBAdapter::getBlockById(BlockId blockId, Sliver &block, bool &found) const {
-  found = false;
+Status DBAdapter::getBlockById(BlockId blockId, Sliver &block) const {
   const auto blockKey = DBKeyManipulator::genBlockDbKey(blockId);
   Sliver blockNodeSliver;
   if (auto status = db_->get(blockKey, blockNodeSliver); status.isNotFound()) {
```
toly-kournik: I like this syntax

dartdart26:
> LGTM - much cleaner and shorter

@toly-kournik Thanks for suggesting that change!

PR #449, opened 2020-03-13
[Apollo] Explicitly defining the metrics port in the Replica object

teoparvanov:
So far, the metrics (UDP) port was assumed to be the replica's listen port, incremented by 1000. This is true for the TesterReplicas used by Apollo's default BftTestNetwork. However, the port number may follow a different pattern when running against an external (containerized) BFT network.

This PR provides the option to set an arbitrary metrics port via the Replica object.


teoparvanov:
@andrewjstone can you help me with a review on this PR?

Thanks!

PR #451, opened 2020-03-16
Long run test: add basic fast path test case

Liorgal28:
1) add a basic test case of fast path scenario - read your writes with an assertion that all the writes on the fast path.

2) adjust start_all_replicas to go through all replicas and try to start if its part of long-run testing.

I tested this long-run scenario for 2 hours on my local machine

### DIFF
--- a/tests/apollo/test_skvbc_long_running.py
+++ b/tests/apollo/test_skvbc_long_running.py
@@ -16,10 +16,7 @@ import trio
 from os import environ
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
 from test_skvbc import SkvbcTest
-from test_skvbc_auto_view_change import SkvbcAutoViewChangeTest
 from test_skvbc_fast_path import SkvbcFastPathTest
-from test_skvbc_slow_path import SkvbcSlowPathTest
-from test_skvbc_view_change import SkvbcViewChangeTest
 
 # Time consts
 EIGHT_HOURS_IN_SECONDS = 8 * 60 * 60
@@ -34,14 +31,12 @@ def start_replica_cmd(builddir, replica_id):
     """
     statusTimerMilli = "500"
     viewChangeTimeoutMilli = "10000"
-    autoPrimaryRotationTimeoutMilli = "10000"
     path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
     return [path,
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
             "-s", statusTimerMilli,
             "-v", viewChangeTimeoutMilli,
-            "-a", autoPrimaryRotationTimeoutMilli,
             "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower()
                     in set(["true", "on"])
                  else ""]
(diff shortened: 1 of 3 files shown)


tests/apollo/util/bft.py
```diff
@@ -661,3 +662,15 @@ def force_quorum_including_replica(self, replica_id, primary=0):
                 with trio.move_on_after(interval):
                     if await predicate():
                         return
+
+
+    async def num_of_slow_path(self):
```
Liorgal28: just for convenience, a method that taking the number of slow path count using the metrics.

tests/apollo/test_skvbc_long_running.py
```diff
@@ -31,13 +34,17 @@ def start_replica_cmd(builddir, replica_id):
...
     viewChangeTimeoutMilli = "10000"
+    autoPrimaryRotationTimeoutMilli = "10000"
     path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
     return [path,
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
             "-s", statusTimerMilli,
             "-v", viewChangeTimeoutMilli,
-            "-p"]
+            "-a", autoPrimaryRotationTimeoutMilli,
```
teoparvanov: no, please, not the "-a" :) it's not used anywhere for now - it's a kind of possible, but unsupported feature

teoparvanov: let's do view change in a separate PR

tests/apollo/test_skvbc_long_running.py
```diff
@@ -31,13 +34,17 @@ def start_replica_cmd(builddir, replica_id):
     """
     statusTimerMilli = "500"
     viewChangeTimeoutMilli = "10000"
+    autoPrimaryRotationTimeoutMilli = "10000"
```
teoparvanov: remove pls

tests/apollo/test_skvbc_long_running.py
```diff
@@ -13,10 +13,13 @@
...
 import unittest
 import trio
-
+from os import environ
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
 from test_skvbc import SkvbcTest
 from test_skvbc_auto_view_change import SkvbcAutoViewChangeTest
+from test_skvbc_fast_path import SkvbcFastPathTest
+from test_skvbc_slow_path import SkvbcSlowPathTest
+from test_skvbc_view_change import SkvbcViewChangeTest
```
teoparvanov: let's only focus on fast path for this PR

tests/apollo/util/bft.py
```diff
@@ -661,3 +662,15 @@ def force_quorum_including_replica(self, replica_id, primary=0):
...
+        nb_slow_path = 0
+        with trio.move_on_after(seconds=.1):
```
teoparvanov: maybe .5 here?

tests/apollo/test_skvbc_long_running.py
Liorgal28: yeah, you right I forgot this from other tests I've made.

PR #453, opened 2020-03-17
Fix call to linkSTChainFrom() in DBAdapter's ctor

dartdart26:
Fix the call to linkSTChainFrom() in DBAdapter's constructor - linking
should be done from getLastReachableBlock() + 1 instead of
getLatestBlock() + 1 as the latter doesn't make sense and it will not
link anything.


dartdart26:
Pushed (without changes) to force a new CI run. Failure is not related to this change.

PR #454, opened 2020-03-17
Support empty blocks in the merkle tree DBAdapter

dartdart26:
Rationale for supporting empty blocks:
 * not imposing restrictions on users
 * support for marking that an event has happened without any data.

Please note that support for empty blocks with regards to client
filtering is a separate topic and will be added in future commits.

When a request for creating an empty block comes in, we create a block
node without any keys and with a state merkle tree root equal to the one
of the previous non-empty block. Additionally, when saving a leaf into
the DB, we serialize it as a concatenation of the block ID the value was
saved at and the value itself. This allows us to maintain a mapping
between merkle tree versions and block IDs and enables the
implementation of getKeyByReadVersion() and its support for returning
the actual block version of the key that is being searched for.

Remove sparse_merkle::IDBReader::get_leaf() as it is not being used.

Provide a specific unit test to prove empty blocks work as expected.
Additionally, enhance a number of existing tests with empty block cases.

### DIFF
--- a/storage/include/blockchain/merkle_tree_db_adapter.h
+++ b/storage/include/blockchain/merkle_tree_db_adapter.h
@@ -110,13 +110,11 @@ class DBAdapter : public DBAdapterBase {
   const sparse_merkle::Hash &getStateHash() const { return smTree_.get_root_hash(); }
 
  private:
-  concordUtils::Sliver createBlockNode(const concordUtils::SetOfKeyValuePairs &updates,
-                                       BlockId blockId,
-                                       const sparse_merkle::Version &stateRootVersion) const;
+  concordUtils::Sliver createBlockNode(const concordUtils::SetOfKeyValuePairs &updates, BlockId blockId) const;
 
   // Returns a set of key/value pairs that represent the needed DB updates for adding a block as part of the blockchain.
-  concordUtils::SetOfKeyValuePairs lastReachableBlockkDbUpdates(const concordUtils::SetOfKeyValuePairs &updates,
-                                                                BlockId blockId);
+  concordUtils::SetOfKeyValuePairs lastReachableBlockDbUpdates(const concordUtils::SetOfKeyValuePairs &updates,
+                                                               BlockId blockId);
 
   // Try to link the ST temporary chain to the blockchain from the passed blockId up to getLatestBlock().
   concordUtils::Status linkSTChainFrom(BlockId blockId);
@@ -148,9 +146,7 @@ namespace detail {
 
 // Serialize leafs in the DB as the block ID the value was saved at and the value itself.
 struct DatabaseLeafValue {
-  using BlockIdType = BlockId;
-
-  BlockIdType blockId;
+  BlockId blockId;
   sparse_merkle::LeafNode leafNode;
 };
 
--- a/storage/include/blockchain/merkle_tree_db_adapter.h
+++ b/storage/include/blockchain/merkle_tree_db_adapter.h
@@ -72,7 +72,6 @@ class DBAdapter : public DBAdapterBase {
   // - adding the whole block (raw block) in its own key
   // - calculating and filling in the parent digest.
   // Typically called by the application when adding a new block.
-  // Empty blocks are not allowed - this method will return IllegalOperation if an empty set is passed.
   concordUtils::Status addLastReachableBlock(const concordUtils::SetOfKeyValuePairs &updates);
 
   // Adds a block from its raw representation and a block ID.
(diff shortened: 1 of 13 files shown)


storage/src/merkle_tree_db_adapter.cpp
```diff
@@ -328,25 +342,30 @@ Status DBAdapter::getBlockById(BlockId blockId, Sliver &block) const {
   for (const auto &[key, keyData] : blockNode.keys) {
     if (!keyData.deleted) {
       Sliver value;
-      if (const auto status = db_->get(DBKeyManipulator::genDataDbKey(key, blockId), value); !status.isOK()) {
+      if (const auto status = db_->get(DBKeyManipulator::genDataDbKey(key, blockNode.stateRootVersion.value()), value);
```
andrewjstone: Is this because the blockId no longer matches the root version ?

storage/src/merkle_tree_db_adapter.cpp
```diff
@@ -328,25 +342,30 @@ Status DBAdapter::getBlockById(BlockId blockId, Sliver &block) const {
...
   return Status::OK();
 }
 
-concordUtils::SetOfKeyValuePairs DBAdapter::lastReachableBlockkDbUpdates(const SetOfKeyValuePairs &updates,
-                                                                         BlockId blockId) {
-  const auto updateBatch = smTree_.update(updates);
-
-  // Key updates.
-  auto dbUpdates = batchToDbUpdates(updateBatch);
+SetOfKeyValuePairs DBAdapter::lastReachableBlockkDbUpdates(const SetOfKeyValuePairs &updates, BlockId blockId) {
```
andrewjstone: Nit: Typo `Blockk`

storage/src/merkle_tree_db_adapter.cpp
```diff
@@ -328,25 +342,30 @@ Status DBAdapter::getBlockById(BlockId blockId, Sliver &block) const {
...
+  auto dbUpdates = SetOfKeyValuePairs{};
+  // Create a block with the same state root as the previous one if there are no updates.
+  if (!updates.empty()) {
+    // Key updates.
+    const auto updateBatch = smTree_.update(updates);
+    dbUpdates = batchToDbUpdates(updateBatch, blockId);
+  }
 
   // Block key.
   dbUpdates[DBKeyManipulator::genBlockDbKey(blockId)] = createBlockNode(updates, blockId, smTree_.get_version());
```
andrewjstone: Looking at this I was wondering why we didn't also pass `smTree.get_root_hash(`). It turns out that we don't need it in `createBlockNode` since `smTree_` is a member variable of the class. Therefore we also don't need to pass `smTree_.get_version()` here either. This can be made in a separate change, but I think it makes the code clearer.

storage/include/blockchain/merkle_tree_db_adapter.h
```diff
@@ -131,26 +129,37 @@ class DBAdapter : public DBAdapterBase {
...
   };
 
   sparse_merkle::Tree smTree_;
 };
 
+namespace detail {
+
+// Serialize leafs in the DB as the block ID the value was saved at and the value itself.
+struct DatabaseLeafValue {
+  using BlockIdType = BlockId;
```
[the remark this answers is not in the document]
dartdart26: Rationale is to call `fromBigEndianBuffer()` with the correct type. Can be achieved via `decltype`, doing that instead.

storage/include/blockchain/merkle_tree_serialization.h
```diff
@@ -345,6 +350,22 @@ inline block::detail::Node deserialize<block::detail::Node>(const concordUtils::
   return node;
 }
 
+// Deserializes the state root version from a serialized block::detail::Node .
+inline sparse_merkle::Version deserializeStateRootVersion(const concordUtils::Sliver &buf) {
+  Assert(buf.length() >= block::detail::Node::MIN_SIZE);
+  constexpr auto offset = sizeof(block::detail::Node::BlockIdType) + block::detail::Node::PARENT_DIGEST_SIZE +
```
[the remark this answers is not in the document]
dartdart26: That's a good point and I have also been thinking about it. I agree there is room for improvement here.

An option to consider is to introduce `serialize()` and `deserialize()` methods to types. That will solve the dependency between files. However, it might be a big burden on types that have a different purpose altogether. That is why I only introduced `SIZE_IN_BYTES` members in `sparse_merkle` types and not code or offsets. On the other hand, why would anybody care about size if not for serialization...

Another option might be to remove public data members and enforce usage of constructors so that users can't create an object in an incomplete state. Then, as you mention, we already have tests for serialization and that would help us find issues when there is a breaking change.

The problem is more relevant for types that are not part of the `detail` namespace, e.g. types in the `sparse_merkle` namespace as it would be perfectly valid to change these types without thinking about serialization there (and I think people should be encouraged not to think about it there).

Let me leave as is in this PR and please let's discuss in order to find a solution that will solve it for all cases in a uniform manner.

Thanks!

storage/include/sparse_merkle/tree.h
```diff
@@ -42,6 +42,7 @@ class Tree {
 
   const Hash& get_root_hash() const { return root_.hash(); }
   Version get_version() const { return root_.version(); }
+  const BatchedInternalNode& get_root() const { return root_; }
```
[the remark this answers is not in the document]
dartdart26: Initially, I had added this as I was saving an internal node if the first block (block ID = 1) was empty. However, that is not needed as we rely on the fact that tree versions start from 0 (hence the added test). I will just remove the unused `get_root()` method.

storage/src/merkle_tree_db_adapter.cpp
dartdart26: Yes.

storage/src/merkle_tree_db_adapter.cpp
dartdart26: Thanks, fixing!

storage/src/merkle_tree_db_adapter.cpp
dartdart26: Great catch! Doing that.

dartdart26:
Thanks @andrewjstone, great comments! I have addressed them in the last commit.

storage/include/blockchain/merkle_tree_serialization.h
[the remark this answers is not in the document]
andrewjstone: Ok. Cool. I'm also not sure I want to introduce serialize()/deserialize() methods. We can discuss at a later point.

PR #460, opened 2020-03-17
Add a conan package for RapidCheck

andrewjstone:
RapidCheck is a property based testing tool. This commit adds it as a
dependency.


.conan/rapidcheck_pkg/conanfile.py
```diff
@@ -0,0 +1,34 @@
...
+    scm = {
+        "type": "git",
+        "subfolder": ".",
+        "url": "https://github.com/emil-e/rapidcheck.git",
+        "revision": "258d907da00a0855f92c963d8f76eef115531716"
+    }
+    name = "rapidcheck"
+    version = "258d907da00a0855f92c963d8f76eef115531716"
+    license = "BSD 2-Clause"
+    url = "http://github.com/emil-e/rapidcheck"
```
dartdart26: [Minor]: I guess it doesn't use this URL to fetch (instead, it should be using the `scm.url `one), but maybe we can use `https` here too.

andrewjstone: Good catch. Will fix!

PR #462, opened 2020-03-18
[Apollo] Fix flaky test for RO replica

teoparvanov:
The metrics in this test case are being polled right after starting the replicas, which might explain the occasionally observed KeyError (the metrics not being available yet).
In this PR we preserve the existing test logic, just making the retrieval of metrics retry upon KeyError.

Testing done - ran the test 10 times locally without failure:
`for i in {1..10}; do python3 -m unittest test_skvbc_ro_replica.SkvbcReadOnlyReplicaTest 2>&1 > /dev/null; done`


andrewjstone:
Thanks for fixing this Teo!

PR #464, opened 2020-03-18
Simplify Conan based builds

andrewjstone:
* Remove unnecessary source methods from dependencies.
* Move the direct call to conan_basic_setup into conan_cmake_run.
* Remove CMake generator from conan files
    Since we use [cmake-conan](https://github.com/conan-io/cmake-conan), the
    CMake generator is used automatically. We don't need to add it to
    conanfile.txt or include the generated `conanbuildinfo.cmake` in
    CMakeLists.txt. Also remove the `generators=cmake` line from each of
    conanfile.py dependencies. It has no effect there for our project, as
    it's a top level setting.
* Remove unnecessary `target_include_directories` call in
`util/test/CMakeLists.txt`

### DIFF
--- a/.conan/hdrhistogram_pkg/conanfile.py
+++ b/.conan/hdrhistogram_pkg/conanfile.py
@@ -14,19 +14,12 @@ class HdrhistogramConan(ConanFile):
     url = "http://github.com/HdrHistogram/HdrHistogram_c"
     description = "A HDR histogram library"
     topics = ("HDR histogram")
-    generators = "cmake"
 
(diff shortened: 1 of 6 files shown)


andrewjstone:
@yontyon Please give this a look when you get a chance. Before I merge I'll check to make sure that it passes builds on blockchain CI.

Thanks!

andrewjstone:
It built with cache cleared and in the product. Only failed after the fact in a test that is failing everywhere. Build works though!

PR #465, opened 2020-03-18
Long run test: add VC test case

Liorgal28:
1) on VC test case change the variable name to expected next view, until now when we ran test cases it always remain primary number == view number but on a long run, it's not always true because 0<=primary<=number of replicas, but 0<=view<= infinity so I made the requested changes.

2)Add wait for fast path to be prevalent - tries for 5 seconds to get in a fast path.

3)Run long-run tests for 2 hours locally.

### DIFF
--- a/tests/apollo/test_skvbc_fast_path.py
+++ b/tests/apollo/test_skvbc_fast_path.py
@@ -71,7 +71,7 @@ class SkvbcFastPathTest(unittest.TestCase):
 
         await tracker.run_concurrent_ops(num_ops=numops, write_weight=write_weight)
 
-        await bft_network.assert_fast_path_prevalent(nb_slow_paths_so_far=nb_slow_path)
+        await bft_network.wait_for_fast_path_to_be_prevalent(nb_slow_paths_so_far=nb_slow_path)
 
     @with_trio
     @with_bft_network(start_replica_cmd)
--- a/tests/apollo/test_skvbc_long_running.py
+++ b/tests/apollo/test_skvbc_long_running.py
@@ -9,7 +9,7 @@
 # notices and license terms. Your use of these subcomponents is subject to the
 # terms and conditions of the subcomponent's license, as noted in the LICENSE
 # file.
-
+import itertools
 import os.path
 import unittest
 import trio
@@ -51,9 +51,8 @@ class SkvbcLongRunningTest(unittest.TestCase):
                       selected_configs=lambda n, f, c: n == 7)
     async def test_stability(self, bft_network):
         bft_network.start_all_replicas()
-        i=0
-        with trio.move_on_after(seconds=EIGHT_HOURS_IN_SECONDS):
-            while True:
+        with trio.move_on_after(seconds=EIGHT_HOURS_IN_SECONDS/2):
+            for i in itertools.count():
                 await SkvbcTest().test_get_block_data\
                     (bft_network=bft_network, already_in_trio=True)
                 await trio.sleep(seconds=10)
@@ -67,4 +66,3 @@ class SkvbcLongRunningTest(unittest.TestCase):
                     await SkvbcViewChangeTest().test_single_vc_only_primary_down \
                       (bft_network=bft_network, already_in_trio=True, disable_linearizability_checks=True)
                     await trio.sleep(seconds=10)
-                i += 1
--- a/tests/apollo/test_skvbc_long_running.py
+++ b/tests/apollo/test_skvbc_long_running.py
@@ -17,6 +17,7 @@ from os import environ
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
 from test_skvbc import SkvbcTest
 from test_skvbc_fast_path import SkvbcFastPathTest
+from test_skvbc_view_change import SkvbcViewChangeTest
 
 # Time consts
 EIGHT_HOURS_IN_SECONDS = 8 * 60 * 60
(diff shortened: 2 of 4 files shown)


tests/apollo/test_skvbc_fast_path.py
```diff
@@ -71,7 +71,7 @@ def setUp(self):
 
         await tracker.run_concurrent_ops(num_ops=numops, write_weight=write_weight)
 
-        await bft_network.assert_fast_path_prevalent(nb_slow_paths_so_far=nb_slow_path)
+        await bft_network.wait_for_fast_path_to_be_prevalent(nb_slow_paths_so_far=nb_slow_path)
```
teoparvanov: nice!

tests/apollo/test_skvbc_long_running.py
```diff
@@ -50,7 +51,8 @@ class SkvbcLongRunningTest(unittest.TestCase):
                       selected_configs=lambda n, f, c: n == 7)
     async def test_stability(self, bft_network):
         bft_network.start_all_replicas()
-        with trio.move_on_after(seconds=EIGHT_HOURS_IN_SECONDS/2):
+        i=0
+        with trio.move_on_after(seconds=EIGHT_HOURS_IN_SECONDS):
             while True:
```
teoparvanov: Instead of having a separately maintained index `i`, let's use the following Python idiom for infinite indexed loop.
So we should replace:
```python
while True:
   ...
```
by:
```python
for i in itertools.count():
   ...
```

tests/apollo/test_skvbc_long_running.py
```diff
@@ -60,4 +62,9 @@ class SkvbcLongRunningTest(unittest.TestCase):
...
                 await SkvbcFastPathTest().test_fast_path_read_your_write \
                     (bft_network=bft_network, already_in_trio=True, disable_linearizability_checks=True)
-                await trio.sleep(seconds=10)
\ No newline at end of file
+                await trio.sleep(seconds=10)
+                if i % 250 == 0:
+                    await SkvbcViewChangeTest().test_single_vc_only_primary_down \
+                      (bft_network=bft_network, already_in_trio=True, disable_linearizability_checks=True)
+                    await trio.sleep(seconds=10)
+                i += 1
```
teoparvanov: if you address my comment above, you can drop the `i += 1` here

tests/apollo/test_skvbc_view_change.py
```diff
@@ -318,17 +318,16 @@ class SkvbcViewChangeTest(unittest.TestCase):
...
 
-        initial_primary = 0
-        
-        replcas_to_stop = [ v for v in range(initial_primary,
-                                             initial_primary + num_consecutive_failing_primaries) ]
-        
-        expected_final_primary = initial_primary + num_consecutive_failing_primaries
+        initial_primary = await bft_network.get_current_primary()
+        replcas_to_stop = [ v for v in range((initial_primary % bft_network.config.n),
+                                             (initial_primary % bft_network.config.n) + num_consecutive_failing_primaries) ]
```
teoparvanov: The logic here is wrong on several levels. First of all, `initial_primary` is always in the interval `[0,n]`, as you have `initial_primary = await bft_network.get_current_primary()`.

Also, a range will not work for what you want to achieve here, because you want potentially to have a list of primaries, such as [5,0,1,2].

Please review this logic...

tests/apollo/test_skvbc_view_change.py
```diff
@@ -318,17 +318,16 @@ class SkvbcViewChangeTest(unittest.TestCase):
...
         
+        expected_final_view = (initial_primary + num_consecutive_failing_primaries)
+
         await self._send_random_writes(tracker)
 
         await bft_network.wait_for_view(
-            replica_id=initial_primary,
+            replica_id=initial_primary % bft_network.config.n,
```
teoparvanov: again, I don't understand why you add `% bft_network.config.n`, given that `initial_primary = await bft_network.get_current_primary()`

tests/apollo/test_skvbc_view_change.py
Liorgal28: Teo and I agreed that we need to have 2 methods, one to get the current primary and one to get the current view.

tests/apollo/util/bft.py
```diff
@@ -360,7 +360,17 @@ def get_live_replicas(self):
         current_primary = await self.wait_for_view(
             replica_id=live_replica, expected=None)
 
-        return current_primary
+        return current_primary % self.config.n
```
teoparvanov: This method can simply become:
```python
async def get_current_primary(self):
        """
        Returns the current primary replica id
        """
        return self.get_current_view() % config.n
```

tests/apollo/test_skvbc_long_running.py
```diff
@@ -60,4 +61,8 @@ class SkvbcLongRunningTest(unittest.TestCase):
                 await trio.sleep(seconds=10)
                 await SkvbcFastPathTest().test_fast_path_read_your_write \
                     (bft_network=bft_network, already_in_trio=True, disable_linearizability_checks=True)
-                await trio.sleep(seconds=10)
\ No newline at end of file
+                await trio.sleep(seconds=10)
+                if i % 250 == 0:
```
teoparvanov: 250 looks like what we would call a "magic number". Where does it come from? Does it depend on the overall long run test duration?

PR #470, opened 2020-03-19
Allow detection of possible primary replica fault during pre-execution

yuliasherman:
- Add a new timer - preProcessReqWaitTimeMilli = viewChangeTimerMillisec / 4 for detection of a possible primary replica crash
- Add logic for possible primary replica crash detection. In case a non-primary replica receives ClientPreProcessRequestMsg from the client, it is registered. A periodic check for registered messages not followed by PreProcessRequestMsg from the primary replica is performed. Those messages are passed for the regular (without pre-execution) ReplicaImp handling and decision regarding a view change is taken by its code.
- Make use of enabling/disabling pre-execution feature flag
- Add unit tests

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -115,14 +115,14 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
   const uint16_t firstClientId = numOfReplicas_;
   for (auto i = 0; i < numOfClients_; i++) {
     // Placeholders for all clients
-    ongoingRequests_[firstClientId + i] = make_unique<ClientRequestInfo>();
+    ongoingRequests_[firstClientId + i] = make_shared<ClientRequestState>();
   }
   // Allocate a buffer for the pre-execution result per client
   for (auto id = 0; id < numOfClients_; id++) {
     preProcessResultBuffers_.push_back(Sliver(new char[maxReplyMsgSize_], maxReplyMsgSize_));
   }
   threadPool_.start(numOfClients_);
-  RequestProcessingInfo::init(numOfRequiredReplies(), preProcessReqWaitTimeMilli_);
+  RequestProcessingState::init(numOfRequiredReplies(), preProcessReqWaitTimeMilli_);
   addTimers();
 }
 
@@ -132,13 +132,15 @@ PreProcessor::~PreProcessor() {
 }
 
 void PreProcessor::addTimers() {
+  // This timer is used for a periodic detection of timed out client requests.
+  // Each such request contains requestTimeoutMilli parameter that defines its lifetime.
   if (preExecReqStatusCheckTimeMilli_ != 0)
     requestsStatusCheckTimer_ = TimersSingleton::getInstance().add(
         chrono::milliseconds(preExecReqStatusCheckTimeMilli_), Timers::Timer::RECURRING, [this](Timers::Handle h) {
           onRequestsStatusCheckTimer(h);
         });
 
-  // The timer below is used in the following scenario: a non-primary replica receives a request from the client
+  // This timer is used in the following scenario: a non-primary replica receives a request from the client
   // and sends it to the primary one. After a reasonable time, it expects to receive PreProcessRequestMsg from the
   // primary replica for this request. If this does not happen, it is supposed that the primary replica is faulty
   // and the request is passed to ReplicaImp for further handling.
@@ -163,7 +165,7 @@ void PreProcessor::onRequestsStatusCheckTimer(Timers::Handle timer) {
   // Pass through all ongoing requests and abort those that are timed out.
   for (const auto &clientEntry : ongoingRequests_) {
     lock_guard<mutex> lock(clientEntry.second->mutex);
-    if (clientEntry.second->clientReqInfoPtr && clientEntry.second->clientReqInfoPtr->isReqTimedOut()) {
+    if (clientEntry.second->reqProcessingStatePtr && clientEntry.second->reqProcessingStatePtr->isReqTimedOut()) {
       preProcessorMetrics_.preProcessRequestTimedout.Get().Inc();
       releaseClientPreProcessRequest(clientEntry.second, clientEntry.first);
     }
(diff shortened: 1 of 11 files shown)


bftengine/src/preprocessor/RequestProcessingInfo.cpp
```diff
@@ -69,20 +69,35 @@ auto RequestProcessingInfo::calculateMaxNbrOfEqualHashes(uint16_t &maxNumOfEqual
...
-                                     << reqSeqNum_ << "; reqProcessingTime=" << reqProcessingTime << "; abort request");
+                                     << reqSeqNum_ << "; reqProcessingTime=" << reqProcessingTime);
       return true;
     }
   }
   return false;
 }
 
+// Non-primary replica logic
+bool RequestProcessingInfo::isPreProcessReqMsgReceivedInTime(const uint16_t preProcessReqWaitTimeMilli) const {
```
teoparvanov: Hi Yulia - I couldn't find where you call `isPreProcessReqMsgReceivedInTime` ?
My guess is - the PR is still WIP, so everything is not fully wired yet.

bftengine/src/preprocessor/RequestProcessingInfo.hpp
```diff
@@ -44,6 +44,7 @@ class RequestProcessingInfo {
   bool isReqTimedOut() const;
   uint64_t getReqTimeoutMilli() const { return clientPreProcessReqMsg_->requestTimeoutMilli(); }
   std::string getReqCid() const { return clientPreProcessReqMsg_->getCid(); }
+  bool isPreProcessReqMsgReceivedInTime(const uint16_t preProcessReqWaitTimeMilli) const;
```
teoparvanov: why is the `preProcessReqWaitTimeMilli` passed as a parameter for each timeout check? Does it change?
I'd rather imagine it as a member variable of `RequestProcessingInfo`...

bftengine/src/preprocessor/RequestProcessingInfo.cpp
yuliasherman: Yes, exactly :-)

bftengine/src/preprocessor/RequestProcessingInfo.hpp
yuliasherman: I've changed the approach now.

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -75,11 +76,13 @@ class PreProcessor {
   void registerRequest(ClientPreProcessReqMsgUniquePtr clientReqMsg,
                        PreProcessRequestMsgSharedPtr preProcessRequestMsg);
   void releaseClientPreProcessRequestSafe(uint16_t clientId);
-  void releaseClientPreProcessRequest(ClientRequestInfoSharedPtr clientEntry, uint16_t clientId);
+  void releaseClientPreProcessRequest(const ClientRequestInfoSharedPtr &clientEntry, uint16_t clientId);
```
teoparvanov: Why not pass by value, as usual for shared ptrs?

bftengine/src/preprocessor/RequestProcessingInfo.cpp
```diff
@@ -69,20 +83,35 @@ auto RequestProcessingInfo::calculateMaxNbrOfEqualHashes(uint16_t &maxNumOfEqual
...
       return true;
     }
   }
   return false;
 }
 
+// Non-primary replica logic
+bool RequestProcessingInfo::isPreProcessReqMsgReceivedInTime() const {
+  // Check if the request was registered for too long after been received from the client
+  auto clientRequestWaitingTime = getMonotonicTimeMilli() - entryTime_;
```
teoparvanov: clientRequestWaitingTimeMilli

bftengine/src/preprocessor/tests/preprocessor_test.cpp
```diff
@@ -344,92 +424,60 @@ TEST(requestPreprocessingInfo_test, clientPreProcessMessageConversion) {
...
+  Assert(preProcessor.getOngoingReqIdForClient(clientId) == reqSeqNum);
+  usleep(replicaConfig.preExecReqStatusCheckTimerMillisec * 1000);
+  TimersSingleton::getInstance().evaluate();
+  Assert(preProcessor.getOngoingReqIdForClient(clientId) == 0);
+}
 
-  void updateMetricsForInternalMessage() {}
-  bool isCollectingState() const { return false; }
+TEST(requestPreprocessingInfo_test, primaryCrashDetected) {
+  setUpConfiguration_7();
```
teoparvanov: great unit test!

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -105,23 +122,46 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
...
+  cancelTimers();
+  threadPool_.stop();
+}
+
+void PreProcessor::addTimers() {
+  if (preExecReqStatusCheckTimeMilli_ != 0)
+    requestsStatusCheckTimer_ = TimersSingleton::getInstance().add(
+        chrono::milliseconds(preExecReqStatusCheckTimeMilli_), Timers::Timer::RECURRING, [this](Timers::Handle h) {
+          onRequestsStatusCheckTimer(h);
+        });
```
teoparvanov: as below, can you also explain what this timer is used for?

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -171,22 +225,27 @@ void PreProcessor::onMessage<ClientPreProcessRequestMsg>(ClientPreProcessRequest
...
+            "Received ClientPreProcessRequestMsg reqSeqNum=" << reqSeqNum << " from clientId=" << clientId
+                                                             << ", senderId=" << senderId);
   if (!checkClientMsgCorrectness(clientPreProcessReqMsg, reqSeqNum)) {
     preProcessorMetrics_.preProcReqIgnored.Get().Inc();
     return;
   }
   {
-    auto &clientEntry = ongoingRequests_[clientId];
+    const auto &clientEntry = ongoingRequests_[clientId];
     lock_guard<mutex> lock(clientEntry->mutex);
```
teoparvanov: Let's set the lock first:
```c++
    const auto &clientEntry = ongoingRequests_[clientId];
```

bftengine/src/preprocessor/PreProcessor.hpp
yuliasherman: Though the use of shared pointers has a low overhead, in case we know that a pointer is not expected to be stored/reset/changed in the code it is passed to, better to pass it by const&.

bftengine/src/preprocessor/RequestProcessingInfo.cpp
yuliasherman: OK

bftengine/src/preprocessor/tests/preprocessor_test.cpp
yuliasherman: Thanks!

bftengine/src/preprocessor/PreProcessor.hpp
teoparvanov: Thanks for the clarification.

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: OK

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: That's all the idea - clientEntry (ClientRequestInfoSharedPtr) always exists. What is changed is RequestProcessingInfoUniquePtr. That's why we can safely get ongoingRequests_[clientId], but then it should be locked to get  RequestProcessingInfo.
Actually I think you were right and RequestProcessingInfo should be renamed to RequestProcessingState to be more understandable.

teoparvanov: Thanks Yulia, I appreciate the improvement!

PR #471, opened 2020-03-20
Remove ctest startup delay of 30-40 seconds

andrewjstone:
Two compounding issues triggered any ctest runs to take 30-40 seconds to
startup. First, all cmake variables were passed to all Apollo tests as
the process environment. This causes shell interpolation of the env
variable, which for some reason is very expensive. Second, the number of
cmake variables was drastically increased by the use of the conan-cmake
module. Therefore, each test initialization at ctest runtime triggered
overhead when `add_test` was scanned for each apollo test. This happened
to be 5s per test.

This commit fixes the issue by instead only passing the needed variable
as the environment for each test.

### DIFF
--- a/tests/apollo/CMakeLists.txt
+++ b/tests/apollo/CMakeLists.txt
@@ -1,40 +1,35 @@
-get_cmake_property(_variableNames VARIABLES)
-list (SORT _variableNames)
-set(ALL_ENV_VARS "")
-foreach (_variableName ${_variableNames})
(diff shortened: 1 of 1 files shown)


andrewjstone:
> Thanks Andrew, for now we only use BUILD_ROCKSDB_STORAGE, so this is ok. We can add others as the need arises.
> 
> PS: thanks to your change here, we've gained about 1-2 min per CI job :) Awesome!

YAY!

PR #473, opened 2020-03-20
[Apollo] Fixes for running against a containerized SKVBC deployment

teoparvanov:
This PR introduces the following improvements:
- [TesterReplica] enable injecting a custom `IBlockMetadata` implementation via the InternalCommandsHandler's constructor. This is required for running SKVBC in the product - which has its own `IBlockMetadata` implementation - `ConcordBlockMetadata`.
- [Python] enable "priming" the SKVBC for state transfer in case of an existing deployment

### DIFF
--- a/tests/apollo/util/skvbc.py
+++ b/tests/apollo/util/skvbc.py
@@ -181,7 +181,8 @@ class SimpleKVBCProtocol:
             checkpoints_num=2,
             persistency_enabled=True):
         initial_nodes = self.bft_network.all_replicas(without=stale_nodes)
-        [self.bft_network.start_replica(i) for i in initial_nodes]
+        self.bft_network.start_all_replicas()
+        self.bft_network.stop_replicas(stale_nodes)
         client = SkvbcClient(self.bft_network.random_client())
         # Write a KV pair with a known value
         known_key = self.max_key()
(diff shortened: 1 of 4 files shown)


tests/apollo/util/skvbc.py
```diff
@@ -181,7 +181,8 @@ def max_key(cls):
             checkpoints_num=2,
             persistency_enabled=True):
         initial_nodes = self.bft_network.all_replicas(without=stale_nodes)
-        [self.bft_network.start_replica(i) for i in initial_nodes]
+        self.bft_network.start_all_replicas()
+        self.bft_network.stop_replicas(stale_nodes)
```
andrewjstone: Makes sense. This is because in the containerized environment the nodes are already running

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
```diff
@@ -53,9 +53,8 @@ int InternalCommandsHandler::execute(uint16_t clientId,
 }
 
 void InternalCommandsHandler::addMetadataKeyValue(SetOfKeyValuePairs &updates, uint64_t sequenceNum) const {
-  concord::kvbc::BlockMetadata metadata(*m_storage);
-  Sliver metadataKey = metadata.getKey();
-  Sliver metadataValue = metadata.serialize(sequenceNum);
+  Sliver metadataKey = m_blockMetadata->getKey();
```
andrewjstone: Just a note that this doesn't match the google style guide which uses suffix underscores rather than `m_` prefixes. However, this does conform to the existing style of the file so I support this choice. We should change all of the member variables together in another commit.

tests/apollo/util/skvbc.py
teoparvanov: Yup!

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
teoparvanov: Thanks Andrew - indeed, I preferred to conform to the existing code in this instance. We can clean this up in a separate effort.

PR #474, opened 2020-03-22
Refactoring and cleanup for DBAdapter

toly-kournik:
Removed ILocalKeyValueStorageReadOnlyIterator
move util/include/kv_types.hpp -> kvbc/include/kv_types.hpp including appropriate namespacing


kvbc/include/kv_types.hpp
```diff
@@ -10,32 +10,42 @@
...
-typedef Sliver Value;
+typedef concordUtils::Sliver Key;
+typedef concordUtils::Sliver Value;
 typedef std::pair<Key, Value> KeyValuePair;
 typedef std::unordered_map<Key, Value> SetOfKeyValuePairs;
 typedef std::map<Key, Value> OrderedSetOfKeyValuePairs;
 typedef std::vector<Key> KeysVector;
 typedef KeysVector ValuesVector;
-typedef uint64_t BlockId;
+typedef std::uint64_t BlockId;
```
dartdart26: Can we include `<cstdint>` for that?

toly-kournik: sure

kvbc/include/kv_types.hpp
```diff
@@ -10,32 +10,42 @@
...
 
-OrderedSetOfKeyValuePairs order(const SetOfKeyValuePairs &unordered);
-}  // namespace concordUtils
+template <typename ContainerIn, typename ContainerOut>
+ContainerOut order(const ContainerIn& unordered) {
+  ContainerOut out;
+  for (auto&& kv : unordered) out.insert(kv);
+  return out;
+}
```
dartdart26: I think it would be better to not allow users to specify types in order to avoid cases such as:

```
order<OrderedSetOfKeyValuePairs, UnorderedSetOfKeyValuePairs>()
```

In that case, this function will not order.

kvbc/include/ReplicaImp.h
```diff
@@ -159,77 +154,9 @@ class ReplicaImp : public IReplica,
...
                                           bool &outRes) const override;
 
-    virtual concord::storage::blockchain::ILocalKeyValueStorageReadOnlyIterator *getSnapIterator() const override;
-
-    virtual Status freeSnapIterator(
-        concord::storage::blockchain::ILocalKeyValueStorageReadOnlyIterator *iter) const override;
     virtual void monitor() const override;
   };
 
-  class StorageIterator : public concord::storage::blockchain::ILocalKeyValueStorageReadOnlyIterator {
```
dartdart26: Nice!

kvbc/test/order_test.cpp
```diff
@@ -0,0 +1,77 @@
+// Copyright 2018 VMware, all rights reserved
+/**
+ * Test the Sliver class.
+ *
+ * While these tests check that Sliver "works", the most interesting test is to
+ * run this under valgrind, and make sure that it doesn't find any
+ * leaks/double-frees/etc. `valgrind --leak-check=full test/SliverTests`
+ */
```
dartdart26: I think we should change that comment to reflect the tests in this file.

kvbc/include/kv_types.hpp
toly-kournik: agree

kvbc/test/order_test.cpp
toly-kournik: agree

kvbc/test/CMakeLists.txt
```diff
@@ -9,3 +9,9 @@ if(Boost_FOUND)
         kvbc
     )
 endif()
+
+add_executable(order_test order_test.cpp $<TARGET_OBJECTS:logging_dev>)
+add_test(order_test order_test)
+target_include_directories(order_test PUBLIC ${libkvbc_SOURCE_DIR}/include)
+target_link_libraries(order_test GTest::Main util)
+target_compile_options(order_test PUBLIC -Wno-sign-compare)
```
dartdart26: [Minor]: New line.

kvbc/include/kv_types.hpp
```diff
@@ -10,32 +10,42 @@
 // file.
 //
 
-#ifndef CONCORD_BFT_UTIL_KV_TYPES_H_
-#define CONCORD_BFT_UTIL_KV_TYPES_H_
+#pragma once
 
+#include <string_view>
```
dartdart26: Why do we need the `<string_view>` include here?

toly-kournik: check out the hash

dartdart26: Thanks, missed that.

andrewjstone:
> I also have a concern that moving types from util to kvbc creates dependencies that we might not want, for example between storage and kvbc. It should not hold this PR, but is something we might want to consider.

This change should be reverted, or better yet, we should move those types into the `storage` library. As it stands, this creates a circular dependency. kvbc -> corebft -> storage -> kvbc

andrewjstone:
> > I also have a concern that moving types from util to kvbc creates dependencies that we might not want, for example between storage and kvbc. It should not hold this PR, but is something we might want to consider.
> 
> This change should be reverted, or better yet, we should move those types into the `storage` library. As it stands, this creates a circular dependency. kvbc -> corebft -> storage -> kvbc

We discussed this offline, and decided the types do live in blockchain, and also some other things should be moved around to fix any circular dependencies.

PR #476, opened 2020-03-22
[Apollo] Explicit test for checkpoint creation

teoparvanov:
Checkpoint creation is implicitly tested by the persistency and state transfer tests. However, it also makes sense to explicitly test the behavior of the checkpoint protocol. Additionally, doing so enables us to include this scenario in long running or product CI tests.

PS: with this PR we also make `fill_and_wait_for_checkpoint()` idempotent by taking into account previously existing checkpoints (useful for long-running tests).

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -586,13 +586,19 @@ class BftTestNetwork:
                     nursery.start_soon(self.wait_for_checkpoint, replica_id,
                             checkpoint_num)
 
-    async def wait_for_checkpoint(self, replica_id, checkpoint_num):
+    async def wait_for_checkpoint(self, replica_id, expected_checkpoint_num=None):
+        """
+        Wait for a single replica to reach the expected_checkpoint_num.
+        If none is provided, return the last stored checkpoint.
+        """
         key = ['bc_state_transfer', 'Gauges', 'last_stored_checkpoint']
-        while True:
-            with trio.move_on_after(.5): # seconds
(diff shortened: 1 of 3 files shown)


tests/apollo/util/bft.py
```diff
@@ -586,13 +586,18 @@ def force_quorum_including_replica(self, replica_id, primary=0):
...
                             checkpoint_num)
 
-    async def wait_for_checkpoint(self, replica_id, checkpoint_num):
+    async def wait_for_checkpoint(self, replica_id, expected_checkpoint_num=None):
+        """
+        Wait for a single replica to reach the expected_checkpoint_num.
+        If none is provided, return the last stored checkpoint.
+        """
         key = ['bc_state_transfer', 'Gauges', 'last_stored_checkpoint']
         while True:
```
andrewjstone: One thing I noticed is that we never actually give up on waiting for the checkpoint to complete. Maybe we should have some upper bound (a few minutes for that, where the test fails from a `trio.fail_after` call). I suppose it's also fine to leave it like this if the test should timeout at a higher level. In fact, that's likely the right thing to do, allow the failure wrapper at the caller since the functions compose. We don't actually do that checking in `fill_and_wait_for_checkpoint` or `test_checkpoint_creation` though.

teoparvanov: Thanks Andrew, great catch! I've added a `trio.fail_after` within the `wait_for_checkpoint()` method as it can be used anywhere in the code. This also avoids the clutter of doing `trio.fail_after` every time we use it...

PR #480, opened 2020-03-24
Enable/disable pre-execution feature via configuration

yuliasherman:
Control pre-execution feature availability via a configuration file.

https://jira.eng.vmware.com/browse/BC-1808

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -66,6 +66,9 @@ struct ReplicaConfig {
   // a time interval in milliseconds, represents the timeout for automatically replacing the primary
   uint16_t autoPrimaryRotationTimerMillisec = 0;
 
+  // preExecutionFeatureEnabled=0/1; preExecutionFeatureEnabled=1 enables the pre-execution feature
+  uint8_t preExecutionFeatureEnabled = 0;
+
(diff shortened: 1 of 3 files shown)


bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -66,6 +66,9 @@ struct ReplicaConfig {
   // a time interval in milliseconds, represents the timeout for automatically replacing the primary
   uint16_t autoPrimaryRotationTimerMillisec = 0;
 
+  // preExecutionFeatureEnabled=0/1; preExecutionFeatureEnabled=1 enables the pre-execution feature
+  uint8_t preExecutionFeatureEnabled = 0;
```
teoparvanov: why not `bool` ?

yuliasherman: Do we like to serialize booleans?

yuliasherman: Iknow that we do that, but it is a bad practice.

teoparvanov: I'd prefer to be consistent with the rest of the code - and do an eventual clean-up separately.

yuliasherman: Hm. OK.

PR #481, opened 2020-03-24
Key-Value Bloclchain - Storage separation

toly-kournik:
In order to have more clear layering and dependencies between components
kvbc-related code is moved to kvbc lib.

The goal is to have the following clear layers separation where upper layers depend on lower ones, but not the opposite:

                  kvbc
                  / |
                /   bft
               |    |
              storage

There are still dependencies on kvbc which will be removed in subsequent commit, dealing with dbadapter re-factoring.
Namespace changes will be done in susequent commit as well.


kvbc/include/db_interfaces.h
```diff
@@ -0,0 +1,52 @@
...
+/** Defines interface for blockchain data keys generation
+ *
+ */
+class IDataKeyGenerator {
+ public:
+  virtual concordUtils::Sliver blockKey(const BlockId&) const = 0;
+  virtual concordUtils::Sliver dataKey(const Key&, const BlockId&) const = 0;
+
+  virtual ~IDataKeyGenerator() = default;
+};
```
dartdart26: Nice!

bftengine/src/simplestatetransfer/SimpleStateTran.cpp
```diff
@@ -279,7 +278,7 @@ SimpleStateTran::SimpleStateTran(
   config.cVal = cVal;
   config.numReplicas = 3 * fVal + 2 * cVal + 1;
   config.pedanticChecks = pedanticChecks;
-  auto comparator = concord::storage::memorydb::KeyComparator(new concord::storage::blockchain::DBKeyComparator());
+  auto comparator = concord::storage::memorydb::KeyComparator();
```
dartdart26: If I am not mistaken, removing the comparator will force lexicographical comparison. Will that be OK in this context?

dartdart26:
@toly-kournik Thanks for the refactoring! Looks good! I only have a minor comment which should not be critical.

PR #482, opened 2020-03-24
Add merkle DBAdapter property-based tests

dartdart26:
Add property-based tests for the merkle tree DBAdapter. Focus on the
public interface that allows adding/getting blocks, key lookups and
getting lastReachable/latest block IDs. Moreover, provide tests that
simulate state transfer. Testing of key/block deletion will be added in
future commits.

Property-based tests are only run with RocksDB, due to test timing
constraints. Future commits may improve timing and include memorydb
tests too.

Remove the unused blockId parameter from v2MerkleTree::block::create() .

### DIFF
--- a/storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
+++ b/storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
@@ -1,0 +1,313 @@
+// Copyright 2020 VMware, all rights reserved
+
+#include "gtest/gtest.h"
+#include "rapidcheck/rapidcheck.h"
+#include "rapidcheck/extras/gtest.h"
+
+#include "storage_test_common.h"
+
+#include "blockchain/merkle_tree_block.h"
+#include "blockchain/merkle_tree_db_adapter.h"
+#include "kv_types.hpp"
+#include "sliver.hpp"
+#include "sparse_merkle/base_types.h"
+#include "storage/db_interface.h"
+
+#include <array>
+#include <cstdint>
+#include <cstring>
+#include <memory>
+#include <set>
+#include <string>
+#include <utility>
+#include <vector>
+
+using ::bftEngine::SimpleBlockchainStateTransfer::StateTransferDigest;
+using ::concord::kvbc::Key;
+using ::concord::kvbc::SetOfKeyValuePairs;
+using ::concord::kvbc::Value;
+using ::concord::storage::IDBClient;
+using ::concord::storage::blockchain::BlockId;
+using ::concord::storage::blockchain::v2MerkleTree::block::create;
+using ::concord::storage::blockchain::v2MerkleTree::block::getData;
+using ::concord::storage::blockchain::v2MerkleTree::block::getParentDigest;
+using ::concord::storage::blockchain::v2MerkleTree::block::getStateHash;
+using ::concord::storage::blockchain::v2MerkleTree::block::BLOCK_DIGEST_SIZE;
+using ::concord::storage::blockchain::v2MerkleTree::DBAdapter;
+using ::concord::storage::sparse_merkle::Hash;
+using ::concordUtils::Sliver;
+
+using ParentDigest = std::array<std::uint8_t, BLOCK_DIGEST_SIZE>;
+using HashArray = std::array<std::uint8_t, Hash::SIZE_IN_BYTES>;
+
+namespace rc {
+
+template <>
+struct Arbitrary<Sliver> {
+  static auto arbitrary() { return gen::construct<Sliver>(gen::string<std::string>()); }
+};
+
+template <>
+struct Arbitrary<Hash> {
+  static auto arbitrary() { return gen::construct<Hash>(gen::arbitrary<HashArray>()); }
+};
+
+// Generate key/value sets with arbitrary keys and non-empty values.
+template <>
+struct Arbitrary<SetOfKeyValuePairs> {
+  static auto arbitrary() {
+    return gen::container<SetOfKeyValuePairs>(gen::arbitrary<Sliver>(),
+                                              gen::construct<Sliver>(gen::nonEmpty(gen::string<std::string>())));
+  }
+};
+
+}  // namespace rc
+
+namespace {
+
+const auto zeroDigest = ParentDigest{};
+
+bool equal(const void *lhs, const ParentDigest &rhs) { return std::memcmp(lhs, rhs.data(), BLOCK_DIGEST_SIZE) == 0; }
+
+bool equal(const void *lhs, const StateTransferDigest &rhs) {
+  return std::memcmp(lhs, rhs.content, BLOCK_DIGEST_SIZE) == 0;
+}
+
+::testing::AssertionResult addBlocks(const std::vector<SetOfKeyValuePairs> &blockUpdates, DBAdapter &adapter) {
+  for (const auto &updates : blockUpdates) {
+    const auto status = adapter.addLastReachableBlock(updates);
+    if (!status.isOK()) {
+      return ::testing::AssertionFailure() << status;
+    }
+  }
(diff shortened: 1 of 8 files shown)


storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+    const auto stateTransferOffset = *rc::gen::inRange(2, 10000);
+
+    auto adapter = DBAdapter{GetParam()->db()};
+
+    // First, add blockchain updates.
+    RC_ASSERT(addBlocks(blockchainUpdates, adapter));
+
+    // Then, add state transfer blocks at an offset, ensuring a hole.
+    for (auto i = 0ul; i < stateTransferUpdates.size(); ++i) {
+      const auto blockId = adapter.getLatestBlock() + stateTransferOffset + i;
```
andrewjstone: Won't this cause a gap between every block, since it includes the `stateTransferOffset`?

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+      RC_ASSERT(adapter.addBlock(create(stateTransferUpdates[i], parentDigest.data(), stateHash), blockId).isOK());
+      RC_ASSERT(adapter.getLatestBlock() == blockId);
+    }
+
+    // Ensure there is a hole.
```
andrewjstone: This comment doesn't seem to match the assertion.

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+        RC_ASSERT(equal(getParentDigest(block), blockDigest(i, prevBlock)));
+      }
+    }
+  };
+
+  ASSERT_TRUE(rc::check(test));
+}
+
+// Test that when adapter.addBlock(block, getLastReachableBlock() + 1) is called and there is a hole between the
+// blockchain and the state transfer chain, the block is added at the back of the blockchain.
```
andrewjstone: Maybe clarify to say after `LastReachableBlock` (in the gap).

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    // Verify that the block was added at the back.
+    RC_ASSERT(adapter.getLastReachableBlock() >= blockchainUpdates.size() + 1);
+    RC_ASSERT(adapter.getLastReachableBlock() <= adapter.getLatestBlock());
+  };
+
+  ASSERT_TRUE(rc::check(test));
+}
+
+// Test that getLastReachableBlock() <= getLatestBlock() holds during a simulated state transfer.
```
andrewjstone: Excellent property to check!

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+using ::concord::storage::blockchain::v2MerkleTree::DBAdapter;
+using ::concord::storage::sparse_merkle::Hash;
+using ::concordUtils::Sliver;
+
+using ParentDigest = std::array<std::uint8_t, BLOCK_DIGEST_SIZE>;
+using HashArray = std::array<std::uint8_t, Hash::SIZE_IN_BYTES>;
+
+namespace rc {
+
+template <>
```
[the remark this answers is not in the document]
dartdart26: Yes, agreed. We should do that for the generic ones.

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+  RC_ASSERT(equal(getParentDigest(block), parentDigest));
+  RC_ASSERT(getStateHash(block) == stateHash);
+}
+
+// Test that the blockchain property is maintained. In pseudocode:
+//  * [getData(block) == blockUpdate] - the passed updates are the same as the fetched ones after adding the block
+//  * [getParentDigest(block) == blockDigest(prevBlock)] - blockchain property
+//  * [getLastReachableBlock() == getLatestBlock() == blockUpdates.size()] - last reachable block is equal to the latest
+//  know block ID
+TEST_P(db_adapter_block_tests, blockchain_property) {
```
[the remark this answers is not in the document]
dartdart26: It is a GTest macro that allows testing with parameters. I don't see a corresponding one in rapidcheck and that is why I use it and call `ASSERT_TRUE(rc::check(test));` explicitly.

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
dartdart26: Great catch! What I meant to write is `adapter.getLastReachableBlock() + stateTransferOffset + i;` . In that case, `getLastReachable() + stateTransferOffset` should be constant during the loop and we should be adding blocks at a constant offset. We should change the names of these methods...

I will make this test clearer.

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
dartdart26: Agreed. I have made it clearer. Please see the comment above.

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
dartdart26: Done. Also changed the test so that we explicitly have a gap of at least 2 blocks. That way, we will expect that the chains are not linked at the end of the test. Linking is tested in the state transfer simulation test. Another test we might want is a test that adds blocks in random order and verifies that linking works.

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    auto adapter = DBAdapter{GetParam()->db()};
+    RC_ASSERT(addBlocks(blockUpdates, adapter));
+
+    for (auto i = 0ul; i < blockUpdates.size(); ++i) {
+      for (const auto &kv : blockUpdates[i]) {
+        auto value = Value{};
+        auto actualBlockId = BlockId{0};
+        RC_ASSERT(adapter.getKeyByReadVersion(i + 1, kv.first, value, actualBlockId).isOK());
+        const auto [referenceBlockId, referenceValue] = getValueByKeyAndVersion(blockUpdates, kv.first, i + 1);
```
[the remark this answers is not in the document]
dartdart26: Great catch, thanks! Doing as you suggest.

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+      }
+    }
+    RC_ASSERT(adapter.getLastReachableBlock() == adapter.getLatestBlock());
+    RC_ASSERT(adapter.getLatestBlock() == latestBlockId);
+  };
+
+  ASSERT_TRUE(rc::check(test));
+}
+
+// Test that getting the value of a key by version V returns the value of the key at version Y such that Y <= V .
```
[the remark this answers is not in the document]
dartdart26: You are absolutely right. Splitting to two tests as you suggest:
- `get_key_at_version_it_was_written_at`
- `get_all_keys_at_last_version`

Thanks!

storage/test/sparse_merkle_storage/db_adapter_property_test.cpp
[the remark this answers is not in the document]
andrewjstone: Ah, thanks for the explanation.

PR #484, opened 2020-03-25
[Apollo] Fast and slow commit path test improvements

teoparvanov:
The following improvements are made to the existing tests:
- make the tests non-trivial by increasing the number of processed requests
- make them more stable and deterministic
- accelerate by removing static `trio.sleep()` statements
- where possible, make the tests idempotent, so they can run as part of product CI, or even long-running tests.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -608,8 +608,8 @@ class BftTestNetwork:
                     try:
                         await self.assert_slow_path_prevalent(
                             as_of_seq_num, nb_slow_paths_so_far, replica_id)
-                    except KeyError:
-                        # metrics not yet available, continue looping
+                    except (KeyError, AssertionError):
+                        # continue polling
                         continue
                     else:
                         # slow path prevalent - done.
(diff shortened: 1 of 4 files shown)


teoparvanov:
> 
> 
> Everything looks good to me, thanks.

Thanks for looking into this!

tests/apollo/util/bft.py
```diff
@@ -692,12 +706,16 @@ def force_quorum_including_replica(self, replica_id, primary=0):
...
+        with trio.fail_after(seconds=5):
+            while True:
+                with trio.move_on_after(seconds=.5):
+                    try:
+                        metric_key = ['replica', 'Counters', 'slowPathCount']
+                        nb_slow_path = await self.metrics.get(0, *metric_key)
+                        return nb_slow_path
+                    except KeyError:
+                        # metrics not yet available, continue looping
+                        pass
```
andrewjstone: Nit: I guess pass is interchangeable with continue here. I think continue might be clearer, but it doesn't really matter.

teoparvanov: Yes, continue would be a bit better - but it's equivalent in this case. I'd prefer not to change it right now.

PR #485, opened 2020-03-26
Add minor improvements ot the kvbc library

dartdart26:
Improve the following:
 * link corebft to kvbc as kvbc uses it
 * remove the readOnly parameter from DBAdapters - it doesn't make sense
   to use them in read-only mode
 * use an std::unique_ptr<IDataKeyGenerator> instead of a raw pointer in
   the v1DirectKeyValue::DBAdapter's ctor to make it clear that the
   adapter owns the generator

### DIFF
--- a/kvbc/CMakeLists.txt
+++ b/kvbc/CMakeLists.txt
@@ -16,7 +16,7 @@ add_library(kvbc  src/ClientImp.cpp
                   src/sparse_merkle/walker.cpp
 )
 
-target_link_libraries(kvbc PUBLIC threshsign util concordbft_storage)
+target_link_libraries(kvbc PUBLIC threshsign util concordbft_storage corebft)
 
 target_include_directories(kvbc PUBLIC include)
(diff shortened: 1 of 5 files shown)


kvbc/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ add_library(kvbc  src/ClientImp.cpp
                   src/sparse_merkle/walker.cpp
 )
 
-target_link_libraries(kvbc PUBLIC threshsign util concordbft_storage)
+target_link_libraries(kvbc PUBLIC threshsign util concordbft_storage corebft)
```
toly-kournik: consider arget_link_libraries(kvbc PUBLIC corebft)

since corebft is linked with threshsign util concordbft_storage

dartdart26: I am thinking that if the `corebft` depdendencies change in the future, `kvbc` will still work if we leave them explicit here.

toly-kournik: it's up to you

PR #486, opened 2020-03-27
Initial tests for ViewChangeSafetyLogic.

HristoStaykov:
Added initial 2 tests for ViewChangeSafety logic.


bftengine/tests/CMakeLists.txt
```diff
@@ -2,3 +2,4 @@ add_subdirectory(simpleStorage)
 add_subdirectory(bcstatetransfer)
 add_subdirectory(testSerialization)
 add_subdirectory(metadataStorage)
+add_subdirectory(testViewChangeSafetyLogic)
```
teoparvanov: let's just call this viewChange, as it will also host other VC-related tests

PR #487, opened 2020-03-27
Make the rapidcheck dependency optional

dartdart26:
In order to support a smooth transition for pre-Conan users of the
library, make the rapidcheck dependency optional. Future commits may
make it required again.


andrewjstone:
Thanks @dartdart26

PR #488, opened 2020-03-27
[Apollo] Bug fixes following preliminary runs of the checkpoint test …

teoparvanov:
…in product CI

This PR contains the following fixes:
- [BFT code] Fix a corner case when `checkpointNumber` is equal to `maxNumOfStoredCheckpoints_`, resulting in a breach of the following invariant: `[Assert(lastStoredCheckpoint - firstStoredCheckpoint + 1 <= maxNumOfStoredCheckpoints_);`
https://github.com/vmware/concord-bft/blob/master/bftengine/src/bcstatetransfer/BCStateTran.cpp#L2123
- [Test code] Made `_wait_to_receive_st_msgs()` more robust around replica restarts, when the metrics are not immediately available

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -440,7 +440,7 @@ STDigest BCStateTran::checkpointReservedPages(uint64_t checkpointNumber, DataSto
 // Remove old checkpoints from the data store
 void BCStateTran::deleteOldCheckpoints(uint64_t checkpointNumber, DataStoreTransaction *txn) {
   uint64_t minRelevantCheckpoint = 0;
-  if (checkpointNumber > maxNumOfStoredCheckpoints_) {
+  if (checkpointNumber >= maxNumOfStoredCheckpoints_) {
     minRelevantCheckpoint = checkpointNumber - maxNumOfStoredCheckpoints_ + 1;
   }
 
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -530,10 +530,13 @@ class BftTestNetwork:
         """
         while True:
             with trio.move_on_after(.5): # seconds
-                key = ['replica', 'Counters', 'receivedStateTransferMsgs']
-                n = await self.metrics.get(replica.id, *key)
-                if n > 0:
-                    cancel_scope.cancel()
+                try:
+                    key = ['replica', 'Counters', 'receivedStateTransferMsgs']
+                    n = await self.metrics.get(replica.id, *key)
+                    if n > 0:
+                        cancel_scope.cancel()
+                except KeyError:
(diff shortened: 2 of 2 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -440,7 +440,7 @@ STDigest BCStateTran::checkpointReservedPages(uint64_t checkpointNumber, DataSto
 // Remove old checkpoints from the data store
 void BCStateTran::deleteOldCheckpoints(uint64_t checkpointNumber, DataStoreTransaction *txn) {
   uint64_t minRelevantCheckpoint = 0;
-  if (checkpointNumber > maxNumOfStoredCheckpoints_) {
+  if (checkpointNumber >= maxNumOfStoredCheckpoints_) {
```
andrewjstone: I fail to understand how this makes any difference, in light of the fact that right below, we ignore `minRelevantCheckpoint < 2`. With 1minRelevantCheckpoint = 3` and `checkpointNumber = 4`, this should still only store checkpoints 2-4, which is 3 checkpoints. 

I think something more complicated is going on here. It's almost like the transaction isn't working as expected.

teoparvanov: Hi Andrew, thanks for looking into this. I believe the problem is actually quite simple and it is indeed a corner case. 
Let's put this into concrete terms. When the assert fails, we have the following state:
`2020-03-27T10:53:32.975 [139864041720448] INFO  state-transfer %% ********* firstStoredCheckpoint = 0, lastStoredCheckpoint = 3, psd_->hasCheckpointDesc(firstStoredCheckpoint) = 0, maxNumOfStoredCheckpoints_ = 3`

In this case indeed, `lastStoredCheckpoint - firstStoredCheckpoint + 1 = 4 > 3 = maxNumOfStoredCheckpoints_` which is a failure of the invariant.

This "failed" state is possible because of the "if" statement I'm fixing in this PR. If we keep the strict inequality `>` at line 443, then `minRelevantCheckpoint=0` which means that `firstStoredCheckpoint=0` remains and the 0-th checkpoint is not deleted.

To summarize - it's not an issue with the transaction, but rather with the logic that deletes old checkpoints.

andrewjstone: Aha. I didn't consider that firstStoredCheckpoint could equal 0. Thank you for pointing that out Teo. So what we really are doing here is ensuring that we are setting `firstStoredCheckpoint` to 1 immediately. 

I find this logic confusing with a few to many conditionals. Nice job tracking this down and fixing it.  There really is no such thing as a 0th stored checkpoint which is why this is so confusing. It's a failure of the unsigned type having a bottom value that is a valid number, but not really a valid checkpoint. Using a `std::optional` would have made this explicit.

PR #491, opened 2020-03-28
PrePrepareMsg: Allocate based on requests

robem:
Prior to this change, every (non-null) PrePrepareMsg allocates the maximum
message size upfront. Depending on the maximum this introduces a lot of
overhead. With this change, we allocate the amount of memory based on all
client requests that are in the queue. However, we still won't allocate more
than the maximum external message size.


robem:
Locally some tests for `make test` are failing. I need to look into this.

robem:
> Locally some tests for `make test` are failing. I need to look into this.

All tests pass. This was a local setup issue.

andrewjstone:
Yay!

PR #494, opened 2020-03-30
Introduce IDbAdapter interface.

toly-kournik:
The new interface is used for easy transition from existing key-value bloclchain to a merkle tree representation,
as well as to strore blockchain in different types of storage.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -265,58 +265,64 @@ void BCStateTran::loadMetrics() {
 void BCStateTran::init(uint64_t maxNumOfRequiredStoredCheckpoints,
                        uint32_t numberOfRequiredReservedPages,
                        uint32_t sizeOfReservedPage) {
-  Assert(!running_);
-  Assert(replicaForStateTransfer_ == nullptr);
-  Assert(sizeOfReservedPage == sizeOfReservedPage_);
+  try {
+    Assert(!running_);
+    Assert(replicaForStateTransfer_ == nullptr);
+    Assert(sizeOfReservedPage == sizeOfReservedPage_);
 
-  maxNumOfStoredCheckpoints_ = maxNumOfRequiredStoredCheckpoints;
-  numberOfReservedPages_ = numberOfRequiredReservedPages;
-  metrics_.number_of_reserved_pages_.Get().Set(numberOfReservedPages_);
-  metrics_.size_of_reserved_page_.Get().Set(sizeOfReservedPage_);
+    maxNumOfStoredCheckpoints_ = maxNumOfRequiredStoredCheckpoints;
+    numberOfReservedPages_ = numberOfRequiredReservedPages;
+    metrics_.number_of_reserved_pages_.Get().Set(numberOfReservedPages_);
+    metrics_.size_of_reserved_page_.Get().Set(sizeOfReservedPage_);
 
-  memset(buffer_, 0, maxItemSize_);
+    memset(buffer_, 0, maxItemSize_);
 
-  LOG_INFO(STLogger,
-           "Init BCStateTran object:"
-               << " maxNumOfStoredCheckpoints_=" << maxNumOfStoredCheckpoints_ << " numberOfReservedPages_="
-               << numberOfReservedPages_ << " sizeOfReservedPage_=" << sizeOfReservedPage_);
+    LOG_INFO(STLogger,
+             "Init BCStateTran object:"
+                 << " maxNumOfStoredCheckpoints_=" << maxNumOfStoredCheckpoints_ << " numberOfReservedPages_="
+                 << numberOfReservedPages_ << " sizeOfReservedPage_=" << sizeOfReservedPage_);
 
-  if (psd_->initialized()) {
-    LOG_INFO(STLogger, "BCStateTran::init - loading existing data from storage");
+    if (psd_->initialized()) {
+      LOG_INFO(STLogger, "BCStateTran::init - loading existing data from storage");
 
-    checkConsistency(pedanticChecks_);
+      checkConsistency(pedanticChecks_);
 
-    FetchingState fs = getFetchingState();
-    LOG_INFO(STLogger, "starting state is " << stateName(fs));
+      FetchingState fs = getFetchingState();
+      LOG_INFO(STLogger, "starting state is " << stateName(fs));
 
-    if (fs == FetchingState::GettingMissingBlocks || fs == FetchingState::GettingMissingResPages) {
-      SetAllReplicasAsPreferred();
+      if (fs == FetchingState::GettingMissingBlocks || fs == FetchingState::GettingMissingResPages) {
+        SetAllReplicasAsPreferred();
+      }
+      loadMetrics();
+    } else {
+      LOG_INFO(STLogger, "BCStateTran::init - initializing a new object");
+
+      AssertAND(maxNumOfRequiredStoredCheckpoints >= 2,
+                maxNumOfRequiredStoredCheckpoints <= kMaxNumOfStoredCheckpoints);
+
+      AssertAND(numberOfRequiredReservedPages >= 2, numberOfRequiredReservedPages <= maxNumOfReservedPages_);
+      DataStoreTransaction::Guard g(psd_->beginTransaction());
+      g.txn()->setReplicas(replicas_);
+      g.txn()->setMyReplicaId(myId_);
+      g.txn()->setFVal(fVal_);
+      g.txn()->setMaxNumOfStoredCheckpoints(maxNumOfRequiredStoredCheckpoints);
+      g.txn()->setNumberOfReservedPages(numberOfRequiredReservedPages);
+      g.txn()->setLastStoredCheckpoint(0);
+      g.txn()->setFirstStoredCheckpoint(0);
+      // TODO(TK) - check max transaction size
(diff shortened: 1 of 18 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -265,58 +265,64 @@ void BCStateTran::loadMetrics() {
...
+      g.txn()->setFirstRequiredBlock(0);
+      g.txn()->setLastRequiredBlock(0);
+      g.txn()->setAsInitialized();
+
+      Assert(getFetchingState() == FetchingState::NotFetching);
+    }
+  } catch (const std::exception &e) {
+    LOG_FATAL(STLogger, e.what());
+    exit(1);
   }
```
dartdart26: Maybe we can `catch(...)` just in case.

bftengine/src/bftengine/IncomingMsgsStorageImp.cpp
```diff
@@ -126,39 +126,37 @@ IncomingMsg IncomingMsgsStorageImp::popThreadLocal() {
...
-        msg.internal->handle();
-    };
+          break;
+        case IncomingMsg::INTERNAL:
+          msg.internal->handle();
+      };
+    }
+  } catch (const std::exception& e) {
+    LOG_FATAL(GL, "Exception: " << e.what() << "exiting ...");
+    exit(1);
```
dartdart26: Same as above - maybe we should `catch (...)` too.

kvbc/src/ReplicaImp.cpp
```diff
@@ -128,14 +128,13 @@ Status ReplicaImp::getBlockData(BlockId blockId, SetOfKeyValuePairs &outBlockDat
   // TODO(GG): check legality of operation (the method should be invoked from
   // the replica's internal thread)
 
-  Sliver block = getBlockInternal(blockId);
-
-  if (block.length() == 0) {
+  try {
+    Sliver block = getBlockInternal(blockId);
+    outBlockData = block::getData(block);
+  } catch (const NotFoundException &e) {
```
dartdart26: I am thinking if it will it be safe to throw any other exceptions out of this function. Most of the code doesn't expect exceptions and it might be safer to catch all exceptions here, at least for now. What do you think?

kvbc/include/db_adapter.h
```diff
@@ -4,5 +4,51 @@
...
+  virtual RawBlock getRawBlock(const BlockId& blockId) const = 0;
+
+  // If found, return the actual version and the value.
+  virtual std::optional<std::pair<Value, BlockId>> getValue(const Key&, const BlockId& blockVersion) const = 0;
+
+  // Delete a block from the database
+  virtual void deleteBlock(const BlockId& blockId) = 0;
+
+  // Used to retrieve the latest block.
+  virtual BlockId getLastBlockchainBlockId() const = 0;
```
dartdart26: I think naming it that way is confusing as `blockchain` implies part of the blockchain. In case of state transfer, as blocks arrive in reverse order, they are not yet part of the blockchain. This is valid for the merkle tree DB adapter, not necessarily valid for the direct KV adapter. Nevertheless, my suggestion is to have a distinction between blocks that are part of the blockchain and ones that are still in the process of ST, i.e. calling these methods as:

```
// equivalent to getLastRechableBlockId(), i.e. returning the ID of the last block that is part of the blockchain
virtual BlockId getLastBlockchainBlockId() const = 0;

// returns the latest known block ID (in case ST doesn't run in reverse order, it will return the received block with the highest ID)
virtual BlockId getLatestKnownBlockId() const = 0;
```

I am open to other naming suggestions too, but would prefer that it reflects the distinction mentioned above.

kvbc/include/db_adapter.h
```diff
@@ -4,5 +4,51 @@
...
+
+  /**
+   * Used to retrieve the last reachable block
+   * Searches for the last reachable block.
+   * From ST perspective, this is maximal block number N
+   * such that all blocks 1 <= i <= N exist.
+   * In the normal state, it should be equal to last block ID.
+   */
```
dartdart26: I think this comment can be made clearer in terms of reachable and last. Also, when we start deleting blocks, genesis block will not always have ID = 1.

kvbc/include/db_adapter.h
toly-kournik: I think lastReachableBlock reflects better its meaning. state transfer may "know" the real last block and in case of ST in progress they are not equal. Besides it's something in common termonology

kvbc/include/db_adapter.h
toly-kournik: yep, from the current beginning

kvbc/src/ReplicaImp.cpp
```diff
@@ -277,39 +259,20 @@ void ReplicaImp::insertBlockInternal(BlockId blockId, Sliver block) {
...
-    // TODO(SG): To do something smarter
-    LOG_ERROR(logger, "An error occurred in get block");
-    return Sliver();
-  }
-
-  if (!found) {
-    return Sliver();
-  } else {
-    return retVal;
-  }
```
dartdart26: Nice! :)

bftengine/src/bcstatetransfer/BCStateTran.cpp
toly-kournik: I don't think so, there's no exceptions other than std::exception thrown.

kvbc/src/ReplicaImp.cpp
toly-kournik: In every other case the application has exited; now the runtime_exception will be caught at the topmost layer

toly-kournik: other exceptions here means problems with db/consistency

bftengine/src/bcstatetransfer/BCStateTran.cpp
dartdart26: Yep, I know, but I was saying that if somebody gets creative and does that in the future, we will not exit here.

kvbc/include/db_adapter.h
toly-kournik: done

storage/include/storage/db_interface.h
```diff
@@ -62,7 +62,6 @@ class IDBClient {
   virtual Status multiGet(const KeysVector& _keysVec, OUT ValuesVector& _valuesVec) = 0;
   virtual Status multiPut(const SetOfKeyValuePairs& _keyValueMap) = 0;
   virtual Status multiDel(const KeysVector& _keysVec) = 0;
-  virtual void monitor() const = 0;
```
dartdart26: Nice, I agree this method is unclear and, I think, unused.

kvbc/include/db_adapter.h
```diff
@@ -4,5 +4,55 @@
...
+
+  // Used to retrieve the latest block.
+  virtual BlockId getLastestBlockId() const = 0;
+
+  // Used to retrieve the last reachable block.
+  // From ST perspective, this is maximal block number N such that all blocks
+  // START <= i <= N exist, where START is usually 1, if pruning is not enabled.
+  // In the normal state, it should be equal to last block ID.
+
+  virtual BlockId getLastReachableBlockId() const = 0;
```
andrewjstone: Nit: comment should have no space before function.

kvbc/include/direct_kv_db_adapter.h
```diff
@@ -54,21 +54,9 @@ class DBKeyManipulator : public DBKeyManipulatorBase {
...
-  static uint64_t extractCheckPointFromKey(const char *_key_data, size_t _key_length);
-  static std::pair<uint32_t, uint64_t> extractPageIdAndCheckpointFromKey(const char *_key_data, size_t _key_length);
-
- private:
-  static concordUtils::Sliver generateReservedPageKey(EDBKeyType, uint32_t pageid, uint64_t chkpt);
-  static bool copyToAndAdvance(char *_buf, size_t *_offset, size_t _maxOffset, char *_src, size_t _srcSize);
 };
 
-class DBAdapter : public DBAdapterBase {
+class DBAdapter : public IDbAdapter, public DBAdapterBase {
```
andrewjstone: Can we avoid the use of multiple inheritance? It's can be especially confusing to inherit from an abstract class and a concrete class. Maybe in a future commit we can pull out all the shared information in DBAdapterBase and instead use composition.

kvbc/src/ReplicaImp.cpp
```diff
@@ -277,39 +259,20 @@ void ReplicaImp::insertBlockInternal(BlockId blockId, Sliver block) {
...
     }
-  } else {
-    s = m_bcDbAdapter->addBlock(block, blockId);
-    if (!s.isOK()) {
-      // TODO(SG): What to do?
-      printf("Failed to add block");
-      exit(1);
-    }
+  } catch (const NotFoundException &e) {
+    m_bcDbAdapter->addRawBlock(block, blockId);  // TODO [TK] will be replaced after integration of hasBlock
```
andrewjstone: Is the expectation that this throws on error and is purposefully uncaught?

kvbc/include/direct_kv_db_adapter.h
toly-kournik: Sure, we have already discussed this with @dartdart26

kvbc/src/ReplicaImp.cpp
toly-kournik: Yes, previously there were exit(1) directly in places they were thrown, and now the exceptions are caught at the most outer layer

PR #496, opened 2020-04-01
Improve merkle storage interfaces and namespaces

dartdart26:
Make v2MerkleTree::DBAdapter an implementation of the IDbAdapter
interface. The deleteBlock() method will be provided in a separate
commit.

Cleanup DBAdapterBase and inherit privately from it in DB adapter
implementations.

Move concord::storage::sparse_merkle to concord::kvbc::sparse_merkle.
Move concord::storage::blockchain::v2MerkleTree to
concord::kvbc::v2MerkleTree. The v2MerkleTree namespace is going to be
renamed in a future commit.

Fix a typo in IDbAdapter - getLastestBlockId() -> getLatestBlockId().

Rework merkle tree DB adapter tests so that they use the new interfaces.
Add a test for the return value of the DBAdapter::addBlock() method.


kvbc/include/base_db_adapter.h
```diff
@@ -23,18 +23,9 @@ class DBAdapterBase {
...
   }
 
- public:
-  std::shared_ptr<storage::IDBClient> getDb() const { return db_; }
-
-  storage::IDBClient::IDBClientIterator *getIterator() { return db_->getIterator(); }
-
-  concordUtils::Status freeIterator(storage::IDBClient::IDBClientIterator *_iter) { return db_->freeIterator(_iter); }
-
  protected:
```
toly-kournik, line 26: Do we still need this class at all?

kvbc/include/merkle_tree_block.h
```diff
@@ -12,29 +12,25 @@
...
-namespace block {
+namespace concord::kvbc::v2MerkleTree::block {
 
 inline constexpr auto BLOCK_DIGEST_SIZE = bftEngine::SimpleBlockchainStateTransfer::BLOCK_DIGEST_SIZE;
 
 // Creates a block that includes a set of key/values. The passed parentDigest buffer must be of size BLOCK_DIGEST_SIZE
 // bytes.
 
-concordUtils::Sliver create(const concord::kvbc::SetOfKeyValuePairs &updates,
+concordUtils::Sliver create(const SetOfKeyValuePairs &updates,
```
toly-kournik, line 22: RawBlock instead of Sliver?

kvbc/include/merkle_tree_block.h
```diff
@@ -12,29 +12,25 @@
...
                             const void *parentDigest,
                             const sparse_merkle::Hash &stateHash);
 
 // Returns the block data in the form of a set of key/value pairs.
-concord::kvbc::SetOfKeyValuePairs getData(const concordUtils::Sliver &block);
+SetOfKeyValuePairs getData(const RawBlock &block);
 
 // Returns the parent digest of size BLOCK_DIGEST_SIZE bytes.
-const void *getParentDigest(const concordUtils::Sliver &block);
+const void *getParentDigest(const RawBlock &block);
```
toly-kournik, line 30: I know it's not you but why is it void*?

kvbc/include/merkle_tree_db_adapter.h
```diff
@@ -15,120 +16,117 @@
...
-namespace concord {
-namespace storage {
-namespace blockchain {
-namespace v2MerkleTree {
+namespace concord::kvbc::v2MerkleTree {
 
 class DBKeyManipulator {
  public:
-  static concordUtils::Sliver genBlockDbKey(kvbc::BlockId version);
+  static concordUtils::Sliver genBlockDbKey(BlockId version);
```
toly-kournik, line 23: These gen...() functions should return Key

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -300,21 +290,24 @@ BlockId DBAdapter::getLatestBlock() const {
...
-    Sliver parentBlock;
-    const auto status = getBlockById(blockId - 1, parentBlock);
-    if (!status.isOK() || parentBlock.empty()) {
-      LOG_FATAL(logger_, "createBlockNode: no block or block data for parent block ID " << blockId - 1);
+    try {
+      const auto parentBlock = getRawBlock(blockId - 1);
+      computeBlockDigest(blockId - 1, parentBlock.data(), parentBlock.length(), &parentBlockDigest);
+    } catch (const std::exception &e) {
+      LOG_FATAL(logger_, "Failed to get parent block ID = " + std::to_string(blockId - 1) + ", reason: " + e.what());
+      std::exit(1);
```
toly-kournik, line 306: I prefer to have exit(1) in one place where the exceptions are caught

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -325,14 +318,22 @@ Sliver DBAdapter::createBlockNode(const SetOfKeyValuePairs &updates, BlockId blo
...
   const auto blockKey = DBKeyManipulator::genBlockDbKey(blockId);
   Sliver blockNodeSliver;
-  if (auto status = db_->get(blockKey, blockNodeSliver); status.isNotFound()) {
+  if (auto statusNode = db_->get(blockKey, blockNodeSliver); statusNode.isNotFound()) {
     // If a block node is not found, look for a state transfer block.
-    return db_->get(DBKeyManipulator::generateSTTempBlockKey(blockId), block);
-  } else if (!status.isOK()) {
-    return status;
+    Sliver block;
+    if (auto statusSt = db_->get(DBKeyManipulator::generateSTTempBlockKey(blockId), block); statusSt.isNotFound()) {
```
toly-kournik, line 327: generateSTTempBlockKey() - is it a state transfer related function?

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -399,45 +399,47 @@ BatchedInternalNode DBAdapter::Reader::get_internal(const InternalNodeKey &key)
   return deserialize<BatchedInternalNode>(res);
 }
 
-Status DBAdapter::addLastReachableBlock(const SetOfKeyValuePairs &updates) {
+BlockId DBAdapter::addBlock(const SetOfKeyValuePairs &updates) {
   for (const auto &kv : updates) {
     if (kv.second.empty()) {
       const auto msg = "Adding empty values in a block is not supported";
       LOG_ERROR(logger_, msg);
```
toly-kournik, line 406: I think we don't need to log error twice - it's logged in place it's caught

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -448,48 +450,44 @@ Status DBAdapter::writeSTLinkTransaction(const Key &sTBlockKey, const Sliver &bl
...
+  txn->commit();
 }
 
-Status DBAdapter::addBlock(const Sliver &block, BlockId blockId) {
-  const auto lastReachableBlock = getLastReachableBlock();
+void DBAdapter::addRawBlock(const RawBlock &block, const BlockId &blockId) {
+  const auto lastReachableBlock = getLastReachableBlockId();
   if (blockId <= lastReachableBlock) {
     const auto msg = "Cannot add an existing block ID " + std::to_string(blockId);
     LOG_ERROR(logger_, msg);
```
toly-kournik, line 461: same as above

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -448,48 +450,44 @@ Status DBAdapter::writeSTLinkTransaction(const Key &sTBlockKey, const Sliver &bl
...
-    if (!status.isOK()) {
-      LOG_ERROR(logger_, "Failed to add a block to the end of the blockchain on state transfer, block ID " << blockId);
-      return status;
+    addBlock(block::getData(block));
+
+    try {
+      linkSTChainFrom(blockId + 1);
+    } catch (const std::exception &e) {
+      LOG_FATAL(logger_, "Aborting due to failure to link chains after block has been added, reason: "s + e.what());
+      std::exit(1);
```
toly-kournik, line 472: same as above about exit(1)

kvbc/include/base_db_adapter.h
dartdart26: Good point. I don't think so, removing it.

kvbc/include/merkle_tree_block.h
dartdart26: Good point, yes, doing that.

kvbc/include/merkle_tree_block.h
dartdart26: The idea is to just treat as a buffer. We should wrap it in some proper type.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: It is not used by state transfer itself. Instead, it is used by the merkle tree DBAdapter to accumulate blocks as they arrive from state transfer.

kvbc/include/merkle_tree_db_adapter.h
dartdart26: Good point, done.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: I agree - in that particular case we can just throw, doing that.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: This is a special case in that if we don't exit at the spot, we risk leaving the system in an invalid state, i.e. the block has been added, but the chains are not linked. If we just throw, we rely on the exception handler terminating the program and I prefer we be explicit here so that future code changes won't silently introduce a bug. When starting the system after the exit() call, the DBAdapter constructor will link the chains.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: I think you are right. However, if there is a bug and there is a missing exception handler or a missing log call in the handler, we might miss the event. Additionally, since these are errors and won't happen often it wouldn't be a big burden on the log. I will leave them for now and if we have an issue, we can easily remove.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: Please see comment above.

dartdart26:
> I know that it's a temporary state refactoring, so it's looks good for me now.
> I have some minor comments/questions which can be addressed in a subsequent PRs.

Great comments, thanks @toly-kournik ! I have addressed them in the last commit.

PR #500, opened 2020-04-03
[SparseMerkle] Fix internal node property test

andrewjstone:
The test ``any_insert_order_of_unique_keys_maintains_same_keys` was
failing due to different version order of the LeafKeys. We must ignore
versions and only ensure the hashes of the keys are different for this
to work.

The seed of the failure for GCC builds only was: `5478480958984474772`.

Also, change the name of the fixed test to be more accurate.

Additionally, fix an unused variable error in release mode.


andrewjstone:
This is a fix for the failing job seen in #499.

PR #502, opened 2020-04-05
MsgsCertificate Test

oded282:
- Created unit tests for 'MsgsCertificare'.
- Repaired 'InConsistent' function in 'MsgsCertificate',
 and tested the desirable consistency behaviour.
- Include needed files.

### DIFF
--- a/bftengine/src/bftengine/messages/MsgsCertificate.hpp
+++ b/bftengine/src/bftengine/messages/MsgsCertificate.hpp
@@ -288,7 +288,7 @@ bool MsgsCertificate<T, SelfTrust, SelfIsRequired, KeepAllMsgs, ExternalFunc>::i
 
 template <typename T, bool SelfTrust, bool SelfIsRequired, bool KeepAllMsgs, typename ExternalFunc>
 bool MsgsCertificate<T, SelfTrust, SelfIsRequired, KeepAllMsgs, ExternalFunc>::isInconsistent() const {
-  return (numOfClasses >= maxFails + 1);
+  return (numOfClasses > maxFails + 1);
 }
 
 template <typename T, bool SelfTrust, bool SelfIsRequired, bool KeepAllMsgs, typename ExternalFunc>
--- a/bftengine/tests/testMsgsCertificate/CMakeLists.txt
+++ b/bftengine/tests/testMsgsCertificate/CMakeLists.txt
@@ -3,6 +3,12 @@ find_package(GTest REQUIRED)
 add_executable(msgsCertificate_test msgsCertificate_test.cpp $<TARGET_OBJECTS:logging_dev>)
 add_test(msgsCertificate_test msgsCertificate_test)
 
+target_include_directories(msgsCertificate_test
+        PRIVATE
+        ${bftengine_SOURCE_DIR}/src/preprocessor
+
+        )
+
 target_link_libraries(msgsCertificate_test PUBLIC
         GTest::Main
         corebft
--- a/bftengine/tests/testMsgsCertificate/msgsCertificate_test.cpp
+++ b/bftengine/tests/testMsgsCertificate/msgsCertificate_test.cpp
@@ -1,8 +1,146 @@
+#include <messages/CheckpointMsg.hpp>
 #include "gtest/gtest.h"
+#include "SimpleClientImp.cpp"
+#include "messages/MsgsCertificate.hpp"
+#include "messages/ClientReplyMsg.hpp"
+#include "CheckpointInfo.hpp"
+#include "Digest.hpp"
 
-using namespace std;
+using namespace bftEngine;
 
-TEST(msgsCertificate_test, initial_test) {}
+namespace {
+
+    const uint16_t fval = 1;
+    const uint16_t cval = 1;
+    const uint16_t numOfReplicas = 3 * fval + 2 * cval + 1;
+    const uint16_t numOfRequired = 2 * fval + cval + 1;
+    const uint16_t selfReplicaId = 0;
+    //char reply[] = "1234567890";
+    std::string reply{""};
+    int replyLen = reply.length();
+
+    ReplicaId randomReplicaeId() {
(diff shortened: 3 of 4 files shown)


bftengine/tests/testMsgsCertificate/CMakeLists.txt
```diff
@@ -3,6 +3,12 @@ find_package(GTest REQUIRED)
 add_executable(msgsCertificate_test msgsCertificate_test.cpp $<TARGET_OBJECTS:logging_dev>)
 add_test(msgsCertificate_test msgsCertificate_test)
 
+target_include_directories(msgsCertificate_test
+        PRIVATE
+        ${bftengine_SOURCE_DIR}/src/preprocessor
```
teoparvanov, line 8: why do you need the preprocessor as a dependency ?

bftengine/src/bftengine/messages/MsgsCertificate.hpp
```diff
@@ -288,7 +288,7 @@ bool MsgsCertificate<T, SelfTrust, SelfIsRequired, KeepAllMsgs, ExternalFunc>::i
 
 template <typename T, bool SelfTrust, bool SelfIsRequired, bool KeepAllMsgs, typename ExternalFunc>
 bool MsgsCertificate<T, SelfTrust, SelfIsRequired, KeepAllMsgs, ExternalFunc>::isInconsistent() const {
-  return (numOfClasses >= maxFails + 1);
+  return (numOfClasses > maxFails + 1);
```
teoparvanov, line 291: which test case discovered this problem?

@yuliasherman this looks like a missed corner case - what do you think?

oded282: - I read the code, and understood the number doesn't fit the protocol.
  Also checked with Guy to be sure, he confirmed that indeed this is incorrect.

teoparvanov: Great job! Can you also write a unit test that specifically reproduces this problem?

teoparvanov: As discussed offline, the test already exists, and will be mentioned in the PR description.

bftengine/tests/testMsgsCertificate/CMakeLists.txt
oded282: In order to use 'ClientReplyMsg' used in my test, I have needed to include '${bftengine_SOURCE_DIR}/src/preprocessor'.

bftengine/tests/testMsgsCertificate/msgsCertificate_test.cpp
```diff
@@ -1,10 +1,132 @@
...
+
+  uint8_t reqSeqNum = 0;
+
+  // Create two identical messages and add them.
+
+  ReplicaId replicaId = randomReplicaeId();
+  auto bestMsg = new ClientReplyMsg(replicaId, reqSeqNum, reply, replyLen);
+  replysCertificate.addMsg(bestMsg, randomReplicaeId());
+
+  replicaId = randomReplicaeId();
```
teoparvanov, line 93: nit - `randomReplicaeId` -> `randomReplicaId`

bftengine/tests/testMsgsCertificate/msgsCertificate_test.cpp
```diff
@@ -1,10 +1,132 @@
...
+      numOfReplicas, fval, numOfRequired, selfReplicaId);
+
+  uint8_t reqSeqNum = 0;
+
+  // Add bunch of new messages
+  for (; reqSeqNum < 3; reqSeqNum++) {
+    ReplicaId replicaId = randomReplicaeId();
+    auto newMsg = new ClientReplyMsg(replicaId, reqSeqNum, reply, replyLen);
+    replysCertificate.addMsg(newMsg, replicaId);
+  }
```
HristoStaykov, line 41: I think the "replicaId" shouldn't be drawn from the RNG. If I understand correct we want different replica ID to add in every round.

bftengine/tests/testMsgsCertificate/msgsCertificate_test.cpp
```diff
@@ -1,10 +1,132 @@
...
+  replicaId = randomReplicaeId();
+  auto identBestMsg = new ClientReplyMsg(replicaId, reqSeqNum, reply, replyLen);
+  replysCertificate.addMsg(identBestMsg, randomReplicaeId());
+
+  // Add bunch of new messages
+  for (; reqSeqNum < 4; reqSeqNum++) {
+    replicaId = randomReplicaeId();
+    auto newMsg = new ClientReplyMsg(replicaId, reqSeqNum, reply, replyLen);
+    replysCertificate.addMsg(newMsg, replicaId);
+  }
```
HristoStaykov, line 102: Same thing here, if we draw from RNG, it could happen that some replicaId-s will coincide.

bftengine/tests/testMsgsCertificate/CMakeLists.txt
teoparvanov: can't you use another message, which is not in the pre-processor?

oded282: Correction:
It was in order to use 'SimpleClientImp' and the method implemented there.

teoparvanov: I don't get it... SimpleClientImp is not defined in the preprocessor.

oded282: SimpleClientImp include "messages/ClientPreProcessRequestMsg.hpp".
My class use SimpleClientImp and this dir is unknown for my class, that's why I added it.

teoparvanov: Alright, I see - thanks for clarifying.

PR #503, opened 2020-04-06
Introduce a diagnostics library

andrewjstone:
This commit is the start of a diagnostics library for building
interactive CLI programs for interacting with an individual replica
built on concord-bft. The first part of the diagnostics library focuses
on retrieving status from different parts of the system in a thread-safe
manner.

This commit introduces the first two components of this library:
 * A `Registrar`, which allows registering `StatusHandler`s and
 operating on those handlers.
 * A `protocol` which defines a string based protocol for
 interacting with command line clients.

 Both of these components are tested in diagnostics_tests, which also
 serves to demonstrate usage.

 Future commits will include:
  * A unix domain socket based server for handling protocol requests
  * A simple bash or python client script for interacting with server
  * Initialization of a global Registrar
  * Implementation and registration of StatusHandlers

### DIFF
--- a/diagnostics/include/diagnostics.h
+++ b/diagnostics/include/diagnostics.h
@@ -1,0 +1,84 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+#include <functional>
+#include <map>
+#include <mutex>
+#include <stdexcept>
+
+namespace concord::diagnostics {
+
+// Every component that can return status should register a status handler with the Registrar.
+struct StatusHandler {
+  StatusHandler(std::string name, std::string description, std::function<std::string()> fun)
+      : name(std::move(name)), description(std::move(description)), f(std::move(fun)) {}
+  std::string name;
+  std::string description;
+  std::function<std::string()> f;
+};
+
+class Registrar {
+ public:
+  void registerStatusHandler(StatusHandler handler) {
+    std::lock_guard<std::mutex> guard(mutex_);
+    if (status_handlers_.count(handler.name)) {
+      throw std::invalid_argument(std::string("StatusHandler already exists: ") + handler.name);
+    }
+    status_handlers_.insert({handler.name, handler});
+  }
+
+  std::string getStatus(const std::string& name) const {
+    std::lock_guard<std::mutex> guard(mutex_);
+    if (status_handlers_.count(name)) {
+      return status_handlers_.at(name).f();
+    }
+    return "*--STATUS_NOT_FOUND--*";
+  }
+
+  std::string describeStatus(const std::string& name) const {
+    std::lock_guard<std::mutex> guard(mutex_);
+    if (status_handlers_.count(name)) {
+      return status_handlers_.at(name).description;
+    }
+    return "*--DESCRIPTION_NOT_FOUND--*";
+  }
+
+  std::string describeStatus() const {
+    std::lock_guard<std::mutex> guard(mutex_);
+    std::string output;
+    for (const auto& [_, handler]: status_handlers_) {
+      (void)_; // undefined variable hack
+      output += "\n" + handler.name + "\n  ";
+      output += handler.description + "\n";
+    }
+    return output;
+  }
+
+  std::string listStatusKeys() const {
+    std::lock_guard<std::mutex> guard(mutex_);
+    std::string output;
+    for (const auto& [key, _]: status_handlers_) {
+      (void)_; // unused variable hack
+      output += key + "\n";
+    }
+    return output;
+  }
+
+ private:
+  std::map<std::string, StatusHandler> status_handlers_;
+  mutable std::mutex mutex_;
+};
+
+}  // namespace concord::diagnostics
+
--- a/diagnostics/include/protocol.h
+++ b/diagnostics/include/protocol.h
@@ -1,0 +1,83 @@
+
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+// This file describes the wire protocol for the diagnostics server. The protocol is an ASCII based
+// protocol as it is expected to be interacted with over the command line. For security purposes we
+// expect to only run the protocol over a local unix domain socket.
+
+#pragma once
+
+#include <array>
+#include <variant>
+
+#include "diagnostics.h"
+
+namespace concord::diagnostics {
+
+std::string usage() {
+  // TODO: Make the name of the program dynamic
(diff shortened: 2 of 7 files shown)


diagnostics/include/diagnostics.h
```diff
@@ -0,0 +1,82 @@
...
+};
+
+class Registrar {
+ public:
+  void registerStatusHandler(StatusHandler handler) {
+    std::lock_guard<std::mutex> guard(mutex_);
+    if (status_handlers_.count(handler.name)) {
+      throw std::invalid_argument(std::string("StatusHandler already exists: ") + handler.name);
+    }
+    status_handlers_.insert({handler.name, handler});
```
dartdart26, line 38: [Minor]: Just a thought - that operation can be done with a single lookup by calling `insert()` directly and checking the return value to determine whether an insertion took place. I doubt there will be lots of handlers and it shouldn't have a big impact.

diagnostics/include/diagnostics.h
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  std::string getStatus(const std::string& name) const {
+    std::lock_guard<std::mutex> guard(mutex_);
+    if (status_handlers_.count(name)) {
+      return status_handlers_.at(name).f();
+    }
+    return "*--STATUS_NOT_FOUND--*";
```
dartdart26, line 46: [Minor]: Same as above - can be done with a single call to `find()`.

diagnostics/include/diagnostics.h
```diff
@@ -0,0 +1,82 @@
...
+      (void)_; // undefined variable hack
+      output += "\n" + handler.name + "\n  ";
+      output += handler.description + "\n";
+    }
+    return output;
+  }
+
+  std::string listStatusKeys() const {
+    std::string output;
+    for (const auto& [key, _]: status_handlers_) {
```
dartdart26, line 70: I think we need to lock the mutex here too as we are accessing `status_handlers_`.

diagnostics/include/diagnostics.h
```diff
@@ -0,0 +1,82 @@
...
+    }
+    return output;
+  }
+
+ private:
+  std::map<std::string, StatusHandler> status_handlers_;
+  mutable std::mutex mutex_;
+};
+
+}  // namespace concord::diagnostics
```
dartdart26, line 82: [Minor]: New line.

diagnostics/include/protocol.h
```diff
@@ -0,0 +1,80 @@
...
+    }
+    if (command == "list-keys") {
+      if (tokens.size() != 2) return usage();
+      return registrar.listStatusKeys();
+    }
+  }
+  return usage();
+}
+
+}  // namespace concord::diagnostics
```
dartdart26, line 80: [Minor]: New line.

diagnostics/test/diagnostics_tests.cpp
```diff
@@ -0,0 +1,117 @@
...
+  registrar.registerStatusHandler(async_handler2);
+
+  // No parameters trigger usage to be displayed.
+  ASSERT_EQ(0, std::memcmp("Usage:", run({}, registrar).c_str(), 6));
+
+  // Bad subject triggers usage
+  ASSERT_EQ(0, std::memcmp("Usage:", run({"no_such_subject"}, registrar).c_str(), 6));
+
+  // Good subject, bad command triggers usage
+  ASSERT_EQ(0, std::memcmp("Usage:", run({"status", "bad_command"}, registrar).c_str(), 6));
```
dartdart26, line 92: [Minor]: I think we can use string literals, i.e. `"Usage"s` to avoid specifying the length.

diagnostics/include/protocol.h
```diff
@@ -0,0 +1,80 @@
...
+    if (command == "describe") {
+      if (tokens.size() == 2) {
+        return registrar.describeStatus();
+      } else {
+        return accumulate(tokens.begin() + 2, tokens.end(), [&registrar](const auto& key) -> std::string {
+          return registrar.describeStatus(key);
+        });
+      }
+    }
+    if (command == "get") {
```
dartdart26, line 66: [Minor]: Just a thought - maybe we can use `else if` for readability here and below.

dartdart26:
@andrewjstone I also have a question - do we plan on supporting write commands? What I mean is changing settings at runtime. I assume not, but please correct me if I am wrong.

andrewjstone:
> This is great - thank you!
> 
> Minor: Function names should start with upper case
> https://google.github.io/styleguide/cppguide.html#Function_Names

Yeah, this is true, but it's basically different from almost all the rest of our code. I started doing it like a year ago but then stopped again, as nobody else seemed to be following this rule. I'm open to the change just pointing out that we should make a decision for consistency purposes.

diagnostics/include/diagnostics.h
andrewjstone: That's true, but it doesn't have the same behavior. In the code as it currently is written, the first handler is never overwritten. In that case a caller could theoretically handle the exception and try a different name, etc... It's an unlikely problem, but I prefer to keep it this way.

diagnostics/include/diagnostics.h
andrewjstone: There'd still need to be an iterator check though. I don't think it shortens things much.

diagnostics/include/diagnostics.h
andrewjstone: Great catch!

diagnostics/test/diagnostics_tests.cpp
andrewjstone: I'm not sure I follow. I just want to compare that "Usage:" is a prefix of the output from run.

diagnostics/include/protocol.h
andrewjstone: I don't like the readability either, but I'm not sure an `else if` will make it any clearer. Each clause is intended to have an early return. I'll see if adding a space makes this standout and also look at adding an `else if`.

andrewjstone:
> @andrewjstone I also have a question - do we plan on supporting write commands? What I mean is changing settings at runtime. I assume not, but please correct me if I am wrong.

Yes, I think we will eventually. They will be part of a new "subject" though. I expect a way to set log levels and also one for enabling message tracing.

andrewjstone:
Thank you very much for the reviews @dartdart26 @robem. I'll clean this up tomorrow.

diagnostics/include/diagnostics.h
dartdart26: You are absolutely right. I should stop doing reviews yearly in the morning...

diagnostics/include/diagnostics.h
dartdart26: I suggested it in order to avoid the two lookups in `count()` and `at()` and do a single lookup in `find()` instead. That is not a big deal, just a minor suggestion.

diagnostics/test/diagnostics_tests.cpp
dartdart26: Again, I should stop doing reviews early in the morning... :) My thought was to avoid `std::memcmp()`. I guess could be done by `run(...).find("Usage:") == 0`, but you can safely ignore that comment, not a big deal.

andrewjstone:
> @andrewjstone, I think we should discuss the design of this feature, as it very much resembles the metrics. It looks like both features can use the same infra both for protocol and for the udp server.

Yup! As we discussed offline, we can likely remove the aggregator and the eventual consistency from metrics by registering a callable and using that instead. It will probably be a few commits down the road for that though, as it will also require a change in the product.

andrewjstone:
I finally got back around to fixing up the code as requested in this review. @robem is going to take over this library from here this sprint.

Just so everyone is caught up, we are likely going to change the return type for status such that each callable returns a specific type wrapped in a variant. For example, rather than just returning a string, we could return a string or a metric, or a list of values. Then we can take these and output them in human readable format, json, etc... according to the returned types. We can't do that if we only return strings that are formatted in an ad-hoc manner. This suggestion comes from my prior experience building [clique](https://github.com/basho/clique#type-assembly-functions)

PR #506, opened 2020-04-07
Infrastructure for sending span context with Concord-BFT protocol

f-squirrel:
The PR is an initial part of the effort to support Open Tracing[1].

Open Tracing helps to pinpoint where failures occur and what causes poor performance in distributed systems.

The essential part of Open Tracing is span. A span is an object created for tracking the time spent executing a transaction. A span might be sent over the network in order to continue it in another process.

There are several ways to send a span over the network. In Concord-BFT, a span is sent as a binary array containing the span's context.

The PR introduces the infrastructure to send the span context as a part of Concord-BFT messages.

The new structure of the messages:
```
---------------------------------------------------------------------------------------------------------------------------------
|     Message base header      |   Message specific header   | Span context bytes | Payload(request bytes, cid, signatures, etc |
---------------------------------------------------------------------------------------------------------------------------------
|<Msg type>:<Span context size>| Any message specific fields |<Span context bytes>|           <Payload bytes>                   |
---------------------------------------------------------------------------------------------------------------------------------
```
The span context is stored before the payload because it is considered as a part of the internal/utility data of a packet.
Since the context may have a variable size, its length is stored as a part of the message base header.

Since most of the messages are supposed to support the span context, here is the list of the messages that do not support the context:
* `ClientReplyMsg` does not need to carry a span unless the BFT-Client is synchronous.
* `SimpleAckMsg`

Besides the addition of the context, there is a list of changes within the message's codebase:
* `ViewChangeMsg::addElement`: removed the unused parameter `replicasInfo`.
* All of the message headers were renamed to Header, with the exception of `ClientRequestMsg` and `ClientReplyMsg`, because they are defined in a separate header.
* Added a generic function, which returns the maximum size of messages `maxMessageSize`.
* Added a generic function, which returns the size of each message' header`sizeOfHeader`.
* Changed `ReqMissingDataMsg`, which introduces bit fields to easily handle message flags.

As part of the PR, unit tests were added to the messages.

[1] https://opentracing.io/docs/overview/what-is-tracing/


andrewjstone:
@f-squirrel This is a very large commit with a lot of nuance. The commit message is inadequate to cover what is going on. Can you please enhance the commit message to explain the context and rationale behind why this change was made the way it was, and how the messages change? For example, why is a span size now part of each message header? What is a span context? Why is its size variable, and how did we arrive at a max size? This context is important not only for a reviewer, but anyone else not versed in the rationale behind the commit and exactly what a span and context is.

Thanks!

util/pyclient/bft_msgs.py
```diff
@@ -65,11 +65,21 @@ def pack_request_header(header, pre_process=False):
...
+    end = start + header.span_context_size
+    span_context = data[start:end]
+
+    start = end
+    end = start + header.length
+    msg = data[start:end]
+
+    start = end
+    cid = data[start:].decode()
+    return header, span_context, msg, cid
```
andrewjstone, line 82: This is a nit, but I find it inconsistent that the `span_context` gets returned before `msg, but `cid` after message. To me, they are really doing a very similar job, and are metadata after the header and message. I feel like they should either both be before message, or both after. Additionally, this should be reflected in how the messages themselves get constructed. If CID comes after the message in a packed struct, then why shouldn't the span_context ?

util/pyclient/test_msgs.py
```diff
@@ -28,20 +28,29 @@ def test_unpack_reply(self):
...
         timeout_milli = 5000
+        span_context = b'span context'
+        msg = b'hello'
         cid = str(req_seq_num)
-        packed = bft_msgs.pack_request(client_id, req_seq_num, read_only, timeout_milli, cid, msg)
-        (header, unpacked_msg, cid_) = bft_msgs.unpack_request(packed, len(cid))
-        self.assertEqual(cid, cid_)
+
+        packed = bft_msgs.pack_request(client_id, req_seq_num, read_only,
+                                timeout_milli, cid, msg, False, span_context)
```
andrewjstone, line 40: Can we please avoid boolean parameters? If we must use them, let's make them keyword parameters. With a boolean parameter, it's impossible to tell at the call sight what it means.

bftengine/include/bftengine/ClientMsgs.hpp
```diff
@@ -21,6 +21,7 @@ namespace bftEngine {
 #pragma pack(push, 1)
 struct ClientRequestMsgHeader {
   uint16_t msgType;  // always == REQUEST_MSG_TYPE
+  uint32_t spanContextSize = 0u;
```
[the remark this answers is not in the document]
f-squirrel: The two first (msg type and spanContextSize) parameters of ClientRequestMsgHeader are just a copy of the header of MessageBase.
For an unknown reason, ClientRequestMsg exposes MessageBase by deriving from it but does not use it's types.

bftengine/src/bftengine/messages/CheckpointMsg.cpp
```diff
@@ -15,32 +15,28 @@
...
+CheckpointMsg::CheckpointMsg(
+    ReplicaId senderId, SeqNum seqNum, const Digest& stateDigest, bool stateIsStable, const std::string& spanContext)
+    : MessageBase(senderId, MsgCode::Checkpoint, spanContext.size(), sizeof(CheckpointMsgHeader)) {
   b()->seqNum = seqNum;
   b()->stateDigest = stateDigest;
   b()->flags = 0;
   if (stateIsStable) b()->flags |= 0x1;
-}
-
-CheckpointMsg* CheckpointMsg::clone() {
```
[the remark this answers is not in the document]
f-squirrel: It is not used.

bftengine/include/bftengine/ClientMsgs.hpp
```diff
@@ -21,6 +21,7 @@ namespace bftEngine {
 #pragma pack(push, 1)
 struct ClientRequestMsgHeader {
```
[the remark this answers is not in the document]
f-squirrel: No, it does not as long as client requests are synchronous.

bftengine/src/bftengine/messages/ReqMissingDataMsg.hpp
```diff
@@ -18,55 +18,67 @@ namespace impl {
...
+  void setSlowPathHasStarted() { b()->flags.bits.slowPathHasStarted = 1; }
 
+  std::string spanContext() const override {
+    return std::string(body() + sizeof(ReqMissingDataMsgHeader), spanContextSize());
+  }
   void validate(const ReplicasInfo&) const override;
 
  protected:
 #pragma pack(push, 1)
+  struct BitFields {
```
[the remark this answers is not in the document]
f-squirrel: The struct is defined under the 'pack' directive, the bit field magic is a part of C-standard.
The size of BitFields is guaranteed by a static assert:
`static_assert(sizeof(Flags) == sizeof(uint16_t));`

bftengine/src/bftengine/messages/ReqMissingDataMsg.hpp
```diff
@@ -18,55 +18,67 @@ namespace impl {
...
+    uint8_t partialProofIsMissing : 1;
+    uint8_t partialPrepareIsMissing : 1;
+    uint8_t partialCommitIsMissing : 1;
+    uint8_t fullCommitProofIsMissing : 1;
+    uint8_t fullPrepareIsMissing : 1;
+    uint8_t fullCommitIsMissing : 1;
+    uint8_t slowPathHasStarted : 1;
+  };
+
+  union Flags {
```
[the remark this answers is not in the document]
f-squirrel: The `uint16_t` parameter of the union is the one that defines the actual size of the struct.
Moreover, sometimes we need all flags at once.

util/pyclient/bft_msgs.py
f-squirrel: The parts of the message are returned in the same order they're stored(header, span, msg_body, cid). The CID is actually stored after the message body and it is specific for only for client request.
I believe that it is preferable to behave like every other binary protocol:
header/payload, in our case everything that is not message body is header(even if it contains fields with variable length).
Besides, `cid` is supposed to be before the message body but I am afraid this is beyond the scope of this commit.

bftengine/tests/messages/ReplicaStatusMsg_test.cpp
```diff
@@ -0,0 +1,215 @@
...
+  EXPECT_EQ(msg.getLastStableSeqNum(), lastStable);
+  EXPECT_EQ(msg.getLastExecutedSeqNum(), lastExecuted);
+  EXPECT_EQ(msg.currentViewIsActive(), viewIsActive);
+  EXPECT_EQ(msg.currentViewHasNewViewMessage(), hasNewChangeMsg);
+  EXPECT_EQ(msg.hasListOfPrePrepareMsgsInActiveWindow(), listOfPrePrepareMsgsInActiveWindow);
+  EXPECT_EQ(msg.hasListOfMissingViewChangeMsgForViewChange(), listOfMissingViewChangeMsgForViewChange);
+  EXPECT_EQ(msg.hasListOfMissingPrePrepareMsgForViewChange(), listOfMissingPrePrepareMsgForViewChange);
+
+  EXPECT_NO_THROW(msg.validate(replicaInfo));
+  /* the next methods crash the app with an assert ¯\_(ツ)_/¯
```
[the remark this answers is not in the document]
f-squirrel: The actual problem is that the messages are static i.e. once initialized they never change their size.
In this particular test, the message was declared with
```
bool listOfPrePrepareMsgsInActiveWindow = false;
bool listOfMissingViewChangeMsgForViewChange = false;
bool listOfMissingPrePrepareMsgForViewChange = false;
```
It means that there is the place for storing the 'lists' was not allocated and the tested methods cannot check the condition and as a result crash with assert.
I think those getters have to return false which I've already mentioned in our slack channel.
But again, the PR is already quite large...
The only thing I wanted to emphasize was that I did not call that in the unit test for a purpose.

bftengine/src/bftengine/PrimitiveTypes.hpp
```diff
@@ -35,6 +35,7 @@ typedef uint16_t NodeIdType;  // TODO(GG): change name
 
 typedef uint32_t MsgSize;
 typedef uint16_t MsgType;
+typedef uint32_t SpanContextSize;
```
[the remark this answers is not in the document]
f-squirrel: Would you like to remove it/remove it together with MsgSize/leave as is?

util/pyclient/test_msgs.py
f-squirrel: Actually, I haven't introduced this boolean, but nevertheless I've made it keyword

bftengine/src/bftengine/messages/MessageBase.hpp
```diff
@@ -24,14 +24,16 @@ class MessageBase {
 #pragma pack(push, 1)
   struct Header {
     MsgType msgType;
+    SpanContextSize spanContextSize = 0u;
   };
 #pragma pack(pop)
 
-  static_assert(sizeof(Header) == 2, "MessageBase::Header is 2B");
+  static_assert(sizeof(Header) == 6, "MessageBase::Header is 2B");
```
teoparvanov, line 31: 6B I guess...

f-squirrel: Done

teoparvanov:
PS: @andrewjstone @f-squirrel , my final suggestion is the following. Ideally, the "infrastructure" part (i.e. refactoring MessageBase to support span context) should have been done separately. Now I guess it's too late, as Dima has already implemented the whole thing and it's maybe not reasonable to make him start over and split it. For practical reasons, I would suggest to "bite the bullet" and review this PR in a detailed way, and then merge it with its current scope.

andrewjstone:
> PS: @andrewjstone @f-squirrel , my final suggestion is the following. Ideally, the "infrastructure" part (i.e. refactoring MessageBase to support span context) should have been done separately. Now I guess it's too late, as Dima has already implemented the whole thing and it's maybe not reasonable to make him start over and split it. For practical reasons, I would suggest to "bite the bullet" and review this PR in a detailed way, and then merge it with its current scope.

I already spent about 90 minutes reviewing this PR. I don't think I have any more in me. I'm not likely to find bugs, even with line by line analysis of the tests. That's the problem. The good news is that these messages are used everywhere and will likely be tested well by Apollo. The tests also appear to be comprehensive. I still stand by my request to refactor to a template method though.

teoparvanov:
> 
> 
> > PS: @andrewjstone @f-squirrel , my final suggestion is the following. Ideally, the "infrastructure" part (i.e. refactoring MessageBase to support span context) should have been done separately. Now I guess it's too late, as Dima has already implemented the whole thing and it's maybe not reasonable to make him start over and split it. For practical reasons, I would suggest to "bite the bullet" and review this PR in a detailed way, and then merge it with its current scope.
> 
> I already spent about 90 minutes reviewing this PR. I don't think I have any more in me. I'm not likely to find bugs, even with line by line analysis of the tests. That's the problem. The good news is that these messages are used everywhere and will likely be tested well by Apollo. The tests also appear to be comprehensive. I still stand by my request to refactor to a template method though.

I also think Apollo covers most of these flows, so I'm not too worried about regressions. It's simply hard to review from a human reviewer standpoint x)). +1 for the refactoring too.

bftengine/src/bftengine/messages/CheckpointMsg.hpp
```diff
@@ -46,7 +52,7 @@ class CheckpointMsg : public MessageBase {
     uint8_t flags;
   };
 #pragma pack(pop)
-  static_assert(sizeof(CheckpointMsgHeader) == (2 + 8 + DIGEST_SIZE + 1), "CheckpointMsgHeader is 43B");
+  static_assert(sizeof(CheckpointMsgHeader) == (6 + 8 + DIGEST_SIZE + 1), "CheckpointMsgHeader is 49B");
```
teoparvanov, line 55: nit - you add 4 bytes, so it should be 47B, right?

bftengine/src/bftengine/messages/ReplicaStatusMsg.hpp
```diff
@@ -73,13 +78,15 @@ class ReplicaStatusMsg : public MessageBase {
     uint8_t flags;
   };
 #pragma pack(pop)
-  static_assert(sizeof(ReplicaStatusMsgHeader) == (2 + 8 + 8 + 8 + 1), "ReplicaStatusMsgHeader is 27B");
+  static_assert(sizeof(ReplicaStatusMsgHeader) == (6 + 8 + 8 + 8 + 1), "ReplicaStatusMsgHeader is 30B");
```
teoparvanov, line 81: 31B ?

andrewjstone:
@teoparvanov @f-squirrel and I spoke offline. He's going to refactor to a template method. So it may be worth it to wait to do further review. Thanks again for looking. Always good to have another pair of eyes.

bftengine/include/bftengine/ClientMsgs.hpp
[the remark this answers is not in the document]
andrewjstone: Good point! I agree with your assesment.

util/pyclient/bft_msgs.py
andrewjstone: I agree with your assessment. Thanks!

util/pyclient/test_msgs.py
andrewjstone: Thanks!

bftengine/src/bftengine/PrimitiveTypes.hpp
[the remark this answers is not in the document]
andrewjstone: Oh, sorry. I guess you can just leave it as is for now.

bftengine/src/bftengine/messages/AskForCheckpointMsg.hpp
```diff
@@ -13,27 +13,37 @@
...
 #include "MessageBase.hpp"
 #include "Digest.hpp"
+#include "assertUtils.hpp"
 
 namespace bftEngine::impl {
 
 class AskForCheckpointMsg : public MessageBase {
  public:
-  static MsgSize maxSizeOfAskForCheckpointMsg() { return sizeof(AskForCheckpointMsgHeader); }
+  static MsgSize maxSizeOfAskForCheckpointMsg() { return sizeof(AskForCheckpointMsgHeader) + SPAN_CONTEXT_MAX_SIZE; }
```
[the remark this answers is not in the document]
f-squirrel: Done

bftengine/src/bftengine/messages/AskForCheckpointMsg.hpp
```diff
@@ -13,27 +13,37 @@
...
 
-  AskForCheckpointMsg(ReplicaId senderId)
-      : MessageBase(senderId, MsgCode::AskForCheckpoint, sizeof(AskForCheckpointMsgHeader)) {}
+  AskForCheckpointMsg(ReplicaId senderId, const std::string& spanContext = "")
+      : MessageBase(senderId, MsgCode::AskForCheckpoint, spanContext.size(), sizeof(AskForCheckpointMsgHeader)) {
+    char* position = body() + sizeof(AskForCheckpointMsgHeader);
+    memcpy(position, spanContext.data(), spanContext.size());
+  }
+
+  std::string spanContext() const override {
```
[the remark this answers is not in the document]
f-squirrel: Done

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -52,27 +61,32 @@ bool ClientRequestMsg::isReadOnly() const { return (msgBody()->flags & READ_ONLY
...
 }
 
 void ClientRequestMsg::setParams(NodeIdType sender,
                                  ReqId reqSeqNum,
                                  uint32_t requestLength,
                                  uint8_t flags,
+                                 uint64_t reqTimeoutMilli,
                                  const std::string& cid,
-                                 uint64_t reqTimeoutMilli) {
+                                 const std::string& spanContext) {
```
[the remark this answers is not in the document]
f-squirrel: Done.

bftengine/src/bftengine/messages/MessageBase.hpp
```diff
@@ -56,17 +58,19 @@ class MessageBase {
 
   MsgType type() const { return msgBody_->msgType; }
 
+  SpanContextSize spanContextSize() const { return msgBody_->spanContextSize; }
+  virtual std::string spanContext() const { return ""; };
```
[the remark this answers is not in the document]
f-squirrel: Replaced with a template method

bftengine/src/bftengine/messages/ReplicaStatusMsg.hpp
f-squirrel: Done

bftengine/src/bftengine/messages/CheckpointMsg.hpp
f-squirrel: Done

bftengine/src/bftengine/messages/AskForCheckpointMsg.hpp
```diff
@@ -13,27 +13,37 @@
...
+  }
 
   AskForCheckpointMsg* clone() { return new AskForCheckpointMsg(*this); }
 
   void validate(const ReplicasInfo& repInfo) const override {
     Assert(type() == MsgCode::AskForCheckpoint);
     Assert(senderId() != repInfo.myId());
 
-    if (size() > sizeof(AskForCheckpointMsgHeader)) throw std::runtime_error(__PRETTY_FUNCTION__);
+    if (size() > sizeof(AskForCheckpointMsgHeader) + spanContextSize()) {
```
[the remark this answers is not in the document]
f-squirrel: I support the idea to have a base class method to check the validity of the messages.
However, since the validity conditions vary from message to message, it requires a deeper analysis.
Since PR is already quite large, I'd prefer to have a separate PR for this change.
WDYT?

f-squirrel:
@andrewjstone and @teoparvanov , I wanted to thank you guys very much for the detailed review!
You provided very good comments and usable points.
I addressed them in my comments, in my updated summary of the PR and within the actual code :)

bftengine/src/bftengine/messages/AskForCheckpointMsg.hpp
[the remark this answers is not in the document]
andrewjstone: Makes sense to me. Fine to leave as is for now.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -77,11 +73,21 @@ void PrePrepareMsg::validate(const ReplicasInfo& repInfo) const {
...
+                             CommitPath firstPath,
+                             const std::string& spanContext,
+                             bool isNull,
+                             size_t size)
+    : MessageBase(sender,
+                  MsgCode::PrePrepare,
+                  spanContext.size(),
+                  (((size + sizeof(Header)) < maxMessageSize<PrePrepareMsg>())
+                       ? (size + sizeof(Header))
+                       : maxMessageSize<PrePrepareMsg>() - spanContext.size()))
```
teoparvanov, line 90: here you assume that `maxMessageSize()` includes the span context size, which may not be true in general...

teoparvanov, line 90: Just by looking at hte size calculation logic, I can't be 100% it is equivalent to the old logic:
```

          (((size + sizeof(PrePrepareMsgHeader)) < maxSizeOfPrePrepareMsg()) ? (size + sizeof(PrePrepareMsgHeader))
                                                                             : maxSizeOfPrePrepareMsg()))
```

bftengine/src/bftengine/messages/ReqMissingDataMsg.hpp
```diff
@@ -18,57 +18,69 @@ namespace impl {
...
-  void setFullCommitIsMissing() { b()->flags |= 0x80; }
-  void setSlowPathHasStarted() { b()->flags |= 0x100; }
+  void setPrePrepareIsMissing() { b()->flags.bits.prePrepareIsMissing = 1; }
+  void setPartialProofIsMissing() { b()->flags.bits.partialProofIsMissing = 1; }
+  void setPartialPrepareIsMissing() { b()->flags.bits.partialPrepareIsMissing = 1; }
+  void setPartialCommitIsMissing() { b()->flags.bits.partialCommitIsMissing = 1; }
+  void setFullCommitProofIsMissing() { b()->flags.bits.fullCommitProofIsMissing = 1; }
+  void setFullPrepareIsMissing() { b()->flags.bits.fullPrepareIsMissing = 1; }
+  void setFullCommitIsMissing() { b()->flags.bits.fullCommitIsMissing = 1; }
+  void setSlowPathHasStarted() { b()->flags.bits.slowPathHasStarted = 1; }
```
teoparvanov, line 47: this is cool indeed, but it's absolutely unrelated to the span context changes... Such improvements should be done in separate, more targetted PR-s

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
f-squirrel: The idea is the following: if the payload size(`size`) + header size is less than the maximum size + span size, then use it.
In this case, the context' size is added to the given size in the `MessageBase`.
Otherwise, use the  `maximum size - span size` because of ^^.
I agree the memory management model is unclear and hard to follow, but I only tried to incorporate the new functionality into it.

bftengine/src/bftengine/messages/ReqMissingDataMsg.hpp
f-squirrel: Agreed, just wanted to add a small improvement on the way...

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
f-squirrel: I think it is the same with one exception which the span size.
However, I might miss something. Probably it worth it to have another pair of eyes on it.
@andrewjstone , could you have a look?

f-squirrel:
It seems the pull request did not pass Travis tested after being merged.

PR #509, opened 2020-04-09
Unit test for two prepare certs with same seq

HristoStaykov:
Unit test to verify that when we have two prepare certificates for the same sequence number the one with the higher view is considered during the View Change process.
During this process ViewChangeSafetyLogic searches all View Change messages for Prepare Certificates in the working window of sequence numbers. For every sequence number it should use the Prepare Certificate with highest View number in order to guarantee safety.


bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -147,6 +197,105 @@ void setUpConfiguration_4() {
 
 void cleanupConfiguration_4() { delete sigManager_; }
 
+TEST(testViewchangeSafetyLogic_test, computeRestrictions_two_prepare_certs_for_same_seq_no) {
```
teoparvanov, line 200: @HristoStaykov can you please document this test method, explaining the overall approach, as well as details about the test steps? A bit like what we do for each Apollo test. Thanks!

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -52,28 +52,36 @@ class myViewsManager : public ViewsManager {
...
+    dynamic_cast<IncomingMsgsStorageImp*>(incomingMsgsStorage_)->stop();
+    delete incomingMsgsStorage_;
   }
-  ~DummyReplica() { delete incomingMsgsStorage_; }
 
   void onPrepareCombinedSigFailed(SeqNum seqNumber, ViewNum view, const set<uint16_t>& replicasWithBadSigs) {
-    combine_operation_promise.set_value(PrepareCombinedSigOperationStatus::OPERATION_FAILED);
+    std::unique_lock<std::mutex> lock(m_);
+    operationStatus_ = PrepareCombinedSigOperationStatus::OPERATION_FAILED;
+    cv_.notify_one();
```
teoparvanov, line 71: what's the purpose of `cv_` ?

HristoStaykov: It replaces the future/promise objects used in previous version. Here we have multiple calls to sinc, because when we add msg-s to the seqNumInfo object it checks if enough signature shares have arrived to combine them. This combination happens on the thread pool of the replica, and after it is completed the thread that has made the combination calls either onPrepareCombinedSigSucceeded(), or onPrepareCombinedSigFailed(), so basically after the main thread has added enough signature shares it has to wait for the combining thread to actually combine them and generate PrepareFullMsg.

teoparvanov: Then you should simply include this explanation in the PR's description, so that reviewers can find their way around the changes...

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -52,28 +52,36 @@ class myViewsManager : public ViewsManager {
 class DummyReplica : public InternalReplicaApi {
```
teoparvanov, line 52: As discussed offline, let's remove this complex infrastructure, and simply replace it with a `MockThresholdVerifier` whose `verify()` method always returns `true`.

bftengine/tests/testViewChange/testViewChange.cpp
HristoStaykov: Using mock for threshold signatures generation/verification makes the mock-ed Replica no longer necessary.

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -165,73 +138,182 @@ TEST(testViewchangeSafetyLogic_test, computeRestrictions) {
...
+// protocol with the right PrePrepare, matching the Prepare Certificate with
+// the highest View number.
+TEST(testViewchangeSafetyLogic_test, computeRestrictions_two_prepare_certs_for_same_seq_no) {
+  char buff[32]{};
+  bftEngine::impl::SeqNum lastStableSeqNum = 100;
+  const uint32_t kRequestLength = 2;
+
+  uint64_t expectedLastValue1 = 12345;
+  const uint64_t requestBuffer1[kRequestLength] = {(uint64_t)200, expectedLastValue1};
+  ViewNum curView1 = 0;
```
teoparvanov, line 206: It's not the current view per se, it's "a view" (possibly not the current). So I'd suggest to call this view1.

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -165,73 +138,182 @@ TEST(testViewchangeSafetyLogic_test, computeRestrictions) {
...
+  // View=0, this is going to be the Prepare Certificate that has to be
+  // discarded for the sequence number -> assignedSeqNum, because we are
+  // going to generate a Prepare Certificate for this sequence number with
+  // a higher view.
+  PrepareFullMsg* pfMsg1 =
+      PrepareFullMsg::create(curView1, assignedSeqNum, pRepInfo->primaryOfView(curView1), buff, sizeof(buff));
+
+  uint64_t expectedLastValue2 = 1234567;
+  const uint64_t requestBuffer2[kRequestLength] = {(uint64_t)200, expectedLastValue2};
+  ViewNum curView2 = 1;
```
teoparvanov, line 236: same here, let's call this simply `view2`

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -165,73 +138,182 @@ TEST(testViewchangeSafetyLogic_test, computeRestrictions) {
...
+                                pfMsg2->signatureLen(),
+                                pfMsg2->signatureBody());
+
+  auto VCS = ViewChangeSafetyLogic(N, F, C, &dummyVerifier_, PrePrepareMsg::digestOfNullPrePrepareMsg());
 
   SeqNum min{}, max{};
   VCS.computeRestrictions(viewChangeMsgs, 0, min, max, restrictions);
 
   for (int i = 0; i < kWorkWindowSize; i++) {
     if (i == assignedSeqNum - 1) {
```
teoparvanov, line 296: why `assignedSeqNum - 1` instead of `assignedSeqNum` ?

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -309,10 +391,9 @@ TEST(testViewchangeSafetyLogic_test, empty_correct_VC_msgs) {
 
 int main(int argc, char** argv) {
   ::testing::InitGoogleTest(&argc, argv);
-  assert(system("../../../tools/GenerateConcordKeys -n 4 -f 1 -o replica_keys_") != -1);
+  Assert(system("../../../tools/GenerateConcordKeys -n 4 -f 1 -o replica_keys_") != -1);
```
teoparvanov, line 394: do we still need this, given that we no longer create a dummy replica?

teoparvanov:
PS: can you also change the PR & commit title, to mention "view change"? It's not very clear right now... Thanks!

bftengine/tests/testViewChange/testViewChange.cpp
HristoStaykov: My mistake, in computeRestrictions() the second argument is the stable sequence number extracted from the View Change messages. This function returns "min" and "max" sequence numbers in the current working window, so basically "i == assignedSeqNum - min" should be the check.

bftengine/tests/testViewChange/testViewChange.cpp
HristoStaykov: We don't need it anymore, thanks :)

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -165,73 +164,184 @@ TEST(testViewchangeSafetyLogic_test, computeRestrictions) {
...
-  VCS.computeRestrictions(viewChangeMsgs, 0, min, max, restrictions);
+  VCS.computeRestrictions(viewChangeMsgs, VCS.calcLBStableForView(viewChangeMsgs), min, max, restrictions);
 
   for (int i = 0; i < kWorkWindowSize; i++) {
-    if (i == assignedSeqNum - 1) {
+    if (i == assignedSeqNum - min) {
       Assert(!restrictions[i].isNull);
-      Assert(pp->digestOfRequests().toString() == restrictions[i].digest.toString())
+      // Assert the prepare certificate with higher view number is selected for assignedSeqNum
+      Assert(ppMsg2->digestOfRequests().toString() == restrictions[i].digest.toString());
```
teoparvanov, line 327: Let's replace `i` with `assignedSeqNum - min`, to be more explicit.

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -165,73 +164,184 @@ TEST(testViewchangeSafetyLogic_test, computeRestrictions) {
...
+      Assert(ppMsg2->digestOfRequests().toString() != ppMsg1->digestOfRequests().toString());
```
teoparvanov, line 328: let's make this assert first

PR #511, opened 2020-04-09
Make the MDC handling of concordlogger thread safe

tdimitrov:
When USE_LOG4CPP is disabled, the persistency_test is failing. The reason is concurrent access to the unordered_map in class Logger, which holds the MDCs.

The fix is to add a mutex which protects the unordered_map in Logger. As the mutex type is not copyable I had to implement copy constructor and operator= for Logger.

### DIFF
--- a/logging/include/Logging.hpp
+++ b/logging/include/Logging.hpp
@@ -34,14 +34,13 @@ class Logger {
   std::string _name;
   std::array<std::string, 6> LEVELS_STRINGS = {"TRACE", "DEBUG", "INFO", "WARN", "ERROR", "FATAL"};
   std::unordered_map<std::string, std::string> mdc_;
-  mutable std::mutex mdc_mut_;
+  mutable std::mutex mdc_mutex_;
   // If you add new members, don't forget to handle them in copy constructor and operator=
 
  public:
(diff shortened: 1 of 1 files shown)


logging/include/Logging.hpp
```diff
@@ -31,30 +32,36 @@ constexpr LogLevel CURRENT_LEVEL = LogLevel::info;
 
 class Logger {
   std::string _name;
-  std::string LEVELS_STRINGS[6] = {"TRACE", "DEBUG", "INFO", "WARN", "ERROR", "FATAL"};
+  std::array<std::string, 6> LEVELS_STRINGS = {"TRACE", "DEBUG", "INFO", "WARN", "ERROR", "FATAL"};
   std::unordered_map<std::string, std::string> mdc_;
+  mutable std::mutex mdc_mut_;
```
andrewjstone, line 37: Nit: Using a full word like `mdc_mutex_` makes it more apparent that mut means mutex and not mutable. For most of our variables we try to use full words.

tdimitrov: Thanks @andrewjstone. I've pushed another commit with the mutex renamed.

PR #514, opened 2020-04-10
Revert "PrePrepareMsg: Allocate based on requests"

robem:
Reverts vmware/concord-bft#491


robem:
I'm testing a fix right now so maybe the revert isn't needed.

PR #516, opened 2020-04-11
PrePrepareMsg: Mandatory request size

robem:
This change refactors the PrePrepareMsg constructor. Instead of taking an
optional `isNull` and `size` parameter, this change gets rid of the former and
makes the latter mandatory. With the required size parameter in place, there is
no need for a separate create-null-prepreparemsg function anymore due to the
parameter should make it clear.

### DIFF
--- a/bftengine/src/bftengine/ViewsManager.cpp
+++ b/bftengine/src/bftengine/ViewsManager.cpp
@@ -674,7 +674,7 @@ bool ViewsManager::tryToEnterView(ViewNum v,
       if (restrictionsOfPendingView[idx].isNull) {
         Assert(prePrepareMsgsOfRestrictions[idx] == nullptr);
         // TODO(GG): do we want to start from the slow path in these cases?
-        PrePrepareMsg* pp = PrePrepareMsg::createNullPrePrepareMsg(myId, myLatestActiveView, i);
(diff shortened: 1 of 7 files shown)


bftengine/src/bftengine/ViewsManager.cpp
```diff
@@ -680,8 +680,7 @@ bool ViewsManager::tryToEnterView(ViewNum v,
       const int64_t idx = i - minRestrictionOfPendingView;
       if (restrictionsOfPendingView[idx].isNull) {
         Assert(prePrepareMsgsOfRestrictions[idx] == nullptr);
-        // TODO(GG): do we want to start from the slow path in these cases?
```
andrewjstone, line 683: Are you sure that we do indeed want to start in slow path? Should you at least leave the comment?

robem: I read the comment different. We were starting in the slow path the whole time and so this question only made sense to me if we were starting in the fast path instead. Well, I can see it the other way. I can put the comment back.

robem: In the force-push below, I rebased and added the comment again.

PR #519, opened 2020-04-14
Add unit tests for adaptive tuning of slow path timer

cloudnoize:
## Final part of testing ControllerWithSimpleHistory.

Required the  following changes in the tested class `ControllerWithSimpleHistory`:
1. Bug fix  - BC-2167 - Wrong calculation of `timeToStartSlowPath` timer. 
2. Moving constants from cpp file to class scope.
3. helper method and macro for better reuse, readability and **testability** (couldn't perform stateless tests without it) .

### Testing the following functionality
The timeToStartSlowPath timer, determines the duration limit of a request in the fast path, if a request  exceeds that duration, it will be downgraded to slow path (ReplicaImp::tryToStartSlowPaths).

**The timer is being adaptively tuned per evaluation period (currently 64 blocks), 'learning' the average time it takes to execute a request.**


bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
```diff
@@ -93,18 +89,14 @@ bool ControllerWithSimpleHistory::onNewSeqNumberExecution(SeqNum n) {
...
+  auto duration = std::chrono::duration_cast<std::chrono::microseconds>(now - s.prePrepareTime);
 
   const auto MAX_DURATION_MICRO = microseconds(2 * 1000 * 1000);
   const auto MIN_DURATION_MICRO = microseconds(100);
 
-  if (duration > MAX_DURATION_MICRO)
-    s.durationMicro_ = MAX_DURATION_MICRO;
-  else if (duration < MIN_DURATION_MICRO)
-    s.durationMicro_ = MIN_DURATION_MICRO;
+  s.durationMicro_ = normalizeToRange(MIN_DURATION_MICRO, MAX_DURATION_MICRO, duration);
```
cloudnoize, line 97: `s.durationMicro_ = normalizeToRange(MIN_DURATION_MICRO, MAX_DURATION_MICRO, duration);` As you can see prior this change, if a duration was within the desired range, MIN_DURATION_MICRO and MAX_DURATION_MICRO, it was ignored, probably a line was omitted by mistake.
The byproduct of this bug is that not only the value is ignored, the duration set to 0, pushing the average to the minimum.

bftengine/src/bftengine/ControllerWithSimpleHistory.hpp
```diff
@@ -66,6 +76,14 @@ class ControllerWithSimpleHistory : public ControllerBase {
   // Returns true in case a sequence number is within the range of the current window.
   bool insideActiveWindow(const SeqNum& n);
 
+  // Returns lower or upper bound, in case val is out of range, otherwise returns val.
+  template <typename T>
+  const T& normalizeToRange(const T& lower, const T& upper, const T& val) {
```
f-squirrel, line 81: The function is generic and technically can serve any types supporting operators `<` and `>`.
I think it might be good to make it free and move to `util` directory.

bftengine/src/bftengine/ControllerWithSimpleHistory.hpp
```diff
@@ -23,13 +23,23 @@
 namespace bftEngine {
 namespace impl {
 
+#define RELATIVE_LOWER_BOUND(rel, val) (uint32_t)((1 - rel) * val)
```
f-squirrel, line 26: Can we have a template method instead of macro?

bftengine/src/bftengine/ControllerWithSimpleHistory.hpp
cloudnoize: Correct, for simplicity I left it there, but I'll move it to as a util

bftengine/src/bftengine/ControllerWithSimpleHistory.hpp
cloudnoize: Sure

teoparvanov: +1

bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
teoparvanov: Can you point out in the description, which test case reproduces this bug?

bftengine/src/bftengine/ControllerWithSimpleHistory.hpp
cloudnoize: I don't know if it will be used.
I thought about whether it's worth exporting it, my original choice was to leave it within the class.
Though it makes sense also to define it as util.

cloudnoize: I'll leave it in the class and define it static

bftengine/src/bftengine/ControllerWithSimpleHistory.hpp
```diff
@@ -66,6 +73,26 @@ class ControllerWithSimpleHistory : public ControllerBase {
   // Returns true in case a sequence number is within the range of the current window.
   bool insideActiveWindow(const SeqNum& n);
 
+  // Returns lower or upper bound, in case val is out of range, otherwise returns val.
+  template <typename T>
+  static const T& normalizeToRange(const T& lower, const T& upper, const T& val) {
```
f-squirrel, line 78: Actually, I expected these 3 functions to be free(not a part of a particular class).
Leaving them as a part of the class makes its interface less clean.
Honestly, I believe that they are a good candidate for being an independent utility:)

teoparvanov: I think it only makes sense to extract them as an independent utility only if another component needs to use them. Which would those components be? I think a static method is also an acceptable model for representing "pure" functions (no side effects).

cloudnoize: I tend to leave it as is, since utils tend to become a mess of unused functions, and beside this class I don't see any usage.  I agree with @f-squirrel that it could go to a util lib as well.
Anyway its such a minor functionality that it's not worth the effort.

PR #520, opened 2020-04-16
Unit test with PrepareCert who's sequence number is below last stable.

HristoStaykov:
The purpose of the test is to verify that during a View Change the Prepare
Certificates with Sequence numbers lower than the last stable will be ignored.


bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -343,6 +343,138 @@ TEST(testViewchangeSafetyLogic_test, computeRestrictions_two_prepare_certs_for_s
   delete pfMsg2;
   delete ppMsg2;
 }
+// The purpose of this test is to verify that a Prepare Certificate
+// who's Sequence number is below the Last Stable Sequence number
```
teoparvanov, line 347: who's -> whose

bftengine/tests/testViewChange/testViewChange.cpp
```diff
@@ -343,6 +343,138 @@ TEST(testViewchangeSafetyLogic_test, computeRestrictions_two_prepare_certs_for_s
...
+  VCS.computeRestrictions(viewChangeMsgs, VCS.calcLBStableForView(viewChangeMsgs), min, max, restrictions);
+
+  for (int i = 0; i < kWorkWindowSize; i++) {
+    if (i == assignedSeqNum - min) {
+      // Check that the sequence number above the last stable is considered
+      Assert(!restrictions[assignedSeqNum - min].isNull);
+      Assert(ppMsg1->digestOfRequests().toString() == restrictions[assignedSeqNum - min].digest.toString());
+    } else {
+      // Check that all others are NULL, and thus the Prepare Certificate
+      // for the sequence number below the last stable is ignored.
```
teoparvanov, line 461: can you add an explicit check for this, outside of the "else" statement?

HristoStaykov: Yes, we could also check that after calculating 
`VCS.computeRestrictions(viewChangeMsgs, VCS.calcLBStableForView(viewChangeMsgs), min, max, restrictions);`
the value of "min" is above "assignedSeqNumIgnored".

PR #521, opened 2020-04-16
Unit test for 'seqNumGeneratorForClientRequestsImp'

oded282:
In order to test 'seqNumGeneratorForClientRequestsImp' I have made few changes:

1. For 'seqNumGeneratorForClientRequestsImp' to be testsable, I added another, name identical, function. The initial one gets one time_point parameter and the new one won't get params.
This way we can preserve the initial behaviour, but can send specific time_point and be more flexible in the unit tests.

2. Created mock class of the 'seqNumGeneratorForClientRequestsImp' with basic functionalities 
needed, in order to test different behaviours.

3 Included dir in the cmake file, for using 'SimpleClientImp.cpp'.

4. Wrote unit-tests.

### DIFF
--- a/bftengine/include/bftengine/SimpleClient.hpp
+++ b/bftengine/include/bftengine/SimpleClient.hpp
@@ -16,6 +16,7 @@
 #include <string>
 #include <set>
 #include "ICommunication.hpp"
+#include <chrono>
 
 namespace bftEngine {
 // Parameters // TODO(GG): move to client configuration
@@ -73,6 +74,8 @@ class SeqNumberGeneratorForClientRequests {
   static SeqNumberGeneratorForClientRequests* createSeqNumberGeneratorForClientRequests();
 
   virtual uint64_t generateUniqueSequenceNumberForRequest() = 0;
+
+  virtual uint64_t generateUniqueSequenceNumberForRequest(std::chrono::time_point<std::chrono::system_clock> now) = 0;
 
   virtual ~SeqNumberGeneratorForClientRequests() = default;
 };
--- a/bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
+++ b/bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
@@ -1,8 +1,96 @@
 #include "gtest/gtest.h"
+#include "SimpleClientImp.cpp"
+#include <list>
 
-TEST(seqNumForClientRequest_test, initial_test) {}
+using namespace std;
+using namespace bftEngine;
 
-int main(int argc, char* argv[]) {
+namespace {
+
+class MockSeqNumGenerator : public SeqNumberGeneratorForClientRequestsImp {
+ public:
+  uint32_t getLastCountOfUniqueFetchID() { return lastCountOfUniqueFetchID_; }
+
+  void setLastCountOfUniqueFetchID(uint64_t count) { lastCountOfUniqueFetchID_ = count; }
+};
+
+TEST(seqNumForClientRequest_test, monotonicity_check) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new MockSeqNumGenerator();
+  std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+  list<u_int64_t> seqNums;
+
+  for (int i = 0; i < 10; i++) {
+    seqNums.push_back(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+    now.operator+=(chrono::milliseconds(1000));
+    usleep(1000);
+  }
+
+  // Checks that generated SN is monotonically increasing.
+  Assert(is_sorted(seqNums.begin(), seqNums.end()));
+  now.operator+=(std::chrono::hours(24));
+  auto newSeqNum = seqNumGen->generateUniqueSequenceNumberForRequest(now);
+  auto lastSeqNum = seqNums.back();
+
+  // Two SN generated on same time in different days, stays monotonic.
+  ASSERT_TRUE(newSeqNum > lastSeqNum);
+  delete seqNumGen;
+}
+
+TEST(seqNumForClientRequest_test, send_requests_same_time_uniqueness_preserved) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new MockSeqNumGenerator();
+  std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+  set<u_int64_t> seqNums;
+
+  for (int i = 0; i < 10; i++) {
+    seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+  }
+
+  // assert 10 unique SN was generated, although practically sent at the same time.
+  Assert(seqNums.size() == 10);
+  delete seqNumGen;
+}
+
+TEST(seqNumForClientRequest_test, exceed_counter_limit_uniqueness_preserved) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new MockSeqNumGenerator();
+  std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+  set<u_int64_t> seqNums;
+
+  seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+  auto mockSeqNumGen = dynamic_cast<MockSeqNumGenerator *>(seqNumGen);
+
+  // Set lastCount biggest value.
+  mockSeqNumGen->setLastCountOfUniqueFetchID(0x3FFFFF);
+  seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+
+  // assert unique SN was generated, although counter limit exceeded.
+  ASSERT_EQ(seqNums.size(), 2);
+  delete seqNumGen;
+}
+
+TEST(seqNumForClientRequest_test, invoke_counter_increasement) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new MockSeqNumGenerator();
+  set<u_int64_t> seqNums;
+
+  // generate bunch of sequence numbers
+  for (int i = 0; i < 10; i++) {
+    seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest());
+    usleep(1000);
+  }
+  // assert 10 unique sequence numbers was generated.
+  ASSERT_EQ(seqNums.size(), 10);
+
+  std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+  now.operator-=(chrono::milliseconds(1000));
+  seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+  auto mockSeqNumGen = dynamic_cast<MockSeqNumGenerator *>(seqNumGen);
+
+  // assert 'lastCountOfUniqueFetchID_' variable increased.
+  ASSERT_EQ(mockSeqNumGen->getLastCountOfUniqueFetchID(), 1);
+  delete seqNumGen;
+}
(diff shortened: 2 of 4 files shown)


bftengine/include/bftengine/SimpleClient.hpp
```diff
@@ -72,7 +73,8 @@ class SeqNumberGeneratorForClientRequests {
  public:
   static SeqNumberGeneratorForClientRequests* createSeqNumberGeneratorForClientRequests();
 
-  virtual uint64_t generateUniqueSequenceNumberForRequest() = 0;
+  virtual uint64_t generateUniqueSequenceNumberForRequest(
+      std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now()) = 0;
```
teoparvanov, line 77: this is not very elegant... Let's preserve the original method and instead use method overloading:
```
virtual uint64_t generateUniqueSequenceNumberForRequest() = 0;
virtual uint64_t generateUniqueSequenceNumberForRequest(std::chrono::time_point<std::chrono::system_clock> now) = 0;
```

Then in `generateUniqueSequenceNumberForRequest()`, you get `std::chrono::system_clock::now()` and pass it to the overloaded `generateUniqueSequenceNumberForRequest(now)`

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+#include "SimpleClientImp.cpp"
 
-TEST(seqNumForClientRequest_test, initial_test) {}
+using namespace std;
+using namespace bftEngine;
 
-int main(int argc, char* argv[]) {
+namespace {
+
+class mockSeqNumGenerator : public SeqNumberGeneratorForClientRequestsImp {
```
teoparvanov, line 9: `mockSeqNumGenerator` -> `MockSeqNumGenerator`

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+ public:
+  uint32_t getLastCountOfUniqueFetchID() { return lastCountOfUniqueFetchID_; }
+
+  void setLastCountOfUniqueFetchID(uint64_t count) { lastCountOfUniqueFetchID_ = count; }
+};
+
+TEST(seqNumForClientRequest_test, send_request_fast_still_persistence) {
```
teoparvanov, line 16: what does "..fast_still_presistence" mean?

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+  SeqNumberGeneratorForClientRequests *seqNumGen = new mockSeqNumGenerator();
```
teoparvanov, line 17: Why not `auto* seqNumGen = new mockSeqNumGenerator();`? This way you would probably not need to do dynamic casting several lines below...

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+    seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest());
+  }
+
+  Assert(seqNums.size() == 10);
+
+  delete seqNumGen;
+}
+
+TEST(seqNumForClientRequest_test, reach_counter_max_value) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new mockSeqNumGenerator();
```
teoparvanov, line 31: same here - `auto* seqNumGen = new mockSeqNumGenerator();` ?

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+  auto mockSeqNumGen = dynamic_cast<mockSeqNumGenerator *>(seqNumGen);
+
+  // Set lastCount biggest value.
+  mockSeqNumGen->setLastCountOfUniqueFetchID(0x3FFFFF);
+
+  seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+
+  ASSERT_EQ(seqNums.size(), 2);
+
+  delete seqNumGen;
```
teoparvanov, line 48: we don't need an empty line everywhere...

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+}
+
+TEST(seqNumForClientRequest_test, invoke_counter_increasment) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new mockSeqNumGenerator();
```
teoparvanov, line 52: `auto*` ?

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+  seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+
+  auto mockSeqNumGen = dynamic_cast<mockSeqNumGenerator *>(seqNumGen);
+
+  // Set lastCount biggest value.
+  mockSeqNumGen->setLastCountOfUniqueFetchID(0x3FFFFF);
+
+  seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+
+  ASSERT_EQ(seqNums.size(), 2);
```
teoparvanov, line 46: shouldn't we also check the content of seqNums? For example, that they are always increasing, or other, more advanced properties?

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,80 @@
...
+TEST(seqNumForClientRequest_test, send_request_fast_still_persistence) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new mockSeqNumGenerator();
+
+  set<u_int64_t> seqNums;
+
+  for (int i = 0; i < 10; i++) {
+    seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest());
+  }
+
+  Assert(seqNums.size() == 10);
```
teoparvanov, line 25: What does this assert tell you? Can you maybe add comments on each assert, in order to explain what property/invariant you are testing?

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
oded282: The content here is a number generated by a unique time stamp, I mostly tested the uniqueness this class trying to achieve.

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
oded282: Sure.

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
oded282: The generated key is derieved from system clock, I have tested that the class generate unique SN, even if requests sent with no break between each other. 
I will think of some more meaningful name.

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
oded282: The function that I am testing declared at 'SeqNumberGeneratorForClientRequestsImp' as private, so I can't use it that way.

bftengine/include/bftengine/SimpleClient.hpp
oded282: In order to test the class behaviours, I need flexibility in
'std::chrono::time_point<std::chrono::system_clock> now' variable, move it back and forward to achieve behaviours that I won't be able to achieve, because this class depends on time.

andrewjstone:
I haven't reviewed the code, but I saw the comments fly by and noticed that we are using system clock. We should be using the steady_clock from chrono which is guaranteed to be monotonic and not the system_clock. If this is not the case, we have a bug in the client.

oded282:
Hello @andrewjstone ,

Maybe I did not understood what the problem using system_clock,
But the function who use system_clock convert it by calling the time_since_epoch function.

Please, tell me if I didn't understand the problem you referred to.

bftengine/include/bftengine/SimpleClient.hpp
teoparvanov: what I'm suggesting achieves the same result as what you had proposed - simply by using method overloading

oded282: ok, I understand what you meant.

andrewjstone:
> Hello @andrewjstone ,
> 
> Maybe I did not understood what the problem using system_clock,
> But the function who use system_clock convert it by calling the time_since_epoch function.
> 
> Please, tell me if I didn't understand the problem you referred to.

The problem with using the system clock is that it is not monotonically increasing. It can actually go backwards in time, which can result in reusing timestamps. I'm not sure if this is actually a problem here, since I believe we save the last timestamp anyway to reuse. But in general, we should always use the steady clock. There is no practical use for us to ever use the system_clock unless we need to display user visible time, which we don't do anywhere.

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
teoparvanov: Aren't the generated sequence numbers also supposed to be monotonically increasing? If so, let's also test this.

bftengine/tests/testSeqNumForClientRequest/seqNumForClientRequest_test.cpp
```diff
@@ -1,8 +1,73 @@
...
+TEST(seqNumForClientRequest_test, exceed_counter_limit_uniqueness_preserved) {
+  SeqNumberGeneratorForClientRequests *seqNumGen = new MockSeqNumGenerator();
+  std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+  set<u_int64_t> seqNums;
+
+  seqNums.insert(seqNumGen->generateUniqueSequenceNumberForRequest(now));
+  auto mockSeqNumGen = dynamic_cast<MockSeqNumGenerator *>(seqNumGen);
+
+  // Set lastCount biggest value.
+  mockSeqNumGen->setLastCountOfUniqueFetchID(0x3FFFFF);
```
teoparvanov, line 39: Where does this 0x3FFFFF number come from?

oded282: The class generate SN out of time_point.
In order to preserve the uniqueness, we use to variables:
last_milli -> the sampled time_point.
last_count -> used to preserve uniqueness.
Before we return SN we shift to the right (0x3FFFFF), the last_milli and use bitwise 'or' with last count.
If we will try to generate another SN and the last_milli will be identical (time sampled is equal to the last one sampled), we will increase last_count by 1, and do the bitwise or with different number, therefore we will get unique SN.
The last_count is limited to increase up to 0x3FFFFF, the size we shifted the last_milli variable.

The class handle the scenario of exceeding the last_count limit, as tested in 'exceed_counter_limit_uniqueness_preserved'

PR #530, opened 2020-04-20
add phase-1 metrics (partially)

yontyon:
Add new metrics to implement phase1 matrics plan.

We have added metrics that count the number of sent status messages and reqMissingData messages as well as counters for the messages that are being sent as a response for status and ReqMissingData.
see BC-2318 on jira

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -110,6 +110,11 @@ void ReplicaImp::send(MessageBase *m, NodeIdType dest) {
   ReplicaBase::send(m, dest);
 }
 
+void ReplicaImp::sendAndIncrementMetric(MessageBase *m, NodeIdType id, CounterHandle &counterMetric) {
+  send(m, id);
+  counterMetric.Get().Inc();
+}
+
 void ReplicaImp::onReportAboutInvalidMessage(MessageBase *msg, const char *reason) {
   LOG_WARN(GL,
            "Node " << config_.replicaId << " received invalid message from Node " << msg->senderId()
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1509,7 +1513,10 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
 
       for (SeqNum i = beginRange; i <= endRange; i = i + checkpointWindowSize) {
         CheckpointMsg *checkMsg = checkpointsLog->get(i).selfCheckpointMsg();
-        if (checkMsg != nullptr) send(checkMsg, msgSenderId);
+        if (checkMsg != nullptr) {
+          send(checkMsg, msgSenderId);
+          metric_sent_checkpoint_msg_due_to_status_.Get().Inc();
```
yuliasherman, line 1518: You use the same pair of operators again and again: send() and increase of some metric.
I propose wrapping send() function for this purpose and pass the corresponding metric (CounterHandle) to it.
This will safe a lot of additional code lines.

yuliasherman:
Do we have an issue number for this effort?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -264,7 +264,10 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
     if (minBatchSize > maxReasonableMinBatchSize) minBatchSize = maxReasonableMinBatchSize;
```
yuliasherman, line 264: Do we have an issue number for this effort? If positive, please mention it in the description, otherwise please create it first.

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: I would prefer to leave it as it is.
I want to see explicitly where the metrics are taken and taking metrics is not part of "send" functionality...
wrapping it in send may be confusing

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: done

PR #532, opened 2020-04-22
[WIP] BCStateTran: Don't delete item cache when switching sources

robem:
This change removes deleting the item data message cache when we switch replica
sources. During manual testing we found that the late spawning replica will be
stuck in an endless loop of retrieving chunks and switching replica sources.
To reproduce the issue, this change adds a test (not run by default) with three
different configurations. In short, the test starts 3/4 replicas, waits for the
first checkpoint, starts the 4th replica and waits for state transfer to
complete. Without this change, only the first (8KB) configuration succeeds.
With this change, also the 16KB configuration succeeds. However, the last (1MB)
configuration still fails due to invalid messages and the frequent timeout. Due
to the latter, we run out of preferred replicas and delete the cache
eventually. We still need to investigate why transferring chunks takes so long
(timeout 5 seconds) and hence triggers a source switch.

### DIFF
--- a/tests/apollo/test_skvbc_st_reply_size.py
+++ b/tests/apollo/test_skvbc_st_reply_size.py
@@ -1,0 +1,107 @@
+# Concord
+#
+# Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+#
+# This product is licensed to you under the Apache 2.0 license (the "License").
+# You may not use this product except in compliance with the Apache 2.0 License.
+#
+# This product may include a number of subcomponents with separate copyright
+# notices and license terms. Your use of these subcomponents is subject to the
+# terms and conditions of the subcomponent's license, as noted in the LICENSE
+# file.
+import os.path
+import unittest
+from os import environ
+
+import trio
+
+from util import skvbc as kvbc
+from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
+
+
+def build_cmd(reply_size=None, page_size=None, num_pages=None):
+    def start_replica_cmd(builddir, replica_id):
+        """
+        Return a command that starts an skvbc replica when passed to
+        subprocess.Popen.
+        """
+        statusTimerMilli = "500"
+        path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
+        cmd = [
+          path,
+          "-k", KEY_FILE_PREFIX,
+          "-i", str(replica_id),
+          "-s", statusTimerMilli,
+        ]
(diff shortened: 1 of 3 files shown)


tests/apollo/test_skvbc_st_reply_size.py
```diff
@@ -0,0 +1,107 @@
...
+          cmd.extend(["--size-of-reserved-page", str(page_size)])
+
+        return cmd
+    return start_replica_cmd
+
+class SkvbcStReplySize(unittest.TestCase):
+
+    __test__ = False  # so that PyTest ignores this test scenario
+
+    async def state_transfer_after_first_checkpoint(self, bft_network):
```
teoparvanov, line 53: Let's make this a private method prefixed with `_`, and move it after the actual test methods that invoke it.

tests/apollo/test_skvbc_st_reply_size.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Note, we only run 4 skvbc test replicas. Each replica is configured to
+    # accomodate up to 100 client proxies and hence the formula for maximum
+    # number of reserved pages becomes: 100 * 4 * reply_size / page_size
+
+    @with_trio
+    @with_bft_network(start_replica_cmd=build_cmd(reply_size=1<<13,
+                                                  page_size=1<<12,
+                                                  num_pages=800),
+                      selected_configs=lambda n,f,c: n==4 and f==1 and c==0)
```
teoparvanov, line 86: You could simply say: `selected_configs=lambda n,f,c: n==4`, or even `selected_configs=lambda n,_,_: n==4`

tests/apollo/test_skvbc_st_reply_size.py
```diff
@@ -0,0 +1,107 @@
...
+          cmd.extend(["--max-reply-message-size", str(reply_size)])
+        if page_size is not None:
+          cmd.extend(["--max-num-of-reserved-pages", str(num_pages)])
+        if page_size is not None:
+          cmd.extend(["--size-of-reserved-page", str(page_size)])
+
+        return cmd
+    return start_replica_cmd
+
+class SkvbcStReplySize(unittest.TestCase):
```
teoparvanov, line 49: why not wire this test suite into CI? If we don't do so, we risk the test starting to "decay", and become irrelevant over time...

toly-kournik:
@robem please close this PR without merging

robem:
Thanks for the reminder. There is a scenario where the replica that is receiving the chunks won't make progress. If we retrieve at least one chunk of bad data from a malicious replica per block then the block verification will fail and we are effectively in a live-lock. If the cache is cleared between switching sources then we will talk to a non-malicious replica eventually.

PR #533, opened 2020-04-22
Add metrics for phase1 - client metrics

yontyon:
In this PR we have added a metrics framework to the bft client, including two metrics:
1. number of retransmission per client
2. the retransmission timer

Note that different clients can't use the same aggregator.

### DIFF
--- a/kvbc/include/ClientImp.h
+++ b/kvbc/include/ClientImp.h
@@ -35,6 +35,8 @@ class ClientImp : public IClient {
                                     uint32_t* outActualReplySize,
                                     const std::string& = "") override;
 
+  virtual void setMetricsAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) override;
+
  protected:
   ClientImp() = default;
   ~ClientImp() override = default;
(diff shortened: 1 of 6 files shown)


kvbc/include/ClientImp.h
```diff
@@ -35,6 +35,8 @@ class ClientImp : public IClient {
                                     uint32_t* outActualReplySize,
                                     const std::string& = "") override;
 
+  virtual void setMetricsAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) override;
```
toly-kournik, line 38: minor: don't use 'virtual' when overriding

tests/simpleTest/simple_test_client.hpp
```diff
@@ -74,6 +74,8 @@ class SimpleTestClient {
     ICommunication* comm = bftEngine::CommFactory::create(conf);
 
     SimpleClient* client = SimpleClient::createSimpleClient(comm, id, cp.numOfFaulty, cp.numOfSlow);
+    auto aggregator = std::make_shared<concordMetrics::Aggregator>();
```
toly-kournik, line 77: minor: singleliner?

kvbc/include/ClientImp.h
yontyon: All other methods in this class are defined that way.

tests/simpleTest/simple_test_client.hpp
yontyon: I'm using the aggregator at the end of the test.

PR #534, opened 2020-04-22
Add metrics functionality

yontyon:
We have added to the concord metrics framework the opportunity to get numeric metrics in a generic struct named Metrics (in the same way as with Prometheus metrics).

The main purpose of it is to get rid of the metrics configuration file in the product.

### DIFF
--- a/util/include/Metrics.hpp
+++ b/util/include/Metrics.hpp
@@ -21,7 +21,6 @@
 #include <memory>
 #include <list>
 #include <variant>
-#include "histogram.hpp"
 
 namespace concordMetrics {
 
@@ -110,7 +109,7 @@ class Counter {
 struct metric_ {
   std::string component;
   std::string name;
-  std::variant<Counter, Gauge, Status, concordUtils::Histogram> value;
+  std::variant<Counter, Gauge, Status> value;
 };
 
 class Values {
--- a/util/include/Metrics.hpp
+++ b/util/include/Metrics.hpp
@@ -20,6 +20,7 @@
 #include <mutex>
 #include <memory>
 #include <list>
+#include <variant>
 
 namespace concordMetrics {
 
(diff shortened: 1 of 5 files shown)


util/include/Metrics.hpp
```diff
@@ -29,6 +30,14 @@ class Gauge;
 class Status;
 class Counter;
 
+// A generic struct that may represent a counter or a gauge
+// the motivation is to eliminate that need to know the exact
+// metric name before getting it from the aggregator
+typedef struct {
+  std::string component;
+  std::string name;
+  uint64_t value;
```
andrewjstone, line 39: I talked to @yontyon offline and he's going to change the type of 'value' to return `std::variant<Status, Counter, Gauge, Histogram>` so we don't lose context and can return non-numeric types.

yontyon: done

util/include/Metrics.hpp
```diff
@@ -96,6 +104,15 @@ class Counter {
   uint64_t val_;
 };
 
+// A generic struct that may represent a counter or a gauge
+// the motivation is to eliminate that need to know the exact
+// metric name before getting it from the aggregator
+struct metric_ {
+  std::string component;
+  std::string name;
+  std::variant<Counter, Gauge, Status, concordUtils::Histogram> value;
```
andrewjstone, line 113: Oh sorry, I misspoke in my comment and you took me literally :) I thought we might have added a Histogram type to metrics.  Since we haven't lets just remove concordUtils::Historgram from the variant.

Otherwise, this looks great. Thank you for the change.

yontyon:
We have added to the concord metrics framework the opportunity to get numeric metrics in a generic struct named Metrics (in the same way as with Prometheus metrics).

The main purpose of it is to get rid of the metrics configuration file in the product.

PR #535, opened 2020-04-22
Test successful pre-execution for a single request.

yuliasherman:
Introduce an Apollo test verifying a single pre-execution request correct handling.

BC-1708 Test successful pre-execution of a single request (black box - no metrics)

### DIFF
--- a/tests/apollo/test_skvbc_preexecution.py
+++ b/tests/apollo/test_skvbc_preexecution.py
@@ -1,0 +1,61 @@
+# Concord
+#
+# Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+#
+# This product is licensed to you under the Apache 2.0 license (the "License").
+# You may not use this product except in compliance with the Apache 2.0 License.
+#
+# This product may include a number of subcomponents with separate copyright
+# notices and license terms. Your use of these subcomponents is subject to the
+# terms and conditions of the subcomponent's license, as noted in the LICENSE
+# file.
+
+import os.path
+import unittest
+
+from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
+from util import skvbc as kvbc
+
+
+def start_replica_cmd(builddir, replica_id):
+    """
+    Return a command that starts an skvbc replica when passed to
+    subprocess.Popen.
+
+    The replica is started with a short view change timeout and with RocksDB
+    persistence enabled (-p).
+
+    Note each arguments is an element in a list.
+    """
+
+    status_timer_milli = "500"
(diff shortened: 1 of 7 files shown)


tests/apollo/test_skvbc_preexecution.py
```diff
@@ -0,0 +1,65 @@
...
+    async def test_single_pre_process_request(self, bft_network):
+        """
+        Ensure that we can create a block using pre-execution feature and retrieve its KV pairs through
+        GetBlockData API request.
+        """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        client = bft_network.random_client()
+
+        # Perform read first to avoid sending write message to all replicas, which causes messages to be ignored
```
teoparvanov, line 57: I guess this happens because initially the client doesn't know which is the primary. It only learns it after the first request. Does this mean that if the first request is a `write` with `pre_process=True`, it will always fail?
If so, I think we should add a separate test, to highlight this behaviour (TBD if it's a bug).

yuliasherman: You are right: in case the primary replica is unknown, the request is sent to all replicas. Sending a read request as a first operation saves a bit of test time, but it works. I think we can get rid of this read request.

teoparvanov: Thanks Yulia, it looks cleaner now.

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -0,0 +1,61 @@
...
+            "-s", status_timer_milli,
+            "-v", view_change_timeout_milli,
+            "-p"
+            ]
+
+
+class SkvbcPreExecutionTest(unittest.TestCase):
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs = lambda n, f, c: n == 7)
```
teoparvanov, line 47: I'd also suggest to run for all BFT configurations, instead of just for `n == 7`

yuliasherman: Do you think all pre-execution tests should run with all configurations?

teoparvanov: I think yes, in general we should run tests for all configurations. If you look into the other Apollo tests, you'll notice that specifying the `selected_configs` is rather an exception. We might want to specify a given config when a test takes a long time to execute, or when reproducing a bug, which is not the case here.

teoparvanov: LGTM, thanks Yulia!

PR #537, opened 2020-04-22
WIP Fix memory leaks

HristoStaykov:
The purpose of this branch is to clean memory leaks found with Address and Leak Clang sanitizers.
To build with sanitizers enabled cmake should be runned with -DLEAKCHECK=true
Currently I have found memory leaks thanks to the following tests:
- serializable_test
- TestThresholdSerialization
- test_serialization


CMakeLists.txt
```diff
@@ -9,7 +9,11 @@ project(concord-bft VERSION 0.1.0.0 LANGUAGES CXX)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
+if(DEFINED IS_LEAKCHECK)
```
andrewjstone, line 12: Nit: Can we change the name of this option to be more declarative? Maybe just `LEAKCHECK`.

We probably should also default it to `TRUE`.

CMakeLists.txt
```diff
@@ -64,7 +68,8 @@ string(APPEND CMAKE_CXX_FLAGS " -Wbuiltin-macro-redefined")
 string(APPEND CMAKE_CXX_FLAGS " -pedantic")
 string(APPEND CMAKE_CXX_FLAGS " -Werror")
 if(IS_LEAKCHECK)
-    string(APPEND CMAKE_CXX_FLAGS " -fsanitize=leak")
+    string(APPEND CMAKE_CXX_FLAGS " -fsanitize=leak -fsanitize=address -fno-omit-frame-pointer")
```
andrewjstone, line 71: OOH. Thanks for this change :) Can we always run with `-fno-omit-frame-pointer`, rather than only doing it when using `IS_LEAKCHECK`?

 Omitting the frame pointer is the silliest build default ever on new hardware. The optimizer doesn't need an extra general purpose register to work with, but debugging and performance tools certainly work much better with a frame pointer.

bftengine/src/bftengine/Bitmap.hpp
```diff
@@ -41,6 +41,18 @@ class Bitmap {
     }
   }
 
+  Bitmap(Bitmap&& other) : numBits_(other.numBits_), p_(nullptr) {
+    if (numBits_ == 0) {
```
andrewjstone, line 45: This if statement is a little strange to me. Either we assume `numBits_ == 0 && `p_ == nullptr` or otherwise this move constructor can trigger leaks if `p != nullptr`. If it must always be nullptr, as I think it should in this case, we can just skip the if altogether and assign directly. We can possibly throw an assert in for good measure.

bftengine/src/bftengine/Bitmap.hpp
```diff
@@ -69,6 +81,25 @@ class Bitmap {
...
   }
 
+  Bitmap& operator=(Bitmap&& other) {
+    if (numBits_ > 0) {
+      Assert(p_ != nullptr);
+      std::free(p_);
+    }
+
+    numBits_ = other.numBits_;
+    if (numBits_ == 0) {
```
andrewjstone, line 91: Same as in the move constructor.

bftengine/src/bftengine/PersistentStorageDescriptors.cpp
```diff
@@ -273,7 +273,9 @@ void DescriptorOfLastExecution::deserialize(char *buf, size_t bufLen, uint32_t &
   buf += seqNumSize;
 
   uint32_t bitMapSize = 0;
-  validRequests = *Bitmap::createBitmapFromBuffer(buf, bufLen - seqNumSize, &bitMapSize);
+  auto* temp = Bitmap::createBitmapFromBuffer(buf, bufLen - seqNumSize, &bitMapSize);
```
andrewjstone, line 276: This is good. I don't understand why we even need to allocate Bitmaps on the heap to begin with.

threshsign/src/bls/relic/BlsThresholdSigner.cpp
```diff
@@ -98,6 +98,7 @@ void BlsThresholdSigner::deserializeDataMembers(istream &inStream) {
   deserialize(inStream, id_);
   BNT idNum(id_);
   idNum.toBytes(serializedId_, sizeof(id_));
+  delete params;
```
andrewjstone, line 101: This fix works, but it doesn't make the code exception safe. We probably want to look at using unique_ptrs everywhere instead of this in future commits. 

Same is true everywhere else we use delete outside a destructor also. We'll need to clean up the memory management. Fine to leave as is for now.

robem: I second this. Smart pointers is the only way forward.

CMakeLists.txt
HristoStaykov: Yes, will change the name. To default it to TRUE, though will require to fix all the tests, some are currently crashing with sanitizers enabled. The three I mentioned that helped me find leaks I have fixed to Pass.

CMakeLists.txt
HristoStaykov: Yes, I think so too, the optimization when removing the frame pointer is minor. Making it mandatory is very useful for debugging and performance measuring tools.

threshsign/src/bls/relic/BlsThresholdSigner.cpp
HristoStaykov: Definitely Smart pointers are the right way to go. Here I try to fix the current leaks without major changes, might be a separate task should be made to do the conversion from raw pointers to smart pointers?

bftengine/src/bftengine/Bitmap.hpp
HristoStaykov: Yes, will add proper Assertions and remove the if.

threshsign/src/bls/relic/BlsThresholdSigner.cpp
andrewjstone: Totally. No need to make that change in this commit.

bftengine/src/bftengine/PersistentStorageDescriptors.cpp
HristoStaykov: Not really necessary, a Bitmap stores only an int and a pointer, could be allocated on the stack.

HristoStaykov: Factory method is now removed and a proper constructor is added.

PR #545, opened 2020-04-26
Add client id to the client metrics component

yontyon:
Needed to distinguish between clients that run on the same process.


f-squirrel:
LGTM!

PR #547, opened 2020-04-28
PlainUDPImpl::sendAsyncMessage error handle

oded282:
Added log error in PlainUDPImpl::sendAsyncMessage.
In case message size exceed 64k the message can't be sent by UDP, therefore I added logs to describe that the message couldn't be sent.

### DIFF
--- a/communication/src/PlainUDPCommunication.cpp
+++ b/communication/src/PlainUDPCommunication.cpp
@@ -71,6 +71,9 @@ class PlainUDPCommunication::PlainUdpImpl {
 
   NodeNum selfId;
 
+  // Max UDP packet bytes can be sent, without headers.
+  static constexpr uint16_t MAX_UDP_PAYLOAD_SIZE = 65535 - 20 - 8;
+
   /** Flag to indicate whether the current communication layer still runs. */
   std::atomic<bool> running;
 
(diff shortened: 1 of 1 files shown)


communication/src/PlainUDPCommunication.cpp
```diff
@@ -71,6 +71,9 @@ class PlainUDPCommunication::PlainUdpImpl {
 
   NodeNum selfId;
 
+  // Max UDP packet bytes can be sent.
+  const uint16_t dataLimit = 65507;
```
andrewjstone, line 75: Where did this number come from? 2^16 == 65536.

oded282: UDP datagram composed of:
65507 data bytes
20 bytes IP header
8 bytes UDP header

teoparvanov: Let's make this more explicit (and static), as in `static constexpr uint16_t MAX_UDP_PAYLOAD_SIZE = 65535 - 20 - 8`

communication/src/PlainUDPCommunication.cpp
```diff
@@ -233,6 +236,11 @@ class PlainUDPCommunication::PlainUdpImpl {
   int sendAsyncMessage(const NodeNum &destNode, const char *const message, const size_t &messageLength) {
     int error = 0;
 
+    if (messageLength > MAX_UDP_PAYLOAD_SIZE) {
+      LOG_INFO(_logger, "Error, exceeded UDP payload size limit, message length: " << std::to_string(messageLength));
```
teoparvanov, line 240: LOG_ERROR

PR #549, opened 2020-04-29
Use local S3 service for Apollo RO replica tests

tdimitrov:
The patch contains the following changes:
* The S3 service is minio.
* If concord-bft is compiled with USE_S3_OBJECT_STORE enabled, the
latest version of the minio binary is downloaded in build dir, while
configuring the build.
* Each RO replica started by Apollo uses special S3 configuration
pointing to the local minio instance. Note that this is true if and only
if concord-bft is compiled with USE_S3_OBJECT_STORE.
* Apollo's BftTestNetwork class is modified to pass the test network
configuration to the start_replica callback. This is required for
distinguishing RO and regular replicas.

I'd like to point out some concerns I have got:
* The minio server binary is fetched during cmake configures the build system. I don't like this solution very much but this is the best automated approach I can think of. Another solution is to just use find_program() and throw an error if the binary is missing, but it will require manual work.
* The minio binary is executed locally as separate process. The alternative is to use Docker but I think it's an overkill. Nice side effect of this approach is that the S3 bucket is just a dir on the filesystem and it can be easily inspected from a test.

### DIFF
--- a/tests/apollo/CMakeLists.txt
+++ b/tests/apollo/CMakeLists.txt
@@ -1,4 +1,23 @@
-set(APOLLO_TEST_ENV "BUILD_ROCKSDB_STORAGE=${BUILD_ROCKSDB_STORAGE}")
+set(APOLLO_TEST_ENV "BUILD_ROCKSDB_STORAGE=${BUILD_ROCKSDB_STORAGE} USE_S3_OBJECT_STORE=${USE_S3_OBJECT_STORE}")
+set(MINIO_BINARY_PATH "${CMAKE_BINARY_DIR}/minio")
+
+if(USE_S3_OBJECT_STORE)
+        find_program(CHMOD_PROG chmod)
+        if(NOT CHMOD_PROG)
+                message(FATAL_ERROR "chmod not found")
+        endif()
+        
+        if(NOT EXISTS ${MINIO_BINARY_PATH})
+                message(STATUS "Downloading minio binary")
+                file(DOWNLOAD "https://dl.min.io/server/minio/release/linux-amd64/minio" ${MINIO_BINARY_PATH} SHOW_PROGRESS)
+                execute_process(COMMAND ${CHMOD_PROG} 755 ${MINIO_BINARY_PATH})
+        else()
+                message(STATUS "minio binary exists - skipping download")
+        endif()
+
+        # This env var is used by apollo tests to locate the path to the binary
+        set(APOLLO_TEST_ENV "${APOLLO_TEST_ENV} CONCORD_BFT_MINIO_BINARY_PATH=${MINIO_BINARY_PATH}")
+endif()
 
 set(STORAGE_TYPES "v1direct" "v2merkle")
 if (CI_TEST_STORAGE_TYPE)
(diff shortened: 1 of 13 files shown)


tdimitrov:
@toly-kournik @teoparvanov @andrewjstone could you please share your opinions?

tests/apollo/test_skvbc.py
```diff
@@ -21,7 +21,7 @@
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
 
 
-def start_replica_cmd(builddir, replica_id):
+def start_replica_cmd(builddir, replica_id, config):
```
teoparvanov, line 24: Can you make the `config` parameter optional, so that we don't need to pass it every time?

tests/apollo/CMakeLists.txt
```diff
@@ -1,4 +1,23 @@
-set(APOLLO_TEST_ENV "BUILD_ROCKSDB_STORAGE=${BUILD_ROCKSDB_STORAGE}")
+set(APOLLO_TEST_ENV "BUILD_ROCKSDB_STORAGE=${BUILD_ROCKSDB_STORAGE} USE_S3_OBJECT_STORE=${USE_S3_OBJECT_STORE}")
```
teoparvanov, line 1: shouldn't you also pass the `USE_S3_OBJECT_STORE` parameter via travis.yml, so that it runs as part of CI ?

tests/apollo/test_skvbc.py
tdimitrov: I wanted to avoid this too, but I didn't figure out a way to do it.

The problem is that at the place start_replica_cmd is called there is no knowledge if the replica is read-only or not. Initially I tried to add new method - start_ro_replica_cmd() but this way the modification is even more intrusive.

tests/apollo/CMakeLists.txt
tdimitrov: You have a point here. We can do this, but if USE_S3_OBJECT_STORE will be the default behaviour, why not change the build itself to always use it?

teoparvanov: As discussed offline, let's wire this via `.travis.yml`, with USE_S3_OBJECT_STORE always active.

Hint: you can use the `env.global` section to achieve this.

tests/apollo/test_skvbc.py
teoparvanov: Here's the way the `start_replica_cmd()` (and stop_) are implemented in Hermes:
https://gitlab.eng.vmware.com/blockchain/vmwathena_blockchain/blob/master/hermes/suites/apollo_bft_tests.py#L54

teoparvanov: PS: the purpose of the Apollo/Hermes integration was to run the same BFT tests we have, in the context of a containerized VMware Blockchain deployment. You can find more details about this integration here:
https://confluence.eng.vmware.com/pages/viewpage.action?spaceKey=BLOC&title=Apollo+testing+framework#Apollotestingframework-[Apollo&Hermes]TestingBFTaspartofproductCI(viaTEE)

tdimitrov:
I've updated the PR with the comments from @teoparvanov

toly-kournik:
If @teoparvanov has no more concerns I'm good

toly-kournik:
One optional thing is maybe it's better to copy and edit an s3 config file to the working dir in order the test to be self contained

PR #550, opened 2020-04-29
Introduce block deletion (pruning) interfaces

dartdart26:
Introduce the ILocalKeyValueStorageReadOnly::getGenesisBlock() method to
aid with changing genesis block ID when deleting blocks.

Introduce the IBlocksDeleter interface and implement it in ReplicaImp.
It provides methods allowing an application to delete the genesis block
or a block range starting from the genesis block up to a certain block.


kvbc/src/ReplicaImp.cpp
```diff
@@ -153,6 +155,30 @@ Status ReplicaImp::addBlock(const SetOfKeyValuePairs &updates, BlockId &outBlock
...
+BlockId ReplicaImp::deleteBlocksUntil(BlockId until) {
+  const auto genesisBlock = m_bcDbAdapter->getGenesisBlockId();
+  if (genesisBlock == 0) {
+    throw std::logic_error{"Cannot delete a block range from an empty blockchain"};
+  } else if (until <= genesisBlock) {
+    throw std::invalid_argument{"Invalid 'until' value passed to deleteBlocksUntil()"};
+  }
+
+  const auto lastBlock = getLastBlock();
+  const auto lastDeletedBlock = std::min(lastBlock, until - 1);
```
toly-kournik, line 175: So theoretically the whole blockchain can be deleted?

dartdart26: In theory, currently we can delete N - 1 blocks in the merkle tree adapter, where N is the number of blocks in the blockchain. We need to leave at least one in order to not create blocks with the same ID in subsequent addBlock() calls.

In terms of `deleteBlocksUntil()` - yes, you are right, we can delete them all.

PR #551, opened 2020-04-29
Fix RocksDB Client to take ownership of Comparator object.

HristoStaykov:
Make Client take ownership of the Comparator object it is passed in construction. Up until now Comparator object is allocated on the heap and not properly deallocated in most use cases.

### DIFF
--- a/bftengine/tests/metadataStorage/metadataStorage_test.cpp
+++ b/bftengine/tests/metadataStorage/metadataStorage_test.cpp
@@ -96,10 +96,10 @@ TEST(metadataStorage_test, multi_write) {
 
 int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
-  KeyComparator *kk = new KeyComparator(new DBKeyComparator);
+  auto kk = make_unique<KeyComparator>(new DBKeyComparator);
   const string dbPath = "./metadataStorage_test_db";
   remove(dbPath.c_str());
-  Client *dbClient = new Client(dbPath, kk);
+  Client *dbClient = new Client(dbPath, std::move(kk));
   dbClient->init();
   metadataStorage =
       new DBMetadataStorage(dbClient, std::make_unique<concord::storage::v1DirectKeyValue::MetadataKeyManipulator>());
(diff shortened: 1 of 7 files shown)


kvbc/src/direct_kv_storage_factory.cpp
```diff
@@ -32,8 +32,8 @@ namespace {
 auto createRocksDBClient(const std::string &dbPath) {
   // Since the client doesn't own the comparator, use the same instance for all clients. This is RocksDB's default
   // behavior - using a static instance returned by BytewiseComparator(). See the ::rocksdb::Options class.
```
dartdart26, line 34: Can we please remove this comment as it is no longer relevant?

bftengine/tests/metadataStorage/metadataStorage_test.cpp
```diff
@@ -96,10 +96,10 @@ TEST(metadataStorage_test, multi_write) {
 
 int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
-  KeyComparator *kk = new KeyComparator(new DBKeyComparator);
+  auto kk = make_unique<KeyComparator>(new DBKeyComparator);
```
dartdart26, line 99: [Minor]: I think we can remove the `kk` variable and directly call `make_unique()` when constructing the `Client`.

kvbc/test/multiIO_test.cpp
```diff
@@ -64,14 +64,13 @@ class multiIO_test : public ::testing::Test {
  protected:
   void SetUp() override {
     keyGen_.reset(new concord::kvbc::v1DirectKeyValue::RocksKeyGenerator);
-    comparator_ = new KeyComparator(new DBKeyComparator());
-    dbClient.reset(new Client(dbPath_, comparator_));
+    auto comparator_ = std::make_unique<KeyComparator>(new DBKeyComparator{});
```
dartdart26, line 67: [Minor]: I think we can remove the `comparator_` variable and call `std::make_unique()` directly when constructing the `Client`.

PR #552, opened 2020-04-29
Fix all tests to run with sanitizers

HristoStaykov:
[WIP] Fixes for errors reported by tests compiled with clang sanitizers in order to make them pass correctly.
For every fix is a description from which test it was reported and what the problem was.


f-squirrel:
LGTM!
The need to use `std::weak_ptr<Aggregator> aggregator_;` seems to be a sign of redundant usage of `std::shared_ptr`.
Probably, we need to reconsider using it there.

andrewjstone:
> LGTM!
> Btw, the need to use `std::weak_ptr<Aggregator> aggregator_;` seems to be a sign of redundant usage of `std::shared_ptr`.
> Probably, we need to reconsider using it there.

I think the weak pointer is appropriate. The components must share a pointer to the same aggregate in some manner. The only other option is a raw pointer, which is much less supportable, since there's no way to handle its destruction safely. In this case, when the aggregator gets destroyed on shutdown, at least the components will not misuse the pointer.

If there's a better way, I'm certainly open to it, but I think this scenarios is exactly why weak pointers exist.

PR #553, opened 2020-04-30
Added span context size to ClientReplyMsgHeader

f-squirrel:
@andrewjstone thanks for pointing out the problem!


f-squirrel:
@andrewjstone , fixed

PR #555, opened 2020-05-03
Add metrics for dashboards

yontyon:
Here we add new metrics to concord-bft in order to support more dashboards

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2872,7 +2872,7 @@ ReplicaImp::ReplicaImp(bool firstTime,
       metric_not_enough_client_requests_event_{metrics_.RegisterCounter("notEnoughClientRequestsEvent")},
       metric_total_finished_consensuses_{metrics_.RegisterCounter("totalOrderedRequests")},
       metric_total_slowPath_{metrics_.RegisterCounter("totalSlowPaths")},
-      metric_total_fastPath_{metrics_.RegisterCounter("totalFastPath")} {
+      metric_total_fastPath_{metrics_.RegisterCounter("totalFastPaths")} {
   Assert(config_.replicaId < config_.numReplicas);
   // TODO(GG): more asserts on params !!!!!!!!!!!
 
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2867,7 +2870,9 @@ ReplicaImp::ReplicaImp(bool firstTime,
       metric_sent_fullCommitProof_msg_due_to_reqMissingData_{
           metrics_.RegisterCounter("sentFullCommitProofMsgDueToReqMissingData")},
       metric_not_enough_client_requests_event_{metrics_.RegisterCounter("notEnoughClientRequestsEvent")},
-      metric_total_finished_consensuses_{metrics_.RegisterCounter("totalOrderedRequests")} {
+      metric_total_finished_consensuses_{metrics_.RegisterCounter("totalOrderedRequests")},
+      metric_total_slowPath_{metrics_.RegisterCounter("totalSlowPaths")},
+      metric_total_fastPath_{metrics_.RegisterCounter("totalFastPath")} {
```
teoparvanov, line 2875: hw doeslowPathCount

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2867,7 +2870,9 @@ ReplicaImp::ReplicaImp(bool firstTime,
       metric_sent_fullCommitProof_msg_due_to_reqMissingData_{
           metrics_.RegisterCounter("sentFullCommitProofMsgDueToReqMissingData")},
       metric_not_enough_client_requests_event_{metrics_.RegisterCounter("notEnoughClientRequestsEvent")},
-      metric_total_finished_consensuses_{metrics_.RegisterCounter("totalOrderedRequests")} {
+      metric_total_finished_consensuses_{metrics_.RegisterCounter("totalOrderedRequests")},
+      metric_total_slowPath_{metrics_.RegisterCounter("totalSlowPaths")},
```
teoparvanov, line 2874: How does `totalSlowPaths` differ from the existing `slowPathCount` ? Aren't those redundant?

yontyon: The slowPathCounts seems to be not accurate. (we have seen in the dashboards that sometimes sloePathCount > lastExecuteSeqNum which makes no sense.
I've prefered not to touch slowPathCount as there are tests that rely upon its value. and instead, I've added the totalSlowPaths which count the slow path only after the execution phase.

teoparvanov: Thanks, we'll also have to switch the fast/slow path Apollo tests to your new metrics. They seem to make more sense.

bftengine/src/bftengine/ReplicaImp.cpp
teoparvanov, line 2875: let's also make it consistent - should be totalFastPath**s**

PR #557, opened 2020-05-04
Add logs to TlsTcpComm

oded282:
After discussion with @salieri11, I replaced the assertions with log errors and exceptions.
Crashing the replica in this case seems to be the right way to handle this situation, if a replica send's message that exceeds the message range size, there is nothing we can do inside 'AsyncTlsConnection::send' function.
I replaced the asserts with exceptions because we would like the replica to stop in RELEASE mode as well.

### DIFF
--- a/communication/src/TlsTCPCommunication.cpp
+++ b/communication/src/TlsTCPCommunication.cpp
@@ -910,8 +910,15 @@ class AsyncTlsConnection : public std::enable_shared_from_this<AsyncTlsConnectio
    * @param length data length
    */
   void send(const char *data, uint32_t length) {
-    assert(data);
-    assert(length > 0 && length <= _maxMessageLength - MSG_HEADER_SIZE);
+    if (!data) {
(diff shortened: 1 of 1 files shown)


communication/src/TlsTCPCommunication.cpp
```diff
@@ -910,8 +910,15 @@ class AsyncTlsConnection : public std::enable_shared_from_this<AsyncTlsConnectio
    * @param length data length
    */
   void send(const char *data, uint32_t length) {
-    assert(data);
-    assert(length > 0 && length <= _maxMessageLength - MSG_HEADER_SIZE);
+    if (!data) {
+      LOG_ERROR(_logger, "Error, message has never been initialized")
+      throw std::runtime_error("data = nullptr");
```
robem, line 915: Both conditions check input parameters. Using `std::invalid_argument` might be better suited?!

oded282: std::invalid_argument seems suitable, thanks!

PR #560, opened 2020-05-04
Fix pre-execution issues

yuliasherman:
- Fix a crash caused by a null pointer dereference
- Fix wrong message size when header name different than "Header" is used
- Fix messages conversion not identified correctly by ReplicaImp
- Decrease possible primary replica crush timeout detection to allow retries to pass successfully
-  Ignore client retries when any message from the same client is in progress
- Unify asserts use in the simpleKVBC code


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -252,15 +254,13 @@ void PreProcessor::onMessage<ClientPreProcessRequestMsg>(ClientPreProcessRequest
...
+    if (clientEntry->reqProcessingStatePtr) {
       const ReqId &ongoingReqSeqNum = clientEntry->reqProcessingStatePtr->getReqSeqNum();
-      if (ongoingReqSeqNum != reqSeqNum) {
-        LOG_WARN(GL,
-                 " reqSeqNum=" << reqSeqNum << " from clientId=" << clientId
-                               << " is ignored: previous client request=" << ongoingReqSeqNum << " is in process");
-        preProcessorMetrics_.preProcReqIgnored.Get().Inc();
-        return;
-      }
+      LOG_INFO(GL,
```
teoparvanov, line 259: LOG_DEBUG, maybe (INFO looks a bit excessive inside an onMessage() method)?

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
```diff
@@ -19,7 +19,7 @@ using namespace std;
 using namespace concord::util;
 using namespace bftEngine;
 
-uint16_t PreProcessReplyMsg::maxReplyMsgSize_ = 512;  // Actually, it is sizeof(PreProcessReplyMsgHeader) = 50 + 256
+uint16_t PreProcessReplyMsg::maxReplyMsgSize_ = 1024;  // Actually, it is sizeof(Header) = 58 + 256 + cid.size()
```
teoparvanov, line 22: The comment is not very clear (because of the mix of constants end expressions, I think). Can we say instead:

`Actually we should have maxReplyMsgSize_ >= size(header) + size(signature) + size(cid) >= 58 + 256 + ? bytes`

PS: I'm not sure what the size of the CID is :)

tests/apollo/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ foreach(STORAGE_TYPE ${STORAGE_TYPES})
           WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
   add_test(NAME skvbc_preexecution_tests_${STORAGE_TYPE} COMMAND sh -c
-          "env STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_preexecution 2>&1 > /dev/null"
+          "env STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_preexecution"
```
teoparvanov, line 53: this shouldn't be part of the PR

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
yuliasherman: Me either, but we need to limit the message size somehow.

tests/apollo/CMakeLists.txt
yuliasherman: You are right.

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
teoparvanov: I see. My main suggestion is to clarify the comment - I'm not questioning the need of a `maxReplyMsgSize_`.

yuliasherman: OK

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: All ClientRequest 'ignore' messages are 'INFO' in ReplicaImp.

teoparvanov: OK, thanks for the clarification!

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
teoparvanov: Nice, thanks!

PR #563, opened 2020-05-05
Support empty values and key deletion in v2 Merkle

dartdart26:
Add support for empty values in the v2MerkleTree DBAdapter . Keys with
empty values are no longer treated as deleted keys.

In addition, support an explicit parameter with a set of keys to delete
in the DBAdapter . If a key is deleted in block N, find a leaf for the
same key with a previous version and, if there is such a leaf, mark it
as deleted in block N. When calling getValue() for block X, if we find a
leaf for the deleted key and its 'deleted in block' value is less than
or equal to X, throw a NotFoundException . Else, return the value as per
normal, even if it is marked as deleted.

Marking keys as deleted is done by serializing a 'deleted in block ID'
integer in the key's value itself (the 'DatabaseLeafValue' structure).

Keys passed to addBlock() for deletion that are not part of the
blockchain will not be saved in the new block and there will be no
information kept that the user tried to delete non-existent keys.

Deleted keys are serialized in raw blocks in order to support state
transfer. Deleted keys are ordered before serialization in order to have
deterministic raw blocks, having equal hashes.

Provide tests that prove the functionality works as expected.

Decrease both the max input size and the test count from the default
value of 100 to 50 for the merkle DBAdapter property test in order to
speed it up.

### DIFF
--- a/kvbc/include/merkle_tree_block.h
+++ b/kvbc/include/merkle_tree_block.h
@@ -17,7 +17,7 @@ RawBlock create(const SetOfKeyValuePairs &updates,
                 const BlockDigest &parentDigest,
                 const sparse_merkle::Hash &stateHash);
 
-// Creates a block that adds a set of key/values and a deletes a set of keys.
+// Creates a block from a set of key/value pairs and a set of keys to delete.
 RawBlock create(const SetOfKeyValuePairs &updates,
                 const OrderedKeysSet &deletes,
                 const BlockDigest &parentDigest,
@@ -93,14 +93,14 @@ inline bool operator==(const Node &lhs, const Node &rhs) {
 }
 
 // Merkle-specific data added to raw blocks.
-struct RawBlockMerklelData {
+struct RawBlockMerkleData {
   static constexpr auto MIN_SIZE = sparse_merkle::Hash::SIZE_IN_BYTES;
   static constexpr auto STATE_HASH_SIZE = sparse_merkle::Hash::SIZE_IN_BYTES;
   static constexpr auto MIN_KEY_SIZE = sizeof(KeyLengthType);
 
-  RawBlockMerklelData() = default;
+  RawBlockMerkleData() = default;
 
-  RawBlockMerklelData(const sparse_merkle::Hash &pStateHash, const OrderedKeysSet &pDeletedKeys = OrderedKeysSet{})
+  RawBlockMerkleData(const sparse_merkle::Hash &pStateHash, const OrderedKeysSet &pDeletedKeys = OrderedKeysSet{})
       : stateHash{pStateHash}, deletedKeys{pDeletedKeys} {}
 
   sparse_merkle::Hash stateHash;
@@ -108,7 +108,7 @@ struct RawBlockMerklelData {
   OrderedKeysSet deletedKeys;
 };
 
-inline bool operator==(const RawBlockMerklelData &lhs, const RawBlockMerklelData &rhs) {
+inline bool operator==(const RawBlockMerkleData &lhs, const RawBlockMerkleData &rhs) {
   return lhs.stateHash == rhs.stateHash && lhs.deletedKeys == rhs.deletedKeys;
 }
 
--- a/kvbc/include/merkle_tree_block.h
+++ b/kvbc/include/merkle_tree_block.h
@@ -12,14 +12,22 @@
 
 namespace concord::kvbc::v2MerkleTree::block::detail {
 
-// Creates a block that includes a set of key/values. The passed parentDigest buffer must be of size BLOCK_DIGEST_SIZE
-// bytes.
+// Creates a block that adds a set of key/values.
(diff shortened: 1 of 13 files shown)


kvbc/test/sparse_merkle_storage/db_adapter_property_test.cpp
```diff
@@ -325,6 +317,69 @@ TEST_P(db_adapter_kv_tests, get_all_keys_at_last_version) {
   ASSERT_TRUE(rc::check(test));
 }
 
+// Test that keys are:
+//  * available before key deletion
+//  * unavailable at the deletion block version and subsequent block versions that don't add the same keys
+//  * available again if added in subsequent block versions
+TEST_P(db_adapter_kv_tests, key_deletion) {
```
andrewjstone, line 324: This is a fantastic test! Thank you for this.

kvbc/include/merkle_tree_block.h
```diff
@@ -12,15 +12,23 @@
...
 namespace concord::kvbc::v2MerkleTree::block::detail {
 
-// Creates a block that includes a set of key/values. The passed parentDigest buffer must be of size BLOCK_DIGEST_SIZE
-// bytes.
+// Creates a block that adds a set of key/values.
 RawBlock create(const SetOfKeyValuePairs &updates,
                 const BlockDigest &parentDigest,
                 const sparse_merkle::Hash &stateHash);
 
+// Creates a block that adds a set of key/values and a deletes a set of keys.
```
robem, line 20: The latter is probably "a set of keys to delete"

kvbc/src/direct_kv_block.cpp
```diff
@@ -114,14 +114,33 @@ SetOfKeyValuePairs getData(const Sliver &block) {
...
 }
 
+Sliver getUserData(const Sliver &block) {
+  constexpr auto headerSize = sizeof(detail::Header);
+  assert(block.length() >= headerSize);
+  const auto header = reinterpret_cast<const detail::Header *>(block.data());
+
+  // If there are no elements, we only have a Header and, optionally, user data.
+  if (header->numberOfElements == 0) {
+    assert(block.length() >= headerSize);
```
robem, line 132: This has been checked above already.

kvbc/include/merkle_tree_block.h
```diff
@@ -84,6 +92,26 @@ inline bool operator==(const Node &lhs, const Node &rhs) {
          lhs.stateRootVersion == rhs.stateRootVersion && lhs.keys == rhs.keys;
 }
 
+// Merkle-specific data added to raw blocks.
+struct RawBlockMerklelData {
```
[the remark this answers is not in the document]
dartdart26: Thanks, fixing that!

kvbc/src/merkle_tree_block.cpp
```diff
@@ -12,22 +12,36 @@ namespace concord::kvbc::v2MerkleTree::block::detail {
...
+// Use the v1DirectKeyValue implementation and just add the merkle-specific data at the back. We can do that, because
+// users are not expected to interpret the returned buffer themselves.
+
 RawBlock create(const SetOfKeyValuePairs &updates, const BlockDigest &parentDigest, const Hash &stateHash) {
-  SetOfKeyValuePairs out;
-  return v1DirectKeyValue::block::detail::create(
-      updates, out, parentDigest, stateHash.dataArray().data(), stateHash.dataArray().size());
+  const auto merkleData = v2MerkleTree::detail::serialize(RawBlockMerklelData{stateHash});
+  auto out = SetOfKeyValuePairs{};
+  return v1DirectKeyValue::block::detail::create(updates, out, parentDigest, merkleData.data(), merkleData.size());
```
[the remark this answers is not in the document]
dartdart26: Yes, that's correct.

kvbc/src/merkle_tree_block.cpp
```diff
@@ -12,22 +12,36 @@ namespace concord::kvbc::v2MerkleTree::block::detail {
...
+}
+
+RawBlock create(const SetOfKeyValuePairs &updates,
+                const OrderedKeysSet &deletes,
+                const BlockDigest &parentDigest,
+                const Hash &stateHash) {
+  const auto merkleData = v2MerkleTree::detail::serialize(RawBlockMerklelData{stateHash, deletes});
+  auto out = SetOfKeyValuePairs{};
```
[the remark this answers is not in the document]
dartdart26: Yes, agreed. It is being used in the v1 DB adapter. We can improve the API and have two functions. I will do that in a separate commit.

kvbc/include/merkle_tree_block.h
dartdart26: Thanks, fixing it!

kvbc/src/direct_kv_block.cpp
dartdart26: Great catch, removing it!

PR #574, opened 2020-05-10
Fix a race condition in pre-execution

yuliasherman:
- Add Apollo test for concurrent requests from different clients.
   https://jira.eng.vmware.com/browse/BC-1710
- Fix a race condition caused a crash

### DIFF
--- a/tests/apollo/test_skvbc_preexecution.py
+++ b/tests/apollo/test_skvbc_preexecution.py
@@ -95,8 +95,8 @@ class SkvbcPreExecutionTest(unittest.TestCase):
         bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
 
-        max_concurrency = len(bft_network.clients) // 2
+        max_concurrency = len(bft_network.clients)
         clients = bft_network.random_clients(max_concurrency)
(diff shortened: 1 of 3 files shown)


tests/apollo/test_skvbc_preexecution.py
```diff
@@ -44,24 +46,57 @@ def start_replica_cmd(builddir, replica_id):
...
+    async def test_concurrent_pre_process_requests(self, bft_network):
+        """
+        Launch concurrent requests from different clients in parallel. Ensure that created block are as expected.
+        """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        max_concurrency = len(bft_network.clients) // 2
+        clients = bft_network.random_clients(max_concurrency)
+        num_of_requests = 50 * len(clients)
```
teoparvanov, line 100: Why `50 *`? Won't this result in most requests being dropped, because we can only pre-process up to `num_clients` requests?

yuliasherman: Changed.

teoparvanov:
Thanks Yulia, looks great!

PR #577, opened 2020-05-11
Change logging of client request status in Replica

andrewjstone:
The bft client resends messages if it hasn't received enough replies
after a certain time. These retries are normal and expected. An adaptive
timeout system should have the client learning and sending them at
different rates. Non-the less replicas may receive them semi-frequently.
Rather than logging these at info level, this commit reduces to logging
them at DEBUG level.

Furthermore, since overflowing the primary queue is a much more
significant condition, we log that at warning level.


jkk-vmw:
Thank you for addressing this!

andrewjstone:
Closing in favor of a new request that adds this an other logs.

PR #580, opened 2020-05-12
Avoid parallel pre-execution retries on a primary replica

yuliasherman:
- In case the primary replica's calculated pre-execution result is different from that reached a consensus, we retry the pre-execution on the primary. Pre-execution consensus requires the collection of (f + 1) equal responses. When additional responses arrive after we reached the consensus, pre-execution retry should not be triggered again.

- Added Apollo test for verifying a view change while running the pre-execution

https://jira.eng.vmware.com/browse/BC-2694
https://jira.eng.vmware.com/browse/BC-2535

### DIFF
--- a/tests/apollo/test_skvbc_preexecution.py
+++ b/tests/apollo/test_skvbc_preexecution.py
@@ -30,7 +30,7 @@ def start_replica_cmd(builddir, replica_id):
     """
 
     status_timer_milli = "500"
-    view_change_timeout_milli = "10000"
+    view_change_timeout_milli = "5000"
 
     path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
     return [path,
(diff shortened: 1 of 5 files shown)


teoparvanov:
Thanks Yulia, the new name makes more sense IMO!

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -49,13 +49,25 @@ async def send_single_read(self, skvbc, client):
...
+                pass
+            await trio.sleep(.1)
+
+    async def send_single_write_with_pre_execution_and_kv(self, skvbc, kv, client):
         reply = await client.write(skvbc.write_req([], kv, 0), pre_process=True)
         reply = skvbc.parse_reply(reply)
         self.assertTrue(reply.success)
 
+    async def send_single_write_with_pre_execution(self, skvbc, client):
+        kv = [(skvbc.random_key(), skvbc.random_value()),
```
teoparvanov, line 67: `kv` -> `write_set`

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -49,13 +49,25 @@ async def send_single_read(self, skvbc, client):
...
+    async def send_indefinite_pre_execution_requests(self, skvbc, clients):
+        while True:
+            client = random.choice(list(clients))
+            try:
+                await self.send_single_write_with_pre_execution(skvbc, client)
+            except:
+                pass
+            await trio.sleep(.1)
+
+    async def send_single_write_with_pre_execution_and_kv(self, skvbc, kv, client):
```
teoparvanov, line 61: `kv` -> `write_set`

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -98,3 +110,44 @@ async def test_concurrent_pre_process_requests(self, bft_network):
...
+        """
+        Crash the primary replica and verify that the system triggers a view change and moves to a new view
+        """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        clients = bft_network.clients.values()
+        client = random.choice(list(clients))
+        key_before_vc = skvbc.random_key()
+        value_before_vc = skvbc.random_value()
+        kv = [(key_before_vc, value_before_vc)]
```
teoparvanov, line 126: write_set

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -30,7 +30,7 @@ def start_replica_cmd(builddir, replica_id):
     """
 
     status_timer_milli = "500"
-    view_change_timeout_milli = "10000"
+    view_change_timeout_milli = "5000"
```
teoparvanov, line 33: why 5sec here? With less than 10 seconds I've seen weird view change behaviors. View change can be a slow process (in the order of 10-s of seconds)

tests/apollo/test_skvbc_preexecution.py
yuliasherman: ok

tests/apollo/test_skvbc_preexecution.py
yuliasherman: ok

tests/apollo/test_skvbc_preexecution.py
yuliasherman: ok

tests/apollo/test_skvbc_preexecution.py
yuliasherman: Should be OK with status_timer_milli of 500ms, but I'll change.

PR #581, opened 2020-05-13
Add zlib1g-dev as install depenency

tdimitrov:
On vanilla bionic installation cmake fails with this error:

    Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR)

I don't know why this works in travis, probably the package is in their default image.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@ set -u
 
 install_basic_deps() {
   sudo apt-get update
-  sudo apt-get install -y ccache cmake clang clang-format-9 python3-pip python3-setuptools libgmp3-dev g++ parallel libs3-dev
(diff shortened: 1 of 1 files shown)


f-squirrel:
@tdimitrov , isn't this `libz-dev`?

tdimitrov:
@f-squirrel you are right, its virtual. Is it more future-proof to use the virtual package instead of zlib1g-dev?

f-squirrel:
@tdimitrov ,
While working on the `dockerized Concord-BFT` I've used `zlib-dev`.
However, if your install it apt says the following:
```
(base) ~|⇒ sudo apt-get install libz-dev

[sudo] password for ddanilov:
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'zlib1g-dev' instead of 'libz-dev'
```
So I have no preferences, probably `libz-dev` is deprecated, let's leave is as is.

PR #582, opened 2020-05-13
Bug fix: State transfer test fails in Travis CI 

toly-kournik:
v1DirectKeyValue::DBAdapter::deleteBlock asserted when deleting the last block.
This happened most probably during replica sync state when crashes at the very start.

### DIFF
--- a/kvbc/src/direct_kv_db_adapter.cpp
+++ b/kvbc/src/direct_kv_db_adapter.cpp
@@ -449,7 +449,7 @@ void DBAdapter::deleteBlock(const BlockId &blockId) {
     if (Status s = db_->multiDel(keysVec); !s.isOK())
       throw std::runtime_error(__PRETTY_FUNCTION__ + std::string(": failed: blockId: ") + std::to_string(blockId) +
                                std::string(" reason: ") + s.toString());
-    Assert(hasBlock(blockId - 1));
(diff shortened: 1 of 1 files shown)


kvbc/src/direct_kv_db_adapter.cpp
```diff
@@ -449,7 +449,8 @@ void DBAdapter::deleteBlock(const BlockId &blockId) {
     if (Status s = db_->multiDel(keysVec); !s.isOK())
       throw std::runtime_error(__PRETTY_FUNCTION__ + std::string(": failed: blockId: ") + std::to_string(blockId) +
                                std::string(" reason: ") + s.toString());
-    Assert(hasBlock(blockId - 1));
+    if (blockId > 1)
+      Assert(hasBlock(blockId - 1));
     setLatestBlock(blockId - 1);
     setLastReachableBlockNum(blockId - 1);
```
dartdart26, line 455: I think this code will set the latest/last reachable blocks incorrectly if `latest != last reachable` and if `blockId` is not the `latest` one prior to the call. I guess we cannot set the latest and last reachable reliably here. Maybe we should get them from storage after the delete call. I guess if we do that, we need a DB transaction for that to be safe. Please let me know if I am missing anything.

andrewjstone:
Thanks for the quick turnaround Toly! Much appreciated :)

PR #585, opened 2020-05-13
Opentracing wrapper

f-squirrel:
This commit adds a wrapper around Opentracing classes and functions.
Rationale:
* If the feature is compiled out(`-DUSE_OPENTRACING=OFF`), the wrapper has to behave as a dummy object.
* If the tracer is not initialized by the user of Concord-BFT library, the wrapper has to handle it.

Changes:
[+] `SpanWrapper` class with a few free functions for a span creation

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -8,7 +8,8 @@ add_library(util STATIC
     src/histogram.cpp
     src/status.cpp
     src/sliver.cpp
-    src/hex_tools.cpp)
+    src/hex_tools.cpp
+    src/OpenTracing.cpp)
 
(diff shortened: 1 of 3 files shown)


toly-kournik:
It's great that we have only one optional interface library dependency instead of the whole bunch of libs. 
Personally, I'd prefer an interface for open tracing with inherited  NullTracing, OpenTracing and others if required. 
What's missing for my understanding is its usage.

f-squirrel:
Summary:
During an offline conversation agreed to leave the implementation as-is.
In case we have more than one kind of behavior the class will be refactored to meet Toly's proposal

salieri11:
LGTM as well

PR #590, opened 2020-05-14
Add logs to tryToSendPrePrepareMsg.

oded282:
Added log to tryToSendPrePrepareMsg.
If the request size is bigger than the remaining size in the PrePrepare message, it won't be added.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -299,23 +299,38 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
   PrePrepareMsg *pp =
       new PrePrepareMsg(config_.replicaId, curView, (primaryLastUsedSeqNum + 1), firstPath, primaryCombinedReqSize);
 
-  ClientRequestMsg *nextRequest = requestsQueueOfPrimary.front();
-  while (nextRequest != nullptr && nextRequest->size() <= pp->remainingSizeForRequests()) {
-    MDC_CID_PUT(GL, nextRequest->getCid());
-    if (clientsManager->noPendingAndRequestCanBecomePending(nextRequest->clientProxyId(),
-                                                            nextRequest->requestSeqNum())) {
-      pp->addRequest(nextRequest->body(), nextRequest->size());
-      clientsManager->addPendingRequest(nextRequest->clientProxyId(), nextRequest->requestSeqNum());
+  uint16_t initialStorageForRequests = pp->remainingSizeForRequests();
+
+  ClientRequestMsg *nextRequest = (!requestsQueueOfPrimary.empty() ? requestsQueueOfPrimary.front() : nullptr);
+
+  while (nextRequest != nullptr) {
+    if (nextRequest->size() <= pp->remainingSizeForRequests()) {
+      MDC_CID_PUT(GL, nextRequest->getCid());
+      if (clientsManager->noPendingAndRequestCanBecomePending(nextRequest->clientProxyId(),
+                                                              nextRequest->requestSeqNum())) {
+        pp->addRequest(nextRequest->body(), nextRequest->size());
+        clientsManager->addPendingRequest(nextRequest->clientProxyId(), nextRequest->requestSeqNum());
+      }
+      primaryCombinedReqSize -= nextRequest->size();
+    } else if (nextRequest->size() > initialStorageForRequests) {
+      // The message is too big
+      LOG_ERROR(GL,
+                "Request sent by: " << nextRequest->senderId() << ", Request size is: " << nextRequest->size()
+                                    << ", PrePrepare maximum size allowed for requests is : "
+                                    << initialStorageForRequests << ", request was ignored");
     }
-    primaryCombinedReqSize -= nextRequest->size();
     delete nextRequest;
     requestsQueueOfPrimary.pop();
-    nextRequest = (requestsQueueOfPrimary.size() > 0 ? requestsQueueOfPrimary.front() : nullptr);
+    nextRequest = (!requestsQueueOfPrimary.empty() ? requestsQueueOfPrimary.front() : nullptr);
+  }
+
+  if (pp->numberOfRequests() == 0) {
+    LOG_ERROR(
+        GL, "Number of PrePrepare requests added is: " << pp->numberOfRequests() << ", none of the requests were sent");
+    return;
   }
 
   pp->finishAddingRequests();
-
-  Assert(pp->numberOfRequests() > 0);
 
   if (config_.debugStatisticsEnabled) {
     DebugStatistics::onSendPrePrepareMessage(pp->numberOfRequests(), requestsQueueOfPrimary.size());


oded282:
Hi Andrew, thank you for your comment.
I wanted to error-handle specific situation, which the first message sent is bigger than the capacity of PrePrepare message, I understand it would invoke more than enough log messages.
I'll do it more elegant.

oded282:
Hi @andrewjstone ,
I made some changes in the PR. I am working on a bug that occurs when the first requests sent by the client, is bigger than the initial PrePrepare size for requests. We assert this situation and that may lead us to fail the replica, it means that the replica will crash if a client sent by mistake or as a malicious behaviour a message that is to big. After consultation with part of the team, my fix ignores the first requests which are to big and error-handles the situation.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -293,23 +293,34 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
...
+  while (nextRequest != nullptr) {
+    if (nextRequest->size() <= pp->remainingSizeForRequests()) {
+      MDC_CID_PUT(GL, nextRequest->getCid());
+      if (clientsManager->noPendingAndRequestCanBecomePending(nextRequest->clientProxyId(),
+                                                              nextRequest->requestSeqNum())) {
+        pp->addRequest(nextRequest->body(), nextRequest->size());
+        clientsManager->addPendingRequest(nextRequest->clientProxyId(), nextRequest->requestSeqNum());
+      }
+      primaryCombinedReqSize -= nextRequest->size();
+    } else if (pp->numberOfRequests() <= 0) {  // case the first message is to big
```
andrewjstone, line 307: Grammar: `to` -> `too`

As another nit here, clang-format tend to have trouble if the comment is appended to a line, rather than on its own line. I'd reword the comment to be: "The first message is too big." and then put it on the next line.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -117,7 +117,6 @@ void PrePrepareMsg::addRequest(const char* pRequest, uint32_t requestSize) {
 void PrePrepareMsg::finishAddingRequests() {
   Assert(!isNull());
   Assert(!isReady());
-  Assert(b()->numberOfRequests > 0);
```
andrewjstone, line 120: I don't think you should remove this Assert. This method should not be called with no pre-prepares. Even in the code you modified you return if there are no requests before finalizing. Otherwise, we enter this code, and then immediately send a `PrePrepareMsg` with no actual requests. In that case we don't want to send the PrePrepareMsg at all. There are of course other ways to perform this logic, but it seems that the rest of the code assumes this function is a gate to prevent generating useless PrePrepareMsgs and then sending them.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -293,23 +293,34 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
...
+      LOG_ERROR(GL,
+                "Request sent by: " << nextRequest->senderId() << " Request size is: " << nextRequest->size()
```
andrewjstone, line 309: I think you should make this error message more explicit. It should say why the request is being dropped. It's not because it doesn't fit in the PrePrepare, it's because its larger than the maximum allowed size of a request.

oded282: done.

bftengine/src/bftengine/ReplicaImp.cpp
oded282: done.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
oded282: I understand. thanks

teoparvanov:
LGTM overall, not sure about the corner cases though...

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -299,23 +299,36 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
...
+  while (nextRequest != nullptr) {
+    if (nextRequest->size() <= pp->remainingSizeForRequests()) {
+      MDC_CID_PUT(GL, nextRequest->getCid());
+      if (clientsManager->noPendingAndRequestCanBecomePending(nextRequest->clientProxyId(),
+                                                              nextRequest->requestSeqNum())) {
+        pp->addRequest(nextRequest->body(), nextRequest->size());
+        clientsManager->addPendingRequest(nextRequest->clientProxyId(), nextRequest->requestSeqNum());
+      }
+      primaryCombinedReqSize -= nextRequest->size();
+    } else if (pp->numberOfRequests() <= 0) {
```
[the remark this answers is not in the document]
oded282: - In the current implementation probably won’t be negative value of requests.
The number of requests type is an int, I thought that it will be safer in case of unpredictable changes in the future, what do you think?

- The bug triggered once a big request was the first one to pop, in this case the request would have stuck in the queue, number of requests would have been zero and the assert in ‘finishAddingRequest()’ would have crash our replica. But I think you are right it will be helpful to log the next big requests as well. Thanks.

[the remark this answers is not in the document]
teoparvanov: * For now let's have a separate error log if `pp->numberOfRequests < 0`, as it's an invariant breach
* I suggest to log an error every time a big request fails to be added to the PP message, not just for the first one.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -299,23 +299,45 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
...
+  while (nextRequest != nullptr) {
+    if (nextRequest->size() <= pp->remainingSizeForRequests()) {
+      MDC_CID_PUT(GL, nextRequest->getCid());
+      if (clientsManager->noPendingAndRequestCanBecomePending(nextRequest->clientProxyId(),
+                                                              nextRequest->requestSeqNum())) {
+        pp->addRequest(nextRequest->body(), nextRequest->size());
+        clientsManager->addPendingRequest(nextRequest->clientProxyId(), nextRequest->requestSeqNum());
+      }
+      primaryCombinedReqSize -= nextRequest->size();
+    } else if (nextRequest->size() > initialStorageForRequests) {
```
teoparvanov, line 315: why not just `else` here? It would be equivalent to `else if nextRequest->size() > pp->remainingSizeForRequests()`. If I understand correctly, `remainingSizeForRequests()` decreases with every new request you add to the PP message, whereas `initialStorageForRequests` always holds the initial storage size.

oded282: This bug I'm trying to fix here occur when a client send a request that is too big to fit the space in PP, the current implementation doesn't handle those requests and leads to failure of the replica. 
As Andrew mentioned in his first comment, "during batching requests aren't evenly sized they unlikely to perfectly fill the available space in a pre-prepare", so it may lead us to spamming the logs. I want to log just those request that are too big to be added to the initial PP space.

teoparvanov: Alright, makes sense, thanks for the clarification.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -299,23 +299,45 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
...
+    return;
+  }
 
-  Assert(pp->numberOfRequests() > 0);
+  if (pp->numberOfRequests() < 0) {
+    LOG_ERROR(GL,
+              "Error, PrePrepare 'numberOfRequests' value is: "
+                  << pp->numberOfRequests() << ", number of requests shouldn't be a negative value!");
+    return;
+  }
```
teoparvanov, line 338: let's remove this - it doesn't make much sense at all...

teoparvanov:
@andrewjstone I don't have the entire context around this change... It looks good to me, but can you please check it fulfills its intended goal, and if so - merge it? Thanks!

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: I don't think we need this either. If you want to use an assert, that's fine, but a whole clause is pointless.

andrewjstone:
> @andrewjstone I don't have the entire context around this PR... It looks good to me, but can you please check it fulfills its intended goal, and if so - merge it? Thanks!

I'll merge it when CI passes.

PR #592, opened 2020-05-16
[WIP] GitHub actions experimenting

tdimitrov:
This is for testing purposes only, please don't merge.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,0 +1,42 @@
+name: Concord-BFT build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        compiler: [gcc-7, clang]
+        CI_BUILD_TYPE:
+          - "-DCMAKE_BUILD_TYPE=DEBUG -DCI_TEST_STORAGE_TYPE=v1direct"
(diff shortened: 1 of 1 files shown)


tdimitrov:
Thanks @andrewjstone 
I think we can't see any action here until the workflow is committed. However here is a run from m fork: https://github.com/tdimitrov/concord-bft/actions/runs/110535377

andrewjstone:
> Thanks @andrewjstone
> I think we can't see any action here until the workflow is committed. However here is a run from m fork: https://github.com/tdimitrov/concord-bft/actions/runs/110535377

Cool. Unfortunately it looks like each job took about twice as long. I wonder if that's related to the underlying hardware perf. The good news is the overall CI run only took about 20% longer since all jobs were able to run in parallel.

tdimitrov:
Are you comparing the overall running times? I noticed that the built I've sent has a cache miss so it includes provisioning. At quick glance build and test times look similar.
I'll try to run another build which will use the cache and compare the results.

andrewjstone:
> Are you comparing the overall running times? I noticed that the built I've sent has a cache miss so it includes provisioning. At quick glance build and test times look similar.
> I'll try to run another build which will use the cache and compare the results.

Yes the overall running times. Thanks!

PR #594, opened 2020-05-17
WIP: State Transfer:

toly-kournik:
- Allow transferring empty vblock, i.e. vblock containing a header only.
  This is a very rare scenario wich happens in conjunction with a View Change. 
- Apollo: stop replicas with terminate() - SIGTERM rather than with kill() - SIGKILL in order to record the stop event in replica's log.
- Assert macros AssertGE, AssertGT, AssertEQ, AssertLT, AssertLE to record arguments in case of assert failure.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -401,7 +401,6 @@ DataStore::CheckpointDesc BCStateTran::createCheckpointDesc(uint64_t checkpointN
   Assert(lastBlock == as_->getLastBlockNum());
   metrics_.last_block_.Get().Set(lastBlock);
 
-  LOG_DEBUG(STLogger, "last block = " << lastBlock);
   STDigest digestOfLastBlock;
 
   if (lastBlock > 0) {
@@ -419,6 +418,12 @@ DataStore::CheckpointDesc BCStateTran::createCheckpointDesc(uint64_t checkpointN
   checkDesc.lastBlock = lastBlock;
   checkDesc.digestOfLastBlock = digestOfLastBlock;
   checkDesc.digestOfResPagesDescriptor = digestOfResPagesDescriptor;
+
+  LOG_DEBUG(STLogger,
+            "CheckpointDesc: "
+                << "checkpoint: " << checkpointNumber << " lastBlock: " << lastBlock
+                << " digestOfLastBlock: " << digestOfLastBlock.toString()
+                << " digestOfResPagesDescriptor: " << digestOfResPagesDescriptor.toString());
 
   return checkDesc;
 }
@@ -1305,7 +1310,7 @@ bool BCStateTran::onMessage(const FetchResPagesMsg *m, uint32_t msgLen, uint16_t
 
   uint32_t vblockSize = getSizeOfVirtualBlock(vblock, config_.sizeOfReservedPage);
 
-  Assert(vblockSize > sizeof(HeaderOfVirtualBlock));
+  AssertGE(vblockSize, sizeof(HeaderOfVirtualBlock));
   Assert(checkStructureOfVirtualBlock(vblock, vblockSize, config_.sizeOfReservedPage));
 
   // compute information about next chunk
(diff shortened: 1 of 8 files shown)


storage/include/storage/db_interface.h
```diff
@@ -37,7 +37,13 @@ class ITransaction {
    public:
     Guard(ITransaction* t) : txn_(t) {}
     virtual ~Guard() {
-      if (!std::uncaught_exception()) txn_->commit();
+      if (!std::uncaught_exception()) {
+        try {
+          txn_->commit();
+        } catch (const std::exception& e) {
+          printf("EXCEPTION: %s\n", e.what());
```
teoparvanov, line 44: LOG_ERROR ?

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -1305,7 +1310,7 @@ bool BCStateTran::onMessage(const FetchResPagesMsg *m, uint32_t msgLen, uint16_t
 
   uint32_t vblockSize = getSizeOfVirtualBlock(vblock, config_.sizeOfReservedPage);
 
-  Assert(vblockSize > sizeof(HeaderOfVirtualBlock));
+  AssertGE(vblockSize, sizeof(HeaderOfVirtualBlock));
```
teoparvanov, line 1313: Is the change here intentional? (GE means >=)

tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -199,7 +199,8 @@ async def test_ro_replica_start_simultaneously (self, bft_network, tracker):
 
     @with_trio
     @with_bft_network(start_replica_cmd=start_replica_cmd, num_ro_replicas=1)
-    async def test_ro_replica_with_s3_failures(self, bft_network):
+    #temporary disabled until is disabled when USE_S3_OBJECT_STORE is unset
+    async def _test_ro_replica_with_s3_failures(self, bft_network):
```
teoparvanov, line 203: As discussed in your other PR, please use @unittest.skip to disable the test.

storage/include/s3/client.hpp
```diff
@@ -165,9 +165,9 @@ class Client : public concord::storage::IDBClient {
   }
 
   concordUtils::Status multiPut(const SetOfKeyValuePairs& _keyValueMap) override {
-    ITransaction::Guard g(beginTransaction());
-    for (auto&& pair : _keyValueMap) g.txn()->put(pair.first, pair.second);
-
+    std::unique_ptr<ITransaction> txn(beginTransaction());
```
teoparvanov, line 168: Why not use ITransaction::Guard here, given that you have fixed it in case of commit exception?

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: Yes

bftengine/src/bcstatetransfer/BCStateTran.cpp
toly-kournik: yes

storage/include/s3/client.hpp
toly-kournik: Because I want the exception to be thrown and it's impossible from a ITransaction::~Guard

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: will do

storage/include/s3/client.hpp
toly-kournik: Since I'm performing a commit only if not during uncaught exception, I'd probably should to the following:
virtual ~Guard() **noexcept(false)** {
      if (!std::uncaught_exception()) txn_->commit();
      delete txn_;
    }

storage/include/storage/db_interface.h
toly-kournik: changed

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: resolved in https://github.com/vmware/concord-bft/pull/595

teoparvanov:
Hi Toly, is this PR related to a specific bug?

andrewjstone:
Thanks Toly! This is great.

PR #595, opened 2020-05-17
Disable S3 Object Store in Apollo

toly-kournik:


tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -199,7 +199,8 @@ async def test_ro_replica_start_simultaneously (self, bft_network, tracker):
 
     @with_trio
     @with_bft_network(start_replica_cmd=start_replica_cmd, num_ro_replicas=1)
-    async def test_ro_replica_with_s3_failures(self, bft_network):
+    #temporary disabled until is disabled when USE_S3_OBJECT_STORE is unset
+    async def _test_ro_replica_with_s3_failures(self, bft_network):
```
teoparvanov, line 203: Please disable the test using Python decoration, instead of changing the method name:
```python
@unittest.skip("unstable")
@with_trio
@with_bft_network(start_replica_cmd=start_replica_cmd, num_ro_replicas=1)
```

PR #596, opened 2020-05-18
Add spans to Concord-BFT

f-squirrel:
This PR adds Opentracing spans for the consensus path:
* Span context arrives in Concord-BFT with ClientRequest message
* The span is transferred with the corresponding messages for both paths: regular(aka slow) and fast.
* Once `CommitFull` or `FullCommitProof` message is received,
    a span context is extracted and used for the creation of span.
    This span is passed to the `execute` method in the form of `concordUtils::SpanWrapper`.

Important notes:
* When Client requests are batched, only the span of the last request is sent as part of the PrePrepare message.
* The same applies to all the Partial* messages.

### DIFF
--- a/bftengine/src/bftengine/messages/PartialProofsSet.cpp
+++ b/bftengine/src/bftengine/messages/PartialProofsSet.cpp
@@ -13,7 +13,9 @@
 // TODO(GG): this class/file should be replaced by an instance of CollectorOfThresholdSignatures
 
 #include "PartialProofsSet.hpp"
+#include <type_traits>
 #include "Logger.hpp"
+#include "OpenTracing.hpp"
 #include "PartialCommitProofMsg.hpp"
 #include "FullCommitProofMsg.hpp"
 #include "Crypto.hpp"
@@ -208,13 +210,15 @@ class AsynchProofCreationJob : public util::SimpleThreadPool::Job {
                          IThresholdAccumulator* acc,
                          Digest& expectedDigest,
                          SeqNum seqNumber,
-                         ViewNum viewNumber) {
+                         ViewNum viewNumber,
+                         const std::string& span_context) {
     this->me = myReplica;
     this->acc = acc;
     this->expectedDigest = expectedDigest;
     this->seqNumber = seqNumber;
     this->view = viewNumber;
     this->verifier = verifier;
+    span_context_ = span_context;
   }
 
   virtual ~AsynchProofCreationJob(){};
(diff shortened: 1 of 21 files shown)


kvbc/include/KVBCInterfaces.h
```diff
@@ -135,7 +135,8 @@ class ICommandsHandler : public bftEngine::IRequestsHandler {
                       const char* request,
                       uint32_t maxReplySize,
                       char* outReply,
-                      uint32_t& outActualReplySize) = 0;
+                      uint32_t& outActualReplySize,
+                      concordUtils::SpanWrapper& span) = 0;
```
teoparvanov, line 139: This is going to be a breaking change for the product. So right after merging this into master, please also bump concord-bft in the product and make the necessary changes.

util/include/OpenTracing.hpp
```diff
@@ -30,21 +31,49 @@ class SpanWrapper {
...
+#ifdef USE_OPENTRACING
+    return (impl().get() != nullptr);
+#else
+    return false;
+#endif
+  }
+
+  template <typename ValueT>
+  void setTag(const std::string& name, const ValueT& value) {
+#ifdef USE_OPENTRACING
```
teoparvanov, line 45: As this is not included in the Travis CI matrix, I suppose you have tested both cases at least locally?

bftengine/src/bftengine/messages/PartialProofsSet.cpp
```diff
@@ -208,20 +210,24 @@ class AsynchProofCreationJob : public util::SimpleThreadPool::Job {
...
+                         ViewNum viewNumber,
+                         const std::string& span_context) {
     this->me = myReplica;
     this->acc = acc;
     this->expectedDigest = expectedDigest;
     this->seqNumber = seqNumber;
     this->view = viewNumber;
     this->verifier = verifier;
+    // DD: not a part of the init list to preserve the initialization order
+    span_context_ = span_context;
```
teoparvanov, line 222: Can you explain a bit more - I don't get what's happening here?

kvbc/include/KVBCInterfaces.h
f-squirrel: Yes, I have a commit for the product.

util/include/OpenTracing.hpp
f-squirrel: Yes, I did not want to add more to the matrix.
The compilation key is tested locally.

bftengine/src/bftengine/messages/PartialProofsSet.cpp
f-squirrel: The variable is initialized in the constructor but not in the initializer list.
I did it because it is the last declared in the class.
Likely, my comment is useless and it is better to remove it.

PR #599, opened 2020-05-19
View change test fixes

teoparvanov:
The failures were happening because once the replicas agree on the new view and activate it, they go ahead and rebuild the entire active window by exchanging `ReqMissingDataMsg`. This may take a significant amount of time (10-s of seconds for 7 replicas) - especially problematic if the 7 replicas are running on 4 cores as in Travis... 

Essentially, this means right after view change has been agreed, the replicas are **practically staggered**, until this active window rebuilding process completes. The test wouldn't previously take this into account. In some cases it would pass if 2f+1 replicas had rebuilt their active window faster than the others. However, in some cases 5 seconds (the UDP client's request timeout) wouldn't be enough, and the test would fail intermittently.

The following fixes are suggested as part of this PR:
* fixing the "current active view" metric to be incremented when we actually start working in the new view
* retry the post-view change write (fails initially because the replicas are rebuilding the active window after the view change and a 2f+1 reply quorum cannot be reached)

### DIFF
--- a/tests/apollo/test_skvbc_view_change.py
+++ b/tests/apollo/test_skvbc_view_change.py
@@ -158,7 +158,7 @@ class SkvbcViewChangeTest(unittest.TestCase):
                 err_msg="Make sure view change has been triggered."
             )
 
-            await tracker.tracked_read_your_writes()
+            await self._wait_for_read_your_writes_success(tracker)
 
             await tracker.run_concurrent_ops(100)
 
@@ -207,7 +207,7 @@ class SkvbcViewChangeTest(unittest.TestCase):
             err_msg="Make sure view change has been triggered."
         )
 
-        await tracker.tracked_read_your_writes()
+        await self._wait_for_read_your_writes_success(tracker)
 
         await tracker.run_concurrent_ops(100)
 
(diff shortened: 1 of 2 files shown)


tests/apollo/test_skvbc_view_change.py
```diff
@@ -158,7 +158,11 @@ async def test_single_vc_primary_isolated(self, bft_network, tracker):
                 err_msg="Make sure view change has been triggered."
             )
 
-            await tracker.tracked_read_your_writes()
+            with trio.fail_after(seconds=60):
+                while True:
+                    with trio.move_on_after(seconds=5):
```
HristoStaykov, line 163: Might be, make this less than the client timeout of 5 seconds, just to be sure it will be the first to expire?

tests/apollo/test_skvbc_view_change.py
```diff
@@ -207,7 +211,11 @@ async def test_single_vc_with_f_replicas_down(self, bft_network, tracker):
             err_msg="Make sure view change has been triggered."
         )
 
-        await tracker.tracked_read_your_writes()
+        with trio.fail_after(seconds=60):
+            while True:
+                with trio.move_on_after(seconds=5):
```
HristoStaykov, line 216: Same as above.

tests/apollo/test_skvbc_view_change.py
```diff
@@ -344,7 +352,11 @@ async def _single_vc_with_consecutive_failed_replicas(
...
```
HristoStaykov, line 357: Same as above.

teoparvanov: now I also catch Exception - which includes trio.TooSlowError, so we're covered

andrewjstone:
Thanks Teo!

PR #601, opened 2020-05-20
set initial values for SkvbcReadOnlyReplicaTest.test_ro_replica_start…

toly-kournik:
…_simultaneously

as it was before s3 tweaks


tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -183,7 +183,7 @@ async def test_ro_replica_start_simultaneously (self, bft_network, tracker):
         # await tracker.skvbc.tracked_fill_and_wait_for_checkpoint(initial_nodes=bft_network.all_replicas(), checkpoint_num=1)     
         with trio.fail_after(seconds=60):
             async with trio.open_nursery() as nursery:
-                nursery.start_soon(tracker.run_concurrent_ops, 900, 1)
+                nursery.start_soon(tracker.send_indefinite_tracked_ops)
                 while True:
```
teoparvanov, line 187: How about extracting the `while True` loop in a separate async method, and also invoke it with `nursery.start_soon()`? This way, both the filling and checking of the lastExecutedSeqNum would happen concurrently.

PR #602, opened 2020-05-20
Make intermediate calls to onTransferComplete during state transfer

tdimitrov:
If a client generates requests indefinitely the RO replica never tries
to catch up and lastExectuedSeqNum metric is never updated.
This patch changes this behaviour so that onTransferComplete is called
after a Checkpoint is transferred suqqessfully. As a result
lastExecutedSeqNum slowly increases.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -1392,7 +1392,7 @@ bool BCStateTran::onMessage(const RejectFetchingMsg *m, uint32_t msgLen, uint16_
 
   Assert(sourceSelector_.isPreferred(replicaId));
 
-  LOG_WARN(STLogger, "Removing replica " << replicaId << " from preferred replicasa");
+  LOG_WARN(STLogger, "Removing replica " << replicaId << " from preferred replicas");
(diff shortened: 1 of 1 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -1997,6 +1997,8 @@ void BCStateTran::processData() {
         Assert(getFetchingState() == FetchingState::GettingMissingResPages);
 
         LOG_DEBUG(STLogger, "moved to GettingMissingResPages");
+        DataStore::CheckpointDesc cp = g.txn()->getCheckpointBeingFetched();
+        replicaForStateTransfer_->onTransferringComplete(cp.checkpointNum);
```
toly-kournik, line 2001: I'm not sure it's a right behaviour: onTransferring complete is called only after the last block which is a vBlock

andrewjstone: Agreed. I think this might cause all sorts of problems. We probably need a different solution to this.

tdimitrov: You are right. I'm closing this PR.

PR #605, opened 2020-05-20
Add a diagnostics server and client

andrewjstone:
We have a diagnostics Registrar that allows us to add status handlers
useful for debugging a live system. This commit adds a TCP server
designated to listen on localhost, that can process commands destined
for the status registrar and return the output over a socket. For
simplicity, and because this is supposed to be a user driven
operation, each client connection is only allowed to send one
command. The client should then wait for the response and exit.

Additionally, a cli client, `concord-ctl`, is provided that takes
arguments on the command line and forwards them as commands to the
server.

Note that neither the registrar or server is integrated in either
concord-bft or the product yet. This will come in follow up commits,
which will also contain real status handlers.

A test server executable that registers 2 dummy status handlers was created for manual testing.

The server can be run from the build directory:
`./build/diagnostics/test/diagnostics_server_main`

The corresponding client can be run from `concord-bft/diagnostics`. Some
examples are shown below.

```
andrewstone@ubuntu:~/concord-bft/diagnostics$ ./concord-ctl
Usage: concord-ctl <SUBJECT> <COMMAND> [ARGS]

  status <COMMAND> [ARGS]

    status get <KEY1> [KEY2]..[KEY_N]
        Get the status of the given key(s).

    status describe [KEY1] [KEY2]..[KEY_N]
        Get the description of the given keys, or all keys if none is given.

    status list-keys
        List all status keys.

andrewstone@ubuntu:~/concord-bft/diagnostics$ ./concord-ctl status list-keys
handler1
handler2

andrewstone@ubuntu:~/concord-bft/diagnostics$ ./concord-ctl status get handler1
handler1 called

andrewstone@ubuntu:~/concord-bft/diagnostics$ ./concord-ctl status get handler2
handler2 called

andrewstone@ubuntu:~/concord-bft/diagnostics$ ./concord-ctl status get handler1 handler2
handler1 called
handler2 called

andrewstone@ubuntu:~/concord-bft/diagnostics$ ./concord-ctl status describe handler1 handler2
handler 1 description
handler 2 description

andrewstone@ubuntu:~/concord-bft/diagnostics$ ./concord-ctl status describe

handler1
  handler 1 description

handler2
  handler 2 description
```

Finally, testing for timeouts and large messages was done directly from
a python shell.

```
>>> s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> s.connect(("127.0.0.1", 6888))
>>> s.send(too_large)
1025
>>> data = s.recv(1024)
>>> print(data)
b'Error: Request exceeded max size: 1024\n'
>>> s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)

>>> s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> s.connect(("127.0.0.1", 6888))
>>> data = s.recv(1024)
>>> print(data)
b'Error: timeout\n'
```

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -1,0 +1,172 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+// This file contains a tcp server listening on localhost that handles the messages in protocol.h.
+
+#pragma once
+
+#include <atomic>
+#include <cassert>
+#include <chrono>
+#include <cstring>
+#include <future>
+#include <sstream>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <thread>
+#include <unistd.h>
+
+#include "Logger.hpp"
+#include "errnoString.hpp"
+#include "protocol.h"
+
+using concordUtils::errnoString;
+
+namespace concord::diagnostics {
+
+static constexpr uint16_t PORT = 6888;
+static constexpr int BACKLOG = 5;
+static constexpr size_t MAX_INPUT_SIZE = 1024;
+
+static concordlogger::Logger logger = concordlogger::Log::getLogger("concord.diagnostics");
+
+// Returns a successfully read line as a string.
+// Throws a std::runtime_error on error.
+std::string readline(int sock) {
+  std::array<char, MAX_INPUT_SIZE> buf;
+  buf.fill(0);
+  int count = 0;
+  auto start = std::chrono::steady_clock::now();
+  auto timeout = std::chrono::microseconds(999999);
+  auto remaining = timeout;
+  while (true) {
+    fd_set read_fds;
+    FD_ZERO(&read_fds);
+    FD_SET(sock, &read_fds);
+    timeval tv;
+    tv.tv_sec = 0;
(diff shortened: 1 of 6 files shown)


andrewjstone:
Additionally, we should probably ignore SIGPIPE in our main signal setup. It doesn't cause any problems on ubuntu, since I don't perform a write on a broken socket more than once, but I'm not sure this behavior is portable. AFAICT, SIGPIPE, is useless for our purposes.

diagnostics/include/diagnostics_server.h
```diff
@@ -0,0 +1,166 @@
...
+    FD_ZERO(&read_fds);
+    FD_SET(sock, &read_fds);
+    timeval tv;
+    tv.tv_sec = 0;
+    tv.tv_usec = remaining.count();
+    auto rv = select(sock + 1, &read_fds, NULL, NULL, &tv);
+    if (rv == 0) {
+      throw std::runtime_error("timeout");
+    }
+    if (rv < 0 && errno == EINTR) continue;
```
robem, line 61: What makes `EINTR` special in this case?

diagnostics/include/diagnostics_server.h
```diff
@@ -0,0 +1,166 @@
...
+    auto end = std::chrono::steady_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+    remaining = timeout - duration;
+  }
+}
+
+void handleRequest(const Registrar& registrar, int sock) {
+  try {
+    std::string command = readline(sock);
+    std::stringstream ss(command);
```
robem, line 92: Those two can be combined

diagnostics/include/diagnostics_server.h
andrewjstone: EINTR is not actually an error. It can happen due to a signal interrupting any blocking syscall. I'm not sure how often this happens anymore, but I've seen and written this code many times where EINTR is checked and ignored. 

See https://github.com/tokio-rs/mio/issues/494 and https://github.com/tokio-rs/mio/issues/16

PR #607, opened 2020-05-21
Simulate long-running transactions with pre-execution

yuliasherman:
- Add a new test for simulation of a long-running pre-execution request and verify that it's processing does not cause the system to enter a new view.
- Combine cleaning of expired requests and possible primary replica failure into one operation to support long-running requests. Before this change, those requests were not handled properly: after the expiration of preProcessReqWaitTimeMilli_ of 500 ms passed to ReplicaImp handling and completed without preprocessing.
- Added some (const &) qualifier to shared pointers that are not expected to take ownership.

https://jira.eng.vmware.com/browse/BC-2601

### DIFF
--- a/tests/apollo/test_skvbc_preexecution.py
+++ b/tests/apollo/test_skvbc_preexecution.py
@@ -18,6 +18,11 @@ import random
 from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
 from util import skvbc as kvbc
 
+NUM_OF_SEQ_WRITES = 100
+MAX_CONCURRENCY = 10
+SHORT_REQ_TIMEOUT_MILLI = 3000
+LONG_REQ_TIMEOUT_MILLI = 15000
+
 def start_replica_cmd(builddir, replica_id):
     """
     Return a command that starts an skvbc replica when passed to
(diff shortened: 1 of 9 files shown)


tests/apollo/test_skvbc_preexecution.py
```diff
@@ -82,44 +87,72 @@ async def run_concurrent_pre_execution_requests(self, skvbc, clients, num_of_req
...
     @with_bft_network(start_replica_cmd)
     async def test_view_change(self, bft_network):
         """
-        Crash the primary replica and verify that the system triggers a view change and moves to a new view
+        Crash the primary replica and verify that the system triggers a view change and moves to a new view.
         """
         bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        await trio.sleep(5)
```
teoparvanov, line 154: Why is this sleep necessary, right after starting the replicas?

teoparvanov:
Some additional comments...

tests/apollo/test_skvbc_preexecution.py
yuliasherman: I realized that the thread pool is not always ready immediately after replicas startup and the first request gets a timeout. It's not an issue in the real environment but causes tests to be unstable.

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -82,44 +87,72 @@ async def run_concurrent_pre_execution_requests(self, skvbc, clients, num_of_req
...
         """
         bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
 
-        for i in range(100):
+        for i in range(NUM_OF_SEQ_WRITES):
             client = bft_network.random_client()
             await self.send_single_write_with_pre_execution(skvbc, client)
 
+    @unittest.skip("unstable")
```
teoparvanov, line 107: PS: is this intentionally skipped?

yuliasherman: Yes. This test reproduces an issue we discussed earlier: ( f + 1) against (2f + 1) required pre-processing/client quorum inconsistency.

PR #611, opened 2020-05-22
Added support for constant client requests in tests

HristoStaykov:
by means of a decorator, creating a client in a separate
set from regular clients used by tests.


tests/apollo/util/bft.py
```diff
@@ -85,6 +86,18 @@ def trio_wrapper(*args, **kwargs):
 
     return trio_wrapper
 
+def with_constant_client_requests(async_fn):
```
teoparvanov, line 89: let's call this `with_constant_load`

teoparvanov, line 89: also, you could add a document string to the decorator

teoparvanov:
Let's rebase and have this merged into master!

PR #613, opened 2020-05-22
Fix diagnostics server compilation in release mode

dartdart26:
If compiling in release mode and asserts in Server::listen() are not
evaluated, the unused variable error kicks in. Additionally, the server
would be broken in release mode as the calls inside the asserts would be
removed.

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -21,6 +21,7 @@
 #include <cstring>
 #include <future>
 #include <sstream>
+#include <utility>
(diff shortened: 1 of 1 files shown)


dartdart26:
> Thanks @dartdart26.
> Now that we have logging, I think it might be better to log the errors with errnoString rather than using the void cast hack.

@andrewjstone Good idea, I agree. I've pushed another commit that logs the errors and exits on socket call failures in `listen()`. Rationale is explained in the commit message. Please let me know if that interferes with the intended behaviour.

PR #615, opened 2020-05-23
Attempt a fix for hanging CI jobs

teoparvanov:
My initial suspicion is the recently introduced use of SIGTERM instead of SIGKILL for stopping replicas.


teoparvanov:
The goal was improved debuggability by gracefully stopping the replicas, but this alters the test behavior:

https://github.com/vmware/concord-bft/pull/594

PR #616, opened 2020-05-23
Add utility to stream values as formatted kvpairs

andrewjstone:
This is useful for adding structured log information at the end of log
lines.

Rather than forcing us to do things like this:

```
LOG_INFO(STLogger,
         "Init BCStateTran object:"
             << " maxNumOfStoredCheckpoints_=" << maxNumOfStoredCheckpoints_ << " numberOfReservedPages_="
             << numberOfReservedPages_ << " config_.sizeOfReservedPage_=" << config_.sizeOfReservedPage);
```

We get nice formatting automatically by doing this:

```
  LOG_INFO(STLogger,
           "Init BCStateTran object:" << KVLOG(
               maxNumOfStoredCheckpoints, numberOfReservedPages_, config_.sizeOfReservedPage));
```

The latter outputs:

```
<LOG PREAMBLE> "Init BCStateTran object: numOfStoredCheckpoints: <VAL>, numOfReservedPages_: <VAL>, config_.sizeOfReservedPage: <VAL>"
```

where `<VAL>` is the actual value of the variable.

It is anticipated, that along with a type trait that determines whether
an expression is streamable, that we can use this to also output the
executed values of expressions in asserts rather than just the
expressions themselves. This is saved for a future commit.

Please note that this is not a replacement for MDC. This is only for printing one off values for a specific log line. We should use them both appropriately, and avoid duplicating anything already in MDC.

PR #617, opened 2020-05-23
 Print expression values in asserts when possible

andrewjstone:
### This includes the commit for #616. Thats should be merged first.

Currently our assert macros only print the value of expressions for
boolean parameters.

However, this is insufficient for our macros that take 2 parameters,
where we want to know what the value of each parameter is: e.g.
`AssertEQ`, `AssertGE`, `AssertGT`, `AssertLT`, `AssertLE`, and the
newly added `AssertNE`.

The macros listed above will now print the values as well as the
expressions in a key value form of "expression: value" for each
expression parameter, if the expression is capable of being written to a
stream. In other words, the value will be written if expression has an
`<<` operator. Otherwise, the value will be written as an underscore,
`_`.

This enhancement for tolerating unprintable values was made to the KvLog
template functions in kvstream.h introduced in #616. Those functions
will now no longer prevent compilation if someone adds an unprintable
type to a KvLog call. This is somewhat unfortunate, but necessary to
allow the macro wrappers in assertUtils.h to work AFAICT.

I tested this with a few different AssertEQ calls:

```
// Two unprintable lvalues
auto busted_replicas = psd_->getReplicas();
busted_replicas.insert(33);
AssertEQ(replicas_, busted_replicas);

// Log output:
53: 3|05-23-2020 22:31:04.544|139700104681984|FATAL|concord|void bftEngine::SimpleBlockchainStateTransfer::impl::BCStateTran::checkConsistency(bool)| AssertEQ| busted_replicas: _, replicas_: _ in function checkConsistency (/home/andrewstone/concord-bft/bftengine/src/bcstatetransfer/BCStateTran.cpp 2131)
```

```
// Two printable lvalues
auto busted_replica_id = 50;
AssertEQ(config_.myReplicaId, busted_replica_id);

// Log output:
53: 3|05-23-2020 22:34:30.500|140625076410880|FATAL|concord|void bftEngine::SimpleBlockchainStateTransfer::impl::BCStateTran::checkConsistency(bool)| AssertEQ| busted_replica_id: 50, config_.myReplicaId: 3 in function checkConsistency (/home/andrewstone/concord-bft/bftengine/src/bcstatetransfer/BCStateTran.cpp 2133)
```

```
// A printable lvalue, and a printable rvalue
auto busted_replica_id = 50;
AssertEQ(config_.myReplicaId, busted_replica_id + 1);

// Log output:
53: 2|05-23-2020 22:43:02.692|140552776358400|FATAL|concord|void bftEngine::SimpleBlockchainStateTransfer::impl::BCStateTran::checkConsistency(bool)| AssertEQ| busted_replica_id + 1: 51, config_.myReplicaId: 2 in function checkConsistency (/home/andrewstone/concord-bft/bftengine/src/bcstatetransfer/BCStateTran.cpp 2133)
```

### DIFF
--- a/util/include/assertUtils.hpp
+++ b/util/include/assertUtils.hpp
@@ -13,11 +13,14 @@
 
 #pragma once
 
-#include "Logger.hpp"
 #include <cxxabi.h>
 #include <cassert>
 #include <cstring>
 #include <execinfo.h>
+#include <sstream>
+
+#include "kvstream.h"
+#include "Logger.hpp"
 
 inline void printCallStack() {
   const uint32_t MAX_FRAMES = 100;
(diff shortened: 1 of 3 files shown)


andrewjstone:
I will make this more flexible in a future commit. Merging so we can use the asserts, etc... in long run tests.

PR #618, opened 2020-05-24
Add storage metrics

yontyon:
Adding storage metrics for the in-memory storage and rocksdb storage.
In addition, I've add a small functionality to the metrics library.
See https://jira.eng.vmware.com/browse/BC-2580


kvbc/src/ReplicaImp.cpp
```diff
@@ -204,7 +204,7 @@ ReplicaImp::ReplicaImp(ICommunication *comm,
   auto dbSet = storageFactory->newDatabaseSet();
   m_bcDbAdapter = std::move(dbSet.dbAdapter);
   m_metadataDBClient = dbSet.metadataDBClient;
-
+  m_metadataDBClient->setAggregator(aggregator);
```
dartdart26, line 207: I think we need to also set an aggregator to the `dataDBClient` from the `dbSet` as it is the one actually used for writing data for the blockchain itself.

storage/include/storage/db_interface.h
```diff
@@ -124,6 +134,18 @@ class IDBClient {
...
+  // Storage metrics
+  // TODO: all metrics have announced as mutable because the get methods are const
+  mutable concordMetrics::Component metrics_;
+  mutable concordMetrics::CounterHandle keys_reads_;
+  mutable concordMetrics::CounterHandle total_read_bytes_;
+  mutable concordMetrics::CounterHandle keys_writes_;
+  mutable concordMetrics::CounterHandle total_written_bytes_;
+  mutable int ops_count = 0;
+  const int update_interval = 100;  // TODO: We update the metrics each 100 ops. Currently this is hardcoded value in
+                                    // the future we may add it to the configuration file
```
dartdart26, line 148: I prefer we don't have any implementation details in the interface. Also, having a public constructor for the interface wouldn't make much sense as users can't properly construct an object via it. Can we maybe separate these details in another class and either contain it or inherit from it in RocksDB and memorydb implementations?

storage/include/rocksdb/client.h
```diff
@@ -59,9 +60,12 @@ class ClientIterator : public concord::storage::IDBClient::IDBClientIterator {
...
 class Client : public concord::storage::IDBClient {
  public:
-  Client(std::string _dbPath) : m_dbPath(_dbPath) {}
+  Client(std::string _dbPath)
+      : m_dbPath(_dbPath), total_db_size_(metrics_.RegisterGauge("storage_rocksdb_db_size", 0)) {}
   Client(std::string _dbPath, std::unique_ptr<const ::rocksdb::Comparator>&& comparator)
-      : m_dbPath(_dbPath), comparator_(std::move(comparator)) {}
+      : m_dbPath(_dbPath),
+        comparator_(std::move(comparator)),
+        total_db_size_(metrics_.RegisterGauge("storage_rocksdb_db_size", 0)) {}
```
dartdart26, line 68: [Minor]: Naming consistency (i.e. camelCase, m_).

storage/src/rocksdb_client.cpp
```diff
@@ -293,6 +299,7 @@ Status Client::multiGet(const KeysVector &_keysVec, OUT ValuesVector &_valuesVec
     }
     _valuesVec.push_back(Sliver(std::move(values[i])));
   }
+  tryToUpdateMetrics();
```
dartdart26, line 302: Will this call update with the correct count of keys fetched (i.e. possibly more than 1)?

dartdart26:
It is great to have these metrics in @yontyon ! Thank you for doing that!

Overall, it looks good. I have two points that I think we should improve:

1. As mentioned in https://github.com/vmware/concord-bft/pull/618#discussion_r431665894, I think we should not have implementation details in the interface.
2. I see that we only have counters in the `Client` methods. However, we fetch data through `ClientIterator`s too (seeking, iterating and so on). If I am not mistaken, we will not count those in properly. Please correct me if I am wrong.

util/include/Metrics.hpp
```diff
@@ -70,6 +70,8 @@ class Gauge {
   explicit Gauge(const uint64_t val) : val_(val) {}
 
   void Set(const uint64_t val) { val_ = val; }
+  void Inc(const uint64_t val) { val_ += val; }
```
andrewjstone, line 73: This should not be part of the public interface. Gauges are not counters and explicitly should *not* be incremented and decremented.

util/include/Metrics.hpp
```diff
@@ -70,6 +70,8 @@ class Gauge {
...
+  void Dec(const uint64_t val) { val_ -= val; }
```
andrewjstone, line 74: Counters should monotonically grow. Why do you need to decrement a counter?

yontyon: This is identical to the Prometheus approach (see https://github.com/jupp0r/prometheus-cpp/blob/master/core/include/prometheus/gauge.h).

It helps to have a ++ like operation for gauges.

util/include/Metrics.hpp
yontyon: See below

kvbc/src/ReplicaImp.cpp
yontyon: This is a shared_ptr. So it doesn't really matter.

storage/src/rocksdb_client.cpp
yontyon: Yeah. It takes the internal rocksdb metrics

storage/include/storage/db_interface.h
yontyon: done

kvbc/src/ReplicaImp.cpp
dartdart26: Yes, you are right. However, that assumes it points to the same object and it might not be the case for all implementations. Please correct me if I am missing something.

yontyon: But in the line before we set ` m_metadataDBClient = dbSet.metadataDBClient;`.
So it does point to the same object...
Am I missing somthing?

util/include/Metrics.hpp
yontyon: Anyway, I noticed that I'm not really using it, so it has been removed

kvbc/src/ReplicaImp.cpp
dartdart26: I meant the other DB client from the set, i.e. `dbSet.dataDBClient` . They can be the same DB client (i.e. `dbSet.dataDBClient == dbSet.dbSet.metadataDBClient`) in case of a normal read-write replica or they can differ, i.e. for a RO replica that stores data on an object store (then the `dbSet.dataDBClient` will be an S3 client).

yontyon: done

dartdart26:
Looks good, thanks for the changes @yontyon ! One last thing I wanted to ask - I see we update counters on put/get/del calls for RocksDB, but not when using iterators. On the other hand, we do that for memorydb. Will RocksDB iterator use be reflected in the count?

yontyon:
@dartdart26 
In concord bft we take once in (approximately) 5 seconds the inner rocksdb metrics.
Therefore it takes all in acount

PR #622, opened 2020-05-26
WIP Re-open failing tests

yuliasherman:


teoparvanov:
Hi Yulia, I've restarted the entire CI run, to make sure we don't let any intermittent failures slip in... Fingers crossed :)

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -100,11 +99,12 @@ async def test_sequential_pre_process_requests(self, bft_network):
         bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
 
+        await trio.sleep(5)
```
teoparvanov, line 102: My only concern with this approach is the entire test will be delayed by a total of 5x3x3 = 45sec

yuliasherman:
I found a bug in the SimpleThreadPool, which caused a quorum issue to show up. 
The pre-processing job just got stuck on some non-primary nodes and never return. In a normal situation, I don't expect this issue to reproduce much. In any case the bug in the pool needs to be resolved. I opened a bug for this issue https://jira.eng.vmware.com/browse/BC-2862.

PR #623, opened 2020-05-27
Make KVLOG functionality more generic

andrewjstone:
Rather than using KvLog calls inside the KVLOG macro, create a generic
KVARGS macro that generates just a parameter list of key-value
parameters for the given string. We can then pass this to any function
with a wrapper macro.

The KVARGS macro was moved to macros.h in util. We can add other system
level macros there when needed.

Additionally, a bugfix was made to allow KVARGS calls with a single
argument, via the use of the `UNUSED` parameter. This is to prevent
errors about variadic macros requiring one argument.

Add forward declarations for KvLog functions so that recursion works
properly at compile time when jumping between printable and
non-printable types.

This change also adds a `Strict` boolean template parameter to the KvLog
function so that we can statically assert in normal KVLOG calls when
passed an argument that is not printable. We add a separate
`KVLOG_FOR_ASSERT` macro that sets this strict parameter to false and allows
asserting on types that are not printable, and only printing the ones
that are. In this latter case we give priority to the Assert over the
printability.

Also, reverse KVARGS macros so args are in correct order.

### DIFF
--- a/util/include/kvstream.h
+++ b/util/include/kvstream.h
@@ -1,254 +1,84 @@
-// Concord
-//
-// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
-//
-// This product is licensed to you under the Apache 2.0 license (the "License").  You may not use this product except in
-// compliance with the Apache 2.0 License.
-//
-// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
-// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
-// file.
-
 #pragma once
 
 #include <sstream>
+
+#include "macros.h"
 #include "type_traits.h"
 
-// Take up to 16 values and output them to a stream as key value pairs.
-//
-// This stringizes the value passed in and uses it as the key. If you want a specific key, just create an alias
-// variable.
-//
-// Usage looks like the following:
-//   KVLOG(seq_num, view, firstStoredCheckpoint)
-//
-// Output looks like:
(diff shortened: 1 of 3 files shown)


util/include/kvstream.h
```diff
@@ -1,254 +1,65 @@
...
+// are allowed. If Strict=true, a static_assert will fire.
+//
 // Right now the entrypoint returns a string, because of the way our logging macros are implemented.
 // We could change them to take an extra parameter that's a stringstream and prevent the additional
 // allocation and copy from returning a string. We'll deal with that if its needed.
-template <typename K,
+template <bool Strict,
+          typename K,
           typename V,
           typename std::enable_if<concord::is_streamable<std::ostream, V>::value>::type * = nullptr>
```
teoparvanov, line 24: Hmm, I'm not sure I understand the use of enable_if here... Isn't there supposed to be a second template param as in enable_if<B, T> ? I don't see who sets the T in your expression :)

andrewjstone: It defaults to void. Since we only care that there is a type, and not what the type is to make things well formed this works.

util/include/kvstream.h
```diff
@@ -1,254 +1,65 @@
...
 void KvLog(std::stringstream &ss, K &&key, V &&val) {
   ss << std::forward<K>(key) << ": " << std::forward<V>(val);
```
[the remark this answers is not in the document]
andrewjstone: Nope. In this case the type implements the `std::ostream<<` operator so the type is printable with and without strict mode.

PR #625, opened 2020-05-27
Add MDC and enhance slow/fast path logs

cloudnoize:
- Adding MDC for primary and path.

- Following the commit path and logging important events.


bftengine/src/bftengine/ControllerWithSimpleHistory.cpp
```diff
@@ -183,7 +193,8 @@ bool ControllerWithSimpleHistory::onEndOfEvaluationPeriod() {
 
   recentActivity.resetAll(maxSeq + 1);
 
-  if (lastFirstPathVal != currentFirstPath) LOG_INFO(GL, "currentFirstPath = " << (int)currentFirstPath);
+  if (lastFirstPathVal != currentFirstPath)
```
salieri11, line 196: can you add "changed from XXX to YYY"?

bftengine/src/bftengine/ControllerBase.hpp
```diff
@@ -45,6 +45,7 @@ class ControllerBase {
   virtual void onMessage(const PreparePartialMsg* m) {}
 
   // TODO(GG): add more methods that may be useful  by controllers
```
robem, line 47: Can we remove this TODO? IMHO, it doesn't describe a concrete action item.

bftengine/src/bftengine/PrimitiveTypes.cpp
```diff
@@ -19,5 +19,20 @@ std::string CommitPathToStr(CommitPath path) {
   }
 }
 
+std::string CommitPathToMDCString(CommitPath path) {
+  switch (path) {
+    case CommitPath::NA:
+      return "NA";
+    case CommitPath::OPTIMISTIC_FAST:
+      return "fast";
```
robem, line 27: You can remove this line, `switch case` is fall-through.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1121,7 +1173,7 @@ void ReplicaImp::onCommitVerifyCombinedSigResult(SeqNum seqNumber, ViewNum v, bo
     ps_->setCommitFullMsgInSeqNumWindow(seqNumber, commitFull);
     ps_->endWriteTran();
   }
-
+  LOG_INFO(GL, "Commit path analysis: request commited, proceeding to try to execute");
```
robem, line 1176: Are we using `INFO` to analyze a current bug? Otherwise this will spam the logs, correct?

robem: More general, should we use `INFO` to log messages in the happy path? I'm afraid that this will just clutter logs.

cloudnoize: It's not for a specific bug.
It's to mark the end of the Consensus and to be able to measure its duration. we have performance issues and we are blind at the analysis of the Consensus.
We just don't know what takes so long

cloudnoize: It will produce a lot of noize.
And I tried to minimize it to significant events. but I don't have a better alternative.
If we'll see it's redundant, I'll reduce its level.

bftengine/src/bftengine/ControllerBase.hpp
cloudnoize: it's Guy's todo I'm afraid to remove it :)

bftengine/src/bftengine/ReplicaImp.cpp
cloudnoize: the "Commit path analysis" prefix is for easy filtering

bftengine/src/bftengine/PrimitiveTypes.cpp
cloudnoize: I believe you meant for the two "fast" is a row.
sure

bftengine/src/bftengine/ReplicaImp.cpp
salieri11: I think that most of recent logs will be moved to DEBUG after system will become stable

andrewjstone: This must be removed ASAP.

PR #628, opened 2020-05-27
WIP: Compute parent block hash in parallel for v2Merkle

dartdart26:
In order to improve performance, compute the parent block hash in
parallel for v2Merkle. This is possible as the parent block hash is not
needed for updating the tree and it can be done during that time, on a
separate thread.

Getting the parent raw block is not done in a separate thread as not
all IDBClient implementations may be thread-safe (an example is
memorydb).

Support a user-supplied thread name and an optional replica ID in
concord::util::Handoff .

Implement the TearDown() method that destructs the DBAdapter in the
'Blockchain' benchmark fixutre as Handoff uses a static storage duration
logger whose destructor will be in a race with Handoff's thread. Reason
for that is that the DBAdapter has a Handoff member.

On my box (i7-9750H CPU @ 2.60GHz), adding a block with 2048
key/value pairs, key size of 32 bytes and value size of 4096 bytes
(i.e. block size is ~8MB), the speedup is around 20-30% .

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -655,7 +655,7 @@ void BCStateTran::handleStateTransferMessage(char *msg, uint32_t msgLen, uint16_
 }
 
 void BCStateTran::handoff(char *msg, uint32_t msgLen, uint16_t senderId) {
-  static concord::util::Handoff handoff_("handoff", config_.myReplicaId);
+  static concord::util::Handoff handoff_("state-transfer", config_.myReplicaId);
   handoff_.push(std::bind(&BCStateTran::handleStateTransferMessageImp, this, msg, msgLen, senderId));
 }
 
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -655,7 +655,7 @@ void BCStateTran::handleStateTransferMessage(char *msg, uint32_t msgLen, uint16_
 }
 
 void BCStateTran::handoff(char *msg, uint32_t msgLen, uint16_t senderId) {
-  static concord::util::Handoff handoff_(config_.myReplicaId);
+  static concord::util::Handoff handoff_("handoff", config_.myReplicaId);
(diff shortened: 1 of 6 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -656,7 +656,7 @@ void BCStateTran::handleStateTransferMessage(char *msg, uint32_t msgLen, uint16_
 }
 
 void BCStateTran::handoff(char *msg, uint32_t msgLen, uint16_t senderId) {
-  static concord::util::Handoff handoff_(config_.myReplicaId);
+  static concord::util::Handoff handoff_("handoff", config_.myReplicaId);
```
toly-kournik, line 659: Let's call it state-transfer

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -266,6 +261,25 @@ RawBlock DBAdapter::getRawBlock(const BlockId &blockId) const {
...
+    if (parentBlock) {
+      parentBlockDigest = computeBlockDigest(blockId - 1, parentBlock->data(), parentBlock->length());
+    }
+    return parentBlockDigest;
+  });
+
+  auto parentBlockDigestFuture = parentBlockDigestTask->get_future();
+
+  // Calculate the parent block digest while updating the tree.
+  asyncExecutor_.push([parentBlockDigestTask]() { (*parentBlockDigestTask)(); });
```
toly-kournik, line 281: Just a question, why do we capture parentBlockDigestTask?

bftengine/src/bcstatetransfer/BCStateTran.cpp
dartdart26: Great, will change it.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: I've done it in such a way, because the function passed to `Handoff` is a `void ()` one and we want to call the task inside of it. Please let me know if there's a way to simplify that.

dartdart26:
WIP until issues reported by TSAN are fully understood.

dartdart26:
Will implement using `std::async` in another PR. Closing that one.

PR #630, opened 2020-05-28
Reject client requests in case the system is not ready

yuliasherman:
When the system starts up the replicas become available within a few minutes and not all together, which causes the system to start a chain of view-changes. To avoid this behavior, SimpleCLient should verify that there are enough live connections to the replicas before starting request handling.

https://jira.eng.vmware.com/browse/BC-2827

### DIFF
--- a/bftengine/src/bftengine/SimpleClientImp.cpp
+++ b/bftengine/src/bftengine/SimpleClientImp.cpp
@@ -38,23 +38,20 @@ class SimpleClientImp : public SimpleClient, public IReceiver {
       ICommunication* communication, uint16_t clientId, uint16_t fVal, uint16_t cVal, SimpleClientParams& p);
   ~SimpleClientImp() override;
 
-  int sendRequest(uint8_t flags,
-                  const char* request,
-                  uint32_t lengthOfRequest,
-                  uint64_t reqSeqNum,
-                  uint64_t timeoutMilli,
-                  uint32_t lengthOfReplyBuffer,
-                  char* replyBuffer,
-                  uint32_t& actualReplyLength,
-                  const std::string& cid = "",
-                  const std::string& span_context = "") override;
-
-  int sendRequestToResetSeqNum() override;
-  int sendRequestToReadLatestSeqNum(uint64_t timeoutMilli, uint64_t& outLatestReqSeqNum) override;
+  OperationResult sendRequest(uint8_t flags,
+                              const char* request,
+                              uint32_t lengthOfRequest,
+                              uint64_t reqSeqNum,
+                              uint64_t timeoutMilli,
+                              uint32_t lengthOfReplyBuffer,
+                              char* replyBuffer,
+                              uint32_t& actualReplyLength,
+                              const std::string& cid,
+                              const std::string& span_context) override;
 
   // IReceiver methods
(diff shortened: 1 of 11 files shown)


teoparvanov:
Hi Yulia, I suppose with this fix you could also remove the 3 "sleeps" for 5 seconds in the pre-execution tests? For example:
https://github.com/vmware/concord-bft/blob/master/tests/apollo/test_skvbc_preexecution.py#L154

yuliasherman:
> Hi Yulia, I suppose with this fix you could also remove the 3 "sleeps" for 5 seconds in the pre-execution tests? For example:
> https://github.com/vmware/concord-bft/blob/master/tests/apollo/test_skvbc_preexecution.py#L154

Actually, this change is not related to the issue I have with a thread pool, which still requires a fix.

teoparvanov:
Hmm, interesting... So your goal is to avoid the initial view changes if the replicas aren't up and running yet? If so, the following test I recently introduced should fail, right?

https://github.com/vmware/concord-bft/blob/master/tests/apollo/test_skvbc.py#L220

yuliasherman:
You don't expect to have view changes in this situation. But all those tests are using UDP, which is connectionless, while my fix makes sense only for TCP/TLS.

teoparvanov:
> 
> 
> You don't expect to have view changes in this situation. But all those tests are using UDP, which is connectionless, while my fix makes sense only for TCP/TLS.

You are right, thanks for the clarification!

bftengine/src/bftengine/SimpleClientImp.cpp
```diff
@@ -108,6 +111,21 @@ class SimpleClientImp : public SimpleClient, public IReceiver {
   concordlogger::Logger logger_ = concordlogger::Log::getLogger("concord.bft.client");
 };
 
+bool SimpleClientImp::isSystemReady() const {
+  uint16_t connectedReplicasNum = 0;
+  if (communication_->isRunning()) {
+    for (uint16_t rid : replicas_) {
+      if (communication_->getCurrentConnectionStatus(rid) == ConnectionStatus::Connected) connectedReplicasNum++;
```
andrewjstone, line 118: Unfortunately, I don't think this works. The implementation of getCurrentConnectionStatus in TLS is [incorrect](https://github.com/vmware/concord-bft/blob/master/communication/src/TlsTCPCommunication.cpp#L1287). I discovered this when working on a refactoring a couple months ago. Unfortunately, I also did not have time to finish that (it's about 95%), so the fix is not in. 

You may be able to take my fix although you may need to modify it to fit the exact names in the current code base: 

```C++
ConnectionStatus TlsTCPCommunication::TlsTcpImpl::getCurrentConnectionStatus(const NodeNum id) const {
  std::lock_guard<std::mutex> lock(connectionsGuard_);
  if (connections_.count(id) == 1) {
    return ConnectionStatus::Connected;
  }
  return ConnectionStatus::Disconnected;
}
```

yuliasherman: Thanks for the observation, Andrew. I've also noticed that this functionality is not implemented, that's why prefixed this PR with 'WIP'.

PR #631, opened 2020-05-28
Use binary assert macros in BCStateTran

andrewjstone:
Where possible, all asserts were changed to use the more specific binary macros
rather than just Assert in BCStateTran. This allows values to be printed
when they implement `std::ostream&<<`.

Additionally, `std::ostream&<<` was implemented for STDigest and
FetchingState types.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -23,7 +23,6 @@
 #include "BCStateTran.hpp"
 #include "STDigest.hpp"
 #include "InMemoryDataStore.hpp"
-#include "assertUtils.hpp"
 
 #include "DBDataStore.hpp"
 #include "storage/db_interface.h"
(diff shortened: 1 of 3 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -308,10 +307,11 @@ void BCStateTran::init(uint64_t maxNumOfRequiredStoredCheckpoints,
     } else {
       LOG_INFO(STLogger, "initializing a new object");
 
-      AssertAND(maxNumOfRequiredStoredCheckpoints >= 2,
-                maxNumOfRequiredStoredCheckpoints <= kMaxNumOfStoredCheckpoints);
+      AssertGE(maxNumOfRequiredStoredCheckpoints, 2);
+      AssertLE(maxNumOfRequiredStoredCheckpoints, kMaxNumOfStoredCheckpoints);
+      AssertGE(numberOfRequiredReservedPages, 2);
+      AssertLE(numberOfRequiredReservedPages, config_.maxNumOfReservedPages);
```
teoparvanov, line 313: this assert didn't exist previously, right?

andrewjstone: It did. There's an `AssertAND` that was replaced above and below these 4 asserts. 2 Asserts for each AssertAnd.

teoparvanov: Yep, you're right, thanks for the clarification!

PR #636, opened 2020-06-01
Logging improvements:

toly-kournik:
1. General
	- Rename namespace: concordlogging -> logging;
	- Both logger implementationds use the same configuration file - logging.properties.
	  Each one uses its own prefix: log4cplus. and log. correspondingly.
	- Several source files took the `<cassert>` include from Logging.hpp.
	  Removed the include and added `#include "assertUtils.hpp"` and corresponding macros.
2. Internal logger implementation:
	- Support loglevel configuration per logging facility;
	- Added configuration file support;
	- Optimized printing macros and logging itself;


toly-kournik:
**Notes for code review:**

most of the files have trivial changes: namespace: concordlogging -> logging;

The important files are under **logging** folder

cloudnoize:
Looks great Toly

toly-kournik:
> Looks great Toly.
> 
> We may want to use a less generic namespace though, like concord::logging, since this is a library embedded into applications.

I had some hesitations about it. At the end I decided that since it's used in bftengine it can be more generic than concord.

andrewjstone:
> > Looks great Toly.
> > We may want to use a less generic namespace though, like concord::logging, since this is a library embedded into applications.
> 
> I had some hesitations about it. At the end I decided that since it's used in bftengine it can be more generic than concord.

I think it's fine for now. I just worry about conflicts with other libraries linked into an application. We may have to change it in the future depending upon users of concord-bft and their dependencies.

PR #639, opened 2020-06-03
Fix for post-view change replica restart

teoparvanov:
## Overview
Certain replica start-up scenarios, such as starting replicas with delays in-between, results in one or more view changes. If a replica is restarted during/after this process, it is unable to load its state from persistent storage.

When a replica initiates view change by calling `MoveToHigherView()`, it persists a `DescriptorOfLastExitFromView` in its metadata store. Later on, if and when a new view is agreed, it persists a `DescriptorOfLastNewView`. If the replica is restarted, it uses those descriptors to load its view.

The problem is, as initially there are insufficient replicas for consensus, the `MoveToHigherView()` happens when the initial replicas haven't processed any requests. This is why, the persisted `DescriptorOfLastExitFromView` is exactly equal to the default one. After restarting the replica, this messes up the logic that calculates `hasDescLastExitFromView`, and the `ReplicaLoader::loadViewInfo()` method returns `InconsistentErr`. This results in the replica being unable to boot (becomes unrecoverable). See the final `else` clause below:

```
if (!hasDescLastExitFromView && !hasDescOfLastNewView) {
  ...
} else if (hasDescLastExitFromView && !hasDescOfLastNewView) {
  ...
} else if (hasDescLastExitFromView && hasDescOfLastNewView &&
		 (descriptorOfLastExitFromView.view == descriptorOfLastNewView.view)) {
  ...
} else if (hasDescLastExitFromView && hasDescOfLastNewView &&
  ...
} else {
  LOG_ERROR(GL,
		  "Failed to load view (inconsistent state): " << KVLOG(
			  replicaId, hasDescLastExitFromView, hasDescOfLastNewView, lastExitedView, lastNewView));
  return InconsistentErr;
}
```

## Suggested fix
In the persistent store, we need to be able to distinguish between a default `DescriptorOfLastExitFromView`, and one which has the same content, but was persisted as a result of a view change. This is why, in this PR I suggest introducing a `isDefault` boolean field which is also persisted, and allows the replica to correctly determine whether the descriptor is default.

Note:The same change is applied to `DescriptorOfLastNewView`, mostly for the sake of symmetry, as a similar problem could occur there, although I haven't actually observed it.

Additionally, I've updated the "delayed replica start-up" Apollo test, to restart a random non-primary after the view change has occurred and make sure it continues participating in consensus within the new view.

### DIFF
--- a/tests/apollo/test_skvbc.py
+++ b/tests/apollo/test_skvbc.py
@@ -267,6 +267,30 @@ class SkvbcTest(unittest.TestCase):
                     self.assertTrue(last_block > current_block)
                     current_block = last_block
 
+            current_primary = current_view % bft_network.config.n
+            non_primaries = bft_network.all_replicas(without={current_primary})
+            restarted_replica = random.choice(non_primaries)
+            print(f"Restart replica #{restarted_replica} to make sure its view is persistent...")
+
+            bft_network.stop_replica(restarted_replica)
+            await trio.sleep(seconds=5)
+            bft_network.start_replica(restarted_replica)
+
+            await trio.sleep(seconds=20)
+
+            # Stop sending requests, and make sure the restarted replica
+            # is up-and-running and participates in consensus
+            nursery.cancel_scope.cancel()
+
+            key = ['replica', 'Gauges', 'lastExecutedSeqNum']
+            last_executed_seq_num = await bft_network.metrics.get(current_primary, *key)
+
+            with trio.fail_after(seconds=30):
+                while True:
+                    last_executed_seq_num_restarted = await bft_network.metrics.get(restarted_replica, *key)
+                    if last_executed_seq_num_restarted >= last_executed_seq_num:
+                        break
+
         except Exception as e:
             print(f"Delayed replicas start-up failed for start-up order: {replicas_starting_order}")
             raise e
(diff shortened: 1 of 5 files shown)


tests/apollo/test_skvbc.py
```diff
@@ -267,6 +267,30 @@ async def test_delayed_replicas_start_up(self, bft_network, skvbc, nursery):
...
+            # The view is stable, stop sending requests
+            nursery.cancel_scope.cancel()
+
+            key = ['replica', 'Gauges', 'lastExecutedSeqNum']
+            last_executed_seq_num = await bft_network.metrics.get(current_primary, *key)
+
+            with trio.fail_after(seconds=30):
+                while True:
+                    last_executed_seq_num_restarted = await bft_network.metrics.get(restarted_replica, *key)
+                    if last_executed_seq_num_restarted >= last_executed_seq_num:
```
andrewjstone, line 290: Why would this almost not always be equivalent, considering you stop the nursery before sending the get request for the metric. I see there is a race, but it seems like more likely than not progress might have already stopped by the time we get here.

In other words, what's the purpose of this check?

teoparvanov: Hi Andrew, the idea is to check that the restarted replica is in the new view after the restart, and executing requests.

PR #640, opened 2020-06-03
[WIP] Wait for the system to be able to make progress.

HristoStaykov:
Initial implementation of waiting for the system to be
able to make progress (execute client requests) after view change.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -430,8 +430,10 @@ class BftTestNetwork:
 
         In case of a timeout, fails with the provided err_msg
         """
+        wait_for_liveness_after_view_change = True
         if expected is None:
             expected = lambda _: True
+            wait_for_liveness_after_view_change = False
 
         matching_view = None
         nb_replicas_in_matching_view = 0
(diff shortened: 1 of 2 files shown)


tests/apollo/util/bft.py
```diff
@@ -430,8 +434,10 @@ async def wait_for_view(self, replica_id, expected=None,
 
         In case of a timeout, fails with the provided err_msg
         """
+        wait_for_liveness_after_view_change = True
```
teoparvanov, line 437: Why do we need this boolean? We always want to `wait_for_liveness` after a view change has completed...

tests/apollo/util/skvbc.py
```diff
@@ -330,6 +331,20 @@ async def read_your_writes(self, test_class):
 
         print(f'[READ-YOUR-WRITES] OK.')
 
+    async def wait_for_system_to_be_live(self, time_to_wait=60):
+        """Wait for system to be able to reach quorums and execute clients' requests"""
+        with trio.fail_after(seconds=time_to_wait):
+            while True:
+                with trio.move_on_after(seconds=5):
+                    try:
+                        k, v = await self.write_known_kv()
```
teoparvanov, line 340: Unfortunately, I believe this approach might not be very reliable... According to my findings here - https://jira.eng.vmware.com/browse/BC-3028, sending requests while the active window is not processed by consensus yet, might trigger additional view changes!

PR #644, opened 2020-06-04
Fix memory leak

f-squirrel:
In the function `IReplica::createNewReplica` in case
`ReplicaLoader::loadReplica` fails the `replicaInternal` is not
released.


f-squirrel:
This is the current fix.
The general approach of using smart pointers in the interfaces will be reviewed during refactoring.

PR #646, opened 2020-06-05
Fix hexPrint()'s output when input is negative

dartdart26:
Convert from signed char to std::uint8_t and then to an unsigned
non-char type so that it prints as an integer.

Previous code was converting from char to uint directly, leading to big
integer values and unexpected hex output.


andrewjstone:
Merging, as it seems good to get in ASAP since we print digests in some cases using this function.

PR #647, opened 2020-06-05
Make InternalMessage handling more explicit

andrewjstone:
This commit attempts to limit action at a distance and make the code
more linear by explicitly showing the internal message handling in a
manner similar to how we handle external messages in ReplicaImp.

Previously, each internal message was a subtype that overrode a `handle`
method that would operate on the `ReplicaImp` via the
`InternalReplicaAPI`. This coupled the message itself with the handling
of the message and meant that the message must know what the handler
wanted to do with it. Indeed an `InternalReplicaAPI*` was included in
every internal message and often required type casting to use.
Furthermore, this required creating a duplicate function for each
handler in `InternalReplicaAPI` as well as ReplicaImp. It was not always
clear that these methods corresponded to internal messages either, since
the names did not indicate this.

Additionally, the previous implementation required heap allocation, even
though the internal messages themselves are small, due to the
inheritance from `InternalMessage`. This PR instead puts all internal
messages in a `std::variant` such that we can see them all in one place
and treat them as a "closed" set. This also allows us to allocate on the
heap when necessary, and on the stack when messages are small.

All messages can now be put directly on the message queue:
`IncomingMsgsStorage`, and handled in `ReplicaImp::onInternalMsg`,
making it easy to see where threads are involved in processing.  This is
the beginning of a strategy to make message passing and dispatching more
explicit and allow easier use of messaging and threads as needed.

This commit is also a pre-requisite for adding new diagnostics related
InternalMessages.

Additionally, this PR contains other cleanup:

 * Removes unused execution proof related messages, tests, and handlers.
 These are from the original research code and we have no intent to use
 them. The client verifiability work mostly supercedes these messages
 for the short and medium term. We can add more compatible ones later if
 we decide to implement this phase of the protocol. The reason this
 was done in this commit was to prevent the need to adapt to the new
 internal message structure for unused code.

 * Deletes move constructor and move assigment from the diagnostics
 `RegistrarSingleton`, which is intended to be used in the a follow up
 commit.


bftengine/src/bftengine/messages/InternalMessage.hpp
```diff
@@ -11,18 +11,44 @@
...
 
+#include <variant>
+#include "messages/FullCommitProofMsg.hpp"
+#include "messages/RetranProcResultInternalMsg.hpp"
+#include "SignatureInternalMsgs.hpp"
+
 namespace bftEngine::impl {
 
-class InternalReplicaApi;
+struct GetStatus {
```
robem, line 21: This seemed to have slipped in prematurely or was this copied from somewhere?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -860,6 +858,47 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
...
+  metric_received_internal_msgs_.Get().Inc();
+
+  // Handle a full commit proof sent by self
+  if (auto *fcp = std::get_if<FullCommitProofMsg *>(&msg)) {
+    return onInternalMsg(*fcp);
+  }
+
+  // Handle prepare related internal messages
+  if (auto *csf = std::get_if<CombinedSigFailedInternalMsg>(&msg)) {
+    return onPrepareCombinedSigFailed(csf->seqNumber, csf->view, csf->replicasWithBadSigs);
```
robem, line 871: All those handlers' signatures effectively want their respective message but instead each handler takes all members of the message struct. That looks like another opportunity to refactor later.

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -430,29 +301,24 @@ PrepareFullMsg* SeqNumInfo::ExFuncForPrepareCollector::createCombinedSignatureMs
...
-InternalMessage* SeqNumInfo::ExFuncForPrepareCollector::createInterVerifyCombinedSigResult(void* context,
-                                                                                           SeqNum seqNumber,
-                                                                                           ViewNum viewNumber,
-                                                                                           bool isValid) {
-  InternalReplicaApi* r = (InternalReplicaApi*)context;
-  return new VerifyCombinedSigResultInternalMsg(r, seqNumber, viewNumber, isValid);
+InternalMessage SeqNumInfo::ExFuncForPrepareCollector::createInterVerifyCombinedSigResult(SeqNum seqNumber,
+                                                                                          ViewNum viewNumber,
+                                                                                          bool isValid) {
+  return VerifyCombinedSigResultInternalMsg(seqNumber, viewNumber, isValid);
```
robem, line 321: Why do we have those `create*` functions in the first place? Maybe another potential refactoring task for later?

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: Yes indeed. I thought about doing that, but kept this particular change a bit smaller since it was already getting quite large. They should probably all become `onInternalMessage(SpecificMessage)`

bftengine/src/bftengine/SeqNumInfo.cpp
andrewjstone: I had the same thought at first, and actually started making that change, but it doesn't work unfortunately. Although, this possibly can be made simpler with another refactoring. The reason we can't do this is that [CollectorOfThresholdSignatures](https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp#L31) uses these wrappers for both prepare and commit messages which are different.

bftengine/src/bftengine/messages/InternalMessage.hpp
andrewjstone: Ah good point. Yes, this should only come in the follow up PR. i'll remove it.

PR #650, opened 2020-06-07
Fix span/cid mismatch in a function call (preprocessor_test)

yuliasherman:
- Fix span/cid order while creating PreProcessRequestMsg
- Add some clarifications to DummyReplica/DummyReceiver functons

### DIFF
--- a/bftengine/src/preprocessor/tests/preprocessor_test.cpp
+++ b/bftengine/src/preprocessor/tests/preprocessor_test.cpp
@@ -28,7 +28,6 @@ using namespace preprocessor;
 
 namespace {
 
-ReplicaConfig replicaConfig;
 const uint16_t numOfReplicas_4 = 4;
 const uint16_t numOfReplicas_7 = 7;
 const uint16_t numOfClients = 4;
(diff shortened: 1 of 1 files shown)


bftengine/src/preprocessor/tests/preprocessor_test.cpp
```diff
@@ -85,35 +86,52 @@ DummyRequestsHandler requestsHandler;
...
-
-  const ReplicaConfig& getReplicaConfig() const { return replicaConfig; }
-
-  IThresholdVerifier* getThresholdVerifierForExecution() { return nullptr; }
-  IThresholdVerifier* getThresholdVerifierForSlowPathCommit() { return nullptr; }
-  IThresholdVerifier* getThresholdVerifierForCommit() { return nullptr; }
-  IThresholdVerifier* getThresholdVerifierForOptimisticCommit() { return nullptr; }
+  explicit DummyReplica(bftEngine::impl::ReplicasInfo replicasInfo) : replicasInfo_(move(replicasInfo)) {}
+
+  void onPrepareCombinedSigFailed(SeqNum seqNumber, ViewNum view, const set<uint16_t>& replicasWithBadSigs) {}
```
andrewjstone, line 91: There's no need to add these empty functions back in. They don't do anything, and are no longer part of` InternalRepliaAPI`, so you no longer have to implement them anytime you want to try to create a usable test replica.

andrewjstone: Unless I'm totally missing something, which is also possible :)

PR #651, opened 2020-06-07
BFT Client pool:Add dynamic libraries

Liorgal28:
In order to create a shared library for BFT client pool, I had to create shared libraries to the linked libraries as well.

In this PR I created 3 dynamic libraries for:
1)bft-communication
2)util
3)bft-client


communication/CMakeLists.txt
```diff
@@ -13,22 +13,32 @@ if(${BUILD_COMM_TCP_TLS})
...
 add_library(bftcommunication ${bftcommunication_src})
+add_library(bftcommunication_shared SHARED ${bftcommunication_src})
 target_include_directories(bftcommunication PUBLIC include)
 target_link_libraries(bftcommunication PUBLIC util)
+target_include_directories(bftcommunication_shared PUBLIC include)
+target_link_libraries(bftcommunication_shared PUBLIC util_shared)
 
 if(${BUILD_COMM_TCP_PLAIN} OR ${BUILD_COMM_TCP_TLS})
-    set(Boost_USE_STATIC_LIBS ON) # only find static libs
+    set(Boost_USE_STATIC_LIBS OFF) # only find static libs
```
yuliasherman, line 23: Change the comment to 'shared lib'

util/CMakeLists.txt
```diff
@@ -11,24 +11,41 @@ add_library(util STATIC
     src/hex_tools.cpp
     src/OpenTracing.cpp)
 
+add_library(util_shared SHARED
```
yuliasherman, line 14: Define a list of files to be compiled once and then use them in both static/shared add_library(...)

Liorgal28: Done

communication/CMakeLists.txt
Liorgal28: Changed the comment to both things, because it's also not just shared

PR #653, opened 2020-06-08
Additional view change edge scenarios

teoparvanov:
This PR introduces the following view change "edge" scenarios:
1) Make sure a replica that was down during view change, catches up and works in the new view when restarted
2) Make sure a replica successfully restarts after view change, and works in the new view

Additionally, this PR introduces certain log INFO level improvements, which have helped in the manual log analysis part while running the above test cases.

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -292,11 +292,16 @@ void ClientsManager::clearAllPendingRequests() {
 Time ClientsManager::timeOfEarliestPendingRequest() const  // TODO(GG): naive implementation - consider to optimize
 {
   Time t = MaxTime;
+  ClientInfo earliestClientWithPendingRequest = indexToClientInfo_.at(0);
 
   for (const ClientInfo& c : indexToClientInfo_) {
-    if (c.timeOfCurrentPendingRequest != MinTime && t > c.timeOfCurrentPendingRequest)
+    if (c.timeOfCurrentPendingRequest != MinTime && t > c.timeOfCurrentPendingRequest) {
       t = c.timeOfCurrentPendingRequest;
+      earliestClientWithPendingRequest = c;
+    }
   }
+
+  LOG_INFO(GL, "Earliest pending client request: " << KVLOG(earliestClientWithPendingRequest.currentPendingRequest));
 
   return t;
 }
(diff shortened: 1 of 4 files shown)


bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -18,6 +18,9 @@
 #include "Logger.hpp"
 #include "ReplicaConfig.hpp"
 
+using std::chrono::duration_cast;
+using std::chrono::milliseconds;
```
teoparvanov, line 22: note to myself - those are actually not used -> to be removed

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1980,10 +1980,19 @@ void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNew
     primaryLastUsedSeqNum = lastStableSeqNum;
     strictLowerBoundOfSeqNums = lastStableSeqNum;
     maxSeqNumTransferredFromPrevViews = lastStableSeqNum;
+    LOG_INFO(GL,
+             "No PrePrepare-s for the new view: " << KVLOG(
+                 primaryLastUsedSeqNum, strictLowerBoundOfSeqNums, maxSeqNumTransferredFromPrevViews));
```
HristoStaykov, line 1985: Might be we should add "***" prefix, to be consistent with the other View Change associated logs.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1980,10 +1980,19 @@ void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNew
...
   } else {
     primaryLastUsedSeqNum = lastPPSeq;
     strictLowerBoundOfSeqNums = firstPPSeq - 1;
     maxSeqNumTransferredFromPrevViews = lastPPSeq;
+    LOG_INFO(GL,
+             "There are PrePrepare-s for the new view: " << KVLOG(firstPPSeq,
+                                                                  lastPPSeq,
+                                                                  primaryLastUsedSeqNum,
+                                                                  strictLowerBoundOfSeqNums,
+                                                                  maxSeqNumTransferredFromPrevViews));
```
HristoStaykov, line 1995: Same thing here, I think the sync of the active window after View Change is good to be consistent with the convention of View Change logs.

tests/apollo/test_skvbc_view_change.py
```diff
@@ -211,6 +211,120 @@ async def test_single_vc_with_f_replicas_down(self, bft_network, tracker):
...
+        await bft_network.wait_for_view(
+            replica_id=random.choice(
+                bft_network.all_replicas(without={initial_primary})),
+            expected=lambda v: v == initial_primary + 1,
+            err_msg="Make sure a view change is triggered."
+        )
+        current_primary = initial_primary + 1
+
+        bft_network.start_replica(initial_primary)
+
```
HristoStaykov, line 308: Might be it would be good to also verify that after restarting the initial Primary, it is going to get in the new View.

teoparvanov: Thanks @HristoStaykov , makes sense!

PR #656, opened 2020-06-08
[WIP] Reduce Apollo test replicas

teoparvanov:
I suspect the CI failures we see in GitHub Actions are because of the lack of CPU cores. Running 7 replicas on 2 CPU cores may have "interesting" side effects.

My suggestion is to try reducing the number of replicas.


teoparvanov:
Actually some tests do require f >= 2, so n >= 7. We have to think of something else...

PR #657, opened 2020-06-08
[Apollo] Increase and unify view change timeout across test cases

teoparvanov:
In certain cases where the system is under load, or the replicas are starving for compute resources (e.g. CI), undesirable view changes may be triggered, compromising the liveness of the system under test, and altering test results.

To mitigate this risk, I'm increasing the view change timeout from 10 to 20 seconds where applicable. This, by the way, is the value currently used in the product.


teoparvanov:
After discussions with Guy and Ittai, I'm closing this for now, as there are likely better avenues...

PR #659, opened 2020-06-09
[EXPERIMENTAL] Reduce the active window size

teoparvanov:
This PR is an experimental fix for https://jira.eng.vmware.com/browse/BC-3028.

As discussed with Ittai and Guy, making a first attempt to reduce the window size from 300 to 30. As a consequence, checkpoints will also happen 10 times more frequently. 

Still, practically speaking if a replica is 30 sequence numbers behind, it would also likely lag even further (probably because it has crashed or there is network partitioning). So, despite the active window advancing 10x faster than before, we don't expect state transfer to happen much more frequently than it did previously.

Also, by doing this we expect to significantly (10x) speed-up view change, especially in the part where it rebuilds the active window.

### DIFF
--- a/tests/apollo/util/skvbc_history_tracker.py
+++ b/tests/apollo/util/skvbc_history_tracker.py
@@ -949,7 +949,7 @@ class SkvbcTracker:
 
     async def tracked_prime_for_state_transfer(
             self, stale_nodes,
-            checkpoints_num=2,
+            checkpoints_num=20,
             persistency_enabled=True):
         initial_nodes = self.bft_network.all_replicas(without=stale_nodes)
         [self.bft_network.start_replica(i) for i in initial_nodes]
@@ -964,7 +964,7 @@ class SkvbcTracker:
         # there.
         client1 = self.bft_network.random_client()
         # Write enough data to checkpoint and create a need for state transfer
-        for i in range(1 + checkpoints_num * 150):
+        for i in range(1 + checkpoints_num * 15):
             key = self.skvbc.random_key()
             val = self.skvbc.random_value()
             kv = [(key, val)]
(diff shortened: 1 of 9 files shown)


teoparvanov:
PS: interestingly, all GitHub Actions jobs pass successfully with this setup :)

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -949,7 +949,7 @@ async def read_and_track_known_kv(self, key, client):
 
     async def tracked_prime_for_state_transfer(
             self, stale_nodes,
-            checkpoints_num=2,
+            checkpoints_num=20,
```
andrewjstone, line 952: Why change this to 20?

teoparvanov: Well, essentially, to make the state transfer test non-trivial. With a window of 30, we would generate a checkpoint for every 15 sequence numbers. So with 2 checkpoints, we would only have 30 blocks to transfer, which happens too fast for our test suites to be able to properly detect and inject failures.

salieri11:
LGTM. Its really interesting to see performance impact. However, all our performance tests didnt involve VC, so I suppose we will not feel it.
It will imply updating metadata for the stable checkpoint 10x times more frequently, but I think that this data is so minor that it will not be noticeable.
Agree with @andrewjstone that we need to run Maestro test before bumping Concord

teoparvanov:
The tests are stable for n=7 (f >= 2), which is our top priority.
 
On the other hand, for n=6 and n=4 I I still observe unexpected view changes, likely because of insufficient quorums. This could be explained with the fact that at different points during the test execution, we restart the primary AND another random non-primary. My guess is, there are some intermediary states where we effectively end up with more than 1 "byzantine" replica.

teoparvanov:
> 
> 
> Occasionally, I see the following error log:
> `Failed to satisfy full proof requirements`
> 

This misleading error message is being fixed here - https://github.com/vmware/concord-bft/pull/672

PR #661, opened 2020-06-09
BFT client pool: external clients parameter to replica configuration

Liorgal28:
In this PR I have added numOfExternalClients parameter to the replicaConfiguration in order to reserve pages to the external client in addition to the internal clients as well.


bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -47,6 +47,11 @@ struct ReplicaConfig {
   // numOfClientProxies >= 1
   uint16_t numOfClientProxies = 0;
 
+  // number of objects that represent external clients.
+  // numOfExternalClients >= 0
+  // By default numOfExternalClients will be 0 unless we configure it will be greater than.
```
yuliasherman, line 52: -> By default, numOfExternalClients is 0 unless configured differently.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2917,7 +2917,9 @@ ReplicaImp::ReplicaImp(bool firstTime,
   }
 
   std::set<NodeIdType> clientsSet;
-  for (uint16_t i = config_.numReplicas; i < config_.numReplicas + config_.numOfClientProxies; i++)
+  for (uint16_t i = config_.numReplicas;
+       i < config_.numReplicas + config_.numOfClientProxies + config_.numOfExternalClients;
```
yuliasherman, line 2921: Don't re-calculate the loop end condition at every iteration: define a const variable before a loop and use it inside: 
const auto numOfEntities = config_.numReplicas + config_.numOfClientProxies + config_.numOfExternalClients;
for (uint16_t i = config_.numReplicas; I < numOfEntities....

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -95,7 +95,8 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
       maxReplyMsgSize_(myReplica.getReplicaConfig().maxReplyMessageSize - sizeof(ClientReplyMsgHeader)),
       idsOfPeerReplicas_(myReplica.getIdsOfPeerReplicas()),
       numOfReplicas_(myReplica.getReplicaConfig().numReplicas),
-      numOfClients_(myReplica.getReplicaConfig().numOfClientProxies),
+      numOfClients_(myReplica.getReplicaConfig().numOfClientProxies +
+                    myReplica.getReplicaConfig().numOfExternalClients),
```
yuliasherman, line 99: Preprocessor works only with external clients.

Liorgal28: so, remove numOfClientProxies? because before that he took only numOfClientProxies.

yuliasherman: Yes. Previously this number was expected to include all possible clients.

bftengine/include/bftengine/ReplicaConfig.hpp
Liorgal28: Fixed.

bftengine/src/preprocessor/PreProcessor.cpp
Liorgal28: For now, we keep using the numOfClientProxies parameter in order not to break the Preprocessor tests.

PR #663, opened 2020-06-09
Use the numOfExternalClients configuration parameter instead of numOfClientProxies

yuliasherman:
- A new parameter numOfExternalClients has been introduced by the BFT Client pool feature and it should be used by the internal Preprocessor configuration instead of numOfClientProxies.
- Configure a number of Preprocessor threads equal to the number of VM CPUs.

https://jira.eng.vmware.com/browse/BC-3095

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -36,6 +36,11 @@ void PreProcessor::addNewPreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunic
                                       bftEngine::IRequestsHandler &requestsHandler,
                                       InternalReplicaApi &replica,
                                       concordUtil::Timers &timers) {
+  if (!ReplicaConfigSingleton::GetInstance().GetNumOfExternalClients()) {
+    LOG_ERROR(logger(), "A number of configured external clients should not be zero!");
+    return;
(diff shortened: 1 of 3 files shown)


tests/apollo/test_skvbc_preexecution.py
```diff
@@ -90,7 +90,6 @@ async def run_concurrent_pre_execution_requests(self, skvbc, clients, num_of_req
             await trio.sleep(.1)
         return read_count + write_count
 
-    @unittest.skip("unstable")
     @with_trio
     @with_bft_network(start_replica_cmd)
```
teoparvanov, line 94: Let's try running those tests with only 4 replicas, for the purposes of CI.

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -104,7 +103,6 @@ async def test_sequential_pre_process_requests(self, bft_network):
             client = bft_network.random_client()
             await self.send_single_write_with_pre_execution(skvbc, client)
 
-    @unittest.skip("unstable")
     @with_trio
     @with_bft_network(start_replica_cmd)
     async def test_concurrent_pre_process_requests(self, bft_network):
```
teoparvanov, line 108: same here (filter for `n ==4`)

PR #665, opened 2020-06-09
Chaotic startup tests.

HristoStaykov:
Moved tests related to chaotic startup of the system to separate python test.
Fixed comments from PR 641.

### DIFF
--- a/tests/apollo/test_skvbc_chaotic_startup.py
+++ b/tests/apollo/test_skvbc_chaotic_startup.py
@@ -18,11 +18,7 @@ from os import environ
 
 import trio
 
-from util import bft_network_partitioning as net
-from util import skvbc as kvbc
 from util.bft import with_trio, with_bft_network, with_constant_load, KEY_FILE_PREFIX
-from util.skvbc_history_tracker import verify_linearizability
-from trio._timeouts import TooSlowError
 
 def start_replica_cmd(builddir, replica_id):
     """
@@ -181,8 +177,8 @@ class SkvbcChaoticStartupTest(unittest.TestCase):
             active_view = 0
             view = 0
             while active_view == view:
-                active_view = await self._get_active_view(r, bft_network)
-                view = await self._get_view(r, bft_network)
+                active_view = await self._get_gauge(r, bft_network, 'currentActiveView')
+                view = await self._get_gauge(r, bft_network, 'view')
                 await trio.sleep(seconds=0.1)
 
         print("STATUS: Early replicas initiated View Change.")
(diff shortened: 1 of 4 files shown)


tests/apollo/test_skvbc_chaotic_startup.py
```diff
@@ -0,0 +1,277 @@
...
+        print(f"STATUS: Starting F={f} replicas.")
+        bft_network.start_replicas(replicas=early_replicas)
+
+        print("STATUS: Wait for early replicas to initiate View Change.")
+        for r in early_replicas:
+            active_view = 0
+            view = 0
+            while active_view == view:
+                active_view = await self._get_active_view(r, bft_network)
+                view = await self._get_view(r, bft_network)
```
andrewjstone, line 185: There's no actual timeout here that can cause the test to fail. I'd suggest not masking the TooSlowError, and possibly using a `fail_after` rather than `move_on_after` in the `get_view` call. Feel free to increase the timeout if you feel it's necessary. Otherwise, the test will take whatever the max test timeout of ctest or the CI system is to fail. Something like 1500 seconds.

tests/apollo/test_skvbc_chaotic_startup.py
```diff
@@ -0,0 +1,277 @@
...
+        await skvbc.fill_and_wait_for_checkpoint(
+            initial_nodes=bft_network.all_replicas(without={replica_to_stop}),
+            checkpoint_num=1,
+            verify_checkpoint_persistency=False,
+            assert_state_transfer_not_started=False
+        )
+        checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=expected_next_primary)
+
+        # Verify that the system is making progress after the View Change.
+        self.assertTrue(checkpoint_after > checkpoint_before)
```
andrewjstone, line 249: I would add one extra check that verifies that every single replica is in view 1, indicating that everyone is caught up together, and another single replica isn't lagging behind. We don't expect this ever to happen, but may as well check.

tests/apollo/test_skvbc_chaotic_startup.py
```diff
@@ -0,0 +1,271 @@
...
+        late_replicas = bft_network.all_replicas(without=early_replicas)
+
+        print(f"STATUS: Starting F={f} replicas.")
+        bft_network.start_replicas(replicas=early_replicas)
+
+        print("STATUS: Wait for early replicas to initiate View Change.")
+        for r in early_replicas:
+            active_view_of_replica = 0
+            view_of_replica = 0
+            while active_view_of_replica == view_of_replica:
```
yontyon, line 185: If the metric client is timedout, then we may end up with active_view_of_replica=None and view_of_replica != None

PR #670, opened 2020-06-10
Compute value hashes in parallel for v2Merkle

dartdart26:
Spawn a separate thread to compute value hashes while updating the
sparse merkle tree. Idea is to have the current value hash ready for the
update while the separate thread is computing another one.

Implement by using std::promise/future pairs. Introduce the ValueHashes
type to encapsulate the mapping between keys, values, promises and
futures.

Assume the same order for subsequent iterations of SetOfKeyValuePairs
objects, provided that we don't mutate them in any way. That, combined
with the assumption that the tree update takes longer than value hash
calculation, should ensure that this change is actually an optimization.
Note: The documentation for std::unordered_map (SetOfKeyValuePairs)
states that read-only operations do not invalidate iterators. Therefore,
we can expect that subsequent iterations will have the same order.

Benchmark data on an i7-9750H CPU @ 2.60GHz:
Before:
```
-----------------------------------------------------------------------
Benchmark                                Time          CPU   Iterations
-----------------------------------------------------------------------
Blockchain/addBlock/128/4/1024    4456033 ns      4456167 ns        154
Blockchain/addBlock/512/32/1024   24865546 ns     24865500 ns        29
Blockchain/addBlock/2048/8/4096  159640857 ns    159627593 ns         4

```
After:
```
-----------------------------------------------------------------------
Benchmark                                Time          CPU   Iterations
-----------------------------------------------------------------------
Blockchain/addBlock/128/4/1024    4051238 ns      3924905 ns        175
Blockchain/addBlock/512/32/1024   22804525 ns     22655801 ns        32
Blockchain/addBlock/2048/8/4096  119120398 ns    118909535 ns         6
```

Remove unused <iostream> include and 'using namespace std' directive.

### DIFF
--- a/kvbc/src/sparse_merkle/tree.cpp
+++ b/kvbc/src/sparse_merkle/tree.cpp
@@ -13,13 +13,57 @@
 #include "sparse_merkle/tree.h"
 #include "sparse_merkle/walker.h"
 
-#include <iostream>
-using namespace std;
+#include <future>
+#include <iterator>
+#include <unordered_map>
 
 using namespace concordUtils;
 
 namespace concord::kvbc::sparse_merkle {
 using namespace detail;
+
+// Computes value hashes in a separate thread.
+class ValueHashes {
+ public:
+  ValueHashes(const SetOfKeyValuePairs& updates) {
+    // A vector of value and hash promise pairs. We pass it to a separate thread that calculates the value hashes and
+    // sets them in their corresponding futures. The futures are indexed by key for user convenience.
+    auto value_promises = std::vector<std::pair<Sliver, std::promise<Hash>>>{};
+    for (const auto& [key, value] : updates) {
+      auto promise = std::promise<Hash>{};
+      hashes_.emplace(key, promise.get_future());
+      value_promises.push_back(std::make_pair(value, std::move(promise)));
+    }
+
+    // Compute value hashes in a separate thread. The passed lambda is noexcept, because:
+    //  * we don't expect the hashing code to throw
+    //  * propagating the exception correctly will make the code unnecessarily complex
+    async_future_ = std::async(std::launch::async, [value_promises = std::move(value_promises)]() mutable noexcept {
+      auto hasher = Hasher{};
+      for (auto& [value, promise] : value_promises) {
+        const auto hash = hasher.hash(value.data(), value.length());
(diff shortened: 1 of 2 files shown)


kvbc/src/sparse_merkle/tree.cpp
```diff
@@ -13,14 +13,58 @@
...
+#include <iterator>
+#include <unordered_map>
 
 using namespace concordUtils;
 
 namespace concord::kvbc::sparse_merkle {
 using namespace detail;
 
+// Computes value hashes in a separate thread.
+class ValueHashes {
```
teoparvanov, line 26: AsyncValueHashes ?

dartdart26: Yep, will rename.

kvbc/src/sparse_merkle/tree.cpp
```diff
@@ -141,11 +185,17 @@ UpdateBatch Tree::update_impl(const concord::kvbc::SetOfKeyValuePairs& updates,
...
+  // should ensure that this is actually an optimization.
+  // Note: The documentation for std::unordered_map (SetOfKeyValuePairs) states that read-only operations do not
+  // invalidate iterators. Therefore, we can expect that subsequent iterations will have the same order.
+  auto value_hashes = ValueHashes(updates);
   for (auto&& [key, val] : updates) {
-    auto leaf_hash = hasher.hash(val.data(), val.length());
     LeafNode leaf_node{val};
     LeafKey leaf_key{hasher.hash(key.data(), key.length()), version};
-    LeafChild child{leaf_hash, leaf_key};
+    LeafChild child{value_hashes[key], leaf_key};
```
andrewjstone, line 198: Doesn't this block, since get is called on the future?

dartdart26: You are correct, it does. Assumption here is that value hash calculation is faster than the tree update. As discussed offline, I am closing this PR for now. Will open another PR that implements an std::async()-like thread pool as an utility. Will further measure performance with it and decide if we should calculate value hashes in parallel and, if yes, how many threads should be used.

dartdart26:
Closing this PR as per: https://github.com/vmware/concord-bft/pull/670#discussion_r442877274

PR #675, opened 2020-06-11
Logging improvements:

toly-kournik:
- Moved logging from an object to a static library;
- Removed deprecated auto_ptr, as we moved to a latest version of log4cplus;
- All submodules are implicitly linked against logging library.
- Split iternal and log4cplus implementations into separate files


logging/CMakeLists.txt
```diff
@@ -1,114 +1,62 @@
...
+
+add_library(logging STATIC src/Logger.cpp )
+target_include_directories(logging PUBLIC include/)
+set_property(TARGET logging PROPERTY POSITION_INDEPENDENT_CODE ON)
+
+if(USE_LOG4CPP)
+	message(STATUS "USE_LOG4CPP")
+	target_sources(logging PRIVATE src/Logging4cplus.cpp)
+
+	if (USE_CONAN)
```
andrewjstone, line 12: WE can probably remove all the conan stuff.

toly-kournik: Next step

logging/src/Logging4cplus.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <log4cplus/fileappender.h>
+
+using namespace log4cplus;
+
+namespace logging {
+
+const char* logPattern = "%X{rid}|%d{%m-%d-%Y %H:%M:%S.%q}|%-5p|%c|%X{thread}|%M|%m%n";
+
+void initLogger(const std::string& configFileName) {
+  std::ifstream infile(configFileName);
```
andrewjstone, line 30: You may need an Initializer here: https://sourceforge.net/p/log4cplus/wiki/CodeExamples/
However, possibly not. If tests all pass and system works, no worries :)

toly-kournik: Fixed with force initialization to avoid log4cplus warnings in tests

andrewjstone: > Fixed with force initialization to avoid log4cplus warnings in tests

What is forced initialization? You're doing too advanced stuff for me haha :)

toly-kournik: Previously I did:
`logging::Logger GL = defaultInit()`
But if in a particular test GL was not used, defultInit() hasn't happened and we got this ugly
`log4cplus:ERROR No appenders could be found for logger (thread-pool).
log4cplus:ERROR Please initialize the log4cplus system properly.
`
Now I do:
`Logger getLogger(const std::string& name) {
  static bool __logging_init__ = defaultInit();  // one time initialization
  (void)__logging_init__;
  return log4cplus::Logger::getInstance(name);
}
`
and defaultInit happens once before _any_ logger is used.

andrewjstone: Thanks for the explanation.

PR #677, opened 2020-06-11
Diagnostics: Add status handler for BcStateTran

andrewjstone:


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -654,6 +654,33 @@ void BCStateTran::handleStateTransferMessage(char *msg, uint32_t msgLen, uint16_
   messageHandler_(msg, msgLen, senderId);
 }
 
+std::string BCStateTran::getStatus() {
+  std::ostringstream oss;
+  auto state = getFetchingState();
```
toly-kournik, line 659: minor: I would use it inline without a variable

PR #683, opened 2020-06-13
WIP: Remove 7 replicas from test config

f-squirrel:
Github Actions VMs have 7GB RAM.
Tester replica memory consumption:
```
   VSZ  RSS
1530336 34700
```
7 replicas running in parallel makes the testing system unstable


teoparvanov:
We can only test so much with 2 CPUs... At some point we'll have to request a bit more powerful VMs. AFAIK, In production, each replica has 4 dedicated CPU cores!

PR #694, opened 2020-06-16
F minus one staggered replicas requesting view change

yontyon:
* closes BC-2848
* some refactoring

PR #695, opened 2020-06-17
[Apollo] Checkpoint creation and propagation with f crashed replicas

shgandhi:
This PR introduces a test case in which:

1. We bring down f replicas and then verify checkpoint creation. 
2. We bring back up those f replica and verify checkpoint propagation in the replicas brought back up.

### DIFF
--- a/tests/apollo/test_skvbc_checkpoints.py
+++ b/tests/apollo/test_skvbc_checkpoints.py
@@ -67,3 +67,84 @@ class SkvbcCheckpointTest(unittest.TestCase):
         checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=0)
 
         self.assertEqual(checkpoint_after, 1 + checkpoint_before)
+
+    @with_trio
+    @with_bft_network(start_replica_cmd)
+    async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_network):
+        """
+        Here we bring down a total of f replicas, verify checkpoint creation and
+        propagation in that scenario.
+
+        1) Given a BFT network, we make sure all nodes are up
+        2) Crash f replicas, excluding the current primary
+        3) Send sufficient number of client requests to trigger checkpoint protocol
+        4) Make sure checkpoint is created
+        4) Bring the f replicas up
+        5) Make sure checkpoint is agreed upon by the crashed replicas
+        """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        n = bft_network.config.n
+        f = bft_network.config.f
+        c = bft_network.config.c
+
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -67,3 +66,65 @@ async def test_checkpoint_creation(self, bft_network):
...
+        self.assertEqual(len(bft_network.procs), n,
+                         "Make sure all replicas are up initially.")
+        current_primary = await bft_network.get_current_primary()
+
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+
+        crashed_replicas = await self._crash_replicas_without_primary(
+            bft_network=bft_network,
+            nb_crashing=f,
+            except_replicas={current_primary}
```
andrewjstone, line 97: You shouldn't have to pass the current primary here. Given the name of the function, you should just have it exclude the current primary.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -67,3 +66,65 @@ async def test_checkpoint_creation(self, bft_network):
...
+            initial_nodes=bft_network.all_replicas(without=crashed_replicas),
+            checkpoint_num=1,
+            verify_checkpoint_persistency=False
+        )
+
+        checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+
+        self.assertEqual(checkpoint_after, 1 + checkpoint_before)
+
+    async def _crash_replicas_without_primary(
```
andrewjstone, line 115: Maybe we can move this to bft.py if it would be useful for other tests in the future. @teoparvanov what do you think?

tests/apollo/test_skvbc_checkpoints.py
shgandhi: Makes sense. I will update it.

tests/apollo/test_skvbc_checkpoints.py
teoparvanov: Actually @yontyon is extracting something similar in his PR (see `random_set_of_replicas`):
https://github.com/vmware/concord-bft/pull/694/files

I'd suggest to wait for him to merge and the refactor this (in a separate PR).

tests/apollo/test_skvbc_checkpoints.py
shgandhi: Updated it Andrew, and added another test-case.

andrewjstone: The latter test seems to be a superset of the first. For runtime purposes, maybe just use that one?

shgandhi: Yep, updated it.

PR #698, opened 2020-06-18
Fix wrong parameter name in PlainTCPCommunication

yuliasherman:
PlainTCPCommunication is not a part of a regular compilation process and not used in the tests, but it still needs to be supported.


yuliasherman:
> LGTM!

Thanks!

andrewjstone:
Thanks Yulia :)

PR #699, opened 2020-06-18
[CI] Drop DEBUG from the test matrix (both Travis & GitHub Actions)

teoparvanov:
This would significantly speed-up CI, and at the same time reduce false-positive failures (due to slower computations by the DEBUG build).

PS: for the local dev build, I suggest to keep debug as the default setting.


toly-kournik:
@teoparvanov , I'd leave at least one DEBUG job in the matrix, because it often reveals problems

andrewjstone:
> @teoparvanov , I'd leave at least one DEBUG job in the matrix, because it often reveals problems

Ooh. good idea.

teoparvanov:
> 
> 
> > @teoparvanov , I'd leave at least one DEBUG job in the matrix, because it often reveals problems
> 
> Ooh. good idea.

How do you run a single job with DEBUG, using the matrix??

PR #700, opened 2020-06-19
Implement async() functionality backed by a pool

dartdart26:
Implement an std::async()-like functionality that is backed by a thread
pool. Return std::future objects with the result to users. Support any
callable object with any return type.

Provide unit tests that also serve as examples.

Based on Mastering the C++17 STL, page 204:
https://books.google.bg/books?id=zJlGDwAAQBAJ&pg=PA205&lpg=PA205#

### DIFF
--- a/util/include/thread_pool.hpp
+++ b/util/include/thread_pool.hpp
@@ -56,6 +56,8 @@ class ThreadPool {
 
  public:
   // Executes the passed function (or any callable) in a pool thread. Returns a future to the result.
+  // Note: The returned future's destructor doesn't block if the future value hasn't been retrieved. This is in contrast
+  // to the futures returned by std::async that block.
   template <class F>
   auto async(F&& func) {
     using ResultType = std::invoke_result_t<std::decay_t<F>>;
--- a/util/include/thread_pool.hpp
+++ b/util/include/thread_pool.hpp
@@ -58,12 +58,16 @@ class ThreadPool {
   // Executes the passed function (or any callable) in a pool thread. Returns a future to the result.
   // Note: The returned future's destructor doesn't block if the future value hasn't been retrieved. This is in contrast
   // to the futures returned by std::async that block.
-  template <class F>
-  auto async(F&& func) {
-    using ResultType = std::invoke_result_t<std::decay_t<F>>;
-    auto ptask = std::packaged_task<ResultType()>{std::forward<F>(func)};
+  template <class F, class... Args>
+  auto async(F&& func, Args&&... args) {
+    using ResultType = std::invoke_result_t<std::decay_t<F>, std::decay_t<Args>...>;
+    auto ptask = std::packaged_task<ResultType(Args...)>{std::forward<F>(func)};
     auto future = ptask.get_future();
-    auto task = GenericTask{[ptask = std::move(ptask)]() mutable { ptask(); }};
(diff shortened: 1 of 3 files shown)


util/include/thread_pool.hpp
```diff
@@ -0,0 +1,123 @@
...
+    std::queue<GenericTask> tasks;
+
+    // A mutex for the queue.
+    std::mutex mutex;
+
+    // A condition variable to signal the queue.
+    std::condition_variable cv;
+  };
+
+  // Each thread has its own queue - rationale is that doing so will lessen the contention compared to having a single
```
toly-kournik, line 109: On the other hand starvation is very probable.
Personally I'd implement a single queue

dartdart26: Good point. Do you mean a case where tasks have different complexities?

I think, in general, we can measure it. Also, we can have a policy to specify if there should be one queue only. Another option is to implement work stealing.

Please let me know what your thoughts are on above.

dartdart26: After an offline discussion, I've pushed a commit that implements by using a single queue.

dartdart26:
Will add a future commit with argument support as in `std::async()`,  i.e. `pool.async(func, arg1, arg2...)`.

util/include/thread_pool.hpp
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+ public:
+  // Executes the passed function (or any callable) in a pool thread. Returns a future to the result.
+  // Note: The returned future's destructor doesn't block if the future value hasn't been retrieved. This is in contrast
+  // to the futures returned by std::async that block.
+  template <class F>
+  auto async(F&& func) {
+    using ResultType = std::invoke_result_t<std::decay_t<F>>;
```
andrewjstone, line 63: I've tried, but still don't fully understand why `std::decay_t<F>` is needed here. Do function pointers not work without it? What callable needs to be decayed?

util/include/thread_pool.hpp
```diff
@@ -0,0 +1,115 @@
...
+    {
+      auto lock = std::lock_guard{task_queue_.mutex};
+      task_queue_.tasks.push(std::move(task));
+    }
+    task_queue_.cv.notify_one();
+    return future;
+  }
+
+ private:
+  using GenericTask = std::packaged_task<void()>;
```
andrewjstone, line 76: Does this type erasure make it possible to add tasks with different return types? If so, can we do that in the tests as well please?

util/include/thread_pool.hpp
dartdart26: It's a bit tricky and not extremely practical: https://stackoverflow.com/questions/37834567/decay-types-before-passing-to-stdresult-of

util/include/thread_pool.hpp
dartdart26: Very good point. I have such a test in my next commit.

util/include/thread_pool.hpp
andrewjstone: Thanks!

dartdart26:
@HristoStaykov Thanks for the offline comment about the `non_blocking_future_dtors` test not doing what it is supposed to. Fixed by making sure the future's dtor is called while the lambda is still running.

HristoStaykov:
Great work, Peter! User friendly implementation and very thorough unit tests.

PR #707, opened 2020-06-24
Remove implicitly deleted move-assignment operator

f-squirrel:
Since the class has the const reference `const IDBClient& db_` as
private member, the move-assignmanet operator is implicitly removed.

For more details refer to: https://en.cppreference.com/w/cpp/language/move_assignment
Section: Deleted implicitly-declared move assignment operator

### DIFF
--- a/storage/include/storage/db_interface.h
+++ b/storage/include/storage/db_interface.h
@@ -101,7 +101,6 @@ class IDBClient {
       Guard(const Guard&) = delete;
       Guard& operator=(const Guard&) = delete;
       Guard(Guard&&) = default;
-      Guard& operator=(Guard&&) = default;
       ~Guard() noexcept {
         try {
           db_.freeIterator(&iter_);


dartdart26:
@f-squirrel You are right. Could you please explain why we should remove the explicit deletion? Does it have to do with overload resolution?

f-squirrel:
@dartdart26 , we delete the explicitly defaulted move-assignment operator `Guard& operator=(Guard&&) = default;` that is implicitly deleted because we declare `const IDBClient& db_;`, which cannot be moved.

dartdart26:
> @dartdart26 , we delete the explicitly defaulted move-assignment operator `Guard& operator=(Guard&&) = default;` that is implicitly deleted because we declare `const IDBClient& db_;`, which cannot be moved.
> In general, the change is not critical but makes the IDE less annoying:)

Oh, you are right, sorry. It is because of the `const`. It is confusing if we leave it. Good catch, thanks!

PR #710, opened 2020-06-25
[Apollo] Continuous checkpoint creation while replicas crash

shgandhi:
In this PR, a test was added for continuous checkpoint creation while replicas crash.
Herein we continuously - crash f replicas, trigger and verify checkpoint creation and then bring back crashed replicas. 
The objective is to verify checkpoint creation amid replicas that continuously go down and come back up.

### DIFF
--- a/tests/apollo/test_skvbc_checkpoints.py
+++ b/tests/apollo/test_skvbc_checkpoints.py
@@ -12,11 +12,9 @@
 import os.path
 import random
 import unittest
-from os import environ
 
 import trio
 
-from util import blinking_replica
 from util import skvbc as kvbc
 from util.bft import with_trio, with_bft_network, with_constant_load, KEY_FILE_PREFIX
 
@@ -96,14 +94,15 @@ class SkvbcCheckpointTest(unittest.TestCase):
 
         checkpoint_before_primary = await bft_network.wait_for_checkpoint(replica_id=current_primary)
 
-        crashed_replicas = await self._crash_replicas_without_primary(
+        crashed_replicas = await self._crash_replicas(
             bft_network=bft_network,
-            nb_crashing=f
+            nb_crashing=f,
+            exclude_replicas={current_primary}
         )
         self.assertFalse(current_primary in crashed_replicas)
 
         self.assertGreaterEqual(
-            len(bft_network.procs), 2 * f + 2 * c + 1,
+            len(bft_network.procs), 2 * f + c + 1,
             "Make sure enough replicas are up to allow a successful checkpoint creation")
 
         await skvbc.fill_and_wait_for_checkpoint(
@@ -117,7 +116,7 @@ class SkvbcCheckpointTest(unittest.TestCase):
         # verify checkpoint creation after f replicas crash
         self.assertEqual(checkpoint_after_primary, 1 + checkpoint_before_primary)
 
-        await self._bring_up_crashed_replicas(bft_network, crashed_replicas)
+        bft_network.start_replicas(crashed_replicas)
 
         stale_node = random.choice(list(crashed_replicas))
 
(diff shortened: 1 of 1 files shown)


shgandhi:
Hi @teoparvanov, thanks for the review. I have updated this PR to only include the first test. I will submit a new PR for view-change with the changes you suggested.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -128,6 +126,32 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+        c = bft_network.config.c
+
+        self.assertEqual(len(bft_network.procs), n,
+                         "Make sure all replicas are up initially.")
+        current_primary = await bft_network.get_current_primary()
+
+        await self._create_checkpoints_while_crashing(bft_network, skvbc, f, c,
+                                                      current_primary, nb_crashes=2)
+
     async def _crash_replicas_without_primary(
```
teoparvanov, line 155: As discussed in a previous PR, let's refactor this using the already existing utility @yontyon introduced some time ago:
https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft.py#L399

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -148,3 +172,37 @@ async def _bring_up_crashed_replicas(
...
+                                                 current_primary,
+                                                 nb_crashes=20):
+        for _ in range(nb_crashes):
+            crashed_replicas = await self._crash_replicas_without_primary(
+                bft_network=bft_network,
+                nb_crashing=f
+            )
+            self.assertFalse(current_primary in crashed_replicas)
+
+            self.assertGreaterEqual(len(bft_network.procs), 2 * f + 2 * c + 1,
```
teoparvanov, line 188: I think 2f+c+1 replicas are enough for checkpoint creation.

tests/apollo/test_skvbc_checkpoints.py
shgandhi: @teoparvanov, updated.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -128,23 +126,79 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+        crash_replicas = bft_network.random_set_of_replicas(nb_crashing, without={current_primary})
 
-        crash_candidates = bft_network.all_replicas(
-            without={current_primary})
-        random.shuffle(crash_candidates)
-        for i in range(nb_crashing):
-            bft_network.stop_replica(crash_candidates[i])
-            crashed_replicas.add(crash_candidates[i])
+        for crash_replica in crash_replicas:
+            bft_network.stop_replica(crash_replica)
```
teoparvanov, line 162: Let's use the `bft_network.stop_replicas()` method here.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -128,23 +126,79 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+                    assert_state_transfer_not_started=False
+                )
+            except trio.TooSlowError:
+                self.fail("Unable to create enough client requests for checkpoints.")
+
+            await bft_network.wait_for_checkpoint(current_primary, expected_checkpoint_num=checkpoint_before + 1)
+
+            await self._bring_up_crashed_replicas(bft_network, crashed_replicas)
+
+            await trio.sleep(30)
```
teoparvanov, line 203: I'm afraid this would significantly slow down the test. Do we really need to wait for 30 seconds here?

shgandhi: For 2 rounds (number of contiguous crashes), 5 seconds wait time is suffice, for 4 or more rounds, higher wait time is required (>= 10 seconds).

andrewjstone: If you'll notice, none of our tests use sleeps. This is because it makes them subject to non-determinism. Any value running on machine that is slower, or has a temporarily descheduled VM/container can cause the test to fail. This is why we have all these `wait_for` conditions. If you ever feel the need to use a sleep, instead you should reach for an existing wait_for or add your own. This allows optimal test runtime, and longer failure timeouts, since the wait_for will retry much more quickly than any sleep, and the failure timeout can be made larger without having to worry about average test runtime, thus avoiding nondeterminism problems with slow machines.

While I don't like to use the word 'never' much, I'd suggest that we should never use a sleep in any test, for any reason, other than maybe testing an implementation of sleep itself :)

shgandhi: Sounds reasonable, @andrewjstone. Will update. :)

shgandhi: I have updated the test s.t., stale nodes now wait for checkpoint to propagate to them after they are brought up, instead of waiting for an arbitrary amount of time.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -128,23 +127,69 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
-        crash_candidates = bft_network.all_replicas(
-            without={current_primary})
-        random.shuffle(crash_candidates)
-        for i in range(nb_crashing):
-            bft_network.stop_replica(crash_candidates[i])
-            crashed_replicas.add(crash_candidates[i])
+            bft_network.start_replicas(crashed_replicas)
 
-        return crashed_replicas
+            await bft_network.wait_for_replicas_to_checkpoint(crashed_replicas, checkpoint_before + 1)
```
teoparvanov, line 193: Instead of doing this, I'd suggest the following:
```
await skvbc.fill_and_wait_for_checkpoint(
                initial_nodes=bft_network.all_replicas(),
                checkpoint_num=1,
                verify_checkpoint_persistency=False,
                assert_state_transfer_not_started=False
            )
```
This should force the stale replicas to rejoin the consensus and participate in the next checkpoint.

teoparvanov: I'd suggest to close this PR in favor of enhancing `test_checkpoint_propagation_after_crashed_replicas_comeback`, to verify the checkpoint propagates to all crashed replicas after they are brought up.

Next up, you could consider adding some network partitioning while creating checkpoints. See the existing examples in https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft_network_partitioning.py

teoparvanov:
@shgandhi , I'd suggest to close this PR in favor of enhancing test_checkpoint_propagation_after_crashed_replicas_comeback, to verify the checkpoint propagates to all crashed replicas after they are brought up.

Next up, you could consider adding some network partitioning while creating checkpoints. See the existing examples in https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft_network_partitioning.py

shgandhi:
> @shgandhi , I'd suggest to close this PR in favor of enhancing test_checkpoint_propagation_after_crashed_replicas_comeback, to verify the checkpoint propagates to all crashed replicas after they are brought up.
> 
> Next up, you could consider adding some network partitioning while creating checkpoints. See the existing examples in https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft_network_partitioning.py

Thanks @teoparvanov! Will do so.

PR #714, opened 2020-06-29
Fix a vector size bug in the merkle benchmark

dartdart26:
Call std::vector::reserve() rather than std::vector::resize() in
randomBuffer() in the sparse merkle benchmark. Calling resize()
generates a vector with an incorrect size.


dartdart26:
Fastest review ever, thanks!

PR #716, opened 2020-06-30
increase test timeout to improve stability

yontyon:
Increase timeout of pre execution test to increase stability (https://travis-ci.com/github/vmware/concord-bft/jobs/355349818)

### DIFF
--- a/tests/apollo/test_skvbc_preexecution.py
+++ b/tests/apollo/test_skvbc_preexecution.py
@@ -26,30 +26,28 @@ MAX_CONCURRENCY = 10
 SHORT_REQ_TIMEOUT_MILLI = 3000
 LONG_REQ_TIMEOUT_MILLI = 15000
 
-def start_replica_cmd(builddir, replica_id):
+def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="10000"):
     """
     Return a command that starts an skvbc replica when passed to
     subprocess.Popen.
-
-    The replica is started with a short view change timeout and with RocksDB
-    persistence enabled (-p).
-
     Note each arguments is an element in a list.
     """
-
-    status_timer_milli = "500"
-    view_change_timeout_milli = "10000"
-
+    statusTimerMilli = "500"
     path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
     return [path,
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
-            "-s", status_timer_milli,
+            "-s", statusTimerMilli,
             "-v", view_change_timeout_milli,
-            "-p",
-            "-t", os.environ.get('STORAGE_TYPE')
-            ]
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower()
+                    in set(["true", "on"])
+                 else "",
+            "-t", os.environ.get('STORAGE_TYPE')]
 
+def start_replica_cmd_with_vc_timeout(vc_timeout):
+    def wrapper(*args, **kwargs):
+        return start_replica_cmd(*args, **kwargs, view_change_timeout_milli=vc_timeout)
+    return wrapper
 
 class SkvbcPreExecutionTest(unittest.TestCase):
 
@@ -327,7 +325,7 @@ class SkvbcPreExecutionTest(unittest.TestCase):
 
 
     @with_trio
-    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: f >= 2)
+    @with_bft_network(start_replica_cmd_with_vc_timeout("20000"), selected_configs=lambda n, f, c: f >= 2)
     @verify_linearizability(pre_exec_enabled=True)
     async def test_alternate_f_isolated(self, bft_network, tracker):
         '''
--- a/tests/apollo/test_skvbc_preexecution.py
+++ b/tests/apollo/test_skvbc_preexecution.py
@@ -94,7 +94,7 @@ class SkvbcPreExecutionTest(unittest.TestCase):
     @unittest.skip("unstable")
     @with_trio
     @with_bft_network(start_replica_cmd)
-    @verify_linearizability(pre_exec_enabled=True)
+    @verify_linearizability(pre_exec_enabled=True, no_conflicts=True)
     async def test_sequential_pre_process_requests(self, bft_network, tracker):
         """
         Use a random client to launch one pre-process request in time and ensure that created blocks are as expected.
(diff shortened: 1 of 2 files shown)


tests/apollo/test_skvbc_preexecution.py
```diff
@@ -369,7 +369,7 @@ async def test_alternate_f_isolated(self, bft_network, tracker):
 
     async def issue_tracked_ops_to_the_system(self, tracker):
         try:
-            with trio.move_on_after(seconds=30):
+            with trio.move_on_after(seconds=40):
```
teoparvanov, line 372: So, your fix is to effectively send more requests? What's the reasoning behind this?

yontyon: From what I've seen, the scenario is as following:
Once we release the first f replicas from the isolation and isolate a different set of f replicas, a view change may be initiated.
If the view change takes too much time the tracker fails from sending any request (because we catch the tooSlowError and pass) and then we end up without creating any new block.

Increasing this timeout gives the tracker more opportunities to issue a successful command.
Although, now that I'm thinking of it, we should try to issue more than 10 requests...

teoparvanov: I'm wondering though... Don't we rather want to avoid such unexpected view changes? You could increase the view change timeout for tests where we don't want view changes, like here:
https://github.com/vmware/concord-bft/blob/master/tests/apollo/test_skvbc_chaotic_startup.py#L307

yontyon: mmm, a view change is a reasonable scenario in this case, so why preventing it artificially?

teoparvanov: I'd argue that if the _f_ crashed replicas do not include the primary, we should not have view change. The system should be able to proceed with the same primary, even with _f_ crashed nodes...

yontyon: They did not crash, in this test we isolate a group A of f replicas, and check the system is able to make progress.
Then we release group A and isolate a distinct group B of f replicas and check that the system is able to make progress.

So it makes sense that sometimes group A will try to perform a view change

And when I'm thinking of it now, if only f replicas are trying to make a view change they should not succeed, but the remaining f+1 replicas will also get stack and they will initiate a view change.
So we will have view change in this test in anyway, we just want to send enough messages such that we will see progress eventually (we can even wait for sure for a view change)

teoparvanov: Well, it depends how you isolate group A :) If they are also isolated from client requests, there won't be view change.

yontyon: Yeah, you write...
I will increase the viewchange timeout

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -369,7 +367,7 @@ async def test_alternate_f_isolated(self, bft_network, tracker):
 
     async def issue_tracked_ops_to_the_system(self, tracker):
         try:
-            with trio.move_on_after(seconds=30):
-                await tracker.run_concurrent_ops(10, write_weight=1)
+            with trio.move_on_after(seconds=40):
+                await tracker.run_concurrent_ops(100, write_weight=1)
```
teoparvanov, line 371: If you're increasing the view change timeout, can we then reduce the number of concurrent ops to 50? I noticed that with your PR, the tests take ~40min, as opposed to ~35min on master...

teoparvanov: PS: otherwise, the rest looks good to me!

tests/apollo/test_skvbc_preexecution.py
```diff
@@ -370,6 +368,6 @@ async def test_alternate_f_isolated(self, bft_network, tracker):
     async def issue_tracked_ops_to_the_system(self, tracker):
         try:
             with trio.move_on_after(seconds=30):
-                await tracker.run_concurrent_ops(10, write_weight=1)
+                await tracker.run_concurrent_ops(50, write_weight=1)
```
teoparvanov, line 371: Instead of increasing the number of requests, let's add another parameter to the `@verify_linearizability` - `with_conflicts=False`. This parameter would then get propagated to the SkvbcTracker & PassThroughSkvbcTracker.
We could use it to modify the behavior of the `readset()` method - to return empty if `with_conflicts` is True.
PS: let's also drop the view change timeout parametrization from this PR

@yontyon what do you think?

yontyon: sounds great
On it

yontyon: done

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -844,7 +848,10 @@ async def get_last_block_id(self, client):
         return kvbc.SimpleKVBCProtocol.parse_reply(await client.read(msg))
 
     async def send_tracked_write(self, client, max_set_size, long_exec=False):
-        readset = self.readset(0, max_set_size)
+        max_read_set_size = max_set_size
+        if self.no_conflicts:
+            max_read_set_size = 0
```
teoparvanov, line 853: A more "pythonistic" way to do this would be to say:
```python
max_read_set_size = 0 if self.no_conflicts else max_set_size
```

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -1011,15 +1018,20 @@ async def tracked_read_your_writes(self):
...
+
         self.skvbc = skvbc
 
         self.bft_network = bft_network
 
     async def send_tracked_write(self, client, max_set_size, long_exec=False):
-        readset = self.readset(0, max_set_size)
+        max_read_set_size = max_set_size
+        if self.no_conflicts:
+            max_read_set_size = 0
```
teoparvanov, line 1033: same here :)

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -377,12 +377,16 @@ class SkvbcTracker:
...
     in the future.
     """
-    def __init__(self, initial_kvpairs={}, skvbc=None, bft_network=None, pre_exec_all=False):
+    def __init__(self, initial_kvpairs={}, skvbc=None, bft_network=None, pre_exec_all=False, no_conflicts=False):
 
         # If this flag is set to True, it means that all the tracker requests will
         # go through the pre_execution mechanism
         self.pre_exec_all = pre_exec_all
 
+        # If set to True, it means that we always sends an empty read set in the bft request, in order to prevent
```
teoparvanov, line 386: ... in bft write requests ...

yontyon: done

tests/apollo/util/skvbc_history_tracker.py
yontyon: done

tests/apollo/util/skvbc_history_tracker.py
yontyon: done

PR #717, opened 2020-06-30
Obscure sensitive logs

oded282:
This PR is about obscuring sensitive information from the logs.
I wrote a script to find those sensitive logs and obscured them.

Changes:
- Removed access key config variable from the logs.

While searching for sensitive information, I have noticed we are logging ,in several files, seeds of PRNG.
Generally a seed shouldn't be logged and that is indeed sensitive information, but this seed is not used for cryptographic randomness, therefore in this case there is no security threat.
I added a comment to explicitly note this seed shouldn't be used for cryptographic randomness.


oded282:
These parts of the code use RELIC PRF for sensitive operations.
srand used for basics, not for key generation etc.

PR #719, opened 2020-06-30
[Apollo] Checkpoint propagation after view change

shgandhi:
In this test, we bring down f - 1 replicas, create a checkpoint, trigger
view change by bringing down the primary (total f replicas down).
We then verify checkpoint propagation in the new view.

### DIFF
--- a/tests/apollo/test_skvbc_checkpoints.py
+++ b/tests/apollo/test_skvbc_checkpoints.py
@@ -125,6 +125,109 @@ class SkvbcCheckpointTest(unittest.TestCase):
 
             self.assertEqual(checkpoint_stale, checkpoint_after_primary)
 
+    @unittest.skip("Unstable due to BC-3451")
+    @with_trio
+    @with_bft_network(start_replica_cmd)
+    async def test_checkpoint_propagation_after_crashed_replicas_comeback_after_vc(self, bft_network):
+        """
+        Here we bring down a total of f - 1 replicas, create checkpoint, trigger
+        view change by bringing down the primary (total f replicas down).
+        We then verify checkpoint propagation in the new view.
+        1) Given a BFT network, we make sure all nodes are up
+        2) Crash f - 1 replicas, excluding the initial primary, and next view primary
+        3) Send sufficient number of client requests to trigger checkpoint protocol
+        4) Make sure checkpoint is created
+        5) Crash the initial primary & send a batch of write requests
+        6) Verify view change has occurred
+        7) Verify checkpoint propagation to current primary in the new view
+        8) Bring the f replicas up
+        9) Verify checkpoint propagation to a stale node of the f replicas brought up
+        """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        n = bft_network.config.n
+        f = bft_network.config.f
+        c = bft_network.config.c
+
+        self.assertEqual(len(bft_network.procs), n,
+                         "Make sure all replicas are up initially.")
+
+        initial_primary = await bft_network.get_current_primary()
+
+        expected_next_primary = initial_primary + 1
+
+        checkpoint_init_primary_before = await bft_network.wait_for_checkpoint(replica_id=initial_primary)
+
+        # crashing f-1 replicas, initial_primary is crashed after checkpoint creation to trigger vc
+        crashed_replicas = await self._crash_replicas(
+            bft_network=bft_network,
+            nb_crashing=f - 1,
+            exclude_replicas={expected_next_primary, initial_primary}
+        )
+        self.assertFalse(initial_primary in crashed_replicas or expected_next_primary in crashed_replicas)
+
+        self.assertGreaterEqual(
+            len(bft_network.procs), 2 * f + 2 * c + 1,
+            "Make sure enough replicas are up to allow a successful view change!")
+
+        await skvbc.fill_and_wait_for_checkpoint(
+            initial_nodes=bft_network.all_replicas(without=crashed_replicas),
+            checkpoint_num=1,
+            verify_checkpoint_persistency=False
+        )
+
+        # verify checkpoint creation at the initial primary after f - 1 replicas down
+        checkpoint_init_primary_after = await bft_network.wait_for_checkpoint(replica_id=initial_primary)
+        self.assertEqual(checkpoint_init_primary_after, 1 + checkpoint_init_primary_before)
+
+        await bft_network.wait_for_view(
+            replica_id=initial_primary,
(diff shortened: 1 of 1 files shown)


teoparvanov:
@HristoStaykov could you please have a look and review this new test by @shgandhi ? Thanks!

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -128,23 +127,111 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+        # crashing f-1 replicas, initial_primary is crashed after checkpoint creation to trigger vc
+        crashed_replicas = await self._crash_replicas(
+            bft_network=bft_network,
+            nb_crashing=f - 1,
+            exclude_replicas={expected_next_primary, initial_primary}
+        )
+        self.assertFalse(expected_next_primary and initial_primary in crashed_replicas)
+
+        self.assertGreaterEqual(
+            len(bft_network.procs), 2 * f + c + 1,
```
HristoStaykov, line 173: I think the Quorum for ViewChange is ["2 * f + 2 * c + 1" ](https://github.com/vmware/concord-bft/blob/0d226cb5ae9def633ea9667d39ae402fd5f7ff1d/bftengine/src/bftengine/ViewsManager.cpp#L752)

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -128,23 +127,111 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+        self.assertEqual(current_primary, expected_next_primary)
+
+        checkpoint_current_primary = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+
+        # verify checkpoint propagation to current primary in the new view
+        self.assertEqual(checkpoint_current_primary, checkpoint_init_primary_after)
+
+        bft_network.start_replicas(crashed_replicas)
+
+        stale_node = random.choice(list(crashed_replicas))
```
HristoStaykov, line 222: Just a suggestion, wouldn't it be better to check that all crashed replicas get to the same checkpoint after they are brought up?

shgandhi:
Thanks for the review @HristoStaykov! I have made the changes you suggested.

shgandhi:
Thanks @teoparvanov! Done.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -123,6 +123,106 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+
+        checkpoint_init_primary_before = await bft_network.wait_for_checkpoint(replica_id=initial_primary)
+
+        # crashing f-1 replicas, initial_primary is crashed after checkpoint creation to trigger vc
+        crashed_replicas = await self._crash_replicas(
+            bft_network=bft_network,
+            nb_crashing=f - 1,
+            exclude_replicas={expected_next_primary, initial_primary}
+        )
+        self.assertFalse(expected_next_primary and initial_primary in crashed_replicas)
```
teoparvanov, line 166: The correct way to do this would be:
`self.assertTrue({initial_primary, expected_next_primary}.issubset(crashed_replicas))`

shgandhi: Thanks, updated.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -123,6 +123,106 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+        finally:
+            # wait for view change
+            await bft_network.wait_for_view(
+                replica_id=random.choice(bft_network.all_replicas(without=crashed_replicas)),
+                expected=lambda v: v == expected_next_primary,
+                err_msg="Make sure view change has been triggered."
+            )
+
+        current_primary = initial_primary + 1
+        self.assertEqual(current_primary, expected_next_primary)
```
teoparvanov, line 209: Sorry, I confused you, it should rather be:
`self.assertFalse(initial_primary in crashed_replicas or expected_next_primary in crashed_replicas)`

shgandhi: No worries, updated.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -123,6 +123,106 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+
+        checkpoint_current_primary = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+
+        # verify checkpoint propagation to current primary in the new view
+        self.assertEqual(checkpoint_current_primary, checkpoint_init_primary_after)
+
+        bft_network.start_replicas(crashed_replicas)
```
teoparvanov, line 216: Let's do a wait_for_view() for all crashed replicas before waiting for the checkpoints...

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -123,6 +123,114 @@ async def test_checkpoint_propagation_after_crashed_replicas_comeback(self, bft_
...
+        for crashed_replica in crashed_replicas:
+            await bft_network.wait_for_view(
+                replica_id=crashed_replica,
+                expected=lambda v: v == expected_next_primary,
+                err_msg=f"Make sure view change has been triggered for stale replica: {crashed_replica}."
+            )
+
+        # verify checkpoint propagation to all the stale nodes after they come back up
+        for crashed_replica in crashed_replicas:
+            checkpoint_stale = await bft_network.wait_for_checkpoint(
```
teoparvanov, line 228: I'm not sure why this is failing... I'd suggest logging a P1 JIRA bug, and decorating the test with `@unittest.skip("Unstable due to BC-XXXX")`. Then we can merge it into master and you'll investigate the bug separately.

As for the bug investigation, I guess there's no way around digging deeper in the logs and the C++ code related to checkpoints. 
Maybe a good starting point would be to track where the `last_stored_checkpoint` metric gets updated.
My guess is, the metric may not be correctly updated when crashed replicas catch-up with the others via `ReplicaStatusMsg`.

shgandhi: Thanks @teoparvanov! Makes sense, will do as you suggested.

teoparvanov:
Note: please avoid creating so many PRs on the same file. It makes both reviewing and merging harder... So mistakes are more likely to happen!

shgandhi:
Sorry for the confusion. Will avoid in the future.

teoparvanov:
No worries, it's just an advice :-)

PR #720, opened 2020-06-30
Add new BFT Client support

andrewjstone:
This new client is intended to be the canonical BFT client for C++
applications. It's expected that in the future other languages will
have their own clients. A future commit will modify the existing Python
client used in Apollo tests to have the same capabilities as this client.

This client introduces support for variably sized quorums and replica
specific information (RSI). Replica specific information is data
specific to each replica returned with a reply, that is excluded from
quorum matching. The RSI for each replica is returned along with the
respons to a `Client::Send` call. The client API exists in
`bft_client.h`, with the primary functionality being the blocking `Send`
function. Send is split into 2 different overloads: one for reading and
one for writing. This enables compiler driven typechecks, since reads
and writes support different quorums. Writes only support `Linearizable`
(2F+c+1) and `ByzantineSafe` (F+1) quorums, while reads support both of
those plus `All` and `MofN` quorums.

The four types of quorums are defined in quorum.h and each can be
parameterized by the destinations to send to. If no destinations are
given, then all replicas in the `ClientConfig` struct are used. If
specific destinations are given, they must contain enough replicas to be
safe. For example, in a `ClientConfig` that contains 4 replicas with an
`f_val` of 1, a `LinearizableQuorum` must contain at least 3
unique destinations, and those destinations must be replicas listed in
the ClientConfig. Additionally, all quorums can be converted to MofN
quorums. This conversion is done internally so that the `Matcher`
component does not need to work with multiple quorums. The reason we
don't just pass an MofN quorum is both for legibility, type safety in
Client::Send calls and to allow reasonable defaults. For example, all
current code is expected to issue writes with `LinearizableQuorums`, and
not need to fill in the destinations. This mimics SimpleClient.

There are 2 primary threads that communicate with each other in client
operations. The user of the client has a thread that calls
`Client::Send`. This thread blocks, waiting for a quorum of matching
replies from the caller. The other thread is the thread running the
`Communication` instance. It sends and receives raw messages to/from the
network. Received messages are translated into `UnmatchedReply` structs
and put on a queue to be sent to the Client::Thread where they can be
matched and quorum can be checked. The loop for this is done in
`bft_client.cpp` and utilizes a `Client::Wait` call to get replies.

This code base is split up into a few files containing core classes:

 * `bft_client.h/.cpp` - The main API of the client.
 * `base_types.h` - Types used throughout the rest of the
 client
 * `config.h` - All configuration structs. Each of these should
 be self explanatory. The top level configurations are `WriteConfig` and
 `ReadConfig` that get passed to the `Client::Send` function, which is
 the main API function of the library.
 * `exception.h` - Exceptions that can be thrown by the library
 * `quorums.h/.cpp` - All quorum structs, validation, and conversion
 *  `seq_num_generator.h` - The mechanism for generating sequence
 numbers used by SimpleClient. It's a separate class/file to emphasize
 that it is optional for users of the client.
 * `matcher.h/.cpp` - Matches a quorum of replies to ensure it can
 return them to the caller of ClientSend.
 * `msg_receiver.h/.cpp` - The implementation of the `IReceiver`
 interface of the communication module. Raw replies received from the
 network are handled in `MsgReceiver::onNewMessage`, converted to a
 message independent format, `UnmatchedReply`, and put onto a queue so
 they can be matched.

There are also unit tests for the MsgReceiver, Matcher, and quorums in
`bft_client_test.cpp`. There are higher level unit tests for the Client
API in `bft_client_api_tests.cpp`. These tests mock out the
communication layer, and utilize `Behaviors` which pretend to be the
replica network and send replies to client requests. A Behavior is just
a callback function that can send replies. The behavior runs in the
context of the `FakeCommunication` thread and therefore must communicate
with the real client API by using the msg receiver to return replies.

### DIFF
--- a/bftclient/include/bftclient/base_types.h
+++ b/bftclient/include/bftclient/base_types.h
@@ -1,0 +1,55 @@
+
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use
+// this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license
+// terms. Your use of these subcomponents is subject to the terms and conditions of the
+// subcomponent's license, as noted in the LICENSE file.
+
+#pragma once
+
+#include <cstdint>
+#include <map>
+#include <vector>
+
+namespace bft::client {
+
+// A typesafe replica id.
+struct ReplicaId {
+  uint16_t val;
+
+  bool operator==(const ReplicaId& other) const { return val == other.val; }
+  bool operator!=(const ReplicaId& other) const { return val != other.val; }
+  bool operator<(const ReplicaId& other) const { return val < other.val; }
+};
+
+// A typesafe client id.
+struct ClientId {
+  uint16_t val;
+
+  bool operator==(const ClientId& other) const { return val == other.val; }
+  bool operator!=(const ClientId& other) const { return val != other.val; }
+  bool operator<(const ClientId& other) const { return val < other.val; }
+};
+
+enum Flags : uint8_t { EMPTY_FLAGS_REQ = 0x0, READ_ONLY_REQ = 0x1, PRE_PROCESS_REQ = 0x2 };
+
+struct ReplicaSpecificInfo {
+  ReplicaId from;
+  std::vector<uint8_t> data;
+};
+
+typedef std::vector<uint8_t> Msg;
+
+// `matched_data` contains any data that must be identical for a quorum of replicas
+// `rsi` contains replica specific information that was received for each replying replica.
+struct Reply {
+  Msg matched_data;
+  std::map<ReplicaId, Msg> rsi;
+};
+
+}  // namespace bft::client
--- a/bftclient/include/bftclient/bft_client.h
+++ b/bftclient/include/bftclient/bft_client.h
@@ -1,0 +1,96 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <memory>
+#include <optional>
+
+#include "communication/ICommunication.hpp"
+#include "Logger.hpp"
+#include "DynamicUpperLimitWithSimpleFilter.hpp"
+
+#include "bftclient/config.h"
+#include "matcher.h"
+#include "msg_receiver.h"
+#include "exception.h"
+
+namespace bft::client {
+
+class Client {
+ public:
+  Client(std::unique_ptr<bft::communication::ICommunication> comm, const ClientConfig& config)
+      : communication_(std::move(comm)),
+        config_(config),
+        quorum_converter_(config_.all_replicas, config_.f_val, config_.c_val),
+        expected_commit_time_ms_(config_.retry_timeout_config.initial_retry_timeout.count(),
+                                 config_.retry_timeout_config.number_of_standard_deviations_to_tolerate,
+                                 config_.retry_timeout_config.max_retry_timeout.count(),
+                                 config_.retry_timeout_config.min_retry_timeout.count(),
+                                 config_.retry_timeout_config.samples_per_evaluation,
+                                 config_.retry_timeout_config.samples_until_reset,
+                                 config_.retry_timeout_config.max_increasing_factor,
+                                 config_.retry_timeout_config.max_decreasing_factor) {
+    communication_->setReceiver(config_.id.val, &receiver_);
+    communication_->Start();
+  }
+
+  void stop() { communication_->Stop(); }
+
+  // Send a message where the reply gets allocated by the callee and returned in a vector.
+  // The message to be sent is moved into the caller to prevent unnecessary copies.
+  //
+  // Throws a BftClientException on error.
+  Reply send(const WriteConfig& config, Msg&& request);
+  Reply send(const ReadConfig& config, Msg&& request);
+
+  // Useful for testing. Shouldn't be relied on in production.
+  std::optional<ReplicaId> primary() { return primary_; }
+
+ private:
+  // Generic function for sending a read or write message.
+  Reply send(const MatchConfig& match_config, const RequestConfig& request_config, Msg&& request, bool read_only);
+
+  // Wait for messages until we get a quorum or a retry timeout.
+  //
+  // Return a Reply on quorum, or std::nullopt on timeout.
+  std::optional<Reply> wait();
+
+  // Send a Msg to all destinations in the configured quorum.
+  void sendToGroup(const MatchConfig& config, const Msg& msg);
+
+  // Extract a matcher configurations from operational configurations
+  //
+  // Throws BftClientException on error.
+  MatchConfig writeConfigToMatchConfig(const WriteConfig&);
+  MatchConfig readConfigToMatchConfig(const ReadConfig&);
+
+  MsgReceiver receiver_;
+
+  std::unique_ptr<bft::communication::ICommunication> communication_;
+  ClientConfig config_;
+  logging::Logger logger_ = logging::getLogger("bftclient");
+
+  // The client doesn't always know the current primary.
+  std::optional<ReplicaId> primary_;
+
+  // Each outstanding request matches replies using a new matcher.
+  // If there are no outstanding requests, then this is a nullopt;
+  std::optional<Matcher> outstanding_request_;
+
+  // A class that takes all Quorum types and converts them to an MofN quorum, with validation.
+  QuorumConverter quorum_converter_;
+
+  // A utility for calculating dynamic timeouts for replies.
+  bftEngine::impl::DynamicUpperLimitWithSimpleFilter<uint64_t> expected_commit_time_ms_;
+};
+
+}  // namespace bft::client
--- a/bftclient/include/bftclient/config.h
+++ b/bftclient/include/bftclient/config.h
@@ -1,0 +1,90 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <chrono>
+#include <map>
+#include <set>
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "bftclient/base_types.h"
+#include "bftclient/quorums.h"
+
+using namespace std::chrono_literals;
+
+namespace bft::client {
+
+// This config is based on the parameters to the DynamicUpperLimitWithSimpleFilter.
+// The defaults are set to the original defaults in SimpleClientImp.
+struct RetryTimeoutConfig {
+  // The starting retry timeout before we have any calculations.
+  std::chrono::milliseconds initial_retry_timeout = 150ms;
+
+  // The minimum that the timeout can be set to dynamically.
+  std::chrono::milliseconds min_retry_timeout = 50ms;
+
+  // The maximum that the timeout can be set to dynamically.
+  std::chrono::milliseconds max_retry_timeout = 1s;
+
+  // The number of standard deviations within the average that we expect reply times to fall into.
+  uint16_t number_of_standard_deviations_to_tolerate = 2;
+
+  // The number of replies we sample before we attempt to recalculate the rolling average and variance.
+  uint16_t samples_per_evaluation = 32;
+
+  // The number of samples before we reset the rolling average and variance to its empty state
+  // A negative number will turn off resets.
+  int16_t samples_until_reset = 1000;
+
+  // The scaling factor at which the timeout can be increased. A factor of 2 means that the upper
+  // limit can be doubled on each evaluation period.
+  double max_increasing_factor = 2;
+
+  // The scaling factor at which the timeout can be decreased. A factor of 2 means that the upper
+  // limit can be halved on each evaluation period.
+  double max_decreasing_factor = 2;
+};
+
+// The configuration for a single instance of a client.
+struct ClientConfig {
+  ClientId id;
+  std::set<ReplicaId> all_replicas;
+  uint16_t f_val;
+  uint16_t c_val;
+  RetryTimeoutConfig retry_timeout_config;
+};
+
+// Generic per-request configuration shared by reads and writes.
+struct RequestConfig {
+  bool pre_execute = false;
+  uint64_t sequence_number;
+  uint32_t max_reply_size = 64 * 1024;
+  std::chrono::milliseconds timeout = 5s;
+  std::string correlation_id = "";
+  std::string span_context = "";
+};
+
+// The configuration for a single write request.
+struct WriteConfig {
+  RequestConfig request;
+  WriteQuorum quorum;
+};
+
+// The configuration for a single read request.
+struct ReadConfig {
+  RequestConfig request;
+  ReadQuorum quorum;
+};
+
+}  // namespace bft::client
--- a/bftclient/include/bftclient/exception.h
+++ b/bftclient/include/bftclient/exception.h
@@ -1,0 +1,46 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <stdexcept>
+#include <string>
+
+#include "bftclient/config.h"
+
+namespace bft::client {
+
+class BftClientException : public std::runtime_error::runtime_error {
+ public:
+  explicit BftClientException(const std::string& what) : runtime_error(what) {}
+};
+
+class BadQuorumConfigException : public BftClientException {
+ public:
+  BadQuorumConfigException(const std::string& what) : BftClientException(what) {}
+};
+
+class InvalidDestinationException : public BftClientException {
+ public:
+  InvalidDestinationException(ReplicaId replica_id)
+      : BftClientException("Replica: " + std::to_string(replica_id.val) + " is not part of the cluster.") {}
+
+  InvalidDestinationException() : BftClientException("MofN quorums must have destinations") {}
+};
+
+class TimeoutException : public BftClientException {
+ public:
+  TimeoutException(uint64_t seq_num, std::string cid)
+      : BftClientException("Timeout for request sequence number: " + std::to_string(seq_num) +
+                           ", and correlation id: " + cid) {}
+};
+
+}  // namespace bft::client
--- a/bftclient/include/bftclient/quorums.h
+++ b/bftclient/include/bftclient/quorums.h
@@ -1,0 +1,82 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <set>
+#include <variant>
+
+#include "bftclient/base_types.h"
+
+namespace bft::client {
+
+// A quorum of 2F + C + 1 matching replies from `destination` must be received for the `send` call
+// to complete.
+struct LinearizableQuorum {
+  std::set<ReplicaId> destinations;
+};
+
+// A quorum of F + 1 matching replies from `destination` must be received for the `send` call to complete.
+struct ByzantineSafeQuorum {
+  std::set<ReplicaId> destinations;
+};
+
+// A matching reply from every replica in `destination` must be received for the `send` call to complete.
+struct All {
+  std::set<ReplicaId> destinations;
+};
+
+// A matching reply from `wait_for` number of replicas from `destination` must be received for the
+// `send` call to complete.
+struct MofN {
+  size_t wait_for;
+  std::set<ReplicaId> destinations;
+};
+
+// Reads and writes support different types of quorums.
+typedef std::variant<LinearizableQuorum, ByzantineSafeQuorum, All, MofN> ReadQuorum;
+typedef std::variant<LinearizableQuorum, ByzantineSafeQuorum> WriteQuorum;
+
+// Convert a ReadQuorum or a WriteQuorum to an MofN Quorum, given:
+//  * All replicas in the cluster
+//  * F value
+//  * C value
+class QuorumConverter {
+ public:
+  QuorumConverter(const std::set<ReplicaId>& all_replicas, uint16_t f_val, uint16_t c_val)
+      : all_replicas_(all_replicas),
+        linearizable_quorum_size_(2 * f_val + c_val + 1),
+        bft_safe_quorum_size_(f_val + 1) {}
+
+  // Convert Quorums to a compatible MofN quorum for use by the matcher.
+  //
+  // This conversion handles quorum validation.
+  // The conversion from MofN to MofN only does validation.
+  //
+  // Throws BftClientException on error.
+  MofN toMofN(const LinearizableQuorum& quorum) const;
+  MofN toMofN(const ByzantineSafeQuorum& quorum) const;
+  MofN toMofN(const All& quorum) const;
+  MofN toMofN(const MofN& quorum) const;
+
+ private:
+  // Ensure that each replica in `destination` is part of `all_replicas`
+  //
+  // Throws an InvalidDestinationException if validation fails.
+  void validateDestinations(const std::set<ReplicaId>& destinations) const;
+
+  std::set<ReplicaId> all_replicas_;
+
+  size_t linearizable_quorum_size_;
+  size_t bft_safe_quorum_size_;
+};
+
+}  // namespace bft::client
--- a/bftclient/include/bftclient/seq_num_generator.h
+++ b/bftclient/include/bftclient/seq_num_generator.h
@@ -1,0 +1,82 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use
+// this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license
+// terms. Your use of these subcomponents is subject to the terms and conditions of the
+// subcomponent's license, as noted in the LICENSE file.
+
+#pragma once
+
+#include <chrono>
+#include <cstdint>
+
+#include "assertUtils.hpp"
+#include "base_types.h"
+
+namespace bft::client {
+
+// This is a basic implementation of a monotonic clock based sequence number.
+//
+// Using this class is safe, but not necessarily live. It is safe because messages from a client
+// with an old (or identical) sequence number are simply ignored. It may not be live because if your machine
+// reboots, its clock may be set "backwards". Within a few seconds however, given NTP,
+// things should be back to normal and new requests will be handled. This is actually a very
+// unlikely scenario, as reboots will generally take longer than any possible clock drift. Therefore
+// this is a good default sequence number generator.
+//
+// Users are allowed to generate their own sequence numbers (they do not
+// have to use this class). Other examples of ways to do this include:
+// (1) A simple counter + store the last counter in a persistent storage
+// (2) A mechanism to retrieve the last used sequence number from the replicas on restart. No such
+//     mechanism is cucrrently implemented, buf if necessary we will add it.
+class SeqNumberGenerator {
+  SeqNumberGenerator(ClientId client_id) : client_id_(client_id) {}
+
+  uint64_t unique() {
+    std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+    return unique(now);
+  }
+
+  uint64_t unique(std::chrono::time_point<std::chrono::system_clock> now) {
+    uint64_t milli = std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()).count();
+
+    if (milli > lastMilliOfUniqueFetchID_) {
+      lastMilliOfUniqueFetchID_ = milli;
+      lastCountOfUniqueFetchID_ = 0;
+    } else {
+      if (lastCountOfUniqueFetchID_ == LAST_COUNT_LIMIT) {
+        LOG_WARN(logger_, "Client SeqNum Counter reached max value. " << KVLOG(client_id_));
+        lastMilliOfUniqueFetchID_++;
+        lastCountOfUniqueFetchID_ = 0;
+      } else {
+        // increase last count to preserve uniqueness.
+        lastCountOfUniqueFetchID_++;
+      }
+    }
+    // shift lastMilli by 22 (0x3FFFFF) in order to 'bitwise or' with lastCount
+    // and preserve uniqueness and monotonicity.
+    uint64_t r = (lastMilliOfUniqueFetchID_ << (64 - 42));
+    Assert(lastCountOfUniqueFetchID_ <= LAST_COUNT_LIMIT);
+    r = r | ((uint64_t)lastCountOfUniqueFetchID_);
+
+    return r;
+  }
+
+ private:
+  // limited to the size lastMilli shifted.
+  const uint64_t LAST_COUNT_LIMIT = 0x3FFFFF;
+  // lastMilliOfUniqueFetchID_ holds the last SN generated,
+  uint64_t lastMilliOfUniqueFetchID_ = 0;
+  // lastCount used to preserve uniqueness.
+  uint32_t lastCountOfUniqueFetchID_ = 0;
+
+  ClientId client_id_;
+
+  logging::Logger logger_ = logging::getLogger("bftclient.seqnum");
+};
+
+}  // namespace bft::client
--- a/bftclient/src/bft_client.cpp
+++ b/bftclient/src/bft_client.cpp
@@ -1,0 +1,165 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use
+// this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license
+// terms. Your use of these subcomponents is subject to the terms and conditions of the
+// subcomponent's license, as noted in the LICENSE file.
+
+#include "bftclient/bft_client.h"
+#include "bftengine/ClientMsgs.hpp"
+#include "assertUtils.hpp"
+
+namespace bft::client {
+
+// This function creates a ClientRequestMsg or a ClientPreProcessRequestMsg depending upon config.
+//
+// Since both of these are just instances of a `ClientRequestMsgHeader` followed by the message
+// data, we construct them here, rather than relying on the type constructors embedded into the
+// bftEngine impl. This allows us to not have to link with the bftengine library, and also allows us
+// to return the messages as vectors with proper RAII based memory management.
+Msg makeClientMsg(const RequestConfig& config, Msg&& request, bool read_only, uint16_t client_id) {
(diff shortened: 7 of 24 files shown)


bftclient/src/matcher.h
```diff
@@ -0,0 +1,95 @@
...
+  Reply reply;
+  std::optional<ReplicaId> primary;
+};
+
+// Match replies for a given quorum.
+class Matcher {
+ public:
+  Matcher(const MatchConfig& config) : config_(config) {}
+
+  std::optional<Match> onReply(UnmatchedReply&& reply);
```
robem, line 60: UpperCaseFunctionName vs lowerCaseFunctionName. Or are you implementing an interface?

bftclient/include/bftclient/bft_client.h
```diff
@@ -0,0 +1,101 @@
...
+  std::optional<ReplicaId> primary() { return primary_; }
+
+ private:
+  // Generic function for sending a read or write message.
+  Reply Send(const MatchConfig& match_config, const RequestConfig& request_config, Msg&& request, bool read_only);
+
+  // Wait for messages until we get a quorum or a retry timeout.
+  //
+  // Return a Reply on quorum, or std::nullopt on timeout.
+  std::optional<Reply> Wait();
```
yuliasherman, line 67: I think we are missing a proper error reply handling in our functionality. For example, in the case of conflicting writes, we don't need to retry an operation.

bftclient/include/bftclient/bft_client.h
```diff
@@ -0,0 +1,101 @@
...
+
+  // Send a Msg to all destinations in the configured quorum.
+  void SendToGroup(const MatchConfig& config, const Msg& msg);
+
+  // A timeout has occurred. Perform any necessary cleanup.
+  Reply Timeout(const RequestConfig& config);
```
yuliasherman, line 73: -> OnTimeout

bftclient/include/bftclient/config.h
```diff
@@ -0,0 +1,89 @@
...
+#include <map>
+#include <set>
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "bftclient/base_types.h"
+#include "bftclient/quorums.h"
+
+using namespace std::chrono_literals;
```
yuliasherman, line 24: Please remove this from the include file

bftclient/include/bftclient/seq_num_generator.h
```diff
@@ -0,0 +1,71 @@
...
+
+  uint64_t Unique(std::chrono::time_point<std::chrono::system_clock> now) {
+    uint64_t milli = std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()).count();
+
+    if (milli > lastMilliOfUniqueFetchID_) {
+      lastMilliOfUniqueFetchID_ = milli;
+      lastCountOfUniqueFetchID_ = 0;
+    } else {
+      if (lastCountOfUniqueFetchID_ == last_count_limit) {
+        LOG_WARN(GL, "Client SeqNum Counter reached max value");
```
yuliasherman, line 47: I think the client-id should be printed in all logs - it is important when used in a pool.

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,162 @@
...
+  auto match_config = WriteConfigToMatchConfig(config);
+  bool read_only = false;
+  return Send(match_config, config.request, std::move(request), read_only);
+}
+
+Reply Client::Send(const ReadConfig& config, Msg&& request) {
+  Assert(!outstanding_request_.has_value());
+  auto match_config = ReadConfigToMatchConfig(config);
+  bool read_only = true;
+  return Send(match_config, config.request, std::move(request), read_only);
```
yuliasherman, line 63: I'm getting a feeling that division to ReadConfig/WriteConfig is artificial. Most of the code is the same for both.
Maybe the addition of a read_only flag to the configuration is enough.

bftclient/src/msg_receiver.cpp
```diff
@@ -0,0 +1,100 @@
...
+// subcomponent's license, as noted in the LICENSE file.
+
+#include "kvstream.h"
+#include "assertUtils.hpp"
+#include "bftengine/ClientMsgs.hpp"
+#include "msg_receiver.h"
+
+namespace bft::client {
+
+void UnmatchedReplyQueue::push(UnmatchedReply&& reply) {
```
yuliasherman, line 19: Looks like you finally moved to a low-case function names :-)

bftclient/src/quorums.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <algorithm>
+
+#include "bftclient/quorums.h"
+#include "bftclient/exception.h"
+
+namespace bft::client {
+
+MofN QuorumConverter::ToMofN(const LinearizableQuorum& quorum) const {
+  MofN new_quorum;
+  new_quorum.wait_for = 2 * f_val_ + c_val_ + 1;
```
yuliasherman, line 22: You can calculate this once and store it in an object.

bftclient/src/quorums.cpp
```diff
@@ -0,0 +1,92 @@
...
+  if (quorum.destinations.empty()) {
+    // If the user doesn't provide destinations, send to all replicas
+    new_quorum.destinations = all_replicas_;
+  } else if (quorum.destinations.size() < new_quorum.wait_for) {
+    throw BadQuorumConfigException(
+        "Destination does not contain enough replicas for a linearizable quorum. Destination size: " +
+        std::to_string(quorum.destinations.size()) +
+        "is less than 2f + c + 1 = " + std::to_string(new_quorum.wait_for));
```
yuliasherman, line 30: This could be defined once as a constant string to simplify a change if required.

bftclient/src/quorums.cpp
```diff
@@ -0,0 +1,92 @@
...
+  } else {
+    ValidateDestinations(quorum.destinations);
+    new_quorum.destinations = quorum.destinations;
+  }
+  return new_quorum;
+}
+
+MofN QuorumConverter::ToMofN(const ByzantineSafeQuorum& quorum) const {
+  MofN new_quorum;
+  new_quorum.wait_for = f_val_ + 1;
```
yuliasherman, line 40: The same as above

bftclient/src/quorums.cpp
```diff
@@ -0,0 +1,92 @@
...
+  if (quorum.destinations.empty()) {
+    // If the user doesn't provide a destination, send to all replicas
+    new_quorum.destinations = all_replicas_;
+  } else if (quorum.destinations.size() < new_quorum.wait_for) {
+    throw BadQuorumConfigException(
+        "Destination does not contain enough replicas for a byzantine fault tolerant quorum. Destination size: " +
+        std::to_string(quorum.destinations.size()) + "is less than f + 1 = " + std::to_string(new_quorum.wait_for));
```
yuliasherman, line 47: The same as above

bftengine/include/bftengine/ClientMsgs.hpp
```diff
@@ -39,8 +40,16 @@ struct ClientReplyMsgHeader {
...
   uint16_t currentPrimaryId;
   uint64_t reqSeqNum;
+
+  // Reply length is the total length of the reply, including any replica specific info.
   uint32_t replyLength;
+
+  // This is the size of the replica specific information. If it is 0, there is no replica specific
+  // information. The offset of the replica specific information from the start of the reply message
+  // is `replyLength - replicaSpecificInfoLength`.
+  uint32_t replicaSpecificInfoLength = 0;
```
yuliasherman, line 50: I would prefer to see replicaSpecificInfo instead of rsi :-)

bftclient/include/bftclient/config.h
```diff
@@ -0,0 +1,89 @@
...
+  int16_t samples_until_reset = 1000;
+
+  // The scaling factor at which the timeout can be increased. A factor of 2 means that the upper
+  // limit can be doubled on each evaluation period.
+  double max_increasing_factor = 2;
+
+  // The scaling factor at which the timeout can be decreased. A factor of 2 means that the upper
+  // limit can be halved on each evaluation period.
+  double max_decreasing_factor = 2;
+};
```
yuliasherman, line 57: I miss uint16_t clientPeriodicResetThresh = 30; here

bftclient/include/bftclient/config.h
```diff
@@ -0,0 +1,89 @@
...
+
+  // The number of standard deviations within the average that we expect reply times to fall into.
+  uint16_t number_of_standard_deviations_to_tolerate = 2;
+
+  // The number of replies we sample before we attempt to recalculate the rolling average and variance.
+  uint16_t samples_per_evaluation = 32;
+
+  // The number of samples before we reset the rolling average and variance to its empty state
+  // A negative number will turn off resets.
+  int16_t samples_until_reset = 1000;
```
yuliasherman, line 48: Is it intentionally not unsigned?

bftclient/include/bftclient/exception.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "bftclient/config.h"
+
+namespace bft::client {
+
+class BftClientException : public std::exception {
+ public:
+  explicit BftClientException(const std::string& what) : msg(what){};
+
+  virtual const char* what() const noexcept override { return msg.c_str(); }
```
f-squirrel, line 26: Is it intentional to inherit from `std::exception` and overloading `what`?
If we use `std::runtime_error` there is no need for overloading and adding the `msg` member.

bftclient/include/bftclient/seq_num_generator.h
```diff
@@ -0,0 +1,71 @@
...
+        lastCountOfUniqueFetchID_ = 0;
+      } else {
+        // increase last count to preserve uniqueness.
+        lastCountOfUniqueFetchID_++;
+      }
+    }
+    // shift lastMilli by 22 (0x3FFFFF) in order to 'bitwise or' with lastCount
+    // and preserve uniqueness and monotonicity.
+    uint64_t r = (lastMilliOfUniqueFetchID_ << (64 - 42));
+    Assert(lastCountOfUniqueFetchID_ <= 0x3FFFFF);
```
f-squirrel, line 58: Was `last_count_limit` supposed to be used here?

bftclient/include/bftclient/seq_num_generator.h
```diff
@@ -0,0 +1,71 @@
...
+    r = r | ((uint64_t)lastCountOfUniqueFetchID_);
+
+    return r;
+  }
+
+ private:
+  // limited to the size lastMilli shifted.
+  const uint64_t last_count_limit = 0x3FFFFF;
```
f-squirrel, line 66: Nit: `constexpr` and probably `LAST_COUNT_LIMIT`?

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,162 @@
...
+Msg MakeClientMsg(const RequestConfig& config, Msg&& request, bool read_only, uint16_t client_id) {
+  uint8_t flags = read_only ? READ_ONLY_REQ : EMPTY_FLAGS_REQ;
+  if (config.pre_execute) {
+    flags |= PRE_PROCESS_REQ;
+  }
+
+  auto header_size = sizeof(bftEngine::ClientRequestMsgHeader);
+
+  Msg msg(header_size + request.size());
+  bftEngine::ClientRequestMsgHeader* header = reinterpret_cast<bftEngine::ClientRequestMsgHeader*>(msg.data());
```
f-squirrel, line 33: This way of creation messages is dangerous because it does not take into account the variable size of the span context. I am afraid we'll end up creating a separate implementation of messages.

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,162 @@
...
+  return Send(match_config, config.request, std::move(request), read_only);
+}
+
+Reply Client::Send(const MatchConfig& match_config,
+                   const RequestConfig& request_config,
+                   Msg&& request,
+                   bool read_only) {
+  outstanding_request_ = Matcher(match_config);
+  receiver_.activate(request_config.max_reply_size);
+  auto msg = MakeClientMsg(request_config, std::move(request), read_only, config_.id.val);
```
f-squirrel, line 72: Since the message is used only as a const reference, it would be better to use `const auto&` to avoid possible `move`s or copies.

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,162 @@
...
+      if (auto match = outstanding_request_->onReply(std::move(req))) {
+        primary_ = match->primary;
+        outstanding_request_ = std::nullopt;
+        return match->reply;
+      }
+    }
+  }
+  // If there are multiple distinct replies, we may want to clear any replies to free memory. This really only matters
+  // for long running/indefinite requests.
+  static constexpr size_t clear_matcher_replies_threshold = 5;
```
f-squirrel, line 117: Nit: Probable `CLEAR_MATCHER_REPLIES_THRESHOLD`.

bftclient/src/matcher.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "matcher.h"
+
+namespace bft::client {
+
+std::optional<Match> Matcher::onReply(UnmatchedReply&& reply) {
+  if (!valid(reply)) return std::nullopt;
+
+  auto key = MatchKey{std::move(reply.metadata), std::move(reply.data)};
+  const auto [it, success] = matches_[key].insert_or_assign(reply.rsi.from, std::move(reply.rsi.data));
+  (void)it;  // unused variable hack needed by GCC
```
f-squirrel, line 21: `std::tie` + `std::ignore` may be used as alternative.

bftclient/src/matcher.h
```diff
@@ -0,0 +1,95 @@
...
+  size_t numDifferentReplies() const { return matches_.size(); }
+
+  void clearReplies() { matches_.clear(); }
+
+ private:
+  // Check the validity of a reply
+  bool valid(const UnmatchedReply& reply);
+
+  // Is the reply from a source listed in the quorum's destination?
+  bool validSource(const ReplicaId source);
```
f-squirrel, line 74: Either `const` is redundant or `const ReplicaId&` to be used.

bftclient/src/msg_receiver.cpp
```diff
@@ -0,0 +1,100 @@
...
+    LOG_WARN(logger_, "Invalid message received. Incorrect Header Type. " << KVLOG(header->msgType));
+    return;
+  }
+
+  auto metadata = ReplyMetadata{};
+  metadata.primary = ReplicaId{header->currentPrimaryId};
+  metadata.seq_num = header->reqSeqNum;
+
+  auto data_len = header->replyLength - sizeof(bftEngine::ClientReplyMsgHeader);
+  const char* start_of_body = message + sizeof(bftEngine::ClientReplyMsgHeader);
```
f-squirrel, line 72: The actual message body starts after the span context:
```
|header|span context|message body|cid|
```

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    if (auto reply = Wait()) {
+      expected_commit_time_ms_.add(
+          std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start).count());
+      return reply.value();
+    }
+  }
+  // This method always throws a  TimeoutException
+  return Timeout(request_config);
```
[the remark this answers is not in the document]
andrewjstone: I think it was leftover code from when there were two calls to Timeout. I inlined it as suggested.

bftclient/test/bft_client_api_tests.cpp
```diff
@@ -0,0 +1,336 @@
...
+  auto reply = client.Send(read_config, Msg({'h', 'e', 'l', 'l', 'o'}));
+  Msg expected{'w', 'o', 'r', 'l', 'd'};
+  ASSERT_EQ(expected, reply.matched_data);
+  for (auto& rsi : reply.rsi) {
+    ASSERT_TRUE(rsi.second.empty());
+  }
+  client.Stop();
+}
+
+static constexpr NodeNum bad_replica_id = 999;
```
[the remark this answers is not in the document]
andrewjstone: hahaha. FIXED!

bftclient/include/bftclient/bft_client.h
```diff
@@ -0,0 +1,101 @@
...
+#include "communication/ICommunication.hpp"
+#include "Logger.hpp"
+#include "DynamicUpperLimitWithSimpleFilter.hpp"
+
+#include "bftclient/config.h"
+#include "matcher.h"
+#include "msg_receiver.h"
+#include "exception.h"
+
+using namespace std::chrono_literals;
```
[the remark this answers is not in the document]
andrewjstone: Done!

bftclient/include/bftclient/bft_client.h
andrewjstone: Isn't this mechanism just  an error response that matches from a quorum of replicas? The client itself is unaware of the contents of messages and our current client replies don't include any sort of success or failure marker, so I don't know how we could do this otherwise. This implementation mirrors the SimpleClient behavior.

bftclient/include/bftclient/bft_client.h
andrewjstone: Removed as suggested by Rob.

bftclient/include/bftclient/config.h
andrewjstone: Doing that means making all the defaults much more verbose. I'd be really surprised if this caused serious problems for callers. I think the readability of the existing usage outweighs any risk of some user wanting to define their own "ms" or "s" suffixes.

bftclient/include/bftclient/seq_num_generator.h
andrewjstone: Will do! Actually, I should also add a new logger, rather than using GL. This was just a copy-paste from simpleClient that I never fixed.

andrewjstone: Hmm, actually, I can't do this currently. There is no clientId, as the seqNumGenerator is outside of the client. It gets passed into the client.

andrewjstone: I added a constructor that takes a ClientId.

bftclient/src/bft_client.cpp
andrewjstone: I don't feel it's artificiall. The semantics and capabilities of read and write requests are entirely different and the caller should be aware of how they work. The specific use case that drove this separation here is that Reads allow for more quorums than Writes. A read can be sent to just 1 or 2 replicas (via `All` or `MofN`), while a write cannot use these types of quorums. With this separation, we now get a *compile-time* check that the caller is appropriately using a quorum. If we had just one call, then we would need to write extra error handling code that would only be triggered at runtime. So we have both more code and less safety. We'd then also have to write tests for that error handling code rather than relying on correctness of the compiler.

 Furthermore, bool parameters are generally bad form, as you don't know what a literal true or false means at the call site and must look at the definition of the function.

bftclient/src/quorums.cpp
andrewjstone: I't be pretty strange for this to change, unless we changed protocols away from SBFT. Changing it inline doesn't feel like a burden.

bftclient/src/quorums.cpp
andrewjstone: Good call. Fixed.

bftclient/src/quorums.cpp
andrewjstone: done.

bftengine/include/bftengine/ClientMsgs.hpp
andrewjstone: I don't follow. Isn't that what's written?

bftclient/include/bftclient/config.h
andrewjstone: It's the same as resetPoint here: https://github.com/vmware/concord-bft/blob/master/util/include/DynamicUpperLimitWithSimpleFilter.hpp#L28

bftclient/include/bftclient/config.h
andrewjstone: Good catch :) 

I changed the name to `clear_matcher_replies_threshold` and hardcoded it at 5.

https://github.com/vmware/concord-bft/pull/720/files#diff-fe45c5cdd94cff5cafdb5f5e922ce0acR117-R120

 It's also slightly more sophisticated than the original in that it actually looks at how many distinct replies there are saved in memory, rather than just counting retransmissions. There's no problem saving all replies unless you are afraid of using too much memory. As far as I can tell there's no good reason to make this configurable with the current implementation so I just hardcoded it at 5. If we get 5 different responses after a few retransmissions we have some heavy contention around some data. 

We can always make it configurable later if need be.

bftclient/include/bftclient/base_types.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <cstdint>
+#include <map>
+#include <vector>
+
+namespace bft::client {
+
+// A typesafe replica id.
+struct ReplicaId {
+  uint16_t val;
```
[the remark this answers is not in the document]
andrewjstone: A constructor doesn't prevent initialization with garbage. You can always just write ReplicaId{}, unless you make the data private, which I think overcomplicates things. Blame C and C++ for even allowing uninitialized member variables.

bftclient/src/bft_client.cpp
andrewjstone: done.

bftclient/src/matcher.cpp
andrewjstone: I'm sticking to the basics for now :) GCC needs to get realize the value of destructuring and ignore this :)

bftclient/include/bftclient/base_types.h
```diff
@@ -0,0 +1,55 @@
...
+// A typesafe client id.
+struct ClientId {
+  uint16_t val;
+
+  bool operator==(const ClientId& other) const { return val == other.val; }
+  bool operator!=(const ClientId& other) const { return val != other.val; }
+  bool operator<(const ClientId& other) const { return val < other.val; }
+};
+
+enum Flags : uint8_t { EMPTY_FLAGS_REQ = 0x0, READ_ONLY_REQ = 0x1, PRE_PROCESS_REQ = 0x2 };
```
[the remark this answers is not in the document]
andrewjstone: I actually don't think the verbosity is worth it here. It's still part of the bft::client namespace

bftclient/include/bftclient/base_types.h
```diff
@@ -0,0 +1,55 @@
...
+
+struct ReplicaSpecificInfo {
+  ReplicaId from;
+  std::vector<char> data;
+};
+
+typedef std::vector<char> Msg;
```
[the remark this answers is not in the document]
andrewjstone: Good call. I was considering this change and will just do it!

bftclient/include/bftclient/base_types.h
```diff
@@ -0,0 +1,55 @@
...
+
+// `matched_data` contains any data that must be identical for a quorum of replicas
+// `rsi` contains replica specific information that was received for each replying replica.
+struct Reply {
+  Msg matched_data;
+  std::map<ReplicaId, Msg> rsi;
```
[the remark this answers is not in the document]
andrewjstone: It doesn't have to be, but I also don't have to implement a hash function for an unsorted map this way, and display is cleaner if ordered.

bftclient/include/bftclient/seq_num_generator.h
andrewjstone: Fixed.

bftclient/include/bftclient/seq_num_generator.h
andrewjstone: You can't use constexpr in member context apparently. I left it here just to leave it where it was prior and next to the other data members.

bftclient/src/bft_client.cpp
andrewjstone: I agree it's dangerous, but it's the least worst of the options IMO, given the current message implementation. We shouldn't be adding any new fields before we switch to a better implementation. If we do, we'll have to remember to fix this. I'm pretty sure we will, because tests will probably start failing like crazy. 

Also, thanks for pointing this out, since I actually had a TODO below about taking into account the spanContextSize and forgot to do that. Will fix that.

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,162 @@
...
+  header->spanContextSize = 0;  // TODO (AJS): FIXME
+  header->idOfClientProxy = client_id;
+  header->flags = flags;
+  header->reqSeqNum = config.sequence_number;
+  header->requestLength = msg.size();
+  header->timeoutMilli = config.timeout.count();
+  header->cid_length = config.correlation_id.size();
+
+  // Copy the correlation ID
+  std::memcpy(msg.data() + header_size, config.correlation_id.data(), config.correlation_id.size());
```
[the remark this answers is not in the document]
andrewjstone: I don't follow.

bftclient/src/bft_client.cpp
andrewjstone: Wow, I did not know that this use of const auto& actually extends the lifetime of the returned temporary. In this case I"m not sure it matters, as we'd get return value optimization anyway. But to make you happy I changed it. Thanks for pointing it out. 

For other reviewers see the section "Extending the lifetime of temporaries" here: https://crascit.com/2016/08/15/avoiding-copies-and-moves-with-auto/

bftclient/src/matcher.h
andrewjstone: fixed.

bftclient/include/bftclient/exception.h
andrewjstone: I just used an example from concord for this. Are you saying to just use a runtime_exception directly instead of a named exception, or inherit from std::runtime_error?

bftclient/src/bft_client.cpp
andrewjstone: I just updated this to include span context properly in the RequestConfig and put parameters in the right order. How does the new code look?

bftclient/include/bftclient/base_types.h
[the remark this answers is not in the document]
andrewjstone: This resulted in some extra casts due to the current interface of ICommunication, but I think it was worth it.

bftclient/src/bft_client.cpp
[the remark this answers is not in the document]
robem: I don't follow either - I must have misread the code. Please ignore.

bftclient/include/bftclient/bft_client.h
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  void Stop() { communication_->Stop(); }
+
+  // Send a message where the reply gets allocated by the callee and returned in a vector.
+  // The message to be sent is moved into the caller to prevent unnecessary copies.
+  //
+  // Throws a BftClientException on error.
+  Reply Send(const WriteConfig& config, Msg&& request);
+  Reply Send(const ReadConfig& config, Msg&& request);
```
dartdart26, line 53: Beautiful :)

bftclient/include/bftclient/config.h
```diff
@@ -0,0 +1,90 @@
...
+  ClientId id;
+  std::set<ReplicaId> all_replicas;
+  uint16_t f_val;
+  uint16_t c_val;
+  RetryTimeoutConfig retry_timeout_config;
+};
+
+// Generic per-request configuration shared by reads and writes.
+struct RequestConfig {
+  bool pre_execute;
```
dartdart26, line 70: Nit: Can `pre_execute` be defaulted to false?

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+#include "bftclient/bft_client.h"
+#include "bftengine/ClientMsgs.hpp"
+#include "assertUtils.hpp"
+
+namespace bft::client {
+
+// This function creates a ClientRequestMsg or a ClientPreProcessRequestMsg depending upon config.
+//
+// Since both of these are just instancds of a `ClientRequestMsgHeader` followed by the message
```
dartdart26, line 20: Typo: `instances`

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,165 @@
...
+  // Copy the correlation ID
+  std::memcpy(position, config.correlation_id.data(), config.correlation_id.size());
+
+  return msg;
+}
+
+Reply Client::Send(const WriteConfig& config, Msg&& request) {
+  Assert(!outstanding_request_.has_value());
+  auto match_config = WriteConfigToMatchConfig(config);
+  bool read_only = false;
```
dartdart26, line 62: Nit: Can we remove the `read_only` variable?

bftclient/include/bftclient/quorums.h
```diff
@@ -0,0 +1,82 @@
...
+
+// A quorum of F + 1 matching replies from `destination` must be received for the `send` call to complete.
+struct ByzantineSafeQuorum {
+  std::set<ReplicaId> destinations;
+};
+
+// A matching reply from every replica in `destination` must be received for the `send` call to complete.
+struct All {
+  std::set<ReplicaId> destinations;
+};
```
dartdart26, line 35: Great representation of these notions, makes it much more readable.

bftclient/src/matcher.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  if (!validSource(reply.rsi.from)) {
+    LOG_WARN(logger_, "Received reply from invalid source: " << reply.rsi.from.val);
+    return false;
+  }
+
+  return true;
+}
+
+bool Matcher::validSource(const ReplicaId& source) {
```
dartdart26, line 56: Can we make this method `const`?

bftclient/src/matcher.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+std::optional<Match> Matcher::match() {
+  auto result = std::find_if(matches_.begin(), matches_.end(), [this](const auto& match) {
+    return match.second.size() == config_.quorum.wait_for;
+  });
+  if (result == matches_.end()) return std::nullopt;
+  return Match{Reply{std::move(result->first.data), std::move(result->second)}, result->first.metadata.primary};
+}
+
+bool Matcher::valid(const UnmatchedReply& reply) {
```
dartdart26, line 40: Can we make this method `const`?

bftclient/src/bft_client.cpp
```diff
@@ -0,0 +1,165 @@
...
+  return Send(match_config, config.request, std::move(request), read_only);
+}
+
+Reply Client::Send(const MatchConfig& match_config,
+                   const RequestConfig& request_config,
+                   Msg&& request,
+                   bool read_only) {
+  outstanding_request_ = Matcher(match_config);
+  receiver_.activate(request_config.max_reply_size);
+  const auto& msg = MakeClientMsg(request_config, std::move(request), read_only, config_.id.val);
```
dartdart26, line 79: Nit: Might be a bit confusing to bind to a reference, could just do `const auto msg`.

bftclient/test/bft_client_test.cpp
```diff
@@ -0,0 +1,361 @@
...
+  ASSERT_EQ(match.value().primary.value(), primary);
+  ASSERT_EQ(3, match.value().reply.rsi.size());
+  for (auto i = 0u; i < match.value().reply.rsi.size(); i++) {
+    const auto& rsi_data = match.value().reply.rsi[ReplicaId{(uint16_t)i}];
+    Msg expected{(uint8_t)i};
+    ASSERT_EQ(expected, rsi_data);
+  }
+}
+
+TEST(matcher_tests, wait_for_3_out_of_4_with_mismatches_and_dupes) {
```
dartdart26, line 185: That's a great test!

bftclient/include/bftclient/bft_client.h
yuliasherman: Error responses could be non-retriable like conflicts detected during a post-execution stage. In our case, we retry all the requests no matter of what result they return. Just something to think about as a functionality improvement.

bftengine/include/bftengine/ClientMsgs.hpp
yuliasherman: You used rsi in Reply and all other places till now.

bftclient/include/bftclient/base_types.h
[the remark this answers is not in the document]
f-squirrel: If could have a ctor with default value:) Anyway, I see your point.

bftclient/include/bftclient/base_types.h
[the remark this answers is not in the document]
f-squirrel: Thanks!

bftclient/include/bftclient/exception.h
f-squirrel: Excise me for being ambiguous. I mean that `BftClientException` is in essence a runtime error.
Thus, I propose to inherit from `std:runtime_error`. As a bonus, there is no need to override `what` method.

bftclient/src/bft_client.cpp
f-squirrel: Since I've proposed to use the reference, I feel like, I need to answer this:)
Using the const reference guarantees that the bound' object lifetime is extended to the reference's lifetime.
Thus, we use the object by aliasing w/o copying or moving it.
While on the contrary, if `const auto` is used either copy or move happens.
The behavior is guaranteed by the standard since C++98(if I am not mistaken).

dartdart26: Oh, I haven't seen that you've discussed that, sorry.

You are right about lifetime extension. I think the reference will bind to a temporary, meaning that without RVO there will still be a move. And with RVO, there won't be any moves/copies. Having said that, I don't think ref/without reference will have any impact on the assembly. My comment was mainly focused on code style, i.e. when I see a reference, I wouldn't expect that this function returns a new object. Rather, I expect it returns a message that's already constructed. What I mean is I prefer to keep it simple in that case :) In general, not a big deal, we can leave as is.

In terms of assembly in O3, you can see they look identical: https://godbolt.org/z/qGF25B

bftclient/src/bft_client.cpp
f-squirrel: @andrewjstone, thank you for fixing it, IMO, it is a very elegant solution.

bftclient/src/bft_client.cpp
dartdart26: To be more precise, C++17's rules are a bit different from the C++14 ones, but I think the result is the same: https://godbolt.org/z/qD5-N8

https://en.cppreference.com/w/cpp/language/copy_elision

I am almost sure of @andrewjstone 's reaction and face expression when reading this... :)

f-squirrel: @dartdart26 , Interesting comparison!
Nit: If I am not mistaken, this case is not `RVO` but `NRVO`, where copy-elision is not guaranteed by the standard.

bftclient/include/bftclient/bft_client.h
andrewjstone: Sure. My point was that there's no way to know from the ClientReplyMsgHeader if a request actually had an error. It's in the serialized response. So an error would have to be returned from a quorum of nodes, which would match, and stop the retry. The caller to Client::Send would handle it, not the client itself.

bftclient/src/bft_client.cpp
andrewjstone: > Having said that, I don't think ref/without reference will have any impact on the assembly. My comment was mainly focused on code style, i.e. when I see a reference, I wouldn't expect that this function returns a new object. Rather, I expect it returns a message that's already constructed. What I mean is I prefer to keep it simple in that case :) In general, not a big deal, we can leave as is.

I actually agree with this sentiment. After reflecting, I think the `const auto&` makes the code less clear. I think I'm going to change it back in the next commit.

bftengine/include/bftengine/ClientMsgs.hpp
andrewjstone: Ah, yeah. I prefer the shorter form for most things, since it just makes the calling code shorter. I'm open to changing in another PR though.  I used the longer form in the message to match the existing style, and also because it's only used in one place in the client :)

bftclient/include/bftclient/bft_client.h
yuliasherman: My point is that we may want to add a new field to the ClientReplyMsgHeader specifying what the error was, at least if it is retriable or not.

andrewjstone: Ah okay! Definitely something to discuss later :)

andrewjstone: We'd still need to wait for at least F+1 of those same error responses though.

yuliasherman: Sure

bftclient/include/bftclient/exception.h
andrewjstone: Got it. Thanks! Will fix.

bftclient/src/bft_client.cpp
andrewjstone: fixex.

bftclient/src/bft_client.cpp
andrewjstone: I think it makes the code clearer. This is the problem with boolean parameters. You don't know what they mean at the call site.

bftclient/src/matcher.cpp
andrewjstone: done.

bftclient/src/matcher.cpp
andrewjstone: done.

bftclient/src/matcher.h
andrewjstone: camelCase won the vote.

bftclient/src/msg_receiver.cpp
andrewjstone: camelCase is now official :)

bftclient/include/bftclient/config.h
andrewjstone: done.

andrewjstone:
I resolved all comments and rebased. Also, we now have camelCase functions everywhere :)

PR #721, opened 2020-07-01
clang-tidy: additional bugprone checks

f-squirrel:
Checks added:
+bugprone-unused-return-value - [only for certain types of objects](https://clang.llvm.org/extra/clang-tidy/checks/bugprone-unused-return-value.html)
+bugprone-too-small-loop-variable
+bugprone-undefined-memory-manipulation
+bugprone-suspicious-string-compare
+bugprone-signed-char-misuse

For detailed information, please refer to https://clang.llvm.org/extra/clang-tidy/checks/list.html.

### DIFF
--- a/bftengine/tests/messages/NewViewMsg_test.cpp
+++ b/bftengine/tests/messages/NewViewMsg_test.cpp
@@ -9,6 +9,8 @@
 // these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
 // file.
 
+#include <limits>
+#include <random>
 #include "gtest/gtest.h"
 
 #include "helper.hpp"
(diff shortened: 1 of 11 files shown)


bftengine/tests/messages/NewViewMsg_test.cpp
```diff
@@ -32,15 +32,15 @@ TEST(NewViewMsg, base_methods) {
   EXPECT_EQ(msg.newView(), viewNum);
   EXPECT_EQ(msg.elementsCount(), 0);
 
-  for (uint8_t i = 1; i <= config.fVal * 2 + config.cVal * 2 + 1; ++i) {
+  for (uint8_t i = 1; i <= static_cast<uint8_t>(config.fVal * 2 + config.cVal * 2 + 1); ++i) {
```
andrewjstone, line 35: This is a super weird digest construction. I guess since it's only a test and we know the quorum size fits in 8 bits it's fine, but I'd still suggest changing how bytes are chosen rather than casting.

f-squirrel: In general, this test makes sure that any data stored in the message is retrievable.
The message itself does not alter or relate to the content of the digest.

Anyway, I agree that the casing looks ugly: fixed the casting and digest construction.

PR #726, opened 2020-07-02
Add bash-completion support to concord-bft build image

tdimitrov:
Install bash-completion and enable it for the root user. This makes the
interactive work with the container more easy. E.g. autocompletion for
ctest test cases.


tdimitrov:
The approach @f-squirrel suggested is far better than bloating the image. I'm closing the PR.

PR #730, opened 2020-07-03
Object store tests with minio

tdimitrov:
WIP PR for testing if the minio object store tests pass on CI. Please do not merge.


f-squirrel:
@tdimitrov, I know the PR is still in progress but if you submit it, could you please update the Makefile with `DUSE_S3_OBJECT_STORE=TRUE`?
I know it is very frustrating, I need to think about how to create a single source of compilation flags and docker image versions.

teoparvanov:
@tdimitrov I'd suggest to set the newly introduced `no_conflicts=True` flag in the `@verify_linearizability()` decorator, to avoid generating requests that contain conflicts that don't actually create blocks.

tdimitrov:
We hit a time limit on Travis if S3 tests are enabled. The solution I propose is to run only tests with direct memory storage on Travis and both types (Direct memory + S3) on Github Actions.
What do you think @toly-kournik

teoparvanov:
> 
> 
> We hit a time limit on Travis if S3 tests are enabled. The solution I propose is to run only tests with direct memory storage on Travis and both types (Direct memory + S3) on Github Actions.
> What do you think @toly-kournik

Hi @tdimitrov, I'd prefer to preserve the 1:1 match between GitHub Actions and Travis tests. To mitigate the time limitation, let's try running all RO tests only for n=7 (`selected_configs=lambda n, f, c: n == 7`).

.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
     - CMAKE_CXX_FLAGS="-DCMAKE_CXX_FLAGS_RELEASE=-O3 -g"
     - USE_LOG4CPP=-DUSE_LOG4CPP=TRUE
     - USE_ROCKSDB=-DBUILD_ROCKSDB_STORAGE=TRUE
-      #- USE_S3_OBJECT_STORE=-DUSE_S3_OBJECT_STORE=FALSE
+    - USE_S3_OBJECT_STORE=-DUSE_S3_OBJECT_STORE=OFF
```
teoparvanov, line 18: `-DUSE_S3_OBJECT_STORE=ON` - see my previous comment

tdimitrov: Okay, this is also a good approach. Let's see if they will pass this way.

Makefile
```diff
@@ -17,7 +17,7 @@ CONCORD_BFT_CMAKE_FLAGS:= \
 			-DCMAKE_CXX_FLAGS_RELEASE=-O3 -g \
 			-DUSE_LOG4CPP=TRUE \
 			-DBUILD_ROCKSDB_STORAGE=TRUE \
-			-DUSE_S3_OBJECT_STORE=FALSE \
+			-DUSE_S3_OBJECT_STORE=TRUE \
```
teoparvanov, line 20: Is `TRUE` the same as `ON` ? Let's unify this everywhere.

tdimitrov: It's the same. What is the preferred naming?
1, On, Yes, True or 0, Off, No, False.

I use ON/OFF in cmake related toggles, but I'm not sure it's an industry standard.

teoparvanov: Thanks for the clarification! I don't have a strong opinion about this, as long as the values are interpreted correctly.

PR #732, opened 2020-07-03
Introduce a sparse merkle tree DB inspection tool

dartdart26:
Introduce a sparse merkle tree DB inspection tool that allows fetching
block data and metadata from a RocksDB database. The tool works on the
command line and includes help output for the supported commands. Data
output is encoded in a simple JSON format. The tool is read-only and
never changes the database.

Rationale for this tool is that the sparse merkle storage is not
convenient for inspection using native RocksDB tools. Reason for that is
the complexity of tree storage on top of RocksDB. For example, the tree
has a number of different key types, constructs blocks on-demand and so
on.

Implement using the v2MerkleTree::DBAdapter class.

Add unit tests for the tool and the needed utility functions.

Example usage:
```
./sparse_merkle_db_inspector /rocksdb-path getGenesisBlockID
{
  "genesisBlockID": "1"
}
```


toly-kournik:
What a lovely std::visitor - no need in inheritance for commands!
Can we add an option to print the entire blockchain?

dartdart26:
> What a lovely std::visitor - no need in inheritance for commands!
> Can we add an option to print the entire blockchain?

Yes, will add support for the `getRawBlockRange` command that prints blocks in the `[start, end]` range.

PR #734, opened 2020-07-05
Add readonly mode support for DatabaseSet

salieri11:
This PR enables read only mode for any type of DatabaseSet. Especially it is critical when using rocksdb as underlying storage, for using concord DB from other utilities while the application is running.


salieri11:
After discussion with @dartdart26 , we decided not to implement RO support for the databaseset, since it implies that the underlying interfaces should not implement any write methods and it will require redesign which we are not going to do right now.

PR #738, opened 2020-07-07
[Apollo] Checkpoint propagation after primary isolation

shgandhi:
In this test, the goal is to verify checkpoint creation and propagation workflow when an adversary intercepts and drops all outgoing packets from the current primary.


teoparvanov:
PS: @shgandhi could you please re-trigger CI with a small commit, as it seems to be stuck? Thanks!

shgandhi:
Thanks for the review @teoparvanov. Re-triggered the CI with a small change.

teoparvanov:
Hi Shruti, I just merged your other test, so this PR needs to be rebased before we can merge it too.

shgandhi:
Hi Teo, this PR has been rebased.

PR #742, opened 2020-07-07
Add server side support for ReplicaSpecificInfo

andrewjstone:
Execution handlers can now return ReplicaSpecificInfo by specifying its
size and embedding it in the reply buffer. The field is already
supported in ClientReplyMsgHeaders and does not have to be used. Any
execution handler may choose to simply ignore it and let it default to
0.

ReplicaSpecificInfo is not currently supported or used for
pre-execution. It is also not currently used in our test binaries.
Nonetheless, the field is mandatory and was added to all usage of the
`execute` callback. A corresponding change will need to be made in the
product, and will be made before this is merged in.

Additionally, an unused code path related to direct proofs, which we do
not intend to support, was removed.

### DIFF
--- a/bftengine/include/bftengine/Replica.hpp
+++ b/bftengine/include/bftengine/Replica.hpp
@@ -44,6 +44,7 @@ class IRequestsHandler {
                       uint32_t maxReplySize,
                       char *outReply,
                       uint32_t &outActualReplySize,
+                      uint32_t &outReplicaSpecificInfoSize,
                       concordUtils::SpanWrapper &parent_span) = 0;
 
   virtual void onFinishExecutingReadWriteRequests() {}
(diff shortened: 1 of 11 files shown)


andrewjstone:
> Well done, Andrew!
> I have a few minor questions:
> 
> 1. If `ReplicaSpecificInfo` is embedded in the reply buffer as is, then I suppose it is supposed to be sent over the network as raw bytes.
>    Since `ReplicaSpecificInfo` contains `std::vector` which is not POD, I am not sure it is safe to embed it as is.
> 2. If BFT client receives a reply containing `ReplicaSpecificInfo`, is it supposed to remove it before returning the actual reply to a user?
> 
> Please correct me if I am wrong or miss something.

Thanks for looking @f-squirrel. 
 1. Replica Specific Info is just raw bytes that comes after the matching data in the same buffer. It's not a vector. It gets [extracted into a vector](https://github.com/vmware/concord-bft/blob/master/bftclient/src/msg_receiver.cpp#L73-L78) by the new client, but that is outside the scope of this change.
2. The client itself has no use for RSI, as it's application specific data which must be returned to the user of the client. The [Reply](https://github.com/vmware/concord-bft/blob/master/bftclient/include/bftclient/base_types.h#L48-L53) returns the matched data as well as unique RSI from all responding replicas in the quorum. The separation is to allow just quorum matching on the non-rsi data.

f-squirrel:
@andrewjstone, thank you for the clarification?

PR #744, opened 2020-07-08
[Apollo] Checkpoint propagation after f non primary replicas isolated

shgandhi:
The goal of this test is to verify checkpoint creation and propagation while an adversary isolates f non primary replicas.


teoparvanov:
This PR also needs rebasing before being merged.

shgandhi:
Hi Teo, rebased the PR.

PR #749, opened 2020-07-10
Add metrics to new client

andrewjstone:


bftclient/include/bftclient/metrics.h
```diff
@@ -0,0 +1,43 @@
...
+// file.
+
+#pragma once
+
+#include "Metrics.hpp"
+#include "base_types.h"
+
+namespace bft::client {
+
+struct Metrics {
```
teoparvanov, line 19: Just curious - why `struct`, given that you still define member variable visibility?

andrewjstone: No great reason. It's what I initially had before adding various visibility specifiers and reordering things. One decent reason though is intention: I personally view structs as meaning we want to access fields directly without getters/setters. Whether this distinction actually makes sense is questionable. However, the Google C++ style guide seems to agree since trailing underscores aren't used in struct members, which is what I wanted.

teoparvanov: Thanks for the clarification!

PR #750, opened 2020-07-12
Increase default rocksdb memtable size

salieri11:
This PR increases default rocksdb memtable size from 64mb to 512mb. As was observed in performance tests, this change is essential for high loads.


salieri11:
Thanks @teoparvanov !

PR #754, opened 2020-07-14
Add controlStateManager via the bft layers

yontyon:


bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -0,0 +1,42 @@
...
+  virtual bool saveControlState(const ControlStateId id, const std::string& controlState) = 0;
+  virtual std::string getControlState(const ControlStateId id) = 0;
+  virtual ~IControlStateManager() = default;
+};
+
+class InMemoryControlStateManager : public IControlStateManager {
+  std::unordered_map<ControlStateId, std::string> controlStateMap_;
+
+ public:
+  virtual bool saveControlState(const ControlStateId id, const std::string& controlState) override {
```
f-squirrel, line 35: Nit:
`virtual` is obsolete because of the `override`.
`const` is obsolete because `id` is passed by value.

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace bftEngine {
+
+class IControlStateManager {
+ public:
+  enum ControlStateId {
+    STOP_AT_NEXT_CHECKPOINT,
+  };
+
+  virtual bool saveControlState(const ControlStateId id, const std::string& controlState) = 0;
```
f-squirrel, line 26: Is `bool` return value a placeholder for the future implementations of the state manager that may not be able to save the state?

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+class InMemoryControlStateManager : public IControlStateManager {
+  std::unordered_map<ControlStateId, std::string> controlStateMap_;
+
+ public:
+  virtual bool saveControlState(const ControlStateId id, const std::string& controlState) override {
+    controlStateMap_[id] = controlState;
+    return true;
+  }
+  virtual std::string getControlState(const ControlStateId id) override { return controlStateMap_[id]; }
```
f-squirrel, line 39: If the `id` does not exist an empty string is returned.
Returning `std::optional` or throwing an exception in case of error may be a reasonable alternative.

bftengine/include/bftengine/Replica.hpp
```diff
@@ -48,6 +49,7 @@ class IRequestsHandler {
                       concordUtils::SpanWrapper &parent_span) = 0;
 
   virtual void onFinishExecutingReadWriteRequests() {}
+  virtual void setControlStateManager(std::shared_ptr<IControlStateManager> controlStateManager) {}
```
f-squirrel, line 52: If the class is abstract, why don't we make the method abstract too?
The same applies to `IReplica`.

bftengine/include/bftengine/Replica.hpp
```diff
@@ -75,6 +77,7 @@ class IReplica {
   // TODO(GG) : move the following methods to an "advanced interface"
   virtual void SetAggregator(std::shared_ptr<concordMetrics::Aggregator> a) = 0;
   virtual void restartForDebug(uint32_t delayMillis) = 0;  // for debug only.
+  virtual void setControlStateManager(std::shared_ptr<IControlStateManager> controlStateManager){};
```
f-squirrel, line 80: In general, `shared_ptr` makes ownership and lifetime of objects unclear. Do we really need it to be shared?
As far as I see, replica owns the manager and passes it to the command handlers.
If so, the manager may be moved to the replica and passed to the command handlers as reference(the lifetime of the handlers is supposed to be shorter than replicas so we're safe, I am not mistaken).
If so, there is no need to use pointers(only delete copy-constructors and operators of the manager base class).

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -0,0 +1,42 @@
...
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+#include <string>
+#include <unordered_map>
+
+namespace bftEngine {
+
+class IControlStateManager {
```
andrewjstone, line 20: I don't see a good reason to start off with an abstract interface here. It seems like premature abstraction to me. IMO there should only be a single, concrete, `ControlStateManager`, that is responsible for managing control state across reserved pages. Furthermore a single replica can only take a single instance of the `ControlStateManager`, so any functionality needed across components will have to use the same instance. The manager itself can allow plugins, but again this seems premature to me. We should start with the simplest thing and then get more complicated if we need it.

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -0,0 +1,42 @@
...
+ public:
+  enum ControlStateId {
```
andrewjstone, line 22: If we make this a [std::variant](https://en.cppreference.com/w/cpp/utility/variant) we don't need separate `ControlStateId`s and values, and we wouldn't need to tie our value type to a string, which likely doesn't make much sense for many uses of this. We could use the [index of the variant](https://en.cppreference.com/w/cpp/utility/variant/index) as the id for serialization into reserved pages.

An example is something like this:

```c++
struct StopAtNextCheckpoint {};
struct RotateKeys{
    std::map<..> newKeys;
}

using ControlStateOperation = std::variant<StopAtNextCheckpoint, RotateKeys>;

using CurrentCheckpointForUpgrade = std::optional<uint64_t>;
using CurrentKeys = std::map<..>;

using ControlState = std::variant<CurrentCheckpointForUpgrade, CurrentKeys>;

// Throws on Error
void setControlState(ControlStateOperation op); 

// We use the index of the variant for retrieval. 
ControlState getControlState(std::size_t index);
```

While the above matches some future plans I have for message based systems, it seems unnecessarily complex for now. Equally unnecessary, would be to use virtual types for operations and states, where each instantiation of `IControlStateOperation` or `IControlState` has a unique ID and can be registered.

For now, since we are pretty limited in what we are using this for, and will likely evolve it in the future, why not just do the simplest possible thing and provide a concrete type with functions for each state? The internal implementation can map the functions to ids for serialization. Something like this:

```
class ControlStateManager {
  public:
   void stopAtNextCheckpoint();
   std::optional<uint64_t> getUpgradeCheckpoint();
   void SetKeys(std::map<..> Keys);
   std::map<..> getKeys();
   
 private:
  ControlState some_state_structure_with_ids_that_lives_in_reserved_pages_;
  IStateTransfer& reserved_pages_api_;
};
```
If you really needed to, you could make this interface abstract, to allow you to use a test version, but that maybe overkill as well. Instead the single, concrete, `ControlStateManager` would take the reserved pages interface as an abstract interface, which we already have defined in [IStateTransfer](https://github.com/vmware/concord-bft/blob/master/bftengine/include/bftengine/IStateTransfer.hpp#L52-L57)

bftengine/include/bftengine/ControlStateManager.hpp
yontyon: The virtual class has been removed

bftengine/include/bftengine/ControlStateManager.hpp
yontyon: The virtual class has been removed

bftengine/include/bftengine/ControlStateManager.hpp
yontyon: removed

bftengine/include/bftengine/Replica.hpp
yontyon: There are some derived classes that don't need to implement this method, I prefer to give a default empty implementation than forcing every derived class to implement this method

bftengine/include/bftengine/Replica.hpp
yontyon: I was trying to use references instead of shared_ptrs but it looks like that using references complicates the code a lot.
Explanation:
The state transfer is initiated in the higher level of the kvbc replica, then we need to pass it through all bft layers and interfaces to the replicaBase and the implementations of the command handler.
If I use references I have to pass the controlStateManager through the constructors, which may complicate the current code flow, and also may depend on the order we initiate the objects (replica and command handler) in Concord (the product).
Thus, I prefer to use shared_ptr and getters and setters over references.

bftengine/include/bftengine/ControlStateManager.hpp
yontyon: The abstract class has been removed

bftengine/include/bftengine/ControlStateManager.hpp
yontyon: I built a basic nonabstract implementation with a method for each controlState (as you suggested) based on the state transfer module.

For now the implementation is empty until @toly-kournik will finish his adaptions for the reserved pages and the state transfer module which will also answer our needs for allocating specific pages for the controlState data

bftengine/include/bftengine/Replica.hpp
f-squirrel: If some derived classes do not need the method then the method is not supposed to be declared in the base class.
Adding a method to the interface class is a very strong statement.
Can it be added only to specific implementations?

bftengine/include/bftengine/Replica.hpp
f-squirrel: I understand the complications but the question is if state transfer and command handler can exist w/o replace?
No pressure but I am really trying to understand it from functional perspective.

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -0,0 +1,31 @@
...
+// file.
+
+#pragma once
+#include <string>
+#include <unordered_map>
+#include "IStateTransfer.hpp"
+
+namespace bftEngine {
+
+class ControlStateManager {
```
f-squirrel, line 21: Do we want to have the ability to copy the instance of the object?
I mean something like:
```
std::shared_ptr< ControlStateManager> m;
auto m1 = *m;
```
If not, it might be better to delete copy/assign operators.

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -0,0 +1,31 @@
...
+ public:
+  void setStopAtNextCheckpoint();
+  bool getStopAtNextCheckpoint();
```
andrewjstone, line 24: Are we really just setting a flag and then getting whether the flag is set? That's what these names read like to me. I think it makes a bit more sense to indicate what they are actually doing. I'd prefer to rename them and provide some extra functionality.

`void stopAtNextCheckpoint(); ` actually calculates the checkpoint to stop at and writes that value into a reserved page.
`std::optional<uint64_t> getNextCheckpointToStopAt` returns the checkpoint number to stop at or `std::nullopt` if it's not set.

bftengine/include/bftengine/ControlStateManager.hpp
andrewjstone: Thank you!

bftengine/include/bftengine/Replica.hpp
yontyon: You right, but the problem is that all upper-level references/pointers are for the base class.
I can make this method abstract of course and add an empty implementation where needed, but I'm afraid it may break API with concord and things like that...
Do you think we should do it in any case?

bftengine/include/bftengine/ControlStateManager.hpp
yontyon: done

bftengine/include/bftengine/Replica.hpp
f-squirrel: As far as I see the method `setControlStateManager` is used only by the classes derived from `ICommandsHandler`.
I think it would be right to add move the method there and declare it abstract.
Regarding the product, I think the only class to be changed is `ConcordCommandsHandler`.
WDYT?

PR #757, opened 2020-07-15
Replace assert with Assert

f-squirrel:
Rationale:
The standard `assert` is ignored if `NDEBUG` compilation flag is set.
Since Concord-BFT uses the `assert` extensively to ensure the consistency of the flow, ignoring it is unacceptable.

Changes:
* Replaced usage of `assert` with `Assert` all over the code base.
* Replaced `assert(false)` with `std::terminate()` in the assertUtils.hpp file

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,6 +2,7 @@
 Checks:             '-*,bugprone-*,
                     -bugprone-branch-clone,
                     -bugprone-exception-escape,
+                    -bugprone-lambda-function-name,
                     -bugprone-misplaced-widening-cast,
                     -bugprone-not-null-terminated-result,
                     performance-*,
@@ -13,6 +14,7 @@ Checks:             '-*,bugprone-*,
 WarningsAsErrors:   'bugprone-*,
                     -bugprone-branch-clone,
                     -bugprone-exception-escape,
+                    -bugprone-lambda-function-name,
                     -bugprone-misplaced-widening-cast,
                     -bugprone-not-null-terminated-result,
                     performance-*,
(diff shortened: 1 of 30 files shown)


yuliasherman:
A long time ago I did this change for all relevant concord-bft files, except tests. From then this assert appeared again - in the new code (and one moved from concord).

f-squirrel:
@yuliasherman, very good point!
I can ensure via CI that no default asserts added to the code.
Of course, if everybody is ok with it.

andrewjstone:
> A long time ago I did this change for all relevant concord-bft files, except tests. From then this assert appeared again - in the new code (and one moved from concord).

This one hits home. One time I spent two weeks fixing all non-determinism in tests at a prior company, and cleaning up all sleeps. It took less time than that for things to break again.

andrewjstone:
> @yuliasherman, very good point!
> I can ensure via CI that no default asserts added to the code.
> Of course, if everybody is ok with it.

I think this is a great idea Dima! I think we should do it for other simple things as well if we can. First up, use. of sleeps in test code :)

f-squirrel:
@yuliasherman , @andrewjstone , I've added another commit that ensures that `assert` is not used in the code.
Unfortunately doing the same for `sleep*` is more complex.

andrewjstone:
> @yuliasherman , @andrewjstone , I've added another commit that ensures that `assert` is not used in the code.
> Unfortunately doing the same for `sleep*` is more complex.

I figured :) Thanks for doing this. We can worry about enhancements in the future.

PR #760, opened 2020-07-16
WIP: Initial efforts for sending noops commands

yontyon:
In this PR we add a basic functionality for creating no-ops commands by the replica itself.

This can be helpful for cases in which the primary decides to bring the system for the next checkpoint as soon as possible (for example, when realizing that an upgrade process is about to start).

The way it works is very simple:
Once the primary decides to "jump" the system to the next checkpoint it creates empty preprepare messages and initiates consensus for each one of them.

We may consider other solutions (such as filling the client requests queue with fake messages) but I believe that the above is the simplest and fastest solution.

EDITING:
I've also added call to the method that fills the working window with noops.
Note that currently, if we have viewchange while sending this noops the new primary won't necessarily continue to send them.
Yet, IMO as we assume the happy path + our very initial stage of development, we can leave it as it is for now.

### DIFF
--- a/bftengine/src/bftengine/ReplicaBase.hpp
+++ b/bftengine/src/bftengine/ReplicaBase.hpp
@@ -128,7 +128,7 @@ class ReplicaBase {
   ///////////////////////////////////////////////////
   // ControlStateManger
   std::shared_ptr<ControlStateManager> controlStateManager_;
-  bool stopAtNextChecnpoint = false;
+  bool stopAtNextChecknpoint_ = false;
 };
 
 }  // namespace bftEngine::impl
--- a/bftengine/src/bftengine/ReplicaBase.hpp
+++ b/bftengine/src/bftengine/ReplicaBase.hpp
@@ -128,7 +128,7 @@ class ReplicaBase {
   ///////////////////////////////////////////////////
   // ControlStateManger
   std::shared_ptr<ControlStateManager> controlStateManager_;
-  bool stopAtNextChecknpoint_ = false;
+  bool stopAtNextCheckpoint_ = false;
 };
 
 }  // namespace bftEngine::impl
--- a/bftengine/src/bftengine/ReplicaBase.hpp
+++ b/bftengine/src/bftengine/ReplicaBase.hpp
@@ -128,6 +128,7 @@ class ReplicaBase {
   ///////////////////////////////////////////////////
   // ControlStateManger
   std::shared_ptr<ControlStateManager> controlStateManager_;
+  bool stopAtNextChecnpoint = false;
 };
 
 }  // namespace bftEngine::impl
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -419,6 +419,10 @@ void ReplicaImp::sendInternalNoopsPrePrepareMsg(CommitPath firstPath) {
 
 void ReplicaImp::bringTheSystemToTheNextCheckpointBySendingNoopsCommands(CommitPath firstPath) {
   if (!isCurrentPrimary()) return;
+  // TODO: According to Ittai, it is better to reach to the next next checkpoint to prevent the follwing:
+  // 1. The current sequence number is 290 (next checkpoint is 300)
+  // 2. We decide on 291 --> 291 + concurrency level - 1 (say 29)
+  // 3. We decide on 290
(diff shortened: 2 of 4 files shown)


yontyon:
I manually tested it (by adding an artificial call to `bringTheSystemToTheNextCheckpointBySendingNoopsCommands` on start). I was able to see that indeed the system process 150 "empty" requests, creates a checkpoint and only then it start to process the incoming client requests.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -370,7 +370,12 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
   }
 
   pp->finishAddingRequests();
+  startConsensusProcess(pp);
```
andrewjstone, line 373: What is the purpose of this change? It seems unrelated to noops as far as I can tell.

yontyon: I wanted to create a single point where we send a pre prepare message and update all the counters, persist the data etc.

once we have this method we can create an empty pp message (as in viewchange) and start a consensus with it - and let the existing code to safely do its work.
Assume we in sequence number No. 140, so sending 10 such empty preprepare messages will bring the system to sequence number 150 and to the nearest checkpoint.

andrewjstone: Thanks for the explanation. Makes sense.

bftengine/src/bftengine/ReplicaBase.hpp
```diff
@@ -128,6 +128,7 @@ class ReplicaBase {
   ///////////////////////////////////////////////////
   // ControlStateManger
   std::shared_ptr<ControlStateManager> controlStateManager_;
+  bool stopAtNextChecnpoint = false;
```
robem, line 131: `stopAtNextCheckpoint` and trailing underscore missing.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -407,6 +412,22 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
   }
 }
 
+void ReplicaImp::sendInternalNoopsPrePrepareMsg(CommitPath firstPath) {
+  PrePrepareMsg *pp = new PrePrepareMsg(config_.replicaId, curView, (primaryLastUsedSeqNum + 1), firstPath, 0);
+  startConsensusProcess(pp);
+}
+
+void ReplicaImp::bringTheSystemToTheNextCheckpointBySendingNoopsCommands(CommitPath firstPath) {
```
robem, line 420: Similar `NoopCommands`

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -407,6 +412,22 @@ void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
   }
 }
 
+void ReplicaImp::sendInternalNoopsPrePrepareMsg(CommitPath firstPath) {
```
robem, line 415: This is one noop command, correct? Then we should get rid of the plural.

bftengine/src/bftengine/ReplicaBase.hpp
yontyon: fixed

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: fixed

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: fixed

bftengine/src/bftengine/ReplicaBase.hpp
```diff
@@ -128,7 +128,7 @@ class ReplicaBase {
   ///////////////////////////////////////////////////
   // ControlStateManger
   std::shared_ptr<ControlStateManager> controlStateManager_;
-  bool stopAtNextChecnpoint = false;
+  bool stopAtNextChecknpoint_ = false;
```
robem, line 131: `stopAtNextCheck*n*point_` :)

yontyon: done

PR #761, opened 2020-07-17
[Apollo] Checkpoint propagation after restarting all replicas

shgandhi:
To test backup and restore, we trigger a checkpoint and then restart all replicas in a random order with 10s delay in-between. We then verify the first checkpoint persisted upon restart and then trigger another checkpoint. We make sure the second checkpoint is propagated to all the replicas and the initial view is stable.

### DIFF
--- a/tests/apollo/test_skvbc_backup_restore.py
+++ b/tests/apollo/test_skvbc_backup_restore.py
@@ -1,0 +1,117 @@
+# Concord
+#
+# Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+#
+# This product is licensed to you under the Apache 2.0 license (the "License").
+# You may not use this product except in compliance with the Apache 2.0 License.
+#
+# This product may include a number of subcomponents with separate copyright
+# notices and license terms. Your use of these subcomponents is subject to the
+# terms and conditions of the subcomponent's license, as noted in the LICENSE
+# file.
+import os.path
+import random
+import unittest
+
+import trio
+
+from util import skvbc as kvbc
+from util.bft import with_trio, with_bft_network, with_constant_load, KEY_FILE_PREFIX
+
+
+def start_replica_cmd(builddir, replica_id):
+    """
+    Return a command that starts an skvbc replica when passed to
+    subprocess.Popen.
+
+    The replica is started with a short view change timeout and with RocksDB
+    persistence enabled (-p).
+
+    Note each arguments is an element in a list.
+    """
+    statusTimerMilli = "500"
+    viewChangeTimeoutMilli = "10000"
+    path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
+    return [path,
+            "-k", KEY_FILE_PREFIX,
+            "-i", str(replica_id),
+            "-s", statusTimerMilli,
+            "-v", viewChangeTimeoutMilli,
+            "-p",
+            "-t", os.environ.get('STORAGE_TYPE')]
+
+
+class SkvbcBackupRestoreTest(unittest.TestCase):
+
+    __test__ = False  # so that PyTest ignores this test scenario
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_checkpoint_propagation_after_restarting_replicas(self, bft_network):
+        """
+        Here we trigger a checkpoint, restart all replicas in a random order with 10s delay in-between,
+        both while stopping and starting. We verify checkpoint persisted upon restart and then trigger
+        another checkpoint. We make sure checkpoint is propagated to all the replicas.
+        1) Given a BFT network, we make sure all nodes are up
+        2) Send sufficient number of client requests to trigger checkpoint protocol
+        3) Stop all replicas in a random order (with 10s delay in between)
+        4) Start all replicas in a random order (with 10s delay in between)
+        5) Make sure the initial view is stable
+        6) Send sufficient number of client requests to trigger another checkpoint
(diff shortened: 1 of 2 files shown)


shgandhi:
@andrewjstone, @HristoStaykov can you please review this PR? Thanks!

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -0,0 +1,134 @@
...
+
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+
+        await skvbc.fill_and_wait_for_checkpoint(
+            initial_nodes=bft_network.all_replicas(),
+            checkpoint_num=1,
+            verify_checkpoint_persistency=False
+        )
+
+        # verify checkpoint propagation to all the nodes
```
andrewjstone, line 81: I don't think you need this. [This call](https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/skvbc.py#L229), made as part of `skvbc.fill_and_wait_for_checkpoint`, already waits for all replicas to reach the checkpoint.

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -0,0 +1,134 @@
...
+                expected_checkpoint_num=checkpoint_before + 1)
+            self.assertEqual(first_checkpoint_after, checkpoint_before + 1)
+
+        # stop n replicas in a random order with a delay of 10s in between
+        stopped_replicas = await self._stop_random_replicas_with_delay(bft_network, n)
+
+        # start stopped replicas in a random order with a delay of 10s in between
+        await self._start_random_replicas_with_delay(bft_network, stopped_replicas)
+
+        # verify checkpoint persisted in all the nodes after restart
```
andrewjstone, line 94: You can call `bft_network.wait_for_replicas_to_checkpoint(bft_network.all_replicas(), checkpoint_before + 1)` to run the checks in parallel rather than having to write the loop manually.

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -0,0 +1,134 @@
...
+
+        self.assertEqual(len(bft_network.procs), n, "Make sure all replicas are up initially.")
+
+        current_primary = await bft_network.get_current_primary()
+
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+
+        await skvbc.fill_and_wait_for_checkpoint(
+            initial_nodes=bft_network.all_replicas(),
+            checkpoint_num=1,
```
andrewjstone, line 77: Not directly related to this review, but we really need to change the name of this parameter to "num_checkpoints". I keep reading it as the specific checkpoint number which makes it very confusing.

tests/apollo/test_skvbc_backup_restore.py
shgandhi: Thanks @andrewjstone for the review. Yes, that's true. The reason I have done it this way is to verify checkpoint creation using assertEqual in the loop.

tests/apollo/test_skvbc_backup_restore.py
shgandhi: Done for the same reason as above.

tests/apollo/test_skvbc_backup_restore.py
andrewjstone: You won't even get to this point unless that condition is true though.

shgandhi: `wait_for_replicas_to_checkpoint` does not take `expected_checkpoint_num` as argument. Hence, the underlying call to `wait_for_checkpoint` can return with only `last_stored_checkpoint`, without waiting for replica to reach the expected checkpoint number.

shgandhi: @andrewjstone, updated as discussed.

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -0,0 +1,118 @@
...
+            "-p",
+            "-t", os.environ.get('STORAGE_TYPE')]
+
+
+class SkvbcBackupRestoreTest(unittest.TestCase):
+
+    __test__ = False  # so that PyTest ignores this test scenario
+
+    @with_trio
+    @with_bft_network(start_replica_cmd)
```
teoparvanov, line 49: given the long delays between replica restarts, let's run this test only for `n==7`

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Here we trigger a checkpoint, restart all replicas in a random order with 10s delay in-between,
+        both while stopping and starting. We verify checkpoint persisted upon restart and then trigger
+        another checkpoint. We make sure checkpoint is propagated to all the replicas.
+
+        1) Given a BFT network, we make sure all nodes are up
+        2) Send sufficient number of client requests to trigger checkpoint protocol
+        3) Stop all replicas in a random order (with 10s delay in between)
+        4) Start all replicas in a random order (with 10s delay in between)
+        5) Make the initial view is stable
```
teoparvanov, line 60: nit - "Make sure"

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -0,0 +1,118 @@
+# Concord
+#
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
```
teoparvanov, line 3: another nit - should be 2020

shgandhi:
> LGTM, thanks Shruti - I have some minor comments before we can merge this.

Thanks @teoparvanov. Updated as suggested.

PR #764, opened 2020-07-19
Change stack allocation to heap allocation

Liorgal28:
There are few places on the code that uses alloca to allocate memory on the stack.
According to https://jira.eng.vmware.com/browse/BC-3446 it's not safe to use it.
I changed the code to allocate this memory on the heap(after consulting with Guy) using a vector instead of a raw pointer.

### DIFF
--- a/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
+++ b/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
@@ -351,7 +351,7 @@ class CollectorOfThresholdSignatures {
       // TODO(GG): can utilize several threads (discuss with Alin)
 
       const uint16_t bufferSize = (uint16_t)verifier->requiredLengthForSignedData();
-      std::vector<char> const bufferForSigComputations(bufferSize);
+      std::vector<char> bufferForSigComputations(bufferSize);
 
       const auto& span_context_of_last_message =
           (reqDataItems - 1) ? sigDataItems[reqDataItems - 1].span_context : std::string{};
@@ -364,7 +364,7 @@ class CollectorOfThresholdSignatures {
 
         acc->setExpectedDigest(reinterpret_cast<unsigned char*>(expectedDigest.content()), DIGEST_SIZE);
 
-        acc->getFullSignedData(const_cast<char*>(bufferForSigComputations.data()), bufferSize);
+        acc->getFullSignedData(bufferForSigComputations.data(), bufferSize);
 
         verifier->release(acc);
       }
--- a/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
+++ b/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
@@ -351,7 +351,7 @@ class CollectorOfThresholdSignatures {
       // TODO(GG): can utilize several threads (discuss with Alin)
 
       const uint16_t bufferSize = (uint16_t)verifier->requiredLengthForSignedData();
-      char* const bufferForSigComputations = (char*)alloca(bufferSize);  // TODO(GG): check
+      std::vector<char> const bufferForSigComputations(bufferSize);
 
       const auto& span_context_of_last_message =
           (reqDataItems - 1) ? sigDataItems[reqDataItems - 1].span_context : std::string{};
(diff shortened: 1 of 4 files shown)


bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
```diff
@@ -351,7 +351,7 @@ class CollectorOfThresholdSignatures {
       // TODO(GG): can utilize several threads (discuss with Alin)
 
       const uint16_t bufferSize = (uint16_t)verifier->requiredLengthForSignedData();
-      char* const bufferForSigComputations = (char*)alloca(bufferSize);  // TODO(GG): check
+      std::vector<char> const bufferForSigComputations(bufferSize);
```
tdimitrov, line 354: The const used here is for the pointer itself, not for the memory, which it points to. Remove the const, because it makes the vector itself const.

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
```diff
@@ -364,12 +364,12 @@ class CollectorOfThresholdSignatures {
 
         acc->setExpectedDigest(reinterpret_cast<unsigned char*>(expectedDigest.content()), DIGEST_SIZE);
 
-        acc->getFullSignedData(bufferForSigComputations, bufferSize);
+        acc->getFullSignedData(const_cast<char*>(bufferForSigComputations.data()), bufferSize);
```
tdimitrov, line 367: If you remove the const on line 354, you won't need this cast.

bftengine/src/bftengine/ViewsManager.cpp
```diff
@@ -341,7 +341,7 @@ void ViewsManager::computeCorrectRelevantViewNumbers(ViewNum* outMaxKnownCorrect
   *outMaxKnownAgreedView = myLatestPendingView;
 
   size_t numOfVC = 0;
-  ViewNum* viewNumbers = reinterpret_cast<ViewNum*>(alloca(N * sizeof(ViewNum)));
+  std::vector<ViewNum> viewNumbers(N * sizeof(ViewNum));
```
tdimitrov, line 344: Here the parameter should be just N instead of sizeof(ViewNum). Alloca accepts size in bytes, while the vector - element count.

bftengine/src/bftengine/messages/PartialProofsSet.cpp
```diff
@@ -215,7 +215,7 @@ class AsynchProofCreationJob : public util::SimpleThreadPool::Job {
 
     auto span = concordUtils::startChildSpanFromContext(span_context_, "bft_create_FullCommitProofMsg");
     const uint16_t bufferSize = (uint16_t)verifier->requiredLengthForSignedData();
-    char* const bufferForSigComputations = (char*)alloca(bufferSize);
+    std::vector<char> const bufferForSigComputations(bufferSize);
```
tdimitrov, line 218: Same about the const.

bftengine/src/bftengine/messages/PartialProofsSet.cpp
```diff
@@ -227,18 +227,23 @@ class AsynchProofCreationJob : public util::SimpleThreadPool::Job {
       return;
     }
 
-    acc->getFullSignedData(bufferForSigComputations, sigLength);
+    acc->getFullSignedData(const_cast<char*>(bufferForSigComputations.data()), sigLength);
```
tdimitrov, line 230: This cast won't be needed after removing the const on 218.

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
Liorgal28: Fixed.

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
Liorgal28: Fixed.

bftengine/src/bftengine/ViewsManager.cpp
Liorgal28: Fixed.

bftengine/src/bftengine/messages/PartialProofsSet.cpp
Liorgal28: Fixed.

bftengine/src/bftengine/messages/PartialProofsSet.cpp
Liorgal28: Fixed.

PR #765, opened 2020-07-19
WIP: Add control handlers for upgrade + calling to upgrade on super stable checkpoint

yontyon:
In this PR we have three parts:
1. More functionality for checkpointInfo
2. Callbacks interface to IrequestHandler for control handlers and particular for an upgrade.
3. Identify super stable checkpoint and invoke the upgrade callback if exist

For part 3 we did the following:
We increased the size of the checkpoint log in one and instead of deleting the new lastStableCheckpoint from the log we delete the previous lastStableCheckpoint (if exist). This way we give time for the previous stableCheckpoint to become super stable (and this time is the time between two consecutive stable sequence numbers).

### DIFF
--- a/bftengine/include/bftengine/Replica.hpp
+++ b/bftengine/include/bftengine/Replica.hpp
@@ -35,6 +35,11 @@ enum MsgFlag : uint8_t {
   HAS_PRE_PROCESSED_FLAG = 0x4
 };
 
+class IControlHandler {
+ public:
+  virtual void upgrade() = 0;
+};
+
 class IRequestsHandler {
  public:
   virtual int execute(uint16_t clientId,
@@ -50,6 +55,8 @@ class IRequestsHandler {
 
   virtual void onFinishExecutingReadWriteRequests() {}
   virtual ~IRequestsHandler() {}
+
+  virtual std::shared_ptr<IControlHandler> getControlHandlers() = 0;
 };
 
 class IReplica {
--- a/bftengine/include/bftengine/Replica.hpp
+++ b/bftengine/include/bftengine/Replica.hpp
@@ -35,9 +35,21 @@ enum MsgFlag : uint8_t {
   HAS_PRE_PROCESSED_FLAG = 0x4
 };
 
-class IControlHandler {
+// The ControlHandlers is a group of method that enables the userRequestHandler to perform infrastructure
+// changes in the system.
+// For example, assuming we want to upgrade the system to new software version, then:
+// 1. We need to bring the system to a stable state (bft responsibility)
+// 2. We need to perform the actual upgrade process (the platform responsibility)
+// Thus, once the bft brings the system to the desired stable state, it needs to invoke the a callback of the user to
+// perform the actual upgrade.
+// More possible scenarios would be:
+// 1. Adding/removing node
+// 2. Key exchange
+// 3. Change DB scheme
+// and basically any management action that is handled by the layer that uses concord-bft.
+class ControlHandlers {
  public:
-  virtual void upgrade() = 0;
+  virtual void onSuperStableCheckpoint() = 0;
 };
 
 class IRequestsHandler {
@@ -56,7 +68,7 @@ class IRequestsHandler {
   virtual void onFinishExecutingReadWriteRequests() {}
   virtual ~IRequestsHandler() {}
 
-  virtual std::shared_ptr<IControlHandler> getControlHandlers() = 0;
+  virtual std::shared_ptr<ControlHandlers> getControlHandlers() = 0;
 };
 
 class IReplica {
--- a/bftengine/src/bftengine/CheckpointInfo.hpp
+++ b/bftengine/src/bftengine/CheckpointInfo.hpp
@@ -42,6 +42,9 @@ class CheckpointInfo {
 
   bool isCheckpointCertificateComplete() const;
 
+  // A replica considers a checkpoint to be super stable if it knows that all n/n replicas have reached to this
+  // checkpoint. This is in contrary to stable checkpoint which means that the replica knows that a byzantine quorum of
+  // replicas have reached to this checkpoint.
   bool isCheckpointSuperStable() const;
 
   CheckpointMsg* selfCheckpointMsg() const;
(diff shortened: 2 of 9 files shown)


bftengine/src/bftengine/CheckpointInfo.hpp
```diff
@@ -42,6 +42,8 @@ class CheckpointInfo {
 
   bool isCheckpointCertificateComplete() const;
 
+  bool isCheckpointSuperStable() const;
```
andrewjstone, line 45: Can you please add a comment indicating what it means for a checkpoint to be super stable?

bftengine/include/bftengine/Replica.hpp
```diff
@@ -35,6 +35,11 @@ enum MsgFlag : uint8_t {
   HAS_PRE_PROCESSED_FLAG = 0x4
 };
 
+class IControlHandler {
```
andrewjstone, line 38: Why does this need to be an abstract class? 

I ask because it looks like the handlers are directly included (composition) via the functions.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2308,6 +2317,11 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
   }
 }
 
+void ReplicaImp::onSeqNumIsSuperStable(SeqNum newSuperStableSeqNum) {
+  if (controlStateManager_->getStopCheckpointToStopAt() == newSuperStableSeqNum) {
+    if (userRequestsHandler->getControlHandlers()) userRequestsHandler->getControlHandlers()->upgrade();
```
andrewjstone, line 2322: The abstraction doesn't seem quite right here. If there is a checkpoint to stop at, then it means an upgrade is expected. But if there is no control handler, then we have a major bug. This should be detectable at compile time. If it's possible to trigger an upgrade, than there must be an upgrade handler. At a minimum, there should be a log message, and likely an exception thrown/assert in this case.

Maybe we need to go back to the drawing board and figure out how we couple the control state to the actions to take based on that control state so that there are no gaps in the semantics.

bftengine/include/bftengine/Replica.hpp
```diff
@@ -50,6 +55,8 @@ class IRequestsHandler {
 
   virtual void onFinishExecutingReadWriteRequests() {}
   virtual ~IRequestsHandler() {}
+
+  virtual std::shared_ptr<IControlHandler> getControlHandlers() = 0;
```
andrewjstone, line 59: I'm a bit confused about why this is part of the interface. Doing this requires all users to pass in an `IControlHandler`, as evidenced by the required modification to tests in the code below. However, all these tests simply return a nullptr, which means that all users of this interface now need to check for a nullptr. It seems to me that every real system should have a set of control handlers, and that the reference should never be null. Also, I don't see where the request handler gets passed in. If it's at a higher level of abstraction, then why not also put the setter there. As with another comment further down, I think we may want to revisit our abstractions.

bftengine/include/bftengine/Replica.hpp
andrewjstone: @yontyon and I discussed this, and it makes sense as an abstract class because right now we expect the handlers to be implemented all or nothing by application code.

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: @yontyon and I discussed this. Since the control handlers are all or nothing and implemented by the application, it's possible some applications (like tests) may choose not to implement them. This is totally fine.

bftengine/include/bftengine/Replica.hpp
andrewjstone: Like above, this makes sense in light of the fact that the command handlers are set at a higher level of abstraction, but used inside concord-bft. This is how they are accessed from concord-bft.

andrewjstone:
Thanks for the discussion @yontyon ! Great job!

bftengine/include/bftengine/Replica.hpp
robem, line 38: This needs a comment. AFAIU, a control handler is responsible for infrastructure/system-level operations which are transparent to application handlers (the "execute" callback). In any case, we should have a comment that explicitly states the use-case/intend and maybe even expands on the notion of "Control State".

bftengine/src/bftengine/ReplicaImp.cpp
robem: I'm confused by the naming. In here we pretty much say "super stable CP == upgrade". This might be true for now and in our use-case but what we actually want is to "just get notified" about the super stable CP and *therafter* invoke upgrade. In the current implementation we tie those two together. What if we want to use a super stable checkpoint for something else? Like a general point for synchronization; for backup, for system health reports, to onboard new clients/replicas ...
My suggestion is that each control handler has to provide an `onSuperStableCP` API or similar which then invokes upgrade ... or whatever else.

yontyon: Agreed.
Also, I believe we will have to also write in the reserved pages the reason for stopping at the next checkpoint.
This way, we will be able to distinguish between the different cases.
(see #774 )

bftengine/src/bftengine/CheckpointInfo.hpp
yontyon: done

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2337,7 +2351,11 @@ void ReplicaImp::onSeqNumIsStable(SeqNum newStableSeqNum, bool hasStateInformati
 
   mainLog->advanceActiveWindow(lastStableSeqNum + 1);
 
-  checkpointsLog->advanceActiveWindow(lastStableSeqNum);
+  if (checkpointsLog->insideActiveWindow(newStableSeqNum - checkpointWindowSize)) {
```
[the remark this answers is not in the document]
yontyon: done

bftengine/include/bftengine/Replica.hpp
yontyon: done

PR #767, opened 2020-07-20
Decrease wait times for Apollo RO replica tests

tdimitrov:
Sometimes tests fail on Travis. This patch decreases
some time related values to make the tests run faster.


Liorgal28:
LGTM! but why 3 seconds and not 2 or 4?

tdimitrov:
> LGTM! but why 3 seconds and not 2 or 4?

The test simulates S3 service going down. Ideally it should be for a longer time, but we should keep you n mind Travis time limits.

Or something n nutshell - it's a bit of a guessing game :(

PR #769, opened 2020-07-20
Rename Assert* to ConcordAssert*

f-squirrel:
Reason:
The name `Assert` is not unique and conflicts with Google Test:
```
/googletest/googlemock/include/gmock/internal/gmock-internal-utils.h:295:36: error: too many arguments provided to function-like macro invocation
inline void Assert(bool condition, const char* file, int line,
```
Using `#undef` seems to be tricky.

The problem was discovered after `assert` was replaced with `Assert`.

Changes:
All the macros `Assert*` were renamed to `ConcordAssert*`.
Example, `AssertOR` -> `ConcordAssertOR`.

Feel free to propose other prefixes.


robem:
So GoogleTest determines that we cannot use `Assert`? This sounds wrong and I haven't fully thought this through but there must be a better solution than renaming every assert.

PR #772, opened 2020-07-21
Disable S3 tests on Travis

tdimitrov:
S3 tests are not stable on Travis and cause a lot of PRs to fail.

This patch disables them on Travis and keeps them active on Github Actions. I think it's better to at least have them run somewhere despite the discrepancies between the two build configs.


tdimitrov:
> LGTM!
> I have a question.
> In one of your PRs you've decreased the timeout(?) because Travis CI jobs were time-limited.
> If we are not going to run the test there, is the timeout still relevant?

The idea was to make the tests more stable (which failed) and to make them run faster. I think it doesn't matter if the service will be down for 3 or 5 seconds, so I prefer to keep it that way to save time. But I haven't got that strong opinion on this matter - so I can switch it back to 5 if it makes more sense.

f-squirrel:
@tdimitrov neither have I, let us leave it as is😊

PR #775, opened 2020-07-21
Add infrastructure for bft multi KeyExchange msg

cloudnoize:
- KeyManager singleton class, that manages all related multi-signature functionality e.g. generates keys, initiates the client request using its ICLient member, etc.

- BFT handler the wraps the user handler and forward requests to it as before, unless the request is a BFT msg such as KeyExchange msg.

- A feature flag to enable this feature.

- Once the feature is enabled, the BFT won't accept any msgs other then KeyExchange msgs until all replicas have published there newly generated key.

- Call back is called each checkpoint.

- KeyExchange struct.

### DIFF
--- a/bftengine/include/bftengine/Replica.hpp
+++ b/bftengine/include/bftengine/Replica.hpp
@@ -32,7 +32,8 @@ enum MsgFlag : uint8_t {
   EMPTY_FLAGS = 0x0,
   READ_ONLY_FLAG = 0x1,
   PRE_PROCESS_FLAG = 0x2,
-  HAS_PRE_PROCESSED_FLAG = 0x4
+  HAS_PRE_PROCESSED_FLAG = 0x4,
+  KEY_EXCHANGE_FLAG = 0x8
 };
 
 // The ControlHandlers is a group of method that enables the userRequestHandler to perform infrastructure
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -118,6 +118,8 @@ struct ReplicaConfig {
 
   // Metrics dump interval
   uint64_t metricsDumpIntervalSeconds = 600;
+
+  bool keyExchangeOnStart = false;
   /**
    * create a singleton instance from this object
    * call to this function will have effect only for the first time
@@ -148,7 +150,8 @@ inline std::ostream& operator<<(std::ostream& os, const ReplicaConfig& rc) {
      << "maxNumOfReservedPages: " << rc.maxNumOfReservedPages << "\n"
      << "sizeOfReservedPage: " << rc.sizeOfReservedPage << "\n"
      << "debugStatisticsEnabled: " << rc.debugStatisticsEnabled << "\n"
-     << "metricsDumpIntervalSeconds: " << rc.metricsDumpIntervalSeconds << "\n";
+     << "metricsDumpIntervalSeconds: " << rc.metricsDumpIntervalSeconds << "\n"
+     << "keyExchangeOnStart: " << rc.keyExchangeOnStart << "\n";
   return os;
 }
 
@@ -165,6 +168,7 @@ class ReplicaConfigSingleton {
 
   uint16_t GetFVal() const { return config_->fVal; }
   uint16_t GetCVal() const { return config_->cVal; }
+  uint16_t GetNumReplicas() const { return config_->numReplicas; }
   uint16_t GetReplicaId() const { return config_->replicaId; }
   uint16_t GetNumOfClientProxies() const { return config_->numOfClientProxies; }
   uint16_t GetNumOfExternalClients() const { return config_->numOfExternalClients; }
@@ -209,6 +213,8 @@ class ReplicaConfigSingleton {
   uint64_t GetMetricsDumpInterval() const { return config_->metricsDumpIntervalSeconds; }
 
   bool GetDebugStatisticsEnabled() const { return config_->debugStatisticsEnabled; }
+
+  bool GetKeyExchangeOnStart() const { return config_->keyExchangeOnStart; }
 
  private:
   friend struct ReplicaConfig;
--- a/bftengine/src/bftengine/KeyManager.cpp
+++ b/bftengine/src/bftengine/KeyManager.cpp
@@ -1,0 +1,75 @@
+#include "KeyManager.h"
+#include "thread"
+#include "ReplicaImp.hpp"
+#include "ReplicaConfig.hpp"
+
+const std::string KeyManager::KeyExchangeMsg::getVersion() const { return "1"; }
+
+KeyManager::KeyExchangeMsg KeyManager::KeyExchangeMsg::deserializeMsg(const char* serializedMsg, const uint32_t& size) {
+  std::stringstream ss;
+  KeyManager::KeyExchangeMsg ke;
+  ss.write(serializedMsg, std::streamsize(size));
+  deserialize(ss, ke);
+  return ke;
+}
+
+void KeyManager::KeyExchangeMsg::serializeDataMembers(std::ostream& outStream) const {
+  serialize(outStream, key);
+  LOG_TRACE(GL, "KEY EXCHANGE MANAGER  ser  key_ " << key);
+  serialize(outStream, signature);
+  LOG_TRACE(GL, "KEY EXCHANGE MANAGER  ser  signature_ " << signature);
+  serialize(outStream, repID);
+  LOG_TRACE(GL, "KEY EXCHANGE MANAGER  ser  repID_ " << repID);
+}
+
+void KeyManager::KeyExchangeMsg::deserializeDataMembers(std::istream& inStream) {
+  deserialize(inStream, key);
+  deserialize(inStream, signature);
+  deserialize(inStream, repID);
+}
+
+std::string KeyManager::KeyExchangeMsg::toString() const {
+  std::stringstream ss;
+  ss << "key [" << key << "] signature [" << signature << "] replica id [" << repID << "]";
+  return ss.str();
+}
+
+void KeyManager::set(bftEngine::IBasicClient* cl, const int& id) {
+  cl_ = cl;
+  repID_ = id;
+}
+
+/*
+Usage:
+  KeyExchangeMsg msg{key,sig,id};
+  std::stringstream ss;
+  concord::serialize::Serializable::serialize(ss, msg);
+  auto strMsg = ss.str();
+  char buff[128];
+  uint32_t actSize{};
+  // magic numbers, need to check valid values.
+  cl_->invokeCommandSynch(strMsg.c_str(),
+                          strMsg.size(),
+                          bftEngine::KEY_EXCHANGE_FLAG,
+                          std::chrono::milliseconds(60000),
+                          44100,
+                          buff,
+                          &actSize);
+*/
+void KeyManager::sendKeyExchange() { LOG_DEBUG(GL, "KEY EXCHANGE MANAGER  send msg"); }
+
+std::string KeyManager::onKeyExchange(const KeyExchangeMsg& kemsg) {
+  static int counter = 0;
+  counter++;
+  LOG_DEBUG(GL, "KEY EXCHANGE MANAGER  msg " << kemsg.toString());
+  auto numRep = bftEngine::ReplicaConfigSingleton::GetInstance().GetNumReplicas();
+  if (counter < numRep) {
+    LOG_DEBUG(GL, "Exchanged [" << counter << "] out of [" << numRep << "]");
+  } else if (counter == numRep) {
+    LOG_INFO(GL, "KEY EXCHANGE: start accepting msgs");
+    keysExchanged = true;
+  }
+  return "ok";
+}
+
+void KeyManager::onCheckpoint(const int& num) { LOG_DEBUG(GL, "KEY EXCHANGE MANAGER check point  " << num); }
--- a/bftengine/src/bftengine/KeyManager.h
+++ b/bftengine/src/bftengine/KeyManager.h
@@ -1,0 +1,41 @@
+#pragma once
+
+#include "Serializable.h"
+#include "IBasicClient.h"
+
+class KeyManager {
+ public:
+  struct KeyExchangeMsg : public concord::serialize::SerializableFactory<KeyExchangeMsg> {
+    std::string key;
+    std::string signature;
+    uint32_t repID;
+    std::string toString() const;
+    static KeyExchangeMsg deserializeMsg(const char* serStr, const uint32_t& size);
+
+   protected:
+    const std::string getVersion() const;
+    void serializeDataMembers(std::ostream& outStream) const;
+    void deserializeDataMembers(std::istream& inStream);
+  };
+
+ private:
+  bftEngine::IBasicClient* cl_{nullptr};
+  int repID_{};
+  KeyManager() {}
+  KeyManager(const KeyManager&) = delete;
+  KeyManager(const KeyManager&&) = delete;
+  KeyManager& operator=(const KeyManager&) = delete;
+  KeyManager& operator=(const KeyManager&&) = delete;
+
+ public:
+  std::atomic_bool keysExchanged{false};
+  void set(bftEngine::IBasicClient* cl, const int& id);
+  void sendKeyExchange();
+  std::string onKeyExchange(const KeyExchangeMsg& kemsg);
+  void onCheckpoint(const int& num);
+
+  static KeyManager& get() {
+    static KeyManager km;
+    return km;
+  }
+};
--- a/bftengine/CMakeLists.txt
+++ b/bftengine/CMakeLists.txt
@@ -57,6 +57,8 @@ set(corebft_source_files
     src/bftengine/messages/ViewChangeMsg.cpp
     src/bftengine/messages/ReplicaStatusMsg.cpp
     src/bftengine/messages/StateTransferMsg.cpp
+    src/bftengine/KeyManager.cpp
+    src/bftengine/RequestHandler.cpp
     src/bftengine/ControlStateManager.cpp
 )
 
--- a/bftengine/include/bftengine/IBasicClient.h
+++ b/bftengine/include/bftengine/IBasicClient.h
@@ -1,0 +1,22 @@
+#pragma once
+#include "status.hpp"
+#include <chrono>
+#include <string>
+
+namespace bftEngine {
+
+class IBasicClient {
+ public:
+  virtual ~IBasicClient() = default;
+  virtual concordUtils::Status invokeCommandSynch(const char* request,
+                                                  uint32_t requestSize,
+                                                  uint8_t flags,
+                                                  std::chrono::milliseconds timeout,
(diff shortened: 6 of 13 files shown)


bftengine/src/bftengine/BftCommandHandler.cpp
```diff
@@ -0,0 +1,47 @@
...
+                               uint8_t flags,
+                               uint32_t requestSize,
+                               const char *request,
+                               uint32_t maxReplySize,
+                               char *outReply,
+                               uint32_t &outActualReplySize,
+                               uint32_t &outReplicaSpecificInfoSize,
+                               concordUtils::SpanWrapper &parent_span) {
+  if (flags & KEY_EXCHANGE_FLAG) {
+    KeyManager::KeyExchangeMsg ke = KeyManager::KeyExchangeMsg::KeyExDes(request, requestSize);
```
f-squirrel, line 18: Could you please change `KeyExDes` to a more readable name?

bftengine/src/bftengine/BftCommandHandler.h
```diff
@@ -0,0 +1,26 @@
+#include <Replica.hpp>
+
+#pragma once
+
+namespace bftEngine {
+
+class BftCommandHandler : public IRequestsHandler {
```
f-squirrel, line 7: Is there a reason that it does not inherit from `ICommandHandler`?

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "thread"
+#include "ReplicaImp.hpp"
+#include "ReplicaConfig.hpp"
+
+// KeyManager::KeyExchangeMsg::KeyExchangeMsg(KeyManager::KeyExchangeMsg&&
+// other):key_(std::move(other.key_)),signature_(other.signature_),repID_(other.repID_){}
+
+const std::string KeyManager::KeyExchangeMsg::getVersion() const { return "1"; }
+
+KeyManager::KeyExchangeMsg KeyManager::KeyExchangeMsg::KeyExDes(const char* serStr, const uint32_t& size) {
```
f-squirrel, line 11: Could you please rename `serStr` to a more descriptive name?

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+  std::stringstream ss;
```
f-squirrel, line 12: ostringstream?

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+KeyManager& KeyManager::get() {
+  static KeyManager km;
+  return km;
+}
+void KeyManager::set(concord::kvbc::IClient* cl, const int& id) {
+  cl_ = cl;
+  repID_ = id;
+}
+
+void KeyManager::start(const int& durMs) {
```
f-squirrel, line 49: Can we make the variable name more readable?
Why is it `const int&`?
Can we pass a specific type aka `std::chrono::milliseconds`?

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+  auto l = [this](const int& durMs) {
+    while (true) {
```
f-squirrel, line 51: Stop condition?

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+      std::this_thread::sleep_for(std::chrono::milliseconds(durMs));
+      sendKeyExchange();
+    }
+  };
+  std::thread t(l, durMs);
+  t.detach();
```
f-squirrel, line 57: Detaching threads is highly not recommended:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-detached_thread

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+void KeyManager::sendKeyExchange() {
+  KeyExchangeMsg msg;
+  msg.key_ = "THIS_IS_THE_KEY_" + std::to_string(keyNum_++);
```
f-squirrel, line 62: Is it some sort of unfinished code?
Do we really need to submit it now?

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+  msg.signature_ = "84E34B434A";
+  msg.repID_ = repID_;
+
+  std::stringstream ss;
+  concord::serialize::Serializable::serialize(ss, msg);
+  auto strMsg = ss.str();
+  char buff[44100];
```
f-squirrel, line 69: I guess it is some intermediate code and we run on heavy machines but still, I am a bit afraid of allocating such kind of memory on the stack.

bftengine/src/bftengine/KeyManager.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include "KVBCInterfaces.h"
+#include "Serializable.h"
+
+class KeyManager {
+ public:
+  struct KeyExchangeMsg : public concord::serialize::SerializableFactory<KeyExchangeMsg> {
+    std::string key_;
```
f-squirrel, line 9: If it is public member, please remove `_`.

bftengine/src/bftengine/BftCommandHandler.h
cloudnoize: Yes, ICommandHandler has an additional method that is not necessary for this class.

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: I think it's an easier interface to pass an int rather than the chrono type.

cloudnoize: but I should mention that this class is a template class for @toly-kournik  to implement the real functionality.
This method will not be in production, but for testing purposes.

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: It's a test function, will be deleted once real generation takes place

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: It's a test function, will be deleted once real generation takes place

bftengine/src/bftengine/KeyManager.cpp
andrewjstone: The std::chrono types are much safer to use, and prevent a lot of conversion errors. I would prefer if we could use them for everything where time is relevant. Effort went into refactoring the timers and other parts of the code base to use std::chrono already, and not using them for convenience doesn't seem like a solid justification. We did in fact have bugs due to improper type conversion between seconds and milliseconds without them and they were very hard to find.

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: It's protected by a feature flag.
I can remove it. it is intermediate

bftengine/src/bftengine/KeyManager.h
cloudnoize: I don't mind doing it, but I'm not familiar with this convention, I do actually prefer marking all members with _

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: yes, you right, It's a skeleton class. I'll clean it

cloudnoize:
> Great commit @cloudnoize!
> Besides the comments, I have a general question: do we actually need the class to be a singleton?
> From my experience singletons make unit tests less effective and may lead to various problems in production(uncontrollable destruction).
> Can it be initialized in the beginning and passed as a reference to all the users?

It's by the design of @toly-kournik , so I let him have the choice,  personally I find singletons useful and didn't have troubles with them, though I know they have bad reputation probably because of the reasons you've mentioned :)

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: istringstream

bftengine/src/bftengine/BftCommandHandler.h
f-squirrel: Are we talking about the same [class](https://github.com/vmware/concord-bft/blob/master/kvbc/include/KVBCInterfaces.h#L128)?

toly-kournik: Yes, ICommandsHandler is defined in kvbc layer :  kvbc/include/KVBCInterfaces.h

bftengine/src/bftengine/KeyManager.cpp
f-squirrel: Right, my bad

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -0,0 +1,96 @@
...
+KeyManager& KeyManager::get() {
+  static KeyManager km;
+  return km;
+}
+void KeyManager::set(concord::kvbc::IClient* cl, const int& id) {
+  cl_ = cl;
+  repID_ = id;
+}
+
+// Method for testing periodic generation
```
f-squirrel, line 47: I really don't like the idea of submitting commented code.
The fact that the feature has a feature flag is not an excuse.
Please remove or move to unit tests.

bftengine/src/bftengine/KeyManager.h
f-squirrel: Here is the convention:
https://google.github.io/styleguide/cppguide.html#Variable_Names.
Please change all over.

bftengine/src/bftengine/KeyManager.h
```diff
@@ -0,0 +1,32 @@
...
+    std::string signature_;
+    uint32_t repID_;
+    std::string toString() const;
+    static KeyExchangeMsg MsgDeserializer(const char* serStr, const uint32_t& size);
+
+   protected:
+    const std::string getVersion() const;
```
f-squirrel, line 16: `const` for the return value is redundant.

bftengine/src/bftengine/BftCommandHandler.h
```diff
@@ -0,0 +1,26 @@
...
+ private:
```
toly-kournik, line 8: Conventions: (sorry for being nagging) :
please put public at a first place and members should end with an '_' underscore

bftengine/src/bftengine/KeyManager.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include "KVBCInterfaces.h"
+#include "Serializable.h"
+
+class KeyManager {
+ public:
+  struct KeyExchangeMsg : public concord::serialize::SerializableFactory<KeyExchangeMsg> {
```
toly-kournik, line 8: TODO @toly-kournik : subclass from `Serializable`

bftengine/src/bftengine/KeyManager.h
```diff
@@ -0,0 +1,32 @@
...
+    void deserializeDataMembers(std::istream& inStream);
+  };
+
+ private:
+  concord::kvbc::IClient* cl_{nullptr};
+  int repID_{};
+
+ public:
+  std::atomic_bool keysExcahnged_{false};
+  static KeyManager& get();
```
toly-kournik, line 27: Let's put the impl here

bftengine/src/bftengine/BftCommandHandler.cpp
```diff
@@ -0,0 +1,47 @@
...
+                               uint8_t flags,
+                               uint32_t requestSize,
+                               const char *request,
+                               uint32_t maxReplySize,
+                               char *outReply,
+                               uint32_t &outActualReplySize,
+                               uint32_t &outReplicaSpecificInfoSize,
+                               concordUtils::SpanWrapper &parent_span) {
+  if (flags & KEY_EXCHANGE_FLAG) {
+    KeyManager::KeyExchangeMsg ke = KeyManager::KeyExchangeMsg::MsgDeserializer(request, requestSize);
```
f-squirrel, line 18: Usually, functions and method' names are verbs and classes/structs are nouns.
Any particular reason for this?
Can we call it `deserialize` -- it is a member of `KeyExchangeMsg` and obviously returns a message.

bftengine/src/bftengine/KeyManager.h
f-squirrel, line 27: If we do like singletons, then ctor has to be private, copy/assign/move-copy/move-assign attributes to be deleted.

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -118,6 +118,8 @@ struct ReplicaConfig {
 
   // Metrics dump interval
   uint64_t metricsDumpIntervalSeconds = 600;
+
+  bool multiSignatureEnabled = false;
```
toly-kournik, line 122: Naming: I think this name is misleading, since it's actually controls the behavior of checking the initial key exchange

bftengine/src/bftengine/KeyManager.h
```diff
@@ -0,0 +1,32 @@
...
+#include "Serializable.h"
+
+class KeyManager {
+ public:
+  struct KeyExchangeMsg : public concord::serialize::SerializableFactory<KeyExchangeMsg> {
+    std::string key_;
+    std::string signature_;
+    uint32_t repID_;
+    std::string toString() const;
+    static KeyExchangeMsg MsgDeserializer(const char* serStr, const uint32_t& size);
```
toly-kournik, line 13: Why upper case?

bftengine/src/bftengine/BftCommandHandler.h
f-squirrel: @toly-kournik , I see, the kvbc is a higher level.
I think we have a problem in the hierarchy: 2 command handlers, where one inherits from `IRequestsHandler` and another from `ICommansHandler`. It is a bit misleading. Maybe we can move `ICommandsHandler` to the `bftEngine` namespace?

toly-kournik: To tell the truth I don't really understand the purpose of `ICommandsHandler`.
Probably it was designed to enhance `IRequestHandler` and be an API for KVBC, but currently it does nothing.
`setControlStateManager()` was added recently by @yontyon

f-squirrel: I was the one who recommended this:)
I do not understand the purpose of IRequestsHandler because nobody inherits from it directly.
I'd be happy to get rid of one of them.

toly-kournik: Since `IRequestHandler` is now used in bftEngine it should remain.
`ICommandsHandler` on the other hand is used by concord

cloudnoize: This is a bad design of interfaces, I don't see any reason to inherit from ICommandsHandler and to Dummy implement stuff

bftengine/src/bftengine/KeyManager.h
cloudnoize: It's required from the Serializable interface

bftengine/src/bftengine/BftCommandHandler.cpp
cloudnoize: deserialize is taken from the Serializable interface, otherwise, I would use it

bftengine/include/bftengine/Replica.hpp
```diff
@@ -32,7 +32,8 @@ enum MsgFlag : uint8_t {
   EMPTY_FLAGS = 0x0,
   READ_ONLY_FLAG = 0x1,
   PRE_PROCESS_FLAG = 0x2,
-  HAS_PRE_PROCESSED_FLAG = 0x4
+  HAS_PRE_PROCESSED_FLAG = 0x4,
+  KEY_EXCHANGE_FLAG = 0x8
```
yontyon, line 36: Why not creating a flag for all internal bft commands something like BFT_INTENAL_CMD_FLAG=0X8

bftengine/src/bftengine/KeyManager.h
f-squirrel: Holy crp, gotcha!

bftengine/src/bftengine/BftCommandHandler.h
f-squirrel: The basic question about the design is if the new command handler is a commands handler?
If it is then it is a bad design to inherit from something else than ICommandsHandler.
Regarding, the dummy implementation, I think we might want to have a few base types for handlers, etc.

I understand that this is beyond the scope of the PR, the feature and obviously I do not ask to change it but we have to make a decision.
Otherwise, the inheritance model of the project will end up being ridiculously complex and unclear.

cloudnoize: Multiple inheritance implies bad design in my opinion, also over complex interfaces. If one wants to use an interface but needs to implement dummy methods, something is not correct.

bftengine/src/bftengine/BftCommandHandler.cpp
f-squirrel: I see but what is the problem to call it the same name?
The method will just call the parent class' method `SerializableFactory::deserialize`.

bftengine/include/bftengine/Replica.hpp
cloudnoize: Then we need another mechanism to differentiate between msgs, it can be done, but not necessary at this phase IMO

bftengine/src/bftengine/BftCommandHandler.h
f-squirrel: Let us take it offline:)

bftengine/src/bftengine/BftCommandHandler.cpp
cloudnoize: It means that every usage of the `deserialize` interface method within that class, needs to be prefixed with `Serializable:: ` .
It's error prone and too verbose IMO

cloudnoize: I'm calling it `deserializeMsg` .

f-squirrel: If the verbosity/complexity is hidden from the class user, then it is fine.
Anyways, I am ok with `deserializeMsg`.

bftengine/src/bftengine/KeyManager.h
f-squirrel: Actually, I'd really like this class to be conformant with the code style.
AFAIK, underscores are not used for public members in the project.

cloudnoize: I can find examples with `_` for example 
`struct Msg {
  std::unique_ptr<char[]> msg_;
  uint32_t len_;
  uint16_t to_;
};`
In my opinion, it's too pedantic and missing the point of code review to nag on nit stuff

PR #776, opened 2020-07-21
Add support for the Concord Message Format (CMF)

andrewjstone:
To enhance building message driven systems and enable architecture and
language independence, a uniform messaging format is required. This
commit defines a formal grammar for a schema language with which to
define messages. It also defines a canonical serialization format,
described informally in `messages/README.md`. Furthermore, code
generations is provided for C++ such that CMF messages can be compiled
to generate C++ structs and their corresponding serialization code.

The README provides much more information. Please read it to get a more
detailed understanding.

It's expected that this will lead to much discussion. However, please
try to focus on the semantics of the messages, and the code generation,
and don't nitpick the syntax of the messages. The schema for the
messages is formally defined by the grammar, and we can add syntax or
change naming as needed by modifying the grammar later.


messages/compiler/walker.py
```diff
@@ -0,0 +1,86 @@
...
+
+def field_type(type):
+    """Extract the outermost type from a field"""
+    if not isinstance(type, dict):
+        if is_primitive(type):
+            return type
+        return "msg"
+    for compound in ["list", "kvpair", "map", "optional", "oneof"]:
+        if compound in type:
+            return compound
```
andrewjstone, line 32: Looks like I need to throw an exception here if nothing is found.

messages/compiler/visitor.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Args:
+            name (str): The name of the field
+            type (str): The high level (outermost) type of the field
+        """
+        pass
+
+    @abstractmethod
+    def field_end(self):
+        """ The field has been fully defined. """
```
andrewjstone, line 81: missing a `pass` statement

robem: Same for `oneof`

messages/compiler/cpp/test_cppgen.py
```diff
@@ -0,0 +1,294 @@
...
+def compile_tests():
+    print("Compiling tests with g++")
+    os.system(
+        f"g++ -std=c++17 -g -o {OUTPUT_DIR}/test_serialization {OUTPUT_DIR}/test_serialization.cpp"
+    )
+
+
+def run_tests():
+    print("Running tests")
+    if os.system("./test_serialization") == 0:
```
robem, line 261: This needs to be `f"./{OUTPUT_DIR}/test_serialization"`

andrewjstone: Ahh, silly last minute changes to use an output directory. Good catch!

messages/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Comments must be on their own line and start with the `#` character. Leading whitespace is allowed.
+
+ # Serialized Representations
+
+* `bool` - `0x00` = False, `0x01` = True
+* `uint8` - The value itself
+* `uint16` - The value itself in little endian
+* `uint32` - The value itself in little endian
+* `uint64` - The value itself in little endian
```
tdimitrov, line 89: Just curious - is the value that travels over the wire is in little endian? If so - why? Network packets are usually encoded in big endian.

messages/compiler/cpp/cpp_visitor.py
```diff
@@ -0,0 +1,261 @@
...
+
+def serialize_start(name):
+    return f"""
+void serialize(std::vector<uint8_t>& output, const {name}& t) {{
+"""
+
+
+def deserialize_start(name):
+    return f"""
+void deserialize(uint8_t*& input, const uint8_t* end, {name}& t) {{
```
f-squirrel, line 32: Is the input mutable? Can it be constant?

messages/compiler/cpp/cpp_visitor.py
```diff
@@ -0,0 +1,261 @@
...
+        ])
+        # output and oneofs_seen accumulate across messages
+        output = self.output
+        oneofs = self.oneofs_seen
+        self.__init__()
+        self.output = output
+        self.oneofs_seen = oneofs
+
+    def field_start(self, name, type):
+        self.struct += "  "  # Indent fields
```
f-squirrel, line 173: Regarding the indents, maybe in the future, we can run clang-format for the generated files?
WDYT?

messages/compiler/cpp/cpp_visitor.py
```diff
@@ -0,0 +1,261 @@
...
+                self.oneof_deserialize,
+                equalop_str(self.msg_name, self.fields_seen),
+                self.serialize,
+                self.deserialize,
+            ] if s != ''
+        ])
+        # output and oneofs_seen accumulate across messages
+        output = self.output
+        oneofs = self.oneofs_seen
+        self.__init__()
```
f-squirrel, line 168: Can we call it `reset` or `clean`?

messages/README.md
andrewjstone: The value that travels over the wire is indeed in little endian.

It's a somewhat arbitrary choice biased for common architectures. Big endian was chosen a long time ago when most machines were big endian. That is no longer the case. Since we define the protocol, there's no reason we can't pick the opposite :)

The rationale is that all Intel machines and clones are little endian and we can optimize the code to do a memcpy rather than a byte by byte conversion on little endian machines, and arm machines can be configured to work in little endian. It also allows for zero copy deserialization on little endian machines, although our current implementation doesn't do that because it's less ergonomic to use and I doubt serialization is going to be our bottleneck in concord. 

@f-squirrel Any big endian machine will do the write thing. This is just what goes out on the wire.

If this is a strong concern, it's trivial to change.

messages/compiler/cmfc.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
andrewjstone: Good catch!

messages/compiler/cpp/cpp_visitor.py
andrewjstone: Input is mutable, otherwise I would have taken it directly by const ref or pointer, or used a const_iterator.  Since deserialization is recursive, we need to move the pointer along the input buffer to know where to continue deserializing from.

messages/compiler/cpp/cpp_visitor.py
andrewjstone: I think it's a good idea.

messages/compiler/cpp/cpp_visitor.py
andrewjstone: __init__() is the "magic" constructor function. I just reused it rather than adding another method.

f-squirrel: I know what `__init__` is. I was not sure if the intention was clear.
Nevermind, I do have a strong opinion on this.

andrewjstone: Thinking about this, I think you are right. It will be clearer to the reader with an alias method. I will add one. Thanks!

messages/README.md
f-squirrel: I know that most of the modern processors are little-endian but if we are building a network protocol it might be reasonable to sent data in network byte order.
AFAIK, most of the ARMs are little-endian. And from my experience ppl don't like to change the default endianness on their platforms.

Anyways, I don't think we have plans to run our SW on some exotic architecture so let us leave it as is.

andrewjstone: FWIW Libra canonical serialization uses little endian as well. https://github.com/libra/libra/tree/6a89e827b95405066dc83eec97eca2cb75bc991d/common/canonical-serialization

messages/compiler/walker.py
andrewjstone: fixed.

messages/compiler/cpp/test_cppgen.py
andrewjstone: fixed.

messages/compiler/visitor.py
andrewjstone: fixed.

PR #777, opened 2020-07-22
[Apollo] Accelerate test scenarios, while preserving stability

teoparvanov:
Once again, we have hit the 50 min limitation for Travis job executions. This PR aims to unblock the pipeline.

PS: Ultimately, we should focus on stabilizing GitHub Actions runs, and drop Travis CI altogether.

### DIFF
--- a/tests/apollo/test_skvbc_backup_restore.py
+++ b/tests/apollo/test_skvbc_backup_restore.py
@@ -102,7 +102,8 @@ class SkvbcBackupRestoreTest(unittest.TestCase):
         )
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


f-squirrel:
@teoparvanov , I've posted a [question/bug](https://github.com/actions/virtual-environments/issues/1281) regarding the `clang-tidy` issue to the GA team.

f-squirrel:
@teoparvanov , I totally agree that we have to focus on GA stability.
However, I am afraid it is almost impossible w/o a dedicated effort of the whole team.

PR #778, opened 2020-07-22
remove loading rocksdb auto generate config file

yontyon:
Currently, we first load the auto-generated config file, and only if it doesn't exist (only in the first run) we load our default configuration file.
This complicates the case in which we start without loading our default configuration file.
Thus, in this PR we removed the loading of the default configuration file and we load only our configuration file.


robem:
@yontyon No one outside VMware can access this link :) Please summarize the issue and solution in the description.

salieri11:
Please remove link to internal JIRA ticket and add bug description (alternatively, you can open issue here, on github). Otherwise, LGTM

PR #783, opened 2020-07-24
Add support for state transfer complete callbacks

dartdart26:
Add support for a subscription mechanism for State Transfer (ST)
on-complete callbacks. Rationale is that multiple parts of the system
might want to know when ST completes so that they can act accordingly to
the newly-received state. Examples of such cases are pruning and time
management, to name a few.

Provide an implementation for BCStateTran only. Implementations for
NullStateTransfer and SimpleStateTran are not provided at that stage as
the use is unclear.

In order to implement, add a generic CallbackRegistry that supports
multiple registrations, deregistration and arbitrary parameters for the
callbacks.

Add an unit test for CallbackRegistry to demonstrate usage and test
common cases.

### DIFF
--- a/bftengine/include/bftengine/IStateTransfer.hpp
+++ b/bftengine/include/bftengine/IStateTransfer.hpp
@@ -74,7 +74,7 @@ class IStateTransfer {
   // Accepts the checkpoint number as a parameter.
   // Callbacks must not throw.
   // Multiple callbacks can be added.
-  virtual void addOnTransferringCompleteCallback(std::function<void(int64_t)>) = 0;
+  virtual void addOnTransferringCompleteCallback(std::function<void(uint64_t)>) = 0;
 };
 
 // This interface may only be used when the state transfer module is runnning
--- a/bftengine/include/bftengine/IStateTransfer.hpp
+++ b/bftengine/include/bftengine/IStateTransfer.hpp
@@ -14,6 +14,7 @@
 #pragma once
 
 #include <cstdint>
+#include <functional>
 #include <string>
 
 namespace bftEngine {
@@ -68,6 +69,12 @@ class IStateTransfer {
   // Return the internal state (member variables, etc...) of the state transfer module as a string. This is used by the
   // diagnostics subsystem.
   virtual std::string getStatus() { return ""; };
+
+  // Registers a function that is called every time State Transfer completes.
+  // Accepts the checkpoint number as a parameter.
+  // Callbacks must not throw.
+  // Multiple callbacks can be added.
+  virtual void addOnTransferringCompleteCallback(std::function<void(int64_t)>) = 0;
 };
 
 // This interface may only be used when the state transfer module is runnning
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -18,6 +18,7 @@
 #include <string>
 #include <sstream>
 #include <functional>
+#include <utility>
 
 #include "assertUtils.hpp"
 #include "hex_tools.h"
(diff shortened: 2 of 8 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -248,6 +249,11 @@ BCStateTran::BCStateTran(const Config &config, IAppState *const stateApi, DataSt
   } else {
     messageHandler_ = std::bind(&BCStateTran::handleStateTransferMessageImp, this, _1, _2, _3);
   }
+
+  // Make sure that the internal IReplicaForStateTransfer callback is always registered, alongside any user-supplied
+  // callbacks.
+  on_transferring_complete_cb_registry_.registerCallback(
```
andrewjstone, line 255: Great reuse :)

bftengine/include/bftengine/IStateTransfer.hpp
```diff
@@ -68,6 +69,12 @@ class IStateTransfer {
   // Return the internal state (member variables, etc...) of the state transfer module as a string. This is used by the
   // diagnostics subsystem.
   virtual std::string getStatus() { return ""; };
+
+  // Registers a function that is called every time State Transfer completes.
+  // Accepts the checkpoint number as a parameter.
+  // Callbacks must not throw.
+  // Multiple callbacks can be added.
+  virtual void addOnTransferringCompleteCallback(std::function<void(int64_t)>) = 0;
```
andrewjstone, line 77: Why an int64_t and not uint64_t?

util/include/callback_registry.hpp
```diff
@@ -0,0 +1,96 @@
...
+  template <typename Func>
+  CallbackHandle registerCallback(Func&& callback) {
+    callbacks_.emplace_back(std::forward<Func>(callback));
+    auto it = callbacks_.cend();
+    return --it;
+  }
+
+  // Deregisters a callback. The corresponding handle is invalidated if this method returns. If the passed handle is
+  // invalid, the behavior is undefined.
+  void deregisterCallback(CallbackHandle handle) { callbacks_.erase(handle.iter_); }
```
andrewjstone, line 78: I think we can prevent the behavior from being undefined by the use of a bool in `GenericCallbackHandle` to signal if the iterator is valid still. It would be a silly user error to make, but it's better than erasing with undefined behavior IMO.

util/test/callback_registry_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+  auto handles = std::vector<CallbackRegistry<>::CallbackHandle>{};
+  const auto callback_count = 5;
+  for (auto i = 0; i < callback_count; ++i) {
+    handles.push_back(reg.registerCallback(callback));
+  }
+  reg.invokeAll();
+  ASSERT_EQ(callback_count, calls);
+}
+
+TEST(callback_registry, deregister) {
```
andrewjstone, line 111: Great test!

bftengine/include/bftengine/IStateTransfer.hpp
dartdart26: I think the existing callback uses int64_t, but I might be mistaken. Will double check when around a PC and change accordingly. Thanks for noticing.

util/include/callback_registry.hpp
dartdart26: It’s a good point. I elected to use the standard contract of the object being invalid after being moved from (as the user cannot call this function without moving).

In theory, if we check it we might as well throw an exception. Otherwise, it can hide a programming error. WDYT?

andrewjstone: I think idiomatic C++ contracts like this are valid, but dangerous, and are part of the giant set of footguns that can be reduced without changing the language. It's all done in the name of performance as far as I know, but that's not relevant here.

Yes, I think we can and should throw an exception.

dartdart26: Cool, will do that.

bftengine/include/bftengine/IStateTransfer.hpp
dartdart26: Changed to uint64_t.

dartdart26:
Thanks to @f-squirrel for suggestions regarding:

- simplifying `CallbackRegistry` method naming - from `(de)registerCallback()` to `add()` and `remove()`
- suppressing clang-tidy `bugprone-use-after-move` errors in the unit test

PR #785, opened 2020-07-26
Prevent replicas from initiating consensus if we decided to stop

yontyon:
If a replica discovers that it needs to stop on the current checkpoint but this checkpoint is not yet super stable checkpoint, we want this replica to stop participating in the consensus protocol (but not the state transfer, for example).

For that, a replica checks in three places whether it should stop:
1. primary - on try to send pre-prepare message
2. non-primary - on receiving pre-prepare message
3. all - on receiving a client request

We also added an enable/disable flag for the controlStateManager to the cases when we don't want to use it (for example, in some of the simple tests)

### DIFF
--- a/bftengine/include/bftengine/ControlStateManager.hpp
+++ b/bftengine/include/bftengine/ControlStateManager.hpp
@@ -34,10 +34,14 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
   ControlStateManager(const ControlStateManager&) = delete;
   ~ControlStateManager() {}
 
+  void disable() { enabled_ = false; }
+  void enable() { enabled_ = true; }
+
  private:
   IStateTransfer* state_transfer_;
   const uint32_t sizeOfReservedPage_;
   std::string scratchPage_;
+  bool enabled_ = true;
 
   // In the control handler manager reserved pages space, each control data should has its own page.
   // This struct define the index of each page in this space.
--- a/bftengine/src/bftengine/ControlStateManager.cpp
+++ b/bftengine/src/bftengine/ControlStateManager.cpp
@@ -24,6 +24,7 @@ namespace bftEngine {
  * given sequence number and mark it as a checkpoint to stop at in the reserved pages.
  */
 void ControlStateManager::setStopAtNextCheckpoint(int64_t currentSeqNum) {
+  if (!enabled_) return;
   uint64_t seq_num_to_stop_at = (currentSeqNum + 2 * checkpointWindowSize);
   seq_num_to_stop_at = seq_num_to_stop_at - (seq_num_to_stop_at % checkpointWindowSize);
   std::ostringstream outStream;
@@ -34,6 +35,7 @@ void ControlStateManager::setStopAtNextCheckpoint(int64_t currentSeqNum) {
 }
 
 std::optional<int64_t> ControlStateManager::getCheckpointToStopAt() {
+  if (!enabled_) return {};
   if (!state_transfer_->loadReservedPage(getUpdateReservedPageIndex(), sizeOfReservedPage_, scratchPage_.data())) {
     return {};
   }
--- a/bftengine/src/bftengine/ControlStateManager.cpp
+++ b/bftengine/src/bftengine/ControlStateManager.cpp
@@ -12,6 +12,7 @@
 // file.
 
 #include "ControlStateManager.hpp"
+#include "Logger.hpp"
 namespace bftEngine {
 
 /*
@@ -43,7 +44,10 @@ std::optional<int64_t> ControlStateManager::getCheckpointToStopAt() {
   inStream.str(scratchPage_);
   controlStateMessages::StopAtNextCheckpointMessage msg;
   concord::serialize::Serializable::deserialize(inStream, msg);
-  if (msg.seqNumToStopAt_ < 0) return {};
+  if (msg.seqNumToStopAt_ < 0) {
+    LOG_WARN(GL, "sequence num to stop at is negative!");
+    return {};
+  }
   return msg.seqNumToStopAt_;
 }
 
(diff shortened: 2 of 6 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -149,6 +149,9 @@ void ReplicaImp::onReportAboutInvalidMessage(MessageBase *msg, const char *reaso
 
 template <>
 void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
+  if (isSeqNumToStopAt(lastExecutedSeqNum))
```
f-squirrel, line 152: It might be useful to add a log message explaining why the message is ignored.
The same for `PrePrepareMsg`.

WDYT?

yontyon: agreed. I will add one

yontyon:
@teoparvanov I'm just about to start designing it

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -268,6 +274,12 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
 }
 
 void ReplicaImp::tryToSendPrePrepareMsg(bool batchingLogic) {
+  if (isSeqNumToStopAt(lastExecutedSeqNum)) {
+    LOG_INFO(GL, "Message ignored because we have to stop at super stable checkpoint");
```
robem, line 278: Let's make this message different from the one above. "Not sending PrePrepare message because ..." or similar.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -489,6 +513,10 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
 
 template <>
 void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
+  if (isSeqNumToStopAt(lastExecutedSeqNum)) {
+    LOG_INFO(GL, "Message ignored because we have to stop at super stable checkpoint");
```
robem, line 517: Same as the other log message. E.g., "Ignore PrePrepare message because ..." or similar.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -149,6 +149,12 @@ void ReplicaImp::onReportAboutInvalidMessage(MessageBase *msg, const char *reaso
 
 template <>
 void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
+  if (isSeqNumToStopAt(lastExecutedSeqNum)) {
+    LOG_INFO(GL, "Message ignored because we have to stop at super stable checkpoint");
```
robem, line 153: "Ignore ClientRequest message because ... " or similar.

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -34,10 +34,14 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
   ControlStateManager(const ControlStateManager&) = delete;
   ~ControlStateManager() {}
 
+  void disable() { enabled_ = false; }
```
robem, line 37: Your commit message doesn't describe why we can/want to "enable"/"disable" the ControlStateManager. What is its purpose? Are you using it as a feature flag?
A comment would help that explains in what scenario you call either.

bftengine/src/bftengine/ControlStateManager.cpp
```diff
@@ -34,13 +35,15 @@ void ControlStateManager::setStopAtNextCheckpoint(int64_t currentSeqNum) {
...
 std::optional<int64_t> ControlStateManager::getCheckpointToStopAt() {
+  if (!enabled_) return {};
   if (!state_transfer_->loadReservedPage(getUpdateReservedPageIndex(), sizeOfReservedPage_, scratchPage_.data())) {
     return {};
   }
   std::istringstream inStream;
   inStream.str(scratchPage_);
   controlStateMessages::StopAtNextCheckpointMessage msg;
   concord::serialize::Serializable::deserialize(inStream, msg);
+  if (msg.seqNumToStopAt_ < 0) return {};
```
robem, line 46: This is unexpected, correct? If so, we should at least log a warning.

bftengine/src/bftengine/ReplicaImp.cpp
robem: To clarify, I know that our current log format logs the function name. Still, I think it doesn't hurt to be more explicit.

bftengine/src/bftengine/ControlStateManager.cpp
andrewjstone: WE shouldn't assert as it's a client message that was received, and that would allow for malicious operators to crash the system. However, we should almost certainly log an error message and/or throw an exception if we can handle it upstream. My guess is that we have many places in this code base where we should throw exceptions, but we don't have any way to handle them. Logging a very loud error is probably the best short term fix.

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: I would suggest, something more abstract even. Super stable checkpoint still seems to be a weird, opaque term that will likely need to be explained over and over again to the few people not dealing with it. I think something like the following is better.

"Ignoring ClientRequest because system is stopped at checkpoint pending control state operation (upgrade, etc...)"

andrewjstone: Adding this will also allow us to remove the comment below, which essentially becomes duplication. On a side note, please put comments above lines and not next to them. This style can cause problems with clang-format.

bftengine/include/bftengine/ControlStateManager.hpp
yontyon: This is for cases when we don't want to use this feature (for example, in some of the simple tests)

PR #790, opened 2020-07-28
Add control command basic tests

yontyon:
In this PR we introduce 2 basic Apollo tests for the Wedge command.
1. Test that the system stops as a response to a wedge command when everything goes well
2. Test that the system stops as a response to a wedge command even when one replica is getting updated by the state transfer mechanism.

This PR shows that the basic functionality of the wedge command is working

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2382,8 +2382,11 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
 void ReplicaImp::onSeqNumIsSuperStable(SeqNum newSuperStableSeqNum) {
   auto seq_num_to_stop_at = controlStateManager_->getCheckpointToStopAt();
   if (seq_num_to_stop_at.has_value() && seq_num_to_stop_at.value() == newSuperStableSeqNum) {
-    if (getRequestsHandler()->getControlHandlers())
+    LOG_INFO(GL, "control state manager marked seqNum: " << newSuperStableSeqNum);
+    if (getRequestsHandler()->getControlHandlers()) {
+      metric_on_call_back_of_super_stable_cp_.Get().Set(1);
       getRequestsHandler()->getControlHandlers()->onSuperStableCheckpoint();
+    }
   }
 }
 void ReplicaImp::onSeqNumIsStable(SeqNum newStableSeqNum, bool hasStateInformation, bool oldSeqNum) {
@@ -3045,6 +3048,7 @@ ReplicaImp::ReplicaImp(bool firstTime,
       metric_current_primary_{metrics_.RegisterGauge("currentPrimary", curView % config_.numReplicas)},
       metric_concurrency_level_{metrics_.RegisterGauge("concurrencyLevel", config_.concurrencyLevel)},
       metric_primary_last_used_seq_num_{metrics_.RegisterGauge("primaryLastUsedSeqNum", primaryLastUsedSeqNum)},
+      metric_on_call_back_of_super_stable_cp_{metrics_.RegisterGauge("OnCallBackOfSuperStableCP", 0)},
       metric_first_commit_path_{metrics_.RegisterStatus(
           "firstCommitPath", CommitPathToStr(ControllerWithSimpleHistory_debugInitialFirstPath))},
       metric_slow_path_count_{metrics_.RegisterCounter("slowPathCount", 0)},
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2385,6 +2385,8 @@ void ReplicaImp::onSeqNumIsSuperStable(SeqNum superStableSeqNum) {
     LOG_INFO(GL, "Informing control state manager that consensus should be stopped: " << KVLOG(superStableSeqNum));
     if (getRequestsHandler()->getControlHandlers()) {
       metric_on_call_back_of_super_stable_cp_.Get().Set(1);
+      // TODO: With state transfered replica, this maight be called twice. Consider to clean the reserved page at the
+      // end of this method
       getRequestsHandler()->getControlHandlers()->onSuperStableCheckpoint();
     }
   }
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2379,10 +2379,10 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
   }
 }
 
-void ReplicaImp::onSeqNumIsSuperStable(SeqNum newSuperStableSeqNum) {
+void ReplicaImp::onSeqNumIsSuperStable(SeqNum superStableSeqNum) {
   auto seq_num_to_stop_at = controlStateManager_->getCheckpointToStopAt();
-  if (seq_num_to_stop_at.has_value() && seq_num_to_stop_at.value() == newSuperStableSeqNum) {
-    LOG_INFO(GL, "control state manager marked seqNum: " << newSuperStableSeqNum);
+  if (seq_num_to_stop_at.has_value() && seq_num_to_stop_at.value() == superStableSeqNum) {
+    LOG_INFO(GL, "Informing control state manager that consensus should be stopped: " << KVLOG(superStableSeqNum));
     if (getRequestsHandler()->getControlHandlers()) {
       metric_on_call_back_of_super_stable_cp_.Get().Set(1);
       getRequestsHandler()->getControlHandlers()->onSuperStableCheckpoint();
@@ -3528,12 +3528,6 @@ void ReplicaImp::executeNextCommittedRequests(concordUtils::SpanWrapper &parent_
   auto span = concordUtils::startChildSpan("bft_execute_next_committed_requests", parent_span);
 
   while (lastExecutedSeqNum < lastStableSeqNum + kWorkWindowSize) {
-    if (!stopAtNextCheckpoint_ && controlStateManager_->getCheckpointToStopAt().has_value()) {
-      // If, following the last execution, we discover that we need to jump to the
-      // next checkpoint, the primary sends noop commands until filling the working window.
-      bringTheSystemToCheckpointBySendingNoopCommands(controlStateManager_->getCheckpointToStopAt().value());
-      stopAtNextCheckpoint_ = true;
-    }
     SeqNum nextExecutedSeqNum = lastExecutedSeqNum + 1;
     SeqNumInfo &seqNumInfo = mainLog->get(nextExecutedSeqNum);
 
@@ -3561,7 +3555,12 @@ void ReplicaImp::executeNextCommittedRequests(concordUtils::SpanWrapper &parent_
       metric_total_fastPath_.Get().Inc();
     }
   }
-
+  if (!stopAtNextCheckpoint_ && controlStateManager_->getCheckpointToStopAt().has_value()) {
+    // If, following the last execution, we discover that we need to jump to the
+    // next checkpoint, the primary sends noop commands until filling the working window.
+    bringTheSystemToCheckpointBySendingNoopCommands(controlStateManager_->getCheckpointToStopAt().value());
+    stopAtNextCheckpoint_ = true;
+  }
   if (isCurrentPrimary() && requestsQueueOfPrimary.size() > 0) tryToSendPrePrepareMsg(true);
 }
 
--- a/tests/apollo/CMakeLists.txt
+++ b/tests/apollo/CMakeLists.txt
@@ -80,6 +80,4 @@ foreach(STORAGE_TYPE ${STORAGE_TYPES})
             "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_ro_replica ${TEST_OUTPUT}"
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
   endif()
-
 endforeach()
-
--- a/tests/apollo/CMakeLists.txt
+++ b/tests/apollo/CMakeLists.txt
@@ -67,6 +67,10 @@ foreach(STORAGE_TYPE ${STORAGE_TYPES})
           "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_backup_restore 2>&1 > /dev/null"
           WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
+  add_test(NAME skvbc_control_commands_tests_${STORAGE_TYPE} COMMAND sh -c
+          "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_control_commands ${TEST_OUTPUT}"
+          WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+
   if (BUILD_ROCKSDB_STORAGE)
     add_test(NAME skvbc_persistence_tests_${STORAGE_TYPE} COMMAND sh -c
             "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_persistence ${TEST_OUTPUT}"
@@ -76,5 +80,6 @@ foreach(STORAGE_TYPE ${STORAGE_TYPES})
             "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_ro_replica ${TEST_OUTPUT}"
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
   endif()
+
 endforeach()
 
--- a/tests/apollo/test_skvbc_control_commands.py
+++ b/tests/apollo/test_skvbc_control_commands.py
@@ -44,16 +44,22 @@ def start_replica_cmd(builddir, replica_id):
 
 class SkvbcControlCommandsTest(unittest.TestCase):
 
-    # __test__ = False  # so that PyTest ignores this test scenario
-
+    """
+        Sends a wedge command and check that the system stops from processing new requests.
+        Note that in this test we assume no failures and synchronized network.
+        The test does the following:
+        1. A client sends a wedge command
+        2. The client verify that the system reached to a super stable checkpoint
+        3. The client tries to initiate a new write bft command and fails
+    """
     @with_trio
     @with_bft_network(start_replica_cmd,  selected_configs=lambda n, f, c: n == 4 and c == 0)
     async def test_wedge_command(self, bft_network):
         bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
         client = bft_network.random_client()
-        rep = await client.write(bytearray(), special_flags=0x9)
-        await skvbc.write_known_kv()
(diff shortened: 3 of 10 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2382,8 +2382,11 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
 void ReplicaImp::onSeqNumIsSuperStable(SeqNum newSuperStableSeqNum) {
   auto seq_num_to_stop_at = controlStateManager_->getCheckpointToStopAt();
   if (seq_num_to_stop_at.has_value() && seq_num_to_stop_at.value() == newSuperStableSeqNum) {
-    if (getRequestsHandler()->getControlHandlers())
+    LOG_INFO(GL, "control state manager marked seqNum: " << newSuperStableSeqNum);
+    if (getRequestsHandler()->getControlHandlers()) {
+      metric_on_call_back_of_super_stable_cp_.Get().Set(1);
       getRequestsHandler()->getControlHandlers()->onSuperStableCheckpoint();
```
andrewjstone, line 2388: The more I read `SuperStableCheckpoint`, the less I like it. This doesn't have to be done in this PR, as it will have ripple effects, but maybe we can change it to something like `NofNStableCheckpoint`. Then we can rename the callback to be `onNofNCheckpointsReceived`

I'm open to other naming ideas as well. If everyone disagrees, we can also keep it as is :)

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @with_trio
+    @with_bft_network(start_replica_cmd,  selected_configs=lambda n, f, c: n == 4 and c == 0)
+    async def test_wedge_command(self, bft_network):
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        client = bft_network.random_client()
+        rep = await client.write(bytearray(), special_flags=0x9)
+        await skvbc.write_known_kv()
+        for replica_id in range(bft_network.config.n):
```
andrewjstone, line 57: I assume this check works because the primary sends no-ops to fill the windows, right?

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,81 @@
...
+            "-t", os.environ.get('STORAGE_TYPE')]
+
+
+class SkvbcControlCommandsTest(unittest.TestCase):
+
+    # __test__ = False  # so that PyTest ignores this test scenario
+
+    @with_trio
+    @with_bft_network(start_replica_cmd,  selected_configs=lambda n, f, c: n == 4 and c == 0)
+    async def test_wedge_command(self, bft_network):
```
andrewjstone, line 51: Can you please add a doc comment explaining what this function does? See our other apollo tests for guidance on what they should look like. This is especially important so that the final try clause is documented as explicitly failing because the system shouldn't accept more commands.

util/pyclient/bft_msgs.py
```diff
@@ -53,13 +53,14 @@ def __init__(self, message):
 ReplyHeader = namedtuple('ReplyHeader', ['span_context_size', 'primary_id',
     'req_seq_num', 'length', 'rsi_length'])
 
-def pack_request(client_id, req_seq_num, read_only, timeout_milli, cid, msg, pre_process=False, span_context=b''):
+def pack_request(client_id, req_seq_num, read_only, timeout_milli, cid, msg, pre_process=False, span_context=b'', special_flags=None):
```
andrewjstone, line 56: This API addition is way too low-level. It requires the caller to know the hex value of the special flags. While that could be replaced with some global constant, it's better to instead not expose the flags, which are a low level detail of the protocol, and can be changed without needing to change the `pack_request` API.

Instead you should expose the behavior you want. In this case I think it would be better to continue with the pattern here and take another boolean that defaults to false : `wedge=false` or `stop_consensus=false`, and then sent the proper flag value in the function. This will also make the code at the call sites be more legible such that readers can understand the behavior without having to figure out what special flags are.

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
```diff
@@ -41,6 +41,17 @@ int InternalCommandsHandler::execute(uint16_t clientId,
                                      uint32_t &outActualReplySize,
                                      uint32_t &outActualReplicaSpecificInfoSize,
                                      concordUtils::SpanWrapper &span) {
+  bool wedgeCommand = flags & 0x9;
```
andrewjstone, line 44: `0x9` doesn't really work as a flag, as it uses two bits: bit 0, and bit 3. If someone added a flag of `0x8`, then `0x8` and `0x1` (read-only) combined would match a wedge command and lead to incorrect behavior. You should just use `0x4` or `0x8` instead.

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
```diff
@@ -41,6 +41,17 @@ int InternalCommandsHandler::execute(uint16_t clientId,
...
+  if (wedgeCommand) {
+    LOG_INFO(m_logger, "A wedge command has been called seqNum: " << sequenceNum);
```
andrewjstone, line 46: Please use the KVLOG wrapper for sequenceNum so this log message looks the same as others. I realize this is just test output, but it's good to get into the habit.

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: let's leach it for a future PR :)

tests/apollo/test_skvbc_control_commands.py
yontyon: It was a mistake.
Fixed

util/pyclient/bft_msgs.py
yontyon: fixed

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
yontyon: I have removed the flag completely (as wedge command should be a write command)

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
yontyon: done

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2382,8 +2382,11 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
 void ReplicaImp::onSeqNumIsSuperStable(SeqNum newSuperStableSeqNum) {
   auto seq_num_to_stop_at = controlStateManager_->getCheckpointToStopAt();
   if (seq_num_to_stop_at.has_value() && seq_num_to_stop_at.value() == newSuperStableSeqNum) {
-    if (getRequestsHandler()->getControlHandlers())
+    LOG_INFO(GL, "control state manager marked seqNum: " << newSuperStableSeqNum);
```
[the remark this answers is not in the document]
yontyon: done

yontyon:
Guys, if you have ideas for more testing, please let me know :)

teoparvanov:
> 
> 
> Guys, if you have ideas for more tests, please let me know :)

As an edge case, I wonder what would happen if you send a request while a minority (<f) replicas are doing state transfer, so they ignore the wedge command.

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+        Sends a wedge command and check that the system stops from processing new requests.
+        Note that in this test we assume no failures and synchronized network.
+        The test does the following:
+        1. A client sends a wedge command
+        2. The client verify that the system reached to a super stable checkpoint
+        3. The client tries to initiate a new write bft command and fails
+    """
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 4 and c == 0)
```
teoparvanov, line 56: If you want to run this only for a single configuration, let's make it `lambda n, f, c: n == 7`, as it's closer to the real environment.
Actually, why don't you want to run the test for all supported configs?

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
+# Concord
+#
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
```
teoparvanov, line 3: nit - 2020

tests/apollo/CMakeLists.txt
```diff
@@ -76,5 +80,6 @@ foreach(STORAGE_TYPE ${STORAGE_TYPES})
             "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_ro_replica ${TEST_OUTPUT}"
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
   endif()
+
```
teoparvanov, line 83: nit - remove empty line

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 4 and c == 0)
+    async def test_wedge_command(self, bft_network):
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        client = bft_network.random_client()
+
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=0)
+
+        last_bid = skvbc.parse_reply(await client.read(skvbc.get_last_block_req()))
+        await client.write(skvbc.write_req([], [], block_id=last_bid, wedge_command=True))
```
teoparvanov, line 65: no need to set the `block_id` field - it's only used if your read-set is non-empty

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+        The test does the following:
+        1. Start all replicas but 1
+        2. A client sends a wedge command
+        3. Validate that all started replicas reached to the next next checkpoint
+        4. Start the late replica
+        5. Validate that the late replica completed the state transfer
+        6. Validate that all replicas stopped at the super stable checkpoint and that new commands are not being processed
+    """
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 4 and c == 0)
```
teoparvanov, line 89: same here - why not run for all configs? or at least for `n==7` ?

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+    async def test_wedge_command_with_state_transfer(self, bft_network):
+        initial_prim = 0
+        late_replica = bft_network.random_set_of_replicas(1, {initial_prim})
```
teoparvanov, line 92: `late_replicas`

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+        on_time_replicas = bft_network.all_replicas(without=late_replica)
+        bft_network.start_replicas(on_time_replicas)
+
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=0)
+
+        client = bft_network.random_client()
+        last_bid = skvbc.parse_reply(await client.read(skvbc.get_last_block_req()))
+        await client.write(skvbc.write_req([], [], block_id=last_bid, wedge_command=True))
```
teoparvanov, line 102: same here - no need to set `block_id`

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+                        if checkpoint_after == checkpoint_before + 2:
+                            break
+
+        bft_network.start_replicas(late_replica)
+
+        await bft_network.wait_for_state_transfer_to_start()
+        for r in late_replica:
+            await bft_network.wait_for_state_transfer_to_stop(initial_prim,
+                                                              r,
+                                                              stop_on_stable_seq_num=False)
```
teoparvanov, line 118: isn't `stop_on_stable_seq_num=False` by default ?

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+    viewChangeTimeoutMilli = "10000"
+    path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
+    return [path,
+            "-k", KEY_FILE_PREFIX,
+            "-i", str(replica_id),
+            "-s", statusTimerMilli,
+            "-v", viewChangeTimeoutMilli,
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower()
+                    in set(["true", "on"])
+            else "",
```
teoparvanov, line 41: formatting: the `else` should be under the if

teoparvanov:
> 
> 
> Guys, if you have ideas for more tests, please let me know :)

What happens if the primary is down and at this point you send a wedge command? Will the wedge command be ignored? Will it be processed after the view change completes? I think this needs to be tested (in a separate PR, of course). WDYT?

andrewjstone:
> > Guys, if you have ideas for more tests, please let me know :)
> 
> What happens if the primary is down and at this point you send a wedge command? Will the wedge command be ignored? Will it be processed after the view change completes? I think this needs to be tested (in a separate PR, of course). WDYT?

One thing to note about this code is that its happy path only. We specifically aren't building a fault tolerant upgrade at this point. We could probably make it so that the wedge would be processed after the view change completes, and should, I'm just not sure it's in scope for this change. The immediate, short term goal is if anything goes wrong to do a manual recovery. 

We definitely need to have a chat about that...

With all the above being said, I actually don't think its that hard to make sure the wedge is processed after view change completes and we'll need to consider that in the future. It should also be noted that the `superStableCheckpoint` i.e. N out of N checkpoints received will change to normal N-F in the future as well, when we can tolerate failures. This is just an intermediate step.

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+            with trio.fail_after(seconds=30):
+                while True:
+                    with trio.move_on_after(seconds=1):
+                        checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=replica_id)
+                        if checkpoint_after == checkpoint_before + 2:
+                            break
+
+        await self.validate_stop_on_super_stable_checkpoint(bft_network, skvbc)
+
+    """
```
andrewjstone, line 77: The doc comment should be placed right under the function definition, not before it.

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
```diff
@@ -111,6 +111,10 @@ bool InternalCommandsHandler::executeWriteCommand(uint32_t requestSize,
                << " PRE_PROCESS_FLAG=" << ((flags & MsgFlag::PRE_PROCESS_FLAG) != 0 ? "true" : "false")
                << " HAS_PRE_PROCESSED_FLAG=" << ((flags & MsgFlag::HAS_PRE_PROCESSED_FLAG) != 0 ? "true" : "false"));
 
+  if (writeReq->header.type == WEDGE) {
+    LOG_INFO(m_logger, "A wedge command has been called in seqNum: " << KVLOG(sequenceNum));
```
andrewjstone, line 115: Nit: No need for "in seqNum". KVLOG will print the output as `sequenceNum: <ActualValue>`.

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+            "-s", statusTimerMilli,
+            "-v", viewChangeTimeoutMilli,
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower()
+                    in set(["true", "on"])
+            else "",
+            "-t", os.environ.get('STORAGE_TYPE')]
+
+
+class SkvbcControlCommandsTest(unittest.TestCase):
+    """
```
andrewjstone, line 46: The doc comment should be under the function definition, not before it.

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,155 @@
...
+        for replica_id in on_time_replicas:
+            with trio.fail_after(seconds=30):
+                while True:
+                    with trio.move_on_after(seconds=1):
+                        checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=replica_id)
+                        if checkpoint_after == checkpoint_before + 2:
+                            break
+
+        bft_network.start_replicas(late_replica)
+
```
andrewjstone, line 113: I think it might be useful to add an additional check here that tries to send another request even though we haven't seen a super stable checkpoint yet. At this point each replica should stop accepting requests since it has reached the sequence number to stop at. The waiting for NofN is just to provide the control state callback to start the upgrade.

yontyon: I think that in this case, the late replica will initiate a view change (eventually).
@teoparvanov, what do you think?

yontyon:
> > Guys, if you have ideas for more tests, please let me know :)
> 
> What happens if the primary is down and at this point you send a wedge command? Will the wedge command be ignored? Will it be processed after the view change completes? I think this needs to be tested (in a separate PR, of course). WDYT?

As @andrewjstone explained, we are mainly interested in the happy case only (i.e. all replicas are up and the network is synchronized).
Yet, following the design, if we managed to execute the command, eventually, all replicas (even thus that were updated by the state transfer) should stop at this super stable checkpoint.
I also think that would be a good test case for future PR.

tests/apollo/test_skvbc_control_commands.py
yontyon: Fixed. Run for all configurations

tests/apollo/test_skvbc_control_commands.py
yontyon: fixed

tests/apollo/CMakeLists.txt
yontyon: fixed

tests/apollo/test_skvbc_control_commands.py
yontyon: fixed

tests/apollo/test_skvbc_control_commands.py
yontyon: Fixed. Run for all configurations

tests/apollo/test_skvbc_control_commands.py
yontyon: fixed

tests/apollo/test_skvbc_control_commands.py
yontyon: fixed

tests/apollo/test_skvbc_control_commands.py
yontyon: set to true

tests/apollo/test_skvbc_control_commands.py
yontyon: fixed

tests/apollo/test_skvbc_control_commands.py
yontyon: fixed

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
yontyon: fixed

tests/apollo/test_skvbc_control_commands.py
yontyon: fixed

PR #791, opened 2020-07-29
Add span to addBlock method

f-squirrel:
Reason:
The method `addBlock` is overridden in the product and its performance is measured by Opentracing.
The initial implementation did not change the base method but used a tricky workaround based on
saving the parent span between the calls of the `execute` method.

After introducing Pre-execution the method `execute` may be called in the context of different threads.
Unfortunately, spans' behavior is undefined while used together with multithreading.

Solution:
The solution is to avoid saving spans and to create them on demand which involves the change in the base API.

Changes:
* Added `SpanWrapper` to the `addBlock` method.
* Minor changes in the Opentracing utilities.

### DIFF
--- a/kvbc/include/db_interfaces.h
+++ b/kvbc/include/db_interfaces.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include "OpenTracing.hpp"
 #include "kv_types.hpp"
 #include "status.hpp"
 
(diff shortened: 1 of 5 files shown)


kvbc/include/db_interfaces.h
```diff
@@ -34,7 +35,9 @@ class ILocalKeyValueStorageReadOnly {
  */
 class IBlocksAppender {
  public:
-  virtual concordUtils::Status addBlock(const SetOfKeyValuePairs& updates, BlockId& outBlockId) = 0;
+  virtual concordUtils::Status addBlock(const SetOfKeyValuePairs& updates,
+                                        BlockId& outBlockId,
+                                        const concordUtils::SpanWrapper& parent_span = concordUtils::SpanWrapper{}) = 0;
```
dartdart26, line 40: Can we possibly consider using overloads (one with and one without the span) instead of default parameters for virtual methods? It might make the code more generic, clearer and safer.

PR #793, opened 2020-07-29
[WIP] Apollo SkvbcTracker tracker support for predeployed blockchain

tdimitrov:
The purpose of the PR is to be able to run linearizability checks against predeployed blockchain. The current implementation doesn't support this for two main reasons:
1. It assumes that the test is run against empty blockchain, created by TesterReplica.
2. It assumes that all blocks are result of operations which are run via the SkvbcTracker instanse.

If one wants to run linearizability checks against real blockchain these conditions won't be satisfied. The blockchain may or may not contain blocks which were created outside the tracker and they will cause PhantomBlock exceptions.

The approach implemented in this PR is to be able to set initial block id for the SkvbcTracker and all linearizability checks to be performed from this block to the end. With this functionality the following can be accomplished:
- Linearizability checks can be executed against real blockchain already containing one or more blocks.
- Run multiple Apollo tests against single blockchain without clearing it after each test.

The PR 'sort of' works and is marked as WIP, but I do want to get some feedback for it.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -182,7 +182,7 @@ class BftTestNetwork:
 
     def __init__(self, is_existing, origdir,
                  config, testdir, builddir, toolsdir,
-                 procs, replicas, clients, metrics):
+                 procs, replicas, clients, metrics, new_client_cb):
         self.is_existing = is_existing
         self.origdir = origdir
         self.config = config
@@ -194,9 +194,13 @@ class BftTestNetwork:
         self.clients = clients
         self.metrics = metrics
         self.reserved_clients = {}
+        if new_client_cb:
+            self.new_client_cb = new_client_cb
+        else:
+            self.new_client_cb = self._create_new_udp_client
 
     @classmethod
-    def new(cls, config):
+    def new(cls, config, new_client_cb=None):
         builddir = os.path.abspath("../../build")
         toolsdir = os.path.join(builddir, "tools")
         testdir = tempfile.mkdtemp()
@@ -211,11 +215,12 @@ class BftTestNetwork:
             replicas=[bft_config.Replica(i, "127.0.0.1", 3710 + 2*i, 4710 + 2*i)
                 for i in range(0, config.n + config.num_ro_replicas)],
             clients = {},
-            metrics = None
+            metrics = None,
+            new_client_cb = new_client_cb
         )
 
         #copy loggging.properties file
-        shutil.copy(os.path.abspath("../simpleKVBC/scripts/logging.properties"), testdir);
+        shutil.copy(os.path.abspath("../simpleKVBC/scripts/logging.properties"), testdir)
 
         print("Running test in {}".format(bft_network.testdir))
 
@@ -228,7 +233,7 @@ class BftTestNetwork:
         return bft_network
 
     @classmethod
-    def existing(cls, config, replicas, clients):
+    def existing(cls, config, replicas, clients, new_client_cb=None):
         bft_network = cls(
             is_existing=True,
             origdir=None,
@@ -239,7 +244,8 @@ class BftTestNetwork:
             procs={r.id: r for r in replicas},
             replicas=replicas,
             clients={i: clients[i] for i in range(len(clients))},
-            metrics=None
+            metrics=None,
+            new_client_cb=new_client_cb
         )
 
         bft_network._init_metrics()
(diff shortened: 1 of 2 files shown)


tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -46,6 +46,12 @@ async def wrapper(*args, **kwargs):
                 skvbc = kvbc.SimpleKVBCProtocol(bft_network)
                 init_state = skvbc.initial_state()
                 tracker = SkvbcTracker(init_state, skvbc, bft_network, pre_exec_enabled, no_conflicts)
+
+                if uses_real_blockchain:
+                    client = await bft_network.new_client()
+                    initial_block_id = await tracker.get_last_block_id(client)
+
+                tracker.set_initial_block_id(initial_block_id)
```
andrewjstone, line 54: is `initial_block_id` set before this is called if `uses_real_blockchain = False` ?

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -511,9 +524,11 @@ def get_missing_blocks(self, last_block_id):
         blocks and call self.fill_in_missing_blocks().
 
         After missing blocks are filled in, successful reads can be linearized.
+
+        All blocks are fetched from the genesis (1) to last_known_block
         """
         missing_blocks = set()
-        for i in range(self.last_consecutive_block + 1, self.last_known_block):
```
andrewjstone, line 516: This feels wrong. We shouldn't need to retrieve blocks before they are tracked.

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -573,7 +588,9 @@ def _match_filled_blocks(self):
                     break
                 else:
                     unmatched.append(req)
-            if not success:
+            # Don't raise PhantomBlockError for blocks created before the
+            # instantiation of the tracker
+            if not success and block_id > self.initial_block_id:
```
andrewjstone, line 593: This also feels wrong. We shouldn't need to check blocks prior to self.initial_block_id.

andrewjstone:
> Hi @tdimitrov. I think this is a great idea!
> 
> Your commit description is very clear. Thank you.
> 
> However, I think we want to make the implementation more like the description. For instance, in a few places, you start from the genesis block during checking. Instead, I think we should not pull blocks before the initial block used in the tracker, and also only run the check over those blocks. Checks before that both waste time, and also may lead to unexpected errors. It's possible the latter won't happen, but without thorough analysis we can't be sure. We should just skip the expense and complexity of retrieving and testing blocks before the ones that are tracked.
> 
> I'm happy to chat more with you offline.

I thought more about this, and I believe you did it this way because you don't know the current value of the keys without the prior blocks. However, in most of these tests, the key space is limited, so I believe we can start from the initial block by reading the current values of all known keys in some initialization routine and then using them to set [`initial_kvpairs`](https://github.com/vmware/concord-bft/blob/736e0a4898b195160b3692cea5bf7b8bdbeb71fe/tests/apollo/util/skvbc_history_tracker.py#L386). This would allow us to periodically run linearizability checks against long running blockchains without a reset. However, the flip side of this is that if the keyspace is very large, we now have to read all possible keys. In essence it's a tradeoff of the size of the key space vs the pre-existing blocks. 

However, with our typical workload, the keyspace is very large, and there are many new keys, which lends itself to your current strategy as long as there aren't an enormous amount of already generated blocks. Maybe we can just do a keylist during initialization of any active keys, rather than pull all blocks before the start of tracking. Even if this is heavy, it's guaranteed to be lighter weight than getting the blocks.

tests/apollo/util/bft.py
```diff
@@ -182,7 +182,7 @@ def __exit__(self, *args):
 
     def __init__(self, is_existing, origdir,
                  config, testdir, builddir, toolsdir,
-                 procs, replicas, clients, metrics):
+                 procs, replicas, clients, metrics, new_client_cb):
```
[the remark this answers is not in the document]
tdimitrov: I agree about the rename. Looking at it again today - it's misleading. I'll fix it after Teo shares his opinion about the keyword arg.

tests/apollo/util/skvbc_history_tracker.py
tdimitrov: Yes, it's set to 0 in the constructor. 

Ideally this check should have been in the constructor itself, but it's a chicken-egg problem. get_last_block_id() is part of the tracker and I think copy-pasting its implementation is worse than a setter.

tdimitrov:
@andrewjstone I completely understand why some of the changes (the for loops) look wrong. Actually they are result of some trial-error test. 
In nutshell - yes, I need the KV currently in the blockchain or the tracker fails. My assumption was that the linearizability checks will be executed against real but still dedicated for the test blockchain, which will not contain lots of values and we can safely prefetch them.

Let's discuss this offline and I'll update the PR with a summary of our chat.

teoparvanov:
I suggest to close this PR for now and focus on linearizability checks starting from a clean blockchain. To achieve this in Hermes, we could for example run every Apollo test case in a separate suite.

tdimitrov:
> I suggest to close this PR for now and focus on linearizability checks starting from a clean blockchain. To achieve this in Hermes, we could for example run every Apollo test case in a separate suite.

Let's keep it open for a bit more. @andrewjstone gave me some good ideas today and I want to try them here.

PR #794, opened 2020-07-30
RocksDB parameter tunning to improve performance

biplavk:
Based on multiple experiments with different parameters and values. These parameter and values combination provides optimal performance for heavy write workload.


biplavk:
Yes , I ran experiments on real system with this configuration. I expect a performance improvement of around 9% and this should increase with more heavier workload. Yes , we can run it before merging to master.

andrewjstone:
This is really great stuff!

@biplavk and I discussed offline and we are going to be disabling mmap, because I'm worried about uncontrolled paging of data and memory overhead for very large databases. He doesn't think it will have much if any impact on the perf measurements as those are mostly based around writes and not reads. Will merge after that change.

andrewjstone:
Merged. I should have squashed :(

PR #795, opened 2020-07-30
[Apollo] Checkpoint propagation after restarting all replicas under load

shgandhi:
In this test, under a constant time-service like load, we first trigger a checkpoint. We then stop all replicas in a random order
with a delay in between of view_change_timeout/n, as we don't want replicas to start a view change while they stop. We then restart replicas in a random order with a delay of 10s, we start the primary at the end, so as to avoid view change being triggered while replicas restart. We verify checkpoint persisted upon restart and then trigger another checkpoint. We make sure checkpoint is propagated to all the replicas.

### DIFF
--- a/tests/apollo/test_skvbc_backup_restore.py
+++ b/tests/apollo/test_skvbc_backup_restore.py
@@ -19,7 +19,7 @@ from util import skvbc as kvbc
 from util.bft import with_trio, with_bft_network, with_constant_load, KEY_FILE_PREFIX
 
 
-def start_replica_cmd(builddir, replica_id):
+def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="10000"):
     """
     Return a command that starts an skvbc replica when passed to
     subprocess.Popen.
@@ -28,15 +28,20 @@ def start_replica_cmd(builddir, replica_id):
     Note each arguments is an element in a list.
     """
     statusTimerMilli = "500"
-    viewChangeTimeoutMilli = "10000"
     path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
     return [path,
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
             "-s", statusTimerMilli,
-            "-v", viewChangeTimeoutMilli,
+            "-v", view_change_timeout_milli,
             "-p",
             "-t", os.environ.get('STORAGE_TYPE')]
+
+
+def start_replica_cmd_with_vc_timeout(vc_timeout):
+    def wrapper(*args, **kwargs):
+        return start_replica_cmd(*args, **kwargs, view_change_timeout_milli=vc_timeout)
+    return wrapper
 
 
 class SkvbcBackupRestoreTest(unittest.TestCase):
@@ -76,10 +81,10 @@ class SkvbcBackupRestoreTest(unittest.TestCase):
         )
 
         # stop n replicas in a random order with a delay of 10s in between
-        stopped_replicas = await self._stop_random_replicas_with_delay(bft_network, n)
+        stopped_replicas = await self._stop_random_replicas_with_delay(bft_network, delay=5)
 
         # start stopped replicas in a random order with a delay of 10s in between
-        await self._start_random_replicas_with_delay(bft_network, stopped_replicas)
+        await self._start_random_replicas_with_delay(bft_network, stopped_replicas, current_primary, delay=5)
 
         # verify checkpoint persistence
         await bft_network.wait_for_replicas_to_checkpoint(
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -101,19 +106,128 @@ async def test_checkpoint_propagation_after_restarting_replicas(self, bft_networ
...
+            verify_checkpoint_persistency=False,
+            assert_state_transfer_not_started=False
+        )
+
     @staticmethod
-    async def _stop_random_replicas_with_delay(
-            bft_network, replica_set_size, delay=5, exclude_replicas=None):
-        random_replicas = bft_network.random_set_of_replicas(size=replica_set_size, without=exclude_replicas)
-        for replica in random_replicas:
+    async def _stop_random_replicas_with_delay(bft_network, delay=10, exclude_replicas=None):
```
teoparvanov, line 191: please reduce the delay to 5s, otherwise Travis timeouts

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -101,19 +106,128 @@ async def test_checkpoint_propagation_after_restarting_replicas(self, bft_networ
...
+            print("stopping replica: ", replica)
             bft_network.stop_replica(replica)
             await trio.sleep(delay)
-        return list(random_replicas)
+        return list(all_replicas)
 
     @staticmethod
-    async def _start_random_replicas_with_delay(
-            bft_network, stopped_replicas, delay=5):
+    async def _start_random_replicas_with_delay(bft_network, stopped_replicas, initial_primary, delay=10):
```
teoparvanov, line 201: same here - `delay=10`

shgandhi: With lower delay, say, 5s, 2f + 1 replicas may come up in the initial view (vc_timeout is set to be 20s to give time to restarted replicas to recover), hence, not triggering a view change. If we reduce vc_timeout, say to 10s, we end up triggering unexpected view changes

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -101,19 +106,198 @@ async def test_checkpoint_propagation_after_restarting_replicas(self, bft_networ
...
+                      selected_configs=lambda n, f, c: n == 7)
+    @with_constant_load
+    async def test_checkpoint_propagation_after_restarting_all_replicas_under_load(self, bft_network, skvbc, nursery):
+        """
+        Here we trigger a checkpoint, restart all replicas in a random order with 10s delay in-between,
+        both while stopping and starting. We verify checkpoint persisted upon restart and then trigger
+        another checkpoint. We make sure checkpoint is propagated to all the replicas.
+        1) Given a BFT network, we make sure all nodes are up
+        2) Send sufficient number of client requests to trigger checkpoint protocol
+        3) Stop all replicas in a random order (with 10s delay in between)
```
HristoStaykov, line 120: 10/n seconds

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -101,19 +106,198 @@ async def test_checkpoint_propagation_after_restarting_replicas(self, bft_networ
...
+
+        self.assertEqual(len(bft_network.procs), n, "Make sure all replicas are up initially.")
+
+        current_primary = await bft_network.get_current_primary()
+
+        initial_view = 0
+
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+
+        await skvbc.fill_and_wait_for_checkpoint(
```
HristoStaykov, line 139: Though it is unlikely that we will get 2 checkpoints here, for consistency I think we should use _fill_and_wait_for_checkpoint_under_constant_load.

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -101,19 +106,198 @@ async def test_checkpoint_propagation_after_restarting_replicas(self, bft_networ
...
+            initial_nodes=bft_network.all_replicas(),
+            num_of_checkpoints_to_add=1,
+            verify_checkpoint_persistency=False
+        )
+
+        checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=current_primary)
+        self.assertEqual(checkpoint_before + 1, checkpoint_after)
```
HristoStaykov, line 146: Again for consistency I suggest to use assertGreaterEqual and to emphasize that with constant load we could get an additional checkpoint.

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -101,19 +106,198 @@ async def test_checkpoint_propagation_after_restarting_replicas(self, bft_networ
...
+        # verify checkpoint persistency
+        await bft_network.wait_for_replicas_to_checkpoint(bft_network.all_replicas(),
+                                                          expected_checkpoint_num=lambda ecn: ecn >= checkpoint_after)
+
+        await self._fill_and_wait_for_checkpoint_under_constant_load(skvbc,
+                                                                     bft_network,
+                                                                     initial_nodes=bft_network.all_replicas(),
+                                                                     num_of_checkpoints_to_add=1,
+                                                                     verify_checkpoint_persistency=False,
+                                                                     assert_state_transfer_not_started=False)
```
HristoStaykov, line 182: I suggest we get the checkpoint_before and the checkpoint_after and assert that checkpoint_after >= checkpoint_before.

shgandhi:
Thanks for the review @HristoStaykov! I have made the changes you suggested.

PR #796, opened 2020-07-30
[Apollo] KeyError when reading checkpoint metric

teoparvanov:
In some cases (for example after replica restart), the metrics may not be immediately available. This results in KeyError, which needs to be retried, if it occurred.


Liorgal28:
LGTM!

PR #797, opened 2020-07-31
Support for external create client function in Apollo BftTestNetwork

tdimitrov:
This PR is stripped version of https://github.com/vmware/concord-bft/pull/793

It contains only the first commit, which implemented an external factory method for creating BFT client. The idea is to use either the UdpClient in concord-bft or other user provided implementation.

Please check the initial PR for more context.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -182,7 +182,7 @@ class BftTestNetwork:
 
     def __init__(self, is_existing, origdir,
                  config, testdir, builddir, toolsdir,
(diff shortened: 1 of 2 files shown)


tests/apollo/util/bft.py
```diff
@@ -194,9 +194,13 @@ def __init__(self, is_existing, origdir,
         self.clients = clients
         self.metrics = metrics
         self.reserved_clients = {}
+        if new_client_factory:
```
teoparvanov, line 197: Let's call this simply `client_factory`

PR #798, opened 2020-08-01
[Apollo] Checkpoint propagation after restarting n-f replicas under load

shgandhi:
In this test, we trigger a checkpoint, restart n-f replicas in a random order with a delay in-between, both while stopping and starting. We verify checkpoint persisted upon restart and then trigger  another checkpoint. We make sure checkpoint is propagated to all the replicas

### DIFF
--- a/tests/apollo/test_skvbc_backup_restore.py
+++ b/tests/apollo/test_skvbc_backup_restore.py
@@ -52,13 +52,13 @@ class SkvbcBackupRestoreTest(unittest.TestCase):
     @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
     async def test_checkpoint_propagation_after_restarting_replicas(self, bft_network):
         """
-        Here we trigger a checkpoint, restart all replicas in a random order with 10s delay in-between,
+        Here we trigger a checkpoint, restart all replicas in a random order with 5s delay in-between,
         both while stopping and starting. We verify checkpoint persisted upon restart and then trigger
         another checkpoint. We make sure checkpoint is propagated to all the replicas.
         1) Given a BFT network, we make sure all nodes are up
         2) Send sufficient number of client requests to trigger checkpoint protocol
-        3) Stop all replicas in a random order (with 10s delay in between)
-        4) Start all replicas in a random order (with 10s delay in between)
+        3) Stop all replicas in a random order (with 5s delay in between)
+        4) Start all replicas in a random order (with 5s delay in between)
         5) Make sure the initial view is stable
         6) Send sufficient number of client requests to trigger another checkpoint
         7) Make sure checkpoint propagates to all the replicas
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -80,10 +80,10 @@ async def test_checkpoint_propagation_after_restarting_replicas(self, bft_networ
             verify_checkpoint_persistency=False
         )
 
-        # stop n replicas in a random order with a delay of 10s in between
+        # stop n replicas in a random order with a delay of 5s in between
```
salieri11, line 83: Was it sensitive to the time? What was the reason to change the delay?

shgandhi: If replicas are stopped with 10 seconds delay in between, they can start view change while stopping after the view change timer expires.

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -183,7 +183,112 @@ async def test_checkpoint_propagation_after_restarting_all_replicas_under_load(s
...
+        n = bft_network.config.n
+        f = bft_network.config.f
+
+        initial_primary = await bft_network.get_current_primary()
+
+        # choose a random set of f replicas to stop
+        f_replicas_stopped_early = bft_network.random_set_of_replicas(f, without={initial_primary})
+        bft_network.stop_replicas(f_replicas_stopped_early)
+
+        replicas_up = [i for i in bft_network.all_replicas() if i not in f_replicas_stopped_early]
```
teoparvanov, line 218: I think you could also say, equivalently:
```python
replicas_up = bft_network.all_replicas(without=f_replicas_stopped_early)
```

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -183,7 +183,112 @@ async def test_checkpoint_propagation_after_restarting_all_replicas_under_load(s
...
+            num_of_checkpoints_to_add=1,
+            verify_checkpoint_persistency=False,
+            assert_state_transfer_not_started=False)
+
+        checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=initial_primary)
+        self.assertGreaterEqual(checkpoint_after, checkpoint_before + 1)
+
+        # stop n replicas in a random order with a delay of 10/n seconds in between, so that
+        # all replicas are stopped by 10 seconds, and no view change is triggered during stopping
+        # of replicas.
```
teoparvanov, line 239: Why avoid the view change while stopping the replicas? In a real-world scenario of backup-restore, we can't guarantee that the primary won't be stopped before the others, or the length of the delays.

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -183,7 +183,112 @@ async def test_checkpoint_propagation_after_restarting_all_replicas_under_load(s
                                                                      assert_state_transfer_not_started=False)
 
         next_checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=current_primary)
-        self.assertGreaterEqual(next_checkpoint_before + 1, next_checkpoint_after)
+        self.assertGreaterEqual(next_checkpoint_after, next_checkpoint_before + 1)
+
+    @with_trio
```
teoparvanov, line 188: Given this is an advanced scenario, the focus should be on running it locally and manually reviewing the logs, to make sure everything happens in the right order. I'd suggest to not enable it in CI, as it takes a long time to execute, and also it may be unstable. To do so, you can add the following decorator on top of the test method:
```python
@unittest.skip("Advanced manual scenario - not part of CI")
```

tests/apollo/test_skvbc_backup_restore.py
shgandhi: Yes, that makes sense.

tests/apollo/test_skvbc_backup_restore.py
shgandhi: done.

PR #800, opened 2020-08-02
Stabilize control state tests

yontyon:
After enabling Apollo tests for the wedge command feature (#790) we discovered they were unstable for configurations with n > 4 replicas and we had to revert them (#802).

We suspect that due to the fact that CI runs with UDP configuration, part of the replicas were not able to receive n/n checkpoint messages.
For that, we added a retransmitting mechanism that is activated once the wedge flag is raised by the command handler.
We believe that this addition will help to stabilize the feature even when facing network errors and packages dropping.

### DIFF
--- a/tests/apollo/test_skvbc_control_commands.py
+++ b/tests/apollo/test_skvbc_control_commands.py
@@ -42,7 +42,7 @@ def start_replica_cmd(builddir, replica_id):
 
 class SkvbcControlCommandsTest(unittest.TestCase):
     @with_trio
-    @with_bft_network(start_replica_cmd)
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
     async def test_wedge_command(self, bft_network):
         """
              Sends a wedge command and check that the system stops from processing new requests.
@@ -72,7 +72,7 @@ class SkvbcControlCommandsTest(unittest.TestCase):
 
     # @unittest.skip("Test case assumes the happy path")
     @with_trio
-    @with_bft_network(start_replica_cmd)
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
     async def test_wedge_command_with_state_transfer(self, bft_network):
         """
             This test checks that even a replica that received the super stable checkpoint via the state transfer mechanism
--- a/tests/apollo/test_skvbc_control_commands.py
+++ b/tests/apollo/test_skvbc_control_commands.py
@@ -70,7 +70,6 @@ class SkvbcControlCommandsTest(unittest.TestCase):
 
         await self.validate_stop_on_super_stable_checkpoint(bft_network, skvbc)
 
-    # @unittest.skip("Test case assumes the happy path")
     @with_trio
     @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
     async def test_wedge_command_with_state_transfer(self, bft_network):
--- a/tests/apollo/test_skvbc_control_commands.py
+++ b/tests/apollo/test_skvbc_control_commands.py
@@ -1,0 +1,151 @@
+# Concord
+#
+# Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+#
+# This product is licensed to you under the Apache 2.0 license (the "License").
+# You may not use this product except in compliance with the Apache 2.0 License.
+#
+# This product may include a number of subcomponents with separate copyright
+# notices and license terms. Your use of these subcomponents is subject to the
+# terms and conditions of the subcomponent's license, as noted in the LICENSE
+# file.
+import os.path
+import random
+import unittest
+from os import environ
+
+import trio
+
+from util import blinking_replica
+from util import skvbc as kvbc
+from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
+
+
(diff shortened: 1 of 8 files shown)


f-squirrel:
@yontyon , AFAIK, all the CI tests run on the same machine within the same docker container.
How come some of the messages are not being delivered?
If we do think that UDP diagrams are lost while running the CI we should probably switch to TCP that already has a retransmission mechanism.
WDYT?

yontyon:
@f-squirrel, I don't know how it is possible that some of the messages are not begin delivered.
The fact is that adding the retransmitting mechanism does stabilize the test, so I guess this is what actually happens.

I also know that we run on UDP in purpose because it may simulate an unstable network (in which packages are being lost) and hele up to identify problems and bugs.

f-squirrel:
@yontyon , from my experience testing anything related to networking, has to be done in a controlled environment.
A good example is the set of tests using `iptables` created by @teoparvanov.
IMO, adding timeout or retransmission w/o understanding the nature of the problem only makes our tests less stable and predictable.
I understand that this is beyond the scope of the PR but I do not like the idea of retransmission on top of UDP.

P.S. if we assume some packet loss then we need to explicitly test it together with the retransmission.

teoparvanov:
> 
> 
> @yontyon , from my experience testing anything related to networking, has to be done in a controlled environment.
> A good example is the set of tests using `iptables` created by @teoparvanov.
> IMO, adding timeout or retransmission w/o understanding the nature of the problem only makes our tests less stable and predictable.
> I understand that this is beyond the scope of the PR but I do not like the idea of retransmission on top of UDP.
> 
> P.S. if we assume some packet loss then we need to explicitly test it together with the retransmission.

Is it possible that the `CheckpointMsg`, instead of lost on the network, is ignored by the receiving replica? Maybe some corner case in `ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg)` ?

yontyon:
@teoparvanov, @f-squirrel, I've rechecked the logs.
The fact is that in one of the replicas (number 3) the following line happened only 5 times out of 7 (https://github.com/vmware/concord-bft/blob/c40d0d664a526e657ecb5e76edefc693c90bfe5b/bftengine/src/bftengine/ReplicaImp.cpp#L1430).
While for all other replicas, this line was printed 7 times (because 7 replicas have sent the checkpoint message).
Replica 3 did not die, it simply stopped from receiving the checkpoint messages and this is before any check on the message validity...
I have no idea why would that happen unless we lost messages when using UDP.
If you have any way to test it on Travis with TCP that would be the best (I've tried, but it looks like the TCP module is not well maintained)

teoparvanov:
> 
> 
> @teoparvanov, @f-squirrel, I've rechecked the logs.
> The fact is that in one of the replicas (number 3) the following line happened only 5 times out of 7 (
> 
> https://github.com/vmware/concord-bft/blob/c40d0d664a526e657ecb5e76edefc693c90bfe5b/bftengine/src/bftengine/ReplicaImp.cpp#L1430
> ).
> While for all other replicas, this line was printed 7 times (because 7 replicas have sent the checkpoint message).
> Replica 3 did not die, it simply stopped from receiving the checkpoint messages and this is before any check on the message validity...
> I have no idea why would that happen unless we lost messages when using UDP.
> If you have any way to test it on Travis with TCP that would be the best (I've tried, but it looks like the TCP module is not well maintained)

Can you maybe try to move the log statement on line 1424, just in case?

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,151 @@
...
+            with trio.fail_after(seconds=30):
+                while True:
+                    with trio.move_on_after(seconds=1):
+                        checkpoint_after = await bft_network.wait_for_checkpoint(replica_id=replica_id)
+                        if checkpoint_after == checkpoint_before + 2:
+                            break
+
+        await self.validate_stop_on_super_stable_checkpoint(bft_network, skvbc)
+
+    # @unittest.skip("Test case assumes the happy path")
```
teoparvanov, line 73: instead of commenting, let's just remove the line

tests/apollo/test_skvbc_control_commands.py
```diff
@@ -0,0 +1,151 @@
...
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_wedge_command_with_state_transfer(self, bft_network):
```
teoparvanov, line 76: It could be interesting to run this test as a separate test suite in Hermes (similar to [ApolloBftTests](https://gitlab.eng.vmware.com/blockchain/vmwathena_blockchain/blob/master/hermes/suites/apollo_bft_tests.py#L78)). Those tests run SKVBC in Concord, so they do use TLS instead of UDP. I expect this would help bring clarity on the missing messages problem (if still applicable) ...

yontyon: Yeah, that is a great idea!
Let's take it offline

yontyon:
@andrewjstone.
Basically, once we stopped at this n/n checkpoint we are at a dead end. We can get out of it only by external intervene.
Thus, disabling the retransmission is not such critical now, but I agree that this would be better.
Let's leave it to a future PR.
Thanks!

PR #805, opened 2020-08-05
Fix replica in inactive view calling tryToEnterView without enough view change messages

shgandhi:
Replicas if stopped in a random order with a sufficient delay in between can initiate view changes, but may end up being stopped in inactive view, without the view change ever stabilizing.
In such a scenario, a replica stopped in an inactive view will also start in an inactive view. If another replica initiates view change at the same time the replica in the inactive view restarts, the restarted replica, upon receiving the ViewChangeMsg may try to enter the new view without ever receiving 2f + 2c + 1 ViewChangeMsg. This is because for the replica
restarted from the persistent storage, the curView is set as the latest active view. If a replica was stopped in an inactive view, the curView is set to the latest active view and not the last known inactive view it moved into. So, if a replica was stopped with curView = 1, latest_active_view = 0, viewIsActive(curView) = False, it will be recovered with curView = 0,
latest_active_view = 0, viewIsActive(curView) = False, which is inconsistent, and eventually leads to the replica calling the function tryToEnterView without ever receiving 2f + 2c + 1 ViewChangeMsgs, and fails at the assert `v > myLatestActiveView` in tryToEnterView().

This flow can be verified by running the backup and restore test - test_checkpoint_propagation_after_restarting_all_replicas_under_load, with in-between delay for stopping replicas set to 10 seconds. The failure is persistent irrespective of the view number in which the replica was stopped.

To avoid the failure, when the replica is reloaded from the persistent storage, we update its curView to lastAgreedView + 1, if it was stopped in an inactive view.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2818,8 +2818,18 @@ ReplicaImp::ReplicaImp(const LoadedReplicaData &ld,
 
   ps_ = persistentStorage;
 
-  curView = ld.viewsManager->latestActiveView();
-  lastAgreedView = curView;
+  lastAgreedView = ld.viewsManager->latestActiveView();
+
+  if (ld.viewsManager->viewIsActive(lastAgreedView)) {
+    curView = lastAgreedView;
+  } else {
+    curView = lastAgreedView + 1;
+    ViewChangeMsg *t = ld.viewsManager->getMyLatestViewChangeMsg();
+    ConcordAssert(t != nullptr);
+    ConcordAssert(t->newView() == curView);
+    t->finalizeMessage();  // needed to initialize the VC message
+  }
+
   metric_view_.Get().Set(curView);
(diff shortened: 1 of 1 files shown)


teoparvanov:
@HristoStaykov could you please have an additional look, before this PR gets merged? Thanks!

HristoStaykov:
I would recommend asking Guy Gueta to review the change also, having in mind View Change is one of the most complicated parts of the protocol.

HristoStaykov:
Its a good idea to locally run the edge case tests that are not part of CI from apollo/test_skvbc_backup_restore.py and /apollo/test_skvbc_chaotic_startup.py to verify they pass successfully before merging the fix to master.

andrewjstone:
@HristoStaykov @teoparvanov Shruti and I had a discussion about this on Friday afternoon. We believe there is a deeper issue and this is not a sufficient fix. We can catch up later, but this shouldn't be merged as is.

shgandhi:
The fix has been updated. If a replica was recovered in an inactive view, its `curView` is updated to `lastAgreedView + 1`.

PR #806, opened 2020-08-05
CMF: Add support for python codegen

andrewjstone:
Generating python source code from CMF files is now supported. Example code can
be generated with the following command:

```
cd messages/compiler
./cmfc.py --input ../example.cmf --output example --language python
```

Tests can be run like so:

```
cd messages/compiler/python
pytest -q test_pygen.py
```

A future commit will hook the python and cpp tests into our cmake build.
This requires updating install_deps.sh and pushing a new version of the
build container to dockerhub.

The code generation for python is similar to that for C++ in that a
visitor is implemented that constructs code as the AST is walked. The
primary difference is that python is dynamically typed and therefore all
the serialization logic itself is implemented in serialize.py.

CMFSerializer and CMFDeserializer implement recursive serialization
functions that take a value and a list of "serializers". These are
strings or tuples that instruct the serialize and deserialize methods
how to recursively encode/decode the corresponding value. The
serializer lists are constructed by the visitor, thus making the visitor
code very simple.

One other thing to note is that due to the dynamic typing of python,
oneofs and optionals are implemented by simply setting the field to the
corresponding type or None in case of an empty optional.

This code also includes a bugfix for C++ that will ensure that booleans
can only be deserialized if they are a 0 or a 1.


messages/compiler/python/serialize.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def optional(self, val, serializers):
+        if val is None:
+            self.bool(False)
+        else:
+            self.bool(True)
+            self.serialize(val, serializers)
+
+    def oneof(self, val, msgs):
+        if val.__class__.__name__ in msgs.keys():
```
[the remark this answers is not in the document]
andrewjstone: Great catch! I meant to use OrderedDict, and think we should.

messages/compiler/python/serialize.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def int64(self):
+        if self.pos + 8 > len(self.buf):
+            raise NoDataLeftError()
+        val = struct.unpack_from('<q', self.buf, self.pos)
+        self.pos += 8
+        return val[0]
+
+    def string(self):
+        if self.pos + 4 > len(self.buf):
```
andrewjstone, line 268: This is redundant, as `uint32()` already does this check.

messages/compiler/python/serialize.py
```diff
@@ -0,0 +1,330 @@
...
+            raise NoDataLeftError()
+        size = self.uint32()
+        if self.pos + size > len(self.buf):
+            raise NoDataLeftError()
+        val = str(self.buf[self.pos:self.pos + size], 'utf-8')
+        self.pos += size
+        return val
+
+    def bytes(self):
+        if self.pos + 4 > len(self.buf):
```
andrewjstone, line 278: This is redundant, as uint32() already does this check.

messages/compiler/python/serialize.py
```diff
@@ -0,0 +1,330 @@
...
+        self.pos += bytes_read
+        return val
+
+    def kvpair(self, serializers):
+        key = self.deserialize(serializers)
+        val = self.deserialize(serializers[1:])
+        return (key, val)
+
+    def list(self, serializers):
+        if self.pos + 4 > len(self.buf):
```
andrewjstone, line 299: This is redundant, as uint32() already does this check.

messages/compiler/python/serialize.py
```diff
@@ -0,0 +1,330 @@
...
+            raise NoDataLeftError()
+        size = self.uint32()
+        if self.pos + size > len(self.buf):
+            raise NoDataLeftError()
+        return [self.deserialize(serializers) for _ in range(0, size)]
+
+    def map(self, serializers):
+        if self.pos + 4 > len(self.buf):
```
andrewjstone, line 307: This is redundant, as uint32() already does this check.

messages/compiler/python/serialize.py
```diff
@@ -0,0 +1,330 @@
...
+            for _ in range(0, size)
+        }
+
+    def optional(self, serializers):
+        if not self.bool():
+            return None
+        return self.deserialize(serializers)
+
+    def oneof(self, msgs):
+        if self.pos + 4 > len(self.buf):
```
andrewjstone, line 323: This is redundant, as uint32() already does this check.

messages/compiler/python/serialize.py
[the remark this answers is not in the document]
andrewjstone: Hmm. I don't think it actually matters here. Where it matters is in maps. We don't actually want insertion order. We want lexicographic order when serializing. This means that the output of any dict serialization must be sorted lexicographically.

robem:
Still, looks good to me.

andrewjstone:
@robem Ready for final review I think :)

messages/compiler/python/test_pygen.py
```diff
@@ -0,0 +1,181 @@
...
+
+def test_NewViewElement(codegen):
+    import example
+
+    msg = example.NewViewElement()
+    msg.replica_id = 5
+    msg.digest = b"somedigest"
+    assert_roundtrip(msg)
+
+    with pytest.raises(Exception):
```
robem, line 54: All tests are checking for any exception but we should check whether the proper exception is thrown

messages/compiler/python/pygen.py
```diff
@@ -0,0 +1,30 @@
...
+    Walk concord message format(CMF) AST and generate Python code.
+
+    Return Python code as a string.
+    """
+    with open(os.path.join(os.path.dirname(__file__), "serialize.py")) as f:
+        base_serializers = f.read() + '\n'
+    visitor = PyVisitor()
+    walker = Walker(ast, visitor)
+    walker.walk()
+    return base_serializers + visitor.output
```
robem, line 30: Somewhere, we should add the "AUTO GENERATED CODE" warning into the resulting `.py` file.

messages/compiler/cmfc.py
```diff
@@ -25,13 +25,19 @@ def translate(ast, language, namespace, output):
...
         from cpp import cppgen
-
         header, impl = cppgen.translate(ast, output + ".hpp", namespace)
         with open(output + ".hpp", "w") as f:
             f.write(header)
         with open(output + ".cpp", "w") as f:
             f.write(impl)
 
+    elif language == "python":
+        print("Generating Python source code")
```
robem, line 35: At some point we should replace all `print` statements with proper logging. No need to change now.

robem (approved):
The PR aged well and got even better over time :) I ran the tests locally according to the PR description. Thanks for the the C++ fixes and the test code in addition to the actual Python support.

Only minor comments below.

messages/compiler/python/serialize.py
```diff
@@ -0,0 +1,329 @@
...
+                s[1]) is str:
+            return self.msg(s[1])
+        elif is_primitive(s):
+            return getattr(self, s)()
+        else:
+            raise CmfDeserializeError(f'Invalid serializer: {s}')
+
+    def bool(self):
+        if self.pos + 1 > len(self.buf):
+            raise NoDataLeftError()
```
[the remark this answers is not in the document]
andrewjstone: Good point. Will leave this for a future cleanup.

messages/compiler/python/test_pygen.py
andrewjstone: Excellent point. I just found an exception I didn't expect :) Fixing...

messages/compiler/python/pygen.py
andrewjstone: Fixed.

PR #810, opened 2020-08-09
Synchronize msg processing thread with recovery

cloudnoize:
The main thread should handle recovery before msg processing thread starts accepting msgs.
When the main thread finishes handling recovery in the replica implementation, it notifies the msg processing thread to start accepting msgs.

### DIFF
--- a/bftengine/src/bftengine/IncomingMsgsStorage.hpp
+++ b/bftengine/src/bftengine/IncomingMsgsStorage.hpp
@@ -29,7 +29,6 @@ class IncomingMsgsStorage {
 
   virtual void start() = 0;
   virtual void stop() = 0;
-  virtual void notifyOnSynch() = 0;
 
   virtual bool isRunning() const = 0;
 
--- a/bftengine/src/bftengine/IncomingMsgsStorage.hpp
+++ b/bftengine/src/bftengine/IncomingMsgsStorage.hpp
@@ -29,6 +29,7 @@ class IncomingMsgsStorage {
 
   virtual void start() = 0;
   virtual void stop() = 0;
+  virtual void notifyOnSynch() = 0;
 
   virtual bool isRunning() const = 0;
 
--- a/bftengine/src/bftengine/IncomingMsgsStorageImp.cpp
+++ b/bftengine/src/bftengine/IncomingMsgsStorageImp.cpp
@@ -123,10 +123,21 @@ IncomingMsg IncomingMsgsStorageImp::popThreadLocal() {
   }
 }
 
+void IncomingMsgsStorageImp::notifyOnSynch() {
+  startAccept_ = true;
+  acceptCV_.notify_one();
+}
+
 void IncomingMsgsStorageImp::dispatchMessages(std::promise<void>& signalStarted) {
   signalStarted.set_value();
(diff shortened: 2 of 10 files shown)


cloudnoize:
> As far as I understand the idea is to process the recovered messages before processing the new ones.
> If so, can we just not call [msgsCommunicator_->startMsgsProcessing(config_.replicaId)](https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaBase.cpp#L53) in `ReplicaBase::start` but call it in `ReplicaImp::start`?
> Thus, the dispatcher thread will be launched only when Replica is ready to go.

@f-squirrel  That was the first imp choice but encountered issues like read-only replica needs the same mechanism and a dependency on the order of operations, which I currently don't remember the error I received.

f-squirrel:
@cloudnoize , I am not sure I understand the problem.
`msgsCommunicator_->startMsgsProcessing(config_.replicaId);` is the last operation in `ReplicaBase::start`.
All the "replicas" that inherit from the `ReplicaBase` do `msgsCommunicator_->startMsgsProcessing(config_.replicaId);` in their implementation of `start`(`ReplicaImp`, `ReplicaForStateTransfer`).
The `ReadOnlyReplica` does not need to do it because it calls `ReplicaForStateTransfer::start()`.
This way we do need to use the additional synchronization primitives.

cloudnoize:
@f-squirrel  think I recall. if you extract `msgsCommunicator_->startMsgsProcessing(config_.replicaId);` from base, you need to call it in `start` of each replica inheritance (imp,st,ro) but imp and ro ,need to call start of state-transfer, so where would you put the statement in state-transfer replica?

f-squirrel:
@cloudnoize , you're right, I got lost in the inheritance hierarchy...

As far as I see we actually instantiate only two types of replicas: [`ReplicaImp` and `ReadOnlyReplica`](https://github.com/vmware/concord-bft/blob/master/bftengine/include/bftengine/Replica.hpp#L79). The `ReplicaForStateTransfer` is never instantiated but only used as a base class.
Hence we may to call `startMsgsProcessing` only in the end of `ReplicaImp::start` and `ReadOnlyReplica::start`.
WDYT?

If there are no other implications, I suggest considering the simpler solution i.e. just moving the function call to the upper hierarchy.

salieri11:
Just a general question: doesn't the recovery process use the same message dispatching mechanism?

cloudnoize:
> Just a general question: doesn't the recovery process use the same message dispatching mechanism?

@salieri11 nope, it process committed msgs that it knows that their execution was aborted, on start. must be before any msg processed by the dispatching mechanism.

salieri11:
> > Just a general question: doesn't the recovery process use the same message dispatching mechanism?
> 
> @salieri11 nope, it process committed msgs that it knows that their execution was aborted, on start. must be before any msg processed by the dispatching mechanism.

So who provides the thread for the processing?

cloudnoize:
> > > Just a general question: doesn't the recovery process use the same message dispatching mechanism?
> > 
> > 
> > @salieri11 nope, it process committed msgs that it knows that their execution was aborted, on start. must be before any msg processed by the dispatching mechanism.
> 
> So who provides the thread for the processing?

main thread is responsible for the recovery.
it also calls replica base start, that starts the processing thread

PR #812, opened 2020-08-11
Implemented proper exit for skvbc_replica, necessary for Leak sanitizers

HristoStaykov:
to generate reports.

### DIFF
--- a/tests/simpleKVBC/TesterReplica/main.cpp
+++ b/tests/simpleKVBC/TesterReplica/main.cpp
@@ -29,6 +29,8 @@
 
 namespace concord::kvbc::test {
 
+std::atomic_bool timeToExit = false;
+
 void run_replica(int argc, char** argv) {
   const auto setup = TestSetup::ParseArgs(argc, argv);
   logging::initLogger(setup->getLogPropertiesFile());
@@ -54,17 +56,25 @@ void run_replica(int argc, char** argv) {
   // registration of metrics from the replica with the aggregator and don't
   // return empty metrics from the metrics server.
   setup->GetMetricsServer().Start();
-  while (replica->isRunning()) std::this_thread::sleep_for(std::chrono::seconds(1));
+  while (replica->isRunning()) {
+    if (timeToExit) {
+      setup->GetMetricsServer().Stop();
+      replica->stop();
+    } else {
+      std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+  }
 }
 }  // namespace concord::kvbc::test
 
 using namespace std;
 
-void signal_handler(int signal_num) {
+namespace {
+static void signal_handler(int signal_num) {
   LOG_INFO(GL, "Program received signal " << signal_num);
-  exit(0);
+  concord::kvbc::test::timeToExit = true;
 }
-
+}  // namespace
 int main(int argc, char** argv) {
   signal(SIGINT, signal_handler);
   signal(SIGTERM, signal_handler);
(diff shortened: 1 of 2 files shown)


tests/simpleKVBC/TesterReplica/main.cpp
```diff
@@ -54,17 +56,25 @@ void run_replica(int argc, char** argv) {
...
 }  // namespace concord::kvbc::test
 
 using namespace std;
 
-void signal_handler(int signal_num) {
+namespace {
+static void signal_handler(int signal_num) {
   LOG_INFO(GL, "Program received signal " << signal_num);
-  exit(0);
+  concord::kvbc::test::timeToExit = true;
```
dartdart26, line 75: Nice, that is clean solution!

Nit: I don't want to be a language lawyer here, especially when we log in signal handlers via global logger objects, but if the atomic variable  is not lock-free, that won't be safe as it implies a mutex that can be held by the same thread running the handler. That obviously matters mainly for portability.

Not sure if we should handle that as we have C-style signal handling all over the place. Might be a good idea to still check if the variable is lock-free or don't use `signal()` for signal-handling. We can use `sigaction` instead. Or better yet - boost.

Ref: https://en.cppreference.com/w/cpp/atomic/atomic_is_lock_free

util/src/MetricsServer.cpp
```diff
@@ -82,11 +82,27 @@ void Server::RecvLoop() {
     }
     running_lock_.unlock();
 
+    struct timeval timeout;
+    timeout.tv_sec = 5;
+    timeout.tv_usec = 0;
+
+    if (setsockopt(sock_, SOL_SOCKET, SO_RCVTIMEO, (char*)&timeout, sizeof(timeout)) < 0) {
+      LOG_ERROR(logger_, "Failed to set socket timeout!" << concordUtils::errnoString(errno));
```
dartdart26, line 90: Should we continue if we can't set this option? Maybe abort?

util/src/MetricsServer.cpp
```diff
@@ -82,11 +82,27 @@ void Server::RecvLoop() {
...
     if (len < 0) {
-      LOG_ERROR(logger_, "Failed to recv msg: " << concordUtils::errnoString(errno));
+      if (errno != EAGAIN) {
+        LOG_ERROR(logger_, "Failed to recv msg: " << concordUtils::errnoString(errno));
+      }
+
+      std::unique_lock<std::mutex> lock(running_lock_);
+      if (!running_) {
+        return;
+      }
```
dartdart26, line 104: Nit: Might be redundant as we check that in the beginning of the loop.

util/src/MetricsServer.cpp
```diff
@@ -82,11 +82,27 @@ void Server::RecvLoop() {
...
+    timeout.tv_usec = 0;
+
+    if (setsockopt(sock_, SOL_SOCKET, SO_RCVTIMEO, (char*)&timeout, sizeof(timeout)) < 0) {
+      LOG_ERROR(logger_, "Failed to set socket timeout!" << concordUtils::errnoString(errno));
+    }
+
     socklen_t addrlen = sizeof(cliaddr);
     len = recvfrom(sock_, buf_, MAX_MSG_SIZE, 0, (sockaddr*)&cliaddr, &addrlen);
 
     if (len < 0) {
```
dartdart26, line 96: Nit: I realise that is not part of this PR, but If `recvfrom()` fails, then we will busy loop. Maybe we can improve on that?

PR #814, opened 2020-08-11
[Pre-execution] Fix span context/cid order when creating PreProcessRequestMsg

teoparvanov:
Both fields are `std::string`, which is why the compiler did not detect the wrong parameter order.


andrewjstone:
This is exactly the reason for new type wrappers! Unfortunately, in C++ every time you create a type wrapper, you have to reimplement all operations :(

f-squirrel:
@andrewjstone , there is a PR fixing it:
https://github.com/vmware/concord-bft/pull/816

andrewjstone:
> @andrewjstone , there is a PR fixing it:
> #816

Saw that. Thank you :)

PR #809, opened 2020-08-07
Add support in HaveYouStopped RO command in command handler

yontyon:
Add Apollo support in Replica Specific Info. 
Specially, we add support in "Have you stopped" read-only command to support more reconfiguration tests.
Also, we have added a simple test in the reconfiguration tests to test the replica specific information feature in Apollo.

PR #815, opened 2020-08-11
Add support in specific replica info and special quorums

yontyon:
In this PR wee add support in replica specific information messages as well as in configurable quorums.
Please see #809 to understand the context


util/pyclient/replica_specific_info.py
```diff
@@ -0,0 +1,82 @@
...
+from collections import namedtuple
+import bft_msgs
+
+CommonReplyHeader = namedtuple('CommonReplyHeader', ['span_context_size', 'primary_id',
+                                                     'req_seq_num', 'length'])
+
+MatchedReplyKey = namedtuple('MatchedReplyKey', ['req_seq_num', 'header', 'data'])
+
+
+class MsgWithSpecificReplicaInfo:
```
andrewjstone, line 24: Please stick to the conventional name: `ReplicaSpecificInfo` rather than `SpecificReplicaInfo`

util/pyclient/test_msgs.py
```diff
@@ -14,6 +14,119 @@
 import struct
 
 import bft_msgs
+import replica_specific_info as rsi
+
+
+class TestSpecificReplyInfoManager(unittest.TestCase):
```
andrewjstone, line 20: Great tests!

The name can probably be simplified to `TestRepliesManager` though.

util/pyclient/test_client.py
```diff
@@ -181,6 +181,23 @@ async def _testPrimaryWrite(self):
            self.assertEqual(2, self.read_val(read))
            self.assertNotEqual(None, udp_client.primary)
 
+    async def _testMofNQuorum(self):
```
andrewjstone, line 184: Nice :)

util/pyclient/bft_client.py
```diff
@@ -173,58 +203,41 @@ async def send_to_primary(self, request):
...
-    async def recv(self, cancel_scope):
+    async def recv(self, required_quorum_size, cancel_scope):
         """
         Receive reply messages until a quorum is achieved or the enclosing
         cancel_scope times out.
         """
         while True:
             data, sender = await self.sock.recvfrom(self.config.max_msg_size)
-            header, reply = bft_msgs.unpack_reply(data)
+            sri_msg = rsi.MsgWithSpecificReplicaInfo(data, sender)
```
andrewjstone, line 225: Please stick to `rsi` over `sri`

util/pyclient/bft_client.py
```diff
@@ -173,58 +203,41 @@ async def send_to_primary(self, request):
...
+                    self.rsi_replies = self.replies_manager.get_rsi_replies(sri_msg.get_matched_reply_key())
+                    self.primary = self.replicas[header.primary_id]
+                    cancel_scope.cancel()
 
     def valid_reply(self, header):
-        """Return true if the sequence number is correct"""
         return self.req_seq_num.val() == header.req_seq_num
 
-    def has_quorum(self):
+    def get_rsi_replies(self):
```
andrewjstone, line 238: Why is this method needed? Is it only because you return just the reply and not the rsi_data when the send call returns?

I think a better solution in the long term would be to return the matching reply and rsi dict together, but that would probably require a change at all callers, so it makes sense not to do that now.

util/pyclient/bft_client.py
```diff
@@ -69,26 +95,26 @@ def __init__(self, config, replicas):
...
-        self.replies = dict()
         self.primary = None
         self.reply = None
         self.retries = 0
         self.msgs_sent = 0
-        self.reply_quorum = 2*config.f + config.c + 1
-        self.port = BASE_PORT + 2*self.client_id
+        self.port = BASE_PORT + 2 * self.client_id
         self.sock_bound = False
+        self.replies_manager = rsi.RepliesManager()
```
andrewjstone, line 104: There needs to either be a mechanism to clear replies, or a new `RepliesManager` needs to be used on each request.

util/pyclient/bft_client.py
```diff
@@ -173,58 +203,41 @@ async def send_to_primary(self, request):
...
-            header, reply = bft_msgs.unpack_reply(data)
+            sri_msg = rsi.MsgWithSpecificReplicaInfo(data, sender)
+            header, reply = sri_msg.get_common_reply()
             if self.valid_reply(header):
-                self.replies[sender] = (header, reply)
-            if self.has_quorum():
-                # This cancel will propagate upward and gracefully terminate all
-                # coroutines. self.reply will get set in self.has_quorum()
-                cancel_scope.cancel()
+                quorum_size = self.replies_manager.add_reply(sri_msg)
```
andrewjstone, line 228: I don't think this works exactly like it should. The problem is that the replies_manager has no notion of the specific MofN quorum. This allows replies to get added from replicas that are different from the ones required. My suggestion is to instead construct the replies_manager with the specific quorum per request so it can successfully add only replies from replicas in its required list.

util/pyclient/replica_specific_info.py
```diff
@@ -0,0 +1,82 @@
...
+
+# This code requires python 3.5 or later
+
+from collections import namedtuple
+import bft_msgs
+
+CommonReplyHeader = namedtuple('CommonReplyHeader', ['span_context_size', 'primary_id',
+                                                     'req_seq_num', 'length'])
+
+MatchedReplyKey = namedtuple('MatchedReplyKey', ['req_seq_num', 'header', 'data'])
```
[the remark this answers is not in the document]
yontyon: fixed

util/pyclient/replica_specific_info.py
```diff
@@ -0,0 +1,82 @@
...
+    def is_valid(self, req_seq_num):
+        return self.common_header.req_seq_num == req_seq_num
+
+
+class RepliesManager:
+    def __init__(self):
+        self.replies = dict()
+
+    def validate_reply(self, rsi_msg):
+        return True
```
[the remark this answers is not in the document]
yontyon: fixed

util/pyclient/replica_specific_info.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def add_reply(self, rsi_message):
```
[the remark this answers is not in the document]
yontyon: done

util/pyclient/replica_specific_info.py
```diff
@@ -0,0 +1,82 @@
...
+        key = rsi_message.get_matched_reply_key()
+        if key not in self.replies.keys():
+            self.replies[key] = dict()
+        self.replies[key][rsi_message.get_sender_id()] = rsi_message.get_rsi_data()
+        return len(self.replies[key])
+
+    def pop(self, matched_reply_key):
+        return self.replies.pop(matched_reply_key)
+
+    def length(self):
```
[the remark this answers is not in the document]
yontyon: done

util/pyclient/replica_specific_info.py
yontyon: done

util/pyclient/test_msgs.py
yontyon: done

util/pyclient/bft_client.py
yontyon: done

util/pyclient/bft_client.py
yontyon: You right. But I wouldn't want to change the whole interface now

util/pyclient/bft_client.py
yontyon: done

util/pyclient/bft_client.py
yontyon: I prefer to separate the specific replica info infra from the m out of n quorums. I've added this check in the bft client itself such that a reply is valid only if came from one of the destination replicas

PR #816, opened 2020-08-11
Opentracing: type for span context

f-squirrel:
Rationale:
Originally `std::string` was used to store the span context.
Since the string is a generic type, there were several issues when arbitrary strings were passed instead of actual context(span name or `cid`).
In order to solve the issue, a new type is introduced.

Changes:
Added class `SpanContext`

### DIFF
--- a/util/include/OpenTracing.hpp
+++ b/util/include/OpenTracing.hpp
@@ -21,7 +21,15 @@
 
 namespace concordUtils {
 
-using SpanContext = std::string;
+class SpanContext {
+ public:
+  SpanContext() = default;
+  explicit SpanContext(const std::string& data) : data_{data} {}
+  const std::string& data() const { return data_; }
+
+ private:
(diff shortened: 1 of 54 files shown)


util/include/OpenTracing.hpp
```diff
@@ -21,7 +21,15 @@
 
 namespace concordUtils {
 
-using SpanContext = std::string;
+class SpanContext {
+ public:
+  SpanContext() = default;
+  explicit SpanContext(const std::string& data) : data_{data} {}
+  std::string data() const { return data_; }
```
andrewjstone, line 28: Maybe return a reference here, rather than a copy?

f-squirrel: Good suggestion!
If a developer uses `auto s = ctx.data()` then `auto` won't be a reference type.
Thanks!

PR #823, opened 2020-08-13
Add reconfiguration tests

yontyon:
Basing on #815, we have added tests for the reconfiguration mechanism in concord-bft.
We have added two tests:
1. send a Wedge command and make sure that all replicas have been stopped - using the read-only and replica specific info framework
2. simulation of the semi-manual upgrade. Apollo (acting as a system operator) sends a Wedge command, then it stops all replicas and restarts them again to verify that they are able to process new write requests.


bftengine/src/bftengine/ControlStateManager.cpp
```diff
@@ -43,7 +43,11 @@ std::optional<int64_t> ControlStateManager::getCheckpointToStopAt() {
   std::istringstream inStream;
   inStream.str(scratchPage_);
   controlStateMessages::StopAtNextCheckpointMessage msg;
-  concord::serialize::Serializable::deserialize(inStream, msg);
+  try {
+    concord::serialize::Serializable::deserialize(inStream, msg);
+  } catch (std::exception& e) {
```
andrewjstone, line 48: We should definitely at least log log an error here. But it may make even more sense to abort the process or just let the exception be thrown and not caught. 

Is there any normal case reason that deserialization will fail? If there is it's a bug, as we need to know that in this case the storage is corrupt.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3269,6 +3269,8 @@ void ReplicaImp::addTimers() {
 
 void ReplicaImp::start() {
   ReplicaForStateTransfer::start();
+  if (controlStateManager_ && controlStateManager_->getCheckpointToStopAt().has_value())
```
andrewjstone, line 3272: I would add a comment here stating that this is only safe if the system doesn't crash and restart on its own, but is done purposefully during a manual upgrade.

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -0,0 +1,229 @@
...
+class SkvbcReconfigurationTest(unittest.TestCase):
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_wedge_command(self, bft_network):
+        """
+             Sends a wedge command and check that the system stops from processing new requests.
+             Note that in this test we assume no failures and synchronized network.
+             The test does the following:
+             1. A client sends a wedge command
+             2. The client verify that the system reached to a super stable checkpoint
```
andrewjstone, line 57: Grammar: "The client verifies that the system reached a super stable checkpoint."

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -0,0 +1,229 @@
...
+            "-p" if os.environ.get('BUILD_ROCKSDB_STORAGE', "").lower() in set(["true", "on"]) else "",
+            "-t", os.environ.get('STORAGE_TYPE')]
+
+
+class SkvbcReconfigurationTest(unittest.TestCase):
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_wedge_command(self, bft_network):
+        """
+             Sends a wedge command and check that the system stops from processing new requests.
```
andrewjstone, line 53: Grammar: "Sends a wedge command and checks that the system stops processing new requests."

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -0,0 +1,229 @@
...
+        # Here the system operator runs a manual upgrade
+
+        bft_network.start_all_replicas()
+
+        await skvbc.assert_kv_write_executed(key, val)
+        await skvbc.write_known_kv()
+
+    async def validate_stop_on_super_stable_checkpoint(self, bft_network, skvbc):
+        for replica_id in range(bft_network.config.n):
+            with trio.fail_after(seconds=90):
```
andrewjstone, line 207: Rather than giving each replica 90 seconds, I think you should give them all 90 seconds or maybe 120 total. You can do this by moving the `trio.fail_after` above the for loop.

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -0,0 +1,229 @@
...
+         """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        client = bft_network.random_client()
+
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=0)
+
+        await client.write(skvbc.write_req([], [], block_id=0, wedge_command=True))
+
+        for replica_id in range(bft_network.config.n):
```
andrewjstone, line 68: This loops seems to be used repeatedly in these tests. Please put it in it's own async function that is parameterized by the the replicas to check.

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -0,0 +1,229 @@
...
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_wedge_command_and_specific_replica_info(self, bft_network):
+        """
+             Sends a wedge command and check that the system stops from processing new requests.
+             Note that in this test we assume no failures and synchronized network.
+             The test does the following:
+             1. A client sends a wedge command
+             2. The client then sends a "Have you stopped" read only command such that each replica answers "I have stopped"
+             3. The client validate with the mterics that all replicas have stopped
```
andrewjstone, line 139: grammar: validate -> validates

bftengine/src/bftengine/ControlStateManager.cpp
yontyon: If the reserved page is empty, the serializer throws an error.
This may happen if we try to read the checkpoint before we set anything into there (or if we cleared the reserved page)

yontyon: I will add a log message though

yontyon: done

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: done

tests/apollo/test_skvbc_reconfiguration.py
yontyon: done

tests/apollo/test_skvbc_reconfiguration.py
yontyon: done

tests/apollo/test_skvbc_reconfiguration.py
yontyon: done

tests/apollo/test_skvbc_reconfiguration.py
yontyon: done

tests/apollo/test_skvbc_reconfiguration.py
yontyon: done

tests/apollo/CMakeLists.txt
```diff
@@ -71,8 +71,8 @@ foreach(STORAGE_TYPE ${STORAGE_TYPES})
           "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} python3 -m unittest test_skvbc_backup_restore 2>&1 > /dev/null"
           WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
-  add_test(NAME skvbc_control_commands_tests_${STORAGE_TYPE} COMMAND sh -c
```
teoparvanov, line 74: Can you delete the already obsolete `test_skvbc_control_commands` module?

util/pyclient/replica_specific_info.py
```diff
@@ -0,0 +1,85 @@
+# Concord
+#
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
```
teoparvanov, line 3: nit - 2019

PR #825, opened 2020-08-13
Add internal BFT client

cloudnoize:
In order to perform key exchange, the KeyManager of each replica needs to publish the key via the bft consensus.
It needs to initiate consensus from within the replica.
This PR introduces a general internal BFT client that knows to send a client request via the dispatcher queue. 

- In order to support the internal client, ClientManager data structures were extended. since each replica has one internal client. the size of the data structure is now: |non-internal-clients| + cluster-size.

- The Client manager identifies each client by its ID. The internal client IDs are assigned after the initialization process of the non-internal clients. The formula for an internal client ID is to add the replica ID to the highest ID of the non-internal client, plus one.

- After the execution of the internal client request, the replica won't send response to it.

- Unit tests of the Client manager were added to test the existing and added functionality.


bftengine/src/bftengine/KeyManager.h
```diff
@@ -19,8 +21,13 @@ class KeyManager {
   };
 
  private:
-  bftEngine::IBasicClient* cl_{nullptr};
   int repID_{};
+  uint32_t clusterSize_{};
+  std::set<NodeIdType> exchangedReplicas_;  // TODO to/from Storage
+  std::string generateCid();
+  // Raw pointer is ok, since this class does not manage this resource.
+  InternalBFTClient* client_{nullptr};
```
toly-kournik, line 29: shared_ptr?

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -37,17 +46,50 @@ ClientsManager::ClientsManager(ReplicaId myId, std::set<NodeIdType>& clientsSet,
...
-  if (ReplicaConfigSingleton::GetInstance().GetMaxReplyMessageSize() % sizeOfReservedPage != 0)
-    reservedPagesPerClient_++;
+  reservedPagesPerClient_ = reservedPagesPerClient(sizeOfReservedPage, maxReplysize_);
+
+  numOfClients_ = (uint16_t)clientsSet.size();
+
+  requiredNumberOfPages_ = (numOfClients_ * reservedPagesPerClient_);
+}
+
+uint32_t ClientsManager::reservedPagesPerClient(const uint32_t& sizeOfReservedPage, const uint32_t& maxReplysize) {
```
toly-kournik, line 60: const method

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -37,17 +46,50 @@ ClientsManager::ClientsManager(ReplicaId myId, std::set<NodeIdType>& clientsSet,
...
+    LOG_DEBUG(GL,
+              "Adding internal client, id [" << currClId << "] as index [" << currIdx << "] vactor size "
+                                             << indexToClientInfo_.size());
+  }
+}
+
+NodeIdType ClientsManager::getHighestIdOfNonInternalClient() { return highestIdOfNonInternalClient_; }
 
-  uint16_t numOfClients = (uint16_t)clientsSet.size();
+bool ClientsManager::isInternal(NodeIdType clientId) { return clientId > highestIdOfNonInternalClient_; }
```
toly-kournik, line 88: const method, can be defined inline

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -37,17 +46,50 @@ ClientsManager::ClientsManager(ReplicaId myId, std::set<NodeIdType>& clientsSet,
...
 
-  requiredNumberOfPages_ = (numOfClients * reservedPagesPerClient_);
+int ClientsManager::getIndexOfClient(const NodeIdType& id) {
```
toly-kournik, line 90: const method

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -92,6 +138,9 @@ void ClientsManager::loadInfoFromReservedPages() {
 
 ReqId ClientsManager::seqNumberOfLastReplyToClient(NodeIdType clientId) {
   uint16_t idx = clientIdToIndex_.at(clientId);
+  LOG_DEBUG(GL,
+            "seqNumberOfLastReplyToClient id [" << clientId << "] as index [" << idx << "] vec size "
```
toly-kournik, line 142: No need to put a function name in the log message - it's added automatically

bftengine/tests/clientsManager/ClientsManager_test.cpp
```diff
@@ -0,0 +1,66 @@
+// Concord
```
andrewjstone, line 1: Thanks for the tests :)

bftengine/src/bftengine/InternalBFTClient.h
```diff
@@ -0,0 +1,22 @@
...
+class InternalBFTClient {
+ public:
+  InternalBFTClient(const int& id, const NodeIdType& nonInternalNum, MsgsCommunicator* msgComm);
+  NodeIdType getClientId();
+  void sendRquest(uint8_t flags, uint32_t requestLength, const char* request, const std::string& cid);
+
+ private:
+  uint32_t repID_{};
+  NodeIdType startIdForInternalClient_{};
+  MsgsCommunicator* msgComm_{nullptr};
```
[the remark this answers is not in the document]
cloudnoize: It's not the owner of the resource. 
For users of resources, I think it's valid, safe and reduces overhead to use raw pointers. 
what do you think?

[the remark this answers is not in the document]
andrewjstone: I think it would be better to use a shared_ptr, as it will guarantee proper destruction. In theory the raw pointer may not be valid because it doesn't know the lifetime of the original pointer. If you could guarantee that the pointer would live at least as long as the owner it would be safe. But in that case you should just us a reference, since the pointer is already created.

The common case for using a pointer when not owning is when just passing to a local function in the call chain that doesn't store the pointer.

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -34,41 +39,42 @@ std::string KeyManager::KeyExchangeMsg::toString() const {
...
+std::string KeyManager::generateCid() {
+  std::string cid{"KEY-EXCHANGE-"};
+  auto now = getMonotonicTime().time_since_epoch();
+  auto now_ms = std::chrono::duration_cast<std::chrono::milliseconds>(now);
+  auto sn = now_ms.count();
+  cid += std::to_string(repID_) + "-" + std::to_string(sn);
+  return cid;
+}
 
 std::string KeyManager::onKeyExchange(const KeyExchangeMsg& kemsg) {
```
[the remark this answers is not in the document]
cloudnoize: I don't know. BFT does not allow empty replies, so I guess dummy "ok".

bftengine/src/bftengine/KeyManager.h
cloudnoize: same as above, if it's not an owner I don't see a reason for ownership semantic

andrewjstone: The whole point of using shared_ptrs is to safely manage lifetimes when multiple users reference the pointer. If only the owner/creator had it, then there would be no need for shared pointers to exist. In that case you'd just always use a unique_ptr and reference.

andrewjstone: There's actually likely a good argument that could be made that we don't need shared pointers here or in most our uses because we can manually track the lifetime and ownership easy enough. My guess is the reason it was specifically used in this case was because this object in particular is used across multiple threads.

cloudnoize: Shared pointer is for multiple users in different threads, where no specific owner is defined.
here the `InternalBFTClient` is tied to the lifetime of the replicaImp. there is no scenario where the KeyManager will own the resource exclusively. 
I think once a resource is being owned and managed, there is no need to use shared_ptr

bftengine/src/bftengine/ClientsManager.cpp
cloudnoize: it's  static

bftengine/src/bftengine/KeyManager.h
andrewjstone: In that case, this is fine. Thanks for the explanation!

PR #827, opened 2020-08-15
BftClient: Init sequence_number to 0 in config

andrewjstone:
A default allows wrapper code to check if it was set, and initialize if
not.

Also, this was the only member without a default, so it makes sense to
set one.


andrewjstone:
Merging as this is trivial, and not in use here anyway.

PR #828, opened 2020-08-16
Add summary metric for measuring the commit size in merkle tree

yontyon:
Adding a summary metric to measure the size of the commit size in Merkle tree

### DIFF
--- a/kvbc/include/db_adapter_interface.h
+++ b/kvbc/include/db_adapter_interface.h
@@ -7,6 +7,7 @@
 #include "block_digest.h"
 #include "kv_types.hpp"
 #include <utility>
+#include "Metrics.hpp"
 
 namespace concord::storage {
 class IDBClient;
@@ -74,6 +75,7 @@ class IDbAdapter {
   // TODO [TK] not sure it's needed for long term
   virtual std::shared_ptr<storage::IDBClient> getDb() const = 0;
 
+  virtual void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) = 0;
   virtual ~IDbAdapter() = default;
 };
 
--- a/kvbc/src/merkle_tree_db_adapter.cpp
+++ b/kvbc/src/merkle_tree_db_adapter.cpp
@@ -51,7 +51,9 @@ using namespace detail;
 constexpr auto MAX_BLOCK_ID = std::numeric_limits<BlockId>::max();
 
 // Converts the updates as returned by the merkle tree to key/value pairs suitable for the DB.
-SetOfKeyValuePairs batchToDbUpdates(const sparse_merkle::UpdateBatch &batch, BlockId blockId) {
+SetOfKeyValuePairs batchToDbUpdates(const sparse_merkle::UpdateBatch &batch,
+                                    BlockId blockId,
+                                    int64_t &update_raw_size) {
   SetOfKeyValuePairs updates;
   const Sliver emptySliver;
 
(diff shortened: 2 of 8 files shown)


kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -354,7 +367,10 @@ BatchedInternalNode DBAdapter::Reader::get_internal(const InternalNodeKey &key)
 
 BlockId DBAdapter::addBlock(const SetOfKeyValuePairs &updates, const OrderedKeysSet &deletes) {
   const auto blockId = getLastReachableBlockId() + 1;
-  const auto status = db_->multiPut(lastReachableBlockDbUpdates(updates, deletes, blockId));
+  int64_t data_raw_size = 0;
+  auto data = lastReachableBlockDbUpdates(updates, deletes, blockId, data_raw_size);
+  writeCommitSizeSummary_.Get().Observe(data_raw_size);
```
yuliasherman, line 372: Why do we need to calculate data_raw_size in many different places instead of doing it here from the resulting lastReachableBlockDbUpdates SetOfKeyValuePairs?

dartdart26: I agree with @yuliasherman 's comment - we can just use the `dbUpdates` return variable in `lastReachableBlockDbUpdates()`. It will also cover the case of replica receiving a block over State Transfer.

kvbc/include/db_adapter_interface.h
```diff
@@ -74,6 +75,7 @@ class IDbAdapter {
   // TODO [TK] not sure it's needed for long term
   virtual std::shared_ptr<storage::IDBClient> getDb() const = 0;
 
+  virtual void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) = 0;
```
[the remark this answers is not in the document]
dartdart26: I am thinking if we actually need this to be part of the base class. We can start by only adding it in the merkle DBAdapter and add to the base class in the future, if needed.

It might be done by passing the agregator to the `v2MerkleTree::RocksDBStorageFactory`'s constructor. The factory will then call a non-virtual `setAggregator()` method on the instances it produces.

Just something to consider to avoid further inheritance and adding methods to a base class.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -749,10 +749,10 @@ TEST_P(db_adapter_custom_blockchain, delete_last_reachable_block) {
   for (auto i = 1; i <= numBlocks - 1; ++i) {
     ASSERT_EQ(adapter.addBlock(getOffsetUpdates(i, numBlocks)), i);
   }
-
+  int64_t updates_raw_data = 0;
```
dartdart26, line 752: Maybe we can call this differently, i.e. `blockUpdatesByteSize` or similar. Additionally, I think it might be an unsigned type.

kvbc/src/merkle_tree_db_adapter.cpp
yontyon: I would prefer to avoid rotating over the updates again. This may harm performance, isn't it?

yuliasherman: Don't think so. It's a quick cpu operation.

yontyon: And what if the size of the update map is very big?
Isn't it better to make only a single loop over it?

yuliasherman: You can measure how long it takes for 1000 keys :-)

kvbc/include/db_adapter_interface.h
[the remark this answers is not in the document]
yontyon: I think in the future we might want to add metrics also in the direct key value store (for example, to compare performance). So I think that would be good thing to have it with a default imp in the interface, what do you think?

yontyon:
Implementation moved to #839

PR #829, opened 2020-08-17
Diagnostic tool - print answer to file

Liorgal28:
I've added a print to file to the concord-ctl since it requested by the deployment team.
They need it in order to parse the answer from the file and not from the console.
currently, they want to use the diagnostic tool to know the replica health status by sending:
`./concord-ctl status get replica`

After my change, it will print to the console and also always save the last answer to a file.

### DIFF
--- a/diagnostics/concord-ctl
+++ b/diagnostics/concord-ctl
@@ -10,3 +10,6 @@ cmd = b' '.join(sys.argv[1:]) + b'\n'
 s.send(cmd)
 data = s.recv(64*1024)
 print(data.decode())
+f = open("status.txt", "w")
+f.write(data.decode())
+f.close()


diagnostics/concord-ctl
```diff
@@ -10,3 +10,6 @@ cmd = b' '.join(sys.argv[1:]) + b'\n'
 s.send(cmd)
 data = s.recv(64*1024)
 print(data.decode())
+f = open("status.txt", "w")
```
f-squirrel, line 13: I'd suggest 2 things:
1. use `with`:
```
with open(...) as f:
    f.write(...)
```
2. It may be useful to add a parameter `-o` to set the path

andrewjstone: > LGTM!
> But I have a general question: why can't they do just:
> `./concord-ctl status get replica > whatever_file_they_want.txt`?
> 
> P.S.
> I've left a few minor comments.

I agree. I don't think we should add this at all. I think it is completely unnecessary. Shell redirect should be used. 

Let's try not to add options for things that are easy to do in the shell. Options should only be added for different types of diagnostics.

Liorgal28: 1) I'll change thanks.
2) I'm not sure if we need to set a path - don't want to add more parameters to the call - I think that the root is ok.

About your suggestion I thought about it also, the only thing I can thought on is that they want it to happen every time and automatically. but I'm waiting for them to connect today to ask actually.

PR #834, opened 2020-08-18
Architecture decision record for CMF

robem:
Using our own message format will always be a source of discussion. This change
adds an architecture decision record (ADR) which contains thoughts and
arguments that went into the decision making process.
This document was created in collaboration with @andrewstone and Ittai Abraham.

### DIFF
--- a/docs/adr-cmf.md
+++ b/docs/adr-cmf.md
@@ -1,0 +1,197 @@
+# Concord Message Format (CMF)
+
+## Context and Problem Statement
+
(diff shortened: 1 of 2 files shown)


andrewjstone:
Hmm. I wonder if instead of going in `/docs`, this should go in the messages folder and a link inserted into the messages README.

robem:
> Hmm. I wonder if instead of going in `/docs`, this should go in the messages folder and a link inserted into the messages README.

I assume and hope that this won't be the last ADR and hence `docs` might be a good fit in order to have all "big decisions" in one place. I will add a link from the CMF README to this ADR.

PR #835, opened 2020-08-18
Remove metadata with sparse_merkle_db_inspector

f-squirrel:
Rationale:
A database created by one replica can be used by another one only if the database does not
contain any data specific to the original replica(metadata or state transfer data).

Changes:
* Added "removeMetadata" command line to sparse_merkle_db_inspector.
* Since the inspector is not read-only anymore it is renamed to sparse_merkle_db_editor.


kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
```diff
@@ -185,14 +185,33 @@ struct GetValue {
...
 
+struct RemoveMetadata {
+  std::string description() const {
+    return "removeMetadata\n"
+           "Removes metadata and state transfer data from RocksDB";
+  }
+  std::string execute(const v2MerkleTree::DBAdapter &db_adapter, const CommandArguments &) const {
+    const concordUtils::Sliver b{std::string{static_cast<char>(storage::v2MerkleTree::detail::EDBKeyType::BFT),
+                                             static_cast<char>(storage::v2MerkleTree::detail::EBFTSubtype::Metadata)}};
+    const concordUtils::Sliver e{std::string{static_cast<char>(storage::v2MerkleTree::detail::EDBKeyType::Key)}};
```
dartdart26, line 196: Nit: Maybe we can call them `begin` and `end` to make it clearer.

dartdart26, line 196: I think we should add some checks to make sure we are safe in case the enum is changed. For example:

```
// Pseudocode

// Ensure Key is the next key type.
static_assert(EDBKeyType::BFT == EDBKeyType::Key + 1);

// Ensure Metadata is the first enum value.
static_assert(std::is_unsigned_v<EBFTSubtype>);
static_assert(EBFTSubtype::Metadata == 0);
```

Above won't solve the case of adding types to the `EBFTSubtype` enum after `Metadata` that we don't want to delete...

Another option is to not use `rangeDel()` and loop and only delete the types we want.

WDYT?

kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
```diff
@@ -185,14 +185,33 @@ struct GetValue {
...
+
+    const auto status = db_adapter.getDb()->rangeDel(b, e);
+    if (!status.isOK()) {
+      throw std::runtime_error{"Failed to delete metadata and state transfer data: " + status.toString()};
+    }
+    return "Sucessfully deleted metadata and state transfer data";
```
dartdart26, line 202: Nit: Since all other commands return JSON outputs, we might do the same here. Something similar to:

```
{ "result": "true" }
```

dartdart26:
Could you please also add a test for this command in the unit test?

kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
f-squirrel: Done

kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
f-squirrel: Done

kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
f-squirrel: I reviewed the solution and understood that we want to delete all the `BFT` keys.
Please take a look at the updated file.

dartdart26: Ohk,, cool.

f-squirrel:
> Could you please also add a test for this command in the unit test as it is a potentially dangerous one?

Done

kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
```diff
@@ -185,14 +185,37 @@ struct GetValue {
   }
 };
 
+struct RemoveMetadata {
+  std::string description() const {
+    return "removeMetadata\n"
+           "Removes metadata and state transfer data from RocksDB";
+  }
+  std::string execute(const v2MerkleTree::DBAdapter &db_adapter, const CommandArguments &) const {
```
yuliasherman, line 193: Nit: be consistent in the use of function parameter names in the declaration.

yuliasherman (approved):
LGTM

kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
f-squirrel: Done.

kvbc/test/sparse_merkle_storage/sparse_merkle_db_editor_test.cpp
```diff
@@ -333,46 +350,56 @@ TEST_F(SparseMerkleDbInspectorTests, get_value_with_version_0) {
...
+  ASSERT_EQ(EXIT_SUCCESS,
+            run(CommandLineArguments{{"sparse_merkle_db_editor_test", rocksDbPath(), "removeMetadata"}}, out_, err_));
+  ASSERT_THAT(out_.str(), StartsWith(toJson(std::string{"result"}, std::string{"true"})));
+  ASSERT_TRUE(err_.str().empty());
+
+  const auto db = TestRocksDb::create();
+  const auto& kvp = generateMetadataAndStateTransfer();
+  for (const auto& kv : kvp) {
+    ASSERT_TRUE(db->has(kv.first).isNotFound());
+  }
```
dartdart26, line 398: When we call `TestRocksDb::create()`, we will essentially create a new DB handle **and** call `cleanup()`, leading to always having an empty DB.

I think what we can do instead is to make the DB object a member of the fixture(Test) and just use it to verify keys are deleted after the `removeMedatada` call. Or any other method of making sure we use the same DB on disk.

Another thing I think might make sense is to verify that other key types are present after calling `removeMetadata`, i.e. `Block`, `Key` and so on.

kvbc/test/sparse_merkle_storage/sparse_merkle_db_editor_test.cpp
```diff
@@ -48,6 +51,23 @@ using namespace std::string_literals;
 
 Sliver getSliver(unsigned value) { return toBigEndianStringBuffer(value); }
 
+SetOfKeyValuePairs generateMetadataAndStateTransfer() {
+  constexpr auto num_metadata_object_ids = MAX_METADATA_PARAMS_NUM - INITIALIZED_FLAG;
```
dartdart26, line 55: Nit: Maybe we can have a note on why this particular number is used. I am not sure I understand it.

dartdart26 (changes requested):
It looks great, thanks for expanding the tool @f-squirrel !

I think we should fix one minor thing, though. I've left a comment about the `TestRocksDb::create()` call in the unit test.

toly-kournik:
A general question: don't we want to support v1 kvbc data store type?

f-squirrel:
> A general question: don't we want to support v1 kvbc data store type?

Meanwhile, `v2` is in use.
Once we need to support both, we may add it.

kvbc/test/sparse_merkle_storage/sparse_merkle_db_editor_test.cpp
f-squirrel: done

kvbc/test/sparse_merkle_storage/sparse_merkle_db_editor_test.cpp
f-squirrel: I did it first but then I had 2 instances of the DB object(fixture and the code under test).
As a result, the code under test could not open the DB(the `LOCK` file issue).
As far as I know, `SetUp` and `TearDown` are called for each test, aren't they?

I've removed `cleanup()` from `create` and call explicitly in `SetUp`.
Thus, both the editor and the test use the same DB.
Added checks to make sure that blocks still exist.

dartdart26: I think you are right, calling `cleanup()` in `create()` might not be the best idea.

Thanks!

dartdart26 (approved):
Thanks for addressing my comments!

PR #836, opened 2020-08-19
Multisignatures

toly-kournik:
- Switch to multisignatures in all paths
- Dynamic public key calculation during Multisig verification;
- ViewChangeMsg length uint16_t => size_t, as multisig certs are longer (include bitvector of signers);
- Threshsigh/bls logging.

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -109,7 +109,7 @@ struct ReplicaConfig {
   bool debugPersistentStorageEnabled = false;
 
   // Messages
-  uint32_t maxExternalMessageSize = 65536;
+  uint32_t maxExternalMessageSize = 131072;
   uint32_t maxReplyMessageSize = 8192;
 
   // StateTransfer
--- a/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
+++ b/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
@@ -14,6 +14,7 @@
 #include <type_traits>
 #include <unordered_map>
 #include <set>
+#include <iterator>
 
 #include "OpenTracing.hpp"
 #include "PrimitiveTypes.hpp"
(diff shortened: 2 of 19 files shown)


f-squirrel:
ViewChangeMsg length used to be fixed type(`uint16_t`).
Is it not better to use `uint64_t` rather than platform-dependent `size_t`?

toly-kournik:
> ViewChangeMsg length used to be fixed type(`uint16_t`).
> Is it not better to use `uint64_t` or another fixed type rather than platform-dependent `size_t`?

Yes, it's in my list, I was planning to do it in a subsequent PR

f-squirrel:
> > ViewChangeMsg length used to be fixed type(`uint16_t`).
> > Is it not better to use `uint64_t` or another fixed type rather than platform-dependent `size_t`?
> 
> Yes, it's in my list, I was planning to do it in a subsequent PR

I see, btw it might be good to have a `typedef` for it.
It will be easier to change the type in the future.

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -109,7 +109,7 @@ struct ReplicaConfig {
   bool debugPersistentStorageEnabled = false;
 
   // Messages
-  uint32_t maxExternalMessageSize = 65536;
+  uint32_t maxExternalMessageSize = 131072;
```
cloudnoize, line 112: This param is also defined in the configuration file, may override this value I think.

cloudnoize: which use case requires the increase ? just for general knowledge

cloudnoize (commented):
LGTM

cloudnoize (approved):

bftengine/include/bftengine/ReplicaConfig.hpp
toly-kournik: External messages are basically all the BFT messages sent between replicas + Client[Request/Reply]Msg.
In this case the ViewChangeMsg's size has changed, but I guess it could be the max size for ClientRequestMsg...

PR #838, opened 2020-08-19
Multi-signature key store

cloudnoize:
Each replica has a key store which is a queue, limited in its size (currently 2).
A cluster key store is a vector of replica key stores with size equals to the cluster dimensions.

Operations against the cluster key store are triggered via interaction with the key manager.

- On key exchange - pushes new key to the corresponding replica queue and notifies registered components.

- On check point - checks whether to rotate a key by iterating over all replicas and checking whether the queue has two keys, and the corresponding check point is the second checkpoint since the key has arrived. If a key was rotated, notification is sent to registered components.

Replica key store is a serializable object that will be mapped to the replica's key-reserve page, in order to do so, support for deque has been added to the serializable api.

Most of the code is covered by UT.


bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -64,18 +30,40 @@ std::string KeyManager::generateCid() {
   return cid;
 }
 
-std::string KeyManager::onKeyExchange(const KeyExchangeMsg& kemsg) {
-  LOG_DEBUG(GL, "KEY EXCHANGE MANAGER  msg " << kemsg.toString());
+std::string KeyManager::onKeyExchange(KeyExchangeMsg& kemsg, const uint64_t& sn) {
```
f-squirrel, line 33: Reference is redundant for primitive types.
Please remove all over.

bftengine/src/bftengine/KeyManager.h
```diff
@@ -1,47 +1,50 @@
...
-    KeyExchangeMsg(std::string key, std::string signature, int repID);
-    std::string toString() const;
-    static KeyExchangeMsg deserializeMsg(const char* serStr, const int& size);
-
-   protected:
-    const std::string getVersion() const;
-    void serializeDataMembers(std::ostream& outStream) const;
-    void deserializeDataMembers(std::istream& inStream);
-  };
+  static KeyManager& get(InternalBFTClient* cl = nullptr, const int id = 0, const uint32_t clusterSize = 0) {
```
f-squirrel, line 19: The attribute `const` is redundant for the arguments passed by value.

bftengine/src/bftengine/KeyStore.h
```diff
@@ -0,0 +1,92 @@
...
+  std::deque<ReplicaKey> keys_;
+  uint16_t numOfKeysLimit_{2};
+  uint16_t seqNumsPerChkPoint_{150};  // TODO init from config
+  uint16_t checkPointsForRotation_{2};
+};
+
+// Holds all replicas key store.
+// Perform operations like rotation and push.
+// will be responsible on reserved pages operations.
+class ClsuterKeyStore {
```
f-squirrel, line 82: `ClusterKeyStore`?

bftengine/src/bftengine/KeyStore.cpp
```diff
@@ -0,0 +1,169 @@
...
+  for (auto ike : registryToExchange) {
+    ike->onNewKey(clusterKeys_[kem.repID].current().msg);
+  }
+  return true;
+}
+
+bool ClsuterKeyStore::rotate(const uint64_t& chknum, const std::vector<IKeyExchanger*>& registryToExchange) {
+  bool ret{};
+  auto idx = 0;
+  for (auto& rks : clusterKeys_) {
```
f-squirrel, line 149: Please give more meaningful name to `rks`.

bftengine/src/bftengine/KeyManager.h
```diff
@@ -1,47 +1,50 @@
...
+
+  uint16_t repID_{};
   uint32_t clusterSize_{};
   std::set<NodeIdType> exchangedReplicas_;  // TODO to/from Storage
   std::string generateCid();
   // Raw pointer is ok, since this class does not manage this resource.
   InternalBFTClient* client_{nullptr};
 
-  KeyManager(InternalBFTClient* cl, const int& id, const uint32_t& clusterSize);
+  std::vector<IKeyExchanger*> registryToExchange_;
```
f-squirrel, line 42: `KeyManager` does not own `IKeyExchanger*`, is it right?

f-squirrel (commented):
LGTM!
P.S. A few minor comments;)

bftengine/src/bftengine/KeyManager.h
cloudnoize: Disagree, const is meant for the reader as for the compiler. it doesn't matter if it has a side effect. it tells the reader that this variable shouldn't change.
Furthermore, inside the function one can change the value by mistake., and initialize a member with it for example.

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: const ref is c++ idiom, of course, that for primitives it saves nothing but idioms are important.
and if I relate to the comment below, if I remove the ref I should remove the const as well, which IMO is incorrect function signature for the purpose.

bftengine/src/bftengine/KeyStore.h
cloudnoize: yep 👍

bftengine/src/bftengine/KeyStore.cpp
cloudnoize: yep, you right

bftengine/src/bftengine/KeyManager.h
cloudnoize: right

bftengine/src/bftengine/KeyManager.cpp
f-squirrel: I do not see why the "idiom" is important and how it helps here.
Changes made to a variable passed by value do not affect the original value.
P.S. I meant to remove both `const` and `&`.

bftengine/src/bftengine/KeyManager.h
f-squirrel: The reader cannot change the passed variable anyways, since it is passed by value.
The fact that the variable is passed by value makes it clear to reader that it cannot be changed(since it is copied).
The copied variable can be changed by mistake:
```
auto id1 = id;
++id;
```
If we take a look at various functions and classes of the standard library, primitive types are always passed w/o const and references:
https://en.cppreference.com/w/cpp/container/vector/resize

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: I understand what you meant, I think that removing const is wrong.
See my comment below.

cloudnoize: and idioms are very important, they help the code to be more concise, readable, and clear to any c++ programmer.
Any c++ programmer knows the meaning of const ref signature

bftengine/src/bftengine/KeyManager.h
cloudnoize: the reader doesn't know that the function can't change the value inside. which he shouldn't.
const tells him that. as I said const is for the reader to understand the code by looking at the signature.
without the const I lose information, why does it bother you?
and the standard is not a good example, it's not an open source product.

f-squirrel: Ok man, if C++ standard library is not a good example of C++ code and its code is not an open-source project I am afraid this conversation is meaningless.
P.S. [GCC](https://github.com/gcc-mirror/gcc/tree/a1c9c9ff06ab15e697d5bac6ea6e5da2df840cf5/libstdc%2B%2B-v3), [clang](https://github.com/llvm-mirror/libcxx/tree/78d6a7767ed57b50122a161b91f59f19c9bd0d19).

bftengine/src/bftengine/KeyStore.h
```diff
@@ -0,0 +1,92 @@
...
+
+   protected:
+    const std::string getVersion() const;
+    void serializeDataMembers(std::ostream& outStream) const;
+    void deserializeDataMembers(std::istream& inStream);
+  };
+
+  bool push(const KeyExchangeMsg& kem, const uint64_t& sn);
+  void pop();
+  inline void setKeysLimit(const uint16_t& l) { numOfKeysLimit_ = l; };
```
toly-kournik, line 57: minor: methods defined inline are inline by default :-)

toly-kournik (approved):
LGTM!

bftengine/src/bftengine/KeyManager.h
cloudnoize: let's open a file in the standard library and see home much code is understandable.
I happened to look at it quite a lot

PR #839, opened 2020-08-19
Revert prometheus dep

yontyon:
After our offline discussion, we decided to keep concord-bft with less dependency as possible.
Thus, we revert our former PR and added an interface for injecting summaries from a higher level into concord-bft.
 
+Adding a summary to merkle tree


f-squirrel:
Please delete the new version of the docker image from Docker Hub after you merge this.

util/include/Metrics.hpp
```diff
@@ -107,28 +107,13 @@ class Counter {
...
-  };
-
-  using InitQuantiles = std::vector<std::tuple<double, double>>;
-
-  Summary(const InitQuantiles& quantiles);
-  void Observe(double value);
-  SummaryDescription Collect();
-
-  std::string ToJson();
+  Summary(ISummary* summary) : summaryImp_(summary) {}
```
andrewjstone, line 113: Please use `unique_ptr<ISummary>` and get rid of the destructor.

util/include/Statistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+struct SummaryDescription {
+  Quantiles quantiles_;
+  int64_t samples_sum_;
+  int64_t samples_count_;
+};
+class ISummary {
+ public:
+  virtual void Observe(double value) = 0;
+  virtual SummaryDescription Collect() = 0;
+  virtual ~ISummary() {}
```
andrewjstone, line 34: I don't believe we need a destructor here if we just use a unique_ptr.

util/include/Statistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+class EmptySummary : public ISummary {
+ public:
+  void Observe(double value) override {}
+  SummaryDescription Collect() override { return {}; }
+};
+
+class DefaultStatisticFactory : public IStatisticsFactory {
+ public:
+  virtual ISummary* createSummary(const Quantiles& quantiles) { return new EmptySummary(); }
```
andrewjstone, line 51: unique_ptr

util/include/Statistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+class ISummary {
+ public:
+  virtual void Observe(double value) = 0;
+  virtual SummaryDescription Collect() = 0;
+  virtual ~ISummary() {}
+};
+
+class IStatisticsFactory {
+ public:
+  virtual ISummary* createSummary(const Quantiles& quantiles) = 0;
```
andrewjstone, line 39: unique_ptr

util/include/Statistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+  virtual ISummary* createSummary(const Quantiles& quantiles) { return new EmptySummary(); }
+};
+
+class StatisticsFactory {
+  IStatisticsFactory* pImp = nullptr;
+
+ public:
+  StatisticsFactory() : pImp(new DefaultStatisticFactory()) {}
+
+  ISummary* createSummary(const Quantiles& quantiles) { return pImp->createSummary(quantiles); }
```
andrewjstone, line 60: unique_ptr

util/include/Statistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+  SummaryDescription Collect() override { return {}; }
+};
+
+class DefaultStatisticFactory : public IStatisticsFactory {
+ public:
+  virtual ISummary* createSummary(const Quantiles& quantiles) { return new EmptySummary(); }
+};
+
+class StatisticsFactory {
+  IStatisticsFactory* pImp = nullptr;
```
toly-kournik, line 55: unique_ptr, then you don't need a destructor

util/include/Statistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+ public:
+  StatisticsFactory() : pImp(new DefaultStatisticFactory()) {}
+
+  ISummary* createSummary(const Quantiles& quantiles) { return pImp->createSummary(quantiles); }
+
+  ~StatisticsFactory() {
```
toly-kournik, line 62: No need in destructor if is held as a unique_ptr

util/include/Statistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+    if (pImp) delete pImp;
+  }
+
+  static StatisticsFactory& get() {
+    static StatisticsFactory sf;
+    return sf;
+  }
+
+  static void setImp(StatisticsFactory& sf, IStatisticsFactory* pImp) { sf.pImp = pImp; }
```
toly-kournik, line 71: Why not just 
`
static void setImp(IStatisticsFactory* pImp) {
  StatisticsFactory::get().pImp = pImp
}
?
`

toly-kournik (changes requested):
> Looks good @yontyon. However, please don't use raw pointers in new code. Just replace everything with unique_ptr and remove the corresponding destructors.

I think the factory methods in general should return raw pointers, and the caller should decide how to hold it.
In this case I agree it's a unique_ptr.

util/include/Metrics.hpp
yontyon: Concord bft metrics framework makes some copy by value, return by value, and reference operations.
In this case, using uniqe_ptr may cause problems (for example, how to define a summary handler, as a reference to unique_ptr?).
This why I prefer to keep using raw pointers in this case...
What do you think?

util/include/Statistics.hpp
yontyon: done

util/include/Statistics.hpp
yontyon: done

andrewjstone:
> > Looks good @yontyon. However, please don't use raw pointers in new code. Just replace everything with unique_ptr and remove the corresponding destructors.
> 
> I think the factory methods in general should return raw pointers, and the caller should decide how to hold it.
> In this case I agree it's a unique_ptr.

If you return a `unique_ptr` it documents the exact ownership semantics. There is no work required to use it directly, and you don't have to worry about if there's a destructor or not preventing you from taking ownership, as in the case of many of our older interfaces. The only thing you would possibly want to do with it besides use it as a unique_ptr is to upgrade it to a shared_ptr which you can do in a safe manner atomically.

util/include/Metrics.hpp
andrewjstone: How does a raw pointer help you with this problem vs a unique_ptr? Maybe I'm not following the problem, but if you want a reference or value to be returned from any pointer, including smart pointers, can't you just dereference it with `*`?

PR #842, opened 2020-08-20
Introduce preExecMaxResultSize configuration parameter

yuliasherman:
The pre-execution result buffer size should be configurable independently from maxReplyMessageSize. This way we don't increase a number of reserved pages required to store client reply messages, as pre-execution result does not return to the clients.


toly-kournik (approved):
lgtm

PR #854, opened 2020-08-26
Support DB comparison in sparse_merkle_db_editor

dartdart26:
Add support for the `compareTo` command in sparse_merkle_db_editor that
compares two sparse merkle databases. Useful for debugging
non-determinism in a single or multiple replicas by comparing to a
deterministic one.

Outputs if the common block range is equivalent or there is a mismatch.
In case of a mismatch, the block ID of the first mismatch is returned.

Implement by exploiting the blockchain property and do a logaritmhic
search instead of a linear one.

Add an on/off flag for linking the State Transfer temporary chain to the
blockchain in v2MerkleTree::DBAdapter. Rationale is that we don't want
the editor to change the DB state.

Add tests for the `compareTo` command.

Add support for multiple storage test DBs by introducing a DB
identifier.

Remove the call to cleanup() when creating test databases.

### DIFF
--- a/kvbc/include/merkle_tree_db_adapter.h
+++ b/kvbc/include/merkle_tree_db_adapter.h
@@ -48,13 +48,18 @@ namespace concord::kvbc::v2MerkleTree {
 // getLastReachableBlockId() + 1. Additionally, ReachableSTBlock <= getLatestBlockId().
 class DBAdapter : public IDbAdapter {
  public:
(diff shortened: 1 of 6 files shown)


HristoStaykov:
LGTM, thanks!

HristoStaykov (approved):

f-squirrel (approved):
Excellent! Thank you @dartdart26 for the PR!

PR #855, opened 2020-08-26
Crash the replica in case of pre-execution failure

yuliasherman:
When execute() function returns a zero-sized reply or error, the replica should crash to avoid a blockchain state corruption.
Initial fix for this issue for the pre-execution threw an exception, which was caught by the thread pool code and as a result, a replica did not crash. Now it throws an assert and crashes.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -567,9 +567,8 @@ uint32_t PreProcessor::launchReqPreProcessing(uint16_t clientId,
                                          replicaSpecificInfoLen,
                                          span);
   if (status != 0 || !resultLen) {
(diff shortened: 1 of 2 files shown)


f-squirrel:
As far as I remember, there was a plan to reduce the usage of asserts.
Actually, I'd prefer to remove the `catch` from the [SimpleThreadPool](https://github.com/vmware/concord-bft/blob/master/util/src/SimpleThreadPool.cpp#L88). The only thing it does is hiding real issues.
If the implementation of `Job::execute` did not bother to add `try {} catch {..}` they probably wanted to crash the replica.

yuliasherman:
In general I agree with you - that's why we put an 'exception' originally. But I think that this 'try and catch' might have been be put in the thread pool on purpose and its removal could break other things, that's why I used and 'assert'.

teoparvanov (approved):
LGTM

PR #857, opened 2020-08-27
Terminate when exception leaves Job::execute

f-squirrel:
Rationale:
If the implementation of `Job::execute` does not catch an exception, then `SimpleThreadPool` just hides the problem.

Changes:
* `LOG_FATAL` exception message
* `std::terminate`

### DIFF
--- a/util/src/SimpleThreadPool.cpp
+++ b/util/src/SimpleThreadPool.cpp
@@ -11,6 +11,7 @@
 
 #include "SimpleThreadPool.hpp"
 #include "Logger.hpp"
+#include <exception>
 #include <iostream>
 #include <mutex>
 
@@ -86,9 +87,11 @@ void SimpleThreadPool::execute(Job* j) {
   try {
     j->execute();
   } catch (std::exception& e) {
-    LOG_ERROR(SP, "SimpleThreadPool: exception during execution of " << typeid(*j).name() << " Reason: " << e.what());
+    LOG_FATAL(SP, "SimpleThreadPool: exception during execution of " << typeid(*j).name() << " Reason: " << e.what());
+    std::terminate();
   } catch (...) {
-    LOG_ERROR(SP, "SimpleThreadPool: unknown exception during execution of " << typeid(*j).name());
+    LOG_FATAL(SP, "SimpleThreadPool: unknown exception during execution of " << typeid(*j).name());
+    std::terminate();
   }
 }
 


dartdart26 (changes requested):
I think it is a good idea to pass exceptions to users. However, I think the way the code is written as of now, there is no meaningful context in which we can throw the exception - we can't pass it from the threads to the user in any way. We will just crash the process as we will throw in the lambda passed to the threads.

Additionally, if `execute()` throws, then we will not call `j->release()`:
https://github.com/vmware/concord-bft/blob/d3e34d49bac8114e0bf2c2795e83a99bd31e78da/util/src/SimpleThreadPool.cpp#L34

https://github.com/vmware/concord-bft/blob/d3e34d49bac8114e0bf2c2795e83a99bd31e78da/util/src/SimpleThreadPool.cpp#L62

Please let me know if I am missing something.

f-squirrel:
> I think it is a good idea to pass exceptions to users. I also agree that the code just swallows the exceptions. However, I think the way the code is written as of now, there is no meaningful context in which we can throw the exception - we can't pass it from the threads to the user in any way. We will just crash the process as we will throw in the lambda passed to the threads.
> 
> I don't think we can change the behaviour without changing the interfaces here...
> 
> Additionally, if `execute()` throws, then we will not call `j->release()`:
> https://github.com/vmware/concord-bft/blob/d3e34d49bac8114e0bf2c2795e83a99bd31e78da/util/src/SimpleThreadPool.cpp#L34
> 
> https://github.com/vmware/concord-bft/blob/d3e34d49bac8114e0bf2c2795e83a99bd31e78da/util/src/SimpleThreadPool.cpp#L62
> 
> Please let me know if I am missing something.
> 
> P.S. Maybe your intention is to crash the process. Is that correct? If yes, maybe we can be more explicit and maybe make the `Job::execute()` method `noexcept` or call `std::exit()` or something along those lines. But I think that requires changes in the contract for the `Job::execute()` method and also verifying current use cases.

@dartdart26 , good point, we may call `std::terminate` in the event of an exception.
Thus, the exit code reflects the fact that the process was terminated and we leave an option to override the behavior by setting the [terminate handler](terminate_handler).
WDYT?

dartdart26:
> > I think it is a good idea to pass exceptions to users. I also agree that the code just swallows the exceptions. However, I think the way the code is written as of now, there is no meaningful context in which we can throw the exception - we can't pass it from the threads to the user in any way. We will just crash the process as we will throw in the lambda passed to the threads.
> > I don't think we can change the behaviour without changing the interfaces here...
> > Additionally, if `execute()` throws, then we will not call `j->release()`:
> > https://github.com/vmware/concord-bft/blob/d3e34d49bac8114e0bf2c2795e83a99bd31e78da/util/src/SimpleThreadPool.cpp#L34
> > 
> > https://github.com/vmware/concord-bft/blob/d3e34d49bac8114e0bf2c2795e83a99bd31e78da/util/src/SimpleThreadPool.cpp#L62
> > 
> > Please let me know if I am missing something.
> > P.S. Maybe your intention is to crash the process. Is that correct? If yes, maybe we can be more explicit and maybe make the `Job::execute()` method `noexcept` or call `std::exit()` or something along those lines. But I think that requires changes in the contract for the `Job::execute()` method and also verifying current use cases.
> 
> @dartdart26 , good point, we may call `std::terminate` in the event of an exception.
> Thus, the exit code reflects the fact that the process was terminated and we leave an option to override the behavior by setting the [terminate handler](terminate_handler).
> WDYT?

I would leave as is and add a comment on `Job::execute()` that exceptions are ignored and it is safe to throw.

If we really want to stop the program, I'd say maybe making `Job::execute() noexcept` is the preferred way for me, because it is explicit. However, I am not sure this is a good API for such a generic utility.

dartdart26 (approved):
After a convincing discussion with @f-squirrel offline, we've agreed to log and call `std::terminate()`. That might be the safest thing to do. Thanks @f-squirrel !

PR #858, opened 2020-08-27
Fix uninitialized memory issues in Key Store

dartdart26:
Make sure we don't call `ReplicaKeyStore::deserializeReplicaKeyStore()`
with an empty buffer. Doing so silently fails and leaves the
deserialized object in an arbitrary state. A number of failures in the
Key Manager test in CI were observed.

Make sure a `repID` member is initialized properly in the Key Manager
ReplicaKeyStore.ser_der test. Reported by valgrind.

### DIFF
--- a/bftengine/src/bftengine/KeyStore.cpp
+++ b/bftengine/src/bftengine/KeyStore.cpp
@@ -11,6 +11,8 @@
 
 #include "KeyStore.h"
 
+#include "assertUtils.hpp"
+
 namespace bftEngine::impl {
 
 ////////////////////////////// KEY EXCHANGE MSG//////////////////////////////
@@ -132,6 +134,7 @@ ClusterKeyStore::ClusterKeyStore(const uint32_t& clusterSize,
                                  IReservedPages& reservedPages,
                                  const uint32_t& sizeOfReservedPage)
     : clusterKeys_(clusterSize), reservedPages_(reservedPages), buffer_(sizeOfReservedPage, 0) {
+  ConcordAssertGT(sizeOfReservedPage, 0);
   loadAllReplicasKeyStoresFromReservedPages();
 }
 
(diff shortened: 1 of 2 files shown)


yuliasherman (approved):
LGTM! Thanks for fixing that.

dartdart26:
Restarting the failed job (reconfiguration test) after a discussion with @yontyon - assuming it is a timing issue and it is not related. Reconfiguration tests might be disabled at some point.

bftengine/src/bftengine/KeyStore.cpp
```diff
@@ -153,6 +154,11 @@ void ClusterKeyStore::loadAllReplicasKeyStoresFromReservedPages() {
 
 std::optional<ReplicaKeyStore> ClusterKeyStore::loadReplicaKeyStoreFromReserevedPages(const uint16_t& repID) {
   reservedPages_.loadReservedPage(resPageOffset() + repID, buffer_.size(), buffer_.data());
+  if (buffer_.empty()) {
```
toly-kournik, line 157: you can check the `loadReservedPage()` status

dartdart26: I think that's a good idea, thanks. For it to work in the test, we can change `ReservedPagesMock::loadReservedPage()` so that it returns `false` in case the passed length is 0.

However, in the actual implementation, if there is a pending reserved page, we will call `psd_->getPendingResPage()` with a size of zero and still return `true`:
https://github.com/vmware/concord-bft/blob/master/bftengine/src/bcstatetransfer/BCStateTran.cpp#L562

WDYT?

dartdart26: Actually, as we've discussed offline, the `buffer_` is not supposed to be empty. I will add an assert when constructing the `ClusterKeyStore` instead. It is solely a test issue, it seems - we just need to pass a valid reserved page size instead of 0. Will change the PR accordingly.

dartdart26:
> I would still check the return value to cover the case when `if (lastCheckpoint == 0) return false;`

Do you refer to `reservedPages_.loadReservedPage()`? If yes, any idea what the correct behaviour is if returns `false` - return `nullopt`?

dartdart26:
> > I would still check the return value to cover the case when `if (lastCheckpoint == 0) return false;`
> 
> Do you refer to `reservedPages_.loadReservedPage()`? If yes, any idea what the correct behaviour is if returns `false` - return `nullopt`?

I've added a check of the return value, but I think there is a TODO that we should address, maybe in a separate issue:

```
  // TODO: The check of the return value doesn't discriminate between a missing reserved page + the last checkpoint
  // being 0 and a failure to load. However, the check is still needed as a failure to load might leave the buffer_
  // variable in a wrong state. Additionally, we need to be sure that returning an empty optional is the right thing to
  // do.
```

PR #859, opened 2020-08-27
Use external message size as reply buffer size

Liorgal28:
On this PR i changed maxPreExecResultSize_ to use maxExternalMessageSize and not maxReplyMessageSize, since it already initialized by 16777216 to support pre-execution replies.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -99,7 +99,7 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
       requestsHandler_(requestsHandler),
       myReplica_(myReplica),
       myReplicaId_(myReplica.getReplicaConfig().replicaId),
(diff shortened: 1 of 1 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -99,7 +99,7 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
       requestsHandler_(requestsHandler),
       myReplica_(myReplica),
       myReplicaId_(myReplica.getReplicaConfig().replicaId),
-      maxPreExecResultSize_(myReplica.getReplicaConfig().maxReplyMessageSize - sizeof(ClientReplyMsgHeader)),
+      maxPreExecResultSize_(myReplica.getReplicaConfig().maxExternalMessageSize - sizeof(ClientReplyMsgHeader)),
```
yuliasherman, line 102: It should be just 'myReplica.getReplicaConfig().maxExternalMessageSize'. It does not have any connection to the reply msg.

Liorgal28: Fixed

yuliasherman (approved):
LGTM.

PR #860, opened 2020-08-28
Use '::rocksdb::WriteOptions::sync = true' for all writes to the data…

toly-kournik:
…base


dartdart26 (approved):
Thanks for that @toly-kournik !

Nit: If we decide to merge to `master` at some point, we may want to introduce a function that creates a default `WriteOptions` object (that sets `sync = true`) and use it instead of doing it every time we create an object.

toly-kournik:
We also can make it configurable

andrewjstone:
I'm really happy you made this change for safety sakes Toly. Thanks! I'm also very curious to see how it affects performance. Hopefully not too much.

PR #862, opened 2020-08-30
Decrease amount of unneeded client request retries

yuliasherman:
As we have a distributed system and the replicas perform actions on their own, we always will have unneeded client request retries, but we have to minimize them to fit the feature performance requirements.


teoparvanov (approved):
LGTM!

PR #863, opened 2020-08-30
Introduce Apollo log frame work and replace print calls

oded282:
I used python logging, in order to replace the print calls with logger calls.
I made the following changes:

- Created 'basic_logger.py', the file contains one function which creates the logger, set the configuration and return it.
**configurations:**
default log level - INFO
format - '%(date&time) - %(module_name) - %(level_name) - %(message)', '%d-%m-%Y %H:%M:%S'
logs sent to console

- Imported 'basic_logger.py' and saved an instance of the logger in each test file.
- Used the logger in each test to replace the prints functions calls.

*It seems like a massive change, but most of the changes are replacing the print calls with logger calls.


tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -295,7 +297,7 @@ async def _stop_random_replicas_with_delay(bft_network, delay=10, exclude_replic
         all_replicas = bft_network.all_replicas(without=exclude_replicas)
         random.shuffle(all_replicas)
         for replica in all_replicas:
-            print("stopping replica: ", replica)
+            SkvbcBackupRestoreTest.logger.info(f"stopping replica: {replica}")
```
teoparvanov, line 300: why not simply say: `logger.info(f"stopping replica: {replica}")` ?

Same everywhere else - why would we need to prepend the class name?

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -17,6 +17,7 @@
 
 from util import skvbc as kvbc
 from util.bft import with_trio, with_bft_network, with_constant_load, KEY_FILE_PREFIX
+import basic_logger
```
teoparvanov, line 20: Let's call this apollo_logging.py instead

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -47,6 +48,7 @@ def wrapper(*args, **kwargs):
 class SkvbcBackupRestoreTest(unittest.TestCase):
 
     __test__ = False  # so that PyTest ignores this test scenario
+    logger = basic_logger.get_logger(__name__)
```
teoparvanov, line 51: For simplicity, let's call this `log` instead of `logger`

teoparvanov:
Hi @andrewjstone, thanks, I actually didn't know about Eliot. It looks awesome, and it seems to go in the right direction for us.

In this case, it doesn't make sense to proceed with introducing Python logging at the moment. 

I suggest to close this PR and instead experiment with Eliot a little bit, especially to make sure the integration with Trio works correctly. I'll sync additionally with @oded282 later today to refocus the effort.

PR #867, opened 2020-08-31
Fix taking in consideration that we do not have ViewChangeMsg-s and

HristoStaykov:
NewViewMsg for View 0.


teoparvanov (approved):
LGTM, thanks @HristoStaykov !

PR #870, opened 2020-09-02
Release pre-execution request memory before archiving

yuliasherman:
- Archiving the full client messages is memory consuming and actually is not required => release them before saving in a history.
- Pass ClientPreProcessReqMsg to the primary replica only when registration passes to reduce unneeded request retries and lower network load.


teoparvanov (approved):
LGTM!

PR #871, opened 2020-09-02
Manage SeqNumberGeneratorForClientRequests

robem:
Raw pointers shouldn't imply ownership of the underlying memory. Instead, smart
pointers is the way to go. As a side effect, this change fixes a small memory
leak.


f-squirrel (approved):
Great, thank you @robem

PR #872, opened 2020-09-03
Move BLS Multisig signers and verifiers to a CryptoManager from ReplicaConfig

toly-kournik:
- Introduce CryptoManager class holding all the BLS signers and verifiers.
- ReplicaConfig holds only BLS configuration, i.e. system type, subtype, number of signers and bootstrap keys.
- disable tests/simpleTest/persistency_test as it runs 4 replicas in threads and it doesn't conform to a CryptoManager singleton paradigm.


f-squirrel (changes requested):
Thank you for the great PR, I think it makes the code base much cleaner.
In my opinion, if a singleton does not conform with the unit tests, the singleton has to be replaced with something else, not the unit tests. For example, `CryptoManager` can be passed as a constant reference.

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+  IThresholdSigner* thresholdSignerForSlowPathCommit() const { return thresholdSigner_.get(); }
+  IThresholdVerifier* thresholdVerifierForSlowPathCommit() const { return thresholdVerifierForSlowPathCommit_.get(); }
+
+  IThresholdSigner* thresholdSignerForCommit() const { return thresholdSigner_.get(); }
+  IThresholdVerifier* thresholdVerifierForCommit() const { return thresholdVerifierForCommit_.get(); }
+
+  IThresholdSigner* thresholdSignerForOptimisticCommit() const { return thresholdSigner_.get(); }
+  IThresholdVerifier* thresholdVerifierForOptimisticCommit() const {
+    return thresholdVerifierForOptimisticCommit_.get();
```
cloudnoize, line 37: If I understand correctly, those pointers can be invalidated on rotation, right ? 
Are they being held by any client of those methods?

bftengine/src/bftengine/ReplicaConfigSerializer.cpp
```diff
@@ -117,14 +117,12 @@ void ReplicaConfigSerializer::serializeDataMembers(ostream &outStream) const {
...
-  serializePointer(config_->thresholdVerifierForSlowPathCommit, outStream);
-
-  serializePointer(config_->thresholdSignerForCommit, outStream);
-  serializePointer(config_->thresholdVerifierForCommit, outStream);
-
-  serializePointer(config_->thresholdSignerForOptimisticCommit, outStream);
-  serializePointer(config_->thresholdVerifierForOptimisticCommit, outStream);
+  // Serialize threshold crypto system
+  concord::serialize::Serializable::serialize(outStream, config_->thresholdSystemType_);
+  concord::serialize::Serializable::serialize(outStream, config_->thresholdSystemSubType_);
```
cloudnoize, line 122: A lot of `serialize` we should rename it :)

bftengine/src/bftengine/ReplicaConfigSerializer.cpp
```diff
@@ -117,14 +117,12 @@ void ReplicaConfigSerializer::serializeDataMembers(ostream &outStream) const {
...
+  serializeKey(config_->thresholdPrivateKey_, outStream);  // TODO [TK] serialize to a secret key store
+  serializeKey(config_->thresholdPublicKey_, outStream);
```
cloudnoize, line 124: the same question as above, it's for the bootstrap only keys, right?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3148,7 +3147,8 @@ ReplicaImp::ReplicaImp(bool firstTime,
                                 config_.replicaPrivateKey,
                                 config_.publicKeysOfReplicas);
     repsInfo = new ReplicasInfo(config_, dynamicCollectorForPartialProofs, dynamicCollectorForExecutionProofs);
-    viewsManager = new ViewsManager(repsInfo, sigManager, config_.thresholdVerifierForSlowPathCommit);
+    viewsManager =
+        new ViewsManager(repsInfo, sigManager, CryptoManager::instance().thresholdVerifierForSlowPathCommit());
```
cloudnoize, line 3151: I guess this one caches the pointer and will need to implement the interface for key exchange registration

bftengine/src/bftengine/messages/PartialProofsSet.cpp
```diff
@@ -304,19 +305,12 @@ void PartialProofsSet::tryToCreateFullProof() {
...
-  IThresholdVerifier* verifier;
-
   // TODO: ALIN: Not sure if the commented code below would be the desired behavior
   if (cPath == CommitPath::OPTIMISTIC_FAST) {
-    verifier = replica->getThresholdVerifierForOptimisticCommit();
+    return CryptoManager::instance().thresholdVerifierForOptimisticCommit();
   } else /* if (cPath == CommitPath::FAST_WITH_THRESHOLD) */ {
-    verifier = replica->getThresholdVerifierForCommit();
-    //} else {
+    return CryptoManager::instance().thresholdVerifierForCommit();
```
cloudnoize, line 312: I might be wrong, but it seems that the class `PartialProofsSet` caches the pointer e.g. `IThresholdAccumulator* thresholdAccumulatorForFast`, might it be a problem in case of rotation?

cloudnoize (commented):
LGTM left few questions

bftengine/include/bftengine/CryptoManager.hpp
toly-kournik: These pointers are not held by anybody

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -82,24 +83,18 @@ struct ReplicaConfig {
   // If equals to 0, a default number of min(thread::hardware_concurrency(), numOfClients) is used
   uint16_t preExecConcurrencyLevel = 0;
 
-  // public keys of all replicas. map from replica identifier to a public key
+  // RSA public keys of all replicas. map from replica identifier to a public key
   std::set<std::pair<uint16_t, const std::string>> publicKeysOfReplicas;
 
-  // private key of the current replica
+  // RSA private key of the current replica
   std::string replicaPrivateKey;
```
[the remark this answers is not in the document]
toly-kournik: No, those are RSA keys used for ViewChange.
Eventually they'll be replaced by the same single  Multisig key pair.

bftengine/src/bftengine/ReplicaConfigSerializer.cpp
toly-kournik: Right.
Eventually the private key will be sent to a secret store and public keys will "live" in ReservedPages

bftengine/src/bftengine/ReplicaImp.cpp
toly-kournik: This is a good catch! I'll change it not to pass it as an argument.

bftengine/src/bftengine/messages/PartialProofsSet.cpp
toly-kournik: yes, we should remove the usage of
` IThresholdAccumulator* thresholdAccumulatorForFast;`
 ` IThresholdAccumulator* thresholdAccumulatorForOptimisticFast;`

bftengine/src/bftengine/ReplicaConfigSerializer.cpp
toly-kournik: > Thank you for the great PR, I think it makes the code base much cleaner.
> In my opinion, if a singleton does not conform with the unit tests, the singleton has to be replaced with something else, not the unit tests. For example, `CryptoManager` can be passed as a constant reference.

This unittest always was a problematic one because of the way it was written. I don't think it's essential and needed to be maintained.

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include "threshsign/ThresholdSignaturesTypes.h"
+#include "threshsign/IThresholdSigner.h"
+#include "threshsign/IThresholdVerifier.h"
+#include "ReplicaConfig.hpp"
+
+namespace bftEngine {
+
+class CryptoManager {
+ public:
+  static CryptoManager& instance(const ReplicaConfig* config = nullptr, Cryptosystem* cryptoSys = nullptr) {
```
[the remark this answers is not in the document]
toly-kournik: I'll put it in my list of minor leftovers. I also wanted to rename the file later to reflect the class defined inside.

bftengine/src/bftengine/ReplicaLoader.cpp
```diff
@@ -153,6 +140,15 @@ ReplicaLoader::ErrorCode loadConfig(shared_ptr<PersistentStorage> &p, LoadedRepl
                                  replicasSigPublicKeys);
 
   ld.repsInfo = new ReplicasInfo(ld.repConfig, dynamicCollectorForPartialProofs, dynamicCollectorForExecutionProofs);
+
+  Cryptosystem *cryptoSys = new Cryptosystem(ld.repConfig.thresholdSystemType_,
```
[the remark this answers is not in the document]
toly-kournik: It happens at system start and if there is exception the process won't start.
The raw pointer is immediately passed to the CryptoManager c-r where it's held by unique_ptr.

toly-kournik:
@cloudnoize I want to commit this PR to let you start working on the integration.
Would you mind if we fix two issues above not in this PR?

cloudnoize:
> @cloudnoize I want to commit this PR to let you start working on the integration.
> Would you mind if we fix two issues above not in this PR?

I don't mind, we'll add them to a ticket in order not to forget

cloudnoize (approved):

PR #874, opened 2020-09-03
chore: replace exit(1) with std::terminate

f-squirrel:
Rationale:
Exit code 1 usually means that the process has failed in a controlled way.
For example: `grep -r "non-existing string" .` returns 1 which
means that it failed to find the string but did not crash.
In Concord-BFT `exit(1)` is mostly used when an unexpected exception is caught and the library has to crash.
`std::terminate` crashes the process with `SIGABRT` and generates a core dump.
However, if a user of Concord-BFT would like to change the behavior they can set a custom terminate handler[1].

[1] - https://en.cppreference.com/w/cpp/error/set_terminate

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,10 @@ signed our contributor license agreement (CLA), our bot will update the issue wh
 questions about the CLA process, please refer to our [FAQ](https://cla.vmware.com/faq). For more detailed information,
 refer to [CONTRIBUTING.md](CONTRIBUTING.md).
 
+## Notes
+The library calls `std::terminate()` when it cannot continue in a safe manner.
+In that way, users can install a handler that does something different than just calling `std::abort()`.
+
 ## Community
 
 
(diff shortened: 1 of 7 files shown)


dartdart26 (approved):
Thanks @f-squirrel , I think that makes sense!

Maybe we can document somewhere (maybe in the README?) that the library calls `std::terminate()` when it cannot continue in a safe manner. In that way, users can install a handler that does something different than just calling `std::abort()`.

Lastly, I have a single question. Is the intention also to abort as quickly as possible and **not** call static variable destructors, automatic variable destructors and so on - as done by `std::abort()`?

f-squirrel:
@dartdart26 , thank you for the review, I've updated the README.
Since we call `std::terminate` when the state of the library is undefined I am not sure that calling destructors makes sense.
It may lead to another exception(in the destructors).
Anyways, a user may set a handler that does something more meaningful.
WDYT?

dartdart26:
> @dartdart26 , thank you for the review, I've updated the README.
> Since we call `std::terminate` when the state of the library is undefined I am not sure that calling destructors makes sense.
> It may lead to another exception(in the destructors).
> Anyways, a user may set a handler that does something more meaningful.
> WDYT?

Thanks for the update @f-squirrel .

It makes sense about destructors and the handler, I am thinking the same.

PR #876, opened 2020-09-08
Add functionality for removing metadata before shutting a replica down.

yontyon:
In some cases, we want to remove the metadata storage before shutting the replica down.
For example:
1. On removing a node we want the replicas to load a different configuration file than the one it has.
2. On adding a node we want the replicas to load a different configuration file than the one it has.

For that, we have added a core functionality for removing metadata of a replica.
A replica that removes its metadata before shutting down will be reloaded with all its previous state but without stored metadata.
Instead, it will consume the metadata from the configuration file.

### DIFF
--- a/bftengine/include/bftengine/MetadataStorage.hpp
+++ b/bftengine/include/bftengine/MetadataStorage.hpp
@@ -48,6 +48,9 @@ class MetadataStorage {
   virtual void beginAtomicWriteOnlyBatch() = 0;
   virtual void writeInBatch(uint32_t objectId, char *data, uint32_t dataLength) = 0;
   virtual void commitAtomicWriteOnlyBatch() = 0;
+
+  // In some cases, we would like to load a new metadata after a crash (for example, on reconfiguration actions).
+  virtual void setDontLoadStorageOnStartupFlag() = 0;
 };
 
 }  // namespace bftEngine
--- a/bftengine/src/bcstatetransfer/DBDataStore.cpp
+++ b/bftengine/src/bcstatetransfer/DBDataStore.cpp
@@ -401,6 +401,22 @@ void DBDataStore::deleteCoveredResPageInSmallerCheckpoints(uint64_t minChkp) {
   inmem_->deleteCoveredResPageInSmallerCheckpoints(minChkp);
 }
 
+void DBDataStore::clearDataStoreData() {
+  del(Initialized);
+  del(MyReplicaId);
+  del(MaxNumOfStoredCheckpoints);
+  del(NumberOfReservedPages);
+  del(LastStoredCheckpoint);
+  del(FirstStoredCheckpoint);
+  del(IsFetchingState);
+  del(fVal);
+  del(FirstRequiredBlock);
+  del(LastRequiredBlock);
+  del(Replicas);
+  del(CheckpointBeingFetched);
+  deleteAllPendingPages();
+}
+
 /** ******************************************************************************************************************/
 }  // namespace impl
 }  // namespace SimpleBlockchainStateTransfer
--- a/bftengine/src/bftengine/DbMetadataStorage.cpp
+++ b/bftengine/src/bftengine/DbMetadataStorage.cpp
@@ -159,7 +159,12 @@ void DBMetadataStorage::cleanDB() {
 
 DBMetadataStorage::~DBMetadataStorage() {
   if (dontLoadStorageOnStartup) {
-    cleanDB();
+    try {
+      cleanDB();
+    } catch (std::exception &e) {
+      LOG_ERROR(logger_, e.what());
+      std::terminate();
+    }
   }
 }
 
--- a/bftengine/src/bftengine/DbMetadataStorage.cpp
+++ b/bftengine/src/bftengine/DbMetadataStorage.cpp
@@ -162,7 +162,7 @@ DBMetadataStorage::~DBMetadataStorage() {
     try {
       cleanDB();
     } catch (std::exception &e) {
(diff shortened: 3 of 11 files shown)


bftengine/src/bftengine/DbMetadataStorage.cpp
```diff
@@ -146,6 +146,30 @@ Status DBMetadataStorage::multiDel(const ObjectIdsVector &objectIds) {
...
+  // Second, to clean things up, clear all the records from the DB
+  ObjectIdsVector objectIds;
+  for (const auto &id : objectIdToSizeMap_) {
+    objectIds.push_back(id.first);
+  }
+  multiDel(objectIds);
+}
+
+DBMetadataStorage::~DBMetadataStorage() {
+  if (dontLoadStorageOnStartup) {
```
f-squirrel, line 169: Since exceptions are not allowed to leave destructors, it might be useful to catch exceptions here, log them, and `std::terminate()` the application.

f-squirrel (commented):
I suggest to **delete** all the metadata fields including `objectsNumParameterId_` instead setting it to `0` because Metadata storage checks the existence of the field in the DB.
Consider deleting all the State Transfer data as `db_editor` does. (@toly-kournik please correct me if I am wrong).

bftengine/tests/simpleStorage/FileStorage.cpp
```diff
@@ -43,6 +43,9 @@ FileStorage::FileStorage(Logger &logger, const string &fileName) : logger_(logge
 }
 
 FileStorage::~FileStorage() {
+  if (dontLoadStorageOnStartup) {
```
toly-kournik, line 46: Do we need FileStorage at all?

bftengine/include/bftengine/MetadataStorage.hpp
```diff
@@ -48,6 +48,9 @@ class MetadataStorage {
   virtual void beginAtomicWriteOnlyBatch() = 0;
   virtual void writeInBatch(uint32_t objectId, char *data, uint32_t dataLength) = 0;
   virtual void commitAtomicWriteOnlyBatch() = 0;
+
+  // In some cases, we would like to load a new metadata after a crash (for example, on reconfiguration actions).
+  virtual void setDontLoadStorageOnStartupFlag() = 0;
```
toly-kournik, line 53: What's the trigger for calling this method, and who does it?

toly-kournik:
@yontyon, are you going to do the same for State Transfer?
Why the dbeditor's ability added by @f-squirrel is not enough?

bftengine/src/bftengine/DbMetadataStorage.cpp
yontyon: done

bftengine/tests/simpleStorage/FileStorage.cpp
yontyon: I don't know, it is there so I have handled it too.

bftengine/include/bftengine/MetadataStorage.hpp
yontyon: The trigger is a reconfiguration action that requires removing the old metadata before shutting down the replicas.
For example, on removing a node, we need to update the node's configuration to commit this node details.
So, this flag will be set, and before the replicas will be shut down, the metadata will be deleted.
Thus, once the replicas reloaded, they will start with the previous state and with no metadata (similar to what @f-squirrel is doing).
The reason why we don't use @f-squirrel tool is that its tool is only for the Merkle tree storage implementation and we want this feature to be part of the reconfiguration feature which is agnostic to the storage implementation.

yontyon:
@toly-kournik see https://github.com/vmware/concord-bft/pull/876#discussion_r484940297

bftengine/include/bftengine/MetadataStorage.hpp
toly-kournik: > The reason why we don't use @f-squirrel tool is that its tool is only for the Merkle tree storage implementation and we want this feature to be part of the reconfiguration feature which is agnostic to the storage implementation.
It's doable - just use a proper key manipulator.

bftengine/tests/simpleStorage/FileStorage.cpp
toly-kournik: ask @yuliasherman , we can probably through it away

bftengine/src/bcstatetransfer/DBDataStore.cpp
```diff
@@ -401,6 +401,22 @@ void DBDataStore::deleteCoveredResPageInSmallerCheckpoints(uint64_t minChkp) {
   inmem_->deleteCoveredResPageInSmallerCheckpoints(minChkp);
 }
 
+void DBDataStore::clearDataStoreData() {
+  del(Initialized);
```
toly-kournik, line 405: it's enough to delete Initialized only

toly-kournik (commented):
> Consider deleting all the State Transfer data as `db_editor` does. (@toly-kournik please correct me if I am wrong).
Yep, I personally don't like the idea of deleting it from inside the process.

bftengine/src/bcstatetransfer/DBDataStore.cpp
yontyon: @f-squirrel recommended to delete it all, so if it isn't matter I will keep it that way.
As for the file storage, I will ask @yuliasherman but in any case, this is not part of this PR.

f-squirrel (approved):
LGTM!
Thank you for addressing the comments.

PR #878, opened 2020-09-09
Decentralized key generation:

toly-kournik:
- Add share secret and verification key pair generation ability to CrytptoSystem.
- Add corresponding API to CryptoManager.
- Add test for above key generation to TestGeneratedKeys.

### DIFF
--- a/bftengine/include/bftengine/CryptoManager.hpp
+++ b/bftengine/include/bftengine/CryptoManager.hpp
@@ -37,28 +37,40 @@ class CryptoManager {
     return thresholdVerifierForOptimisticCommit_.get();
   }
 
- private:
-  CryptoManager(const ReplicaConfig* config, Cryptosystem* cryptoSys) {
-    multiSigCryptoSystem_.reset(cryptoSys);
-    thresholdSigner_.reset(multiSigCryptoSystem_->createThresholdSigner());
-    thresholdVerifierForSlowPathCommit_.reset(
-        multiSigCryptoSystem_->createThresholdVerifier(config->fVal * 2 + config->cVal + 1));
-    thresholdVerifierForCommit_.reset(
-        multiSigCryptoSystem_->createThresholdVerifier(config->fVal * 3 + config->cVal + 1));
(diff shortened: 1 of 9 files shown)


cloudnoize (approved):
LVGTM

threshsign/include/threshsign/ThresholdSignaturesTypes.h
```diff
@@ -210,13 +157,20 @@ class Cryptosystem {
    */
   void generateNewPseudorandomKeys();
 
+  /**
+   *  Generate a single key pair.
+   *  Used for key rotation.
+   *
+   *  @return pair<shareSecretKey, shareVerificationKey>
+   */
+  std::pair<std::string, std::string> generateNewKeyPair();
```
cloudnoize, line 166: This method should be reached via the CryptoManager API

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -38,28 +38,38 @@ class CryptoManager {
...
+  }
+
+  void updateMultisigKeys(const std::string& secretKey, const std::string& verificationKey) {
+    multiSigCryptoSystem_->updateKeys(secretKey, verificationKey);
+    init();
+  }
+
+  void updateVerificationKeyForSigner(const std::string& verificationKey, const std::uint16_t& signerIndex) {
+    multiSigCryptoSystem_->updateVerificationKey(verificationKey, signerIndex);
+    init();
```
cloudnoize, line 61: Can this method get a boolean `init` , in order to select whether to init on update, since most of the time I'll need to update multiple public keys and call it multiple times.

toly-kournik: You always need to do init on update as you don't know if there will be another one.

cloudnoize: I know if multiple public keys need to be updated at once, in that case, it's better to call init once

toly-kournik: It's a rare event we can not worry about performance here, we can update one by one

PR #881, opened 2020-09-10
Parameterize diagnostics server with host and port

andrewjstone:
This allows flexibility for docker deployments, such accessing the
server from other containers on the same host.

### DIFF
--- a/diagnostics/test/main.cpp
+++ b/diagnostics/test/main.cpp
@@ -19,6 +19,8 @@
 
 using namespace concord::diagnostics;
 
+static constexpr uint16_t PORT = 6888;
+
 int main() {
   Registrar registrar;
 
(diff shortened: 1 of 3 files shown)


Liorgal28 (approved):
Amazing :) Thanks

diagnostics/test/main.cpp
```diff
@@ -29,7 +31,7 @@ int main() {
   registrar.registerStatusHandler(handler2);
 
   concord::diagnostics::Server diagnostics_server;
-  diagnostics_server.start(registrar);
+  diagnostics_server.start(registrar, INADDR_LOOPBACK, PORT);
```
robem, line 34: Did you test that this works with container to container communication? I thought we need to use the docker bridge device or `INADDR_ANY`.

andrewjstone: Nope, this is just a test main that you have to run manually anyway. I only use it for validating that ./concop works as expected. The change in the product will pass in INADDR_ANY

PR #882, opened 2020-09-10
Fix for stuck rebuild of the previous View working window.

HristoStaykov:
In case the Primary has missing information for a given Sequence ID
it has to ask all other Replicas.


teoparvanov (approved):
Great fix, thank you @HristoStaykov !

andrewjstone (commented):
Awesome!

PR #884, opened 2020-09-14
Added option GRACEFUL_SHUTDOWN to control the shutdown of Replicas in…

HristoStaykov:
… Apollo tests.

This option is set to TRUE if LEAKCHECK=TRUE, because it is needed in order to
get reports from Sanitizers.
GRACEFUL_SHUTDOWN=TRUE - shutdown with SIGTERM
GRACEFUL_SHUTDOWN=FALSE - shutdown with SIGKILL


teoparvanov (approved):
LGTM, thanks @HristoStaykov !

PR #888, opened 2020-09-15
Pre-execution: resend PreProcessRequestMsg to replicas that have previously rejected it

yuliasherman:
In case non-primary replicas are busy processing requests, they reject PreProcessRequestMsg and this could cause a primary never collecting enough replies for the pre-execution consensus. This fix changes the pre-execution protocol adding a response status field to the PreProcessReply message , which could contain REJECT value. In this case the primary replica resends the PreProcessRequestMsg message.


teoparvanov (approved):
LGTM, thanks Yulia!

PR #889, opened 2020-09-15
chore(valgrind): Replace char* buffer with vector

f-squirrel:
Valgrind falsely reports this as a memory leak


andrewjstone (approved):
LGTM! Thanks for the fix @f-squirrel

PR #893, opened 2020-09-21
State Transfer persistence inconsistency bug fix

toly-kournik:
- Remove usage of key indirection in DBDataStore.
- Add concord::storage::ISTKeyManipulator::getReservedPageKeyPrefix().
- Improved a way Handoff is shutting down.
- Logging improvements.

### DIFF
--- a/bftengine/src/bcstatetransfer/DBDataStore.cpp
+++ b/bftengine/src/bcstatetransfer/DBDataStore.cpp
@@ -205,31 +205,7 @@ void DBDataStore::deleteCheckpointBeingFetched() {
   del(CheckpointBeingFetched);
   inmem_->deleteCheckpointBeingFetched();
 }
-/** *******************************************************************************************************************
- *  Reserved Pages
- *
- *  C = lastStoredCheckpoint
- *  n = maxNumOfStoredCheckpoints
- *  K = numberOfReservedPages
- *  #keys = n x K
- *
- *               |page1        |page2        |       |pageK        |
- *               |1|2|3|4|...|n|1|2|3|4|...|n| ..... |1|2|3|4|...|n|
- *               | | | |     |                       | | | |     |
- *         rp11 -| | | |     |                       | | | |     |-- rpKn
- *         rp12 ---| | |     |                       | | | |         ...
- *         rp13 -----| |     |                       | | | |-------- rpK4
- *         rp14 -------|     |                       | | |---------- rpK3
- *         ...               |                       | |------------ rpK2
- *         rp1n -------------|                       |-------------- rpK18
- *
- * Keys indirection table maps keys in the pages array to the actual (dynamic page key):
- *
- * rp_pid_idx -> page_pid_checkpoint
- *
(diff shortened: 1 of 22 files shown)


bftengine/src/bcstatetransfer/DBDataStore.cpp
```diff
@@ -256,40 +232,30 @@ void DBDataStore::deserializeResPage(
...
+    std::istringstream iss(
+        std::string(reinterpret_cast<const char*>(keyValue.second.data()), keyValue.second.length()));
+    uint32_t pageid;
+    uint64_t checkpoint;
+    STDigest digest;
+    char* page = new char[inmem_->getSizeOfReservedPage()];
+    deserializeResPage(iss, pageid, checkpoint, digest, page);
+    inmem_->setResPage(pageid, checkpoint, digest, page);
+    delete[] page;  // is copied
+    LOG_DEBUG(logger(), KVLOG(pageid, checkpoint, digest.toString()));
```
cloudnoize, line 248: I would add a comment describing the algorithm

cloudnoize (changes requested):
LGTM

bftengine/src/bcstatetransfer/DBDataStore.cpp
toly-kournik: sure

andrewjstone:
With regards to the commit message: What was the actual bug, and how does this code fix it?

toly-kournik:
In State Transfer persistence module we maintain in-memory as well as persistent states of the reserved pages which in some scenarios were not aligned. This is exactly what was reported by an assert in checkStoredCheckpoints() method which is a part of checking State Transfer consistency process.

The problem was the way we used to store Reserved Pages in rocksdb using keys indirection:
For every page k there were n (max number of stored checkpoints) static keys: [k,0], [k,1] ... [k,n].
Keys indirection table mapped static keys to the actual dynamic page key:

[pageId, chckpId mod n] -> [pageId, chckpId]  -> ResPageData
In several scenarios the value for a new reserved page has overwritten the value of a previous reserved page which yielded the same index (mod n), while this page wasn't still removed from the in-memory state and was taken for calculating of the reserved pages digest.

andrewjstone:
> In State Transfer persistence module we maintain in-memory as well as persistent states of the reserved pages which in some scenarios were not aligned. This is exactly what was reported by an assert in checkStoredCheckpoints() method which is a part of checking State Transfer consistency process.
> 
> The problem was the way we used to store Reserved Pages in rocksdb using keys indirection:
> For every page k there were n (max number of stored checkpoints) static keys: [k,0], [k,1] ... [k,n].
> Keys indirection table mapped static keys to the actual dynamic page key:
> 
> [pageId, chckpId mod n] -> [pageId, chckpId] -> ResPageData
> In several scenarios the value for a new reserved page has overwritten the value of a previous reserved page which yielded the same index (mod n), while this page wasn't still removed from the in-memory state and was taken for calculating of the reserved pages digest.

Thank you for the great explanation Toly!

PR #894, opened 2020-09-21
Skip unstable tests

yuliasherman:
Two recently added pre-execution tests are unstable - disabling (example: https://travis-ci.com/github/vmware/concord-bft/jobs/388036688).


teoparvanov (approved):
LGTM, will have a look at those two when I'm back.

PR #897, opened 2020-09-23
State Transfer bug

toly-kournik:
When deleting old checkpoints during the current state checkpointing,
if previous checkpoints have been received via ST, new firstStoredCheckpoint might have no checkpoint description.
This leads to checkFirstAndLastCheckpoint() to assert.
This logic is checked when we a receive a new vBlock (and deleting old checkpoints).
Solution: use deleteOldCheckpoints() in both places, and check for checkpoint description existence.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -452,20 +452,23 @@ STDigest BCStateTran::checkpointReservedPages(uint64_t checkpointNumber, DataSto
 
 void BCStateTran::deleteOldCheckpoints(uint64_t checkpointNumber, DataStoreTransaction *txn) {
   uint64_t minRelevantCheckpoint = 0;
-  if (checkpointNumber >= maxNumOfStoredCheckpoints_) {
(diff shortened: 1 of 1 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -452,20 +452,23 @@ STDigest BCStateTran::checkpointReservedPages(uint64_t checkpointNumber, DataSto
...
-  if (checkpointNumber >= maxNumOfStoredCheckpoints_) {
+  if (checkpointNumber >= maxNumOfStoredCheckpoints_)
     minRelevantCheckpoint = checkpointNumber - maxNumOfStoredCheckpoints_ + 1;
-  }
 
   const uint64_t oldFirstStoredCheckpoint = txn->getFirstStoredCheckpoint();
 
+  if (minRelevantCheckpoint > 0)
+    while (minRelevantCheckpoint < checkpointNumber && !txn->hasCheckpointDesc(minRelevantCheckpoint))
+      minRelevantCheckpoint++;
```
cloudnoize, line 462: nit: why not using sone curly braces

cloudnoize (approved):
LGTM

PR #900, opened 2020-09-24
Add Apollo's logs to GitHub Action artifacts

oded282:
I added the replica and Eliot logs to GithubHub Action artifacts.

- Set "KEEP_APOLLO_LOGS"=True - saves the logs to /build/tests/apollo/logs directory.
- Set "KEEP_APOLLO_LOGS"=True when CI runs.
- Commented Eliot's logs to console option - Eliot's logs together with the replicas logs in unreadable in the console.

Re-arrange Apollo's logs.

- Separated between the replicas **v2merkle** and **v1direct** logs.
- Created two directories for each test, one for **v2merkle** and another one for **v1direct**.

*Each directory has Eliot’s logs and the replicas logs for each test configuration.

Example:
Screen Shot 2020-09-24 at 17 18 49

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -47,7 +47,7 @@ jobs:
                               ${{ matrix.use_s3_obj_store }} \
                               -DUSE_OPENTRACING=ON \
                               -DOMIT_TEST_OUTPUT=OFF\" "\
-              && script -q -e -c "make test"
+              && script -q -e -c "make test -DKEEP_APOLLO_LOGS=TRUE"
         - name: Prepare artifacts
           if: failure()
           run: |
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -46,14 +46,14 @@ jobs:
                               -DBUILD_ROCKSDB_STORAGE=TRUE \
                               ${{ matrix.use_s3_obj_store }} \
                               -DUSE_OPENTRACING=ON \
-                              -DOMIT_TEST_OUTPUT=OFF\" "\
-              && script -q -e -c "make test -DKEEP_APOLLO_LOGS=TRUE"
+                              -DOMIT_TEST_OUTPUT=OFF\
+                              -DKEEP_APOLLO_LOGS=TRUE\" "\
+              && script -q -e -c "make test"
         - name: Prepare artifacts
           if: failure()
           run: |
             sudo chown -R $USER:$GROUP ${PWD}/build
             mv ${PWD}/build ${{ github.workspace }}/artifact/
-            cp /tmp/apollo ${{ github.workspace }}/artifact/
             ls -lh ${{ github.workspace }}/artifact/cores
             sudo chown -R $USER:$GROUP ${{ github.workspace }}/artifact/
         - name: Upload artifacts
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -53,6 +53,7 @@ jobs:
           run: |
             sudo chown -R $USER:$GROUP ${PWD}/build
             mv ${PWD}/build ${{ github.workspace }}/artifact/
+            cp /tmp/apollo ${{ github.workspace }}/artifact/
             ls -lh ${{ github.workspace }}/artifact/cores
             sudo chown -R $USER:$GROUP ${{ github.workspace }}/artifact/
         - name: Upload artifacts
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,8 @@ CONCORD_BFT_CMAKE_FLAGS:= \
 			-DUSE_S3_OBJECT_STORE=TRUE \
 			-DUSE_OPENTRACING=ON \
 			-DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
-			-DOMIT_TEST_OUTPUT=OFF
+			-DOMIT_TEST_OUTPUT=OFF \
+			-DKEEP_APOLLO_LOGS=TRUE
 
 # The consistency parameter makes sense only at MacOS.
 # It is ignored at all other platforms.
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -22,6 +22,7 @@ import subprocess
 from collections import namedtuple
 import tempfile
 from functools import wraps
+from datetime import datetime
 import inspect
 
 import trio
@@ -138,7 +139,9 @@ def with_bft_network(start_replica_cmd, selected_configs=None, num_clients=None,
                                         stop_replica_cmd=None,
                                         num_ro_replicas=num_ro_replicas)
                     with BftTestNetwork.new(config) as bft_network:
-                        bft_network.current_test = async_fn.__name__ + "_n=" + str(bft_config['n']) \
+                        storage_type = os.environ.get("STORAGE_TYPE")
+                        bft_network.current_test = async_fn.__name__ + "_" + storage_type \
+                                                                     + "_n=" + str(bft_config['n']) \
                                                                      + "_f=" + str(bft_config['f']) \
                                                                      + "_c=" + str(bft_config['c'])
                         print(f'Running {async_fn.__name__} '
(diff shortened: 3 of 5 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -47,12 +47,13 @@ jobs:
                               ${{ matrix.use_s3_obj_store }} \
                               -DUSE_OPENTRACING=ON \
                               -DOMIT_TEST_OUTPUT=OFF\" "\
-              && script -q -e -c "make test"
+              && script -q -e -c "make test -DKEEP_APOLLO_LOGS=TRUE"
         - name: Prepare artifacts
           if: failure()
           run: |
             sudo chown -R $USER:$GROUP ${PWD}/build
             mv ${PWD}/build ${{ github.workspace }}/artifact/
```
teoparvanov, line 55: To reduce the bundle size, let's comment out this line as follows:
```
# Uncomment if core dump analysis is needed
# mv ${PWD}/build ${{ github.workspace }}/artifact/
```

Makefile
```diff
@@ -20,7 +20,8 @@ CONCORD_BFT_CMAKE_FLAGS:= \
 			-DUSE_S3_OBJECT_STORE=TRUE \
 			-DUSE_OPENTRACING=ON \
 			-DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
-			-DOMIT_TEST_OUTPUT=OFF
+			-DOMIT_TEST_OUTPUT=OFF \
+			-DKEEP_APOLLO_LOGS=TRUE
```
teoparvanov, line 24: We should "KEEP_APOLLO_LOGS" only in CI. When running locally, it should be up to the developer to activate this or not.

tests/apollo/util/bft.py
```diff
@@ -394,13 +397,27 @@ def start_replica(self, replica_id):
         with log.start_action(action_type="start_replica"):
             stdout_file = None
             stderr_file = None
+            storage_type = os.environ.get("STORAGE_TYPE")
+            tests_names = [m for m in sys.modules.keys() if m.startswith("test_")]
```
teoparvanov, line 401: @oded282 you are maybe right about the env variable. I don't like to have STORAGE_TYPE propagate down here.
Let's try your suggestion with the TEST_NAME env variable.

.github/workflows/build_and_test.yml
```diff
@@ -47,12 +47,13 @@ jobs:
                               ${{ matrix.use_s3_obj_store }} \
                               -DUSE_OPENTRACING=ON \
                               -DOMIT_TEST_OUTPUT=OFF\" "\
-              && script -q -e -c "make test"
+              && script -q -e -c "make test -DKEEP_APOLLO_LOGS=TRUE"
```
f-squirrel, line 50: The test fails because `make` does not accept variables in this way, error message:
```
2020-09-24T16:07:01.4392760Z Build finished. The binaries are in /home/runner/work/concord-bft/concord-bft/build
2020-09-24T16:07:01.4622684Z make: invalid option -- 'D'
2020-09-24T16:07:01.4623320Z make: invalid option -- 'K'
2020-09-24T16:07:01.4623799Z make: invalid option -- 'E'
2020-09-24T16:07:01.4624271Z make: invalid option -- 'E'
2020-09-24T16:07:01.4624745Z make: invalid option -- 'P'
2020-09-24T16:07:01.4625213Z make: invalid option -- '_'
2020-09-24T16:07:01.4625682Z make: invalid option -- 'A'
2020-09-24T16:07:01.4626154Z make: invalid option -- 'P'
```

.github/workflows/build_and_test.yml
f-squirrel: @teoparvanov , in this way we'll loose possibly useful core dumps.
I'd prefer to discuss it within an open forum with the rest of the team.

.github/workflows/build_and_test.yml
```diff
@@ -47,12 +47,13 @@ jobs:
...
         - name: Prepare artifacts
           if: failure()
           run: |
             sudo chown -R $USER:$GROUP ${PWD}/build
             mv ${PWD}/build ${{ github.workspace }}/artifact/
+            cp /tmp/apollo ${{ github.workspace }}/artifact/
```
f-squirrel, line 56: `cp` copies a folder but not it's content, use: `cp -r ...`

Makefile
f-squirrel: @teoparvanov , IMO, it is better to turn on, so that a developer would have the same logs as in the CI.
If somebody does not need it, they may disable it explicitly.

tests/apollo/util/bft.py
f-squirrel: @teoparvanov , @oded282 , please notice that`make test` does not trigger either cmake generation or compilation. The reason is that tests are not supposed to change the pre-built product.
Please keep this behavior.

tests/apollo/util/bft.py
```diff
@@ -394,13 +397,27 @@ def start_replica(self, replica_id):
...
+                # Single Apollo module loaded, test name known.
+                test_name = f"{tests_names.pop()}_{storage_type}"
+
+            test_dir = f"/tmp/apollo/{test_name}/{self.current_test}/"
+            test_log = f"{test_dir}stdout_{replica_id}.log"
+
             if os.environ.get('KEEP_APOLLO_LOGS', "").lower() in ["true", "on"]:
                 try:
-                    os.mkdir(f"/tmp/apollo/{self.current_test}/")
+                    os.makedirs(test_dir)
```
f-squirrel, line 415: IMO, the exception thing is redundant here.
We can just use: `os.makedirs(test_dir, exist_ok=True)`

tests/apollo/util/eliot_logging.py
```diff
@@ -16,18 +15,24 @@ def set_file_destination():
         # Single Apollo module loaded, test name known.
         test_name = f"{tests_names.pop()}_{storage_type}"
 
-    # Create logs directory if not exist
-    if not os.path.isdir("logs"):
-        os.mkdir("logs")
+    logs_dir = '/tmp/apollo/'
```
f-squirrel, line 18: `/tmp/apollo` is not mounted to the docker image. So it won't be available for the host(GitHub Action runner).
In general, I'd prefer all the logs to be stored in the `build` directory:
* it is ignored by git
* it is a part of the project' directory and can be naturally searched for logs
* by writing to the build we do not pollute the system(when built "natively")
* tests do not rely on system permissions
* `build` is actually mounted to the docker

f-squirrel (changes requested):

f-squirrel (commented):
@oded282 ,
I see that you pass `KEEP_LOGS` to make and retrieve it in Apollo but I could not find the place where you actually provide the tests with the env variable.
As far as I see, you need to check in the `CMakelists.txt`if the flag is set and if so, set environment variable in a way similar to `STORAGE_TYPE`.

tests/apollo/util/bft.py
oded282: ok

tests/apollo/util/bft.py
```diff
@@ -138,7 +139,9 @@ async def wrapper(*args, **kwargs):
                                         stop_replica_cmd=None,
                                         num_ro_replicas=num_ro_replicas)
                     with BftTestNetwork.new(config) as bft_network:
-                        bft_network.current_test = async_fn.__name__ + "_n=" + str(bft_config['n']) \
+                        storage_type = os.environ.get("STORAGE_TYPE")
+                        bft_network.current_test = async_fn.__name__ + "_" + storage_type \
```
teoparvanov, line 143: can't we use the TEST_NAME variable here?

oded282: In case someone runs it locally and won't provide the storage name properly or won't provide it at all, extracting the storage type out of the TEST_NAME may produce strange log files names.

teoparvanov: AFAIK, you can't even run the tests locally without providing STORAGE_TYPE. Have you tried?

tests/apollo/util/eliot_logging.py
teoparvanov: Oded, please consider Dima's comment here, with respect to the failure we saw in GitHub Actions:
https://github.com/vmware/concord-bft/pull/900/checks?check_run_id=1181065805

Wondering why this only fails in one of the jobs, and not all of them though ...

tests/apollo/util/bft.py
oded282: You can run a single test with TEST_NAME equals any name, Therefore extracting the storage type from the test name will cause us problems.

teoparvanov: Can't you just say:
```python
bft_network.current_test = TEST_NAME
```

tests/apollo/util/eliot_logging.py
```diff
@@ -5,29 +5,31 @@
 
 
 def set_file_destination():
-    storage_type = os.environ.get('STORAGE_TYPE')
-    tests_names = [m for m in sys.modules.keys() if "test_" in m]
+    test_name = os.environ.get('TEST_NAME')
 
-    if len(tests_names) > 1:
+    if not test_name:
         # Multiple Apollo tests modules loaded, test name unknown.
```
teoparvanov, line 11: nit - the comment is not relevant anymore

tests/apollo/util/bft.py
teoparvanov: OK, agreed.

teoparvanov:
@oded282 can you please rerun CI with your latest changes? Thanks!

tests/apollo/util/eliot_logging.py
oded282: Sure.

teoparvanov (approved):
LGTM, thanks @oded282 !

Once you merge, let's use this for debugging the occasional GitHub Actions test failures. I suggest you also show how this can be done in your demo.

PR #901, opened 2020-09-24
Add an abstract channel type and bounded mpsc impl

andrewjstone:
In order to allow separation of functionality into various threads, and
allow message passing across those threads, an abstract channel
interface was created in channel.h. The rationale for the design is
given in comments in channel.h. Please read that. It's important to call
out here that the design is reflective of expectations of communication
patterns among our components, such that we must deal with cyclic,
transitive flows, and cannot allow any sender to block arbitrarily.

A lock based, bounded MPSC channel implementation was created in
mpsc_bounded_channel.h. This is useful for an initial implemnentation,
so that we can get started, but also useful to serve as an example of
how to implement the interface defined in the NullChannel in channel.h.

It's expected that the interface can be implemented using any queue,
including non-blocking ones such as boost::lockfree::spsc_queue.

A test is provided that shows how to use the BoundedMpscSender and
BoundedMpscReceiver types that compose a BoundedMpscChannel.

### DIFF
--- a/util/include/channels/channel.h
+++ b/util/include/channels/channel.h
@@ -26,6 +26,9 @@
 // multiple senders. With seperate sender and receiver types we make it so that we can make the sender copyable, but the
 // receiver unique.
 //
+// We distinguish between bounded and unbounded channels, since each is appropriate in different scenarios,
+// and the interfaces for the senders should be slightly different.
+//
 // Since a channel is based on static polymorphism, and concepts aren't supported in C++ 17, we
 // define the interfaces below with an example "NullChannel". All channels should support this
 // interface. The rationale for static polymorphism is three fold:
@@ -51,9 +54,9 @@ class NoReceiversError : public std::runtime_error {
   NoReceiversError() : std::runtime_error("All receivers have been destructed.") {}
 };
 
-// The interface for the sender side of a channel
+// The interface for the sender side of a bounded channel
 template <typename Msg>
-class NullSender {
+class BoundedNullSender {
   // Send a message over a channel.
   //
   // Sending a message on a channel must be non-blocking to prevent distributed deadlocks as a
@@ -77,9 +80,31 @@ class NullSender {
   // value will not be available at all. In those cases std::nullopt should be returned.
   std::optional<size_t> size() const { return std::nullopt; }
 
-  // Bounded channels should return how many messages they can store in their internal buffer.
-  // Unbounded channels should return std::nullopt;
-  std::optional<size_t> capacity() const { return std::nullopt; }
+  // Return how many messages can be store in the internal buffer.
+  size_t capacity() const { return 0; }
+};
+
+// The interface for the sender side of an unbounded channel
+template <typename Msg>
+class UnboundedNullSender {
+  // Send a message over a channel.
+  //
+  // Sending a message on a channel succeeds excepted in the following cases:
+  //   * There are no receivers - In this case a NoReceiversError will be thrown.
+  //   * Memory allocation fails - Only use unbounded channels when you don't expect this to be an issue.
+  //
+  // Throws NoReceiversError if all receivers have been destructed.
+  //
+  // It's important to note that there is a race condition, whereby a receiver can be destructed
+  // immediately after a successful send. There is no absolute guarantee of delivery. The
+  // purpose of returning an error is to prevent endlessly retrying and never noticing that the
+  // channel is effectively destroyed.
+  void send(Msg&& msg) {}
+
+  // Return how many elements are in the channel's internal buffer. For some channels, only an
+  // estimate will be avaialable, and for some others (mainly lock-free/wait-free) channels, this
+  // value will not be available at all. In those cases std::nullopt should be returned.
+  std::optional<size_t> size() const { return std::nullopt; }
 };
 
 // The interface for the receiver side of a channel
--- a/util/include/channels/channel.h
+++ b/util/include/channels/channel.h
@@ -72,7 +72,7 @@ class NullSender {
   // channel is effectively destroyed.
   std::optional<Msg> send(Msg&& msg) { return std::nullopt; }
 
-  // Return how many elements are in the channels internal buffer. For some channels, only an
+  // Return how many elements are in the channel's internal buffer. For some channels, only an
   // estimate will be avaialable, and for some others (mainly lock-free/wait-free) channels, this
   // value will not be available at all. In those cases std::nullopt should be returned.
   std::optional<size_t> size() const { return std::nullopt; }
@@ -101,14 +101,14 @@ class NullReceiver {
   // Return std::nullopt if a timeout occurred.
   std::optional<Msg> recv(std::chrono::milliseconds timeout) { return std::nullopt; };
 
-  // Return how many elements are in the channels internal buffer. For some channels, only an
+  // Return how many elements are in the channel's internal buffer. For some channels, only an
   // estimate will be avaialable, and for some others (mainly lock-free/wait-free) channels, this
   // value will not be available at all. In those cases std::nullopt should be returned.
-  std::optional<size_t> size() { return std::nullopt; }
+  std::optional<size_t> size() const { return std::nullopt; }
 
   // Bounded channels should return how many messages they can store in their internal buffer.
   // Unbounded channels should return std::nullopt;
-  std::optional<size_t> capacity() { return std::nullopt; }
+  std::optional<size_t> capacity() const { return std::nullopt; }
 };
 
 }  // namespace concord::channel
--- a/util/include/channels/channel.h
+++ b/util/include/channels/channel.h
@@ -1,0 +1,114 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the
+// LICENSE file.
+
+#pragma once
+#include <stdexcept>
+
+// A generic channel interface based around static polymorphism (templates).
+//
+// We expect to have multiple channel implementations for different purposes, but the sending and
+// receiving code should be agnostic to this fact. We want to be able to swap implementations based
+// on performance and/or capability.
+//
+// A channel is made up of a sender and receiver pair. This split prevents misuse of concrete
+// channels. For example, if we only returned an "IChannel" type, and the channel was an MPSC channel, we could
+// accidentally have two receivers for it, since we would be forced to allow cloning of the type so that we could have
+// multiple senders. With seperate sender and receiver types we make it so that we can make the sender copyable, but the
+// receiver unique.
+//
+// Since a channel is based on static polymorphism, and concepts aren't supported in C++ 17, we
+// define the interfaces below with an example "NullChannel". All channels should support this
+// interface. The rationale for static polymorphism is three fold:
+//    * Performance - avoiding virtual call overhead
+//    * Avoiding the need to expose the sender and receiver types as pointers or references
+//    * Dynamic polymorphism may not be particularly helpful, because we already must parameterize on a message type. It
+//      would require all channels to send the same type if stored in the same map, etc...
+//
+// These channels are also optimized for move operations and as such do not require Msg to be copyable, although in
+// almost all instances they will be. This means that if we implement this interface with something like
+// boost::lockfree::spsc_queue (which requires copying), and our messages are large, we will likely lose some speed.
+namespace concord::channel {
+
+// Thrown when receiver.recv() is called but there are no corresponding senders for the channel anymore.
+class NoSendersError : public std::runtime_error {
+ public:
+  NoSendersError() : std::runtime_error("All senders have been destructed.") {}
+};
+
+// Thrown when sender.send() is called but there are no corresponding receivers for the channel anymore.
+class NoReceiversError : public std::runtime_error {
+ public:
+  NoReceiversError() : std::runtime_error("All receivers have been destructed.") {}
+};
+
+// The interface for the sender side of a channel
+template <typename Msg>
+class NullSender {
+  // Send a message over a channel.
+  //
+  // Sending a message on a channel must be non-blocking to prevent distributed deadlocks as a
+  // result of cyclic topologies. To enable this, if the channel is full we return the message to
+  // the sender. The sender can then determine whether to drop it or buffer it and apply
+  // backpressure.
+  //
+  // Returns std::nullopt when the message is succesfully sent.
+  // Returns Msg when the channel is full.
+  //
+  // Throws NoReceiversError if all receivers have been destructed.
+  //
+  // It's important to note that there is a race condition, whereby a receiver can be destructed
+  // immediately after a successful send. There is no absolute guarantee of delivery. The
+  // purpose of returning an error is to prevent endlessly retrying and never noticing that the
+  // channel is effectively destroyed.
+  std::optional<Msg> send(Msg&& msg) { return std::nullopt; }
+
+  // Return how many elements are in the channels internal buffer. For some channels, only an
+  // estimate will be avaialable, and for some others (mainly lock-free/wait-free) channels, this
+  // value will not be available at all. In those cases std::nullopt should be returned.
+  std::optional<size_t> size() const { return std::nullopt; }
+
+  // Bounded channels should return how many messages they can store in their internal buffer.
+  // Unbounded channels should return std::nullopt;
+  std::optional<size_t> capacity() const { return std::nullopt; }
+};
+
+// The interface for the receiver side of a channel
+template <typename Msg>
+class NullReceiver {
+  // Receive a message over a channel.
+  // Block waiting indefinitely.
+  //
+  // This example implies that Msg must be default constructable. In a real channel this is not
+  // necessary.
+  //
+  // Throws NoSendersError if all senders have been destructed and there are no more messages left
+  // in the communication buffer.
+  Msg recv() { return Msg(); }
+
+  // Receive a message over a channel.
+  // Wait only until the timeout.
+  //
+  // Return std::nullopt if a timeout occurred.
+  std::optional<Msg> recv(std::chrono::milliseconds timeout) { return std::nullopt; };
+
+  // Return how many elements are in the channels internal buffer. For some channels, only an
+  // estimate will be avaialable, and for some others (mainly lock-free/wait-free) channels, this
+  // value will not be available at all. In those cases std::nullopt should be returned.
+  std::optional<size_t> size() { return std::nullopt; }
+
+  // Bounded channels should return how many messages they can store in their internal buffer.
+  // Unbounded channels should return std::nullopt;
+  std::optional<size_t> capacity() { return std::nullopt; }
+};
+
+}  // namespace concord::channel
--- a/util/include/channels/mpsc_bounded_channel.h
+++ b/util/include/channels/mpsc_bounded_channel.h
@@ -13,7 +13,6 @@
 
 #pragma once
 
-#include <cassert>
 #include <chrono>
 #include <condition_variable>
 #include <memory>
@@ -21,6 +20,7 @@
 #include <queue>
 #include <utility>
 
+#include "assertUtils.hpp"
 #include "channel.h"
 
 namespace concord::channel {
@@ -52,7 +52,7 @@ class BoundedMpscQueue {
     return buf_.size();
   }
 
-  std::optional<size_t> capacity() const {
+  size_t capacity() const {
     // No need for a lock, as capacity_ is immutable
     return capacity_;
   }
@@ -121,7 +121,7 @@ class BoundedMpscSender {
   }
 
   std::optional<size_t> size() const { return queue_->size(); }
-  std::optional<size_t> capacity() const { return queue_->capacity(); }
+  size_t capacity() const { return queue_->capacity(); }
 
   ~BoundedMpscSender() {
     // We must check if a queue still exists because of move semantics. A move doesn't alter the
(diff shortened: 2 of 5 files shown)


util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,238 @@
...
+    const std::lock_guard<std::mutex> lock(mutex_);
+    valid_receiver_ = false;
+  }
+
+  size_t capacity_;
+  std::queue<Msg> buf_;
+  mutable std::mutex mutex_;
+  std::condition_variable cv_;
+  size_t num_senders_ = 0;
+  bool valid_receiver_ = true;
```
yuliasherman, line 87: Why not make all simple data members atomic?

andrewjstone: It doesn't actually work for mutation. You can end up with lost updates that way. With the use of a condition variable, they must all be protected by a mutex. I actually did have num_senders and valid_receiver as atomic, but ended up with lost notifications and a deadlock when I didn't take the lock to mutate. 

This explains it better than I can: https://www.modernescpp.com/index.php/c-core-guidelines-be-aware-of-the-traps-of-condition-variables

andrewjstone: For capacity, though, since it's fixed, we could avoid taking the lock altogether though I believe.

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,238 @@
...
+class BoundedMpscReceiver {
+ public:
+  // Receive a message over a channel.
+  // Block waiting indefinitely.
+  //
+  // Throws NoSendersError if all senders have been destructed and there are no more messages left
+  // in the communication buffer.
+  Msg recv() {
+    std::unique_lock<std::mutex> lock(queue_->mutex_);
+    if (!queue_->buf_.empty()) {
```
andrewjstone, line 157: Looks like these aren't needed since the predicate form is used and they are coverd in waitPred(). This was a misunderstanding on my part.

andrewjstone: Fixed.

util/include/channels/mpsc_bounded_channel.h
andrewjstone: I went ahead and removed the lock from `BoundedMpscQueue::capacity()`

util/include/channels/channel.h
```diff
@@ -0,0 +1,114 @@
...
+  // in the communication buffer.
+  Msg recv() { return Msg(); }
+
+  // Receive a message over a channel.
+  // Wait only until the timeout.
+  //
+  // Return std::nullopt if a timeout occurred.
+  std::optional<Msg> recv(std::chrono::milliseconds timeout) { return std::nullopt; };
+
+  // Return how many elements are in the channels internal buffer. For some channels, only an
```
dartdart26, line 104: Typo: channel's

util/include/channels/channel.h
```diff
@@ -0,0 +1,114 @@
...
+  // estimate will be avaialable, and for some others (mainly lock-free/wait-free) channels, this
+  // value will not be available at all. In those cases std::nullopt should be returned.
+  std::optional<size_t> size() { return std::nullopt; }
+
+  // Bounded channels should return how many messages they can store in their internal buffer.
+  // Unbounded channels should return std::nullopt;
+  std::optional<size_t> capacity() { return std::nullopt; }
```
dartdart26, line 111: Nit: Maybe we can make these `const` so that users can expect it for all channels.

util/include/channels/channel.h
```diff
@@ -0,0 +1,114 @@
...
+  // Returns std::nullopt when the message is succesfully sent.
+  // Returns Msg when the channel is full.
+  //
+  // Throws NoReceiversError if all receivers have been destructed.
+  //
+  // It's important to note that there is a race condition, whereby a receiver can be destructed
+  // immediately after a successful send. There is no absolute guarantee of delivery. The
+  // purpose of returning an error is to prevent endlessly retrying and never noticing that the
+  // channel is effectively destroyed.
+  std::optional<Msg> send(Msg&& msg) { return std::nullopt; }
```
dartdart26, line 73: Nice idea for the return type. I was going to suggest that we don't conflate the status return with the message itself, but after thinking about it I believe it is quite useful that way, especially with the move semantics. I guess it can be done in some other way in order to avoid the move if the channel is full, but I like the current interface more :)

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+  friend class ::concord::channel::BoundedMpscSender;
+  template <typename T>
+  friend class ::concord::channel::BoundedMpscReceiver;
+
+  std::optional<size_t> size() {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    return buf_.size();
+  }
+
+  std::optional<size_t> capacity() {
```
dartdart26, line 54: Nit: const?

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+  BoundedMpscQueue(size_t capacity) : capacity_(capacity) {}
+
+ private:
+  template <typename T>
+  friend class ::concord::channel::BoundedMpscSender;
+  template <typename T>
+  friend class ::concord::channel::BoundedMpscReceiver;
+
+  std::optional<size_t> size() {
+    const std::lock_guard<std::mutex> lock(mutex_);
```
dartdart26, line 50: Nit: Can we make the `mutex_` mutable and this method `const`?

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+    return capacity_;
+  }
+
+  void addSender() {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    num_senders_++;
+  }
+
+  void removeSender() {
+    bool num_senders = 0;
```
dartdart26, line 65: I would assume this should be of type `size_t`.

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+  BoundedMpscSender& operator=(BoundedMpscSender&&) noexcept = default;
+
+ private:
+  BoundedMpscSender(const std::shared_ptr<impl::BoundedMpscQueue<Msg>>& queue) : queue_(queue) { queue_->addSender(); }
+
+  std::shared_ptr<impl::BoundedMpscQueue<Msg>> queue_;
+
+  template <typename T>
+  friend std::pair<BoundedMpscSender<T>, BoundedMpscReceiver<T>> makeBoundedMpscChannel(size_t capacity);
+};  // namespace concord::channel
```
dartdart26, line 145: Nit: This shouldn't be the namespace bracket.

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+  bool waitPred() { return queue_->num_senders_ == 0 || !queue_->buf_.empty(); }
+
+  std::shared_ptr<impl::BoundedMpscQueue<Msg>> queue_;
+
+  template <typename T>
+  friend std::pair<BoundedMpscSender<T>, BoundedMpscReceiver<T>> makeBoundedMpscChannel(size_t capacity);
+};
+
+template <typename Msg>
+std::pair<BoundedMpscSender<Msg>, BoundedMpscReceiver<Msg>> makeBoundedMpscChannel(size_t capacity) {
```
dartdart26, line 220: So nice :)

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+  ~BoundedMpscSender() {
+    if (!queue_) return;
+    queue_->removeSender();
+  }
+
+  BoundedMpscSender(const BoundedMpscSender& other) : BoundedMpscSender(other.queue_) {}
+  BoundedMpscSender& operator=(const BoundedMpscSender& other) {
+    *this = BoundedMpscSender(other);
+    return *this;
+  }
```
dartdart26, line 134: If I am not mistaken, these can be made `= default`. Maybe even omit the `= default` move ones. Please correct me if I miss something.

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+    queue_->cv_.notify_one();
+
+    return std::nullopt;
+  }
+
+  std::optional<size_t> size() { return queue_->size(); }
+  std::optional<size_t> capacity() { return queue_->capacity(); }
+
+  ~BoundedMpscSender() {
+    if (!queue_) return;
```
dartdart26, line 126: Is that check really necessary? I am asking not because of performance, but just to be sure I don't miss something. I think if a non-null ptr is passed to the ctor (which is a precondition anyway), it will never be null.

dartdart26, line 126: If I am not mistaken, this can be null due to move semantics only. Can we maybe add a comment to be more explicit about it? It is a bit tricky.

util/include/channels/mpsc_bounded_channel.h
```diff
@@ -0,0 +1,226 @@
...
+  // We do allow moving a receiver though
+  BoundedMpscReceiver(BoundedMpscReceiver&&) noexcept = default;
+  BoundedMpscReceiver& operator=(BoundedMpscReceiver&&) noexcept = default;
+
+ private:
+  BoundedMpscReceiver(const std::shared_ptr<impl::BoundedMpscQueue<Msg>>& queue) : queue_(queue) {}
+
+  Msg pop() {
+    auto msg = std::move(queue_->buf_.front());
+    queue_->buf_.pop();
```
dartdart26, line 207: Nit: If we ever use a buffer type for which `pop()` can throw, then that would be exception-unsafe as we've already moved. Not sure if this is worth documenting.

dartdart26 (approved):
I think it looks great @andrewjstone ! I like how simple it is and how it enforces correct usage via the language. Thanks for doing it!

The only thing I think should be fixed is the `bool num_senders` variable.

util/include/channels/mpsc_bounded_channel.h
andrewjstone: You are correct! That is exactly the scenario I ran into. The `makeBoundedMpscChannel` function creates a pair with an rvalue for each of the sender and receiver. It then immediately moves them resulting in a null queue. It's very silly, and I don't know why the move can't be elided entirely in this case, similar to return value optimization. I will add a comment. Thank you for the feedback!

f-squirrel (approved):
Looks great to me!
I like the clear interface and after Petar's and Yulia's comments I have nothing to add:)

util/include/channels/channel.h
andrewjstone: I copied it from rust 😄 https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html

util/include/channels/mpsc_bounded_channel.h
andrewjstone: wow, good catch.

util/include/channels/mpsc_bounded_channel.h
andrewjstone: They can be default, but we must be explicit about them due to the custom destructor and rule of 5.

util/include/channels/mpsc_bounded_channel.h
andrewjstone: In this case, we use a std::deque, and pop cannot throw since we always call this on a non-empty deque. I will add a comment.

util/include/channels/mpsc_bounded_channel.h
andrewjstone: Nope, the default constructors do not work. The queue does not appear to get copied correctly in that case.

util/include/channels/channel.h
andrewjstone: So one difference that I note here from rust, is that in the case of a `NoReceiversError` we don't return the value. I could change the exception to contain the value, or change the call to return a more complex error type in the std::optional, similar to the rust version. 

This also got me thinking that regardless of the direction we go with this, that if you disregard the missing receiver case, there is a fundamental difference between bounded and unbounded channels. Unbounded channels can only fail when we run out of memory which is not a scenario worth considering. I'm thinking that maybe we should have two separate interfaces: One for bounded channels and one for unbounded channels, where unbounded channels are void if they throw an exception or std::optional if they fail because there is no receiver (or they reuse whatever error return type we create). 

@dartdart26  @f-squirrel What are your thoughts around returning the unsent value in the exception and/or separating bounded and unbounded channels into two separate interfaces?

dartdart26: In terms of interfaces, my opinion is that two separate interfaces is clearer - callers don't need to check the return value for unbounded channels.

I also have a question - if you don't have a receiver, what can you do with the message, even if you get it back? What I am getting at is that if you can't create a receiver for an existing queue, why bother with the message that was destined for that channel? Do I miss something?

Bottom line - I would leave the interface as is for bounded queues and make the method void for unbounded.

util/include/channels/mpsc_bounded_channel.h
dartdart26: {}-init syntax :)

util/include/channels/mpsc_bounded_channel.h
dartdart26: Oh, lord, the non-default dtor... You are right. That's extremely tricky, especially in this case. If you don't mention the copy and move ctors at all, you won't get moves if you have a non-default dtor. But if you make both copy and move `= default` it should have moves as in the existing code. Maybe it is still worth it to make the copy ops `= default` for simplicity. Please correct me if I am wrong.

util/include/channels/channel.h
andrewjstone: Thanks @dartdart26. I concur. I think the rationale for returning the value in case of error is if you have a single use channel. Then you can create a new one. It's kind of a weird use case we don't have. We can, and probably should use std::async for such things.

util/include/channels/mpsc_bounded_channel.h
andrewjstone: I tried making the copy ops default, but I got a NoSendersError. It appears they don't do what we expect.

dartdart26: What I mean is that the default copy and move for the single shared_ptr member should suffice as that's what the custom copy ctor is doing. If not, I must be missing something.

andrewjstone: Ah, I get it. Yes, just making the copy assignment operator default, but leaving the copy constructor implemented makes it all work..

andrewjstone: Fixed.

util/include/channels/channel.h
andrewjstone: I added an interface for an `UnboundedSender`. Notably, an UnboundedSender has a void return for the `send` function, but also does not have a `capacity` getter. 

Note, importantly, that both share the same receiver interface. There is no need for the user of a receiver to have to know whether a channel is bounded or not. Thus one interface can be shared and we get more polymorphism.

util/include/channels/mpsc_bounded_channel.h
dartdart26: I knew I was missing something - the copy ctor is calling the ctor taking the queue and, in effect, calling `addSender()`.  I didn't see that in the beginning. Sorry  for the noise and thanks for addressing it.

dartdart26: Having said above, if I don't miss something else (don't count on it :D), then `operator=()` should call `removeSender()` on the sender's current queue and then call `addSender()` on the new one.

Since the only way of creating a sender is to create a pair of a sender and a receiver for a specific channel, we might also consider disabling `operator=()`. I don't have a strong opinion on that, though.

toly-kournik:
@andrewjstone , from  your presentation I understood you wanted to introduce a general message passing interface which could be implemented in an in-process as well as an inter-process way, making it seamless for the interface users.
From this PR it's hard to see how it conforms to this vision.

Several notes:

**MPSC**
Basically there are two basic types of producer/consumer problems - SPSC and MPMC
Since you mentioned boost, it has implementations for only two of the above cases.
So don't think MPSC is a separate case, but rather a subset of MPMC. The desired MPSC behavior can be achieved on top of it.

**Static Polymorphism**
1. To my mind, virtual function call performance overhead is too much demonized, I wouldn't take it as a basic assumption.
Using  lockfree queues will contribute to overall performance much more than not using polymorphism with senders and receivers.
2. Don't see any problem exposing the sender/receiver types.
3. Of course,  (dynamic) polymorphism in this case is non-usable because of the taken approach, but what about the case with the current hierarchical BFT messages implementation? Would we have to use a channel per message type, or still pass `MessageBase` ?

**Bounded vs. Unbounded Sender API**
Having different APIs will require to decide at compile time which sender to use, and make it difficult to use a different approach. For instance we'd want to offload some processing to another process on the same or a different machine and this will require to switch to a bounded queue.
Instead, the common API with a timeout notion could be introduced to achieve the required behavior (zero timeout will return immediately).

**To Summarize**
I suggest to revise an _a priori_ taken decision about polymorphism overhead against the overall seamless distributed message passing vision, which I totally support.
Polymorphism will make it simple to replace underlying implementations of senders, receivers, queues, etc.
Combination matrix of types `[MPMC|SPSC|MPSC]/[bounded|unbounded]/[in-process|inter-process|distributed]` can be achieved in a more effective and brief way.

andrewjstone:
Thank you so much for the detailed feedback @toly-kournik. This is the type of counter argument needed to get this stuff right. 

My immediate goal here was concerned with C++ and I fully agree this isn't generic enough for cross language/IPC use cases. Your comment makes me think that we should go back to the drawing board... I'll respond to each piece here first though:

**MPSC** - You are probably right here. I thought for lockfree channels there was something that made MPSC faster and so it wasn't a subset, but that may not be true.

**Static Polymorphism*** - I agree that virtual function call overhead is not a big deal. I originally had the channel have a virtual interface but the message as a template, but realized that all channels would still need to take the same message to be able to put things into a container or make them truly polymorphic with the template message. I thought this may be unlikely, if we wanted to be more flexible on what we put into each channel. That flexibility on what to put into the channel thought (with a template message) came at the cost of the flexibility of storing channels in containers, which is a great loss. In short, I agree this is a problem that we can likely fix with dynamic polymorphism.

> Of course, (dynamic) polymorphism in this case is non-usable because of the taken approach, but what about the case with the current hierarchical BFT messages implementation? Would we have to use a channel per message type, or still pass MessageBase ?

We could have a single channel between components. I think we should always strive for that. We could do that with the current implementation or CMF.

**Bounded vs Unbounded** - In many cases, you really don't need to know whether you are dealing with a bounded or unbounded channel. However, in the same respect you can make some simplifying assumptions if you do know. Whether that's a useful tradeoff is up for discussion. My expectation is that inside the system, we will have unbounded channels, and at the edges, bounded ones. We can discuss this more offline. 

In summary, let's chat offline and get this right :)

dartdart26:
Great discussion, thanks for it!

I agree with most of what has been said. I just want to share my opinion on a few topics.

**Common Interfaces**
I think it is good to strive for common interfaces, but only when they are absolutely identical. In case of bounded vs unbounded, I think the interfaces are not identical (as mentioned in one of my previous comments). It might be a small difference now, but future changes may make it bigger and pollute the base interface if we were to use dynamic polymorphism.
Additionally, I completely agree that the performance overhead of virtual methods is not problematic. The only concern I have is that it forces us to have a common interface for things that might not share one.

**Policy-based Design**
I don't think we should use dynamic polymorphism **just** to implement a common interface. I think it makes sense to use dynamic polymorphism when we are going to change channel/sender/receiver type **at runtime**. I don't think we are planning to do something like that, at least for now. What we can do instead is to use policy-based design and combine different parts of the implementation and still have a common interface (or almost common). For example:

```
template <typename BoundPolicy, template TransportPolicy> 
class Sender {
 public:
  // Common (or semi-common) public API
  ...
 private:
  TransportPolicy transport_;
  BoundPolicy bound_;
};
```

The BoundPolicy may define how we handle a bounded queue and the TransportPolicy may define whether we use an in-memory queue, IPC or networking. Above is just a simple example and not a design. Doing something like that will allow us to have simpler and clearer interfaces that can evolve **separately**.

**Lock-Free**
I think it is not a given that a lock-free queue will be faster. Could actually be slower, we need to measure it.

Finally, I didn't fully get the problem with containers of channels. Could you please elaborate more?

andrewjstone:
Thanks for the great feedback @dartdart26. I agree that we shouldn't use dynamic polymorphism **just** for an identical interface. My point was just that with dynamic polymorphism, you can store many channels of different concrete types that conform to the same interface, whereas with static polymorphism, you can only store a single concrete channel type in a container.

As for policy based design, I really like it! I just had a good discussion with @ittaiab about DOS protection and fairness and we need to consider those in our channel design as well. The issue comes down to the fact that we may not want a strict FIFO removal policy on our channels. This is useful in other places as well, such as system status messages. If a bounded queue is full for example, we may still want to be able to receive system status messages.

We can implement these sophisticated queues in many ways:
1.  A single queue and a giant search selection, ala [selective receive in erlang](https://www.oreilly.com/library/view/programming-erlang-2nd/9781941222454/f_0107.html) 
2. Multiple queues with round robin selection [select in go](https://tour.golang.org/concurrency/5).
3. A single queue interface with multiple queues underneath that takes policies for selection.

Option three is likely the best for us, since it allows us to reuse existing SPSC or MPSC queues, and build policies on top. We need a  deeper discussion here.

andrewjstone:
Also, related to using a single queue interface with multiple queues underneath:
 This is similar to select, except we can build it with templates and tuples and remove the need to store channels in containers. This allows us to have different message types underneath without dynamic polymorphism.

dartdart26:
On the topic of runtime polymorphism, an interesting way to do it without inheritance is the Dyno library: https://github.com/ldionne/dyno

andrewjstone:
Closing this as YAGNI for now. We can revisit in the future. The branch with the code isn't going away.

PR #904, opened 2020-09-28
[Apollo] Fix pre-execution tests with constant load in the background

teoparvanov:
The proposed solution is to wait for the pre-processor's thread pool to become ready. The two following improvements are suggested for the BFT client:
* BFT client should not clean replies received so far, unless byzantine replies are detected (more than f non-matching replies). This mimics the behavior of the C++ SimpleClientImp.
* (trivial) we still have quorum, if we have received more than 2f+1 replies

### DIFF
--- a/tests/apollo/test_skvbc_preexecution.py
+++ b/tests/apollo/test_skvbc_preexecution.py
@@ -147,7 +147,6 @@ class SkvbcPreExecutionTest(unittest.TestCase):
                                                 err_msg="Make sure the view did not change.")
                 await trio.sleep(seconds=5)
 
-    @unittest.skip("Unstable")
     @with_trio
     @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
     @with_constant_load
@@ -157,6 +156,7 @@ class SkvbcPreExecutionTest(unittest.TestCase):
         concurrently with a constant system load in the background.
         """
         bft_network.start_all_replicas()
+        await trio.sleep(SKVBC_INIT_GRACE_TIME)
 
         write_set = [(skvbc.random_key(), skvbc.random_value()),
                      (skvbc.random_key(), skvbc.random_value())]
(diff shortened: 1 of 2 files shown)


tests/apollo/test_skvbc_preexecution.py
```diff
@@ -256,8 +255,9 @@ async def test_pre_execution_with_added_constant_load(self, bft_network, skvbc,
         This test validates that pre-execution and normal execution coexist correctly.
         """
         bft_network.start_all_replicas()
-        num_preexecution_requests = 200
+        await trio.sleep(SKVBC_INIT_GRACE_TIME)
```
andrewjstone, line 258: Is there a way to poll this instead? Maybe an an atomic variable or something, or a sample request can be sent.

I just really hate sleeps, and this looks like one that could cause flaky test runs.

teoparvanov: Well, not sure what exactly to poll for... 

This problem is specific to how the pre-processor is initialized.

I generally agree with your comment, but here the system is simply not up yet, so I think it makes sense to wait.

Another option would be to try running a pre-execution request until it succeeds, but it would create an undesired entry in the blockchain...

andrewjstone: You're waiting for the thread pool to start up right? Can you use some sort of out of band mechanism? If everything was in C++ you could poll on an atomic or use a condition variable. But for testing, maybe you just instead publish an environment variable or write a file to be read. There has to be a better way than just using an arbitrary sleep.

Now the question is whether what I mentioned is actually worth the effort. In most cases the answer is likely yes, but in this case maybe not. So if not, how long to do you make the timeout? And what happens if it's not long enough?

teoparvanov: A timeout of `SKVBC_INIT_GRACE_TIME=2` (seconds) seems to be sufficient (used in other pre-execution tests too).

teoparvanov:
@yuliasherman can you please have a look and approve if OK? Thanks!

yuliasherman (approved):
LGTM

PR #905, opened 2020-09-28
Exit on non-existent DB in sparse_merkle_db_editor

dartdart26:
If any of the passed RocksDB DBs to sparse_merkle_db_editor are
non-existent or empty, exit with an error instead of reporting empty
values. Prior to this change, the editor was first creating an empty DB
and then all commands were working by reporting default/empty values.

Since RocksDB doesn't offer a consistent method for checking if a DB
exists at a given path, first check if the passed directory exists. If
it doesn't, report an error. If it does, check if there are any keys in
the DB - if there are none, report an error. Above behavior can be
simplified when the following RocksDB issue is resolved:
https://github.com/facebook/rocksdb/issues/5029

Add tests for non-existent and empty databases for commands accepting
both one and two DB paths.

### DIFF
--- a/kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
+++ b/kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
@@ -37,6 +37,16 @@
 #include <variant>
 #include <vector>
 
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace fs = std::filesystem;
+#elif __has_include(<experimental/filesystem>)
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#else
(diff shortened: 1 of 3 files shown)


f-squirrel (approved):
LGTM!

kvbc/tools/sparse_merkle_db/include/sparse_merkle_db_editor.hpp
```diff
@@ -37,6 +37,16 @@
 #include <variant>
 #include <vector>
 
+#if __has_include(<filesystem>)
```
tdimitrov, line 40: Portability to the limit ;)

tdimitrov (approved):

andrewjstone (approved):
LGTM. Thanks for the fix @dartdart26

PR #907, opened 2020-09-29
Path prefix for S3 object store is set as configuration parameter

tdimitrov:
Path prefix is additional key (consider it something like a directory)
created in the target S3 bucket.
This patch changes its creation from random value to value specified in
the S3 configuration. The prefix can't be a random value because all
fetched state in S3 is lost on RO replica restart.

### DIFF
--- a/kvbc/test/s3_storage_factory_test.cpp
+++ b/kvbc/test/s3_storage_factory_test.cpp
@@ -1,0 +1,35 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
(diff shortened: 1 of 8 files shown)


kvbc/test/s3_storage_factory_test.cpp
```diff
@@ -0,0 +1,26 @@
+// Copyright 2019 VMware, all rights reserved
```
dartdart26, line 1: Nit: 2020

kvbc/test/s3_storage_factory_test.cpp
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Tests for S3StorageFactory
```
dartdart26, line 3: Nit: Current tests are not related to the factory itself - instead, focus is on key generation. Maybe we can change the test name or include it in the existing `ecs_s3_test`. Just a nit...

dartdart26 (approved):
Looks good, thanks @tdimitrov !

PR #909, opened 2020-09-29
Run CMF python codegen tests via CMAKE

andrewjstone:
This will trigger them in CI.

Running the C++ tests in CI will require some changes due to dependence
on g++.


robem (approved):
Nice change!

PR #911, opened 2020-09-30
Replace Apollo's print statements with Eliot's logs

oded282:
In this PR I replaced all Apollo's print statements with Eliot's log statements.
Added Eliot's `@log_call` decorator to `def interesting_configs(selected=None)` - prints the configurations under which the test will run. 

Decorator example:
image

*Eliot's logs total size is 70MB.
56MB comes from one test 'test_skvbc_persistence.py' - can be decreased significantly by removing the action from `async def is_fetching(self, replica_id)`

*Removing the action mentioned above, reduced 24MB from Eliot's logs file.

### DIFF
--- a/tests/apollo/test_skvbc_backup_restore.py
+++ b/tests/apollo/test_skvbc_backup_restore.py
@@ -17,6 +17,7 @@ import trio
 
 from util import skvbc as kvbc
 from util.bft import with_trio, with_bft_network, with_constant_load, KEY_FILE_PREFIX
+from util import eliot_logging as log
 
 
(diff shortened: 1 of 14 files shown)


f-squirrel (approved):
LGTM!
Is it a final decision not to print to stdout anymore in Apollo(+ @teoparvanov)?

teoparvanov:
@oded282 as you suggested, let's do the following:
_"56MB comes from one test 'test_skvbc_persistence.py' - can be decreased significantly by removing the action from async def is_fetching(self, replica_id)"_

teoparvanov (approved):
LGTM!

andrewjstone:
Woooo! Awesome work @oded282

PR #912, opened 2020-09-30
Fix bug in reconfiguration

yontyon:
Once we stop processing new requests (for example, due to wedge command) the following scenario may happen:
1. n-f replicas are getting to a stable checkpoint and advance the mainLog (https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L2460)
2. The f remaining replicas are trying to get updated by sending reqMissingInfo messages
3. Yet, the first n-f replicas reject their messages due to the following condition: https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L2649
4. Following No. 3, the f late replicas can only get updated using state transfer, but due to the fact that the system does not produce new checkpoints they cannot get to the following conditions: https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L1537
5. The late f replicas never get to a n/n checkpoint and the test fails
Our solution is to enable re-adding stable checkpoints to the checkpoint table https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L1496. This should not harm performance (in the worst case we will have a dozen of replicas, and lopping over the table won't make any effect)

From reading the logs, it seems that with this change, the late f replicas eventually initiate the state transfer (a thing that did not happen without this change) end eventually they reach to a n/n checkpoint

EDIT:
After discussing with @guyg8 I have removed the retransmitting solution #800 and use the existing status messages mechanism.

To test the stability of reconfiguration I've run some the reconfiguration test for hundreds of times on my local machine without any failure.
In addition, I have triggered the CI tests several times and it passed in all of them.

Note:
We won't merge this change before OCT cut release.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1500,9 +1500,7 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
       if (pos != tableOfStableCheckpoints.end()) delete pos->second;
       CheckpointMsg *x = new CheckpointMsg(msgSenderId, msgSeqNum, msgDigest, msgIsStable);
       tableOfStableCheckpoints[msgSenderId] = x;
-      LOG_INFO(GL,
-               "Added stable Checkpoint message to tableOfStableCheckpoints: " << KVLOG(
-                   msgSenderId, tableOfStableCheckpoints.size()));
+      LOG_INFO(GL, "Added stable Checkpoint message to tableOfStableCheckpoints: " << KVLOG(msgSenderId));
 
       if ((uint16_t)tableOfStableCheckpoints.size() >= config_.fVal + 1) {
         uint16_t numRelevant = 0;
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1494,13 +1494,15 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
   if (msgIsStable && msgSeqNum > lastExecutedSeqNum) {
     auto pos = tableOfStableCheckpoints.find(msgSenderId);
     if (pos == tableOfStableCheckpoints.end() || pos->second->seqNumber() < msgSeqNum ||
-        (pos->second->seqNumber() == msgSeqNum
-         && mainLog->insideActiveWindow(lastExecutedSeqNum)
-         && (getMonotonicTime() - mainLog->get(lastExecutedSeqNum).lastUpdateTimeOfCommitMsgs() > milliseconds(timeToWaitBeforeStartingStateTransferInMainWindowMilli)))) {
+        (pos->second->seqNumber() == msgSeqNum && mainLog->insideActiveWindow(lastExecutedSeqNum) &&
+         (getMonotonicTime() - mainLog->get(lastExecutedSeqNum).lastUpdateTimeOfCommitMsgs() >
+          milliseconds(timeToWaitBeforeStartingStateTransferInMainWindowMilli)))) {
       if (pos != tableOfStableCheckpoints.end()) delete pos->second;
       CheckpointMsg *x = new CheckpointMsg(msgSenderId, msgSeqNum, msgDigest, msgIsStable);
       tableOfStableCheckpoints[msgSenderId] = x;
-      LOG_INFO(GL, "Added stable Checkpoint message to tableOfStableCheckpoints: " << KVLOG(msgSenderId, tableOfStableCheckpoints.size()));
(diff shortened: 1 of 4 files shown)


guyg8:
> 4\. Following No. 3, the f late replicas can only get updated using state transfer, but due to the fact that the system does not produce new checkpoints they cannot get to the following conditions: https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L1537

If I remember correctly, according the original design the "late" f replicas should receive the checkpoints by sending ReplicaStatusMsg messages to the other n-f replicas. See https://github.com/vmware/concord-bft/blob/1e2bc080057f7a348375c28d91bb9db51d27c17b/bftengine/src/bftengine/ReplicaImp.cpp#L1735-L1770

yontyon:
> https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L1537

It is true that this code eventually gives the late replica the desired checkpoint.
Yet, due to the "<" (instead of "<=") in https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L1495 and the fact that we don't produce new checkpoint, the late replica will never invoke state transfer and won't be able to get to n/n checkpoint.
@guyg8, Am I missing something?

guyg8:
Yes. I understand that you want to support “reconfiguration” that is based on collaboration of all replicas. I guess it can be implemented in a few different ways.
I just wanted to say that if you use the existing checkpoint messages, then you should probably use the retransmission mechanism in ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg). In the current codebase, this mechanism is used to (re-)send checkpoints to the “late” replicas.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1763,6 +1768,7 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
       for (SeqNum i = beginRange; i <= endRange; i = i + checkpointWindowSize) {
         CheckpointMsg *checkMsg = checkpointsLog->get(i).selfCheckpointMsg();
         if (checkMsg != nullptr) {
+          LOG_INFO(GL, " *** sends checkpoint *** " << KVLOG(i, msgSenderId));
```
andrewjstone, line 1771: Presumably, you'll remove this before merge, or make it a debug message.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2646,8 +2650,7 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
   const ReplicaId msgSender = msg->senderId();
   SCOPED_MDC_SEQ_NUM(std::to_string(msgSeqNum));
   LOG_INFO(GL, "Received ReqMissingDataMsg. " << KVLOG(msgSender, msg->getFlags()));
-  if ((currentViewIsActive()) && (msgSeqNum > strictLowerBoundOfSeqNums) && (mainLog->insideActiveWindow(msgSeqNum)) &&
-      (mainLog->insideActiveWindow(msgSeqNum))) {
+  if ((currentViewIsActive()) && (msgSeqNum > strictLowerBoundOfSeqNums) && (mainLog->insideActiveWindow(msgSeqNum))) {
```
andrewjstone, line 2653: After a wedge, no client commands are every processed, right?

Is it ever possible that a view becomes inactive somehow after a wedge, and therefore this condition doesn't hold. 

I think a way to test it would be to partition a replica at the right time, send some client requests, wait for view change timeout, then see if you can rejoin the replica and reach N/N. Maybe this is overkill though.

andrewjstone (commented):
@yontyon This looks great. It's much simpler than the prior solution. As long as the view can never become inactive it should be safe to merge.

Thank you very much @guyg8 for the helpful feedback. The code is much better with your helpful suggestions!

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: I don't think it is possible, but anyway, even if it is, it doesn't matter. The system is wedged only w.r.t running consensus on new client request.
Vire change still can happen on the system

teoparvanov (approved):
LGTM, thanks @yontyon !

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: Ah ok. Cool. Thank you!

andrewjstone (approved):

PR #913, opened 2020-09-30
WIP : initial key rotation

cloudnoize:
- Initial key exchange generates a key pair using the crypto manager.

 Private key rotation - A private key has three states:
- published to consensus.
-  after consensus i.e. outstanding
-  After all replicas exchange keys on initial key-exchange the key is promoted to the private key of the replica

File interface for saving and loading the current key view of a replica, key view contains all replicas public keys and all states of the private keys.

Handling failing to verify a request when starting from the replica loader and rebuilding the window.
The handling is lazy i.e. upon failure we try to build the next crypto system from a key view and to verify with it.

Adding to Apolo an initial key-exchange method that performs the initial key exchange and can be added to tests as the first action to perform in order to validate its correctness with an initial key exchange


tests/apollo/test_skvbc.py
```diff
@@ -57,13 +58,16 @@ async def test_state_transfer(self, bft_network):
         cluster with f=1 we should be able to stop a different node after state
         transfer completes and still operate correctly.
         """
+        await bft_network.do_key_exchange()
```
teoparvanov, line 61: shouldn't the replicas be started before we do the key exchange?

tests/apollo/test_skvbc_view_change.py
```diff
@@ -59,6 +60,8 @@ async def test_request_block_not_written_primary_down(self, bft_network):
         4) Verify the BFT network eventually transitions to the next view.
         5) Validate that there is no new block written.
         """
+        bft_network.do_key_exchange()
```
teoparvanov, line 63: `await  bft_network.do_key_exchange()`, otherwise it's not executed

tests/apollo/test_skvbc.py
cloudnoize: This method starts the replicas, performs key exchange, and stops the replicas.
The idea is to add this method as an initial step to an arbitrary test. Then the test starts as it used to start, for example, if a test needs to start with a partial set of the replicas, this method won't impose any changes.

The test can test its functionality with the insurance that it uses an updated start-up sequence but without any changes to the test itself.

Of course, it does create additional side effects that can't be avoided, like starting from a last executed sequence number which is greater than 0.

tests/apollo/test_skvbc_view_change.py
cloudnoize: thanks

tests/apollo/test_skvbc.py
teoparvanov: I see, thanks Eran. I would then suggest the following improvement:
* add an optional parameter to the decorator `def with_bft_network(await bft_network.do_key_exchange(), ..., rotate_keys=False)`
* move the call to `await bft_network.do_key_exchange()` in the `with_bft_network` decorator. Only run `if rotate_keys`.
* update the decorator for tests where you want to do key rotation (e.g. `with_bft_network(start_replica_cmd, rotate_keys=True)`)

tests/apollo/util/bft.py
```diff
@@ -898,3 +899,28 @@ async def num_of_slow_path(self):
                         except KeyError:
                             # metrics not yet available, continue looping
                             pass
+    
+    async def do_key_exchange(self):
+        self.start_all_replicas()
+        with trio.fail_after(seconds=20):
```
teoparvanov, line 905: as demonstrated by @oded282, let's wrap this as an Eliot action for better logging

tests/apollo/test_skvbc.py
cloudnoize: It's a great improvement, I hope I'll be able to do it for this PR, since I'm on a very tight schedule.
If not I'll open a ticket and implement it after this.

cloudnoize: you wrote ```def with_bft_network(await bft_network.do_key_exchange(), ..., rotate_keys=False)```
do I need the ```await bft_network.do_key_exchange()``` as a param ?

teoparvanov: Oops, of course not, it was a copy/paste glitch ;) just add the `rotate_keys` param. And sure, this improvement can be in a separate PR!

toly-kournik (approved):
LGTM, thanks for you dedication and hard work!

PR #915, opened 2020-10-02
Diagnostics histograms

andrewjstone:
Support for histograms was added to the diagnostics library. Histograms are based on the [C Implementation](https://github.com/HdrHistogram/HdrHistogram_c/) of [HdrHistogram](http://hdrhistogram.org/).

This PR consists of 2 logical commits. The first commit is a refactoring of the diagnostics status code. The second is support for histograms. Please read and review the commits independently.

A follow up commit, in a separate PR, will register components and histograms for the concord-bft codebase.

As a side note, I just noticed that there are no units in the histogram output. Each histogram registration should likely include units. I will add that. I just wanted to get this out ASAP for review.

### DIFF
--- a/diagnostics/include/performance_handler.h
+++ b/diagnostics/include/performance_handler.h
@@ -1,0 +1,245 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+#include <map>
+#include <memory>
+#include <mutex>
+#include <ostream>
+#include <stdexcept>
+
+#include <hdr/hdr_interval_recorder.h>
+
+#include "assertUtils.hpp"
+#include "Logger.hpp"
+
+namespace concord::diagnostics {
+
+static logging::Logger DIAG_LOGGER = logging::getLogger("concord.diagnostics");
+
+enum class Unit {
+  NANOSECONDS,
+  MICROSECONDS,
+  MILLISECONDS,
+  SECONDS,
+  MINUTES,
+
+  BYTES,
+  KB,
+  MB,
+  GB,
+
+  // Things like queue length, size of a map, etc..
+  COUNT
+};
+
+// A recorder is a thread-safe type that should always be created in a shared pointer to ensure
+// proper destruction. The recorder is used to add values to the histogram in the recording thread.
+// Interval histograms can be extracted in other threads.
+struct Recorder {
+  // These values, (except for unit), come directly from hdr histogram.
+  // https://github.com/HdrHistogram/HdrHistogram_c/blob/master/src/hdr_interval_recorder.h#L28-L32
+  Recorder(int64_t lowest_trackable_value, int64_t highest_trackable_value, int significant_figures, Unit unit)
+      : unit(unit) {
+    ConcordAssert(lowest_trackable_value > 0);
+    auto rv =
+        hdr_interval_recorder_init_all(&recorder, lowest_trackable_value, highest_trackable_value, significant_figures);
+    ConcordAssertEQ(0, rv);
+  }
+
+  ~Recorder() { hdr_interval_recorder_destroy(&recorder); }
+  Recorder(const Recorder&) = delete;
+  Recorder& operator=(const Recorder&) = delete;
+
+  void record(int64_t val) {
+    if (!hdr_interval_recorder_record_value(&(recorder), val)) {
+      // We don't track the name in recorder, which we would do just for this, which is almost impossible to hit.
+      LOG_WARN(DIAG_LOGGER, "Failed to record value: " << KVLOG(val, unit));
+    }
+  }
+
+  hdr_interval_recorder recorder;
+  Unit unit;
+};
+
+struct Histogram {
+  Histogram(const std::shared_ptr<Recorder>& recorder)
+      : recorder(recorder), start(std::chrono::system_clock::now()), snapshot_start(start), snapshot_end(start) {
+    snapshot = hdr_interval_recorder_sample_and_recycle(&(recorder->recorder), snapshot);
+    auto rv = hdr_init(
+        snapshot->lowest_trackable_value, snapshot->highest_trackable_value, snapshot->significant_figures, &history);
+    ConcordAssertEQ(0, rv);
+  }
+
+  ~Histogram() {
+    hdr_close(snapshot);
+    hdr_close(history);
+    snapshot = nullptr;
+    history = nullptr;
+  }
+
+  Histogram(const Histogram&) = delete;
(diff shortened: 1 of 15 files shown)


andrewjstone:
Need to add some more error checking asserts around hdr_XXX function return values.

andrewjstone:
There seems to be a bug (although not necessarily a corruption of the history) that zeroes out the snapshot when the same component is passed twice to `snapshot`.

```
root@f8d066d67825:/concord-bft/diagnostics# ./concord-ctl perf snapshot component1 component1

root@f8d066d67825:/concord-bft/diagnostics# ./concord-ctl perf get component1 histogram1
Start time (UTC): 2020-10-02 05:08:09
Snapshot Start time (UTC): 2020-10-02 05:48:12
Snapshot End time (UTC): 2020-10-02 05:48:12

History
--------------------
Memory Used: 163944
Count: 47094
Max: 1000447
Min: 32
Standard Deviation: 289937
Percentiles: 
  10: 100671
  25: 248319
  50: 500479
  75: 754175
  90: 904703
  95: 951295
  99: 990719
  99.9: 999423
  99.99: 999935
  99.999: 1000447

Snapshot
--------------------
Memory Used: 163944
Count: 0
Max: 0
Min: 0
Standard Deviation: 0
Percentiles: 
  10: 0
  25: 0
  50: 0
  75: 0
  90: 0
  95: 0
  99: 0
  99.9: 0
  99.99: 0
  99.999: 0
```

diagnostics/include/performance_handler.h
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  Histogram(const Histogram&) = delete;
+  Histogram& operator=(const Histogram&) = delete;
+
+  void takeSnapshot() {
+    snapshot_start = snapshot_end;
+    snapshot_end = std::chrono::system_clock::now();
+    // Add the previous snapshot to the history
+    hdr_add(history, snapshot);
```
andrewjstone, line 69: Since the histograms are the exact same size/shape this should never return discarded entries (return value > 0). I don't think it warrants an assert, as we don't want to crash. But it may be worth logging.

diagnostics/include/performance_handler.h
```diff
@@ -0,0 +1,201 @@
...
+};
+
+class PerformanceHandler {
+ public:
+  void registerComponent(const std::string& name, const Recorders&);
+
+  // List all components
+  std::string list() const;
+
+  // List all metrics for a given component
```
andrewjstone, line 176: s/metrics/histograms

diagnostics/src/status_handlers.cpp
```diff
@@ -0,0 +1,63 @@
...
+  }
+  status_handlers_.insert({handler.name, handler});
+}
+
+std::string StatusHandlers::get(const std::string& name) const {
+  std::lock_guard<std::mutex> guard(mutex_);
+  if (status_handlers_.count(name)) {
+    return status_handlers_.at(name).f();
+  }
+  return "*--STATUS_NOT_FOUND--*";
```
tdimitrov, line 31: nit: Maybe add this as static const std::string STATUS_NOT_FOUND {"*--STATUS_NOT_FOUND--*"} to avoid copy-pasting the text all around the code.

diagnostics/src/status_handlers.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+std::string StatusHandlers::describe(const std::string& name) const {
+  std::lock_guard<std::mutex> guard(mutex_);
+  if (status_handlers_.count(name)) {
+    return status_handlers_.at(name).description;
+  }
+  return "*--DESCRIPTION_NOT_FOUND--*";
```
tdimitrov, line 39: nit: Another const

tdimitrov (approved):

andrewjstone:
> There seems to be a bug (although not necessarily a corruption of the history) that zeroes out the snapshot when the same component is passed twice to `snapshot`.
> 
> ```
> root@f8d066d67825:/concord-bft/diagnostics# ./concord-ctl perf snapshot component1 component1
> 
> root@f8d066d67825:/concord-bft/diagnostics# ./concord-ctl perf get component1 histogram1
> Start time (UTC): 2020-10-02 05:08:09
> Snapshot Start time (UTC): 2020-10-02 05:48:12
> Snapshot End time (UTC): 2020-10-02 05:48:12
> 
> History
> --------------------
> Memory Used: 163944
> Count: 47094
> Max: 1000447
> Min: 32
> Standard Deviation: 289937
> Percentiles: 
>   10: 100671
>   25: 248319
>   50: 500479
>   75: 754175
>   90: 904703
>   95: 951295
>   99: 990719
>   99.9: 999423
>   99.99: 999935
>   99.999: 1000447
> 
> Snapshot
> --------------------
> Memory Used: 163944
> Count: 0
> Max: 0
> Min: 0
> Standard Deviation: 0
> Percentiles: 
>   10: 0
>   25: 0
>   50: 0
>   75: 0
>   90: 0
>   95: 0
>   99: 0
>   99.9: 0
>   99.99: 0
>   99.999: 0
> ```

This is not actually a bug. It's correct behavior. Two snapshots get taken without any samples in between them. So the last snapshot has nothing in it.

andrewjstone:
> > There seems to be a bug (although not necessarily a corruption of the history) that zeroes out the snapshot when the same component is passed twice to `snapshot`.
> > ```
> > root@f8d066d67825:/concord-bft/diagnostics# ./concord-ctl perf snapshot component1 component1
> > 
> > root@f8d066d67825:/concord-bft/diagnostics# ./concord-ctl perf get component1 histogram1
> > Start time (UTC): 2020-10-02 05:08:09
> > Snapshot Start time (UTC): 2020-10-02 05:48:12
> > Snapshot End time (UTC): 2020-10-02 05:48:12
> > 
> > History
> > --------------------
> > Memory Used: 163944
> > Count: 47094
> > Max: 1000447
> > Min: 32
> > Standard Deviation: 289937
> > Percentiles: 
> >   10: 100671
> >   25: 248319
> >   50: 500479
> >   75: 754175
> >   90: 904703
> >   95: 951295
> >   99: 990719
> >   99.9: 999423
> >   99.99: 999935
> >   99.999: 1000447
> > 
> > Snapshot
> > --------------------
> > Memory Used: 163944
> > Count: 0
> > Max: 0
> > Min: 0
> > Standard Deviation: 0
> > Percentiles: 
> >   10: 0
> >   25: 0
> >   50: 0
> >   75: 0
> >   90: 0
> >   95: 0
> >   99: 0
> >   99.9: 0
> >   99.99: 0
> >   99.999: 0
> > ```
> 
> This is not actually a bug. It's correct behavior. Two snapshots get taken without any samples in between them. So the last snapshot has nothing in it. It is counter intuitive though, and accidentally doing this can make a user lose a snapshot they cared about. I will make it impossible to pass the same component twice.

Fixed.

diagnostics/include/performance_handler.h
andrewjstone: Logs added.

diagnostics/src/status_handlers.cpp
andrewjstone: These were all just moved to a different file. I prefer to postpone this change.

diagnostics/src/status_handlers.cpp
andrewjstone: Same as above.

PR #916, opened 2020-10-02
Add implementation for each subclass of MessageBase to be created from MessageBase.

HristoStaykov:
Fixes BC-4800.
This is needed during deserialization (constructing the messages from the
raw char array which is holding a packed struct). In this case we incorporate a larger
msgBody_ into a MessageBase object, having the header required for MessageBase in the beginning.
We need to later create the concrete object in order to call the accessor methods and
the virtual validate method for the complete msgBody_ (holding the header of the MessageBase
followed by the header of the concrete message followed by the payload).


f-squirrel (approved):
LGTM,
Thank you @HristoStaykov!
I suggest updating the PreProcessor and StateTransfer message handlers as well.

toly-kournik (approved):
LGTM, thanks Hristo

PR #917, opened 2020-10-02
Proper handling for empty prefix in S3KeyGenerator

tdimitrov:
The key for the object store can't contain two slashes. However if empty
prefix is set, the generated key is something like '//1/rawblock'.

This patch adds proper handling for empty prefix.


teoparvanov (approved):
LGTM!

PR #918, opened 2020-10-04
Change location for fake communication

Liorgal28:
In this PR I changed the location of the fake communication in order to use it in another place.


yontyon (approved):

andrewjstone (commented):
I really would prefer if this was *not* in the production include directory. It is test only code. Can we please move this to the test directory?

PR #920, opened 2020-10-05
Save key view to file

cloudnoize:


Liorgal28 (approved):
LGTM!

bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -294,12 +297,32 @@ bool KeyManager::KeysView::load() {
...
+  myfile.close();
 }
 
-std::string KeyManager::PersistentSaverLoader::load() { return ps->getKeysView(); }
+std::string KeyManager::FileSecureStore::load() {
+  std::ifstream inFile;
+  inFile.open(fileName.c_str());
+  if (!inFile.good()) {
+    LOG_WARN(KEY_EX_LOG, "key file wasn't loaded " << fileName);
+    return "";
```
f-squirrel, line 320: Nit: If the empty string is to identify a failure, it might be better to change the return type to `std::optional<std::string>>`.

f-squirrel (approved):

bftengine/src/bftengine/KeyManager.cpp
cloudnoize: It's a good idea, I'll refactor it in a next PR

PR #921, opened 2020-10-06
Save Apollo logs when running natively

oded282:
I turned on the 'KEEP_APOLLO_LOGS' flag so that a developer who builds natively would have the same logs as in the CI and the dockerized build.

*Removed the code which opens the old logs directory.


teoparvanov (approved):
LGTM, thanks @oded282 !

PR #924, opened 2020-10-07
BC-4947: Disabling unstable pre-execution test

teoparvanov:
This test runs pre-execution commands while dropping a small amount of UDP packets. This failure has lower priority, given we are going to move to a TLS setup in the near future.


yuliasherman (approved):
LGTM

PR #928, opened 2020-10-08
test_state_transfer wouldn't fail correctly for f>1

KTeofanidis:
Due to only stopping one node the test would always succeed irrespective of stale_node working correctly or not when f>1

### DIFF
--- a/tests/apollo/test_skvbc.py
+++ b/tests/apollo/test_skvbc.py
@@ -54,9 +54,9 @@ class SkvbcTest(unittest.TestCase):
         Test that state transfer starts and completes.
 
         Stop one node, add a bunch of data to the rest of the cluster, restart
-        the node and verify state transfer works as expected. In a 4 node
-        cluster with f=1 we should be able to stop a different node after state
-        transfer completes and still operate correctly.
+        the node and verify state transfer works as expected. We should be able
+        to stop a different set of f nodes after state transfer completes and
+        still operate correctly.
         """
         if exchange_keys:
             await bft_network.do_key_exchange()
@@ -75,9 +75,7 @@ class SkvbcTest(unittest.TestCase):
         await bft_network.wait_for_state_transfer_to_start()
         await bft_network.wait_for_state_transfer_to_stop(0, stale_node)
         await skvbc.assert_successful_put_get(self)
-        random_replica = random.choice(
-            bft_network.all_replicas(without={0, stale_node}))
-        bft_network.stop_replica(random_replica)
+        await bft_network.force_quorum_including_replica(stale_node)
         await skvbc.assert_successful_put_get(self)
 
     @with_trio
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -624,24 +624,16 @@ class BftTestNetwork:
                     count_if_replica_in_view, r, view)
         return nb_replicas_in_view
 
-    def force_quorum_including_replica(self, replica_id, primary=0):
+    async def force_quorum_including_replica(self, replica_id):
         """
         Bring down a sufficient number of replicas (excluding the primary),
         so that the remaining replicas form a quorum that includes replica_id
         """
         with log.start_action(action_type="force_quorum_including_replica") as action:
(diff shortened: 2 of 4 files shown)


tests/apollo/test_skvbc.py
```diff
@@ -75,9 +75,8 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
         await bft_network.wait_for_state_transfer_to_start()
         await bft_network.wait_for_state_transfer_to_stop(0, stale_node)
         await skvbc.assert_successful_put_get(self)
-        random_replica = random.choice(
-            bft_network.all_replicas(without={0, stale_node}))
-        bft_network.stop_replica(random_replica)
+        bft_network.stop_replicas(bft_network.random_set_of_replicas(
+            bft_network.config.f, without={0, stale_node}))
```
teoparvanov, line 79: This is great, but can you actually move this logic into the `bft.force_quorum_including_replica()` method and use it wherever applicable?

teoparvanov (commented):
Great find, thanks @KTeofanidis ! I have a suggestion how to make this even better...

tests/apollo/util/bft.py
```diff
@@ -630,18 +630,9 @@ def force_quorum_including_replica(self, replica_id, primary=0):
...
-                action.log(f'Stopping backup replica {backup_replica_id} in order '
-                      f'to force a quorum including replica {replica_id}...')
-                self.stop_replica(backup_replica_id)
-                if len(self.procs) == 2 * self.config.f + self.config.c + 1:
-                    break
-
-            assert len(self.procs) == 2 * self.config.f + self.config.c + 1
+            assert len(self.procs) >= 2 * self.config.f + self.config.c + 1
+            self.stop_replicas(self.random_set_of_replicas(
+                len(self.procs) - (2 * self.config.f + self.config.c + 1), without={primary, replica_id}))
```
teoparvanov, line 635: why not just 
```python
            self.stop_replicas(self.random_set_of_replicas(f, without={primary, replica_id}))
```
?

KTeofanidis: The original code was correct. We need at least 2f + c + 1 nodes for the quorum, so for example in the edge case n6 c1 f1 we need to stop 2 instead of 1 to ensure the relevant replica_id is a non failed part of the quorum of 4.

teoparvanov: ok, makes sense

tests/apollo/test_skvbc.py
```diff
@@ -75,9 +75,7 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
         await bft_network.wait_for_state_transfer_to_start()
         await bft_network.wait_for_state_transfer_to_stop(0, stale_node)
         await skvbc.assert_successful_put_get(self)
-        random_replica = random.choice(
-            bft_network.all_replicas(without={0, stale_node}))
-        bft_network.stop_replica(random_replica)
+        bft_network.force_quorum_including_replica(stale_node)
```
teoparvanov, line 78: can you also use this function in the test you recently added in the network partitioning suite?

teoparvanov (approved):
LGTM, thanks!

tests/apollo/util/bft.py
teoparvanov, line 635: Passing the primary as a parameter could lead to inconsistencies. Could you instead dynamically retrieve it, such as:
```python
primary = await bft_network.get_current_primary()
```
and also remove it from the method's parameters?

teoparvanov (changes requested):
one last thing before we merge this - see my comment above

tests/apollo/util/bft.py
KTeofanidis: This is going to add some overhead considering currently the method isn't async. Are you sure?

teoparvanov: Yes

teoparvanov: PS: don't forget to add `await` when calling the force quorum method

teoparvanov (approved):
Looks great, thanks @KTeofanidis !

PR #929, opened 2020-10-08
Fix memory leak reported by LeakSanitizer.

HristoStaykov:
This leak was reported during local run of skvbc_basic_tests.


f-squirrel (approved):
LGTM!
Imo, a safer option would be to change the signature of [virtual IThresholdAccumulator *newAccumulator(bool withShareVerification)
](https://github.com/vmware/concord-bft/blob/master/threshsign/include/threshsign/IThresholdVerifier.h#L28) to return a `unique_ptr`.

PR #931, opened 2020-10-09
Add histograms for merkle based storage

andrewjstone:
Add many histograms to cover sparse merkle code as well as DBAdapter
code for the sparse merkle tree based storage backend.

Since computing of parent block hashes happens in another thread using
std::async, we need to use atomic updates for the related histograms.
The diagnostics Recorder and TimeRecorder classes were enhanced to
support atomic update using the underlying HdrHistogram support.

### DIFF
--- a/kvbc/include/sparse_merkle/histograms.h
+++ b/kvbc/include/sparse_merkle/histograms.h
@@ -11,6 +11,8 @@
 // terms and conditions of the sub-component's license, as noted in the
 // LICENSE file.
 
+#pragma once
+
 #include "diagnostics.h"
 #include "sparse_merkle/base_types.h"
 
@@ -124,6 +126,6 @@ struct Recorders {
   std::shared_ptr<Recorder> dba_hashed_parent_block_size = std::make_shared<Recorder>(1, MAX_VAL_SIZE, 3, Unit::BYTES);
 };
 
-static Recorders histograms;
+inline const Recorders histograms;
 
 }  // namespace concord::kvbc::sparse_merkle::detail
--- a/kvbc/include/sparse_merkle/histograms.h
+++ b/kvbc/include/sparse_merkle/histograms.h
@@ -1,0 +1,129 @@
+
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the sub-component's license, as noted in the
+// LICENSE file.
+
+#include "diagnostics.h"
+#include "sparse_merkle/base_types.h"
+
+namespace concord::kvbc::sparse_merkle::detail {
+
+struct Recorders {
+  static constexpr int64_t MAX_NS = 1000 * 1000 * 1000;       // 1s
+  static constexpr int64_t MAX_VAL_SIZE = 1024 * 1024 * 100;  // 100MB
+
+  using Recorder = concord::diagnostics::Recorder;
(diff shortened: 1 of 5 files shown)


kvbc/include/sparse_merkle/histograms.h
```diff
@@ -0,0 +1,129 @@
...
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the sub-component's license, as noted in the
+// LICENSE file.
+
```
dartdart26, line 13: Nit: Missing include guard.

kvbc/include/sparse_merkle/histograms.h
```diff
@@ -0,0 +1,129 @@
...
+  std::shared_ptr<Recorder> dba_get_leaf_key_val_at_most_version =
+      std::make_shared<Recorder>(1, MAX_NS, 3, Unit::NANOSECONDS);
+  std::shared_ptr<Recorder> dba_delete_keys_for_block = std::make_shared<Recorder>(1, MAX_NS, 3, Unit::NANOSECONDS);
+  std::shared_ptr<Recorder> dba_has_block = std::make_shared<Recorder>(1, MAX_NS, 3, Unit::NANOSECONDS);
+  std::shared_ptr<Recorder> dba_keys_for_version = std::make_shared<Recorder>(1, MAX_NS, 3, Unit::NANOSECONDS);
+  std::shared_ptr<Recorder> dba_hash_parent_block = std::make_shared<Recorder>(1, MAX_NS, 3, Unit::NANOSECONDS);
+  std::shared_ptr<Recorder> dba_hashed_parent_block_size = std::make_shared<Recorder>(1, MAX_VAL_SIZE, 3, Unit::BYTES);
+};
+
+static Recorders histograms;
```
dartdart26, line 127: If I am not missing something, I think there is a problem with using a `static` variable in a header in this context. What I mean is that we will have multiple `Recorders` instances in different translation units. However, the constructors of these instances will add to a common `singleton`, throwing exceptions for existing components in `PerformanceHandler::registerComponent()`.

I think a simple solution would be to declare the variable as `inline const Recorders histograms;`.

kvbc/src/sparse_merkle/tree.cpp
```diff
@@ -114,8 +120,18 @@ void remove(Walker& walker, const Hash& key_hash) {
   }
 }
 
+void updateBatchHistograms(const UpdateBatch& batch) {
```
dartdart26, line 123: Nit: Can be declared as `static`.

dartdart26 (changes requested):
Wow, thanks for doing that @andrewjstone ! It will be really useful.

I have put a few minor comments. One of them that I think should be fixed is related to multiple registrations for the same component.

andrewjstone:
> Wow, thanks for doing that @andrewjstone ! Looks great and it will be really useful.
> 
> I have put a few minor comments. One of them that I think should be fixed is related to multiple registrations for the same component.

Thanks for the great review @dartdart26 . I was clearly confused about the usage of static and inline. I had it backwards, but then again the naming is not exactly ideal :) Really appreciate the help here. You saved me a bunch of time with your suggestion.

kvbc/src/sparse_merkle/tree.cpp
andrewjstone: fixed.

kvbc/include/sparse_merkle/histograms.h
andrewjstone: Fixed.

dartdart26 (approved):
Thanks @andrewjstone !

PR #932, opened 2020-10-11
Coverity - 86318 Argument cannot be negative

oded282:
Validate the socket value, and return -1 if the socket couldn't be open.

### DIFF
--- a/communication/src/PlainUDPCommunication.cpp
+++ b/communication/src/PlainUDPCommunication.cpp
@@ -164,6 +164,11 @@ class PlainUDPCommunication::PlainUdpImpl {
     // Initialize socket.
     udpSockFd = socket(AF_INET, SOCK_DGRAM, 0);
 
+    if (udpSockFd < 0) {
+      LOG_FATAL(_logger, "Failed to initialize socket, error: " << strerror(errno));
+      std::terminate();
+    }
+
     // Name the socket.
     sAddr.sin_family = AF_INET;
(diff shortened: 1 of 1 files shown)


communication/src/PlainUDPCommunication.cpp
```diff
@@ -164,6 +164,11 @@ class PlainUDPCommunication::PlainUdpImpl {
     // Initialize socket.
     udpSockFd = socket(AF_INET, SOCK_DGRAM, 0);
 
+    if (udpSockFd < 0) {
+      LOG_DEBUG(_logger, "Couldn't open the socket");
```
f-squirrel, line 168: Since it seems to be a serious issue, I think `LOG_ERROR` fits more.

communication/src/PlainUDPCommunication.cpp
```diff
@@ -164,6 +164,11 @@ class PlainUDPCommunication::PlainUdpImpl {
...
+      return -1;
```
f-squirrel, line 169: In the [code](https://github.com/vmware/concord-bft/blob/master/communication/src/PlainUDPCommunication.cpp#L178) below we terminate the application in the event of failure.
Probably, we might want to do the same here:
`ConcordAssert(udpSockFd >= 0, std::srting{"Failed to initialize socket, error: "} + strerror(errno));`

f-squirrel (changes requested):

f-squirrel (approved):
LGTM!
Thank you @oded282 for addressing the comments!

PR #934, opened 2020-10-12
create_tls_certs.sh: extend user certificate generation options

glevkovich:
Here we modify this script to give the user more control over how
certificates are generated:
1) new optional input - 'output folder'. If not given, we keep ./certs
as default.
2) new optional input - 'start node ID'. If not given, we keep 0 as the
start node id.

Both of these capabilities are needed for testing purposes (To support Apollo TLS communication tests).
Additional remarks:
1) We also made some "cosmetic" changes in the script.
2) The lines starting with 'openssl req ' has some bug, double quotes
inside double quotes are not allowed in bash.
3) Keep the positional and optional input convention for now. Examples
are given above.

Issue: BC-4932

### DIFF
--- a/scripts/linux/create_tls_certs.sh
+++ b/scripts/linux/create_tls_certs.sh
@@ -1,38 +1,54 @@
 #!/usr/bin/env bash
-# creates simple self signed certificates to use with TCP TLS module
-# by default, it will create "certs" folder in the current folder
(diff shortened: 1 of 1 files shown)


glevkovich:
opened a new one after joining VMware on Github

PR #937, opened 2020-10-12
Modify tests config to accept command line TLS certificate root path

glevkovich:
Until now the only option for a simpleKVBC TesterReplica,
was to look for its TLS certificates in the current working folder under
"certs". Apollo tests run under some random folder under /tmp, and
the certificate is generated into this folder. To save time, we would
like to allow the user to point into this folder
and override the default case. We add an option -c / --cert-root-path.
This is only optional, if not given, a replica will look under ./certs.
This option should be used only when a replica supports TCP/TLS
communication, else it is ignored.


teoparvanov (approved):
LGTM!

PR #938, opened 2020-10-12
Upgrade OpenSSL and install net-tools on concord-bft Docker image

glevkovich:
1) OpenSSL:
The current version 1.1.1 is quite old, from 2018. It has a bug that
prevents running the script create_tls_certs.sh inside a container.
We need to run this script to run Apollo tests with TLS communication.
This version is currently the default when installing using the
official Ubuntu package manager.
Version 1.1.1g fixes that bug and is much recent (and stable).
Here we change install_deps.sh to download, configure, build and install
OpenSSL from https://www.openssl.org.

2) net-tools:
We need netstat and (might) need other utilities from this package
for our tests. Let's add it to the image.


f-squirrel (approved):
LGTM!

PR #941, opened 2020-10-13
Using new metrics to provide fast/slow path testing

KTeofanidis:
New metrics that count requests rather than batches, so the client can ensure a certain number of requests that were sent were processed on the expected path.

Rewritten the wait_for methods to send requests as necessary and check the counts until the requested path is prevalent or a timeout.

Retained the assert_X_path_prevalent option for cases where the tests want to check if a predetermined series of events happened on the expected path similar to the previous functionality except they should be deterministic with an exact threshold of maximum number requests processed on the undesired path and minimum on the desired.

_wait_for_consensus_path_to_be_prevalent algorithm:
  read new metrics to set baseline number
  use the callback to send requests
  read the new metrics
  if there are any requests processed on the undesired path wait 5 seconds and send new requests
  if there are no undesired requests, but not enough requests on the desired path wait .5 seconds and read the metrics again. If the desired number(threshold input parameter) isn't reached within 15 seconds sends new requests.
  if neither failure mode happens i.e. we have enough desired path requests to clear the threshold and no undesired requests we succeed.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -32,6 +32,7 @@ sys.path.append(os.path.abspath("../../util/pyclient"))
 import bft_config
 import bft_client
 import bft_metrics_client
+from enum import Enum
 from util import bft_metrics, eliot_logging as log
 from util.eliot_logging import log_call
 from util import skvbc as kvbc
@@ -48,6 +49,11 @@ TestConfig = namedtuple('TestConfig', [
     'stop_replica_cmd',
     'num_ro_replicas'
 ])
+
+class ConsensusPathPrevalentResult(Enum):
+   OK = 0
+   TOO_FEW_REQUESTS_ON_EXPECTED_PATH = 1
+   TOO_MANY_REQUESTS_ON_UNEXPECTED_PATH = 2
 
 KEY_FILE_PREFIX = "replica_keys_"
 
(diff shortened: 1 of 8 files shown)


tests/apollo/util/bft.py
```diff
@@ -825,35 +849,49 @@ async def assert_state_transfer_not_started_all_up_nodes(self, up_replica_ids):
...
+        with log.start_action(action_type="assert_%s_path_prevalent" % ("fast" if fast else "slow")):
+            total_nb_fast_paths = await self.num_of_fast_path_requests(replica_id)
+            total_nb_slow_paths = await self.num_of_slow_path_requests(replica_id)
+
+            log.log_message("assert on %s path | slow %d/%d fast %d/%d" % ("fast" if fast else "slow",
+                total_nb_slow_paths, nb_slow_paths_to_ignore, total_nb_fast_paths, nb_fast_paths_to_ignore))
+
+            if fast:
+                if total_nb_slow_paths > nb_slow_paths_to_ignore:
+                    return ConsensusPathPrevalentResult.TOO_MANY_WRONG_PATH_REQUESTS  # we can't have any slow
```
teoparvanov, line 883: `TOO_MANY_REQUESTS_ON_UNEXPECTED_PATH` ?

tests/apollo/util/bft.py
```diff
@@ -825,35 +849,49 @@ async def assert_state_transfer_not_started_all_up_nodes(self, up_replica_ids):
...
+                if total_nb_fast_paths <= nb_fast_paths_to_ignore:
+                    return ConsensusPathPrevalentResult.NOT_ENOUGH_RIGHT_PATH_REQUESTS  # we don't have enough fast
```
teoparvanov, line 885: `TOO_FEW_REQUESTS_ON_EXPECTED_PATH`

teoparvanov (approved):
LGTM, some minor naming-related comments from me.

PR #942, opened 2020-10-13
[Apollo] Reproduce BC-4982 outside of long-running tests

teoparvanov:
1) Select two clients - one that runs "long" and one that runs "short" requests
2) Set the timeout of the long client to LONG_REQ_TIMEOUT_MILLI
3) Concurrently:
  3.1) Run a long executing request via the "long" client
  3.2) Trigger a view change during the long execution
4) Run a short write using the "short" client (succeeds)
5) Run a short write using the "long" client (fails because the long client is not freed in the pre-processor)


yuliasherman (approved):
LGTM!

PR #946, opened 2020-10-13
Improve TLS exception handling

salieri11:
Wrap potential IO exception with try:catch and add more informative logging for connection handling


yuliasherman (approved):
LGTM

PR #949, opened 2020-10-14
Refactor TLS code

andrewjstone:
Split the code out of a single file. All TlsTCP communication code uses
pImpl to call into the implemntation class TlsTcpImpl. This class
manages and keeps track of all connections and moves them through
multiple phases of the handshake protocol by storing them in different
maps. This allows easier global tracking of connections. This tracking
is made visible to diagnostics through a status handler.

Actual TLS negotiation and connection management, reads, writes, and
timers are managed for each connection individually in
AsyncTLSConnection.

The structure of this code base now more clearly defines a concrete
separation between the tracking of connections, and the individual usage
of the connections.

Additionally, methods were created and named to clearly define
individual actions taken in the code base, and made to do one thing
only. The naming follows the event driven flow of ASIO and makes the
code a bit easier to follow.

It should be noted that all the core ASIO code remains as is, including
the cert pinning code. This is purely a structural refactoring, with
minor behavioral changes to fix bugs or enhance safety. Big thanks to
@salieri11 for doing all the initial hard work of actually getting TLS
working. Essentially none of that code required changing.

One change that is worth mentioning is regarding the size message
header. It is now passed in network byte order rather than from raw
bytes in memory making it safe across platforms. This is in anticipation
of future changes to messaging that will also be platform independent.

Additional logging was also added, and fixed such that port and
connection numbers are accurately reported.

Enhancements were also made, such that old messages that have been
waiting in queue and are likely stale, are dropped before they get sent.
A max send queue size is also set to prevent unbounded memory growth on
the send side.

This was used tested in the product and significant tests passed.


communication/src/TlsTcpImpl.cpp
```diff
@@ -0,0 +1,325 @@
...
+  resolving_.insert(i);
+  status_->num_resolving = resolving_.size();
+  auto node = config_.nodes.at(i);
+  boost::asio::ip::tcp::resolver::query query(boost::asio::ip::tcp::v4(), node.host, std::to_string(node.port));
+  resolver_.async_resolve(query, [this, node, i, query](const auto& error_code, auto results) {
+    if (error_code) {
+      LOG_WARN(
+          logger_,
+          "Failed to resolve node " << i << ": " << node.host << ":" << node.port << " : " << error_code.message());
+      return;
```
andrewjstone, line 261: Need to clean the `resolving_` map and `status_` before returning.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+        LOG_ERROR(logger_,
+                  "Reading message of size <<" << getReadMsgSize() << " failed for node " << peer_id_.value() << ": "
+                                               << error_code.message());
+        return dispose();
+      }
+    }
+
+    // The Read succeeded.
+    boost::system::error_code _ec;
+    read_timer_.cancel(_ec);
```
f-squirrel, line 69: Do we want to make sure that the `_ec` does not return an error?
We might want to log it.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+  // We send in network byte order
+  return ntohl(*reinterpret_cast<uint32_t*>(read_size_buf_.data()));
+}
+
+void AsyncTlsConnection::dispose() {
+  ConcordAssert(!disposed_);
+  LOG_ERROR(logger_, "Closing connection to node " << peer_id_.value());
+  disposed_ = true;
+  boost::system::error_code _;
+  read_timer_.cancel(_);
```
f-squirrel, line 115: Nit: log the error code if present?

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+            // The socket has already been cleaned up and any references are invalid. Just return.
+            return;
+          }
+          LOG_WARN(logger_,
+                   "Write failed to node " << peer_id_.value() << " for message with size "
+                                           << out_queue_.front().msg.size() << ": " << ec.message());
+          return dispose();
+        }
+        // The write succeeded.
+        write_timer_.cancel();
```
f-squirrel, line 167: AFAIK, `cancel()` throws in the event of an error, maybe we can pass `boost:error` as it is done in the code above?

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+  boost::system::error_code ec;
+  ssl_context_.set_verify_callback(
+      [this, self, destination](auto preverified, auto& ctx) -> bool {
+        if (self.expired()) return false;
+        return verifyCertificateClient(preverified, ctx, destination);
+      },
+      ec);
+  if (ec) {
+    LOG_ERROR(logger_, "Unable to set client verify callback" << ec.message());
```
f-squirrel, line 204: I think it might be better to use `LOG_FATAL`, because we terminate the app.
I guess that `LOG_FATAL` might flush the log immediately(just speculation :) ).

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+    abort();
```
f-squirrel, line 205: `std::terminate` might be a better [choice](https://github.com/vmware/concord-bft#notes) because it gives a possible user to set the termination handler.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+    return false;
+  }
+  X509_NAME_oneline(X509_get_subject_name(cert), subject.data(), 256);
+  auto [valid, _] = checkCertificate(cert, "server", subject, expected_dest_id);
+  (void)_;  // unused variable hack
+  return valid;
+}
+
+bool AsyncTlsConnection::verifyCertificateServer(bool preverified, boost::asio::ssl::verify_context& ctx) {
+  std::string subject(512, 0);
```
f-squirrel, line 293: Extract `512` to a variable?

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+  fs::path path;
+  try {
+    path = fs::path(tlsTcpImpl_.config_.certificatesRootPath) / std::to_string(remotePeerId) / connectionType /
+           std::string(connectionType + ".cert");
+  } catch (std::exception& e) {
+    LOG_FATAL(logger_, "Failed to construct filesystem path: " << e.what());
+    ConcordAssert(false);
+  }
+
+  FILE* fp = fopen(path.c_str(), "r");
```
f-squirrel, line 362: I know that the code does not throw exceptions but I think it might be safer to wrap the pointer in a unique_ptr.

f-squirrel (approved):
LGTM! Only a few non-critical comments.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+        std::lock_guard<std::mutex> guard(write_lock_);
+        queued_size_in_bytes_ -= out_queue_.front().msg.size();
+        out_queue_.pop_front();
+        if (!out_queue_.empty()) {
+          write();
+        }
+      });
+}
+
+void AsyncTlsConnection::createSSLSocket(boost::asio::ip::tcp::socket&& socket) {
```
robem, line 177: Probably just a copy and paste but the coding standard wants us to use `Ssl` instead.
"[...] prefer to capitalize abbreviations as single words, e.g., StartRpc() rather than StartRPC()"

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+  }
+
+  NodeNum remotePeerId;
+  try {
+    remotePeerId = stoul(remPeer, nullptr);
+  } catch (const std::invalid_argument& ia) {
+    LOG_ERROR(logger_, "cannot convert OU, " << subject << ", " << ia.what());
+    return std::make_pair(false, 0);
+  } catch (const std::out_of_range& e) {
+    LOG_ERROR(logger_, "cannot convert OU, " << subject << ", " << e.what());
```
robem, line 338: Not worth factoring this out for both exceptions, so, duplication it is. Thanks C++.

communication/src/AsyncTlsConnection.h
```diff
@@ -0,0 +1,190 @@
...
+  boost::asio::steady_timer write_timer_;
+
+  // On every read, we must read the size of the incoming message first. This buffer stores that size.
+  std::array<char, MSG_HEADER_SIZE> read_size_buf_;
+
+  // Last read message
+  std::vector<char> read_msg_;
+
+  // We must maintain ownership of the in_flight_message until the asio::buffer wrapping it has actually been sent by
+  // the underling io_service. We will know this is the case when the write completion handler gets called.
```
robem, line 175: "underlying" although "underling" kinda works too :D

communication/src/TlsTcpImpl.cpp
```diff
@@ -0,0 +1,327 @@
...
+  }
+  for (auto& [_, conn] : connected_waiting_for_handshake_) {
+    (void)_;  // unused variable hack
+    syncCloseConnection(conn);
+  }
+
+  for (auto& [_, conn] : connected_waiting_for_handshake_) {
+    (void)_;  // unused variable hack
+    syncCloseConnection(conn);
+  }
```
robem, line 65: Why are we doing the same thing twice?

communication/src/TlsTcpImpl.cpp
```diff
@@ -0,0 +1,327 @@
...
+bool TlsTCPCommunication::TlsTcpImpl::isRunning() const {
+  std::lock_guard<std::mutex> l(startStopGuard_);
+  return io_thread_.get() != nullptr;
+}
+
+void TlsTCPCommunication::TlsTcpImpl::setReceiver(NodeNum nodeId, IReceiver* receiver) {
+  // We don't allow setting a receiver after startup
+  ConcordAssert(!isRunning());
+  receiver_ = receiver;
+}
```
robem, line 115: `nodeId`?

communication/src/TlsTcpImpl.h
```diff
@@ -0,0 +1,252 @@
...
+  TlsTcpConfig config_;
+
+  // The lifetime of the receiver must be at least as long as the lifetime of this object
+  IReceiver *receiver_;
+
+  // This thread runs the asio io_service. It's the main thread of the TlsTcpImpl.
+  std::unique_ptr<std::thread> io_thread_;
+
+  // io_thread_ lifecycle management
+  mutable std::mutex startStopGuard_;
```
robem, line 206: We should rename this to `io_thread_guard_`

robem (approved):
Great change - thanks for the hard work 💪  🚀 

The change looks good to me and all my comments are no blockers. Please address them as you see fit.

communication/src/AsyncTlsConnection.cpp
andrewjstone: I'm not sure it makes sense to log here.There's really nothing we can do, and it's not fatal. An error is only created when the handlers:

-   have already been invoked; or
-   have been queued for invocation in the near future.

While the latter is something we would never expect, the former may end up being a common case and littering our log with unnecessary warnings. I prefer to leave this as is.

Relatedly, I feel like `_` is a pretty common convention and prefer to leave that as well.

communication/src/AsyncTlsConnection.cpp
andrewjstone: same as above.

communication/src/AsyncTlsConnection.cpp
andrewjstone: Good catch!

communication/src/AsyncTlsConnection.cpp
andrewjstone: good catch!

communication/src/AsyncTlsConnection.cpp
andrewjstone: Good catch! I used `ConcordAssert(false);` as seems to be the pattern elsewhere, and correlates with Igor's latest changes to the existing TLS code.

communication/src/AsyncTlsConnection.cpp
andrewjstone: Agreed. Done.

communication/src/TlsTcpImpl.cpp
andrewjstone: Great catch! This should be `accepted_waiting_for_handshake`.

communication/src/TlsTcpImpl.h
andrewjstone: Yup. I also changed `connectionsGuard_` to `connections_guard_`.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,390 @@
...
+void AsyncTlsConnection::send(std::vector<char>&& raw_msg) {
+  std::lock_guard<std::mutex> guard(write_lock_);
+  if (queued_size_in_bytes_ + raw_msg.size() > MAX_QUEUE_SIZE_IN_BYTES) {
+    LOG_WARN(logger_, "Outgoing Queue is full. Dropping message with size: " << raw_msg.size());
+    return;
+  }
+
+  queued_size_in_bytes_ += raw_msg.size();
+  out_queue_.emplace_back(OutgoingMsg(std::move(raw_msg)));
+  if (out_queue_.size() == 1) {
```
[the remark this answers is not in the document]
andrewjstone: If `size > 1` then the io thread is already writing. I'll put a comment. Good point!

communication/src/AsyncTlsConnection.cpp
andrewjstone: going to leave this one for now.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -0,0 +1,417 @@
...
+  EC_KEY_free(ecdh);
+
+  // Only allow using the strongest cipher suites.
+  SSL_CTX_set_cipher_list(ssl_context_.native_handle(), tlsTcpImpl_.config_.cipherSuite.c_str());
+}
+
+bool AsyncTlsConnection::verifyCertificateClient(bool preverified,
+                                                 boost::asio::ssl::verify_context& ctx,
+                                                 NodeNum expected_dest_id) {
+  std::string subject(256, 0);
```
salieri11, line 305: In the original code there is a check here:
if (X509_V_ERR_DEPTH_ZERO_SELF_SIGNED_CERT != X509_STORE_CTX_get_error(ctx.native_handle())) {
      return false;
}
This code was present at verify_certificate_server as well. This was required by security team and it basically validates only zero level cert (which is pinned). So please can you add it to both methods?

salieri11 (dismissed):
Look good! Thanks for this refactoring. I just have one comment related to some missing code.

communication/src/AsyncTlsConnection.cpp
andrewjstone: Thanks for pointing this out @salieri11 ! 

I was wondering how I missed this, then remembered the bulk of this refactoring was done about 6 months ago. These 2 changes are the only recent ones, besides the extra checks added last week that I already took into account.

andrewjstone: Fixed and pushed into the first commit with git fixup.

teoparvanov:
Hi Andrew, Gil recently merged the Python TLS client and created additional CI jobs that run Apollo tests in a full TLS setup. Could you rebase your PR, to benefit from those additional checks?

Thanks!
-- Teo

andrewjstone:
Will Do. This is excellent. Thanks for the heads up Teo!
________________________________

bftengine/src/preprocessor/tests/preprocessor_test.cpp
```diff
@@ -435,6 +441,7 @@ TEST(requestPreprocessingState_test, requestTimedOut) {
   usleep(replicaConfig.preExecReqStatusCheckTimerMillisec * 1000);
   timers.evaluate();
   ConcordAssert(preProcessor.getOngoingReqIdForClient(clientId) == 0);
+  clearDiagnosticsHandlers();
```
Liorgal28, line 444: Thank you for including that and handling those changes in the pre-processor.
The reason why I have included the try-catch is exactly that the tests are trying to register handlers that have already been registered.

Liorgal28 (approved):

PR #950, opened 2020-10-16
Prolong the maxmimum source replica selection timeout from 5s to 15s

toly-kournik:
to deal with larger blocks.


teoparvanov (approved):
LGTM!

PR #951, opened 2020-10-17
skvbc_history_tracker: fix bug in send_indefinite_tracked_ops()

glevkovich:
The modified block of code is intended to have a sequence of
random reads and writes. But before change we got 100% writes since in
both cases we called self.send_tracked_write().


andrewjstone (approved):
Great catch @glevkovich !

PR #952, opened 2020-10-18
Stateless docker build

f-squirrel:
Rationale:
Avoid restarting the docker container manually on every update of the image or switch between release branches.

Solution:
Since there is no intermediate state of the build docker container except the `build` directory, we can
make the docker build stateless. As a result, the update of docker image would be easier: no need to restart the
container after the rebase or to make sure that you’re using the latest image because `docker run` would download the
latest automatically.

Changes:
* Every command launches a docker container: replaced all the `exec` commands with `run`
* `make login` launches a new container only if it does not exist, otherwise it enters the existing
* Forwared the error output of `docker container inspect` to `/dev/null`
* Removed `run-c`, `stop-c`, and `remove-c` targets

Usage:
* Once the commit is merged, need to run `docker kill concord-bft`
to allow launching it.
* All the rest is exactly the same

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,15 @@ CONCORD_BFT_CTEST_TIMEOUT:=3000 # Default value is 1500 sec. It takes 2500 to ru
 CONCORD_BFT_USER_GROUP:=--user `id -u`:`id -g`
 
 CONCORD_BFT_ADDITIONAL_RUN_PARAMS:=
-CONCORD_BFT_ADDITIONAL_RUN_COMMANDS:=
+
+BASIC_RUN_PARAMS:=--rm --privileged=true \
+					  --cap-add NET_ADMIN --cap-add=SYS_PTRACE --ulimit core=-1 \
+					  --name="${CONCORD_BFT_DOCKER_CONTAINER}" \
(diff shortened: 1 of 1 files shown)


salieri11 (approved):
looks good

Makefile
```diff
@@ -30,74 +30,46 @@ CONCORD_BFT_CTEST_TIMEOUT:=3000 # Default value is 1500 sec. It takes 2500 to ru
...
-			  ${CONCORD_BFT_CONTAINER_SHELL} -c \
-			  "${CONCORD_BFT_ADDITIONAL_RUN_COMMANDS} \
-			  /usr/bin/tail -f /dev/null"
-	@echo
-	@echo "The container \"${CONCORD_BFT_DOCKER_CONTAINER}\" with the build environment is started as daemon."
-	@echo "Run \"make stop-c\" to stop or \"make remove-c\" to delete."
-
 .PHONY: login
-login: ## Login to the container
+login: ## Login to the container. Note: if the container is already running, logins into exisiting one
```
andrewjstone, line 61: typos: logins -> login, exisiting -> existing

andrewjstone (approved):
LGTM. Thanks so much for doing this @f-squirrel

PR #954, opened 2020-10-18
Modify docker image and add psmisc package, update Docker image to 0.9

glevkovich:
Install package psmisc into concord-bft Docker image
This package is needed for the Apollo tests, for the port discovery by
bft_network_partitioning.py (network partition manager), The command
fuser is used in order to get which process ids use which sockets.


f-squirrel (approved):
LGTM!

PR #956, opened 2020-10-19
Modify Apollo test framwork to run with TLS/TCP

glevkovich:
This is a WIP, some tests still do not pass. UDP is enabled by default. Please follow this Pr commit by commit with the commit messages to better understand the changes (file diff is less recommended).

This commit must not be merged before merging: https://github.com/vmware/concord-bft/pull/954

Series of commits:

1) Modify BftTestNetwork to support TLS/TCP communication
2) Refactor pyclient client class hierarchy, Implement TcpTlsClient client
3) Make required changes when running with partitions -   to start all replicas on test startup, including reserved replicas.
4) Disable test_rough_initial_key_exchange due to **BC-5081** 
5) bft_network_partitioning: modify to support TCP communication
6) Make changes to fix test_client.py on UDP


tests/apollo/util/bft.py
```diff
@@ -56,17 +57,25 @@ class ConsensusPathPrevalentResult(Enum):
    TOO_MANY_REQUESTS_ON_UNEXPECTED_PATH = 2
 
 KEY_FILE_PREFIX = "replica_keys_"
+# To run TLS/UDP clients, replace the next constant with a client implemented in bft_client.py 
+# For now the choices are UdpClient or TcpTlsClient:
+BFT_CLIENT_TYPE = bft_client.UdpClient
```
teoparvanov, line 62: Why is the UdpClient hard-coded here? How does it change if we want to run with TLS?

tests/apollo/util/bft.py
```diff
@@ -225,23 +233,28 @@ def new(cls, config, client_factory=None):
             builddir=builddir,
             toolsdir=toolsdir,
             procs={},
-            replicas=[bft_config.Replica(i, "127.0.0.1", 3710 + 2*i, 4710 + 2*i)
-                for i in range(0, config.n + config.num_ro_replicas)],
+            replicas=[bft_config.Replica(i, "127.0.0.1",
+                                         bft_config.data_port_from_node_id(i), bft_config.metrics_port_from_node_id(i))
```
teoparvanov, line 237: I'd suggest a better name:
`data_port` -> `bft_port`

tests/apollo/util/bft.py
```diff
@@ -431,6 +506,62 @@ def start_replica(self, replica_id):
...
+        return None
+
+    def is_replica_id(self, node_id):
+        """ Return True if a node_id is a replica id (includes a read-only replica) """
+        return node_id < self.num_total_replicas()
+
+    def is_client_id(self, node_id):
+        """ Return True if node_id is a client Id - reserved or non-reserved """
+        start_id = self.num_total_replicas()
+        return start_id <= node_id < (start_id + + self.config.num_clients + RESERVED_CLIENTS_QUOTA)
```
teoparvanov, line 526: `+ +` ? Python doesn't seem to complain, but I'm wondering how does he interpret it 🐍

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,256 @@ def get_rsi_replies(self):
...
+        self.sock.close()
+
+    def __enter__(self):
+        """context manager method for 'with' statements"""
+        return self
+
+    def __exit__(self, *args):
+        """context manager method for 'with' statements"""
+        self.sock.close()
+        pass
```
teoparvanov, line 317: nit - unnecessary

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,256 @@ def get_rsi_replies(self):
...
+        """
+        cert_type = "client" if is_client else "server"
+        return os.path.join(self.config.certs_path, str(replica_id), cert_type, cert_type + ".cert")
+
+    async def _establish_ssl_stream(self, dest_replica):
+        """
+        Try every 250ms to connect to dest_replica. We relay on the caller context that will trigger cancalleation at 
+        some point in time. Break the loop on success and keep the ssl_stream in self.ssl_streams for future use.
+        """
+        server_cert_path = self._get_cert_path(dest_replica.id, False)
```
teoparvanov, line 372: Leverage the optionality of the parameter: `server_cert_path = self._get_cert_path(dest_replica.id)`

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,256 @@ def get_rsi_replies(self):
...
+        client_cert_path = self._get_cert_path(self.client_id, True)
```
teoparvanov, line 373: It'd be a bit more explicit to say:
```python
self._get_cert_path(self.client_id, is_client=True)
```

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,256 @@ def get_rsi_replies(self):
...
+        client_pk_path = self._get_private_key_path(self.client_id, True)
```
teoparvanov, line 374: same here

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,256 @@ def get_rsi_replies(self):
...
+        # Create an SSl context - enable CERT_REQUIRED and check_hostname = True
+        ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        # Verify server certificate using this trusted path
+        ssl_context.load_verify_locations(cafile=server_cert_path)
+        # Load my private key and certificate
+        ssl_context.load_cert_chain(client_cert_path, client_pk_path)
+        # Server hostname to be verified must be taken from create_tls_certs.sh
+        server_hostname = self.CERT_DOMAIN_FORMAT % dest_replica.id
+        while True:
+            try:
```
teoparvanov, line 384: to avoid an infinite wait, does it make sense to wrap with:
```python
with trio.fail_after()
```
?

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,256 @@ def get_rsi_replies(self):
...
+        if dest_addr not in self.ssl_streams.keys():
+            # dest_replica is not connected (crushed? isolated?) 
+            return
+        # first 4 bytes include the data header (message size), then comes the data
+        data_len = len(data)
+        out_buff = bytearray(data_len.to_bytes(self.MSG_HEADER_SIZE, "little"))
+        out_buff += bytearray(data)
+        stream = self.ssl_streams[dest_addr]
+        try:
+            await stream.send_all(out_buff)
```
teoparvanov, line 423: don't we need a timeout here?

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,256 @@ def get_rsi_replies(self):
...
+        except (trio.BrokenResourceError, trio.ClosedResourceError):
+            if dest_addr in self.ssl_streams:
+                del self.ssl_streams[dest_addr]
+                await stream.aclose()
+            return False
+
+    async def _receive_from_replica(self, dest_addr, replicas_addr, required_replies, cancel_scope):
+        """
+        Receive from a single replica. 3 stages:
+        1) Wait for the 1st 4 bytes which consist the header (payload length)
```
teoparvanov, line 446: nit - "... which contain the header ..."

teoparvanov (commented):
Hi Gil, I think on first pass it looks very good! I have several comments/questions as part of this review.

PS: I'd ask for an additional review from Andrew on the TLS client specifically, as he wrote the original UDP client.

.github/workflows/build_and_test.yml
```diff
@@ -12,9 +12,10 @@ jobs:
                 - "CONCORD_BFT_CONTAINER_CC=gcc CONCORD_BFT_CONTAINER_CXX=g++"
                 - "CONCORD_BFT_CONTAINER_CC=clang CONCORD_BFT_CONTAINER_CXX=clang++"
             ci_build_type:
-                - "-DCMAKE_BUILD_TYPE=RELEASE -DCI_TEST_STORAGE_TYPE=v1direct"
-                - "-DCMAKE_BUILD_TYPE=RELEASE -DCI_TEST_STORAGE_TYPE=v2merkle"
-                - "-DCMAKE_BUILD_TYPE=DEBUG -DCI_TEST_STORAGE_TYPE=v2merkle"
+                - "-DCMAKE_BUILD_TYPE=RELEASE -DCI_TEST_STORAGE_TYPE=v1direct -DBUILD_COMM_TCP_TLS=FALSE"
+                - "-DCMAKE_BUILD_TYPE=RELEASE -DCI_TEST_STORAGE_TYPE=v2merkle -DBUILD_COMM_TCP_TLS=FALSE"
+                - "-DCMAKE_BUILD_TYPE=DEBUG -DCI_TEST_STORAGE_TYPE=v2merkle -DBUILD_COMM_TCP_TLS=FALSE"
+                - "-DCMAKE_BUILD_TYPE=DEBUG -DCI_TEST_STORAGE_TYPE=v2merkle -DBUILD_COMM_TCP_TLS=TRUE"
```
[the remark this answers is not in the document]
glevkovich: changed already

tests/apollo/util/bft.py
glevkovich: This was a mistake I've already fixed an hour ago.
You are right.
It now looks at the environment to decide.

tests/apollo/util/bft.py
glevkovich: @teoparvanov, The metrics port isn't bft as well?
I came with this name from previous experience - there is always data port and metrics (usually) and some other ports.
The options are:
1) bft_port
2) bft_main_port
3) bft_msg_port

I like the last option!!
What do you think?

tests/apollo/util/bft.py
glevkovich: This is a typo for sure, no idea how it works?
I tried "1 + + 1" on the interpreter, and it works....weird
fixed..

util/pyclient/bft_client.py
glevkovich: @teoparvanov , I don't mind adding this, although we are not really using the full readability options of python, like parameter types etc on function declaration, right?
In that case, at least I want to force user to pass is_client. Will define the function as (star force):
def _get_cert_path(self, replica_id, *, is_client=False):

util/pyclient/bft_client.py
glevkovich: I made it explicit, same as in _get_cert_path. Actually, for now it is always False, but in case (for any reason) in the future we will have other IDs for clients, who might become TLS servers, ITs good to have an already make helper function. So I keep it with this option.

util/pyclient/bft_client.py
glevkovich: There is a caller fail_after block in bft.py . I wrote a comment in on_started_replicas().
I'm still not sure about the right behavior. 
In the trio guide, the say a good implementation should not put fail_after for reusable functions.
My idea is that a good one should give an optional fail after timeput. But I'm not sure if it's possible to nicely implement a conditional fail_after (can also put infinite time if the user doesn't want to fail, but it's ugly ).

util/pyclient/bft_client.py
glevkovich: No.
all the receive and send are covered by large trio scope:
with trio.move_on_after(self.config.retry_timeout_milli / 1000):
this one do retries.

util/pyclient/bft_client.py
glevkovich: "which contains the header" 👍

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,257 @@ def get_rsi_replies(self):
         This method should be called after the send has done and before initiating a new request
         """
         return self.rsi_replies
+
+    def _process_received_msg(self, data, sender, replicas_addr, required_replies, cancel_scope):
+        """Called by child class to process a received message. At this point it's unknown if message is valid"""
+        try:
+            rsi_msg = rsi.MsgWithReplicaSpecificInfo(data, sender)
+        except (bft_msgs.MsgError , struct.error) as ex:
+            self._process_received_msg_err(ex)
```
andrewjstone, line 256: This method seems unnecessary to me, and breaks flow the normal linear flow control of python. The calling code can just catch the exception instead.

As it's not used in either implementation I suggest just removing it from the abstract class and inheriting classes altogether.

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,257 @@ def get_rsi_replies(self):
...
+        """
+        Certificate is under <certificate root path>/replica_id/<node type>/cert.pem,
+        where node type is "server" or "client".
+        """
+        cert_type = "client" if is_client else "server"
+        return os.path.join(self.config.certs_path, str(replica_id), cert_type, cert_type + ".cert")
+
+    async def _establish_ssl_stream(self, dest_replica):
+        """
+        Try every 250ms to connect to dest_replica. We relay on the caller context that will trigger cancalleation at
```
andrewjstone, line 369: typo: `cancellation`

util/pyclient/bft_client.py
andrewjstone: I think leaving it to the caller is the right move, as they say in the trio docs. Let the caller decide how long to wait. That caller can also decide to scope this call inside an outer nursery etc.. 

When I originally wrote the UDP code I added some timeouts, but I think following the trio guidelines makes the most sense. The author has spent years working on this stuff and thinking about the proper abstractions.

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,257 @@ def get_rsi_replies(self):
...
+            # Failed! close the stream and return failure.
+            if dest_addr in self.ssl_streams:
+                del self.ssl_streams[dest_addr]
+                await stream.aclose()
+            return False
+
+    async def _stream_recv_some(self, out_data, dest_addr, stream, num_bytes):
+        """ Try to receive data from replica. On failure - close the SSL stream. """
+        try:
+            out_data += await stream.receive_some(num_bytes)
```
andrewjstone, line 435: If zero bytes are ever returned we have gotten an EOF I believe, and should return false.

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,257 @@ def get_rsi_replies(self):
...
+                if addr in self.ssl_streams:
+                    stream = self.ssl_streams[addr]
+                    del self.ssl_streams[addr]
+                    await stream.aclose()
+
+    async def _send_data(self, data, dest_replica):
+        """ Try to send data to dest_replica. On exception - close the SSL stream. """
+        dest_addr = (dest_replica.ip, dest_replica.port)
+        if dest_addr not in self.ssl_streams.keys():
+            # dest_replica is not connected (crushed? isolated?)
```
andrewjstone, line 415: This is a perfectly normal scenario right? The replica can simply be down or partitioned as you indicate, and the client can't connect. But the higher level software doesn't know this and attempts to send, and so you simply return here. I'm just trying to clarify that we have the same understanding here. 

Additionally, when a partition heals, or a node restarts, we should reconnect from the client. In our C++ client we continuously try to reconnect in the background. We need some strategy for reconnecting.

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,257 @@ def get_rsi_replies(self):
...
+    CERT_DOMAIN_FORMAT="node%dser"
+    # Taken from TlsTCPCommunication.cpp (we prefer hard-code and not to parse the file)
+    MSG_HEADER_SIZE=4
+
+    def __init__(self, config, replicas, bft_network):
+        super().__init__(config, replicas)
+        self.ssl_streams = dict()
+        self.bft_network = bft_network
+
+    async def on_started_replicas(self, replica_ids):
```
andrewjstone, line 340: I'm not sure this callback really makes sense for a general client. It works when we only use the client for testing, but that's not the only use case. This client is a general client that should be usable with a production system. It's certainly not perfect, and the retry strategy can be improved, but I wouldn't want to limit it simply to being test only.

A better way to do this would be to do what we do in the C++ client and take a configuration, and continuously try to reconnect any broken streams in the background when detected. If a stream isn't currently connected, messages get dropped, but we are constantly trying to reconnect.

andrewjstone (changes requested):
Hey @glevkovich, I reviewed the TLS client code as requested. I think the restructuring looks really good and most of my comments are minor. 

The one thing I'd like to change though is tying connection retries to test callbacks. How the client connects should be independent of what's going on in the network replica. it would be great if we could use this client to interact with production systems in the future. With TLS that is now possible.

tests/apollo/util/bft.py
teoparvanov: `bft_msg_port` sounds good!

glevkovich: @teoparvanov ok great. btw, this changes the names of all macros and functions as well (in many places)

util/pyclient/bft_client.py
glevkovich: @andrewjstone I think I agree with you. But there are some problems with each methodology:
Let's say you put your own timeout: The big issue is - if you don't look at the whole code including the caller's code, you can't even know if your own timeout is effective. I've seen some examples where the caller put 5-sec delay, and the inside code puts a 10-sec delay. This can happen if you decide to change your code after the caller wrote its own code.
On the other hand, letting the caller decide might be an issue as well, in complicated code- many times caller might not have any good idea about "what is a good timeout". So the caller must have a very good idea on what's going on inside.

util/pyclient/bft_client.py
```diff
@@ -241,3 +247,257 @@ def get_rsi_replies(self):
...
+        """
+        cert_type = "client" if is_client else "server"
+        return os.path.join(self.config.certs_path, str(replica_id), cert_type, cert_type + ".cert")
+
+    async def _establish_ssl_stream(self, dest_replica):
+        """
+        Try every 250ms to connect to dest_replica. We relay on the caller context that will trigger cancalleation at 
+        some point in time. Break the loop on success and keep the ssl_stream in self.ssl_streams for future use.
+        """
+        server_cert_path = self._get_cert_path(dest_replica.id, False)
```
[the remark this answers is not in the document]
glevkovich: @andrewjstone . I've made it more explicit according to @teoparvanov  comment.
Now the function definition looks like:
`def _get_cert_path(self, replica_id, *, is_client):`
and the call:
`server_cert_path = self._get_cert_path(dest_replica.id, is_client=False)`

util/pyclient/bft_client.py
```diff
@@ -130,8 +160,9 @@ async def sendSync(self, msg, read_only, seq_num=None, cid=None, pre_process=Fal
          Note that this method also binds the socket to an appropriate port if
          not already bound.
         """
-        if not self.sock_bound:
-            await self.bind()
+        # Call an abstract function to allow each client type to set-up its communication before starting
+        if not self.comm_prepared:
+            await self._comm_prepare(read_only)
```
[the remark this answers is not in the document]
glevkovich: Good catch!
Needed that flag before, unneeded and unused now.
removed.

util/pyclient/bft_client.py
glevkovich: @andrewjstone 
Catching the exception inside won't help since the handling is different for each type of client, so handling it must be done by the client code.
UDP client re-raise the exception inside _process_received_msg_err.
TLS client ignores it inside _process_received_msg_err.
There are some other solutions for that(like passing a flag inside MsgWithReplicaSpecificInfo), but all looked less favorable to me.

util/pyclient/bft_client.py
glevkovich: @andrewjstone 
Thanks, I would add:
I can expect only 2 values to be returned:
1) Header size self.MSG_HEADER_SIZE - 4 bytes
2) Payload size
So I think I'll make it more strict - if len(out_data) < num_bytes just return False, not just for zero.

util/pyclient/bft_client.py
glevkovich: @andrewjstone (adding @teoparvanov here)- I'll try to shortly make things clearer here:
Yes - It's a normal scenario I agree.
At the beginning of the test, always before trying to send, when the replica is started, all clients connect to it. clients are always starting at the beginning (I'm not dealing with external clients for now). So if a replica is isolated or crushed the TCP socket stays in a special state (TIME_WAIT) and not yet closed. This is a 240-sec timeout, so no need to reconnect. If replica returns before that timeout, send will succeed since the socket is still open on both sides of no one explicitly closed it in the application layer. Every time when starting and stopping a replica, all clients are connected (TCP and SSL. handshake) and disconnected respectively.
My first implementation did exactly what you are suggesting - it did a lazy connection whenever trying to send data and connections is closed. This causes many complications, especially with the network partition manager, as the ports given to the client are unknown ports, and every time it happens, the connectivity port matrix in advisory needs to be refreshed. In addition, the client is unaware of who is healed, so it retries to reconnect all the time to all disconnected servers, and that causes a "major shake" in all tests - as it slows down all our estimated timeouts and multiple sends (tracker), and fails many of the tests.

As of now, In all our tests we get isolation/disconnection by:
1) Stopping and starting later some replica.
2) Isolating a replica for less than 4 minutes.
So all our tests pass.

If we want to isolate a replica for more than 4 minutes - it will require adding re-connect. But this is more complicated, especially to write notifications between the network partition manager and bft_client.

To summarize - I am aware of this, but this should be part of the 2nd stage impl., and not 1st stage (we should add a task for it already now)

util/pyclient/bft_client.py
glevkovich: @andrewjstone / @teoparvanov 
I think it related to your previous comment.
Trying to reconnect all the time slows down the test.
Just to understand the number of connections we have in a single test:
32 clients (30 regular and 2 reserved) with 7 replicas ~ 500 connections (SSL handshake) on a single machine every time a network is going up. I've tried to reduce the number of clients, but some tests failed (not adaptable to client number, making assumptions on the number of sends in a time interval, and expect moving from optimizing the fast path to slow path) so for now, returned to 30 clients.
Each replica that goes down triggers all client that is sending to get busy with reconnecting, and the send command would not end until a connection is failed (if the replica is still isolated/down).

THE MAJOR with your suggestion my opinion is that the current test framework implementation does not support this kind of implementation:
We are using blocking I/O on a single thread to deal with everything and avoid multi-threading.
To do what you suggest requires (at least) 1 thread per client that tries to reconnect the replica all the time in the background. for now, we can try to reconnect only during sending, and as I've said that causes many issues in test timeouts.

glevkovich:
@andrewjstone , @teoparvanov:
I would like to summarize my thought about these 2 comments:
https://github.com/vmware/concord-bft/pull/956#discussion_r508901240
https://github.com/vmware/concord-bft/pull/956#discussion_r508902016

Regarding reconnection which @andrewjstone pointer about:
The only fast (but complicated) solution I can think of, for now, is adding in both bft_client.read and bft_client_write calls reconnect attempts at the beginning of both functions.
The reconnection attempts should be shielded (that means, cannot be canceled by outer scopes) and later on add to the outer cancellation scopes the time that passed in the reconnection process.

To read more about shielded scopes:
https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-api-details

I'm not sure how accurate I am here, since I haven't read much about shielded scopes, but that might be the only way for now without major changes.
Anyhow, this should be part of later stages Impl.

util/pyclient/bft_client.py
andrewjstone: I don't think that's how this call works, nor how TCP streams work in general. The stream can return any number of bytes > 0 and <=4 and still be correct. If it returns exactly 0 the stream has reached an EOF indicating closure from the other side. But due to windows, delays, etc.. it's possible this call returns with say, only 2 bytes. Then the remaining 2 bytes need to be read in another call to get the complete header size. 

0 is a sentinel meaning something is broken. Returning less than max bytes (4 in this case) is normal behavior.

util/pyclient/bft_client.py
andrewjstone: I believe you are missing my point. This call: `_process_received_msg` is called by the concrete implementations, and so the UDP code or TLS code that calls this function can handle the exception directly and then do the same thing they would do in the callback. The callback is just an extra level of indirection. 

For example, the TLS callback swallows the exception , as you point out, but it could just as easily do that by wrapping the call to `_process_received_msg` inside `recv_data` with a `try` and a `pass` in the `except` clause.

In the UDP client case, it's even easier. You don't have to change the calling code at all since the exception is re-raised in the callback. Instead of forcing the UDP code to implement this callback and re-raise an exception, you just let the exception flow normally through the call chain by not catching it in the first place. 

Introducing this method just means that some clients have to do more work than they would otherwise, and you further complicate this method by adding the exception catch and then calling the callback.

util/pyclient/bft_client.py
glevkovich: You are right that the stream can return any amount of bytes up to num_bytes, including 0. I've fixed it the wrong way.
But please pay attention that there is already an outer loop in the caller which calls again.

util/pyclient/bft_client.py
glevkovich: I'll explain my logic when writing this way:
_process_received_msg does a lot of things.
When I wrap an except around it, how can I be sure I got it in a specific place like in the current implementation?
I agree it simplifies the code, but It looks riskier to me.. 
I wanted to be sure the exception happens inside rsi.MsgWithReplicaSpecificInfo and not anywhere else. Do you suggest to check the exception location?

andrewjstone: Determining why the error occurred is what the exception type is for. This is similar to a return value in more procedural or functional languages. Taking your existing strategy to it's logical conclusion would require a separate callback for each line that could possibly fail.

glevkovich: Ok, I've made the changes as you suggested. Thanks!

util/pyclient/bft_client.py
```diff
@@ -241,3 +230,258 @@ def get_rsi_replies(self):
...
+    async def __aexit__(self, *args):
+        """async context manager method for 'with' statements"""
+        self.exit_flag = True
+        for lot in self.establish_ssl_stream_parklot.values():
+            lot.unpark()
+        remove = [ k for k in self.ssl_streams ]
+        for k in remove:
+            stream = self.ssl_streams[k]
+            del self.ssl_streams[k]
+            await stream.aclose()
```
teoparvanov, line 472: instead of closing those sequentially, can we do this concurrently in a nursery?

teoparvanov (approved):
Awesome work, thank you @glevkovich !

util/pyclient/bft_client.py
```diff
@@ -241,3 +230,258 @@ def get_rsi_replies(self):
...
+                self.rsi_replies = self.replies_manager.get_rsi_replies(rsi_msg.get_matched_reply_key())
+                self.primary = self.replicas[header.primary_id]
+                cancel_scope.cancel()
+
+class UdpClient(BftClient):
+    """
+    Define a UDP client - sends and receive all data via a single port
+    (connectionless / stateless datagram communication)
+    """
+    def __init__(self, config, replicas, background_nursery):
```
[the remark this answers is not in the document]
glevkovich: @teoparvanov - If we do that I'll have to add a condition at all callers (few testa call this constructor), and we miss the whole idea of a polymorphic constructor. Also, I don't think it's a bad idea to have a preparation for something we have. good chance to use in the future - this is a general-purpose object which allows calling for background tasks.

util/pyclient/bft_client.py
```diff
@@ -241,3 +230,258 @@ def get_rsi_replies(self):
...
+    MSG_HEADER_SIZE=4
+
+    def __init__(self, config, replicas, background_nursery):
+        super().__init__(config, replicas)
+        self.ssl_streams = dict()
+        self.reconnect_nursery = background_nursery
+        self.exit_flag = False
+        self.establish_ssl_stream_parklot = dict()
+        for replica in replicas:
+            # For each thread, start a background task to keep connection established
```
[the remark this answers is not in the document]
glevkovich: That is a mistake of course, will change to "For each TCP destination replica" - it is not yet connected.

util/pyclient/bft_client.py
glevkovich: Good idea. I will.

util/pyclient/bft_client.py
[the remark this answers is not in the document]
teoparvanov: OK, just wanted to understand the reasoning behind it. Thanks!

tests/apollo/test_skvbc.py
```diff
@@ -71,7 +71,7 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
             checkpoints_num=3, # key-exchange channges the last executed seqnum
             persistency_enabled=False
         )
-        bft_network.start_replica(stale_node)
+        await bft_network.start_replica(stale_node)
```
teoparvanov, line 74: Hi @glevkovich I think with your latest changes, start/stop replica don't need to be async anymore.
Please correct me if I'm wrong.

teoparvanov (changes requested):

tests/apollo/test_skvbc.py
glevkovich: Yes, I will change it back.

PR #958, opened 2020-10-19
Serialization enhancements:

toly-kournik:
- add support for std::map, std::unordered_map
- add support for const std::string
- add syntactic sugar to type_traits as acceptible since c++17 (e.g. is_vector_v)
- minor cleanup for several serializable classes, which are not serialized anymore.


cloudnoize (approved):
LGTM

PR #960, opened 2020-10-20
[Apollo] Wait more for source replica replacement during ST

teoparvanov:
In a recent PR, the source replica replacement timeout was increased to 15s:
https://github.com/vmware/concord-bft/pull/950

This seems to cause an occasional failure of the Apollo test which covers this path. In this PR we increase the time Apollo waits for the ST source replica to be replaced.


yuliasherman (approved):
LGTM

PR #968, opened 2020-10-21
[Apollo] Reduce pre-processor log level for SKVBC replicas

teoparvanov:
This went in by mistake, following some local Apollo pre-execution test investigations... It may be why we are observing "disk space" issues in GitHub Actions.


yuliasherman (approved):
LGTM

PR #971, opened 2020-10-21
TlsTCPCommunication: Degrade log message

robem:
This function is called frequently and the common case is that we already have
a connection with that node. If we don't then there are two other log messages
and one will be printed for sure. Meaning, I don't think we loose any
information at runtime by degrading this message to DEBUG.


robem:
This function is being refactored in #949 and depending on how fast we can get it in this PR can be closed.

andrewjstone (approved):
LGTM. Thanks!

PR #973, opened 2020-10-21
Check if the MetricsServer thread is joinable before join

tdimitrov:
The Stop() function of the MetricsServer calls join() on the processing
thread however there are cases when the thread is not joinable.

This causes std::terminate() to be called and breaks the cleanup process
of concord-bft library.

### DIFF
--- a/util/src/MetricsServer.cpp
+++ b/util/src/MetricsServer.cpp
@@ -67,7 +67,7 @@ void Server::Stop() {
   close(sock_);
 
   // Wait for the recvLoop thread to stop
-  thread_.join();
+  if (thread_.joinable()) thread_.join();
 }
 
 void Server::RecvLoop() {


HristoStaykov (approved):
LGTM, thanks.

tdimitrov:
> @tdimitrov , isn't [_running](https://github.com/vmware/concord-bft/blob/master/util/include/MetricsServer.hpp#L63) flag supposed to represent if the thread is running?

I understand your idea and you have a point. No need to try to close the socket again.
I just realised we also have to keep the mutex locked until we initialise the thread in Start(). Otherwise there is a small chance for a race condition.

I'll update the PR tomorrow with these changes.

tdimitrov:
Looking again at the code I think the joinable check is mandatory.

Stop can be:
`void Server::Stop() {
  {
    std::scoped_lock lock(running_lock_);
    if (!running_) {
      //nothing to do
      return;
    }

    // This will cause `recvfrom` to error in `RecvLoop` and therefore allow it
    // to check for running_ = false without requiring a timeout on the socket.
    close(sock_);
  } // lock scope
  
  // Wait for the recvLoop thread to stop
  // Release the lock so it can finish its job
  // Check if the thread is joinable() in case Stop() is called again while we wait on the join().
  // This is highly inlikely but possible and will cause a crash.
  if (thread_.joinable()) {
    **thread_.join();**
  }

  {
    std::scoped_lock lock(running_lock_);
    running_ = false;
  }
}`

And a slight modification in Start():
`void Server::Start() {
  std::scoped_lock lock(running_lock_);

  if (running_) {
    // don't run twice
    return;
  }

  if ((sock_ = socket(AF_INET, SOCK_DGRAM, 0)) < 0) {

// the rest is the same`

If we decide to count on the running_ flag as an indication whether the Server is working or not, we have to protect it a bit more:
1. Hold it during the whole lifetime of Start() due to the swap in the end of the function.
2. Hold it during the whole life time of Stop() but release it while calling join() so that the RecvLoop() can finish.

One way or another we have to keep the joinable check because Stop() can be called twice and at one point both calls will try to join and the second one will fail.

If you ask me - I'd stick with the initial version because it's simpler and will put link to this discussion in the code. But I don't insist on this. 

@f-squirrel @andrewjstone  what do you think?

f-squirrel (approved):
@tdimitrov , I think that the new version looks a bit more logic but comes with an additional complexity.
I am fine with sticking to the original one.

andrewjstone (approved):
Thanks for the fix and explanation.

PR #975, opened 2020-10-21
Add msg parameter to do_with_retry() in s3 client

tdimitrov:
The current implementation of do_with_retry() prints the function name
with typeid(f).name(). However when the function is used with std::bind
the output is unreadable for a human.

This patch adds another msg parameter to do_with_retry() which is a
string used in the log message.

### DIFF
--- a/storage/include/s3/client.hpp
+++ b/storage/include/s3/client.hpp
@@ -106,7 +106,7 @@ class Client : public concord::storage::IDBClient {
     concordUtils::Status res = concordUtils::Status::OK();
     std::function<Status(const concordUtils::Sliver&, OUT concordUtils::Sliver&)> f =
         std::bind(&Client::get_internal, this, _1, _2);
-    do_with_retry(res, f, _key, _outValue);
+    do_with_retry("get_internal", res, f, _key, _outValue);
     return res;
   }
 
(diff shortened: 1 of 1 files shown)


storage/include/s3/client.hpp
```diff
@@ -204,13 +204,13 @@ class Client : public concord::storage::IDBClient {
  protected:
   // retry forever, increasing the waiting timeout until it reaches the defined maximum
   template <typename F, typename... Args>
-  void do_with_retry(Status& r, F&& f, Args&&... args) const {
+  void do_with_retry(const std::string& msg, Status& r, F&& f, Args&&... args) const {
```
dartdart26, line 207: Nit: Maybe we can consider changing the parameter to `std::string_view msg` to avoid copying and potential allocation from string literals.

dartdart26 (approved):
Looks good, thanks!

storage/include/s3/client.hpp
tdimitrov: Sorry I missed this comment. Will open a new PR.

tdimitrov: https://github.com/vmware/concord-bft/pull/978

PR #977, opened 2020-10-22
Fix call to finishAddingRequests for the prePrepare message

yuliasherman:
The finishAddingRequests function should be called for all PrePrepare messages - batched or not.


f-squirrel (approved):
LGTM!

PR #982, opened 2020-10-25
Ignore PreProcessRequestMsg during state transfer

yuliasherman:
PreProcessRequestMsg should be ignored by a replica during a state transfer exactly like ClientPreProcessReqMsg.


teoparvanov (approved):
Makes sense, thanks Yulia!

PR #984, opened 2020-10-25
Makefile: Add an option to list all tests

glevkovich:
Before building a target or before running a single test, user should be
able to list all tests. Until now user had to log into a container and
manually call `cd build && ctest -N`.
User should be able to do this from host.

Changes:
* Added `make list-tests` that prints out all the tests.

Usage:
* Run `make list-tests` to print a list of all tests.


f-squirrel (approved):
Awesome, thank you!

PR #986, opened 2020-10-25
Add unique identity to the PreProcess messages to match between requests and replies including client retries

yuliasherman:
This PR provides a way to match between a PreProcess request and corresponding replies including client retries when both - cid and reqSeqNum are the same. New fields are added to the PreProcessRequest/PreProcessReply messages. They are set to the same number for a request and all its replies. The retry message gets a different value. This approach allows reporting about non-determinism in the correct way and not to mix old replies with a retried request.

### DIFF
--- a/bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
+++ b/bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
@@ -22,12 +22,10 @@ using namespace bftEngine;
 // maxReplyMsgSize_ = sizeof(Header) + sizeof(signature) + cid.size(), i.e 58 + 256 + up to 710 bytes of cid
 uint16_t PreProcessReplyMsg::maxReplyMsgSize_ = 1024;
 
-PreProcessReplyMsg::PreProcessReplyMsg(SigManagerSharedPtr sigManager,
-                                       NodeIdType senderId,
-                                       uint16_t clientId,
-                                       uint64_t reqSeqNum)
+PreProcessReplyMsg::PreProcessReplyMsg(
+    SigManagerSharedPtr sigManager, NodeIdType senderId, uint16_t clientId, uint64_t reqSeqNum, uint64_t reqRetryId)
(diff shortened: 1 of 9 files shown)


teoparvanov (approved):
LGTM, thanks Yulia!

bftengine/src/preprocessor/messages/PreProcessRequestMsg.hpp
```diff
@@ -23,6 +23,7 @@ class PreProcessRequestMsg : public MessageBase {
   PreProcessRequestMsg(NodeIdType senderId,
                        uint16_t clientId,
                        uint64_t reqSeqNum,
+                       uint64_t reqVersion,
```
teoparvanov, line 26: How about `preProcessRequestId` for example ?

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
```diff
@@ -22,12 +22,10 @@ using namespace bftEngine;
 // maxReplyMsgSize_ = sizeof(Header) + sizeof(signature) + cid.size(), i.e 58 + 256 + up to 710 bytes of cid
 uint16_t PreProcessReplyMsg::maxReplyMsgSize_ = 1024;
 
-PreProcessReplyMsg::PreProcessReplyMsg(SigManagerSharedPtr sigManager,
-                                       NodeIdType senderId,
-                                       uint16_t clientId,
-                                       uint64_t reqSeqNum)
+PreProcessReplyMsg::PreProcessReplyMsg(
+    SigManagerSharedPtr sigManager, NodeIdType senderId, uint16_t clientId, uint64_t reqSeqNum, uint64_t replyVersion)
```
teoparvanov, line 26: Same here - `preProcessReplyId` ?

teoparvanov (approved):
LGTM, thanks Yulia! 

PS: I just have a minor comment about naming. Up to you to decide which one makes more sense.

PR #987, opened 2020-10-25
bft_client: catch OverflowError on TLS stream.receive_some()

glevkovich:
Due to an exception:
"OverflowError: signed integer is greater than maximum".

I've found out that if something is wrong with payload size (too big),
client code is not resilient to an uncaught exception.
The trio documentation does not mention any limit.
The ssl.py module supports up to 2^31. That means, we gave a very very
large number.

Resolution: Catch exception, and unpark internal re-connect task to
close the stream and re-connect.


f-squirrel (approved):
LGTM

PR #988, opened 2020-10-26
Remove timeToWaitBeforeStartingStateTransferInMainWindowMilli to avoi…

toly-kournik:
…d unnecessary state transfers

Leave a warning though to prove the thesis.


andrewjstone (commented):
Can you please describe what the hypothesis is that you are making, and why you are making this change?

PR #989, opened 2020-10-26
bft: Make changes in BftTestNetwork.existing() to support hermes tests

glevkovich:
Recent changes did not take into account the need to run Hermes tests
with the new changes. To allow bumping concord-bft into concord and
run Hermes tests with the new changes, we must add these minor changes.


teoparvanov (approved):
LGTM!

PR #990, opened 2020-10-27
take execution in account on vc

yontyon:
In order to stabilize the viewchange action, we want to take in account the replica progression that is not related to the earliest client request timeout. 

For that, we have added another check in the viewchange timer such that if the time from the last execution is less than the viewchange timeout while the time from the earliest pending client request is larger, we double the view change timeout.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2767,15 +2767,31 @@ void ReplicaImp::onViewsChangeTimer(Timers::Handle timer)  // TODO(GG): review/u
 
     if (!hasPendingRequest) return;
 
-    const uint64_t diffMilli1 = duration_cast<milliseconds>(currTime - timeOfLastStateSynch).count();
-    const uint64_t diffMilli2 = duration_cast<milliseconds>(currTime - timeOfLastViewEntrance).count();
-    const uint64_t diffMilli3 = duration_cast<milliseconds>(currTime - timeOfEarliestPendingRequest).count();
+    const uint64_t timeFromLastStateTransfer = duration_cast<milliseconds>(currTime - timeOfLastStateSynch).count();
+    const uint64_t timeFromLastViewEntrance = duration_cast<milliseconds>(currTime - timeOfLastViewEntrance).count();
+    const uint64_t timeFromEarliestPendingRequest =
+        duration_cast<milliseconds>(currTime - timeOfEarliestPendingRequest).count();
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2767,15 +2767,31 @@ void ReplicaImp::onViewsChangeTimer(Timers::Handle timer)  // TODO(GG): review/u
...
+    const uint64_t timeFromEarliestPendingRequest =
+        duration_cast<milliseconds>(currTime - timeOfEarliestPendingRequest).count();
+    const uint64_t timeFromLastExecution = duration_cast<milliseconds>(currTime - timeOfLastExecution).count();
+
+    if (timeFromLastExecution <= viewChangeTimeout && timeFromEarliestPendingRequest > viewChangeTimeout) {
+      // If the replica is progressing, but yet client messages are timed out, it means that the replica is busy with
+      // other requests, thus we double the viewchange timeout to reduce the possibility of unnecessary viewchange.
+      LOG_INFO(VC_LOG,
+               "time from eraliest pending request is greater than viewChangeTimeout but time from last execution is "
+               "not. This may indicate that the replcia is struggling to execute requests amd there is no need to "
```
HristoStaykov, line 2781: minor note: typo "amd"->"and"

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2767,15 +2767,31 @@ void ReplicaImp::onViewsChangeTimer(Timers::Handle timer)  // TODO(GG): review/u
...
-      LOG_INFO(
-          VC_LOG,
-          "Ask to leave view=" << curView << " (" << diffMilli3 << " ms after the earliest pending client request).");
-
+    const uint64_t timeFromLastStateTransfer = duration_cast<milliseconds>(currTime - timeOfLastStateSynch).count();
+    const uint64_t timeFromLastViewEntrance = duration_cast<milliseconds>(currTime - timeOfLastViewEntrance).count();
+    const uint64_t timeFromEarliestPendingRequest =
+        duration_cast<milliseconds>(currTime - timeOfEarliestPendingRequest).count();
+    const uint64_t timeFromLastExecution = duration_cast<milliseconds>(currTime - timeOfLastExecution).count();
+
```
HristoStaykov, line 2775: Just a proposal, we might log if we have a Client operation beyond half of the VC timeout, but less than the VC timeout just as an early indication that the Replica is getting behind.

HristoStaykov (approved):
Great job, @yontyon, just 2 minor comments.

PR #991, opened 2020-10-27
bcstatetransfer_tests: Remove local database when test ends

glevkovich:
When trying to run this test more than once (native Linux or the same
docker container), It always fails.

Resolution: When the test ends, on Teardown, remove the whole folder
`./bcst_db`.


andrewjstone (approved):
LGTM. 
One thing that may be worthwhile is to also try to delete it on startup, in case the test crashed prior without running terminate.

glevkovich:
OK, I will add clean on start as well. That would be good for a developer.

PR #992, opened 2020-10-27
State Transfer tweaks:

toly-kournik:
1. Make max chunk size equal to max block size if using TCP;
2. If inside active window try first to catch up using request for missing data;
3. Clear all pending requests before call to startCollecting state to avoid entering to a new view on request timeout.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -673,7 +673,7 @@ void ReplicaImp::tryToAskForMissingInfo() {
   SeqNum maxSeqNum = 0;
 
   if (!recentViewChange) {
-    const int16_t searchWindow = 4;  // TODO(GG): TBD - read from configuration
+    const int16_t searchWindow = 32;  // TODO(GG): TBD - read from configuration
     minSeqNum = lastExecutedSeqNum + 1;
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2415,8 +2419,8 @@ void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
   }
 
   if (askAnotherStateTransfer) {
```
yontyon, line 2421: As you ask for state transfer also when tryToAskForMissingInfo return false because isCollectingState is true, I think we should also add here that if isCollectingState is true, we won't issue state transfer

yontyon (approved):
I have only a small comment that I think we should fix before merging it.

andrewjstone:
Can you please add details to the commit message about *why* those specific changes are being made?

bftengine/src/bftengine/ReplicaImp.cpp
toly-kournik: we cannot be in collecting state here

toly-kournik:
> Can you please add details to the commit message about why the block size change for tcp ?
To quicker transfer the entire block without dividing it into chunks

PR #995, opened 2020-10-28
[WIP, Proposal] Support a client for partitioned DBs

dartdart26:
Extend the IDBClient and ITransaction interfaces by providing support
for the notion of `partitions`. A `partition` can be thought of as a key
namespace. Two or more partitions can contain the same key and its value
can be different across them.

In terms of design, the IPartitionedDBClient and the
IPartitionedTransaction interfaces are introduced by extending the
IDBClient and the ITransaction interfaces, respectively. The new
interfaces just add another parameter to the key-value methods, namely
the `partition`. An attempt is made to not change the interfaces
significantly. Moreover, methods for partition management are supported,
i.e. looking up, adding, dropping.

In addition, introduce the PartitionDBClient and PartitionTransaction
classes that bridge partition-based interfaces to the existing IDBClient
one. That way, a partition can be used as drop-in replacement for
existing partition-unaware IDBClient implementations.

Remove the IDBClient::freeIterator() method and make the
IDBClient::getIterator() one return an std::unique_ptr. Remove the
corresponding Guard.

Introduce the IDBClient::startTransaction() method that is an alias to
the IDBClient::beginTransaction() method, but returning an
std::unique_ptr. Support upgrade from the std::unique_ptr to an
ITransaction::Guard in order to automatically commit the transaction.
Applies to IPartitionedTransaction too.

Rename IDBClientITerator::isEnd() to valid() and invert its meaning.
That allows for graceful handling of the case of calling previous() on
an iterator that points to the first element. Fix const-correcness for
the valid() and getStatus() methods.

Move common GTest storage code from `kvbc` to `storage`. Add the
`db_partitions` unit test.

This PR assumes that the underlying store can meaningfully support
partitions. For example, partitions map to column families when RocksDB
is used.

A sample implementation is provided in the existing `memorydb` client.

Finally, this PR is just a proposal and a placeholder for discussions.
Suggestions for changes or totally abandoning the idea are welcome. If
accepted, a RocksDB implementation will be added in a future commit.

### DIFF
--- a/bftengine/src/bcstatetransfer/DBDataStore.cpp
+++ b/bftengine/src/bcstatetransfer/DBDataStore.cpp
@@ -237,9 +237,9 @@ void DBDataStore::deserializeResPage(
  * [pageId, checkpoint] => <serialized page>
  */
 void DBDataStore::loadResPages() {
-  auto it(dbc_->getIteratorGuard());
+  auto it = dbc_->getIterator();
   for (auto keyValue = it->seekAtLeast(dynamicResPageKey(0, 0));  // start of dynamic keys space
-       !it->isEnd() && keyValue.first.string_view().find(keymanip_->getReservedPageKeyPrefix().string_view()) == 0;
+       it->valid() && keyValue.first.string_view().find(keymanip_->getReservedPageKeyPrefix().string_view()) == 0;
        keyValue = it->next()) {
     std::istringstream iss(
         std::string(reinterpret_cast<const char*>(keyValue.second.data()), keyValue.second.length()));
--- a/bftengine/src/bcstatetransfer/DataStore.hpp
+++ b/bftengine/src/bcstatetransfer/DataStore.hpp
@@ -181,7 +181,7 @@ class DataStoreTransaction : public DataStore, public ITransaction {
   class Guard : public ITransaction::Guard {
    public:
     Guard(DataStoreTransaction* t) : ITransaction::Guard(t) {}
-    DataStoreTransaction* txn() { return static_cast<DataStoreTransaction*>(txn_); }
+    DataStoreTransaction* txn() { return static_cast<DataStoreTransaction*>(ITransaction::Guard::txn()); }
   };
 
   DataStoreTransaction(std::shared_ptr<DataStore> ds, ITransaction* txn)
--- a/kvbc/src/direct_kv_db_adapter.cpp
+++ b/kvbc/src/direct_kv_db_adapter.cpp
@@ -480,10 +480,10 @@ void DBAdapter::deleteLastReachableBlock() {
  */
 std::pair<Value, BlockId> DBAdapter::getValue(const Key &key, const BlockId &blockVersion) const {
   LOG_TRACE(logger_, "Getting value of key " << key << " for read version " << blockVersion);
-  auto iter = db_->getIteratorGuard();
+  auto iter = db_->getIterator();
   Key searchKey = keyGen_->dataKey(key, blockVersion);
   KeyValuePair p = iter->seekAtLeast(searchKey);
-  if (!iter->isEnd() && DBKeyManipulator::extractTypeFromKey(p.first) == EDBKeyType::E_DB_KEY_TYPE_KEY) {
+  if (iter->valid() && DBKeyManipulator::extractTypeFromKey(p.first) == EDBKeyType::E_DB_KEY_TYPE_KEY) {
     Key foundKey = DBKeyManipulator::composedToSimple(p).first;
     BlockId currentReadVersion = DBKeyManipulator::extractBlockIdFromKey(p.first);
     // TODO(JGC): Ask about reason for version comparison logic
@@ -537,7 +537,7 @@ BlockId DBAdapter::fetchLatestBlockId() const {
 
   // Generate maximal key for type 'block'.
   const auto maxKey = keyGen_->blockKey(std::numeric_limits<BlockId>::max());
-  auto iter = db_->getIteratorGuard();
+  auto iter = db_->getIterator();
   // Since we use the maximal key, SeekAtLeast will take the iterator
   // to one position beyond the key corresponding to the largest block id.
   iter->seekAtLeast(maxKey);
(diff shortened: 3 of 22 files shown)


andrewjstone (commented):
Hey Petar, I finally took a look at your partitioned DB code. It looks really good. I have no
problem with the code at all, and I like all the fixes you made as well regarding smart pointers.
It may make sense to do things this way. I’m not sure. I have two concerns, and I haven’t really
thought much about them beyond what I’m writing here.

 1. Every expansion to IDB client makes it harder to switch DBs later. This change in particular seems to tie us pretty heavily to rocksDB.
 2. How useful is the IDBClient abstraction in the first place?

Both of these come down to essentially the same thing: Using multiple smaller, higher level
interfaces may make more sense long term. For example, our public storage interfaces:
IBlocksAppender, etc… don’t require transactions. Most of our code doesn’t use transactions in
fact. Multi-put is used a lot, but can be implemented in a variety of ways. This also ties back
to my problem with reusing IDBClient for the object store. It makes very little sense to support
interfaces for underlying constructs that can’t support the functionality in those interfaces.

I have a gut feeling that that some of the “partitions” we want to create won’t end up being
backed by our main database at all, and we will end up with 2 different layers of partitions.
Some partitions at the higher level may live in files or use some other mechanism. Maybe this is ok, and there is no conflict, as it's just different layering. Maybe we can also polyfill the partition abstraction with key-prefixes for IDBClient implementers that don't support the notion. 

I'm not sure how relevant my concerns are at this point, but I think we should discuss them before we merge this in. 

Note that these concerns are purely at an abstraction level. Practically speaking, I don't believe we are likely to switch from RocksDB anytime soon. I also think it is extremely valuable to have the capability of using rocksdb column families. That provides us with the ability to do a variety of performance and consistency tuning based on the needs of distinct data sets. Lastly, using such families will also require an ability to configure them independently. We probably don't want that in the interface as it will be distinct across underlying databases, and can be done in the constructor. However, configuration should be considered.

dartdart26:
Thank you for the review @andrewjstone ! Great points!

I agree with everything you say. Let me share my thinking.

IDBClient is already heavily tied to RocksDB (or maybe some ordered KV store), because it assumes:

- key ordering
- provides an interface for key comparison
- supports iteration
- it supports transactions unconditionally

As to whether IDBClient actually allows for changing the underlying DB, I am not really sure, because of the mentioned points. There aren't many options that support all of above. Currently, I think IDBClient is useful for testing, i.e. memorydb almost implements it (without concurrent transactions). Maybe we can rename IDBClient to **IOrderedKVStore**.

In terms of partitions, this PR doesn't try to provide an abstraction for separate stores, i.e. files, etc. It focuses on a single key-value DB that can support partitions. Even if it doesn't, as you say, they can be implemented internally with prefixing. We already do that in our kvbc codebase and we can use the partition abstraction so that users (kvbc) don't need to deal with it.

For different store types, I don't think we should use the same IDBClient abstraction. Maybe these different stores are special (i.e. files) and we don't need abstractions for them. Maybe @ittaiab has an opinion here too as we've discussed that before offline.

Regarding our public higher-level interfaces (IBlocksAppender, etc.), I don't see them being directly dependent on the storage layer in Concord-BFT as we only expose functionality that is provided by kvbc. What I mean is that we can change the storage library (underlying store, interfaces, etc.) and still provide the same service over the public higher-level interfaces. In general, I see kvbc as a client to both IDBClient and other notions (i.e. files) to provide a key-value blockchain.

If we want to go the route of abstraction in the storage library, maybe we need to separate interfaces to something like KV store, ordered KV store, transaction KV store and possibly other properties that, for example, file stores can provide. Personally, I am not sure that makes sense (it might, just can't predict).

Bottom line is I think that IDBClient (or IOrderedKVStore) shouldn't aim at being universal in any way. It provides an interface and if we find another notion, we can define it separately and leave IDBClient as is. Also, Concord-BFT is more about a replicated key-value blockchain and not so much about generic DB stores - kvbc can handle the complexities of using different stores without explicit abstractions.

toly-kournik:
> IDBClient is already heavily tied to RocksDB (or maybe some ordered KV store), because it assumes:
> 
> * key ordering
I believe exists in any KV db.

> * provides an interface for key comparison 
For v1direct only. After switching to natural ordering we can delete it

> * supports iteration
basically supported (or may be implemented for ) by most DBs I guess 

> * it supports transactions unconditionally
What does it mean?

Let's take an S3 example
`concord::storage::s3::Client` was a good example of why we need some form of `IDBClient` interface.
As for support of points above, it implements only a subset of `IDBClient`, for instance iterators were not implemented because we didn't need it.

dartdart26:
> > IDBClient is already heavily tied to RocksDB (or maybe some ordered KV store), because it assumes:
> > 
> > * key ordering
> >   I believe exists in any KV db.
> 
> > * provides an interface for key comparison
> >   For v1direct only. After switching to natural ordering we can delete it
> 
> > * supports iteration
> >   basically supported (or may be implemented for ) by most DBs I guess
> 
> > * it supports transactions unconditionally
> >   What does it mean?
> 
> Let's take an S3 example
> `concord::storage::s3::Client` was a good example of why we need some form of `IDBClient` interface.
> As for support of points above, it implements only a subset of `IDBClient`, for instance iterators were not implemented because we didn't need it.

@toly-kournik Thanks for the comments!

Firstly, I am no S3 expert, please correct me if I am wrong. I agree that S3 can be used over IDBClient, but only partially. And it goes both ways - some of the IDBClient methods cannot be implemented and some of the S3 features cannot be expressed. I think the reason is that the Object Store and Key-Value Store notions differ. I am focusing on S3 below:

1. AFAIK, key-values in S3 are not ordered. This makes the `seek` methods in `IDBClientIterator` unusable. In fact, I think it makes iteration meaningless. I think that also applies to any range methods. What S3 offers instead is `listing` keys by prefix and that implies fetching them in memory. With IDBClient iterators, on the other hand, one can iterate one at a time.
2. IDBClient expects that the underlying store supports transactions. AFAIK, S3 doesn't support that. Instead, the doc says `There is no way to make atomic updates across keys.`: https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.htl#ConsistencyModel
3. Object Stores (S3 included) support metadata, i.e. a key has a value + metadata. Currently, we cannot expose that over IDBClient. We might be able to do so, but not sure if it would make sense for Key-Value Stores.

What I am saying is that the user wouldn't be able to properly use an IDBClient if it points to an S3 instance - only partially. And if it is partially, then that implies checks for the type and that raises questions whether this interface is useful - might as well use the concrete notion.

Having said above, I realise partitions might also not be generic. The focus in this PR is partitions in a single DB. However, even if not natively supported, they can be implemented on top of a single Key-Value Store via prefixing (as already mentioned by @andrewjstone ).

Another thing that you mentioned offline is that maybe **partition** is not a good name. I think that is true, because we support the same keys across them and, if I am not mistaken, a "partition" doesn't imply that.

In general, I think the focus should be on the kbvc sub-library - we can make it operate with multiple different notions, i.e. the ones we use to get the best performance and functionality, such as **IOrderedKVStore(IDBClient)**, **IObjectStore**,  **MemoryMapedFile** or **AnySpecialStore**.

dartdart26:
Decided to go with a concrete RocksDB implementation for Column Families as in https://github.com/vmware/concord-bft/pull/1045 .

PR #996, opened 2020-10-29
Added additional storage to keep information for Sequence Numbers

HristoStaykov:
for up to N previous working windows. This is an optimisation to
help in the situation where a Replica is catching up and is being
slightly behind the others. If in this case the other Replicas
agree on a stable Checkpoint further than the sequence ID the
behind Replica is requesting, currently we will have to wait for
State Transfer to start to be able to catch up. By keeping the
information for the sequence numbers from the previous Working
Window, the behind Replica will be able to catch up by means of
ReqMissingDataMsg messages.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2651,8 +2651,8 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
   const ReplicaId msgSender = msg->senderId();
   SCOPED_MDC_SEQ_NUM(std::to_string(msgSeqNum));
   LOG_INFO(GL, "Received ReqMissingDataMsg. " << KVLOG(msgSender, msg->getFlags()));
-  if ((currentViewIsActive()) && mainLog->IsPressentInHistory(msgSeqNum)) {
-    SeqNumInfo &seqNumInfo = mainLog->getFromHistory(msgSeqNum);
+  if ((currentViewIsActive()) && (mainLog->insideActiveWindow(msgSeqNum) || mainLog->isPressentInHistory(msgSeqNum))) {
+    SeqNumInfo &seqNumInfo = mainLog->getFromActiveWindowOrHistory(msgSeqNum);
 
     if (config_.replicaId == currentPrimary()) {
       PrePrepareMsg *pp = seqNumInfo.getSelfPrePrepareMsg();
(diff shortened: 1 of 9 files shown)


bftengine/src/bftengine/SequenceWithActiveWindow.hpp
```diff
@@ -55,6 +115,10 @@ class SequenceWithActiveWindow {
     return ((n >= beginningOfActiveWindow) && (n < (beginningOfActiveWindow + WindowSize)));
   }
 
+  bool IsPressentInHistory(NumbersType n) const {
```
toly-kournik, line 118: A bit misleading name since it checks both history and active window.
I would make a separate method for history.
Minor: starts with lower case.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2651,9 +2651,8 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
   const ReplicaId msgSender = msg->senderId();
   SCOPED_MDC_SEQ_NUM(std::to_string(msgSeqNum));
   LOG_INFO(GL, "Received ReqMissingDataMsg. " << KVLOG(msgSender, msg->getFlags()));
-  if ((currentViewIsActive()) && (msgSeqNum > strictLowerBoundOfSeqNums) && (mainLog->insideActiveWindow(msgSeqNum)) &&
-      (mainLog->insideActiveWindow(msgSeqNum))) {
-    SeqNumInfo &seqNumInfo = mainLog->get(msgSeqNum);
+  if ((currentViewIsActive()) && mainLog->IsPressentInHistory(msgSeqNum)) {
```
toly-kournik, line 2654: see comment about IsPresentInHistory() below

bftengine/src/bftengine/SequenceWithActiveWindow.hpp
```diff
@@ -63,6 +127,17 @@ class SequenceWithActiveWindow {
     return activeWindow[i];
   }
 
+  ItemType &getFromHistory(NumbersType n) {
```
toly-kournik, line 130: Same as above, misleading name

toly-kournik (changes requested):
Overall it's good for now. I'm sure after the refactoring there will be much less changes most of which will be isolated in the SequenceWithActiveWindow

bftengine/src/bftengine/SequenceWithActiveWindow.hpp
HristoStaykov: Yes, will separate the two. I think to simplify request in ReplicaImp.cpp I'll add a method that explicitly searches both.

toly-kournik (approved):
LGTM for our current needs. 
Thanks @HristoStaykov

PR #998, opened 2020-10-29
[WIP] Docker update to version 1.0

oded282:
Click is a Python package for command-line interfaces. As part of building a CLI tool to parse Concord's logs, I use Click to implement the infrastructure for the parsing tool.

Additionally, reconfigured locale settings for two reasons:
1. Required for Click usage.
2. Solves the error that occurs when trying to use eliot-tree from within a docker container.


f-squirrel (changes requested):
Please change the commit summary to `Docker update to version 1.0` and put all the details in the commit message.

andrewjstone (changes requested):
This shouldn't be a 1.0 version. It should be version 0.10. We are not ready to call the build 1.0 IMO.

Dockerfile
```diff
@@ -30,3 +30,6 @@ path-exclude /usr/share/linda/*' > /etc/dpkg/dpkg.cfg.d/01_nodoc && \
             python3-pip \
             python3-setuptools \
             wget
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
```
andrewjstone, line 35: Nit: missing newline

f-squirrel:
> This shouldn't be a 1.0 version. It should be version 0.10. We are not ready to call the build 1.0 IMO.

Good point, Andrew!

Makefile
```diff
@@ -1,6 +1,6 @@
 CONCORD_BFT_DOCKER_REPO:=concordbft/
 CONCORD_BFT_DOCKER_IMAGE:=concord-bft
-CONCORD_BFT_DOCKER_IMAGE_VERSION:=0.9
+CONCORD_BFT_DOCKER_IMAGE_VERSION:=1.0
```
robem, line 3: Andrew mentioned this but I'm marking it explicitly. Our docker image should probably stay below 1.0 as long as concord-bft isn't 1.0. `0.10` should work.

robem (changes requested):

PR #1000, opened 2020-10-31
Decrease stateTranTimer_'s period.

HristoStaykov:
This is an initial value for this timer's period to perform tests. The goal is to make State Transfer ask for messages more frequently to speed up the process of catching up.


teoparvanov (approved):
LGTM!

PS: congrats for creating PR #1000 😄

PR #1002, opened 2020-11-01
Add has keys request

cloudnoize:
In order for BFT clients to know when the cluster has performed the initial key exchange and it's ready to accept requests

### DIFF
--- a/bftengine/include/bftengine/KeyExchangeMsg.hpp
+++ b/bftengine/include/bftengine/KeyExchangeMsg.hpp
@@ -1,0 +1,52 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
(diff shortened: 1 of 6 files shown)


bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -74,6 +74,12 @@ std::string KeyManager::generateCid() {
 // Key exchange msg for replica has been recieved:
 // update the replica public key store.
 std::string KeyManager::onKeyExchange(KeyExchangeMsg& kemsg, const uint64_t& sn) {
```
salieri11, line 76: What is the string based reply for?

bftengine/include/bftengine/KeyExchangeMsg.hpp
```diff
@@ -0,0 +1,52 @@
...
+  }
+  static KeyExchangeMsg deserializeMsg(const char* serializedMsg, const int& size) {
+    std::stringstream ss;
+    KeyExchangeMsg ke;
+    ss.write(serializedMsg, std::streamsize(size));
+    deserialize(ss, ke);
+    return ke;
+  }
+
+ protected:
```
salieri11, line 38: Shouldn't these methods override the ISerializable methods?

Liorgal28 (approved):
LGTM

PR #1004, opened 2020-11-02
reset signatures on state transfer

yontyon:
reset the signatures if we in the middle of state transfer
This solves a bug in which we get the same commit message after state transfer but never execute it


toly-kournik (approved):
LGTM
@yontyon, great catch!

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -47,6 +47,9 @@ void SeqNumInfo::acquire(SeqNumInfo& rhs) {
   commitMsgsCollector->acquire(rhs.commitMsgsCollector);
   partialProofsSet->acquire(rhs.partialProofsSet);
 }
+void SeqNumInfo::resetCommitSignatres() { commitMsgsCollector->resetAndFree(); }
```
andrewjstone, line 50: typo: `Signatres` -> `Signatures`.

andrewjstone (approved):
NIce fix! Small typo.

PR #1006, opened 2020-11-02
Enhance apollo logging and prevent busy waits

andrewjstone:
More details in the form of KV pairs were added to eliot actions and
logs in bft.py. This makes the data reported in eliot-tree much more
useful.

Additionally, in many cases we are spinning in a tight loop when a
condition is not met in an await. The expectation was tha the timeout
given in `trio.move_on_after` would prevent this, but it only ensures
that an await call will give up after that time. If the call succeeds
but the condition is not met we immediately loop again. This not only
causes 100% cpu usage, but also adds unnecessary data to the logs making
them harder to parse for eliot-tree. Every time we are about to execute
a tight loop we put a short `trio.sleep` call instead to allow the
coroutine to yield, and give time for the condition to become true.

Lastly, a bug was fixed such that when a call to
`write_and_track_known_kv` is made and the reply is checked, we take
into consideration the fact that the call may have timed out. We don't
want to throw an exception from the tracker, because in many cases we
just want the tracker to log the timeout. However, if we do need to
check the result, we can now raise a `ReadTimeoutError` rather than
getting an error when trying to access a the `success` field on a `None`
value.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -402,7 +402,7 @@ class BftTestNetwork:
         If the callback accepts three parameters and one of them is named 'config' - pass the network configuration too.
         Append the SSL certificate path. This is needed only for TLS communication.
         """
-        with log.start_action(action_type="start_replica_cmd"):
+        with log.start_action(action_type="start_replica_cmd", replica=replica_id):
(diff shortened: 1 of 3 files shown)


andrewjstone:

f-squirrel (approved):
Looks great to me 👍 
Maybe after this PR is merged my laptop will stop overheating!

PR #1008, opened 2020-11-02
Enhance logging of sent and received messages

andrewjstone:
A new logger, `MSGS`, was introduced so that debug messages for messages
sent and received by ReplicaImp and ReplicaBase could be turned on
independently. Both of those files were modified to use the new logger
instead of the global logger, `GL`.

Additionally, the name of the MsgCode is now printed instead of the
number. This makes following the logs much easier.


andrewjstone:

teoparvanov (approved):
LGTM!

robem (approved):

andrewjstone:
This was merged in as part of #949

PR #1011, opened 2020-11-03
BCStateTran: Split logging into 2: source and destination

glevkovich:
Here we split all logs in this module into 2:
1) concord.bft.st.src - logs of a regular non-fetching replica
2) concord.bft.st.dst = logs of a fetching replica.

This will make it easier to filter logs according to state transfer
status.


toly-kournik (approved):
LGTM!
Thanks, @glevkovich

PR #1012, opened 2020-11-03
Add Apollo test for catching up using request missing data mecahinsm

yontyon:
Add apollo test for cathing up with request missing info from the previous window

### DIFF
--- a/tests/apollo/test_skvbc.py
+++ b/tests/apollo/test_skvbc.py
@@ -80,6 +80,44 @@ class SkvbcTest(unittest.TestCase):
 
     @with_trio
     @with_bft_network(start_replica_cmd)
+    async def test_request_missing_data_from_previous_window(self, bft_network, exchange_keys=True):
+        """
+        Test that a replica succeeds to ask for missing info from the former window
+
+        1. Start all nodes and process 149 requests.
+        2. Stop one node and process 2 more requests (this will cause to the remaining
+            replicas to proceeds beyond the checkpoint)
+        3. The node should catchup without executing state transfer.
+        """
+
+        if exchange_keys:
+            await bft_network.do_key_exchange()
+        bft_network.start_all_replicas()
+
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        stale_node = random.choice(
+            bft_network.all_replicas(without={0}))
+        for i in range(151):
+            await skvbc.write_known_kv()
+            if i == 149:
+                bft_network.stop_replica(stale_node)
+
+        bft_network.start_replica(stale_node)
+        state_transfer_started = True
+        try:
+            await bft_network.wait_for_state_transfer_to_start()
+        except trio.TooSlowError:
+            state_transfer_started = False
+        self.assertFalse(state_transfer_started, "State Transfer has been started")
+
+        for r in bft_network.all_replicas():
+                last_stable_checkpoint = await bft_network.get_metric(r, bft_network, "Gauges", "lastStableSeqNum")
+                self.assertEqual(last_stable_checkpoint, 150)
+        await skvbc.assert_successful_put_get(self)
+
+    @with_trio
+    @with_bft_network(start_replica_cmd)
     async def test_get_block_data_with_blinking_replica(self, bft_network):
         """
         Test that the cluster continues working when one blinking replica
--- a/tests/apollo/test_skvbc.py
+++ b/tests/apollo/test_skvbc.py
@@ -85,8 +85,8 @@ class SkvbcTest(unittest.TestCase):
         Test that a replica succeeds to ask for missing info from the former window
 
         1. Start all nodes and process 149 requests.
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc.py
```diff
@@ -78,6 +78,44 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
...
+            await bft_network.do_key_exchange()
+        bft_network.start_all_replicas()
+
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        stale_node = random.choice(
+            bft_network.all_replicas(without={0}))
+        for i in range(151):
+            await skvbc.write_known_kv()
+            if i == 149:
```
teoparvanov, line 103: Why not stop the replica at `i == 70`, just to make this a bit more significant (so that we catch up 70 sequences, instead of just one) ?

teoparvanov (approved):
LGTM, with the above remark.

tests/apollo/test_skvbc.py
yontyon: The state transfer is starting after 2 seconds (regardless of our changes). I want to have a small gap such that state transfer won't have the opportunity to start

tests/apollo/test_skvbc.py
```diff
@@ -78,6 +78,44 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
...
         await skvbc.assert_successful_put_get(self)
 
+    @with_trio
+    @with_bft_network(start_replica_cmd)
+    async def test_request_missing_data_from_previous_window(self, bft_network, exchange_keys=True):
+        """
+        Test that a replica succeeds to ask for missing info from the former window
+
+        1. Start all nodes and process 149 requests.
+        2. Stop one node and process 2 more requests (this will cause to the remaining
```
HristoStaykov, line 88: Might be we can stop F Replicas, and before we bring them back, we can verify that the other 2F+C+1 have all reached a Stable Checkpoint. I think this way we will know for sure that the restarted F are catch up from the Inactive Window.

tests/apollo/test_skvbc.py
```diff
@@ -78,6 +78,44 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
...
+
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        stale_node = random.choice(
+            bft_network.all_replicas(without={0}))
+        for i in range(151):
+            await skvbc.write_known_kv()
+            if i == 149:
+                bft_network.stop_replica(stale_node)
+
```
HristoStaykov, line 105: I think it would be good to verify all up nodes reached to the stable Sequence ID before starting the catching up one, to be sure they shifted their Working Window.

HristoStaykov (approved):
LGTM, just 2 suggestions, thanks @yontyon !

tests/apollo/test_skvbc.py
```diff
@@ -78,6 +78,44 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
...
 
+    @with_trio
+    @with_bft_network(start_replica_cmd)
+    async def test_request_missing_data_from_previous_window(self, bft_network, exchange_keys=True):
+        """
+        Test that a replica succeeds to ask for missing info from the former window
+
+        1. Start all nodes and process 149 requests.
+        2. Stop one node and process 2 more requests (this will cause to the remaining
+            replicas to proceeds beyond the checkpoint)
```
andrewjstone, line 89: typo: proceeds -> proceed

tests/apollo/test_skvbc.py
```diff
@@ -78,6 +78,44 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
...
+
+        stale_node = random.choice(
+            bft_network.all_replicas(without={0}))
+        for i in range(151):
+            await skvbc.write_known_kv()
+            if i == 149:
+                bft_network.stop_replica(stale_node)
+
+        bft_network.start_replica(stale_node)
+        state_transfer_started = True
```
andrewjstone, line 107: This whole clause (lines 107-112) can be replaced with the more idiomatic:
```python
with self.assertRaises(trio.TooSlowError):
     await bft_network.wait_for_state_transfer_to_start()
```

tests/apollo/test_skvbc.py
```diff
@@ -78,6 +78,44 @@ async def test_state_transfer(self, bft_network,exchange_keys=True):
...
+        try:
+            await bft_network.wait_for_state_transfer_to_start()
+        except trio.TooSlowError:
+            state_transfer_started = False
+        self.assertFalse(state_transfer_started, "State Transfer has been started")
+
+        for r in bft_network.all_replicas():
+                last_stable_checkpoint = await bft_network.get_metric(r, bft_network, "Gauges", "lastStableSeqNum")
```
andrewjstone, line 115: Please indent 4 spaces. Don't use tabs.

PR #1015, opened 2020-11-04
Apollo: Do not log a function using eliot

andrewjstone:
This is not serializable and causes errors of the form:
```
├── eliot:destination_failure/10/3 2020-11-04 03:11:29Z
    │   │   ├── exception: builtins.TypeError
    │   │   ├── message: {"'replica'": '3', "'expected'": '<function SkvbcSlowPathTest.test_slow_path_view_change.<locals>.<l…
    │   │   └── reason: Object of type 'function' is not JSON serializable
```


teoparvanov (approved):
LGTM, thanks Andrew!

PR #1017, opened 2020-11-04
fix small bug in request missing info mechanism

yontyon:
Before sending a request for missing info, a replica checks if not enough time passed for the sequence number from (a) the last time a request for missing info was sent or (b) the last time we got data from the primary for this sequence number.

Alas, in the current code we check if the time == MinTime which means that we don't actually check this condition (and hence sending unnecessary requests)


toly-kournik:
@yontyon can you please elaborate on it?

andrewjstone (approved):
Great fix @yontyon! It will be really nice to not have these messages unnecessarily filling the logs.

Thanks for the excellent description as well.
Can you please add the description to the commit message itself though, with `git commit --amend`. This will make us easier to see why the change was made in the future with tools like `git blame`.

PR #1025, opened 2020-11-08
Patch for v0.8

yontyon:
This PR contains several patches that should help to avoid endless state transfer.

* Once the primary gets to a stable checkpoint, it takes the current time. It will send the next PrePrepare only if either at least one seconds has passed from that time or if it gets to an n/n checkpoint. This should help late replicas to catch up and thus avoiding from endless state transfer.

* Make all replicas involving in answering to missing PrePrepare request


bftengine/src/bftengine/messages/ReqMissingDataMsg.hpp
```diff
@@ -38,6 +38,19 @@ class ReqMissingDataMsg : public MessageBase {
   bool getSlowPathHasStarted() const { return b()->flags.bits.slowPathHasStarted != 0; }
 
   uint16_t getFlags() const { return b()->flags.flags; }
+  std::string getFlagsAsBits() const {
```
toly-kournik, line 41: os << std::bitset(b()->flags.flags)

PR #1029, opened 2020-11-10
State Transfer onTimer handlig

toly-kournik:
If configured to run in a separate thread, hand off onTimer() processing to a this thread along with message processing.


tdimitrov (approved):
Looks good and solves all the problems I've found.

PR #1031, opened 2020-11-10
Support non-provable keys

f-squirrel:
Reason:
Some keys don't need proof. By saving the non-provable keys outside of the Merkle Tree,
Concord-BFT can save a significant amount of disk space. In some scenarios up to 10 times.

Changes:
* Merkle Tree DB Adapter is initialized with a set of non-provable keys.
* If the updates passed to the method `addBlock` contain the non-provable keys,
    the key-values are stored outside of the Merkle Tree.
* The method `addBlock` throws an exception if the "deletes" set
    contains a non-provable key.
* The non-provable keys are stored in the following way:
    `|EKeyType::Key |EKeySubtype::NonProvable|Block ID|Key1 |  = Value1`
* A block with non-provable keys will look like:
    ```
    |EDBKeyType::Block |Block ID                               |  = Block Node
    |EKeyType::Key     |EKeySubtype::NonProvable|Block ID|Key1 |  = Value1
    |EKeyType::Key     |EKeySubtype::NonProvable|Block ID|Key2 |  = Value2
    ```
* The method `getValue` checks if a given key is non-provable.
    If so, it looks up according to the scheme above, if it fails to find,
    the regular search is done.

Note:
Non-provable keys will not be pruned as of this commit and that a subsequent one will do it.

### DIFF
--- a/kvbc/include/merkle_tree_db_adapter.h
+++ b/kvbc/include/merkle_tree_db_adapter.h
@@ -26,6 +26,7 @@
 #include <future>
 #include <memory>
 #include <optional>
+#include <unordered_set>
 #include <utility>
 
 namespace concord::kvbc::v2MerkleTree {
@@ -48,6 +49,7 @@ namespace concord::kvbc::v2MerkleTree {
 // getLastReachableBlockId() + 1. Additionally, ReachableSTBlock <= getLatestBlockId().
 class DBAdapter : public IDbAdapter {
  public:
+  using NonProvableKeySet = std::unordered_set<Key>;
   // Unless explicitly turned off, the constructor will try to link the blockchain with any blocks in the temporary
   // state transfer chain. This is done so that the DBAdapter will operate correctly in case a crash or an abnormal
   // shutdown has occurred prior to startup (construction). Only a single DBAdapter instance should operate on a
@@ -55,7 +57,11 @@ class DBAdapter : public IDbAdapter {
   // mechanism. The constructor throws if an error occurs.
   // Note1: The passed DB client must be initialized beforehand.
   // Note2: The 'linkTempSTChain' parameter turns of chain linking and is used for testing/tooling purposes.
-  DBAdapter(const std::shared_ptr<concord::storage::IDBClient> &db, bool linkTempSTChain = true);
+  // Note3: The key provided via 'non_provable_key_set' parameter will be stored outside of the Merkle Tree.
+  // Note4: Non-provable keys cannot be deleted for now.
+  DBAdapter(const std::shared_ptr<concord::storage::IDBClient> &db,
+            bool linkTempSTChain = true,
+            const NonProvableKeySet &nonProvableKeySet = NonProvableKeySet{});
 
   // Make the adapter non-copyable.
   DBAdapter(const DBAdapter &) = delete;
@@ -208,6 +214,7 @@ class DBAdapter : public IDbAdapter {
   std::shared_ptr<storage::IDBClient> db_;
   sparse_merkle::Tree smTree_;
   std::unique_ptr<concordMetrics::ISummary> commitSizeSummary_;
+  const NonProvableKeySet nonProvableKeySet_;
 };
 
 namespace detail {
--- a/kvbc/src/merkle_tree_db_adapter.cpp
+++ b/kvbc/src/merkle_tree_db_adapter.cpp
@@ -116,14 +116,17 @@ void add(KeysVector &to, const KeysVector &src) { to.insert(std::end(to), std::c
 
 }  // namespace
 
-DBAdapter::DBAdapter(const std::shared_ptr<IDBClient> &db, bool linkTempSTChain)
+DBAdapter::DBAdapter(const std::shared_ptr<IDBClient> &db,
+                     bool linkTempSTChain,
+                     const NonProvableKeySet &nonProvableKeySet)
     : logger_{logging::getLogger("concord.kvbc.v2MerkleTree.DBAdapter")},
       // The smTree_ member needs an initialized DB. Therefore, do that in the initializer list before constructing
       // smTree_ .
       db_{db},
       smTree_{std::make_shared<Reader>(*this)},
       commitSizeSummary_{concordMetrics::StatisticsFactory::get().createSummary(
-          "merkleTreeCommitSizeSummary", {{0.25, 0.1}, {0.5, 0.1}, {0.75, 0.1}, {0.9, 0.1}})} {
+          "merkleTreeCommitSizeSummary", {{0.25, 0.1}, {0.5, 0.1}, {0.75, 0.1}, {0.9, 0.1}})},
+      nonProvableKeySet_{nonProvableKeySet} {
   if (linkTempSTChain) {
     // Make sure that if linkSTChainFrom() has been interrupted (e.g. a crash or an abnormal shutdown), all DBAdapter
     // methods will return the correct values. For example, if state transfer had completed and linkSTChainFrom() was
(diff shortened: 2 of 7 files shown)


kvbc/include/merkle_tree_db_adapter.h
```diff
@@ -48,14 +49,18 @@ namespace concord::kvbc::v2MerkleTree {
...
   // shutdown has occurred prior to startup (construction). Only a single DBAdapter instance should operate on a
   // database and access to all methods should be either done from a single thread or serialized via a mutex or another
   // mechanism. The constructor throws if an error occurs.
   // Note1: The passed DB client must be initialized beforehand.
   // Note2: The 'linkTempSTChain' parameter turns of chain linking and is used for testing/tooling purposes.
-  DBAdapter(const std::shared_ptr<concord::storage::IDBClient> &db, bool linkTempSTChain = true);
+  // Note3: The key provided via 'non_provable_key_set' parameter will be stored outside of the Merkle Tree.
+  DBAdapter(const std::shared_ptr<concord::storage::IDBClient> &db,
+            bool linkTempSTChain = true,
+            const NonProvableKeySet &non_provable_key_set = NonProvableKeySet{});
```
dartdart26, line 63: Nit: Since this file uses camelCase, maybe we can name it `nonProvableKeySet`.

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -334,18 +358,36 @@ SetOfKeyValuePairs DBAdapter::lastReachableBlockDbUpdates(const SetOfKeyValuePai
...
+  // Split the keys: NonProvable and Provable Keys
+  SetOfKeyValuePairs non_provable_kvs;
+  SetOfKeyValuePairs provable_kvs = updates;
+  for (const auto &k : non_provable_key_set_) {
+    if (deletes.find(k) != std::cend(deletes)) {
+      throw std::runtime_error{"The deletes key set contains a non-provable key: " + k.toString()};
+    }
+    auto iter = provable_kvs.find(k);
+    if (iter != std::cend(provable_kvs)) {
+      non_provable_kvs.emplace(iter->first, iter->second);
```
dartdart26, line 370: Based on https://en.cppreference.com/w/cpp/container/unordered_set/emplace, all iterators might be invalidated if rehashing occurs. I think that is problematic with the loop being done over `non_provable_key_set_`.

storage/include/storage/db_types.h
```diff
@@ -50,8 +50,10 @@ enum class EDBKeyType : std::uint8_t {
 // containing actual application data.
 enum class EKeySubtype : std::uint8_t {
   Internal,
-  Stale,
+  ProvableStale,
+  NonProvableStale,
```
dartdart26, line 54: In order to preserve compatibility with an existing DB, I think we need to add new types at the end of the enum. In particular, adding `NonProvableStale` before `Leaf` will change the value for `Leaf`.

dartdart26: Another thing came to mind - would be nice if we can test that somehow in a unit test. Will be relevant for migrations, as a whole.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -105,7 +106,7 @@ bool hasStaleIndexKeysSince(const std::shared_ptr<IDBClient> &db, const Version
   auto iter = db->getIteratorGuard();
   const auto key = iter->seekAtLeast(DBKeyManipulator::genStaleDbKey(version)).first;
   if (!key.empty() && DBKeyManipulator::getDBKeyType(key) == EDBKeyType::Key &&
-      DBKeyManipulator::getKeySubtype(key) == EKeySubtype::Stale &&
+      DBKeyManipulator::getKeySubtype(key) == EKeySubtype::ProvableStale &&
```
dartdart26, line 109: Maybe that function should be renamed, because it only checks for `ProvableStale` keys.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -183,6 +184,126 @@ TEST_P(db_adapter_custom_blockchain, add_block_return) {
...
+  const Sliver skey2{std::move(key2)};
+  concord::kvbc::v2MerkleTree::DBAdapter::NonProvableKeySet non_provable_key_set = {skey1, skey2};
+  auto adapter = DBAdapter{GetParam()->newEmptyDb(), true, non_provable_key_set};
+
+  const auto updates = SetOfKeyValuePairs{{skey1, defaultSliver}, {skey2, defaultSliver}};
+  const auto deletes = OrderedKeysSet{skey1};
+  ASSERT_THROW(adapter.addBlock(updates, deletes), std::runtime_error);
+}
+
+#ifdef USE_ROCKSDB
```
dartdart26, line 272: In order to test with both ` memorydb` and `RocksDB`, I think you can use `db_adapter_custom_blockchain` and get the test DB by calling `GetParam()->newEmptyDb()` and keeping the pointer. Then it can be used twice for both DBAdapter instances. That way we can avoid the `ifdef`. Please correct me if I miss something.

kvbc/include/merkle_tree_db_adapter.h
```diff
@@ -48,14 +49,18 @@ namespace concord::kvbc::v2MerkleTree {
...
+  using NonProvableKeySet = std::unordered_set<Key>;
   // Unless explicitly turned off, the constructor will try to link the blockchain with any blocks in the temporary
   // state transfer chain. This is done so that the DBAdapter will operate correctly in case a crash or an abnormal
   // shutdown has occurred prior to startup (construction). Only a single DBAdapter instance should operate on a
   // database and access to all methods should be either done from a single thread or serialized via a mutex or another
   // mechanism. The constructor throws if an error occurs.
   // Note1: The passed DB client must be initialized beforehand.
   // Note2: The 'linkTempSTChain' parameter turns of chain linking and is used for testing/tooling purposes.
-  DBAdapter(const std::shared_ptr<concord::storage::IDBClient> &db, bool linkTempSTChain = true);
+  // Note3: The key provided via 'non_provable_key_set' parameter will be stored outside of the Merkle Tree.
```
dartdart26, line 60: Maybe we can put a comment either here or on `addBlock` methods to say that non-provable keys cannot be explicitly deleted for now.

dartdart26:
Thanks @f-squirrel, that is great! Looks simple, yet we can have big gains from it. Thanks!

Most of my comments are minor. The ones I think should be addressed are related to ordering of the key types in the `EKeySubtype` and `std::unordered_set` iterator invalidation.

Another thing is that maybe we want to mention explicitly that non-provable keys will not be pruned as of this PR and that a subsequent one will do it.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: I think above is a mistake, I didn't read the code correctly. Nothing is emplaced in `non_provable_key_set_`. Please ignore.

dartdart26:
P.S. Please ignore the part about `std::unordered` set, my mistake.

storage/include/storage/db_types.h
f-squirrel: Added a `static_assert` in the header file.

f-squirrel:
@dartdart26 , thank you for your detailed review.
I've addressed your comments in the updated commit.

kvbc/include/merkle_tree_db_adapter.h
f-squirrel: Done

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: Done

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: Great idea, done!

kvbc/include/merkle_tree_db_adapter.h
f-squirrel: Done: `  // Note4: Non-provable keys cannot be deleted for now.`

storage/include/storage/db_types.h
dartdart26: I was thinking that maybe we can be explicit about the values themselves in the enum declaration. And maybe we can put a comment to say that we can't change the existing ones.

I also had a look at the file and the comment at the bottom about key ordering is no longer correct (the one with the ASCII art). I think it would be better to remove it and just comment on the enum itself that we cannot change the existing values.

dartdart26 (approved):
Thanks @f-squirrel!

PR #1034, opened 2020-11-10
Pre-execution histograms

Liorgal28:
On this MR I have added histograms in all main methods on the pre-execution flow in order to measure timings inside each method.
We can use these measurements to learn where the bottlenecks might be and if there are any.


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -241,6 +241,7 @@ void PreProcessor::sendRejectPreProcessReplyMsg(NodeIdType clientId,
 
 template <>
 void PreProcessor::onMessage<ClientPreProcessRequestMsg>(ClientPreProcessRequestMsg *msg) {
+  concord::diagnostics::TimeRecorder scoped_timer(*histograms_.onMessage);
```
yuliasherman, line 244: 'onMessage' is a template function, do we combine results for all kind of received messages?

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -470,6 +471,7 @@ void PreProcessor::finalizePreProcessing(NodeIdType clientId) {
   {
     lock_guard<mutex> lock(clientEntry->mutex);
     auto &reqProcessingStatePtr = clientEntry->reqProcessingStatePtr;
+    concord::diagnostics::TimeRecorder scoped_timer(*histograms_.finalizePreProcessing);
```
yuliasherman, line 474: This needs to be added at the beginning of the function to include wait-for-a-lock time, too.

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -175,6 +176,39 @@ class PreProcessor {
   concordUtil::Timers::Handle requestsStatusCheckTimer_;
   const uint64_t preExecReqStatusCheckPeriodMilli_;
   concordUtil::Timers &timers_;
+
+  // 5 Minutes
+  static constexpr int64_t MAX_VALUE_MICROSECONDS = 1000 * 1000 * 60 * 5l;
+  // 60 seconds
+  static constexpr int64_t MAX_VALUE_NANOSECONDS = 1000 * 1000 * 1000 * 60l;
```
yuliasherman, line 183: Remove it

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -175,6 +176,39 @@ class PreProcessor {
   concordUtil::Timers::Handle requestsStatusCheckTimer_;
   const uint64_t preExecReqStatusCheckPeriodMilli_;
   concordUtil::Timers &timers_;
+
+  // 5 Minutes
+  static constexpr int64_t MAX_VALUE_MICROSECONDS = 1000 * 1000 * 60 * 5l;
```
yuliasherman, line 181: Put just a number. 
Add a comment with a value in seconds.

bftengine/src/preprocessor/PreProcessor.cpp
Liorgal28: If I understood you correctly, it is only for ClientPreProcessRequestMsg.

bftengine/src/preprocessor/PreProcessor.cpp
Liorgal28: Done.

bftengine/src/preprocessor/PreProcessor.hpp
Liorgal28: Done.

bftengine/src/preprocessor/PreProcessor.hpp
Liorgal28: Done.

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -177,21 +177,24 @@ class PreProcessor {
...
+      try {
+        registrar.perf.registerComponent("pre-execution",
+                                         {{"onMessage", onMessage},
+                                          {"launchReqPreProcessing", launchReqPreProcessing},
+                                          {"handlePreProcessedReqByNonPrimary", handlePreProcessedReqByNonPrimary},
+                                          {"handlePreProcessedReqByPrimary", handlePreProcessedReqByPrimary},
+                                          {"sendPreProcessRequestToAllReplicas", sendPreProcessRequestToAllReplicas},
+                                          {"finalizePreProcessing", finalizePreProcessing}});
+      } catch (std::invalid_argument &e) {
+        // if component already exist lets keep record on the same histograms
```
yuliasherman, line 196: Nit: exist-> exists

yuliasherman (approved):
LGTM

andrewjstone (approved):
Looks great! Thanks @Liorgal28

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -175,6 +176,42 @@ class PreProcessor {
...
+      try {
+        registrar.perf.registerComponent("pre-execution",
+                                         {{"onMessage", onMessage},
+                                          {"launchReqPreProcessing", launchReqPreProcessing},
+                                          {"handlePreProcessedReqByNonPrimary", handlePreProcessedReqByNonPrimary},
+                                          {"handlePreProcessedReqByPrimary", handlePreProcessedReqByPrimary},
+                                          {"sendPreProcessRequestToAllReplicas", sendPreProcessRequestToAllReplicas},
+                                          {"finalizePreProcessing", finalizePreProcessing}});
+      } catch (std::invalid_argument &e) {
+        // if component already exists lets keep record on the same histograms
```
andrewjstone, line 196: I didn't notice this in my initial review. This doesn't work as intended. Instead, measurements will be silently dropped as I noted when updating the API in [this commit](https://github.com/vmware/concord-bft/pull/949/commits/b38c5fd850df55b70db95b46478947593488eb81) in [the tls pr](https://github.com/vmware/concord-bft/pull/949).

Unfortunately this is a bug that causes tests to fail from dual registration. I'm attempting to fix it, but the change might require input from you @Liorgal28 as it's a semantic issue. 

In short, there can only be one component with a given name. Ignoring the exception is no longer going to be an option because the process will terminate.

andrewjstone: I actually was able to fix this by modifying the test to clear the registrations. This is in general a problem of singletons, in that they introduce these types of side effects. Maybe in the future when we have a better startup sequence we can pass the registrar to components rather than using a singleton. For now this works fine and is easy enough.

Liorgal28: Thanks for addressing that issue, I commented in https://github.com/vmware/concord-bft/pull/949.

PR #1036, opened 2020-11-11
JSON parser utility for dynamic configuration

toly-kournik:
Will be used for various concord-bft parameter tweaks.

Historically (and with no real reason) YAML is used for concord configuration.
Since:
 - We want no dependency on an external configuration parser library;
 - YAML parser is much harder to implement;
 - YAML and JSON are interoperable;
 
JSON was chosen.
But no worries, concord will remain with YAML configuration.

### DIFF
--- a/kvbc/include/categorization/column_families.h
+++ b/kvbc/include/categorization/column_families.h
@@ -17,16 +17,20 @@
 
 namespace concord::kvbc::categorization::detail {
 
-inline const auto IMMUTABLE_KV_CF_SUFFIX = std::string{"_immutable"};
-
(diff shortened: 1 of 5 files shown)


robem:
JSON is not a good format for a configuration file. A quick "google" will show you multiple blogs/articles that argue why JSON isn't a good configuration language and I pretty much agree. I'd prefer [TOML](https://github.com/toml-lang/toml) over [YAML](https://yaml.org/) because IMHO it's easier to read. Also, we should try to find an existing parser instead of writing our own.

PR #1042, opened 2020-11-12
Support Pruning for non-provable keys

f-squirrel:
This commit adds the possibility to mark non-provable keys as
NonProvableStale in case a given key already exists.

Changes:
* The method `addBlock` checks if input non-provable keys already exist
and mark them as stale.
* `DBAdapter::lastReachableBlockKeyDeletes` and `DBAdapter::genesisBlockKeyDeletes` are udpated to support
the cleanup of non-provable keys.
* Added method `DBAdapter::staleIndexKeysForBlock` to look for stale
indices.
* Unit tests.

Additional information:
The non-provable stale keys are stored in the DB in the following way:
```
|EDBKeyType::Key|EKeySubtype::NonProvableStale|Stale since version(block id)|Actual NonProvable key|

### DIFF
--- a/kvbc/include/merkle_tree_db_adapter.h
+++ b/kvbc/include/merkle_tree_db_adapter.h
@@ -58,7 +58,8 @@ class DBAdapter : public IDbAdapter {
   // mechanism. The constructor throws if an error occurs.
   // Note1: The passed DB client must be initialized beforehand.
   // Note2: The 'linkTempSTChain' parameter turns of chain linking and is used for testing/tooling purposes.
-  // Note3: The key provided via 'non_provable_key_set' parameter will be stored outside of the Merkle Tree.
+  // Note3: The key provided via 'nonProvableKeySet' parameter will be stored outside of the Merkle Tree,
+  // the keys must have the same size!
   // Note4: Non-provable keys cannot be deleted for now.
   DBAdapter(const std::shared_ptr<concord::storage::IDBClient> &db,
             bool linkTempSTChain = true,
@@ -182,9 +183,11 @@ class DBAdapter : public IDbAdapter {
 
   block::detail::Node getBlockNode(BlockId blockId) const;
 
-  KeysVector staleIndexKeysForVersion(const sparse_merkle::Version &version) const;
+  KeysVector staleIndexNonProvableKeysForBlock(BlockId blockId) const;
 
-  KeysVector internalKeysForVersion(const sparse_merkle::Version &version) const;
+  KeysVector staleIndexProvableKeysForVersion(const sparse_merkle::Version &version) const;
+
+  KeysVector internalProvableKeysForVersion(const sparse_merkle::Version &version) const;
 
   KeysVector lastReachableBlockKeyDeletes(BlockId blockId) const;
 
@@ -203,6 +206,8 @@ class DBAdapter : public IDbAdapter {
 
   // Return std::nullopt if no temporary ST blocks are present.
   std::optional<BlockId> loadLatestTempSTBlockId() const;
+
+  std::optional<std::pair<Value, BlockId>> getValueForNonProvableKey(const Key &key, const BlockId &blockVersion) const;
 
   class Reader : public sparse_merkle::IDBReader {
    public:
--- a/kvbc/include/merkle_tree_key_manipulator.h
+++ b/kvbc/include/merkle_tree_key_manipulator.h
@@ -26,12 +26,13 @@ namespace concord::kvbc::v2MerkleTree::detail {
 class DBKeyManipulator {
  public:
   static Key genBlockDbKey(BlockId version);
-  static Key genNonProvableBlockDbKey(BlockId block_id, const Key &key);
+  static Key genNonProvableDbKey(BlockId block_id, const Key &key);
   static Key genDataDbKey(const sparse_merkle::LeafKey &key);
   static Key genDataDbKey(const Key &key, const sparse_merkle::Version &version);
   static Key genInternalDbKey(const sparse_merkle::InternalNodeKey &key);
   static Key genStaleDbKey(const sparse_merkle::InternalNodeKey &key, const sparse_merkle::Version &staleSinceVersion);
   static Key genStaleDbKey(const sparse_merkle::LeafKey &key, const sparse_merkle::Version &staleSinceVersion);
+  static Key genNonProvableStaleDbKey(const Key &key, BlockId staleSinceBlock);
   // Version-only stale keys do not exist in the DB. They are used as a placeholder for searching through the stale
   // index. Rationale is that they are a lower bound of all stale keys for a specific version due to lexicographical
   // ordering, the fact that the version comes first and that real stale keys are longer and always follow version-only
@@ -53,10 +54,16 @@ class DBKeyManipulator {
   static sparse_merkle::Hash extractHashFromLeafKey(const Key &key);
 
   // Extract the stale since version from a stale node index key.
-  static sparse_merkle::Version extractVersionFromStaleKey(const Key &key);
+  static sparse_merkle::Version extractVersionFromProvableStaleKey(const Key &key);
+
+  // Extract the stale since block id from a stale node index key.
+  static BlockId extractBlockIdFromNonProvableStaleKey(const Key &key);
 
   // Extract the actual key from the stale node index key.
-  static Key extractKeyFromStaleKey(const Key &key);
+  static Key extractKeyFromProvableStaleKey(const Key &key);
+
+  // Extract the actual key from the non-provable stale key.
+  static Key extractKeyFromNonProvableStaleKey(const Key &key);
 
   // Extract the version of an internal key.
   static sparse_merkle::Version extractVersionFromInternalKey(const Key &key);
--- a/kvbc/src/merkle_tree_db_adapter.cpp
+++ b/kvbc/src/merkle_tree_db_adapter.cpp
@@ -25,6 +25,8 @@
 #include <iterator>
 #include <limits>
 #include <memory>
+#include <optional>
+#include <stdexcept>
 #include <utility>
 
 namespace concord::kvbc::v2MerkleTree {
@@ -93,12 +95,12 @@ auto hash(const Sliver &buf) {
   return hasher.hash(buf.data(), buf.length());
 }
 
-template <typename VersionExtractor>
+template <typename VersionT, typename VersionExtractorT>
 KeysVector keysForVersion(const std::shared_ptr<IDBClient> &db,
                           const Key &firstKey,
-                          const Version &version,
+                          const VersionT &version,
                           EKeySubtype keySubtype,
-                          const VersionExtractor &extractVersion) {
+                          const VersionExtractorT &extractVersion) {
   TimeRecorder scoped(*histograms.dba_keys_for_version);
   auto keys = KeysVector{};
   auto iter = db->getIteratorGuard();
@@ -114,6 +116,20 @@ KeysVector keysForVersion(const std::shared_ptr<IDBClient> &db,
 
 void add(KeysVector &to, const KeysVector &src) { to.insert(std::end(to), std::cbegin(src), std::cend(src)); }
 
+template <typename ContainerT>
+std::pair<ContainerT, ContainerT> splitToProvableAndNonProvable(const ContainerT &updates,
+                                                                const DBAdapter::NonProvableKeySet &nonProvableKeySet) {
+  ContainerT provableKvPairs = updates;
+  ContainerT nonProvableKvPairs;
+  for (const auto &k : nonProvableKeySet) {
+    auto iter = provableKvPairs.find(k);
+    if (iter != std::cend(provableKvPairs)) {
+      nonProvableKvPairs.emplace(iter->first, iter->second);
+      provableKvPairs.erase(iter);
+    }
+  }
+  return {provableKvPairs, nonProvableKvPairs};
+}
 }  // namespace
 
 DBAdapter::DBAdapter(const std::shared_ptr<IDBClient> &db,
@@ -130,6 +146,14 @@ DBAdapter::DBAdapter(const std::shared_ptr<IDBClient> &db,
       commitSizeSummary_{concordMetrics::StatisticsFactory::get().createSummary(
           "merkleTreeCommitSizeSummary", {{0.25, 0.1}, {0.5, 0.1}, {0.75, 0.1}, {0.9, 0.1}})},
       nonProvableKeySet_{nonProvableKeySet} {
+  if (!nonProvableKeySet_.empty()) {
+    const auto length = nonProvableKeySet_.begin()->length();
+    for (const auto &k : nonProvableKeySet_) {
+      if (length != k.length()) {
+        throw std::runtime_error{"Non-provable keys must have the same length"};
+      }
+    }
+  }
   if (linkTempSTChain) {
     // Make sure that if linkSTChainFrom() has been interrupted (e.g. a crash or an abnormal shutdown), all DBAdapter
     // methods will return the correct values. For example, if state transfer had completed and linkSTChainFrom() was
@@ -140,16 +164,26 @@ DBAdapter::DBAdapter(const std::shared_ptr<IDBClient> &db,
   }
 }
 
+std::optional<std::pair<Value, BlockId>> DBAdapter::getValueForNonProvableKey(const Key &key,
+                                                                              const BlockId &blockVersion) const {
+  const auto &blockKey = DBKeyManipulator::genNonProvableDbKey(blockVersion, key);
+  auto iter = db_->getIteratorGuard();
+  const auto [foundBlockKey, foundBlockValue] = iter->seekAtMost(blockKey);
+  if (!foundBlockKey.empty() && DBKeyManipulator::getDBKeyType(foundBlockKey) == EDBKeyType::Key &&
+      DBKeyManipulator::getKeySubtype(foundBlockKey) == EKeySubtype::NonProvable &&
+      DBKeyManipulator::extractKeyFromNonProvableKey(foundBlockKey) == key) {
+    return {{foundBlockValue, DBKeyManipulator::extractBlockIdFromNonProvableKey(foundBlockKey)}};
+  }
+  return std::nullopt;
+}
+
 std::pair<Value, BlockId> DBAdapter::getValue(const Key &key, const BlockId &blockVersion) const {
   TimeRecorder scoped_timer(*histograms.dba_get_value);
   auto stateRootVersion = Version{};
   if (nonProvableKeySet_.find(key) != std::cend(nonProvableKeySet_)) {
-    const auto &blockKey = DBKeyManipulator::genNonProvableBlockDbKey(blockVersion, key);
-    auto iter = db_->getIteratorGuard();
-    const auto [foundBlockKey, foundBlockValue] = iter->seekAtMost(blockKey);
-    if (!foundBlockKey.empty() && DBKeyManipulator::getDBKeyType(foundBlockKey) == EDBKeyType::Key &&
-        DBKeyManipulator::getKeySubtype(foundBlockKey) == EKeySubtype::NonProvable) {
-      return {foundBlockValue, DBKeyManipulator::extractBlockIdFromNonProvableKey(foundBlockKey)};
+    auto ret = getValueForNonProvableKey(key, blockVersion);
+    if (ret) {
+      return std::move(ret.value());
     }
   }
   // Find a block with an ID that is less than or equal to the requested block version and extract the state root
@@ -277,20 +311,31 @@ RawBlock DBAdapter::getRawBlock(const BlockId &blockId) const {
   auto keyValues = SetOfKeyValuePairs{};
   auto deletedKeys = OrderedKeysSet{};
   for (const auto &[key, keyData] : blockNode.keys) {
-    if (!keyData.deleted) {
+    if (nonProvableKeySet_.find(key) != std::cend(nonProvableKeySet_)) {
       Sliver value;
-      if (const auto status = db_->get(DBKeyManipulator::genDataDbKey(key, blockNode.stateRootVersion), value);
-          !status.isOK()) {
+      if (const auto status = db_->get(DBKeyManipulator::genNonProvableDbKey(blockId, key), value); !status.isOK()) {
         // If the key is not found, treat as corrupted storage and abort.
         ConcordAssert(!status.isNotFound());
-        throw std::runtime_error{"Failed to get value by key from DB, block node ID = " + std::to_string(blockId)};
+        throw std::runtime_error{"Failed to get value by non-provable key from DB, block node ID = " +
+                                 std::to_string(blockId)};
       }
-      TimeRecorder scoped(*histograms.dba_deserialize_leaf);
-      const auto dbLeafVal = detail::deserialize<detail::DatabaseLeafValue>(value);
-      ConcordAssert(dbLeafVal.addedInBlockId == blockId);
-      keyValues[key] = dbLeafVal.leafNode.value;
+      keyValues[key] = value;
     } else {
-      deletedKeys.insert(key);
+      if (!keyData.deleted) {
+        Sliver value;
+        if (const auto status = db_->get(DBKeyManipulator::genDataDbKey(key, blockNode.stateRootVersion), value);
+            !status.isOK()) {
+          // If the key is not found, treat as corrupted storage and abort.
+          ConcordAssert(!status.isNotFound());
+          throw std::runtime_error{"Failed to get value by key from DB, block node ID = " + std::to_string(blockId)};
+        }
+        TimeRecorder scoped(*histograms.dba_deserialize_leaf);
+        const auto dbLeafVal = detail::deserialize<detail::DatabaseLeafValue>(value);
+        ConcordAssert(dbLeafVal.addedInBlockId == blockId);
+        keyValues[key] = dbLeafVal.leafNode.value;
+      } else {
+        deletedKeys.insert(key);
+      }
     }
   }
 
@@ -325,10 +370,14 @@ SetOfKeyValuePairs DBAdapter::lastReachableBlockDbUpdates(const SetOfKeyValuePai
 
   // Find keys that are deleted only and not updated. We need that list, because updates take precedence over deletes
   // and we should only try to update deleted keys.
+  // Make sure the deletes do not contain non-provable keys.
   auto deletesOnly = KeysVector{};
   for (const auto &key : deletes) {
     if (updates.find(key) == std::cend(updates)) {
       deletesOnly.push_back(key);
+    }
+    if (nonProvableKeySet_.find(key) != std::cend(nonProvableKeySet_)) {
+      throw std::runtime_error{"The deletes key set contains a non-provable key: " + key.toString()};
     }
   }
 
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: kvbc/include/merkle_tree_db_adapter.h [hunk 1]
```diff
@@ -181,6 +181,8 @@ class DBAdapter : public IDbAdapter {
 
   block::detail::Node getBlockNode(BlockId blockId) const;
 
+  KeysVector staleIndexKeysForBlock(BlockId blockId) const;
+
   KeysVector staleIndexKeysForVersion(const sparse_merkle::Version &version) const;
```
dartdart26, line 186: Maybe we can rename that to indicate these are stale tree keys. `staleIndexKeysForBlock `, on the other hand, can be named in a way to indicate it relates to non-provable keys.

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -137,16 +138,25 @@ DBAdapter::DBAdapter(const std::shared_ptr<IDBClient> &db,
   }
 }
 
+std::optional<std::pair<Value, BlockId>> DBAdapter::getValueForNonProvableKey(const Key &key,
+                                                                              const BlockId &blockVersion) const {
+  const auto &blockKey = DBKeyManipulator::genNonProvableBlockDbKey(blockVersion, key);
```
dartdart26, line 143: Nit: I think it will be more readable if we don't bind a reference to a temporary, i.e. just do `const auto blockKey`. Looking at the const ref, it looks like we are returning some constant internal object that way.

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -387,6 +397,14 @@ SetOfKeyValuePairs DBAdapter::lastReachableBlockDbUpdates(const SetOfKeyValuePai
 
   for (const auto &[k, v] : non_provable_kvs) {
     dbUpdates[DBKeyManipulator::genNonProvableBlockDbKey(blockId, k)] = v;
+    if (blockId) {
+      const auto &stale_kv = getValueForNonProvableKey(k, blockId - 1);
```
dartdart26, line 401: Nit: I think that even though the style for this file is incorrect (camelCase), it might be better to stick to it (i.e. `staleKv`) and then change all occurrences it in one go. Applies to other uses too.

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -387,6 +397,14 @@ SetOfKeyValuePairs DBAdapter::lastReachableBlockDbUpdates(const SetOfKeyValuePai
 
   for (const auto &[k, v] : non_provable_kvs) {
     dbUpdates[DBKeyManipulator::genNonProvableBlockDbKey(blockId, k)] = v;
+    if (blockId) {
```
dartdart26, line 400: Maybe it will be clearer if we compare fro greater than `INITIAL_GENESIS_BLOCK_ID` instead of 0.

kvbc/src/merkle_tree_db_adapter.cpp
f-squirrel: done

kvbc/src/merkle_tree_db_adapter.cpp
f-squirrel: Done

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -387,6 +397,14 @@ SetOfKeyValuePairs DBAdapter::lastReachableBlockDbUpdates(const SetOfKeyValuePai
...
+      const auto &stale_kv = getValueForNonProvableKey(k, blockId - 1);
+      if (stale_kv) {
+        const auto &nonProvableStaleKey = DBKeyManipulator::genNonProvableStaleDbKey(
+            DBKeyManipulator::genNonProvableBlockDbKey(stale_kv.value().second, k), blockId);
```
dartdart26, line 404: Nit: I think it will be easier to read if we write these keys on separate lines and naming them. Another thing that comes to mind is that `block` in the name of `genNonProvableBlockDbKey()` might be redundant.

kvbc/include/merkle_tree_db_adapter.h
f-squirrel: done

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26, line 401: Something that came up after discussions with @andrewjstone is that if we don't save non-provable keys in the block node, then we will need to seek in `getRawBlock()` instead of doing direct lookups for them. We have seen some potential performance issues with iterators and also`getRawBlock()` itself. What I am saying is we might consider handling non-provable similarly to provable ones, i.e. writing them in the block node.

kvbc/src/merkle_tree_db_adapter.cpp
f-squirrel: Done

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -623,8 +648,7 @@ KeysVector DBAdapter::lastReachableBlockKeyDeletes(BlockId blockId) const {
   // Clear the stale index for the last version.
   add(keysToDelete, staleIndexKeysForVersion(blockNode.stateRootVersion));
 
-  // TODO(DD): Add to keysToDelete the NonProvable keys and NonProvableStale(if any).
-  // key_type|NonProvableStale|stale_since_version|
+  add(keysToDelete, staleIndexKeysForBlock(blockId));
```
dartdart26, line 651: Can we add a comment about what we are doing and be explicit that the line above adds the tree stale keys (provable) and that one is for non-provable ones?

kvbc/src/merkle_tree_db_adapter.cpp
f-squirrel: The binding to a const reference guarantees no-copy according to the standard, i.e. better performance.
Since the object is only passed by reference and never copied, I think we can use it.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26, line 651: If I am not mistaken, we are not actually deleting the non-provable keys. Instead, we are only deleting the stale indices for non-provable keys. I think that's also related to the fact we don't put non-provable keys in the block node and we might reconsider doing that as I mentioned above.

f-squirrel: Done

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -642,10 +666,15 @@ KeysVector DBAdapter::genesisBlockKeyDeletes(BlockId blockId) const {
     keysToDelete.push_back(DBKeyManipulator::extractKeyFromStaleKey(staleKey));
   }
 
-  // TODO(DD): Iterate over the NonProvableStale keys and add them to keysToDelete.
-
   // Clear the stale index for the last version.
   add(keysToDelete, staleKeys);
+  const auto &nonProvableKeys = staleIndexKeysForBlock(blockId);
+
+  // Clear the non-provable stale index for the last version.
```
dartdart26, line 673: I think that comment only relates to the `add()` call. Maybe we can be more explicit and separate the lines in a way it is clearer.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: In practice, there will be RVO and there won't be any copies. I am willing to trade for readability :)

kvbc/src/merkle_tree_db_adapter.cpp
f-squirrel: It was discussed offline, the implementation will stay the same for now.

kvbc/include/merkle_tree_key_manipulator.h
```diff
@@ -55,9 +56,15 @@ class DBKeyManipulator {
   // Extract the stale since version from a stale node index key.
   static sparse_merkle::Version extractVersionFromStaleKey(const Key &key);
 
+  // Extract the stale since block id from a stale node index key.
+  static BlockId extractBlockIdFromNonProvableStaleKey(const Key &key);
+
   // Extract the actual key from the stale node index key.
   static Key extractKeyFromStaleKey(const Key &key);
```
dartdart26, line 63: Does it make sense to rename that so it reflects it operates on Provable keys?

kvbc/include/merkle_tree_key_manipulator.h
```diff
@@ -55,9 +56,15 @@ class DBKeyManipulator {
   // Extract the stale since version from a stale node index key.
   static sparse_merkle::Version extractVersionFromStaleKey(const Key &key);
```
dartdart26, line 57: Does it make sense to rename that so it reflects it operates on Provable keys?

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -285,7 +295,7 @@ RawBlock DBAdapter::getRawBlock(const BlockId &blockId) const {
 
   // loookup keys |NonProvable|blockid|
   // add to the keyValues
-  const auto &non_provable_key = DBKeyManipulator::genNonProvableBlockDbKey(blockId, Sliver{});
+  const auto &non_provable_key = DBKeyManipulator::genNonProvableDbKey(blockId, Sliver{});
```
dartdart26, line 298: Does it make sense to name that `firstNonProvableForBlock`?

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -359,34 +369,43 @@ SetOfKeyValuePairs DBAdapter::lastReachableBlockDbUpdates(const SetOfKeyValuePai
...
 
   // Block node with updates and actually deleted keys.
   dbUpdates[DBKeyManipulator::genBlockDbKey(blockId)] =
-      createBlockNode(provable_kvs, actuallyDeleted, blockId, parentBlockDigestFuture.get());
-
-  for (const auto &[k, v] : non_provable_kvs) {
-    dbUpdates[DBKeyManipulator::genNonProvableBlockDbKey(blockId, k)] = v;
+      createBlockNode(provableKvPairs, actuallyDeleted, blockId, parentBlockDigestFuture.get());
+
+  // Marking non-provable keys as stale
```
dartdart26, line 398: Nit: I think that comment belongs a bit lower as here we are both adding the non-provable key and marking it stale.

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -285,7 +295,7 @@ RawBlock DBAdapter::getRawBlock(const BlockId &blockId) const {
 
   // loookup keys |NonProvable|blockid|
   // add to the keyValues
-  const auto &non_provable_key = DBKeyManipulator::genNonProvableBlockDbKey(blockId, Sliver{});
+  const auto &non_provable_key = DBKeyManipulator::genNonProvableDbKey(blockId, Sliver{});
   auto iter = db_->getIteratorGuard();
 
   for (auto kvp = iter->seekAtLeast(non_provable_key);
```
dartdart26, line 301: Sorry I haven't noticed that before, but it seems this will loop until it encounters a different key type/subtype. I think it should stop also when a new version for the same key type/subtype is encountered.

Could we maybe use `nonProvableKeysForBlock()` for that instead of a raw loop here?

Please correct me if I miss something.

kvbc/src/merkle_tree_db_adapter.cpp
f-squirrel: good idea, done

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -1189,6 +1200,135 @@ TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_index) {
   ASSERT_FALSE(hasProvableStaleIndexKeysSince(adapter.getDb(), 2));
 }
 
+TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_key_deletion_for_non_provable_keys) {
+  const Sliver skey1{"key1"};
+  const Sliver skey2{"key2"};
+  concord::kvbc::v2MerkleTree::DBAdapter::NonProvableKeySet non_provable_key_set = {skey1, skey2};
+  auto adapter = DBAdapter{GetParam()->newEmptyDb(), true, non_provable_key_set};
```
dartdart26, line 1207: Nit: camelCase naming.

kvbc/src/merkle_tree_db_adapter.cpp
f-squirrel: Good catch, fixed the condition!
The method `nonProvableKeysForBlock()` returns the `KeyVector` which means we'll need to lookup value for each key.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -1189,6 +1200,135 @@ TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_index) {
...
+                                                std::make_pair(skey2, Sliver{"sk2v3"})}),
+            4);
+
+  // Verify that the leaf key for key1 at version 1 is available.
+  ASSERT_TRUE(adapter.getDb()->has(ver1StaleLeafKey1).isOK());
+
+  // Verify that the leaf key for key2 at version 1 is available.
+  ASSERT_TRUE(adapter.getDb()->has(ver1StaleLeafKey2).isOK());
+  //
+  // Verify that the non-provable key for skey1 at block 1 is available.
```
dartdart26, line 1242: I think that comment is incorrect and it is equivalent to a check for `ver1StaleLeafKey1`. Also applies to the line below.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -1189,6 +1200,135 @@ TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_index) {
...
+  ASSERT_TRUE(adapter.getDb()->has(DBKeyManipulator::genDataDbKey(key1, 1)).isOK());
+
+  // Verify that the non-provable key for skey2 at block 1 is available.
+  ASSERT_TRUE(adapter.getDb()->has(DBKeyManipulator::genDataDbKey(key2, 1)).isOK());
+
+  // Make sure the are internal keys for tree version 1 before deletion.
+  ASSERT_TRUE(hasInternalKeysForVersion(adapter.getDb(), 1));
+
+  // Make sure the are non-provable keys for block 1 before deletion.
```
dartdart26, line 1251: I think that comment is incorrect as the code seems to be checking that there are stale index keys since blocks 3 and 4.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -1189,6 +1200,135 @@ TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_index) {
   ASSERT_FALSE(hasProvableStaleIndexKeysSince(adapter.getDb(), 2));
 }
 
+TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_key_deletion_for_non_provable_keys) {
```
dartdart26, line 1203: I think it might make sense to write this test without any provable (tree) keys as we already have such a test. Using non-provable keys only will simplify it.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: Fixed, this is why I do not like comments: the code gets updated while comments not:)

kvbc/test/sparse_merkle_storage/serialization_unit_test.cpp
```diff
@@ -108,18 +108,47 @@ TEST(key_manipulator, non_provable_key) {
...
 TEST(key_manipulator, extract_key_from_non_provable_key) {
   const auto non_prov_key = Sliver{"non-provable"};
-  const auto key = DBKeyManipulator::genNonProvableBlockDbKey(defaultBlockId, non_prov_key);
+  const auto key = DBKeyManipulator::genNonProvableDbKey(defaultBlockId, non_prov_key);
   const auto expected_key = DBKeyManipulator::extractKeyFromNonProvableKey(key);
   ASSERT_EQ(expected_key, non_prov_key);
 }
 
+TEST(key_manipulator, non_provable_stale_key) {
+  const auto non_prov_key = Sliver{"non-provable"};
```
dartdart26, line 124: Nit: camelCase.

kvbc/test/sparse_merkle_storage/serialization_unit_test.cpp
```diff
@@ -108,18 +108,47 @@ TEST(key_manipulator, non_provable_key) {
...
+  const auto block_key = DBKeyManipulator::genNonProvableDbKey(defaultBlockId, non_prov_key);
```
dartdart26, line 125: I think `block_key` doesn't reflect a `block` key. Maybe can be named differently.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -1189,6 +1200,135 @@ TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_index) {
...
+
+  // Make sure non-provable stale keys for blocks that have been deleted
+  ASSERT_FALSE(hasNonProvableStaleIndexKeysSince(adapter.getDb(), 3));
+  ASSERT_FALSE(hasNonProvableStaleIndexKeysSince(adapter.getDb(), 2));
+  ASSERT_FALSE(hasNonProvableStaleIndexKeysSince(adapter.getDb(), 1));
+
+  ASSERT_TRUE(hasNonProvableStaleIndexKeysSince(adapter.getDb(), 4));
+}
+
+TEST_P(db_adapter_custom_blockchain, delete_blocks_with_non_provable_keys) {
```
dartdart26, line 1283: Nice test!

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: Done: removed the provable keys.

kvbc/src/merkle_tree_db_adapter.cpp
dartdart26: Oh, yes, got it.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -1189,6 +1200,108 @@ TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_index) {
...
+            4);
+
+  // Verify that the non-provable key for skey1 at block 1 is available.
+  ASSERT_TRUE(adapter.getDb()->has(ver1StaleNonProvableKey1).isOK());
+
+  // Verify that the non-provable key for skey2 at block 1 is available.
+  ASSERT_TRUE(adapter.getDb()->has(ver1StaleNonProvableKey2).isOK());
+
+  // Make sure the are non-provable keys for blocks 3, 4 before deletion.
+  ASSERT_TRUE(hasNonProvableStaleIndexKeysSince(adapter.getDb(), 3));
```
dartdart26, line 1234: What is the rationale for checking if there are keys for block 1 and block 2 and stale indices for block 3 and 4?

f-squirrel: I am updating the keys by blocks 3 and 4, so I am checking that the stale keys are actually created.

kvbc/src/merkle_tree_db_adapter.cpp
```diff
@@ -357,35 +406,33 @@ SetOfKeyValuePairs DBAdapter::lastReachableBlockDbUpdates(const SetOfKeyValuePai
...
     // Key updates.
     const auto updateBatch =
-        smTree_.update(provable_kvs, KeysVector{std::cbegin(actuallyDeleted), std::cend(actuallyDeleted)});
+        smTree_.update(provableKvPairs, KeysVector{std::cbegin(actuallyDeleted), std::cend(actuallyDeleted)});
     const auto batchDbUpdates = batchToDbUpdates(updateBatch, blockId);
     dbUpdates.insert(std::cbegin(batchDbUpdates), std::cend(batchDbUpdates));
   }
 
   // Block node with updates and actually deleted keys.
+  // TODO(DD): put all the keys here (provable and non-provable)
```
dartdart26, line 421: I think this comment is no longer relevant.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -212,34 +223,132 @@ TEST_P(db_adapter_custom_blockchain, add_block_with_non_provable_keys) {
...
+  ASSERT_EQ(initialTreeVersion, adapter.getStateVersion());
+  ASSERT_EQ(initialHash, adapter.getStateHash());
+
+  // Make sure that adding blocks with provable keys affects the Merkle Tree state
+  adapter.addBlock(SetOfKeyValuePairs{{Sliver{"provable-key"}, Sliver{"provable-value"}}});
+  ASSERT_EQ(initialTreeVersion + 1, adapter.getStateVersion());
+  ASSERT_NE(initialHash, adapter.getStateHash());
+}
+
+TEST_P(db_adapter_custom_blockchain, raw_block_with_non_provable_and_provable_keys) {
```
dartdart26, line 277: Does it make sense to add blocks that include both provable and non-provable keys (in addition to provable-only and non-provable only blocks)?

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -248,23 +357,126 @@ TEST_P(db_adapter_custom_blockchain, add_block_throws_if_deletes_contain_non_pro
...
+      ASSERT_EQ(blockId, i + 1);
+    }
+    {
+      const auto &[value, blockId] = adapter.getValue(skey2, i + 1);
+      ASSERT_EQ(value, Sliver{"value2" + std::to_string(i)});
+      ASSERT_EQ(blockId, i + 1);
+    }
+  }
+
+  // Make sure that when the last reachable block id is provided, the most updated value is returned
```
dartdart26, line 384: Nice test!

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -248,23 +357,126 @@ TEST_P(db_adapter_custom_blockchain, add_block_throws_if_deletes_contain_non_pro
   ASSERT_THROW(adapter.addBlock(updates, deletes), std::runtime_error);
 }
 
+TEST_P(db_adapter_custom_blockchain, getValue_finds_overlapping_keys) {
```
dartdart26, line 360: Nit: Maybe the name can indicate that we are testing non-provable keys.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -248,23 +357,126 @@ TEST_P(db_adapter_custom_blockchain, add_block_throws_if_deletes_contain_non_pro
...
+    ASSERT_EQ(value, Sliver{"value1" + std::to_string(numBlocks - 1)});
+  }
+  {
+    const auto &[value, blockId] = adapter.getValue(skey2, adapter.getLastReachableBlockId());
+    ASSERT_EQ(blockId, numBlocks);
+    ASSERT_EQ(value, Sliver{"value2" + std::to_string(numBlocks - 1)});
+  }
+}
+
+TEST_P(db_adapter_custom_blockchain, getValue_finds_non_provable_keys) {
```
dartdart26, line 397: I guess this test checks single-versioned non-provable keys, in contrast to the test above. Maybe we can name it such that it is explicit. Do I miss something?

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -248,23 +357,126 @@ TEST_P(db_adapter_custom_blockchain, add_block_throws_if_deletes_contain_non_pro
...
+TEST_P(db_adapter_custom_blockchain, getValue_finds_non_provable_keys_in_bc_with_mixed_keys) {
+  const Sliver skey1{"key1"};
+  const Sliver skey2{"key2"};
+  const Sliver svalue1{"value1"};
+  const Sliver svalue2{"value2"};
+  concord::kvbc::v2MerkleTree::DBAdapter::NonProvableKeySet non_provable_key_set = {skey1, skey2};
+
+  auto adapter = DBAdapter{GetParam()->newEmptyDb(), true, non_provable_key_set};
+  const size_t blockNum = 5u;
+  std::vector<SetOfKeyValuePairs> updates;
```
dartdart26, line 435: This vector seems unused.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -248,23 +357,126 @@ TEST_P(db_adapter_custom_blockchain, add_block_throws_if_deletes_contain_non_pro
...
   {
-    concord::kvbc::v2MerkleTree::DBAdapter::NonProvableKeySet non_provable_key_set = {skey1, skey2};
-    auto adapter = DBAdapter{db, true, non_provable_key_set};
-    for (const auto &k : non_provable_key_set) {
-      auto kvp = adapter.getValue(k, adapter.getLastReachableBlockId());
-      ASSERT_EQ(kvp.first, defaultSliver);
+    auto adapter = DBAdapter{db, true, {skey1, skey2}};
+    for (const auto &k : updates) {
+      auto kvp = adapter.getValue(k.first, adapter.getLastReachableBlockId());
+      ASSERT_EQ(kvp.first, k.second);
```
dartdart26, line 479: Nit: Maybe could name `kvb` and `k` more explicitly, for readability.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -1271,6 +1483,118 @@ TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_index) {
...
 }
 
+TEST_P(db_adapter_custom_blockchain, delete_genesis_and_verify_stale_key_deletion_with_non_provable_keys) {
+  const Sliver skey1{"key1"};
+  const Sliver skey2{"key2"};
+  concord::kvbc::v2MerkleTree::DBAdapter::NonProvableKeySet nonProvableKeySet = {skey1, skey2};
+  auto adapter = DBAdapter{GetParam()->newEmptyDb(), true, nonProvableKeySet};
+
+  const auto ver1StaleNonProvableKey1 = DBKeyManipulator::genNonProvableDbKey(1, skey1);
+  const auto ver1StaleNonProvableKey2 = DBKeyManipulator::genNonProvableDbKey(1, skey2);
```
dartdart26, line 1493: Nit: I think these are not used.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
dartdart26: Got it, thanks!

kvbc/include/merkle_tree_db_adapter.h
```diff
@@ -182,9 +183,11 @@ class DBAdapter : public IDbAdapter {
 
   block::detail::Node getBlockNode(BlockId blockId) const;
 
-  KeysVector staleIndexKeysForVersion(const sparse_merkle::Version &version) const;
+  KeysVector staleIndexNonProvableKeysForBlock(BlockId blockId) const;
 
-  KeysVector internalKeysForVersion(const sparse_merkle::Version &version) const;
+  KeysVector staleIndexProvableKeysForVersion(const sparse_merkle::Version &version) const;
+
+  KeysVector internalProvableKeysForVersion(const sparse_merkle::Version &version) const;
```
dartdart26, line 190: Nit: In terms of naming, I was thinking that maybe we can separate keys to:

- internalTreeKeysForVersion
- staleTreeKeysForVersion
- staleNonProvableKeysForBlock

Rationale is that internal keys are always tree ones and, technically, are not provable. Just something that came to mind.

kvbc/include/merkle_tree_key_manipulator.h
```diff
@@ -53,10 +54,16 @@ class DBKeyManipulator {
   static sparse_merkle::Hash extractHashFromLeafKey(const Key &key);
 
   // Extract the stale since version from a stale node index key.
-  static sparse_merkle::Version extractVersionFromStaleKey(const Key &key);
+  static sparse_merkle::Version extractVersionFromProvableStaleKey(const Key &key);
```
dartdart26, line 57: Nit: As above, can be called `extractVersionFromTreeStaleKey`. Just a suggestion, please only change if you think it is better.

kvbc/include/merkle_tree_key_manipulator.h
```diff
@@ -53,10 +54,16 @@ class DBKeyManipulator {
...
+
+  // Extract the stale since block id from a stale node index key.
+  static BlockId extractBlockIdFromNonProvableStaleKey(const Key &key);
 
   // Extract the actual key from the stale node index key.
-  static Key extractKeyFromStaleKey(const Key &key);
+  static Key extractKeyFromProvableStaleKey(const Key &key);
```
dartdart26, line 63: Nit: As above, can be called `extractKeyFromTreeStaleKey`. Just a suggestion, please only change if you think it is better.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -184,7 +195,7 @@ TEST_P(db_adapter_custom_blockchain, add_block_return) {
   }
 }
 
-TEST_P(db_adapter_custom_blockchain, add_block_with_non_provable_keys) {
+TEST_P(db_adapter_custom_blockchain, add_block_with_provable_and_non_provable_keys) {
```
dartdart26, line 198: Nit: I think it will be clearer if we use different naming from `defaultSliver` to designate provable keys and values, in general.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
```diff
@@ -212,34 +223,132 @@ TEST_P(db_adapter_custom_blockchain, add_block_with_non_provable_keys) {
...
+TEST_P(db_adapter_custom_blockchain, raw_block_only_with_provable_keys) {
+  const Sliver key1{"key1"};
+  const Sliver key2{"key2"};
+  const Sliver key3{"key3"};
+
+  const Sliver value1{"value1"};
+  const Sliver value2{"value2"};
+  const Sliver value3{"value3"};
+
+  auto adapter = DBAdapter{GetParam()->newEmptyDb(), true};
```
dartdart26, line 337: Do we need to configure a set of non-provable keys in this test and then verify that provable ones work? What I am getting at is that the case without non-provable keys (default empty param) is covered in other tests.

dartdart26 (approved):
Thank you so much for supporting non-provable keys @f-squirrel ! Great work!

I have put minor comments, mainly on naming and tests. Please resolve as you see fit.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: It is weird that the compiler lets this code compile. I was sure we disabled unused variables.
Fixed.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: Since we may have blocks with both types of keys, I think it is safer to test it as well.
Mostly for regression testing's sake.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: Renamed the keys, changed values for non-provables.

kvbc/test/sparse_merkle_storage/db_adapter_unit_test.cpp
f-squirrel: Added the set of non-provable keys.

kvbc/include/merkle_tree_db_adapter.h
f-squirrel: I am afraid this PR is being in review for a long time.
If you don't mind we can change it as a separate PR.

PR #1044, opened 2020-11-13
Integrate replica asks to leave view msg in replica status msg

HristoStaykov:
During Status Reports Replicas indicate from which peers they have complaints for the View they are in.
By doing so peers are able to send them the complaints (ReplicaAsksToLeaveViewMSg-s) they are missing.
Resending of the complaint is disabled, because it can be resend during Status Reports.

### DIFF
--- a/bftengine/src/bftengine/messages/ReplicaStatusMsg.cpp
+++ b/bftengine/src/bftengine/messages/ReplicaStatusMsg.cpp
@@ -15,6 +15,7 @@
 
 namespace bftEngine {
 namespace impl {
+namespace {
 static uint8_t powersOf2[] = {
     0x1,
     0x2,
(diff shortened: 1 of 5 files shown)


teoparvanov (approved):
LGTM, thanks! @yontyon could you also have a look, as this PR addresses a concern you had raised about the safety of the view change protocol modifications?

bftengine/src/bftengine/messages/ReplicaStatusMsg.cpp
```diff
@@ -26,18 +27,26 @@ static uint8_t powersOf2[] = {
     0x80,
 };
 
+static const int kWorkWindowBitMaskSize = (kWorkWindowSize + 7) / 8;
```
f-squirrel, line 30: Nit: variables defined in an anonymous namespace are automatically `static`.
If both `kWorkWindowSize` is a compile-time variable, then `kWorkWindowBitMaskSize` can be a `constexpr`.

f-squirrel (approved):
LGTM!

yontyon (approved):
LGTM!
Thanks!

PR #1045, opened 2020-11-13
Wrap RocksDB for safety and column family support

dartdart26:
Wrap the RocksDB client library for:
 * improved memory management - use value semantics instead of raw
   pointers
 * improved error reporting - done via exceptions instead of status
   codes
 * introduce support for column families

Support the following RocksDB features:
 * single-key put/get/del interfaces
 * write batches
 * column families
 * iterators

More features can be added in the future, based on demand.

No dependency on Sliver is required. Instead, use templates that support
any container or span via the data() and size() method pair.

Interoperability with the existing IDBClient interface is perserved via
the asIDBClient() method.

Add a size() method to Sliver so that it behaves like a contiguous
memory container, i.e. having data() and size() members (similar to
std::vector, std::string, etc.).

Move storage_test_common.h from kvbc to storage for reuse. Provide unit
tests for the newly-introduced wrappers.

### DIFF
--- a/storage/include/rocksdb/native_client.h
+++ b/storage/include/rocksdb/native_client.h
@@ -13,13 +13,17 @@
 
 #pragma once
 
+#include "assertUtils.hpp"
 #include "client.h"
 #include "storage/db_interface.h"
 
+#include <rocksdb/env.h>
 #include <rocksdb/slice.h>
 #include <rocksdb/status.h>
+#include <rocksdb/utilities/options_util.h>
 #include <rocksdb/write_batch.h>
 
+#include <algorithm>
 #include <exception>
 #include <map>
 #include <memory>
@@ -38,7 +42,7 @@
 //
 // The Span template parameters to methods allow conversion from any type that has data() and size()
 // members. The data() member should return a char pointer. That includes std::string, std::string_view,
-// std::vector<char>, std::span<char>, Sliver, etc. Raw pointers without size are not supported.
+// std::vector<char>, std::span<char>, Sliver, ::rocksdb::Slice, etc. Raw pointers without size are not supported.
 //
 // Note1: All methods on all classes throw on errors.
 // Note2: All methods without a column family parameter work with the default column family.
@@ -55,7 +59,7 @@ class NativeClient;
 
 class WriteBatch {
  public:
-  WriteBatch(const std::shared_ptr<const NativeClient> &);
+  WriteBatch(const std::shared_ptr<const NativeClient> &) noexcept;
 
   template <typename KeySpan, typename ValueSpan>
   void put(const std::string &cFamily, const KeySpan &key, const ValueSpan &value);
@@ -125,7 +129,7 @@ class Iterator {
   std::string_view valueView() const;
 
  private:
-  Iterator(std::unique_ptr<::rocksdb::Iterator> iter);
+  Iterator(std::unique_ptr<::rocksdb::Iterator> iter) noexcept;
   static std::string sliceToString(const ::rocksdb::Slice &);
   static std::string_view sliceToStringView(const ::rocksdb::Slice &);
 
@@ -181,10 +185,16 @@ class NativeClient : public std::enable_shared_from_this<NativeClient> {
   static std::string defaultColumnFamily();
   static std::unordered_set<std::string> columnFamilies(const std::string &path);
   std::unordered_set<std::string> columnFamilies() const;
-  // If is not an error if it already exists.
-  void createColumnFamily(const std::string &cFamily);
+  // Throws if the column family already exists.
+  void createColumnFamily(const std::string &cFamily,
+                          const ::rocksdb::ColumnFamilyOptions &options = ::rocksdb::ColumnFamilyOptions{});
+  // Return the column family options for an existing column family.
+  ::rocksdb::ColumnFamilyOptions columnFamilyOptions(const std::string &cFamily) const;
   // Drops a column family and its data. It is not an error if the column family doesn't exist.
   void dropColumnFamily(const std::string &cFamily);
+
+  // Return the DB path.
+  const std::string &path() const { return path_; }
 
  private:
   NativeClient(const std::string &path, bool readOnly);
@@ -220,7 +230,7 @@ class NativeClient : public std::enable_shared_from_this<NativeClient> {
 
   ::rocksdb::ColumnFamilyHandle *defaultColumnFamilyHandle() const;
   ::rocksdb::ColumnFamilyHandle *columnFamilyHandle(const std::string &cFamily) const;
-  CfUniquePtr createColumnFamilyHandle(const std::string &cFamily);
+  CfUniquePtr createColumnFamilyHandle(const std::string &cFamily, const ::rocksdb::ColumnFamilyOptions &options);
 
  private:
   std::string path_;
@@ -232,7 +242,7 @@ class NativeClient : public std::enable_shared_from_this<NativeClient> {
   friend class Iterator;
 };
 
-inline WriteBatch::WriteBatch(const std::shared_ptr<const NativeClient> &client) : client_{client} {}
+inline WriteBatch::WriteBatch(const std::shared_ptr<const NativeClient> &client) noexcept : client_{client} {}
 
 template <typename KeySpan, typename ValueSpan>
 void WriteBatch::put(const std::string &cFamily, const KeySpan &key, const ValueSpan &value) {
@@ -242,7 +252,7 @@ void WriteBatch::put(const std::string &cFamily, const KeySpan &key, const Value
 
 template <typename KeySpan, typename ValueSpan>
 void WriteBatch::put(const KeySpan &key, const ValueSpan &value) {
-  return put(NativeClient::defaultColumnFamily(), key, value);
+  put(NativeClient::defaultColumnFamily(), key, value);
 }
 
 template <typename KeySpan>
@@ -253,7 +263,7 @@ void WriteBatch::del(const std::string &cFamily, const KeySpan &key) {
 
 template <typename KeySpan>
 void WriteBatch::del(const KeySpan &key) {
-  return del(NativeClient::defaultColumnFamily(), key);
+  del(NativeClient::defaultColumnFamily(), key);
 }
 
 template <typename BeginSpan, typename EndSpan>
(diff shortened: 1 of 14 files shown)


storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+template <typename KeySpan, typename ValueSpan>
+void WriteBatch::put(const KeySpan &key, const ValueSpan &value) {
+  return put(NativeClient::defaultColumnFamily(), key, value);
+}
+
+template <typename KeySpan>
+void WriteBatch::del(const std::string &cFamily, const KeySpan &key) {
+  auto s = batch_.Delete(client_->columnFamilyHandle(cFamily), NativeClient::toSlice(key));
+  NativeClient::throwOnError("batch del failed"sv, std::move(s));
+}
```
cloudnoize, line 252: Do we expect that a client of this class will rarely give us to delete keys that do not exist, or it can actually be a frequent scenario?

and how should we behave when catching such exception? 
Do we need and have a way to differentiate it from other exceptions?  

I'm trying to understand if exceptions will complicate our implementation on scenarios that might be frequent

cloudnoize (commented):
I like this implementation, it's clean and elegant.
I have doubts regarding the usage of exceptions on any error i.e any return code other than ok which might complicate our code in flows which we expect to tolerate non-ok return codes.

storage/include/rocksdb/native_client.h
dartdart26: Deleting a key that doesn't exist is not an error for both single-key deletes and batches. I've put a comment on the methods and added tests - `del_non_existent_key_in_batch_is_not_an_error` and `del_non_existent_key_is_not_an_error`. This is identical to RocksDB's behaviour.

dartdart26:
> I like this implementation, it's clean and elegant.
> I have doubts regarding the usage of exceptions on any error i.e any return code other than ok which might complicate our code in flows which we expect to tolerate non-ok return codes.

Could you please share an example error type that we might want to report differently (no exception)?

cloudnoize:
> > I like this implementation, it's clean and elegant.
> > I have doubts regarding the usage of exceptions on any error i.e any return code other than ok which might complicate our code in flows which we expect to tolerate non-ok return codes.
> 
> Could you please share an example error type that we might want to report differently (no exception)?

I see that deleting a non-existing key is not reported as an error and that you handle empty gets without exceptions, those are the obvious candidates.
Besides those, I don't have anything concrete in mind.

cloudnoize:
LGTM !

cloudnoize (approved):

andrewjstone:
> > > I like this implementation, it's clean and elegant.
> > > I have doubts regarding the usage of exceptions on any error i.e any return code other than ok which might complicate our code in flows which we expect to tolerate non-ok return codes.
> > 
> > 
> > Could you please share an example error type that we might want to report differently (no exception)?
> 
> I see that deleting a non-existing key is not reported as an error and that you handle empty gets without exceptions, those are the obvious candidates.
> Besides those, I don't have anything concrete in mind.

Petar and I have talked about using [std::expected](https://github.com/TartanLlama/expected) in the past for error handling, rather than exceptions. It forces *a* caller to handle errors and doesn't have the problem with stack unwinding. There are many Cppcon presentations on such result types.

However, that seems like a larger architectural discussion that we probably shouldn't make here.

storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+  // Get iterators from a consistent database state across multiple column families. The order of the returned iterators
+  // match the families input.
+  std::vector<Iterator> getIterators(const std::vector<std::string> &cFamilies) const;
+
+  // Column family management.
+  static std::string defaultColumnFamily();
+  static std::unordered_set<std::string> columnFamilies(const std::string &path);
+  std::unordered_set<std::string> columnFamilies() const;
+  // If is not an error if it already exists.
+  void createColumnFamily(const std::string &cFamily);
```
andrewjstone, line 185: What about configuring a column family?

storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+
+  using CfUniquePtr = std::unique_ptr<::rocksdb::ColumnFamilyHandle, CfDeleter>;
+
+  ::rocksdb::ColumnFamilyHandle *defaultColumnFamilyHandle() const;
+  ::rocksdb::ColumnFamilyHandle *columnFamilyHandle(const std::string &cFamily) const;
+  CfUniquePtr createColumnFamilyHandle(const std::string &cFamily);
+
+ private:
+  std::string path_;
+  // Keep cf_handles_ after the client_ as column family handles need to be deleted before deleting the client. See
```
andrewjstone, line 227: Clever :)

storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+
+template <typename KeySpan>
+void WriteBatch::del(const std::string &cFamily, const KeySpan &key) {
+  auto s = batch_.Delete(client_->columnFamilyHandle(cFamily), NativeClient::toSlice(key));
+  NativeClient::throwOnError("batch del failed"sv, std::move(s));
+}
+
+template <typename KeySpan>
+void WriteBatch::del(const KeySpan &key) {
+  return del(NativeClient::defaultColumnFamily(), key);
```
andrewjstone, line 256: Nit: no need for `return`. Same in other methods.

storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+}
+
+template <typename BeginSpan, typename EndSpan>
+void WriteBatch::delRange(const BeginSpan &beginKey, const EndSpan &endKey) {
+  return delRange(client_->defaultColumnFamily(), beginKey, endKey);
+}
+
+template <typename Container>
+void putInBatch(WriteBatch &batch, const std::string &cFamily, const Container &cont) {
+  for (auto &[key, value] : cont) {
```
andrewjstone, line 273: Nit: `const auto &`

storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+}
+
+template <typename Container>
+void putInBatch(WriteBatch &batch, const Container &cont) {
+  return putInBatch(batch, NativeClient::defaultColumnFamily(), cont);
+}
+
+template <typename Container>
+void delInBatch(WriteBatch &batch, const std::string &cFamily, const Container &cont) {
+  for (auto &key : cont) {
```
andrewjstone, line 285: `const auto &key` ?

storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+
+inline std::vector<Iterator> NativeClient::getIterators(const std::vector<std::string> &cFamilies) const {
+  auto cfHandles = std::vector<::rocksdb::ColumnFamilyHandle *>{};
+  for (auto &cf : cFamilies) {
+    cfHandles.push_back(columnFamilyHandle(cf));
+  }
+  auto rawPtrIterators = std::vector<::rocksdb::Iterator *>{};
+  auto uniquePtrIterators = std::vector<std::unique_ptr<::rocksdb::Iterator>>{};
+  auto status = client_->dbInstance_->NewIterators(::rocksdb::ReadOptions{}, cfHandles, &rawPtrIterators);
+  // Are we supposed to delete the iterators if error is returned...
```
andrewjstone, line 452: Maybe mark this with a `TODO:` to make it stand out.

storage/include/rocksdb/native_client.h
```diff
@@ -0,0 +1,542 @@
...
+  for (auto iter : rawPtrIterators) {
+    uniquePtrIterators.push_back(std::unique_ptr<::rocksdb::Iterator>{iter});
+  }
+  throwOnError("get multiple iterators failed"sv, std::move(status));
+  auto ret = std::vector<Iterator>{};
+  for (auto &iter : uniquePtrIterators) {
+    ret.push_back(std::move(iter));
```
andrewjstone, line 459: If you move this into the prior loop, you optimize for the positive case and only pay the penalty on the negative case. Otherwise we do a double loop on every successful call.

storage/test/native_rocksdb_client_test.cpp
```diff
@@ -0,0 +1,621 @@
...
+  ASSERT_EQ(*dbValue2, value2);
+}
+
+TEST_F(native_rocksdb_test, put_in_batch_in_2_families) {
+  const auto cf1 = "cf1"s;
+  const auto cf2 = "cf2"s;
+  db->createColumnFamily(cf1);
+  db->createColumnFamily(cf2);
+  auto batch = db->getBatch();
+  db->put(cf1, key1, value1);
```
andrewjstone, line 179: Don't these need to be `batch.put` calls?

andrewjstone (approved):
Looks great @dartdart26. Only a few minor comments and questions. It will be excellent to have column family support finally :)

storage/include/rocksdb/native_client.h
dartdart26: Fixed.

storage/include/rocksdb/native_client.h
dartdart26: Fixed.

storage/include/rocksdb/native_client.h
dartdart26: Fixed.

storage/include/rocksdb/native_client.h
dartdart26: You are right, thanks! Not sure why I used two loops, one loop is enough.

storage/include/rocksdb/native_client.h
dartdart26: Thanks! I have changed the comment and explained the rationale. In addition, I have checked the implementation in RocksDB to verify what we do is safe. However, `NewIterators()`'s interface is lacking and is really bad, IMO...

storage/test/native_rocksdb_client_test.cpp
dartdart26: What a great catch, thanks, fixing!

dartdart26:
> Looks great @dartdart26. Only a few minor comments and questions. It will be excellent to have column family support finally :)

Thanks for the great comments!

storage/include/rocksdb/native_client.h
dartdart26: Good point. I have added that in a second commit. It required a number of changes to initialisation.

PR #1047, opened 2020-11-16
Reduce working window size from 300 Sequence numbers to 30.

HristoStaykov:
Fix tests to take in consideration the new size.


util/src/throughput.cpp
```diff
@@ -81,8 +81,8 @@ void Throughput::Stats::reset() {
 
 void Throughput::Stats::calcThroughput() {
   auto duration = std::chrono::steady_clock::now() - start_time_;
-  results_.elapsed_time_ms_ = std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
-  results_.throughput_ = static_cast<uint64_t>((1000 * results_.num_processed_items_) / results_.elapsed_time_ms_);
+  results_.elapsed_time_us_ = std::chrono::duration_cast<std::chrono::microseconds>(duration).count();
+  results_.throughput_ = static_cast<uint64_t>((1000000 * results_.num_processed_items_) / results_.elapsed_time_us_);
```
glevkovich, line 85: I suggest asserting here as well to check for non-zero duration.
```
ConcordAssert(duration != 0);
```

glevkovich (approved):

robem:
@HristoStaykov changing the default values is a big change. Please modify the PR description to include motivation and reasoning for the new values.

PR #1048, opened 2020-11-16
Re-enable rapidcheck property tests

dartdart26:
Fix the way we use CMake to find the rapidcheck library. Installing it
via `make install` is different than the previous Conan method and that
has led to the tests not compiling and not running.

Call Database::cleanup() explicitly in the db_adapter_property_test in
order to start with a fresh DB on every input.


f-squirrel (approved):
Awesome, thank you, @dartdart26 !

andrewjstone (approved):
Looks great. Thanks for the fix!

PR #1050, opened 2020-11-17
Key-view file handling improvements (cherry pick from 0.8)

cloudnoize:
- writing to the key view file only when a real update occurs.

- store a backup


HristoStaykov (approved):
LGTM, thanks!

PR #1051, opened 2020-11-17
Add atomics metrics

yontyon:
In this PR we add new sub metrics types:
* AtomicCounter
* AtomicGauge

These kinds of metrics are thread-safe even when multiple threads are updating the same component.
In addition, a higher level (such as concord) will see them as ordinary counters and gauges. 

Lastly, we changed the storage metrics to be atomic as they accessed from multiple threads (pre-execution and ordinary execution)

### DIFF
--- a/storage/include/storage/storage_metrics.h
+++ b/storage/include/storage/storage_metrics.h
@@ -29,6 +29,8 @@ class InMemoryStorageMetrics {
   concordMetrics::AtomicCounterHandle total_written_bytes_;
 
  private:
+  // update_metrics_ has to be the last declared member due to some synchronization issue. If you have to put something
+  // after it, then explicit release it on the destructor
   std::unique_ptr<concord::util::PeriodicCall> update_metrics_ = nullptr;
 
  public:
@@ -63,6 +65,8 @@ class RocksDbStorageMetrics {
 
   std::shared_ptr<::rocksdb::SstFileManager> sstFm;
   std::shared_ptr<::rocksdb::Statistics> statistics;
+  // update_metrics_ has to be the last declared member due to some synchronization issue. If you have to put something
+  // after it, then explicit release it on the destructor
   std::unique_ptr<concord::util::PeriodicCall> update_metrics_ = nullptr;
 
  public:
@@ -92,7 +96,6 @@ class RocksDbStorageMetrics {
                                ::rocksdb::Tickers::COMPACT_WRITE_BYTES,
                                ::rocksdb::Tickers::FLUSH_WRITE_BYTES,
                                ::rocksdb::Tickers::STALL_MICROS}) {}
-
   void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) {
     rocksdb_comp_.SetAggregator(aggregator);
   }
--- a/storage/include/storage/storage_metrics.h
+++ b/storage/include/storage/storage_metrics.h
@@ -93,7 +93,7 @@ class RocksDbStorageMetrics {
                                ::rocksdb::Tickers::COMPACT_WRITE_BYTES,
                                ::rocksdb::Tickers::FLUSH_WRITE_BYTES,
                                ::rocksdb::Tickers::STALL_MICROS}) {}
-  ~RocksDbStorageMetrics() { update_metrics_.reset();}
+  ~RocksDbStorageMetrics() { update_metrics_.reset(); }
   void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) {
     rocksdb_comp_.SetAggregator(aggregator);
   }
--- a/storage/include/storage/storage_metrics.h
+++ b/storage/include/storage/storage_metrics.h
@@ -39,9 +39,8 @@ class InMemoryStorageMetrics {
         keys_writes_(metrics_.RegisterAtomicCounter("storage_inmemory_total_written_keys")),
         total_written_bytes_(metrics_.RegisterAtomicCounter("storage_inmemory_total_written_bytes")) {
     metrics_.Register();
-    update_metrics_ = std::make_unique<concord::util::PeriodicCall>([&]() { updateMetrics(); }, 1000);
+    update_metrics_ = std::make_unique<concord::util::PeriodicCall>([this]() { updateMetrics(); }, 1000);
   }
-  ~InMemoryStorageMetrics() { update_metrics_.release(); }
   void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) { metrics_.SetAggregator(aggregator); }
   void updateMetrics() { metrics_.UpdateAggregator(); }
 };
@@ -98,12 +97,11 @@ class RocksDbStorageMetrics {
     rocksdb_comp_.SetAggregator(aggregator);
   }
 
-  ~RocksDbStorageMetrics() { update_metrics_.release(); }
   void setMetricsDataSources(std::shared_ptr<::rocksdb::SstFileManager> sourceSstFm,
                              std::shared_ptr<::rocksdb::Statistics> sourceStatistics) {
     sstFm = sourceSstFm;
     statistics = sourceStatistics;
-    update_metrics_ = std::make_unique<concord::util::PeriodicCall>([&]() { updateMetrics(); }, 1000);
+    update_metrics_ = std::make_unique<concord::util::PeriodicCall>([this]() { updateMetrics(); }, 1000);
   }
 
   void updateMetrics() {
--- a/storage/include/storage/storage_metrics.h
+++ b/storage/include/storage/storage_metrics.h
@@ -21,14 +21,15 @@ namespace storage {
  * Recall that the in memory db is quite simple and therefor it has only few relevant metrics to collect.
  */
 class InMemoryStorageMetrics {
-  std::unique_ptr<concord::util::PeriodicCall> update_metrics_;
-
  public:
   concordMetrics::Component metrics_;
   concordMetrics::AtomicCounterHandle keys_reads_;
   concordMetrics::AtomicCounterHandle total_read_bytes_;
   concordMetrics::AtomicCounterHandle keys_writes_;
   concordMetrics::AtomicCounterHandle total_written_bytes_;
+
+ private:
+  std::unique_ptr<concord::util::PeriodicCall> update_metrics_ = nullptr;
 
  public:
   InMemoryStorageMetrics()
@@ -56,13 +57,13 @@ class InMemoryStorageMetrics {
  * operation) the overhead of that approach is negligible.
  */
 class RocksDbStorageMetrics {
-  std::unique_ptr<concord::util::PeriodicCall> update_metrics_;
   concordMetrics::Component rocksdb_comp_;
   std::unordered_map<::rocksdb::Tickers, concordMetrics::AtomicGaugeHandle> active_tickers_;
   concordMetrics::AtomicGaugeHandle total_db_disk_size_;
 
   std::shared_ptr<::rocksdb::SstFileManager> sstFm;
   std::shared_ptr<::rocksdb::Statistics> statistics;
+  std::unique_ptr<concord::util::PeriodicCall> update_metrics_ = nullptr;
 
  public:
   RocksDbStorageMetrics(const std::vector<::rocksdb::Tickers>& tickers)
--- a/storage/include/storage/storage_metrics.h
+++ b/storage/include/storage/storage_metrics.h
@@ -41,6 +41,7 @@ class InMemoryStorageMetrics {
     metrics_.Register();
     update_metrics_ = std::make_unique<concord::util::PeriodicCall>([&]() { updateMetrics(); }, 1000);
   }
+  ~InMemoryStorageMetrics() { update_metrics_.release(); }
   void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) { metrics_.SetAggregator(aggregator); }
   void updateMetrics() { metrics_.UpdateAggregator(); }
 };
@@ -97,6 +98,7 @@ class RocksDbStorageMetrics {
     rocksdb_comp_.SetAggregator(aggregator);
   }
 
+  ~RocksDbStorageMetrics() { update_metrics_.release(); }
   void setMetricsDataSources(std::shared_ptr<::rocksdb::SstFileManager> sourceSstFm,
                              std::shared_ptr<::rocksdb::Statistics> sourceStatistics) {
     sstFm = sourceSstFm;
--- a/storage/include/storage/storage_metrics.h
+++ b/storage/include/storage/storage_metrics.h
@@ -2,6 +2,8 @@
 
 #pragma once
 
+#include <thread>
+#include <chrono>
 #include "Metrics.hpp"
 #ifdef USE_ROCKSDB
 #include <rocksdb/statistics.h>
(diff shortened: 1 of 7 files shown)


util/include/Metrics.hpp
```diff
@@ -104,6 +107,40 @@ class Counter {
   uint64_t val_;
 };
 
+class AtomicCounter {
```
f-squirrel, line 110: Can we make the original classes such as `Counter` and `Gauge` templated and supply the variable type via the template?
Later we can do `using AtomicGauge = Gauge<std::atomic_uint64_t>;
WDYT?

storage/include/storage/storage_metrics.h
```diff
@@ -64,8 +64,8 @@ class InMemoryStorageMetrics {
  */
 class RocksDbStorageMetrics {
```
dartdart26, line 65: I think we need to think about the `last_metrics_update_` variable too as it might be changed concurrently.

util/include/Metrics.hpp
yontyon: I tried, but it is too complicated and I'm not sure it worth the huge refactor we will have to do (also in concord)

f-squirrel: During an offline conversation, we've come up with another idea:
The original `Gauge` can be changed to:
```
template<typename CounterT>
class BasicGauge {
....
```
Then:
```
using Gauge = BasicGauge<uint64_t>;
using AtomicGauge = BasicGauge<std::atomic_uint64_t>;
```
As a result, all the old functionality using the original `Gauge` won't be affected.

andrewjstone: I like it. Great collaboration!

yontyon: done

storage/include/storage/storage_metrics.h
yontyon: done

util/include/Metrics.hpp
```diff
@@ -23,13 +24,20 @@
 #include "Statistics.hpp"
 
 namespace concordMetrics {
+template <class T>
+class basic_gauge;
```
f-squirrel, line 28: Nit: According to the code convention, classes have to be `PascalCase` => `BasicGauge`.

f-squirrel (approved):
Thank you, @yontyon for the changes!

util/include/Metrics.hpp
```diff
@@ -65,17 +73,41 @@ class Aggregator {
...
 // A Gauge is a an integer value that shows the current value of something. It
 // can only be varied by directly setting and getting it.
-class Gauge {
+template <class T>
+class basic_gauge {
  public:
-  explicit Gauge(const uint64_t val) : val_(val) {}
-
+  explicit basic_gauge(const uint64_t val) : val_(val) {}
+  basic_gauge(const basic_gauge& gauge) { val_ = (unsigned long)gauge.val_; }
```
andrewjstone, line 80: I believe what you are doing here is undefined behavior. Presumably the reason you needed the cast was because atomics have deleted copy constructors. This should tell you that the type is not copyable on purpose. 

Why do you need to be able to copy metrics types in the first place?

In case you actually did need to copy them for some reason, you can use [constexpr if](https://en.cppreference.com/w/cpp/language/if#Constexpr_If) and then call `gauge.val_.load()` rather than casting. 

My recommendation is to remove the copy constructor/assignment ops altogether, and try to break out of the habit of using C-style casts :)

util/include/Metrics.hpp
```diff
@@ -65,17 +73,41 @@ class Aggregator {
...
+  T val_;
+};
+
+template <class T>
+class basic_counter {
+ public:
+  explicit basic_counter(const uint64_t val) : val_(val) {}
+  basic_counter(const basic_counter& counter) { val_ = (unsigned long)counter.val_; }
+  basic_counter& operator=(const basic_counter& counter) {
+    val_ = (unsigned long)counter.val_;
```
andrewjstone, line 100: Same as above. This cast looks wrong.

andrewjstone (approved):
Looks good @yontyon. Please just fix the copy assignment one way or the other :)

util/include/Metrics.hpp
yontyon: atomic_uint64_t overloads the ‘unsigned long’ operator, sonI believe the code is correct.
I need a copy constructor because we do copy the values every time we call to updateAggregator

storage/include/storage/storage_metrics.h
dartdart26: Does the mutex also protect the atomic counters? If not, maybe we can also use an atomic for the last update time, because that mutex will be locked often. Please correct me if I am wrong, I am using my phone and I don’t see the code really well.

yontyon: The mutex protects only the `last_metrics_update_`. We cannot use atomic because it supports only primitive types.
This mutex will be taken only once in a while when we update the aggregator (timer depended)

util/include/Metrics.hpp
yontyon: done

storage/include/storage/storage_metrics.h
dartdart26: Don’t we acquire it every time on tryToUpdateMetrics()? That is called on every DB operation, right? I am still on my phone, sorry if I miss something.

Maybe we can use another type for time that can be atomic, just a thought.

util/include/Metrics.hpp
andrewjstone: I guess I'm a bit confused. Yes, there is an operator, but does this code call it? I believe the C-style cast just returns the memory pointed to by `gauge.val_` without running any code.

andrewjstone: Related. Unsigned long doesn't necessarily map to uint64_t.

andrewjstone: Ah, it appears the cast does call the operator. My bad. :)  Can we cast it as a `uint64_t` though to ensure it works on all architectures?

storage/include/storage/storage_metrics.h
yontyon: you right, I will fix it

util/include/Metrics.hpp
yontyon: OK, I will

storage/include/storage/storage_metrics.h
yontyon: done. 
I've changed the update to be operated based instead of time based. And made it atomic.

storage/include/storage/storage_metrics.h
```diff
@@ -3,7 +3,7 @@
 #pragma once
 
 #include "Metrics.hpp"
-
+#include <mutex>
```
dartdart26, line 6: I think we don't need that include anymore.

storage/include/storage/storage_metrics.h
```diff
@@ -21,32 +21,31 @@ namespace storage {
  * Recall that the in memory db is quite simple and therefor it has only few relevant metrics to collect.
  */
 class InMemoryStorageMetrics {
-  std::chrono::seconds metrics_update_interval_ = std::chrono::seconds(5);  // TODO: move to configuration
-  std::chrono::steady_clock::time_point last_metrics_update_ = std::chrono::steady_clock::now();
+  std::atomic<uint32_t> last_metrics_update_ = 0;
+  uint32_t metrics_update_interval_ = 100;  // We update the metrics once in 100 operations.
```
dartdart26, line 25: We were actually updating the metrics every 5 seconds before. If we have lots of operations, doing it every 100 operations will update much more often. I am not sure if that is problematic or not, but just to note it. Maybe we can update less often?

yontyon: I think that if we have many operations we want to see the differences, so it makes sense to update on an operation based rather on time-based, WDY?
I can increase it ofcourse

storage/include/storage/storage_metrics.h
yontyon: will be removed

storage/include/storage/storage_metrics.h
```diff
@@ -21,32 +21,31 @@ namespace storage {
...
+        total_written_bytes_(metrics_.RegisterAtomicCounter("storage_inmemory_total_written_bytes")) {
     metrics_.Register();
   }
 
   void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) { metrics_.SetAggregator(aggregator); }
 
   void tryToUpdateMetrics() {
-    if (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - last_metrics_update_) >
-        metrics_update_interval_) {
+    if (last_metrics_update_++ > metrics_update_interval_) {
```
dartdart26, line 46: I think, in general, using atomics in that way doesn't produce the exact behaviour that is intended, because the reset code that sets to 0 is not atomic with the counter check.

Maybe we can exploit unsigned integer overflow (for simplicity) and just do something like:

```
// Every 256 ops or std::atomic<std::uint16_t> for every 16K ops.
std::atomic<std::uint8_t> last_metrics_update_{0};
...
if (last_metrics_update_++ == 0) {
   ...
}
```

Or maybe another method is using `compare_exchange_strong()` in a loop.

I am not an expert on atomics, maybe your original code is fine and I am just rambling. @andrewjstone could you please assist?

yontyon: no, I think u right. I missed it.
I will use % instead.
I also rethinking it and I think we need also a time-based updating method for a case where we don't make a lot of operations

storage/include/storage/storage_metrics.h
dartdart26: Yes, I agree with you. I am just unsure of how to define the operation count.

storage/include/storage/storage_metrics.h
dartdart26: Great idea, much better!

Nit: Maybe something minor is to use std::uint64_t for `last_metrics_update_`  and `metrics_update_interval_ ` to avoid overflow and make sure we almost always update with the same period (assuming we will almost never overflow 64 bit). And maybe we can rename them to indicate counts and not intervals.

Thanks to @HristoStaykov for assistance!

storage/include/storage/storage_metrics.h
yontyon: @andrewjstone, any idea?

andrewjstone: We use 100ms updates in the replica thread. I would increase this to at least 1000. It really depends on throughput how frequently you want to do it unfortunately. You are going to have sync overhead too much if updating too quickly, or you are not going to see the stats you want if updating too slowly. This is why time based works better. 

I think the problem you ran into was that you are doing the check and update from multiple threads. You don't have to do that though. You can just have a timer in a separate thread that runs an update callback. This costs a thread for a timer, but most of the time it's just sleeping waiting on TimerFd in linux. Probably doable with std::async.

yontyon: done.
I preferred to create my own thread instead of using std::async because I don't want to spawn a new thread on every invocation.

storage/include/storage/storage_metrics.h
yontyon: I've taken @andrewjstone  idea and implemented another thread that updates the metrics on time basis

storage/include/storage/storage_metrics.h
dartdart26: Agreed, having a separate thread makes it simpler and allows us to set periods. Thanks!

storage/include/storage/storage_metrics.h
dartdart26: Great idea, thanks!

storage/include/storage/storage_metrics.h
```diff
@@ -14,41 +15,67 @@
 namespace concord {
 namespace storage {
 
+class StorageMetrics {
+ public:
+  virtual void updateMetrics() = 0;
+};
+
+class StorageMetricsUpdator {
```
dartdart26, line 23: I think this looks good and does the job. I would suggest to use the constructor and destructor to start/stop, without explicit start() and close() calls. That would allow us to not implement the destructors of the metrics objects (InMemoryStorageMetrics and RocksDbStorageMetrics). Additionally, users won't be able to start multiple times and we don' t need to guard against that case.

I was thinking another option might be to have an utility class that runs a function every X seconds, like so:

```
class PeriodicCall {
 public:
   PeriodicCall(const std::function<void()>& f, const std::chrono::milliseconds &interval);
   ~PeriodicCall(); // sets the stop flag and calls t.join()

 private:
  std::thread t;
  std::chrono::milliseconds interval;
  std::atomic_bool stop{false};
};
```
We can then use PeriodicCall as a member in the metrics classes (as now), but avoid inheritance, i.e. remove the `StorageMetrics` interface. For example:

```
InMemoryStorageMetrics::InMemoryStorageMetrics() :
  ... 
  periodic_call_{[this] () { updateMetrics(); }, metrics_update_interval_},
  ...
```

storage/include/storage/storage_metrics.h
```diff
@@ -14,41 +15,67 @@
...
+  StorageMetrics& storage_metrics_;
+  std::thread t;
+  std::chrono::milliseconds metrics_update_interval_ = std::chrono::milliseconds(100);  // TODO: read from configuration
```
dartdart26, line 26: Should we use a bigger interval?

storage/include/storage/storage_metrics.h
```diff
@@ -14,41 +15,67 @@
...
+  void close() {
+    if (!active) return;
+    active = false;
+    t.join();
+  }
+  void start() {
+    active = true;
+    t = std::thread([&] {
+      while (active) {
+        storage_metrics_.updateMetrics();
```
dartdart26, line 41: Can `updateMetrics()` throw? We can maybe be explicit about whether that can happen and, if yes, what is the desired behaviour.

storage/include/storage/storage_metrics.h
dartdart26: Actually, we want to be calling `updateMetrics()` after the constructor has done its job. Is that the reason why you have an explicit `start()` method? I think we can still use constructors by having a default one that doesn't start the thread and using the default move constructor, something like:
```
metrics_updator_ = StorageMetricsUpdator{this};
```

or

```
// Not setting it in the initialiser list.
periodic_call_ = PeriodicCall{[this] () { updateMetrics(); }, metrics_update_interval_};
```

storage/include/storage/storage_metrics.h
```diff
@@ -64,24 +58,22 @@ class InMemoryStorageMetrics {
...
-  concordMetrics::GaugeHandle total_db_disk_size_;
+  std::unordered_map<::rocksdb::Tickers, concordMetrics::AtomicGaugeHandle> active_tickers_;
+  concordMetrics::AtomicGaugeHandle total_db_disk_size_;
 
   std::shared_ptr<::rocksdb::SstFileManager> sstFm;
   std::shared_ptr<::rocksdb::Statistics> statistics;
-
-  std::chrono::seconds metrics_update_interval_ = std::chrono::seconds(5);  // TODO: move to configuration
-  std::chrono::steady_clock::time_point last_metrics_update_ = std::chrono::steady_clock::now();
+  std::unique_ptr<concord::util::PeriodicCall> update_metrics_ = nullptr;
```
dartdart26, line 66: I was thinking that when we run the `InMemoryStorageMetrics`  and `RocksDbStorageMetrics` destructors, we want to ensure that the thread stops before we destruct metrics members as the thread might still call them. Therefore, I think it might make sense to put an explicit comment saying that the `PeriodicCall` pointer should be the last member in these classes, ensuring the thread has stopped. What do you think? Did I get it correctly?

dartdart26 (approved):
Thanks @yontyon for these improvements, appreciate you doing it! I think code is safer and clearer now!

storage/include/storage/storage_metrics.h
yontyon: I have added a comment

PR #1052, opened 2020-11-17
Add histograms for the pre-execution signatures creation and verification

yuliasherman:


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -117,7 +117,7 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
   threadPool_.start(numOfThreads);
   LOG_INFO(logger(),
            KVLOG(firstClientId, numOfClients_, maxPreExecResultSize_, preExecReqStatusCheckPeriodMilli_, numOfThreads));
-  RequestProcessingState::init(numOfRequiredReplies());
+  RequestProcessingState::init(numOfRequiredReplies(), &histograms_);
```
f-squirrel, line 120: Nit: Since the pointer `preProcessorHistograms_` is never checked for `NULL`, I guess it is not optional and has to be always provided.
Maybe it can be passed as a reference to reflect that it is mandatory?
WDYT?

yuliasherman: I can pass the pointer by reference, but not an object. Is that the proposal?

f-squirrel: The proposal is to make the init in the following way:
`void RequestProcessingState::init(uint16_t numOfRequiredReplies, PreProcessorRecorder &histograms)`
and hold the histograms as `preprocessor::PreProcessorRecorder& preProcessorHistograms_`.
Again, just a suggestion.

yuliasherman: It's static...

f-squirrel: Oh, my bad, then please ignore.

f-squirrel (approved):

PR #1053, opened 2020-11-18
GitHub actions split by compiler and release/debug

KTeofanidis:
In order to allow for more control when a retry is needed split up github actions by compiler and release/debug


andrewjstone (commented):
Thanks for doing this @KTeofanidis . I don't have expertise in github actions and so I added @f-squirrel as a reviewer to do a detailed review.

One thign I really don't like here is that we have essentially copy pasted the majority of workflows and now anytime we need to make a change, we need to update every single file. Is there instead a way to include a base yaml file and only modify the relevant bits in each individual workflow?

f-squirrel (approved):
Unfortunately, GHA [neither supports](https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234/62) retriggering a single job nor provides a simple way to import a workflow.

LGTM, @KTeofanidis !

Since we're going to have multiple workflows, please add new badges to the README.

teoparvanov (approved):
LGTM, thanks @KTeofanidis !

PR #1055, opened 2020-11-19
Cache lastReachable/latest/genesis block IDs

dartdart26:
Analisys by @andrewjstone has shown that RocksDB iterator seeks for
lastReachable/latest/genesis block IDs are expensive. Therefore, since
there is an option to cache them in memory, provide that in this PR.

Remove the related histograms as these methods should be called very
rarely given caching.

Change existing unit tests and add ones in order to verify caching
doesn't change behavior.

In addition, since support for non-provable keys is still not ready,
remove the seek for non-provable keys in getRawBlock() and the
corresponding unit test. Future commits will provide full support for
non-provable keys. Until then, non-provable functionality will not work.

### DIFF
--- a/kvbc/include/merkle_tree_db_adapter.h
+++ b/kvbc/include/merkle_tree_db_adapter.h
@@ -50,6 +50,7 @@ namespace concord::kvbc::v2MerkleTree {
 class DBAdapter : public IDbAdapter {
  public:
   using NonProvableKeySet = std::unordered_set<Key>;
+
   // Unless explicitly turned off, the constructor will try to link the blockchain with any blocks in the temporary
   // state transfer chain. This is done so that the DBAdapter will operate correctly in case a crash or an abnormal
   // shutdown has occurred prior to startup (construction). Only a single DBAdapter instance should operate on a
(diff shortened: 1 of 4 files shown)


yontyon (approved):
LGTM! 
Thanks, @dartdart26

kvbc/include/merkle_tree_db_adapter.h
```diff
@@ -192,8 +193,17 @@ class DBAdapter : public IDbAdapter {
...
+  void deleteGenesisBlock();
+
   void deleteKeysForBlock(const KeysVector &keys, BlockId blockId) const;
 
+  BlockId loadGenesisBlockId() const;
+
+  BlockId loadLastReachableBlockId() const;
+
+  // Return std::nullopt if no temporary ST blocks are present.
+  std::optional<BlockId> loadLatestTempSTBlockId() const;
```
andrewjstone, line 205: This little change makes the code more understandable. Brings 😂

andrewjstone (approved):
Looks great @dartdart26! Thanks for the quick turnaround on this important perf fix!

PR #1056, opened 2020-11-20
Improve concord logs

yontyon:
In this PR we present some log improvements as well as changing the order of answering to RFMD.

The benefit from changing the order is that the requester replica has a higher probability to handle first the PrePrepare and commit messages and thus to avoid handling unnecessary messages.


andrewjstone (approved):
LGTM. Thanks!

PR #1058, opened 2020-11-23
Add context to State Transfer logs for causal order

oded282:
In this PR, I added a context to State Transfer protocol logs to parse the logs in a causal order.

**Challenge**
Although most of the protocol runs sequentially, there is no global clock for the replicas. Thus, two replicas may log at the same time, but the time stamp will be different in each of the logs. Hence, we can't rely on the time stamp for the causal ordering of the logs.

**Solution**
Use our monotonically increasing sequence number. In each onMessage function (main event in the protocol) I added the sequence number to the logs with scoped MDC.

**Comments**
The sequence number is generated in the replica that triggered the protocol and passed through the messages to the rest of the replicas. The information in the logs, together with the sequence number should help create causal order.


toly-kournik:
@oded282 nice job! 
Several comments:

1. There's a position log parameter for sequence number 7-th column - %X{sn} , which could be used for the seq num.
2. Pipe '\|' is a log fields delimiter, please don't add any.
3. Exactly for the reason you've mentioned (no global clock), there are sequence number for the each ST message( based on a timestamp, therefore monotonically increasing).
4. I would use these sequence numbers instead of introducing the global sequence and an order notations.

toly-kournik (approved):
LGTM. Short and elegant

PR #1060, opened 2020-11-26
Define the update interface for key categorization

dartdart26:
Key categorization allows users of the Key-Value Blockchain to select
different key categories for their keys. Every key category type has its
own properties and supports a specific set of functionalities. Having
the ability to put keys into different categories enables optimal
storage and performance, based on the requirements users have for
lifetime, lookup and provability.

Aims at:
 * ability to have different properties and functionalities per category
 * support multiple categories of the same type
 * ability to extend key categorization with new category types

### DIFF
--- a/kvbc/include/categorization/details.h
+++ b/kvbc/include/categorization/details.h
@@ -1,0 +1,74 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include "kv_types.hpp"
+#include "sparse_merkle/base_types.h"
+
+#include <optional>
+#include <set>
+#include <unordered_map>
+#include <variant>
+
+namespace concord::kvbc::categorization::detail {
+
+struct MerkleUpdateInfo {
+  struct KeyFlags {
+    bool deleted{false};
+  };
+
+  std::unordered_map<Key, KeyFlags> keys;
+  std::array<std::uint8_t, 32> hash;
+  sparse_merkle::Version version;
+  std::string category_id;
+
+  template <typename T>
+  bool operator<(const T &other) const {
+    return category_id < other.category_id;
+  }
+};
+
+struct KeyValueUpdateInfo {
+  struct KeyFlags {
+    bool deleted{false};
+
+    // Applicable to non-deleted keys only.
+    bool stale_on_update{false};
+  };
+
+  std::unordered_map<Key, KeyFlags> keys;
+  std::optional<std::array<std::uint8_t, 32>> hash;
+  std::string category_id;
+
+  template <typename T>
+  bool operator<(const T &other) const {
+    return category_id < other.category_id;
+  }
+};
+
+struct CategorizedKeyValueUpdateInfo {
+  std::set<Key> keys;
+  std::optional<std::array<std::uint8_t, 32>> hash;
+  std::string category_id;
+
+  template <typename T>
+  bool operator<(const T &other) const {
+    return category_id < other.category_id;
+  }
+};
+
+// Represents updates returned by categories and consumbed by the blockchain (for use in blocks).
+using CategoryUpdateInfos = std::set<std::variant<MerkleUpdateInfo, KeyValueUpdateInfo, CategorizedKeyValueUpdateInfo>>;
+
+}  // namespace concord::kvbc::categorization::detail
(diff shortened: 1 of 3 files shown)


kvbc/include/categorization/updates.h
```diff
@@ -0,0 +1,116 @@
...
+    // Implies that a single categorized update per block is supported.
+    return false;
+  }
+
+  template <typename T>
+  bool operator<(const T &) const {
+    // Since empty category IDs are not supported, use the empty one to designate a categorized update. Implies that the
+    // categorized update is always `less` than other types and that a single categorized update per block is supported.
+    return true;
+  }
```
cloudnoize, line 59: Cool!

kvbc/include/categorization/updates.h
```diff
@@ -0,0 +1,116 @@
...
+// Categorized key-value updates for KVBC blocks. Every category supports different properties and functionalities.
+// Note1: Empty category IDs are invalid and not supported.
+// Note2: Using the same category ID for different category types is an error.
+namespace concord::kvbc::categorization {
+
+// Updates for multiple categories.
+// Persists key-values directly in the underlying key-value store. All key-values are marked stale during the update
+// itself. Explicit deletes are not supported.
+struct CategorizedKeyValueUpdate {
+  struct CategorizedValue {
```
cloudnoize, line 35: I find this name confusing, maybe we can think if more meaningful name

kvbc/include/categorization/details.h
```diff
@@ -0,0 +1,74 @@
...
+  }
+};
+
+struct KeyValueUpdateInfo {
+  struct KeyFlags {
+    bool deleted{false};
+
+    // Applicable to non-deleted keys only.
+    bool stale_on_update{false};
+  };
```
cloudnoize, line 48: Does this bool should be part of the Block as well i.e. the Node for this category will  contain a map<key, KeyFlags >

cloudnoize (approved):

kvbc/include/categorization/details.h
dartdart26: I think yes, because of State Transfer. We should be able to do the exact same operations (marking stale) on the receiving replica. Please correct me if I miss something.

kvbc/include/categorization/updates.h
dartdart26: I agree it might not be ideal. Any suggestions?

Another thing we can do is to maybe avoid such an update type and require users to update a key across (potentially) multiple groups explicitly.

dartdart26: Renamed to `SharedKeyValueUpdates` and `SharedValue` as suggested by @cloudnoize in an offline discussion.

dartdart26:
The types in `details.h` can be replaced by CMF types as they have to be serialized in blocks anyway. Will change in the next commit.

kvbc/include/categorization/details.h
```diff
@@ -0,0 +1,74 @@
...
+  std::optional<std::array<std::uint8_t, 32>> hash;
+  std::string category_id;
+
+  template <typename T>
+  bool operator<(const T &other) const {
+    return category_id < other.category_id;
+  }
+};
+
+struct ShareddKeyValueUpdatesInfo {
```
f-squirrel, line 60: Shared[d]KeyValueUpdatesInfo => SharedKeyValueUpdatesInfo

kvbc/include/categorization/details.h
```diff
@@ -0,0 +1,74 @@
...
+  std::string category_id;
+
+  template <typename T>
+  bool operator<(const T &other) const {
+    return category_id < other.category_id;
+  }
+};
+
+// Represents updates returned by categories and consumbed by the blockchain (for use in blocks).
+using CategoryUpdatesInfo = std::set<std::variant<MerkleUpdateInfos, KeyValueUpdateInfos, ShareddKeyValueUpdatesInfo>>;
```
f-squirrel, line 72: Is `MerkleUpdateInfos` a container?
Nit: "Informations" sounds a bit awkward.

f-squirrel (approved):
Great job!

kvbc/include/categorization/details.h
cloudnoize: I guess that if the stale flag is not part of the value struct that is stored in DB then it's essential.

Anyway, it got me thinking about ST in general, and currently, the raw block enables us to call the Merkle add block as any other client would.
We should ensure that the new Raw block will contain all the information needed to construct the per group update struct as defined in updates.h

kvbc/include/categorization/details.h
dartdart26: Will fix, thanks!

kvbc/include/categorization/details.h
dartdart26: Will change that type to CMF and rename, thanks!

kvbc/include/categorization/details.h
dartdart26: Yes, agreed.

dartdart26:
Closing that one as the effort and discussion is continuing in https://github.com/vmware/concord-bft/pull/1068.

PR #1061, opened 2020-11-26
Make test stable to include in CI.

HristoStaykov:
Because the View Change timer is set to tick with a precision of 1/2
of the View Change Timeout, we usually get it to timeout later
than the specified period. In order to make the test pass consistently
we need to start the Replicas with a delay of at least 3/2 of the
View Change period.

### DIFF
--- a/tests/apollo/test_skvbc_chaotic_startup.py
+++ b/tests/apollo/test_skvbc_chaotic_startup.py
@@ -51,9 +51,8 @@ class SkvbcChaoticStartupTest(unittest.TestCase):
 
     __test__ = False  # so that PyTest ignores this test scenario
 
-    @unittest.skip("Unstable because of BC-5164")
     @with_trio
-    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    @with_bft_network(start_replica_cmd_with_vc_timeout("6500"), selected_configs=lambda n, f, c: n == 7)
     @with_constant_load
     async def test_delayed_replicas_start_up(self, bft_network, skvbc, nursery):
         """
--- a/tests/apollo/test_skvbc_chaotic_startup.py
+++ b/tests/apollo/test_skvbc_chaotic_startup.py
@@ -52,7 +52,7 @@ class SkvbcChaoticStartupTest(unittest.TestCase):
     __test__ = False  # so that PyTest ignores this test scenario
 
     @with_trio
-    @with_bft_network(start_replica_cmd_with_vc_timeout("6500"), selected_configs=lambda n, f, c: n == 7)
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
     @with_constant_load
     async def test_delayed_replicas_start_up(self, bft_network, skvbc, constant_load):
         """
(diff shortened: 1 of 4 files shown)


tests/apollo/test_skvbc_chaotic_startup.py
```diff
@@ -51,9 +51,8 @@ class SkvbcChaoticStartupTest(unittest.TestCase):
 
     __test__ = False  # so that PyTest ignores this test scenario
 
-    @unittest.skip("Unstable because of BC-5164")
     @with_trio
-    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    @with_bft_network(start_replica_cmd_with_vc_timeout("6500"), selected_configs=lambda n, f, c: n == 7)
```
teoparvanov, line 55: I think instead, we should remove the check that a view change happened.
Following the recent improvements of the VC protocol, it should become much less likely, even during chaotic startup.

teoparvanov (changes requested):
Some thoughts on this...

tests/apollo/test_skvbc_chaotic_startup.py
HristoStaykov: During a chaotic startup whether we have a View change depends on the position of Replica 0 in the starting order.
If F+1 replicas start before it and with a delay equal to View Change timeout between, those F+1 will complain
and a View Change will happen. The recent improvements of the VC protocol make the Fast Path very likely in the
case of chaotic startup, because replicas that complain do not actually leave the View until F+1 complaints are
gathered from different Replicas. Unfortunately we still observe problems rarely (in my experiments roughly 1 in 60 runs)
where a Replica moves to a higher View than the others, and the reason for this is that it gathers 2F+2C+1 VC messages, but there is a delay for the New View message sent by the Primary. The following logs are from a recreation of the described situation:
`6|28-11-2020 19:36:25.142|INFO |concord.bft.viewchange|message-processing|||||void bftEngine::impl::ReplicaImp::onViewsChangeTimer(Timers::Handle)|Unable to activate the last agreed view (despite receiving 2f+2c+1 view change msgs). State transfer hasn't kicked-in for a while either. Asking to leave the current view:  curView: 1, timeSinceLastAgreedViewMilli: 10001, timeSinceLastStateTransferMilli: 30096
6|28-11-2020 19:36:25.142|INFO |concord.bft.viewchange|message-processing|||||void bftEngine::impl::ReplicaImp::MoveToHigherView(bftEngine::impl::ViewNum)| curView: 1, nextView: 2, wasInPrevViewNumber: False
6|28-11-2020 19:36:25.142|INFO |concord.bft.viewchange|message-processing|||||void bftEngine::impl::ReplicaImp::MoveToHigherView(bftEngine::impl::ViewNum)|Sending view change message.  newView: 2, wasInPrevViewNumber: False, newPrimary: 2, lastExecutedSeqNum: 0, lastStableSeqNum: 0
6|28-11-2020 19:36:25.144|INFO |concord.bft|message-processing|2|6738536019580157952|||void bftEngine::impl::ReplicaImp::onMessage(bftEngine::impl::ClientRequestMsg *)|ClientRequestMsg is ignored because current view is inactive.  reqSeqNum: 6738536019580157952, clientId: 17
6|28-11-2020 19:36:25.161|INFO |concord.bft.viewchange|message-processing|||||void bftEngine::impl::ReplicaImp::onMessage(bftEngine::impl::ViewChangeMsg *)| generatedReplicaId: 1, msg->newView(): 1, msg->lastStable(): 0, msg->numberOfElements(): 0, msgAdded: True
6|28-11-2020 19:36:25.162|INFO |concord.bft.viewchange|message-processing|||||void bftEngine::impl::ReplicaImp::onMessage(bftEngine::impl::ViewChangeMsg *)| maxKnownCorrectView: 1, maxKnownAgreedView: 1
6|28-11-2020 19:36:25.163|INFO |concord.bft.viewchange|message-processing|||||void bftEngine::impl::ReplicaImp::onMessage(bftEngine::impl::NewViewMsg *)| senderId: 1, msg->newView(): 1, msgAdded: True, curView: 2, viewsManager->viewIsActive(curView): 0
6|28-11-2020 19:36:25.163|INFO |concord.bft.viewchange|message-processing|||||bool bftEngine::impl::ReplicaImp::tryToEnterView()|Called viewsManager->tryToEnterView  curView: 2, lastStableSeqNum: 0, lastExecutedSeqNum: 0, enteredView: False`

teoparvanov (approved):
LGTM!

PR #1062, opened 2020-11-27
Test ReplicaAsksToLeaveView by isolating a replica 

KTeofanidis:
Test makes sure that a single isolated replica will ask for a view change, but it doesn't occur:
        1) Start all replicas
        2) Isolate a single replica for long enough to trigger a ReplicaAsksToLeaveViewMsg
        3) Let the replica rejoin the consensus
        4) Make sure a view change does not happen and the isolated replica
        rejoins the fast path in the existing view

### DIFF
--- a/tests/apollo/test_skvbc_view_change.py
+++ b/tests/apollo/test_skvbc_view_change.py
@@ -23,8 +23,6 @@ from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
 from util.skvbc_history_tracker import verify_linearizability
 from util import eliot_logging as log
 
-viewChangeTimeoutMilli = "10000"
-
 def start_replica_cmd(builddir, replica_id):
     """
     Return a command that starts an skvbc replica when passed to
@@ -32,6 +30,7 @@ def start_replica_cmd(builddir, replica_id):
     Note each arguments is an element in a list.
     """
     statusTimerMilli = "500"
+    viewChangeTimeoutMilli = "10000"
     path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
     return [path,
             "-k", KEY_FILE_PREFIX,
@@ -421,6 +420,11 @@ class SkvbcViewChangeTest(unittest.TestCase):
             bft_network.all_replicas(without={initial_primary}))
         isolated_replicas = set([isolated_node])
 
+        ask_to_leave_msg_count = await bft_network.get_metric(
+            isolated_node, bft_network, "Gauges", "sentReplicaAsksToLeaveViewMsg")
+        self.assertEqual(ask_to_leave_msg_count, 0,
+                         "Make sure the replica is ok before isolating it.")
+
         with net.ReplicaSubsetTwoWayIsolatingAdversary(bft_network, isolated_replicas) as adversary:
             adversary.interfere()
             try:
@@ -433,18 +437,11 @@ class SkvbcViewChangeTest(unittest.TestCase):
             except:
                 pass
 
-            # wait for 1.5 times the view change timeout to ensure the replica asks for a view change
-            await trio.sleep(1.5 * int(viewChangeTimeoutMilli) / 1000)
+            await bft_network.wait_for_replica_to_ask_for_view_change(replica_id=isolated_node)
 
-        # ensure the isolated replica has sent AsksToLeaveView requests
-        with trio.fail_after(seconds=30):
-            while True:
-                ask_to_leave_msg_count = await bft_network.get_metric(
-                    isolated_node, bft_network, "Counters", "sentReplicaAsksToLeaveViewMsgDueToStatus")
-                if ask_to_leave_msg_count > 0:
-                    break
-                else:
-                    await trio.sleep(.5)
+        # ensure the isolated replica has sent AsksToLeaveView requests after ending it's isolation
+        await self._wait_for_replica_to_ask_to_leave_view_due_to_status(
+            bft_network=bft_network, node=isolated_node)
 
         num_fast_req = 10
         async def write_req():
@@ -530,3 +527,13 @@ class SkvbcViewChangeTest(unittest.TestCase):
             crashed_replicas.add(crash_candidates[i])
 
         return crashed_replicas
+
+    async def _wait_for_replica_to_ask_to_leave_view_due_to_status(self, bft_network, node):
+        with trio.fail_after(seconds=30):
+            while True:
+                ask_to_leave_msg_count = await bft_network.get_metric(
+                    node, bft_network, "Counters", "sentReplicaAsksToLeaveViewMsgDueToStatus")
+                if ask_to_leave_msg_count > 0:
+                    break
+                else:
+                    await trio.sleep(.5)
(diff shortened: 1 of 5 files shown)


tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +395,69 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
             num_consecutive_failing_primaries = 2
         )
 
+    @unittest.skipIf(environ.get('BUILD_COMM_TCP_TLS', "").lower() == "true", "Unstable on CI (TCP/TLS only)")
+    @with_trio
+    @with_bft_network(start_replica_cmd, rotate_keys=True)
+    async def test_replica_asks_to_leave_view(self, bft_network, exchange_keys=True):
```
teoparvanov, line 401: remove `exchange_keys`, it's not used

tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +395,69 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
...
+                client.primary = None
+                for _ in range(5):
+                    msg = skvbc.write_req(
+                        [], [(skvbc.random_key(), skvbc.random_value())], 0)
+                    await client.write(msg)
+            except:
+                pass
+
+            # wait for 1.5 times the view change timeout to ensure the replica asks for a view change
+            await trio.sleep(int(viewChangeTimeoutMilli) / 666)
```
teoparvanov, line 437: When I combine the comment and the code, I get the equation `1,5 * x = x / 666`, which yields `x~31` (x, being the view change timeout). Something is fishy 😄

tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +395,69 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
...
+
+        num_fast_req = 10
+        async def write_req():
+            for _ in range(num_fast_req):
+                await skvbc.write_known_kv()
+
+        await bft_network.wait_for_fast_path_to_be_prevalent(
+            run_ops=lambda: write_req(), threshold=num_fast_req)
+
+        new_primary = await bft_network.get_current_primary()
```
teoparvanov, line 457: `new_primary` -> `current_primary`

tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +395,69 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
...
+    @unittest.skipIf(environ.get('BUILD_COMM_TCP_TLS', "").lower() == "true", "Unstable on CI (TCP/TLS only)")
+    @with_trio
+    @with_bft_network(start_replica_cmd, rotate_keys=True)
+    async def test_replica_asks_to_leave_view(self, bft_network, exchange_keys=True):
+        """
+        This test makes sure that a single isolated replica will ask for a view change, but it doesn't occur:
+        1) Start all replicas
+        3) Isolate a single replica for long enough to trigger a ReplicaAsksToLeaveViewMsg
+        4) Let the replica rejoin the consensus
+        5) Make sure a view change does not happen and the isolated replica
```
teoparvanov, line 407: nit - 3), 4), 5) -> 2), 3), 4)

tests/apollo/test_skvbc_view_change.py
KTeofanidis: 10000/666=15 seconds

teoparvanov: isn't it easier to say `trio.sleep(seconds = 1.5 * viewChangeTimeoutMilli / 1000)` ?

tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +395,69 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
...
+                client.primary = None
+                for _ in range(5):
+                    msg = skvbc.write_req(
+                        [], [(skvbc.random_key(), skvbc.random_value())], 0)
+                    await client.write(msg)
+            except:
+                pass
+
+            # wait for 1.5 times the view change timeout to ensure the replica asks for a view change
+            await trio.sleep(1.5 * int(viewChangeTimeoutMilli) / 1000)
```
andrewjstone, line 437: Is this just an "advisory" sleep? I want to ensure we don't *require* something to happen in this time, and as a result cause test failures if it doesn't happen. Even though the safety margin is large, it would be much preferable to use a "wait for" style here, rather than a failure further down the test because a sleep wasn't long enough.

KTeofanidis: It was unlikely to cause issues since we were waiting for the timeout, but I added a metric to make sure. An added advantage is that the wait is a bit shorter now.

tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +394,68 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
...
+        self.assertEqual(len(bft_network.procs), n,
+                         "Make sure all replicas are up initially.")
+
+        initial_primary = await bft_network.get_current_primary()
+
+        isolated_node = random.choice(
+            bft_network.all_replicas(without={initial_primary}))
+        isolated_replicas = set([isolated_node])
+
+        with net.ReplicaSubsetTwoWayIsolatingAdversary(bft_network, isolated_replicas) as adversary:
```
teoparvanov, line 423: Right before activating the adversary, let's first assert that `sentReplicaAsksToLeaveViewMsgDueToStatus` is `0`.

tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +394,68 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
...
+
+        # ensure the isolated replica has sent AsksToLeaveView requests after ending it's isolation
+        with trio.fail_after(seconds=30):
+            while True:
+                ask_to_leave_msg_count = await bft_network.get_metric(
+                    isolated_node, bft_network, "Counters", "sentReplicaAsksToLeaveViewMsgDueToStatus")
+                if ask_to_leave_msg_count > 0:
+                    break
+                else:
+                    await trio.sleep(.5)
```
teoparvanov, line 445: extract in a `self._wait_for_replica_to_ask_to_leave_view()` method

tests/apollo/test_skvbc_view_change.py
teoparvanov: @KTeofanidis what about this one? How can we check that the replica does not complain when the full BFT network is up and running?

Thanks!
-- Teo

tests/apollo/test_skvbc_view_change.py
```diff
@@ -394,6 +394,67 @@ async def test_single_vc_current_and_next_primaries_down(self, bft_network, trac
...
+
+        isolated_node = random.choice(
+            bft_network.all_replicas(without={initial_primary}))
+        isolated_replicas = set([isolated_node])
+
+        with net.ReplicaSubsetTwoWayIsolatingAdversary(bft_network, isolated_replicas) as adversary:
+            ask_to_leave_msg_count = await bft_network.get_metric(
+                isolated_node, bft_network, "Gauges", "sentReplicaAsksToLeaveViewMsg")
+            self.assertEqual(ask_to_leave_msg_count, 0,
+                             "Make sure the replica is ok before isolating it.")
```
teoparvanov, line 427: For better readability, can we make this check outside of the adversary scope?

teoparvanov (approved):
LGTM, thank you @KTeofanidis !

HristoStaykov (approved):
LGTM, great work @KTeofanidis !

PR #1063, opened 2020-11-27
Use big-endian and add fixed-sized lists in CMF

dartdart26:
Change the byte order of integer serialization in CMF to big-endian.
Rationale for doing so is:
 * big-endian integers can be lexicographically ordered - that can be
   useful for having versioned CMF keys in an ordered DB, for example
 * less of reason is that network byte order is defined as big-endian

Add a static_cast to the unsigned version of the type being
deserialized when fetching the next byte. That fixes issues for 64 bit
types, because the left shift was promoted to an integer instead of the
correct 64 bit type.

Add support for the `fixedlist` type, allowing lists of arbitrary types,
but of a fixed size. No length is serialized before the actual list.
Maps to `std::array` in C++. In Python, size checks are done for lists,
not allowing lists of an incorrect type.

Remove redundant checks for buffer size in CMFDeserializer.map() and
.list() methods.

Add support for C++ CMF in CMake by slightly modifying the
`FindCMFC.cmake` file from Concord (credit to @robem). Add support for
namespace configuration.

Make the std::string overloads of (de)serialize `static inline`,
allowing the use of multiple CMF files in the same binary.

Add a cppgen test to verify integer serialization produces correct
results, in addition to round-trip serialization/deserialization. That
test ensures no systematic error is present both in serialization and
deserialization code. Make test_cppgen a first-class test and run it as
part of ctest.

Use negative and positive values for signed integers in test_cppgen.

Add sanitize g++ options when compiling cppgen tests in order to guard
against undefined behavior related to bitwise operations, memory
corruption and leaks:
 * -fsanitize=undefined
 * -fsanitize=address
 * -fsanitize=leak
 * -fno-sanitize-recover


dartdart26:
Handling negative signed integers seems a bit complicated, especially in relation to the right shifts we use for both little and big endian: https://stackoverflow.com/questions/4009885/arithmetic-bit-shift-on-a-signed-integer

We should probably also verify cross-language compatibility.

Can we maybe use boost::endian?

dartdart26:
> Handling negative signed integers seems a bit complicated, especially in relation to the right shifts we use for both little and big endian: https://stackoverflow.com/questions/4009885/arithmetic-bit-shift-on-a-signed-integer
> 
> We should probably also verify cross-language compatibility.
> 
> Can we maybe use boost::endian?

Running with `-fsanitize=undefined` doesn't produce any warnings, though.

messages/compiler/cpp/serialize.cpp
```diff
@@ -37,12 +38,16 @@ class BadDataError : public DeserializeError {
...
+ * All integers are encoded in the byte order specified by cmf::kByteOrder
  ******************************************************************************/
 template <typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
 void serialize(std::vector<uint8_t>& output, const T& t) {
   if constexpr (std::is_same_v<T, bool>) {
     output.push_back(t ? 1 : 0);
+  } else if constexpr (kByteOrder == ByteOrder::kBigEndian) {
+    for (auto i = sizeof(T); i > 0; i--) {
+      output.push_back(255 & t >> ((i - 1) * 8));
+    }
```
cloudnoize, line 50: nit: add parentheses to clarify precedence

messages/compiler/cpp/serialize.cpp
```diff
@@ -69,8 +74,14 @@ void deserialize(uint8_t*& start, const uint8_t* end, T& t) {
...
     t = 0;
-    for (auto i = 0u; i < sizeof(T); i++) {
-      t |= ((*(start + i)) << (i * 8));
+    if constexpr (kByteOrder == ByteOrder::kBigEndian) {
+      for (auto i = 0u; i < sizeof(T); i++) {
+        t |= (static_cast<std::make_unsigned_t<T>>((*(start + sizeof(T) - 1 - i))) << (i * 8));
+      }
+    } else {
+      for (auto i = 0u; i < sizeof(T); i++) {
+        t |= (static_cast<std::make_unsigned_t<T>>((*(start + i))) << (i * 8));
```
cloudnoize, line 83: nit : seems that the middle parentheses ```((*(start + i)))``` is redundent.

messages/compiler/cpp/serialize.cpp
```diff
@@ -69,8 +74,14 @@ void deserialize(uint8_t*& start, const uint8_t* end, T& t) {
       throw NoDataLeftError();
     }
     t = 0;
-    for (auto i = 0u; i < sizeof(T); i++) {
-      t |= ((*(start + i)) << (i * 8));
+    if constexpr (kByteOrder == ByteOrder::kBigEndian) {
+      for (auto i = 0u; i < sizeof(T); i++) {
+        t |= (static_cast<std::make_unsigned_t<T>>((*(start + sizeof(T) - 1 - i))) << (i * 8));
+      }
```
cloudnoize, line 80: From what I understand the previous version didn't work, since a byte was bit shifted?

cloudnoize (approved):

messages/compiler/cpp/serialize.cpp
dartdart26: Good point, will add, thanks!

messages/compiler/cpp/serialize.cpp
dartdart26: Yep, wil remove.

messages/compiler/cpp/serialize.cpp
dartdart26: Yes, instead of type T.

andrewjstone (approved):
Looks great! Thanks for the fixes @dartdart26! It's awesome to have another contributor to CMF :)

PR #1065, opened 2020-11-30
ROReplica backup/restore

toly-kournik:
First steps not to save the ROReplica state in rocksDB:
- don't use bft PersistentStorage in ROReplica (was used for saving lastExecutedSeqNum) as not necessary;
- make naming order with sequence numbers and checkpoints in a confusing otherwise onTransferringComplete()

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2495,23 +2495,22 @@ void ReplicaImp::sendCheckpointIfNeeded() {
   sendToAllOtherReplicas(checkpointMessage, true);
 }
 
-void ReplicaImp::onTransferringCompleteImp(SeqNum newStateCheckpoint) {
+void ReplicaImp::onTransferringCompleteImp(uint64_t newStateCheckpoint) {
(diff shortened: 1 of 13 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2495,23 +2495,22 @@ void ReplicaImp::sendCheckpointIfNeeded() {
...
 
   if (ps_) {
     ps_->beginWriteTran();
   }
-
-  if (newStateCheckpoint <= lastExecutedSeqNum) {
+  SeqNum newCheckpointSeqNum = newStateCheckpoint * checkpointWindowSize;
+  if (newCheckpointSeqNum <= lastExecutedSeqNum) {
     LOG_DEBUG(GL,
               "Executing onTransferringCompleteImp(newStateCheckpoint) where newStateCheckpoint <= lastExecutedSeqNum");
```
teoparvanov, line 2509: nit - you could maybe add a KVLOG here for newCheckpointSeqNum and lastExecutedSeqNum

teoparvanov (approved):
Awesone, thanks Toly!

PR #1066, opened 2020-12-01
Add a flag to direct_kv_db_adapter which controls if KVs should be saved with the raw block or not

tdimitrov:
Besides the raw block itself direct_kv_db adapter saves also all KV pairs from the block in the database. This feature doesn't play well with S3 (which shares the adapter code), because there is a limitation for the size of the keys written to the object store (around 250 for minio and 1024 for AWS S3).  For this reason additional flag is added to the adapter constructor which controls whether to save the KV pairs AND the raw block or only the raw block.

### DIFF
--- a/kvbc/src/direct_kv_db_adapter.cpp
+++ b/kvbc/src/direct_kv_db_adapter.cpp
@@ -364,13 +364,17 @@ KeyValuePair DBKeyManipulator::composedToSimple(KeyValuePair _p) {
   return KeyValuePair(key, _p.second);
 }
 
-DBAdapter::DBAdapter(std::shared_ptr<storage::IDBClient> db, std::unique_ptr<IDataKeyGenerator> keyGen, bool use_mdt)
+DBAdapter::DBAdapter(std::shared_ptr<storage::IDBClient> db,
+                     std::unique_ptr<IDataKeyGenerator> keyGen,
+                     bool use_mdt,
+                     bool save_kv_pairs_separately)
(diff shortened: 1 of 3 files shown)


teoparvanov (approved):
LGTM!

kvbc/src/direct_kv_db_adapter.cpp
```diff
@@ -364,13 +364,17 @@ KeyValuePair DBKeyManipulator::composedToSimple(KeyValuePair _p) {
...
+                     bool use_mdt,
+                     bool save_kv_pairs_separately)
     : logger_{logging::getLogger("concord.kvbc.v1DirectKeyValue.DBAdapter")},
       db_(db),
       keyGen_{std::move(keyGen)},
       mdt_{use_mdt},
       lastBlockId_{fetchLatestBlockId()},
-      lastReachableBlockId_{fetchLastReachableBlockId()} {}
+      lastReachableBlockId_{fetchLastReachableBlockId()},
+      saveKvPairsSeparately_{save_kv_pairs_separately} {}
```
dartdart26, line 377: Nit: I guess that's the existing naming in the file, but it seems inconsistent.

dartdart26 (approved):
Looks good, thanks!

kvbc/src/direct_kv_db_adapter.cpp
tdimitrov: Yes, for some reason function parameter names are with _, while members with camel case and trailing _. I'll leave it like this.

PR #1067, opened 2020-12-01
fix metrics bug

yontyon:


f-squirrel (approved):
Thanks for the quick fix, @yontyon !

PR #1069, opened 2020-12-01
Reduce merkle tree serialization allocation count

dartdart26:
Use an output parameter instead of std::string concatenation when
serializing merkle tree internals. That has the effect of reducing the
count of memory allocations. Special attention is given to serialization
of BatchedInternalNodes as it was creating a big number of temporary
strings that were large enough to trigger allocations (without
taking advantage of small string optimizations).

Add a benchmark for BatchedInternalNodes. Change constants for block
count and key/value count and sizes in order to better reflect real uses
cases.

```
Before:
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
serializeBatchedInternalNode         3229 ns         3229 ns       212506
```

```
After:
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
serializeBatchedInternalNode          680 ns          680 ns       818500
```

Thanks to @andrewjstone for observing and measuring.

### DIFF
--- a/kvbc/include/merkle_tree_serialization.h
+++ b/kvbc/include/merkle_tree_serialization.h
@@ -24,11 +24,11 @@
 #include "storage/db_types.h"
 #include "string.hpp"
 
+#include <algorithm>
 #include <cstdint>
 #include <iterator>
 #include <limits>
 #include <string>
-
 #include <utility>
 #include <variant>
 #include <vector>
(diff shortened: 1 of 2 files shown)


kvbc/include/merkle_tree_serialization.h
```diff
@@ -77,59 +82,112 @@ inline std::string serializeImp(storage::v2MerkleTree::detail::EBFTSubtype type)
...
 
-    std::string &buf_;
+    std::string &out_;
   };
 
-  const auto &children = intNode.children();
-  std::string serializedChildren;
   auto mask = BatchedInternalMaskType{0};
+
+  // We assume the largest batched internal node contains with 15 internal and 16 leaf children. We then factor in the
```
andrewjstone, line 171: Nit: Remove `with`.

andrewjstone (approved):
Looks great. Thanks @dartdart26 for the awesome speed improvement.

kvbc/include/merkle_tree_serialization.h
dartdart26: Fixed.

PR #1070, opened 2020-12-02
ROReplica backup/restore

toly-kournik:
Store State Transfer metadata in S3 storage instead of rocksdb making ROReplica not to use local rocksdb at all.


storage/src/s3/key_manipulator.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "s3/key_manipulator.h"
+#include "string.hpp"
+#include <string>
+
+namespace concord::storage::s3 {
+
+using ::concordUtils::Sliver;
+
+STKeyManipulator::STKeyManipulator(const std::string& prefix) : prefix_(prefix + "/metadata/st/") {}
```
tdimitrov, line 22: Will this work correctly with empty prefix?

If the prefix is empty, it is replaced with a slash (S3 doesn't handle // in urls): https://github.com/vmware/concord-bft/blob/a9e74742d3a77e9b2852d681956f2f965a4697a6/kvbc/include/direct_kv_db_adapter.h#L79

toly-kournik: fixed, although we always provide a meaningful prefix

kvbc/src/ReplicaImp.cpp
```diff
@@ -229,7 +229,8 @@ ReplicaImp::ReplicaImp(ICommunication *comm,
     replicaConfig_.get<uint32_t>("concord.bft.st.sourceReplicaReplacementTimeoutMs", 15000),
     replicaConfig_.get<uint32_t>("concord.bft.st.fetchRetransmissionTimeoutMs", 250),
     replicaConfig_.get<uint32_t>("concord.bft.st.metricsDumpIntervalSec", 5),
-    replicaConfig_.get("concord.bft.st.runInSeparateThread", replicaConfig_.isReadOnly)
+    replicaConfig_.get("concord.bft.st.runInSeparateThread", replicaConfig_.isReadOnly),
+    replicaConfig_.get("concord.bft.st.enableReservedPages", !replicaConfig_.isReadOnly)
```
teoparvanov, line 233: nice :)

kvbc/src/direct_kv_storage_factory.cpp
```diff
@@ -93,7 +93,7 @@ std::unique_ptr<storage::IMetadataKeyManipulator> S3StorageFactory::newMetadataK
 }
 
 std::unique_ptr<storage::ISTKeyManipulator> S3StorageFactory::newSTKeyManipulator() const {
-  return std::make_unique<storage::v1DirectKeyValue::STKeyManipulator>();
+  return std::make_unique<storage::s3::STKeyManipulator>(s3Conf_.pathPrefix);
```
teoparvanov, line 96: If I understand correctly, this is how we forward the ST metadata to S3.

teoparvanov (approved):
Thanks Toly, I think this is a very elegant implementation!

kvbc/src/ReplicaImp.cpp
toly-kournik: for now

kvbc/src/direct_kv_storage_factory.cpp
toly-kournik: this is how ST metadata keys are generated for S3

PR #1071, opened 2020-12-03
Histograms: Add support for async time recording

andrewjstone:
A new type, `AsyncTimeRecorderMap`, was added that enables recording
start and end times for non-linear control flows and automatically
adding them to the histogram. This class is really only useful for
tracking multiple, concurrent, outstanding requests with distinct keys.

For single asynchronous operations, another class, `AsyncTimeRecorder`,
was added.

An example that records the duration of consensus for individual slots
on the primary was added. The output from `concord-ctl` looks like
the following on the primary:

```
root@1e2dda044cbb:/concord# ./concord-ctl perf get replica consensus
Start time (UTC): 2020-12-03 21:46:31
Snapshot Start time (UTC): 2020-12-03 21:47:08
Snapshot End time (UTC): 2020-12-03 21:47:19
Unit: microseconds

History
--------------------
Count: 142
Max: 179967
Min: 3400
Percentiles:
  10: 4065
  25: 4283
  50: 4983
  75: 6019
  90: 7843
  95: 10679
  99: 99711
  99.9: 179967
  99.99: 179967
  99.999: 179967

Snapshot
--------------------
Count: 44
Max: 7967
Min: 3254
Percentiles:
  10: 3459
  25: 4175
  50: 4519
  75: 5383
  90: 6183
  95: 6723
  99: 7967
  99.9: 7967
  99.99: 7967
  99.999: 7967
```

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -495,6 +495,7 @@ void ReplicaImp::startConsensusProcess(PrePrepareMsg *pp, bool isInternalNoop) {
     LOG_INFO(CNSUS,
              "Sending PrePrepare with the following payload of the following correlation ids ["
                  << pp->getBatchCorrelationIdAsString() << "]");
+    consensus_times_.start(primaryLastUsedSeqNum);
   }
 
   SeqNumInfo &seqNumInfo = mainLog->get(primaryLastUsedSeqNum);
@@ -1047,7 +1048,7 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
           (msgSeqNum > lastExecutedSeqNum + config_.getconcurrencyLevel());  // TODO(GG): check/improve this logic
 
       auto execution_span = concordUtils::startChildSpan("bft_execute_committed_reqs", span);
-      executeNextCommittedRequests(execution_span, askForMissingInfoAboutCommittedItems);
+      executeNextCommittedRequests(execution_span, msgSeqNum, askForMissingInfoAboutCommittedItems);
       return;
     } else if (pps.hasFullProof()) {
       const auto fullProofCollectorId = pps.getFullProof()->senderId();
@@ -1537,7 +1538,7 @@ void ReplicaImp::onCommitCombinedSigSucceeded(SeqNum seqNumber,
 
   auto span = concordUtils::startChildSpanFromContext(
       commitFull->spanContext<std::remove_pointer<decltype(commitFull)>::type>(), "bft_execute_committed_reqs");
-  executeNextCommittedRequests(span, askForMissingInfoAboutCommittedItems);
+  executeNextCommittedRequests(span, seqNumber, askForMissingInfoAboutCommittedItems);
 }
 
 void ReplicaImp::onCommitVerifyCombinedSigResult(SeqNum seqNumber, ViewNum view, bool isValid) {
@@ -1581,7 +1582,7 @@ void ReplicaImp::onCommitVerifyCombinedSigResult(SeqNum seqNumber, ViewNum view,
   auto span = concordUtils::startChildSpanFromContext(
       commitFull->spanContext<std::remove_pointer<decltype(commitFull)>::type>(), "bft_execute_committed_reqs");
   bool askForMissingInfoAboutCommittedItems = (seqNumber > lastExecutedSeqNum + config_.getconcurrencyLevel());
-  executeNextCommittedRequests(span, askForMissingInfoAboutCommittedItems);
+  executeNextCommittedRequests(span, seqNumber, askForMissingInfoAboutCommittedItems);
 }
 
 template <>
@@ -2299,6 +2300,8 @@ bool ReplicaImp::tryToEnterView() {
 void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNewView) {
   SeqNum firstPPSeq = 0;
   SeqNum lastPPSeq = 0;
+
+  consensus_times_.clear();
 
   if (!prePreparesForNewView.empty()) {
     firstPPSeq = prePreparesForNewView.front()->seqNumber();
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1046,6 +1047,8 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
       const bool askForMissingInfoAboutCommittedItems =
           (msgSeqNum > lastExecutedSeqNum + config_.getconcurrencyLevel());  // TODO(GG): check/improve this logic
 
+      consensus_times_.end(msgSeqNum);
```
yuliasherman, line 1050: Isn't it better to put this call at the beginning of executeRequestsAndSendResponses() function to have just one closer point?

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -474,9 +478,18 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
         std::make_shared<Recorder>(1, MAX_VALUE_NANOSECONDS, 3, concord::diagnostics::Unit::NANOSECONDS);
     std::shared_ptr<Recorder> onTransferringCompleteImp =
         std::make_shared<Recorder>(1, MAX_VALUE_NANOSECONDS, 3, concord::diagnostics::Unit::NANOSECONDS);
+
+    // Only updated by the primary
+    std::shared_ptr<Recorder> consensus =
+        std::make_shared<Recorder>(1, MAX_VALUE_MILLISECONDS, 3, concord::diagnostics::Unit::MILLISECONDS);
```
yuliasherman, line 484: I would move to MICROSECONDS here

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2300,6 +2304,8 @@ void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNew
   SeqNum firstPPSeq = 0;
   SeqNum lastPPSeq = 0;
 
+  consensus_times_.clear();
```
yuliasherman, line 2307: I did not get at which point this function should be called.

diagnostics/include/performance_handler.h
```diff
@@ -146,6 +146,25 @@ class TimeRecorder {
   Recorder& recorder_;
 };
 
+// This is a wrapper around an unordered_map that records the durations of asynchronous actions.
+// It's useful when the timing being recorded can't tracked in a single scope, and there are
+// multiple outstanding requests that need timing, such as consensus slots.
+//
+// This type is not thread safe and always uses the non-atomic version of record.
```
yuliasherman, line 153: Why not make it thread-safe? In a single-threaded context, it does not have any overhead, but in other cases, like pre-processing, it will be used from multiple contexts.

yuliasherman:
Thanks for implementing this, Andrew. Highly appreciated!

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: It's a good point. Then we only need to call it once also.

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: Since this stat is only useful on the primary, and we don't want to leak the memory for slots that don't commit on the primary, we clear it after a view change.

bftengine/src/bftengine/ReplicaImp.hpp
andrewjstone: I can do that, but I'm not sure how much it helps us in practice.  Practically speaking, I didn't think we could get microsecond resolution. Even if we see `0`, we just know that consensus latency is not a problem. In production we have multiple messages traversing a real network, and with batching they are going to be fairly large. We are going to just end up discarding the last 3 digits. I noticed this problem where I used nanoseconds instead of micros elsewhere, and for most of those I want to change to micros, as anything less is unlikely or just a blip.  If we consistently see `0` we can change to microseconds.

diagnostics/include/performance_handler.h
andrewjstone: Good point. I will make it configurable via template parameter in a manner similar to the recorder.

andrewjstone (commented):
Thanks for the review Yulia. I will update accordingly.

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: We don't actually have the sequence number in that call so we can't do this without modifying the API, which I don't want to do in this commit.

My original thinking was also that we only want to measure consensus. Consensus is complete before that call. It's probably not a big difference, but we don't even guarantee that anything gets executed at that point as there can be gaps in the log or a missing preprepare. To measure the time between consensus and execution we will need at least one more histogram. That work can be done in future commits.

bftengine/src/bftengine/ReplicaImp.hpp
andrewjstone: I changed it to MICROSECONDS.

diagnostics/include/performance_handler.h
andrewjstone: Fixed.

bftengine/src/bftengine/ReplicaImp.cpp
andrewjstone: I ended up changing the API to take the seqnumber and doing it there as you suggested. It is indeed cleaner.

robem (approved):
Looks useful to me :)

diagnostics/include/performance_handler.h
```diff
@@ -144,6 +149,51 @@ class TimeRecorder {
...
+// This allows starting and stopping a timer manually rather than using the destructor. It's useful
+// for async operations without a linear control-flow.
+template <bool IsAtomic = false>
+class AsyncTimeRecorder {
+ public:
+  AsyncTimeRecorder(const std::shared_ptr<Recorder>& recorder) : recorder_(recorder) {}
+  void start() {
+    // If a timer was already started, it will record if start is called again before end.
+    // This behavior can be prevented by explicitly calling clear().
+    timer_.reset(new TimeRecorder<IsAtomic>(*recorder_));
```
andrewjstone, line 186: If we allow manual start and end of TimeRecorder, we can prevent allocation here.

PR #1072, opened 2020-12-03
clear control state manager data on startup

yontyon:
After successful wedge, the instruction to a wedge is still in the reserved pages. Thus, when the replica starts again it will try to wedge again.
For that, we clear all the controlStateManager data on startup.

Notice, that if a replica has crashed after wedge command and before the actual shutdown, this may cause to instability.

### DIFF
--- a/bftengine/include/bftengine/ControlStateManager.hpp
+++ b/bftengine/include/bftengine/ControlStateManager.hpp
@@ -53,7 +53,7 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
   void setEraseMetadataFlag(int64_t currentSeqNum);
   std::optional<int64_t> getEraseMetadataFlag();
 
-  ControlStateManager(IStateTransfer* state_transfer, uint32_t sizeOfReservedPages);
+  ControlStateManager(IStateTransfer* state_transfer, uint32_t sizeOfReservedPages, bool disabled = false);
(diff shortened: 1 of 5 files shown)


Liorgal28 (approved):
LGTM

bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -53,14 +53,16 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
   void setEraseMetadataFlag(int64_t currentSeqNum);
   std::optional<int64_t> getEraseMetadataFlag();
 
-  ControlStateManager(IStateTransfer* state_transfer, uint32_t sizeOfReservedPages);
```
robem, line 56: Looks like as if the `disable` flag is used for testing only; similar the `disable()` function. If that is true then instead of baking the flag into the production code, couldn't we change the test code?
In `simple_test_replica.hpp` we create a `ControlStateManager` only to disable it right away. How about we create a `TestControlStageManager` which implements the same interface? Thereby, we wouldn't need to touch the production ControlStateManager.

robem:
> Notice, that if a replica has crashed after wedge command and before the actual shutdown, this may cause to instability.

Could you provide more details?

yontyon:
Removing data from reserved pages in a single replica may cause ST problems. Hence, this PR is not correct and I'm closing it.

PR #1073, opened 2020-12-03
Conversion from RocksDB IDBClient to NativeClient

dartdart26:
Provide conversions from an IDBClient that points to rocksdb::Client to
NativeClient and vice versa.

Add support for user-supplied RocksDB options in NativeClient. Moreover,
support creating a NativeClient with default options and with the
existing (persisted) options. In effect, NativeClient::newClient() can
be called with:
 * UserOptions
 * DefaultOptions
 * ExistingOptions

Move column family handles from NativeClient to rocksdb::Client.
Rationale is that since we now support conversions between them, we want
rocksdb::Client to be able to open a DB with column families inside of
it. Without that change, it wouldn't be possible, because RocksDB
requires that all column families are specified when opening the DB.

Extract optimizations in rocksdb::Client into its own method and
support an ability to turn them on/off. Additionally, apply these
optimizations **after** having loaded the existing configuration from a
file.

Add a guard against multiple initializations in in ::rocksdb::Client.

Add tests to verify that conversions and options persistence work as
expected.

### DIFF
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -29,7 +29,7 @@ struct Block {
 };
 class KeyValueBlockchain {
  public:
-  KeyValueBlockchain() : native_client_(NativeClient::newClient("/tmp", false)) {}
+  KeyValueBlockchain() : native_client_(NativeClient::newClient("/tmp", false, NativeClient::DefaultOptions{})) {}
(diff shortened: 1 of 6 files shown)


andrewjstone (approved):
@dartdart26 Thanks so much for doing this. It definitely turned out to be more complicated than either of us expected. I'm glad you took care of it instead of me :)

PR #1074, opened 2020-12-03
Add flag to disable has keys request if feature is disabled

cloudnoize:
Has-key request is made in order to synch the key exchange of the BFT library with its client. 
When the feature is `off` this request should return ```true``` by defult.

### DIFF
--- a/bftengine/src/bftengine/KeyManager.cpp
+++ b/bftengine/src/bftengine/KeyManager.cpp
@@ -26,6 +26,7 @@ KeyManager::KeyManager(InitData* id)
       keyStore_{id->clusterSize, *id->reservedPages, id->sizeOfReservedPage},
       multiSigKeyHdlr_(id->kg),
       keysView_(id->sec, id->backupSec, id->clusterSize),
+      keyExchangeOnStart_(id->keyExchangeOnStart),
       timers_(*(id->timers)) {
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -76,7 +77,7 @@ std::string KeyManager::generateCid() {
 std::string KeyManager::onKeyExchange(KeyExchangeMsg& kemsg, const uint64_t& sn) {
   if (kemsg.op == KeyExchangeMsg::HAS_KEYS) {
     LOG_INFO(KEY_EX_LOG, "Has key query arrived, returning " << std::boolalpha << keysExchanged << std::noboolalpha);
-    if (!keysExchanged) return std::string(KeyExchangeMsg::hasKeysFalseReply);
+    if (!keysExchanged && enableHasKeysReq_) return std::string(KeyExchangeMsg::hasKeysFalseReply);
```
robem, line 80: I'm very confused by the variable name. How about we keep the same name that we use in the configuration?
```
if (!keysExchange && keyExchangeOnStart)
```

cloudnoize: sure

robem (approved):
Lgtm. Thanks!

PR #1076, opened 2020-12-04
Incorporate proof in vc msgs

HristoStaykov:
The purpose of this change is to incorporate the quorum of Complaints in the View Change message. 
This way once a Replica receives such View Change message, even if it has not yet received all the necessary
complaints from peers it will be able to verify that it also has to leave the current View.
1. Added list of Complaints to ViewChangeMsg.
2. Added check method for Complaints list.
3. Added ComplaintsIterator to iterate over packed list of Complaints.
4. Added list of Complaints to unit test for ViewChangeMsg.

### DIFF
--- a/bftengine/src/bftengine/messages/ViewChangeMsg.cpp
+++ b/bftengine/src/bftengine/messages/ViewChangeMsg.cpp
@@ -231,7 +231,7 @@ bool ViewChangeMsg::checkComplaints(uint16_t sigSize) const {
     remainingBytes -= sizeof(MsgSize);
     currLoc += sizeof(MsgSize);
 
-    if (*complaintSize <= sizeof(MessageBase::Header)) {
+    if (*complaintSize <= sizeOfHeader<ReplicaAsksToLeaveViewMsg>()) {
       return false;
     }
 
--- a/bftengine/src/bftengine/messages/ViewChangeMsg.cpp
+++ b/bftengine/src/bftengine/messages/ViewChangeMsg.cpp
@@ -365,13 +365,9 @@ bool ViewChangeMsg::ComplaintsIterator::getCurrent(char*& pComplaint, MsgSize& s
   if (end()) return false;
 
   size = *(MsgSize*)(msg->body() + currLoc);
-
   const uint32_t remainingbytes = (endLoc - currLoc) - sizeof(MsgSize);
-
   ConcordAssert(remainingbytes >= size);  // Validate method must make sure we never accept such message
-
   pComplaint = (char*)malloc(size);
-
   memcpy(pComplaint, msg->body() + currLoc + sizeof(MsgSize), size);
 
   return true;
@@ -381,11 +377,8 @@ void ViewChangeMsg::ComplaintsIterator::gotoNext() {
   if (end()) return;
 
   const uint32_t size = *(MsgSize*)(msg->body() + currLoc);
-
   const uint32_t remainingbytes = (endLoc - currLoc) - sizeof(MsgSize);
-
   ConcordAssert(remainingbytes >= size);  // Validate method must make sure we never accept such message
-
   currLoc += sizeof(MsgSize) + size;
 
   nextComplaintNum++;
--- a/bftengine/src/bftengine/messages/ViewChangeMsg.cpp
+++ b/bftengine/src/bftengine/messages/ViewChangeMsg.cpp
@@ -45,10 +45,15 @@ void ViewChangeMsg::setNewViewNumber(ViewNum newView) {
 }
 
 void ViewChangeMsg::getMsgDigest(Digest& outDigest) const {
-  uint32_t bodySize = b()->locationAfterLast;
-  if (bodySize == 0) bodySize = sizeof(Header) + spanContextSize();
+  auto bodySize = getBodySize();
   bodySize += b()->sizeOfAllComplaints;
   DigestUtil::compute(body(), bodySize, (char*)outDigest.content(), sizeof(Digest));
+}
+
+uint32_t ViewChangeMsg::getBodySize() const {
+  uint32_t bodySize = b()->locationAfterLast;
+  if (bodySize == 0) bodySize = sizeof(Header) + spanContextSize();
+  return bodySize;
 }
 
 void ViewChangeMsg::addElement(SeqNum seqNum,
@@ -106,9 +111,8 @@ void ViewChangeMsg::addComplaint(const ReplicaAsksToLeaveViewMsg* const complain
   // +--------------------+-------------------------+--------------------+-------------------------+---
 
   ConcordAssert(b()->numberOfComplaints > 0 || b()->sizeOfAllComplaints == 0);
-  size_t bodySize = b()->locationAfterLast;
-  if (bodySize == 0) bodySize = sizeof(Header) + spanContextSize();
-  uint16_t sigSize = ViewsManager::sigManager_->getMySigLength();
+  auto bodySize = getBodySize();
+  auto sigSize = ViewsManager::sigManager_->getMySigLength();
   bodySize += sigSize + b()->sizeOfAllComplaints;
 
   auto sizeOfComplaint = complaint->size();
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -93,18 +98,49 @@ void ViewChangeMsg::addElement(SeqNum seqNum,
   b()->numberOfElements++;
 }
 
+void ViewChangeMsg::addComplaint(const ReplicaAsksToLeaveViewMsg* const complaint) {
+  // We store complaints in a size/value list
+  // Visual representation for List Of Complaints:
+  // +--------------------+-------------------------+--------------------+-------------------------+---
+  // |Size of next element|ReplicaAsksToLeaveViewMsg|Size of next element|ReplicaAsksToLeaveViewMsg|...
+  // +--------------------+-------------------------+--------------------+-------------------------+---
```
teoparvanov, line 106: nice, very useful for the reviewer 😄

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -93,18 +98,49 @@ void ViewChangeMsg::addElement(SeqNum seqNum,
...
+
+  ConcordAssert(b()->numberOfComplaints > 0 || b()->sizeOfAllComplaints == 0);
+  size_t bodySize = b()->locationAfterLast;
```
teoparvanov, line 109: `auto bodySize` ?

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -93,18 +98,49 @@ void ViewChangeMsg::addElement(SeqNum seqNum,
...
+  if (bodySize == 0) bodySize = sizeof(Header) + spanContextSize();
+  uint16_t sigSize = ViewsManager::sigManager_->getMySigLength();
```
teoparvanov, line 111: `auto sigSize` ?

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -178,6 +216,32 @@ bool ViewChangeMsg::checkElements(uint16_t sigSize) const {
   return true;
 }
 
+bool ViewChangeMsg::checkComplaints(uint16_t sigSize) const {
```
teoparvanov, line 219: This method looks more like `checkComplaintsListIntegrity()`. I don't see any signature verifications here?

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -178,6 +216,32 @@ bool ViewChangeMsg::checkElements(uint16_t sigSize) const {
...
+  uint16_t numOfActualComplaints = 0;
+  uint32_t remainingBytes = size() - sigSize - b()->locationAfterLast;
+  char* currLoc = body() + b()->locationAfterLast + sigSize;
```
teoparvanov, line 222: I must admit, I got lost in the pointer arithmetic here...

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -178,6 +216,32 @@ bool ViewChangeMsg::checkElements(uint16_t sigSize) const {
...
+    numOfActualComplaints++;
+    remainingBytes -= *complaintSize;
+    currLoc += *complaintSize;
+  }
+
+  if (numOfActualComplaints != numberOfComplaints()) return false;
+
+  if (remainingBytes != 0) return false;
+
+  return true;
```
teoparvanov, line 242: shouldn't we also check that we have at least `f+1` signatures?

teoparvanov (commented):
Some thoughts from my first review pass...

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
HristoStaykov: Yes, this will be better declared auto, I agree.

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
HristoStaykov: Same as above. Agree.

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
HristoStaykov: Yes, here we only check the integrity of the list of complaints. The actual check will be carried at a higher level.

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
HristoStaykov: Yes, I have extracted the calculation of the body size in a separate function for better readability and to avoid mistakes. Also updated the visual representation :)

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -32,6 +32,8 @@ ViewChangeMsg::ViewChangeMsg(ReplicaId srcReplicaId,
   b()->genReplicaId = srcReplicaId;
   b()->newView = newView;
   b()->lastStable = lastStableSeq;
+  b()->numberOfComplaints = 0;
```
f-squirrel, line 35: Nit: Since the values are predefined, we may init them in the struct declaration.
Please don't change unless you find it useful.

f-squirrel (approved):
LGTM!

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
HristoStaykov: I think it is a good point to highlight. 
I see in the code that we do not need to set anything to 0, because we memset the chunk of memory to 0 [MessageBase.cpp:86](https://github.com/vmware/concord-bft/blob/1b877d35de2af14335a1099878b8f155d311b645/bftengine/src/bftengine/messages/MessageBase.cpp#L86.). Still, I think stating it explicitly is good for clarity.
On the other hand since we malloc the chunk in MessageBase, we don't call any constructors on it, so unfortunately I think setting it to 0 in the struct declaration won't have effect in our use case.

f-squirrel: Good point!

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -269,5 +338,64 @@ bool ViewChangeMsg::ElementsIterator::goToAtLeast(SeqNum lowerBound) {
...
+
+  const uint32_t remainingbytes = (endLoc - currLoc) - sizeof(MsgSize);
+
+  ConcordAssert(remainingbytes >= size);  // Validate method must make sure we never accept such message
+
+  pComplaint = (char*)malloc(size);
+
+  memcpy(pComplaint, msg->body() + currLoc + sizeof(MsgSize), size);
+
+  return true;
```
teoparvanov, line 377: nit - why do we need an empty space after each line of code here?

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -269,5 +338,64 @@ bool ViewChangeMsg::ElementsIterator::goToAtLeast(SeqNum lowerBound) {
...
+
+  const uint32_t size = *(MsgSize*)(msg->body() + currLoc);
+
+  const uint32_t remainingbytes = (endLoc - currLoc) - sizeof(MsgSize);
+
+  ConcordAssert(remainingbytes >= size);  // Validate method must make sure we never accept such message
+
+  currLoc += sizeof(MsgSize) + size;
+
+  nextComplaintNum++;
```
teoparvanov, line 391: same here - no need of an empty space after each line

teoparvanov (approved):
LGTM, thanks @HristoStaykov !

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
HristoStaykov: I think might be better to leave to higher level to verify the signatures and the quorum size.

PR #1078, opened 2020-12-06
Clear wedge point before shutdown

yontyon:
In wedge, we save the wedge point on the reserved pages in order to transfer it using ST. 
That means, that once the replicas recover, they try to read the reserved page again and thus try to wedge again. Hence we need a mechanism for safely clear the wedge point on a replica recovery.

As the data is saved in the RP, we have to clear the data in a point where we know that no replica is running a ST, and that no ST can be initiated in the meanwhile.
Thus, we clear the wedge point once we reached to an n/n checkpoint in this wedge point.
If we reached to an n/n checkpoint it means that all replicas have done with ST (if there was any) and all replicas are going to stop processing new requests (which means that there won't be another ST until we restart the replicas).

*******
In addition in this PR we added two metrics that help us to analyze performance.


Liorgal28 (approved):
Looks good

PR #1079, opened 2020-12-06
Add more context to ST logs

oded282:
In this PR I added more context to ST logs for causal ordering.
Every source replica will log the destination replica number.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -935,7 +935,7 @@ void BCStateTran::sendAskForCheckpointSummariesMsg() {
   lastTimeSentAskForCheckpointSummariesMsg = getMonotonicTimeMilli();
   lastMsgSeqNum_ = uniqueMsgSeqNum();
   metrics_.last_msg_seq_num_.Get().Set(lastMsgSeqNum_);
-  SCOPED_MDC_SEQ_NUM(std::to_string(lastMsgSeqNum_));
+  SCOPED_MDC_SEQ_NUM(getSequenceNumber(config_.myReplicaId, lastMsgSeqNum_));
 
   msg.msgSeqNum = lastMsgSeqNum_;
   msg.minRelevantCheckpointNum = psd_->getLastStoredCheckpoint() + 1;
@@ -1007,7 +1007,7 @@ void BCStateTran::sendFetchResPagesMsg(int16_t lastKnownChunkInLastRequiredBlock
 //////////////////////////////////////////////////////////////////////////////
 
 bool BCStateTran::onMessage(const AskForCheckpointSummariesMsg *m, uint32_t msgLen, uint16_t replicaId) {
-  SCOPED_MDC_SEQ_NUM(std::to_string(m->msgSeqNum));
+  SCOPED_MDC_SEQ_NUM(getSequenceNumber(replicaId, m->msgSeqNum));
   LOG_DEBUG(getLogger(), "");
 
   ConcordAssert(!psd_->getIsFetchingState());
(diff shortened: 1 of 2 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -932,7 +932,7 @@ void BCStateTran::sendAskForCheckpointSummariesMsg() {
   lastTimeSentAskForCheckpointSummariesMsg = getMonotonicTimeMilli();
   lastMsgSeqNum_ = uniqueMsgSeqNum();
   metrics_.last_msg_seq_num_.Get().Set(lastMsgSeqNum_);
-  SCOPED_MDC_SEQ_NUM(std::to_string(lastMsgSeqNum_));
+  SCOPED_MDC_SEQ_NUM(std::to_string(config_.myReplicaId) + "-" + std::to_string(lastMsgSeqNum_) + "-0" + "-0");
```
robem, line 935: If you introduce a new format then this should be part of the macro itself which then takes those variables as arguments.
IIRC, replica ID should be part of the log message already and hence this would be duplication.

robem (changes requested):

bftengine/src/bcstatetransfer/BCStateTran.cpp
toly-kournik: State Transfer is a separate protocol, we are piggybacking on the bft sequence number positional variable in log format and put there ST "sequence number" which may differ for different parts of protocol - somewhere it receives 2 arguments, somewhere 4. In one plase replica id is my replica id, in the other - id of a peer replica.
So it makes no sense to introduce several new MDC macros for all these cases.

toly-kournik (approved):
@oded282 LGTM!!! Eventually we've got a brief and elegant solution!

bftengine/src/bcstatetransfer/BCStateTran.cpp
robem: The format is simple enough and I wouldn't even notice if it was just one or two occurrences. However, the format (as simple as it is) is duplicated in too many places. If anyone would want to write a parser then what is the source of truth for the parser? Line 935, or 1268, ... ? My point is that this is not a maintainable approach. Humans make mistakes and if we adopt a new line then we might get the format wrong by accident. Or what if we have to change the format for yet unknown reasons?
I know that those are edge case failure scenario. However, what is the harm in building a scalable/maintainable solution from the start?

toly-kournik: @robem , what's your exact suggestion?

robem: There are multiple solutions. Above, I mentioned to use a macro, however, you can also replace the string concatenation with a function that generates the string for you. I'd prefer an inline function over a macro but I'm fine with either.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2502,6 +2499,14 @@ void BCStateTran::SetAggregator(std::shared_ptr<concordMetrics::Aggregator> aggr
   metrics_component_.SetAggregator(aggregator);
 }
 
+inline std::string BCStateTran::getSequenceNumber(uint16_t replicaId,
+                                                  uint64_t seqNum,
+                                                  uint16_t blockNum,
+                                                  uint64_t chunkNum) {
```
teoparvanov, line 2505: What happens if those two aren't provided? Shouldn't they have default values?

oded282: They have default values:
inline std::string getSequenceNumber(uint16_t replicaId, uint64_t seqNum, uint16_t = 0, uint64_t = 0);

robem (approved):
Thanks for the refactoring. Lgtm!

bftengine/src/bcstatetransfer/BCStateTran.cpp
teoparvanov: thanks for the clarification!

teoparvanov (approved):
LGTM, great job @oded282 !

PR #1080, opened 2020-12-07
Add object store (S3) specific metrics

tdimitrov:
Implementation of new metrics component, containing S3 specific metrics.
It is named "s3" and contains the following values:
numKeysTransferred - counter - how many keys are transferred to S3
bytesTransferred - counter - the size of the data (in bytes) transferred
to S3
lastSavedBlockId - status - the id of the last block transferred to S3


storage/include/s3/metrics.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  concordMetrics::Component metrics_component_;
+
+  concordMetrics::CounterHandle numKeysTransferred;
+  concordMetrics::CounterHandle bytesTransferred;
+
+ private:
+  // This function "guesses" if metadata or block is being updated.
+  // In the latter case it updates the metric
+  bool IsBlockKey(std::string_view key) { return key.find("metadata") == std::string_view::npos; }
```
f-squirrel, line 50: `isBlockKey`?

storage/include/s3/metrics.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  concordMetrics::StatusHandle lastSavedBlockId;
```
f-squirrel, line 52: `lastSavedBlockId` => `lastSavedBlockId_` ?

storage/include/s3/metrics.hpp
```diff
@@ -0,0 +1,54 @@
...
+        lastSavedBlockId{
+            metrics_component_.RegisterStatus("last_saved_block_id", "0"),
+        }
+
+  {
+    metrics_component_.Register();
+  }
+
+  void updateLastSavedBlockId(const std::string& key) {
+    if (!IsBlockKey(key)) return;
```
f-squirrel, line 23: AFAIK, the `key` must be the block id.
If not, can we at least log it as a warning?

bftengine/tests/s3/metrics_test.cpp
```diff
@@ -0,0 +1,34 @@
+#include "gtest/gtest.h"
+#include "s3/metrics.hpp"
+#include "direct_kv_storage_factory.h"
+#include "direct_kv_db_adapter.h"
+
+using namespace concord::kvbc;
+// using namespace concord::storage;
```
f-squirrel, line 7: Remove the commented?

storage/include/s3/metrics.hpp
```diff
@@ -0,0 +1,54 @@
...
+        },
+        lastSavedBlockId{
+            metrics_component_.RegisterStatus("last_saved_block_id", "0"),
+        }
+
+  {
+    metrics_component_.Register();
+  }
+
+  void updateLastSavedBlockId(const std::string& key) {
```
f-squirrel, line 22: The keys are naturally stored as `Keys` which is an alias to `Sliver`.
Can we use it instead of `std::string` and using Key::toString()?
AFAIK, `Sliver` also can also return a `string_view`.

storage/include/s3/metrics.hpp
tdimitrov: It can be the block id or metadata. At this stage we don't know it and we have to make a guess.

storage/include/s3/metrics.hpp
tdimitrov: AFAIK method names should be capitalized (https://google.github.io/styleguide/cppguide.html#Function_Names)?

tdimitrov: Nope, my bad. This is exception :)

storage/include/s3/metrics.hpp
tdimitrov: The reason I put std::string as input parameter is that we strictly threat the key as a string in this case and if for some reason it stops being a string in the future - the method will make no sense.

But it will be more consistent with the rest of the code if I get a Sliver. I'll change it.

f-squirrel (approved):
LGTM!
Thank you @tdimitrov for addressing the comments and the explanation.

PR #1081, opened 2020-12-07
Adding categorized block and raw block structs

cloudnoize:
Block contains data per category, about keys that were added at this block.
In addition, it contains the parent digest and the block ID.

Each category is responsible to handle the block struct its ```update_info``` as the result of an update operation.

Rawblock is a unit, which we can transfer and reconstruct a block in a remote replica.
It contains all the relevant updates as received from the client in the initial call to storage.

RawBlock can be constructed from a ```Block``` struct.

In addition, this PR adds support to ```unsigned char``` in RocksDB NativeClient.
Refactoring to the updates and kv_blockchain API.
Unit tests.

### DIFF
--- a/kvbc/cmf/categorized_kvbc_msgs.cmf
+++ b/kvbc/cmf/categorized_kvbc_msgs.cmf
@@ -5,24 +5,24 @@ Msg MerkleUpdatesData 1 {
     list string deletes
 }
 
-Msg ValueWithExpirationData 2{
+Msg ValueWithExpirationData 2 {
     string value
     optional int64 expire_at
     bool stale_on_update
 }
 
-Msg KeyValueUpdatesData 3{
+Msg KeyValueUpdatesData 3 {
     map string ValueWithExpirationData kv
     list string deletes
     bool calculate_hash
 }
 
-Msg SharedValueData 4{
+Msg SharedValueData 4 {
     string value
     list string categories_ids
 }
 
-Msg SharedUpdatesData 5{
+Msg SharedKeyValueUpdatesData 5 {
     map string SharedValueData kv
     bool calculate_hash
 }
@@ -30,31 +30,81 @@ Msg SharedUpdatesData 5{
 
 # Updates info
 
-Msg MerkleKeyFlag 6{
+Msg MerkleKeyFlag 1000 {
     bool deleted
 }
 
-Msg MerkleUpdatesInfo 7{
+
+Msg MerkleUpdatesInfo 1001 {
     map string MerkleKeyFlag keys
-    fixedlist uint8 32 hash
-    uint64 version
+    fixedlist uint8 32 root_hash
+    uint64 state_root_version
 }
 
-Msg KVKeyFlag 8{
+Msg KVKeyFlag 1002 {
     bool deleted
     bool stale_on_update
 }
 
-Msg KeyValueUpdatesInfo 9{
+Msg KeyValueUpdatesInfo 1003 {
     map string KVKeyFlag keys
-    optional fixedlist uint8 32 hash
+    optional fixedlist uint8 32 root_hash
 }
 
-Msg SharedKeyData 10 {
+Msg SharedKeyData 1004 {
     list string categories
 }
 
-Msg SharedKeyValueUpdatesInfo 11{
+Msg SharedKeyValueUpdatesInfo 1005 {
     map string SharedKeyData keys
-    optional fixedlist uint8 32 hash
+    optional fixedlist uint8 32 root_hash
+}
+
+# Blocks
+
+Msg BlockKey 2000 {
+    uint64 block_id
+}
+
+Msg BlockData 2001 {
+    uint64 block_id
+    fixedlist uint8 32 parent_digest
+    optional SharedKeyValueUpdatesInfo shared_updates_info
+    map string oneof{
+        MerkleUpdatesInfo
+        KeyValueUpdatesInfo
+    } categories_updates_info
+}
+
+Msg RawBlockMerkleUpdates 2002 {
+    MerkleUpdatesData updates
+    fixedlist uint8 32 root_hash
+}
+
+Msg RawBlockKeyValueUpdates 2003 {
+    KeyValueUpdatesData updates
+    optional fixedlist uint8 32 root_hash
+}
+
+Msg RawBlockSharedUpdates 2004 {
+    SharedKeyValueUpdatesData updates
+    optional fixedlist uint8 32 root_hash
+}
+
+Msg RawBlockData 2005 {
+    fixedlist uint8 32 parent_digest
+    map string oneof{
+        RawBlockMerkleUpdates
+        RawBlockKeyValueUpdates
+    } category_updates
+
+    optional RawBlockSharedUpdates shared_update
+}
+
+
+# keys
+
+Msg DbKey 3000 {
+    fixedlist uint8 32 hashed_key
+    uint64 version
 }
--- a/kvbc/include/categorization/blocks.h
+++ b/kvbc/include/categorization/blocks.h
@@ -1,0 +1,164 @@
+#include "updates.h"
+#include <utility>
+#include <block_digest.h>
+#include "details.h"
+#include "merkle_tree_serialization.h"
+#include "merkle_tree_key_manipulator.h"
+#include "sliver.hpp"
+
+namespace concord::kvbc::categorization {
+
+// Block is composed out of:
+// - BlockID
+// - Parent digest
+// - Categories' updates_info where update info is essentially a list of keys, metadata on the keys and a hash of the
+// category state.
+struct Block {
+  static const std::string CATEGORY_ID;
+  Block() {
+    data.block_id = 0;
+    data.parent_digest.fill(0);
+  }
+
+  Block(const BlockId block_id) {
+    data.block_id = block_id;
+    data.parent_digest.fill(0);
+  }
+
+  void add(const std::string& category_id, MerkleUpdatesInfo&& updates_info) {
+    data.categories_updates_info.emplace(category_id, std::move(updates_info));
+  }
+
+  void add(const std::string& category_id, KeyValueUpdatesInfo&& updates_info) {
+    data.categories_updates_info.emplace(category_id, std::move(updates_info));
+  }
+
+  void add(SharedKeyValueUpdatesInfo&& updates_info) { data.shared_updates_info = std::move(updates_info); }
+
+  void setParentHash(const BlockDigest& parent_hash) {
+    std::copy(parent_hash.begin(), parent_hash.end(), data.parent_digest.begin());
+  }
+  inline BlockId id() const { return data.block_id; }
+
+  static const serialization::buffer& serialize(const Block& block) {
+    static thread_local serialization::buffer output;
+    output.clear();
+    concord::kvbc::categorization::serialize(output, block.data);
+    return output;
+  }
+
+  static Block deserialize(const serialization::buffer& input) {
+    Block output;
+    concord::kvbc::categorization::deserialize(input, output.data);
+    return output;
+  }
+
+  // Generate block key from block ID
+  // Using CMF for big endian
+  static const serialization::buffer& generateKey(const BlockId block_id) {
+    static thread_local serialization::buffer output;
+    output.clear();
+    BlockKey key{block_id};
+    // think about optimization
+    concord::kvbc::categorization::serialize(output, key);
+    return output;
+  }
+
+  BlockData data;
+};
+
+const std::string Block::CATEGORY_ID{"__blocks__"};
+
+//////////////////////////////////// RAW BLOCKS//////////////////////////////////////
+
+// Reconstructs the updates data as recieved from the user
+// This set methods are overloaded in order to construct the appropriate updates
+
+// Merkle updates reconstruction
+RawBlockMerkleUpdates getRawUpdates(const std::string& category_id,
+                                    const MerkleUpdatesInfo& update_info,
+                                    const BlockId& block_id,
+                                    const storage::rocksdb::NativeClient* native_client) {
+  // For old serialization
+  using namespace concord::kvbc::v2MerkleTree;
+  RawBlockMerkleUpdates data;
+  // Copy state hash (std::copy?)
+  data.root_hash = update_info.root_hash;
+  // Iterate over the keys:
+  // if deleted, add to the deleted set.
+  // else generate a db key, serialize it and
+  // get the value from the corresponding column family
+  for (auto& [key, flag] : update_info.keys) {
+    if (flag.deleted) {
+      data.updates.deletes.push_back(key);
+      continue;
+    }
+    // E.L see how we can optimize the sliver temporary allocation
+    auto db_key =
+        v2MerkleTree::detail::DBKeyManipulator::genDataDbKey(Key(std::string(key)), update_info.state_root_version);
+    auto val = native_client->get(category_id, db_key);
+    if (!val.has_value()) {
+      throw std::logic_error("Couldn't find value for key");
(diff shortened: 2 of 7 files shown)


kvbc/cmf/categorized_kvbc_msgs.cmf
```diff
@@ -22,39 +22,88 @@ Msg SharedValueData 4{
...
+Msg SharedKeyValueUpdatesData 5{
     map string SharedValueData kv
     bool calculate_hash
 }
 
 
 # Updates info
 
-Msg MerkleKeyFlag 6{
+Msg MerkleKeyFlag 10{
```
dartdart26, line 33: Are the ID gaps used to allow new messages of different types? If yes, maybe we can use a much bigger gaps, something like 100 or 1000. What do you think?

kvbc/cmf/categorized_kvbc_msgs.cmf
```diff
@@ -22,39 +22,88 @@ Msg SharedValueData 4{
...
     bool deleted
 }
 
-Msg MerkleUpdatesInfo 7{
+Msg MerkleUpdatesInfo 11{
     map string MerkleKeyFlag keys
-    fixedlist uint8 32 hash
-    uint64 version
+    fixedlist uint8 32 state_hash
+    uint64 state_root_version
```
dartdart26, line 40: Nit: Maybe these can be unified in a new type as they go hand in hand.

kvbc/cmf/categorized_kvbc_msgs.cmf
```diff
@@ -22,39 +22,88 @@ Msg SharedValueData 4{
...
+    optional SharedKeyValueUpdatesInfo shared_updates_info
+    map string oneof{
+        MerkleUpdatesInfo
+        KeyValueUpdatesInfo
+    } categories_updates_info
+}
+
+Msg RawBlockMerkleUpdates 22 {
+    MerkleUpdatesData updates
+    fixedlist uint8 32 state_hash
```
dartdart26, line 80: Nit: Maybe we can call all hashes for all categories `root_hash`.

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+
+// Block is composed out of:
+// - BlockID
+// - Parent digest
+// - Categories' updates_info where update info is essentially a list of keys, metadata on the keys and a hash of the
+// category state.
+struct Block {
+  static const std::string CATEGORY_ID;
+
+  Block() = default;
```
dartdart26, line 19: Nit: Maybe initialise the block ID to 0 in that case?

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+  Block(const BlockId block_id) { data.block_id = block_id; }
+
+  void add(const std::string& category_id, MerkleUpdatesInfo&& updates_info) {
+    data.categories_updates_info.emplace(std::make_pair(category_id, std::move(updates_info)));
```
dartdart26, line 23: Nit: I think we can omit `std::make_pair()` as we are doing emplace() and we can just pass the arguments.

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+  void add(const std::string& category_id, KeyValueUpdatesInfo&& updates_info) {
+    data.categories_updates_info.emplace(std::make_pair(category_id, std::move(updates_info)));
+  }
+
+  void add(SharedKeyValueUpdatesInfo&& updates_info) { data.shared_updates_info = std::move(updates_info); }
+
+  void setParentHash(const BlockDigest& parent_hash) {
+    std::copy(parent_hash.begin(), parent_hash.end(), data.parent_digest.begin());
+  }
+  inline BlockId ID() const { return data.block_id; }
```
dartdart26, line 35: Nit: Naming, think should be `id()`.

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+
+  static const serialization::buffer& serialize(const Block& block) {
+    static thread_local serialization::buffer output;
```
dartdart26, line 38: Nice! :)

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  void add(SharedKeyValueUpdatesInfo&& updates_info) { data.shared_updates_info = std::move(updates_info); }
+
+  void setParentHash(const BlockDigest& parent_hash) {
+    std::copy(parent_hash.begin(), parent_hash.end(), data.parent_digest.begin());
+  }
+  inline BlockId ID() const { return data.block_id; }
+
+  static const serialization::buffer& serialize(const Block& block) {
```
dartdart26, line 37: Btw, can that be a member (non-static) method?

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+//////////////////////////////////// RAW BLOCKS//////////////////////////////////////
+
+// Reconstructs the updates data as recieved from the user
+// This set methods are overloaded in order to construct the appropriate updates
+
+// Merkle updates reconstruction
+RawBlockMerkleUpdates getRawUpdates(const std::string& category_id,
+                                    const MerkleUpdatesInfo& update_info,
+                                    const BlockId& block_id,
+                                    std::shared_ptr<storage::rocksdb::NativeClient>& native_client) {
```
dartdart26, line 75: Nit: As there's no ownership and no write ops, we can just pass `const NativeClient*`.

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+// shared updates reconstruction
+RawBlockSharedUpdates getRawSharedUpdates(const SharedKeyValueUpdatesInfo& update_info,
+                                          const BlockId& block_id,
+                                          std::shared_ptr<storage::rocksdb::NativeClient>& native_client) {
+  RawBlockSharedUpdates data;
+  return data;
+}
+
+// Raw block is the unit which we can transfer and reconstruct a block in a remote replica.
+// It contains all the relevant updates as recieved from the client in the initial call to storage.
```
dartdart26, line 126: Nice comment!

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+// It also contains:
+// - parent digest
+// - state hash per category (if exists) E.L check why do we pass it.
+struct RawBlock {
+  RawBlock(const Block& block, std::shared_ptr<storage::rocksdb::NativeClient> native_client) {
```
dartdart26, line 131: Nit: Maybe we don't need to bump the ref count on the native client (i.e. use non-owning pointer, as mentioned above).

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+    for (auto& [cat_id, update_info] : block.data.categories_updates_info) {
+      std::visit(
+          [category_id = cat_id, &block, this, &native_client](const auto& update_info) {
+            auto category_updates = getRawUpdates(category_id, update_info, block.ID(), native_client);
+            data.category_updates.emplace(std::make_pair(category_id, std::move(category_updates)));
+          },
+          update_info);
+    }
+    if (block.data.shared_updates_info.has_value()) {
+      auto category_updates = getRawSharedUpdates(block.data.shared_updates_info.value(), block.ID(), native_client);
```
dartdart26, line 144: Nit: Maybe no need for the `category_updates` variable.

kvbc/include/categorization/blocks.h
```diff
@@ -0,0 +1,159 @@
...
+struct RawBlock {
+  RawBlock(const Block& block, std::shared_ptr<storage::rocksdb::NativeClient> native_client) {
+    // parent digest (std::copy?)
+    data.parent_digest = block.data.parent_digest;
+    // recontruct updates of categories
+    for (auto& [cat_id, update_info] : block.data.categories_updates_info) {
+      std::visit(
+          [category_id = cat_id, &block, this, &native_client](const auto& update_info) {
+            auto category_updates = getRawUpdates(category_id, update_info, block.ID(), native_client);
+            data.category_updates.emplace(std::make_pair(category_id, std::move(category_updates)));
```
dartdart26, line 139: Nit: Maybe no need for `std::make_pair()` as mentioned above.

kvbc/include/categorization/details.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace concord::kvbc::categorization {
+
+namespace serialization {
+using buffer = std::vector<uint8_t>;
+}  // namespace serialization
+
+// DB key is a CMF stuct of hashed key and the version.
+// On serialization it will be hashed_key || version(big endian)
+DbKey genDbKey(const std::string& key, const BlockId& version) {
```
dartdart26, line 15: Is that function related to the merkle category? If not, maybe we can be explicit about which hash it uses (i.e. `concord::util::SHA3_256`. Also, maybe we can define a hash type to reuse across `categorization`.

storage/include/rocksdb/native_client.h
```diff
@@ -237,6 +237,17 @@ class NativeClient : public std::enable_shared_from_this<NativeClient> {
...
   }
 
+  template <
+      typename Span,
+      std::enable_if_t<std::is_convertible_v<decltype(std::declval<Span>().size()), std::size_t> &&
+                           std::is_pointer_v<decltype(std::declval<Span>().data())> &&
+                           std::is_convertible_v<std::remove_pointer_t<decltype(std::declval<Span>().data())> (*)[],
+                                                 const uint8_t (*)[]>,
+                       int> = 0>
+  static ::rocksdb::Slice toSlice(const Span &span) noexcept {
```
dartdart26, line 247: Nice!

Nit: Could we please update the comment at the top of the file that mentions NativeClient only accepts vector<char>, span<char> and so on to vector<char/uint8_t>, span<char/uint8_t>.

kvbc/include/categorization/updates.h
```diff
@@ -143,26 +143,27 @@ struct MerkleUpdates {
...
                              category_id};
     }
   }
 
-  void addKeyValueUpdates(std::string category_id, KeyValueUpdates&& updates) {
-    if (const auto [itr, inserted] = category_updates_.insert({category_id, std::move(updates.data_)}); inserted) {
+  void add(std::string&& category_id, KeyValueUpdates&& updates) {
+    if (const auto [itr, inserted] = category_updates_.try_emplace(std::move(category_id), std::move(updates.data_));
+        !inserted) {
       *itr;  // disable unused variable
```
dartdart26, line 158: Ah, I wish one could use `[[maybe_unused]]` with structured bindings. Maybe we can use the `(void)itr` hack to avoid any possible runtime cost from `*`. Not that the compiler will do something on that line in O3...

kvbc/include/categorization/updates.h
```diff
@@ -171,10 +172,12 @@ struct Updates {
     return category_updates_;
   }
 
-  std::optional<SharedUpdatesData>& getSharedUpdates() { return shared_update_; }
+  std::optional<SharedKeyValueUpdatesData>& getSharedUpdates() { return shared_update_; }
```
dartdart26, line 175: I think we should remove that method to avoid users calling it (now that `KeyValueBlockchain` is a friend).

kvbc/include/categorization/kv_blockchain.h
```diff
@@ -16,64 +16,87 @@
...
 namespace concord::kvbc::categorization {
 
-struct Block {
-  void add(const std::string& category_id, std::variant<MerkleUpdatesInfo, KeyValueUpdatesInfo> updates_info) {}
-  void add(SharedKeyValueUpdatesInfo updates_info) {}
-};
 class KeyValueBlockchain {
  public:
   KeyValueBlockchain() : native_client_(NativeClient::newClient("/tmp", false, NativeClient::DefaultOptions{})) {}
+  KeyValueBlockchain(std::shared_ptr<NativeClient> native_client) : native_client_(native_client) {}
```
dartdart26, line 30: Nit: `const std::shared_ptr<>&` ?

kvbc/include/categorization/kv_blockchain.h
```diff
@@ -16,64 +16,87 @@
...
   // 1) Defines a new block
   // 2) calls per cateogry with its updates
   // 3) inserts the updates KV to the DB updates set per column family
   // 4) add the category block data into the new block
-  BlockId addBlock(Updates& updates) {
+  BlockId addBlock(Updates&& updates) {
```
dartdart26, line 35: Wow, that function got simpler now :)

kvbc/test/categorization/updates_test.cpp
```diff
@@ -23,10 +24,131 @@ TEST(categorized_updates, merkle_update) {
   ASSERT_TRUE(mu.getData().kv.size() == 1);
 }
 
-TEST(categorized_updates, add_block) {
-  KeyValueBlockchain block_chain;
-  Updates updates;
-  ASSERT_TRUE(block_chain.addBlock(updates) == (BlockId)8);
+TEST(categorized_updates, add_blocks) {
+  auto native_client = NativeClient::newClient("/tmp/add_block", false, NativeClient::DefaultOptions{});
```
dartdart26, line 28: The storage_test_common.h file contains native client handling and cleanup. Can be reused here too.

dartdart26 (approved):
Thanks @cloudnoize ! Looks great! I have put comments that you can decide whether to handle or not or do it later.

kvbc/cmf/categorized_kvbc_msgs.cmf
cloudnoize: I didn't see it complain so I guess it's allowed. I'll enlarge the gap

kvbc/cmf/categorized_kvbc_msgs.cmf
cloudnoize: yep

kvbc/include/categorization/blocks.h
cloudnoize: true

kvbc/include/categorization/blocks.h
cloudnoize: yeah, I made it on purpose, seems nice to have as a utility function under the Block name.
what do you think?

kvbc/include/categorization/blocks.h
cloudnoize: yep

kvbc/include/categorization/blocks.h
cloudnoize: yep, I used to work like this, but some people didn't like it :)

kvbc/include/categorization/details.h
cloudnoize: it's not related to the Merkle, it's related to all but the Merkle :) as they can use the new CMF.
So yes we can define a category hash type and use it.

kvbc/include/categorization/blocks.h
dartdart26: Bad people :) I mean we just use the client in the scope, we don't save it as a member or anything. I think that's the typical pattern for C++. If we were saving it somewhere that's a different topic, but this ctor is just a function.

Better yet - we can use a reference! :)

kvbc/include/categorization/kv_blockchain.h
cloudnoize: indeed

kvbc/include/categorization/blocks.h
dartdart26: I was thinking you need an object to call it and maybe you can make it a `const` method. No big deal.

PR #1082, opened 2020-12-08
BCStateTran: Add ST FetchBlockMsg <-> ItemDataMsg  latency histogram

glevkovich:
Here we add s FetchBlockMsg <->ItemDataMsg  latency histogram.
This message exchange process is between destination (collecting)
replica and a chosen source. Every time such a process ends, a
cumulative percentile latency histogram is snapshotted and logged.

To record the time passed between send and reply, we use an
AsyncTimeRecorderMap atomic recorder. When FetchBlockMsg is sent, we
call start() and insert the message sequence number. When a valid
ItemDataMsg reply arrives, we call end() with the reply sequence number.


andrewjstone (approved):
Looks great! Nice job @glevkovich. I'm super happy to see histograms in state transfer :)

PR #1083, opened 2020-12-09
add read only adapter to sparse_merkle_tree_editor

yontyon:


dartdart26 (approved):
Thank you @yontyon ! That is really useful!

Nit: The flags can be made const.

PR #1084, opened 2020-12-09
Implement the SharedKeyValueCategory::add() method

dartdart26:
Introduce the shared key-value category that persists key-values across
multiple categories. If a key-value is part of multiple categories in a
single block update, the value will be physically stored once only in
one of the categories and other categories will contain a key with a
`pointer` to the value. Additionally, support key-value proofs per
category and per block.

Add a unit test to verify the SharedKeyValueCategory::add() method.

Fix EVPHash's move ctor and assignment operator.

Re-generate code on changes to CMF files or to CMF's C++ serialization
library. Accept non-const reference to pointer to const (i.e. const
std::uint8_t*&) in deserialization methods in order to be able to
consume const containers and avoid const_cast.

Add a benchmark to evaluate any performance gains of using
`thread_local` buffers for serialization. Usage is still TBD.

Make sure categorization and RocksDB NativeClient (and tests) are built
only when RocksDB is being used.

Optimize column family lookups via a cache in NativeClient in order to
support dynamic column family creation.

Remaining SharedKeyValueCategory functionality will be provided in
separate commits.

### DIFF
--- a/kvbc/include/categorization/column_families.h
+++ b/kvbc/include/categorization/column_families.h
@@ -17,8 +17,8 @@
 
 namespace concord::kvbc::categorization::detail {
 
-inline const auto SHARED_KV_DATA_CF_SUFFIX = std::string{"_skvdata"};
-inline const auto SHARED_KV_LATEST_KEY_VER_CF_SUFFIX = std::string{"_skvlatest"};
+inline const auto SHARED_KV_DATA_CF = std::string{"shared_kv_data"};
+inline const auto SHARED_KV_KEY_VERSIONS_CF = std::string{"shared_kv_key_versions"};
 
 inline const auto BLOCKS_CF = std::string{"blocks"};
 
--- a/kvbc/include/categorization/column_families.h
+++ b/kvbc/include/categorization/column_families.h
@@ -1,0 +1,25 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <string>
+
+namespace concord::kvbc::categorization::detail {
+
+inline const auto SHARED_KV_DATA_CF_SUFFIX = std::string{"_skvdata"};
+inline const auto SHARED_KV_LATEST_KEY_VER_CF_SUFFIX = std::string{"_skvlatest"};
+
+inline const auto BLOCKS_CF = std::string{"blocks"};
+
+}  // namespace concord::kvbc::categorization::detail
--- a/kvbc/include/categorization/shared_kv_category.h
+++ b/kvbc/include/categorization/shared_kv_category.h
@@ -44,21 +44,23 @@ class SharedKeyValueCategory {
   SharedKeyValueCategory(const std::shared_ptr<storage::rocksdb::NativeClient> &);
 
   // Add the given block updates and return the information that needs to be persisted in the block.
-  SharedKeyValueUpdatesInfo add(BlockId block_id, SharedKeyValueUpdatesData &&update, storage::rocksdb::WriteBatch &);
+  SharedKeyValueUpdatesInfo add(BlockId block_id,
+                                SharedKeyValueUpdatesData &&update,
+                                storage::rocksdb::NativeWriteBatch &);
 
   // Delete the genesis block.
   // `updates_info` must be the updates for `block_id`.
   // Precondition: `block_id` is the current genesis block ID.
   void deleteGenesisBlock(BlockId block_id,
                           const SharedKeyValueUpdatesInfo &updates_info,
-                          storage::rocksdb::WriteBatch &);
+                          storage::rocksdb::NativeWriteBatch &);
 
   // Delete the last reachable block.
   // `updates_info` must be the updates for `block_id`.
   // Precondition: `block_id` is the current last reachable block ID.
   void deleteLastReachableBlock(BlockId block_id,
                                 const SharedKeyValueUpdatesInfo &updates_info,
-                                storage::rocksdb::WriteBatch &);
+                                storage::rocksdb::NativeWriteBatch &);
 
   // Get the value of a key in `category_id` with a block version up to `max_block_id`.
   // Return std::nullopt if the key doesn't exist in any earlier blocks than `max_block_id`.
--- a/kvbc/include/categorization/shared_kv_category.h
+++ b/kvbc/include/categorization/shared_kv_category.h
@@ -27,7 +27,10 @@ namespace concord::kvbc::categorization::detail {
 // The shared key-value category persists key-values across multiple categories. A category is implemented as a pair of
 // RocksDB column families:
 //  * `data` column family - contains versioned key-values for the given category
-//  * `latest version` column family - contains the latest version of a key for the given category
+//  * `latest version` column family - contains the latest version of a key for the given category. The rationale behind
+//    storing the latest version of a key is to support latest version key lookup without using RocksdDB iterators.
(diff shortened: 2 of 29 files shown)


kvbc/include/categorization/shared_kv_category.h
```diff
@@ -0,0 +1,89 @@
...
+  // Delete the last reachable block.
+  // `updates_info` must be the updates for `block_id`.
+  // Precondition: `block_id` is the current last reachable block ID.
+  void deleteLastReachableBlock(BlockId block_id,
+                                const SharedKeyValueUpdatesInfo &updates_info,
+                                storage::rocksdb::WriteBatch &);
+
+  // Get the value of a key in `category_id` with a block version up to `max_block_id`.
+  // Return std::nullopt if the key doesn't exist in any earlier blocks than `max_block_id`.
+  std::optional<Value> getValueUpToBlock(const std::string &category_id,
```
cloudnoize, line 62: It's a good name but almost identical to ```getValueAtBlock``` with the same signature, I guess one can use the other method by mistake or vice vera

kvbc/include/categorization/shared_kv_category.h
```diff
@@ -0,0 +1,89 @@
...
+
+#include <memory>
+#include <string>
+
+namespace concord::kvbc::categorization::detail {
+
+// The shared key-value category persists key-values across multiple categories. A category is implemented as a pair of
+// RocksDB column families:
+//  * `data` column family - contains versioned key-values for the given category
+//  * `latest version` column family - contains the latest version of a key for the given category
```
cloudnoize, line 30: maybe add a sentence on the usage of this column family

kvbc/src/categorization/shared_kv_category.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include <stdexcept>
+#include <utility>
+
+using namespace std::literals;
+
+namespace concord::kvbc::categorization::detail {
+
+const Buffer &version(BlockId block_id) {
+  static thread_local auto buf = Buffer{};
+  buf.clear();
```
cloudnoize, line 31: I was thinking since we have multiple places where we store those static buffers per method.
And as we spoke offline a buffer can only grow, maybe we should add a clear method that does
```
buf.clear();
if(buf.capacity() > MAX_CAP)buf.shrink_to_fit();
````

File: kvbc/src/categorization/shared_kv_category.cpp [hunk 4]
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      // Versioned key-value.
+      if (!first_category) {
+        batch.put(data_cf, serialize(versioned_key), sharedValueData(std::move(shared_value.value)));
+        first_category = category_id;
+      } else {
+        batch.put(data_cf, serialize(versioned_key), sharedValuePointer(*first_category));
+      }
+
```
cloudnoize, line 116: An idea, maybe it would be simpler to have a dedicated column family for the value (global to the shared cat) where all other ```per category ID``` columns point to.
Then all there is to do is to mark versioned_key->True at the corresponding category ids.

kvbc/src/categorization/shared_kv_category.cpp
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  // Finish hash calculation per category.
+  if (update.calculate_root_hash) {
+    auto category_hashes = std::map<std::string, Hash>{};
+    for (auto &[category_id, hasher] : category_hashers) {
+      category_hashes[category_id] = hasher.finish();
+    }
+    update_info.category_root_hashes = std::move(category_hashes);
+  }
```
cloudnoize, line 132: Want to verify that I understand the functionality. 
A hash of a category ID is the hash of all the KV of that category **only at that version**  right? it doesn't calculate any KV of that category from past versions

kvbc/src/categorization/shared_kv_category.cpp
cloudnoize: For example, how are you going to distinguish between a pointer or a value on get ?

kvbc/include/categorization/column_families.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include <string>
+
+namespace concord::kvbc::categorization::detail {
+
+inline const auto SHARED_KV_DATA_CF_SUFFIX = std::string{"_skvdata"};
+inline const auto SHARED_KV_LATEST_KEY_VER_CF_SUFFIX = std::string{"_skvlatest"};
+
+inline const auto BLOCKS_CF = std::string{"blocks"};
```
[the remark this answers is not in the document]
dartdart26: I was meaning to do it, but forgot. Let me do it as part of this PR.

kvbc/include/categorization/shared_kv_category.h
dartdart26: Good call, doing that.

kvbc/include/categorization/shared_kv_category.h
dartdart26: getValueUntilBlock()?

kvbc/src/categorization/shared_kv_category.cpp
dartdart26: Currently, we serialize a variant and that's how we know which value is a pointer and which one is an actual value.

I think your solution is better, because we don't need to store the variant ID in the value and we don't need to parse it, we can just lookup in the category and find if the key is part of it or not.

Great catch, thanks! Will change it.

kvbc/src/categorization/shared_kv_category.cpp
dartdart26: Yes, that's correct.

storage/include/rocksdb/native_client.h
```diff
@@ -186,70 +91,55 @@ class NativeClient : public std::enable_shared_from_this<NativeClient> {
...
+  // Return the DB path.
+  const std::string &path() const { return client_->m_dbPath; }
 
-  // Column family management.
+  // On-disk column family management.
   static std::string defaultColumnFamily();
+  // Return the column families in the DB pointed to by `path`.
   static std::unordered_set<std::string> columnFamilies(const std::string &path);
+
+  // Client instance column management. Methods below only operate on column families this client is ware of. The actual
```
andrewjstone, line 119: Nit: typo: `ware` -> `aware`

andrewjstone (approved):
Looks great @dartdart26! There's a bunch of code in here I can use ASAP. Specifically the hashing and column family related code.

Can we merge this in and make any necessary changes in a follow up?

kvbc/src/categorization/shared_kv_category.cpp
```diff
@@ -0,0 +1,142 @@
...
+    // Make sure we've inserted the key for the first time.
+    ConcordAssert(inserted_key);
+    auto &key_data = key_it->second;
+
+    const auto value_hash = hash(shared_value.value);
+    auto first_category = std::optional<std::string>{std::nullopt};
+    for (auto &&category_id : shared_value.category_ids) {
+      const auto data_cf = dataCf(category_id);
+      const auto latest_version_cf = latestVersionCf(category_id);
+      createColumnFamilyIfNotExisting(data_cf);
```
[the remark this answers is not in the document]
dartdart26: Yeah, this is totally changed in my next commit (either commit or a new PR). We have 2 CFs only and nothing is created during add.

kvbc/include/categorization/shared_kv_category.h
andrewjstone: My suggestion is to change `getValueAtBlock` to just `get`.

kvbc/src/categorization/shared_kv_category.cpp
dartdart26: After further discussions with @andrewjstone, key versions per category are implemented via the `KeyVersionsPerCategory` type that contains a mapping from a category ID to a list of key versions for that category ID. That allows implementations for `get()` methods without using RocksDB iterators. Assumption is that keys in the shared category will not have many versions. If we need many versions, we can define a new category.

kvbc/include/categorization/shared_kv_category.h
dartdart26: Will change in the next commit that will implement these methods. Thanks!

PR #1085, opened 2020-12-09
Add pruning status to controlManager and controlHandlers

yontyon:
In order to enable async pruning, we want to have the ability to signal the bft layer that we are in the middle of pruning.
For that, we added the necessary logic.
Once the command handler initiates pruning it marks in the controlMnager that pruning is on the fly. From this point until pruning is done, the replicas will answer only read requests (the same way we do with wedge).

Notice, that in case of a crash, the same mechanism that returns pruning will set the pruning indicator to true (in the command handler layer)

### DIFF
--- a/bftengine/include/bftengine/ControlStateManager.hpp
+++ b/bftengine/include/bftengine/ControlStateManager.hpp
@@ -55,6 +55,8 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
 
   void clearCheckpointToStopAt();
 
+  void setPruningProcessOn(bool onPruningProcess) { onPruningProcess_ = onPruningProcess; }
+  bool getPruningProcessStatus() { return onPruningProcess_; }
   ControlStateManager(IStateTransfer* state_transfer, uint32_t sizeOfReservedPages);
   ControlStateManager& operator=(const ControlStateManager&) = delete;
   ControlStateManager(const ControlStateManager&) = delete;
@@ -69,5 +71,6 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
   std::string scratchPage_;
   bool enabled_ = true;
   ControlStatePage page_;
+  bool onPruningProcess_ = false;
 };
 }  // namespace bftEngine
--- a/bftengine/include/bftengine/ControlStateManager.hpp
+++ b/bftengine/include/bftengine/ControlStateManager.hpp
@@ -55,7 +55,7 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
 
   void clearCheckpointToStopAt();
 
-  void setPruningProcessOn(bool onPruningProcess) { onPruningProcess_ = onPruningProcess; }
+  void setPruningProcess(bool onPruningProcess) { onPruningProcess_ = onPruningProcess; }
   bool getPruningProcessStatus() { return onPruningProcess_; }
   ControlStateManager(IStateTransfer* state_transfer, uint32_t sizeOfReservedPages);
   ControlStateManager& operator=(const ControlStateManager&) = delete;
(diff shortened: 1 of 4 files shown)


bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -55,6 +55,8 @@ class ControlStateManager : public ResPagesClient<ControlStateManager,
 
   void clearCheckpointToStopAt();
 
+  void setPruningProcess(bool onPruningProcess) { onPruningProcess_ = onPruningProcess; }
+  bool getPruningProcessStatus() { return onPruningProcess_; }
```
dartdart26, line 59: This can be `const`.

dartdart26:
@yontyon Something that came to mind is what should happen if Pruning is executing and a client sends a read-only request whose response depends on which blocks are present in every replica. If this client request doesn't work with replica-specific information, it can fail. Do I miss something? What do you think?

One more thing I noticed in https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L198 and https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L226 is that we return without deleting the message. I am not sure, but it seems like a memory leak. Do I miss something?

yontyon:
> https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L226

1. Yes, but do we have another such request other than get_latest_prunebale_block?
2. good catch, I fixed it in this PR

yontyon:
@dartdart26 

W should not delete the message after  https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L226  because we enter it to the queue.
As for  https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L198 you are right (the message is ignored)

dartdart26:
> > @yontyon Something that came to mind is what should happen if Pruning is executing and a client sends a read-only request whose response depends on which blocks are present in every replica. If this client request doesn't work with replica-specific information, it can fail. Do I miss something? What do you think?
> > One more thing I noticed in https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L198 and https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L226 is that we return without deleting the message. I am not sure, but it seems like a memory leak. Do I miss something?
> 
> @dartdart26
> 
> 1. Yes, but do we have another such request other than get_latest_prunebale_block?
> 2. good catch, I fixed it in this PR

1. My thinking was that the API of the replica allows users (i.e. execution engines) to lookup old keys and blocks. Therefore, if a generic application accesses historical blocks, it can lead to different responses from different replicas. Maybe we can only allow read-only requests that return RSI if the system is in such a state. Not sure if that is easily done, though.

Bottom line - what I wanted to say it is application-dependent and might not work in all cases.

Another question I have is related to `isSeqNumToStopAt()`. It is called in 3 cases:

- on **ClientRequestMsg**
- on **PrePrepareMsg**
- when calling **checkSendPrePrepareMsgPrerequisites()**

Do you think we should only limit that change to **ClientRequestMsg** ? I am not sure of the possible implications of the other  two cases and the interactions with Pruning.

I think @HristoStaykov can assist us and shed some more light.

dartdart26:
I had an offline discussion with @HristoStaykov about async execution as a whole and it might be a big and complex topic for the future. As we are now single-threaded, it seems we won't be able to respond to any requests during Pruning. If I am not mistaken, Wedge is different, because it doesn't block the execute callback as Pruning does. I was thinking maybe we can postpone that change to the point when we have async execution.

yontyon:
@dartdart26 lets continue offline

dartdart26 (approved):
Thanks @yontyon ! I've been convinced in an offline discussion :)

PR #1086, opened 2020-12-10
[Apollo] Fix flaky checkpoint test

teoparvanov:
As per the analysis by @robem and @shgandhi:
_The test checks whether we are able to create a checkpoint or not. It clearly succeeds however, the test expected that we would have created CP 1 instead of 2. This might be a flaky test due to we are reading "before checkpoint" twice. Once from replica 0 and internally (lib function) from a random replica. If for whatever reason, those replicas are not in sync then it would explain the difference._
...
_We read the `before checkpoint` twice, that is most likely causing this. To avoid the flakiness, we can do away with the assertion check in this specific test, because the call to `fill_and_wait_for_checkpoint` essentially does the same thing._


shgandhi (approved):
Thanks for the change Teo. LGTM.

PR #1087, opened 2020-12-11
Storage related improvements

andrewjstone:
* Add a larger cache and bloom filter to RocksDB configuration
 * Enable more RocksDB stats
 * Remove duplicate configuration of write_buffer_size in RocksDB
 configuration
 * Shorten name of kvbc histograms
 * Use microseconds instead of nanoseconds for several sparse merkle
 histograms and add two more for the sparse merkle tree walker.

### DIFF
--- a/kvbc/include/ReplicaImp.h
+++ b/kvbc/include/ReplicaImp.h
@@ -150,7 +150,7 @@ class ReplicaImp : public IReplica,
   struct Recorders {
     Recorders() {
       auto &registrar = concord::diagnostics::RegistrarSingleton::getInstance();
-      registrar.perf.registerComponent("kvbc-ReplicaImp",
+      registrar.perf.registerComponent("kvbc",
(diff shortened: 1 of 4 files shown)


storage/include/rocksdb/native_client.h
```diff
@@ -79,6 +79,8 @@ class WriteBatch {
   template <typename BeginSpan, typename EndSpan>
   void delRange(const BeginSpan &beginKey, const EndSpan &endKey);
 
+  size_t size();
```
dartdart26, line 82: const?

dartdart26 (approved):
Thanks Andrew!

storage/include/rocksdb/native_client.h
andrewjstone: whoopsy
you are correct.

andrewjstone: fixed.

PR #1088, opened 2020-12-14
Fix CMF fixedlist composition

dartdart26:
Add missing (de)serialize forward declarations for the `fixedlist` type
(i.e. std::array). This was preventing `fixedlist` composition with
other types, i.e. std::vector<std::array<>>.

Fix a comment about std::array initialization in the cppgen test - be
explicit and say we only initialize the first std::array element with a
random value.


cloudnoize (approved):
LGTM

PR #1089, opened 2020-12-14
Add delete block to the categorized DB adapter

cloudnoize:
Support delete block via a similar mechanism used for the add block i.e. std::visit and per-category delete.
Support for delete genesis and last reachable.

The logic was mostly migrated from the Merkle DB adapter and was adopted to support categorization.

Under the block file, a ```BlockChain``` class was added, in order to encapsulate the interaction with the blockchain itself.
This class also defines a ```StateTransfer``` inner class for interaction with the state-transfer chain.

Files were divided into smaller grains and into cpp files as well.

### DIFF
--- a/kvbc/include/categorization/blocks.h
+++ b/kvbc/include/categorization/blocks.h
@@ -21,6 +21,7 @@
 #include "merkle_tree_key_manipulator.h"
 #include "merkle_tree_db_adapter.h"
 #include "sliver.hpp"
+#include "column_families.h"
 
 namespace concord::kvbc::categorization {
 
@@ -53,30 +54,23 @@ struct Block {
   void setParentHash(const BlockDigest& parent_hash) {
     std::copy(parent_hash.begin(), parent_hash.end(), data.parent_digest.begin());
   }
-  inline BlockId id() const { return data.block_id; }
 
-  static const detail::Buffer& serialize(const Block& block) {
-    static thread_local detail::Buffer output;
-    output.clear();
-    concord::kvbc::categorization::serialize(output, block.data);
-    return output;
-  }
+  BlockId id() const { return data.block_id; }
 
-  static Block deserialize(const detail::Buffer& input) {
+  static const detail::Buffer& serialize(const Block& block) { return detail::serialize(block.data); }
+
+  template <typename T>
+  static Block deserialize(const T& input) {
     Block output;
-    concord::kvbc::categorization::deserialize(input, output.data);
+    detail::deserialize(input, output.data);
     return output;
   }
 
   // Generate block key from block ID
   // Using CMF for big endian
   static const detail::Buffer& generateKey(const BlockId block_id) {
-    static thread_local detail::Buffer output;
-    output.clear();
     BlockKey key{block_id};
-    // think about optimization
-    concord::kvbc::categorization::serialize(output, key);
-    return output;
+    return detail::serialize(key);
   }
 
   BlockData data;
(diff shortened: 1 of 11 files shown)


kvbc/include/categorization/blocks.h
```diff
@@ -53,116 +54,132 @@ struct Block {
...
   // Generate block key from block ID
   // Using CMF for big endian
   static const detail::Buffer& generateKey(const BlockId block_id) {
-    static thread_local detail::Buffer output;
-    output.clear();
     BlockKey key{block_id};
-    // think about optimization
-    concord::kvbc::categorization::serialize(output, key);
-    return output;
+    return concord::kvbc::categorization::detail::serialize(key);
```
dartdart26, line 73: Nit: I suppose we can just do `detail::serialize`.

kvbc/include/categorization/blocks.h
```diff
@@ -53,116 +54,132 @@ struct Block {
...
   }
 
   BlockData data;
 };
 
-//////////////////////////////////// RAW BLOCKS//////////////////////////////////////
+class BlockChain {
```
dartdart26, line 79: Nit: I think the typical spelling is `Blockchain`.

kvbc/include/categorization/blocks.h
```diff
@@ -53,116 +54,132 @@ struct Block {
...
+ public:
+  static constexpr auto MAX_BLOCK_ID = std::numeric_limits<BlockId>::max();
+  static constexpr auto INITIAL_GENESIS_BLOCK_ID = BlockId{1};
+
+  BlockChain(const storage::rocksdb::NativeClient& native_client);
```
dartdart26, line 84: Does it make sense to copy the RocksDB shared_ptr as we are storing it as a member?

kvbc/include/categorization/blocks.h
```diff
@@ -53,116 +54,132 @@ struct Block {
...
+  inline void setLastReachableBlockId(const BlockId id) { last_reachable_block_id_ = id; }
+  inline BlockId getLastReachableBlockId() const { return last_reachable_block_id_; }
+
+  // Genesis
+  std::optional<BlockId> loadGenesisBlockId();
+  inline void setGenesisBlockId(const BlockId id) { genesis_block_id_ = id; }
+  inline BlockId getGenesisBlockId() const { return genesis_block_id_; }
+
+  // both
+  inline void setAddedBlockId(const BlockId last) {
```
dartdart26, line 98: `inline` is redundant in these use cases.

kvbc/include/categorization/blocks.h
dartdart26, line 79: I think this class may be put into the `detail` namespace as it is not supposed to be exposed to users.

kvbc/include/categorization/blocks.h
```diff
@@ -53,116 +54,132 @@ struct Block {
...
-            auto category_updates = getRawUpdates(category_id, update_info, block.id(), native_client);
-            data.category_updates.emplace(category_id, std::move(category_updates));
-          },
-          update_info);
-    }
-    if (block.data.shared_updates_info.has_value()) {
-      data.shared_update = getRawSharedUpdates(block.data.shared_updates_info.value(), block.id(), native_client);
-    }
-  }
+  RawBlock(const Block& block, const storage::rocksdb::NativeClient* native_client);
```
dartdart26, line 182: Nit: Maybe we can use a reference for the client.

kvbc/include/categorization/kv_blockchain.h
```diff
@@ -24,88 +24,75 @@ namespace concord::kvbc::categorization {
...
+        block_chain_{*native_client_.get()},
+        state_transfer_block_chain_{*native_client_.get()} {}
+
+  /////////////////////// Add Block ///////////////////////
+
+  BlockId addBlock(Updates&& updates);
+
+  /////////////////////// Delete block ///////////////////////
+
+  bool deleteBlock(const BlockId& blockId);
```
dartdart26, line 38: I think since we don't need to support the IDbAdapter interface anymore, we might split to `deleteLastReachable()` and `deleteGenesis()`.

kvbc/src/categorization/blocks.cpp
```diff
@@ -0,0 +1,132 @@
...
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include "categorization/blocks.h"
+
+namespace concord::kvbc::categorization {
+
+BlockChain::BlockChain(const storage::rocksdb::NativeClient& native_client) : native_client_{native_client} {
+  auto LastReachableBlockId = loadLastReachableBlockId();
```
dartdart26, line 19: Nit: Naming difference - camelCase vs _.

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+/**
+ * The following test suite tests ordering of KeyValuePairs
```
dartdart26, line 15: Nit: Is this comment relevant?

dartdart26 (approved):
Looks good, thanks @cloudnoize !

I think we should add more tests, but maybe it makes sense to do it once more categories are ready.

kvbc/include/categorization/blocks.h
cloudnoize: I think it does actually, I had doubts regarding this, thanks!

kvbc/include/categorization/kv_blockchain.h
cloudnoize: As discussed offline, the deleteBlock method has state transfer blocks handling as well.
We'll define an appropriate interface for the deletion.

PR #1090, opened 2020-12-14
Remove the C++ serialization lib as CMF dependency

dartdart26:
Using Concord-BFT as a submodule is problematic when using a path that
is relative to CMAKE_SOURCE_DIR . Will fix in a future commit.


yuliasherman (approved):
LGTM

PR #1093, opened 2020-12-15
Use gauge for last_saved_block_id S3 metric

tdimitrov:
Gauge is more suitable type for the metric as it represents a numeric
value.

### DIFF
--- a/storage/include/s3/metrics.hpp
+++ b/storage/include/s3/metrics.hpp
@@ -8,17 +8,17 @@ namespace concord::storage::s3 {
 class Metrics {
   public:
   Metrics()
-      : metrics_component_{concordMetrics::Component("s3", std::make_shared<concordMetrics::Aggregator>())},
-        numKeysTransferred{metrics_component_.RegisterCounter("keys_transferred")},
-        bytesTransferred{
-            metrics_component_.RegisterCounter("bytes_transferred"),
+      : metrics_component{concordMetrics::Component("s3", std::make_shared<concordMetrics::Aggregator>())},
+        num_keys_transferred{metrics_component.RegisterCounter("keys_transferred")},
+        bytes_transferred{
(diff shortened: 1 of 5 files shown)


storage/include/s3/metrics.hpp
```diff
@@ -35,10 +36,23 @@ struct Metrics {
...
+      return;
+    } catch(std::exception& e) {
+      LOG_ERROR(logger_, "Unexpected error occured while converting lastSavedBlockId (" << elems[elems.size() - 2] << ") to numeric value.");
+      return;
+    }
+    lastSavedBlockId_.Get().Set(lastSavedBlockVal);
   }
 
-  std::string getLastSavedBlockId() { return lastSavedBlockId_.Get().Get(); }
+  uint64_t getLastSavedBlockId() { return lastSavedBlockId_.Get().Get(); }
```
f-squirrel, line 55: Nit: `BlockId`?

storage/include/s3/metrics.hpp
```diff
@@ -35,10 +36,23 @@ struct Metrics {
     // the format is: "PREFIX/BLOCK_ID/KEY", where PREFIX is optional
     if (elems.size() < 2) return;
 
-    lastSavedBlockId_.Get().Set(elems[elems.size() - 2]);
+    uint64_t lastSavedBlockVal = 0;
```
f-squirrel, line 39: Nit: As far as I see, the file uses snake_case, so probably`last_saved_block_val`?
P.S. It is really frustrating that we have both...

f-squirrel: Actually, it is mixed... so ignore, please.

tdimitrov: Good point. I fixed the naming in the whole file. It's a new source so it should be styled correctly.

PR #1095, opened 2020-12-15
Introduce the ImmutableKeyValueCategory

dartdart26:
Replace the existing `SharedKeyValueCategory` with
`ImmutableKeyValueCategory`.

The ImmutableKeyValueCategory stores tagged keys directly in RocksDB.
Keys are not versioned. Updating a key's value or its tags is undefined
behavior. Explicit deletes are not supported. Key-values are
automatically made stale on addition.

Support multiple categories of the ImmutableKeyValueCategory type.

Proofs for keys tagged with the same tag in a block are supported as:
`root = h(h(k1) || h(v1) || h(k2) || h(v2) || ... || h(kn) || h(vn))`

Add unit tests to verify the new functionality.


kvbc/src/categorization/immutable_kv_category.cpp
```diff
@@ -0,0 +1,172 @@
...
+      continue;
+    }
+
+    const auto update_value = get(update_key);
+    ConcordAssert(update_value.has_value());
+
+    proof.ordered_complement_kv_hashes.push_back(hash(update_key));
+    proof.ordered_complement_kv_hashes.push_back(hash(update_value->data));
+
+    i += 2;
```
cloudnoize, line 166: Why is it +2 ?

kvbc/src/categorization/immutable_kv_category.cpp
```diff
@@ -0,0 +1,172 @@
...
+  auto i = std::size_t{0};
+  for (const auto &[update_key, update_key_tags] : updates_info.tagged_keys) {
+    if (update_key == key) {
+      proof.key_value_index = i;
+      continue;
+    }
+
+    if (std::find(update_key_tags.cbegin(), update_key_tags.cend(), tag) == update_key_tags.cend()) {
+      continue;
+    }
```
cloudnoize, line 158: Is it valid that the input key may not have the given tag but it still affects the proof ?

kvbc/src/categorization/immutable_kv_category.cpp
```diff
@@ -0,0 +1,172 @@
...
+  proof.block_id = value->block_id;
+  proof.key = key;
+  proof.value = std::move(value->data);
+
+  auto i = std::size_t{0};
+  for (const auto &[update_key, update_key_tags] : updates_info.tagged_keys) {
+    if (update_key == key) {
+      proof.key_value_index = i;
+      continue;
+    }
```
cloudnoize, line 154: Why the key is treated differently and not hashed and inserted to ```ordered_complement_kv_hashes``` as the other keys.
Eventually, it will be done in ```calculateRootHash`` if I understand correctly

kvbc/include/categorization/base_types.h
```diff
@@ -43,23 +38,28 @@ struct KeyValueProof {
...
     hasher.init();
+    const auto update = [&hasher](const auto &in) { hasher.update(in.data(), in.size()); };
+
     if (ordered_complement_kv_hashes.empty()) {
-      hasher.update(key_hash.data(), key_hash.size());
-      hasher.update(value_hash.data(), value_hash.size());
+      update(key_hash);
+      update(value_hash);
     } else {
       for (auto i = 0ul; i < ordered_complement_kv_hashes.size(); ++i) {
```
cloudnoize, line 53: nit: I think it's more readable to perform ```return hasher.finish()``` inside the if and return.
Then to remove the following ```else``` which makes it looks less important.

kvbc/include/categorization/base_types.h
```diff
@@ -43,23 +38,28 @@ struct KeyValueProof {
...
+    // root_hash = h(h(k1) || h(v1) || h(k2) || h(v2) || ... || h(kn) || h(vn))
+    auto hasher = Hasher{};
     hasher.init();
+    const auto update = [&hasher](const auto &in) { hasher.update(in.data(), in.size()); };
+
     if (ordered_complement_kv_hashes.empty()) {
-      hasher.update(key_hash.data(), key_hash.size());
-      hasher.update(value_hash.data(), value_hash.size());
+      update(key_hash);
+      update(value_hash);
```
cloudnoize, line 51: nit: I think it's more readable to perform return hasher.finish() inside the if and return.
Then to remove the following else which makes it looks less important.

dartdart26: I think you are right, I will change it. Thanks!

kvbc/src/categorization/immutable_kv_category.cpp
dartdart26: It shouldn't affect the proof. If I am understanding you correctly, we are "safe", because the keys in the updates_info are unique (i.e. they are in a map), meaning if a key is equal to one of the keys in the update, then that is the actual key we are interested in. Also, in the beginning, we check if the key we want proof for has the requested tag.

Please let me know if I got what you mean.

kvbc/src/categorization/immutable_kv_category.cpp
dartdart26: The intention is for the client to calculate the actual hash of the key-value being proven, instead of relying on a hash calculated here.

kvbc/src/categorization/immutable_kv_category.cpp
dartdart26: I got it now in an offline discussion. You are right, we don't check if the `key` is tagged with `tag`. Great catch, fill add a test for it and fix. Thanks!

cloudnoize (approved):
Awesome @dartdart26 !

PR #1096, opened 2020-12-16
Error handling for protocol config parameter in S3 client

tdimitrov:
Check if valid protocol is provided to the S3 client and log error on
failure.


f-squirrel (approved):
LGTM!

PR #1098, opened 2020-12-17
Add handling for state transfer blockchain

cloudnoize:
Defining a class (```detail::Blockchain::StateTransfer```) that manages the interaction with the state transfer blockchain.
Migrating the caching logic of ```last_reachable_block_id``` and ```last_block_id``` i.e. last state transfer block id.

Implementing the ```add``` and ```get``` operations on raw blocks via the kv_blockchain, which may result in linking
blocks to the main blockchain by removing them atomically from the st blockchain.

Changing the raw block CMF structs in order to be compatible with the kv_blockchain ```addBlock``` and enabling ability to ```move``` them.

not related - renaming of metrics.hpp -> s3_metrics.hpp in order to eliminate compilation error on macos filesystem.


kvbc/cmf/categorized_kvbc_msgs.cmf
```diff
@@ -73,28 +73,18 @@ Msg BlockData 2001 {
...
 Msg RawBlockData 2005 {
     fixedlist uint8 32 parent_digest
+    
     map string oneof{
-        RawBlockMerkleUpdates
-        RawBlockKeyValueUpdates
-        RawBlockImmutableUpdates
+        MerkleUpdatesData
+        KeyValueUpdatesData
+	ImmutableUpdatesData
```
dartdart26, line 83: Nit: Indent.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -240,4 +247,78 @@ ImmutableUpdatesInfo KeyValueBlockchain::handleCategoryUpdates(
...
+  for (auto i = block_id; i <= last_block_id.value(); ++i) {
+    auto raw_block = state_transfer_block_chain_.getRawBlock(i);
+    if (!raw_block) {
+      return;
+    }
+    writeSTLinkTransaction(block_id, *raw_block);
+  }
+  // Linking has fully completed and we should not have any more ST temporary blocks left. Therefore, make sure we don't
+  // have any value for the latest ST temporary block ID cache.
+  state_transfer_block_chain_.resetChain();
```
dartdart26, line 310: Great idea to introduce the ST chain as a separate object!

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -240,4 +247,78 @@ ImmutableUpdatesInfo KeyValueBlockchain::handleCategoryUpdates(
...
+}
+
+// Atmoic delete from state transfer and add to blockchain
```
dartdart26, line 313: Typo: Atomic.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -240,4 +247,78 @@ ImmutableUpdatesInfo KeyValueBlockchain::handleCategoryUpdates(
...
+void KeyValueBlockchain::writeSTLinkTransaction(const BlockId block_id, RawBlock& block) {
+  auto write_batch = native_client_->getBatch();
+  state_transfer_block_chain_.deleteBlock(block_id, write_batch);
+  auto new_block_id = addBlock(std::move(block.data.category_updates), write_batch);
+  native_client_->write(std::move(write_batch));
+
+  block_chain_.setAddedBlockId(new_block_id);
+  state_transfer_block_chain_.updateLastIdAfterDeletion(block_id);
```
dartdart26, line 321: I am not sure we need to call `updateLastIdAfterDeletion()` here as we don't change the last ID of the ST chain. Doing so works, but it does a load that in turn seeks and, if I am not missing something, is unnecessary. Please correct me if I am wrong.

kvbc/test/categorization/blockchain_test.cpp
```diff
@@ -0,0 +1,271 @@
...
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+/**
+ * The following test suite tests ordering of KeyValuePairs
```
dartdart26, line 15: Nit: I think that might be an incorrect comment.

kvbc/test/categorization/blockchain_test.cpp
```diff
@@ -0,0 +1,271 @@
...
+using namespace concord::kvbc;
+
+namespace {
+
+class categorized_kvbc : public ::testing::Test {
+  void SetUp() override {
+    cleanup();
+    db = TestRocksDb::createNative();
+    db->createColumnFamily(BLOCKS_CF);
+    db->createColumnFamily(ST_CHAIN_CF);
```
dartdart26, line 41: Maybe the KV blockchain should create the column families (if not existing) on construction?

kvbc/test/categorization/blockchain_test.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+  // delete non-existing block
+  {
+    auto wb = db->getBatch();
+    block_chain.deleteBlock(200, wb);
+    db->write(std::move(wb));
+    ASSERT_EQ(block_chain.loadGenesisBlockId().value(), 1);
+  }
+
+  // delete last and load
```
dartdart26, line 114: Nit: "delete the genesis and load"

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -229,6 +229,255 @@ TEST_F(categorized_kvbc, delete_block) {
...
+  {
+    categorization::RawBlock rb;
+    block_chain.addRawBlock(rb, 6);
+    ASSERT_TRUE(block_chain.getLastStatetransferBlockId().has_value());
+    ASSERT_EQ(block_chain.getLastStatetransferBlockId().value(), 6);
+
+    ASSERT_EQ(block_chain.getLastReachableBlockId(), 2);
+    ASSERT_EQ(block_chain_imp.loadLastReachableBlockId().value(), 2);
+  }
+  // Link the chains
```
dartdart26, line 466: Nice test! Do we need to also verify the contents (via `getBlock()`) of a raw block after it has been linked?

dartdart26 (approved):
Looks great, thanks @cloudnoize ! Very clear and nice separation!

The only thing that we should maybe fix is the removal of the seek when linking the two chains.

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: You're right, in this context is redundant, I'll remove it

(Though it won't actually do anything due to the condition at the start 
```if (!last_block_id_.has_value() || last_block_id_.value() != id) {
        return;
   }```)

kvbc/test/categorization/kv_blockchain_test.cpp
cloudnoize: yeah, but I'll do it after wiring the categories to actually produce the updates.

kvbc/test/categorization/blockchain_test.cpp
cloudnoize: Probably, the constructor now is trivial, I'll start handling it in the following PR I believe.

PR #1099, opened 2020-12-17
Add support for RocksDB PinnableSlice

dartdart26:


andrewjstone (approved):
Looks great. Thanks @dartdart26. Maybe we can add one more test that ensures we can call `get` and `getSlice` for the same key and it works. I can't imagine this won't work, but doesn't hurt, considering the build failed anyway and you have to repush :)

dartdart26:
> Looks great. Thanks @dartdart26. Maybe we can add one more test that ensures we can call `get` and `getSlice` for the same key and it works. I can't imagine this won't work, but doesn't hurt, considering the build failed anyway and you have to repush :)

Hmm, interesting :) Will do!

PR #1100, opened 2020-12-18
Added new View Change Trigger logic when View Change is initiated during a View Change process.

HristoStaykov:
Added quorum of ReplicaAsksToLeaveViewMsg-s in the View Change Message in order to make receiving Replica that has not received all complaints be able to move to higher View.
Added test to verify multiple addition of complaints to VC message.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -265,7 +265,7 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
 
 template <>
 void ReplicaImp::onMessage<ReplicaAsksToLeaveViewMsg>(ReplicaAsksToLeaveViewMsg *m) {
-  if (currentViewIsActive() && m->viewNumber() == getCurrentView()) {
+  if (m->viewNumber() == getCurrentView()) {
     LOG_INFO(VC_LOG,
              "Received ReplicaAsksToLeaveViewMsg " << KVLOG(m->viewNumber(), m->senderId(), m->idOfGeneratedReplica()));
     complainedReplicas.store(std::unique_ptr<ReplicaAsksToLeaveViewMsg>(m));
@@ -2048,9 +2048,9 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
   // msgSenderId's View is active and we have complaints for this View
   /////////////////////////////////////////////////////////////////////////
 
-  if (msg->currentViewIsActive()) {
+  if (msg->getViewNumber() == getCurrentView()) {
     for (const auto &i : complainedReplicas.getAllMsgs()) {
-      if (!msg->hasComplaintFromReplica(i.first) && msg->getViewNumber() == i.second->viewNumber()) {
+      if (!msg->hasComplaintFromReplica(i.first)) {
         sendAndIncrementMetric(i.second.get(), msgSenderId, metric_sent_replica_asks_to_leave_view_msg_due_to_status_);
       }
     }
@@ -2096,10 +2096,8 @@ void ReplicaImp::tryToSendStatusReport(bool onTimer) {
                        listOfMissingVCMsg,
                        listOfMissingPPMsg);
 
-  if (viewIsActive) {
-    for (const auto &i : complainedReplicas.getAllMsgs()) {
-      msg.setComplaintFromReplica(i.first);
-    }
+  for (const auto &i : complainedReplicas.getAllMsgs()) {
+    msg.setComplaintFromReplica(i.first);
   }
   if (listOfPPInActiveWindow) {
     const SeqNum start = lastStableSeqNum + 1;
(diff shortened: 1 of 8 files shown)


bftengine/src/bftengine/messages/ReplicaAsksToLeaveViewMsg.hpp
```diff
@@ -25,7 +25,7 @@ namespace impl {
 
 class ReplicaAsksToLeaveViewMsg : public MessageBase {
  public:
-  enum class Reason : uint8_t { ClientRequestTimeout };
+  enum class Reason : uint8_t { ClientRequestTimeout, PrimaryGetInChargeTimeout };
```
teoparvanov, line 28: NewPrimaryGetInChargeTimeout ?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2986,7 +3011,14 @@ void ReplicaImp::onViewsChangeTimer(Timers::Handle timer)  // TODO(GG): review/u
...
                    << KVLOG(curView, timeSinceLastAgreedViewMilli, timeSinceLastStateTransferMilli));
-      GotoNextView();
+
+      std::unique_ptr<ReplicaAsksToLeaveViewMsg> askToLeaveView(ReplicaAsksToLeaveViewMsg::create(
+          config_.getreplicaId(), curView, ReplicaAsksToLeaveViewMsg::Reason::PrimaryGetInChargeTimeout));
+      sendToAllOtherReplicas(askToLeaveView.get());
+      complainedReplicas.store(std::move(askToLeaveView));
+      metric_sent_replica_asks_to_leave_view_msg_.Get().Inc();
+
+      tryToGotoNextView();
```
teoparvanov, line 3021: At some point, shouldn’t we check that we have f+1 complaints before we move to the next view?

bftengine/src/bftengine/messages/MessageBase.cpp
```diff
@@ -65,6 +65,23 @@ void MessageBase::shrinkToFit() {
   storageSize_ = msgSize_;
 }
 
+bool MessageBase::reallocSize(uint32_t size) {
```
teoparvanov, line 68: Is this being used in the current PR? Maybe you can add some comment to clarify the purpose.

teoparvanov (commented):
LGTM overall, I just have several clarifying questions.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2048,9 +2048,9 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
   // msgSenderId's View is active and we have complaints for this View
   /////////////////////////////////////////////////////////////////////////
 
-  if (msg->currentViewIsActive()) {
```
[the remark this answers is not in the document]
HristoStaykov: Yes, I have forgotten to update now that we also use the new trigger during View Change itself this has to be corrected. thanks :)

bftengine/src/bftengine/messages/ReplicaAsksToLeaveViewMsg.hpp
HristoStaykov: I agree. This will better indicate that the expected New Primary is given a time period to take action.

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: tryToGotoNextView() does check if we have stored F+1 verified complains (ReplicaAsksToLeaveViewMsg-s) from different Replicas for the current View.
`void ReplicaImp::tryToGotoNextView() {
  if (complainedReplicas.hasQuorumToLeaveView()) {
    GotoNextView();
  } else {
    LOG_INFO(VC_LOG, "Insufficient quorum for moving to next view " << KVLOG(getCurrentView()));
  }
}
bool hasQuorumToLeaveView() const { return msgs.size() >= config_.fVal + 1U; }
`

teoparvanov: I see, thanks for the clarification!

bftengine/src/bftengine/messages/MessageBase.cpp
HristoStaykov: Yes, it is used. When we initiate View Change, we allocate a new ViewChangeMsg [ReplicaImp.cpp#L2260](https://github.com/vmware/concord-bft/blob/31ae2ee35413acf879ba10870137fe7dadf27bab/bftengine/src/bftengine/ReplicaImp.cpp#L2260) . If the next expected Primary does not take action during the View Change process, we need to initiate additional View Change to move to the next Primary. In this process we do not release the VeiwChangeMsg, but instead only modify it to reflect the New View [ReplicaImp.cpp#L2223](https://github.com/vmware/concord-bft/blob/31ae2ee35413acf879ba10870137fe7dadf27bab/bftengine/src/bftengine/ReplicaImp.cpp#L2223) but the information for the Sequence Numbers remains the same. We need to put the new complaints from peer Replicas, though, because they will be different from the initial ones. In order to do so we need to reallocate back to the max message size, clear the old complaints, add the new ones, and finally shrink the unused bytes and sign the new contents (without the complaints, only the View number combined with the SeqNo info).

teoparvanov (approved):
LGTM, thanks Hristo!

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2151,7 +2151,7 @@ void ReplicaImp::onMessage<ViewChangeMsg>(ViewChangeMsg *msg) {
   while (iter.getAndGoToNext(complaint, size)) {
     auto baseMsg = MessageBase(msg->senderId(), (MessageBase::Header *)complaint, size, true);
     auto *complaintMsg = new ReplicaAsksToLeaveViewMsg(&baseMsg);
-    LOG_INFO(GL,
+    LOG_INFO(VC_LOG,
              "Got complaint in ViewChangeMsg from " << KVLOG(msg->senderId(),
```
HristoStaykov: After discussion with @oded282 we found this should be in View Change logs.

PR #1101, opened 2020-12-20
Do not crash diagnostic server on failure

yontyon:
Now, when the diagnostic server is crashing, it also crashes the whole replica.
We would like not to crash the whole replica when the diagnosing server is not working

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -137,7 +137,7 @@ class Server {
         auto rv = select(listen_sock_ + 1, &read_fds, NULL, NULL, &tv);
         if (rv == 0) continue;  // timeout
         if (rv < 0 && errno == EINTR) {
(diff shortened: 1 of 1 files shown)


diagnostics/include/diagnostics_server.h
```diff
@@ -134,8 +136,16 @@ class Server {
         tv.tv_usec = 0;
         auto rv = select(listen_sock_ + 1, &read_fds, NULL, NULL, &tv);
         if (rv == 0) continue;  // timeout
-        if (rv < 0 && errno == EINTR) continue;
-        ConcordAssert(rv > 0);
+        if (rv < 0 && errno == EINTR) {
+          LOG_WARN(logger, "While waiting for a client requests, and interruption has occurred");
```
Liorgal28, line 140: NIT: Typo: and -> an

Liorgal28 (approved):

PR #1103, opened 2020-12-21
Enable clang tidy for header files

f-squirrel:
Clang-tidy does not analyze header files by default.
This PR enables the checks by setting the path to Concord-BFT headers.

Please notice that the PR does not add any additional checks but ensures that headers comply with existing ones.
The PR contains 4 commits:
* Enables the checks.
* All the rest represent fixes in a separate category.

Some of the warnings are suppressed, if you have a better idea of how to fix them in an elegant way, please feel free to comment.


storage/include/rocksdb/client.h
```diff
@@ -77,7 +77,7 @@ class Client : public concord::storage::IDBClient {
     if (txn_db_) {
       // If we're using a TransactionDB, it wraps the base DB, so release it
       // instead of releasing the base DB.
-      dbInstance_.release();
+      (void)dbInstance_.release();
```
dartdart26, line 80: What is the purpose of `(void)`?

dartdart26 (approved):
Great improvements, thanks @f-squirrel !

storage/include/rocksdb/client.h
f-squirrel: @dartdart26 , In general, a developer is not supposed to ignore the value returned from `unique_ptr::release`.
Since `clang-tidy` warns about this, I add `(void)` to suppress this warning and show that it is by design.

dartdart26: Got it, thanks!

PR #1104, opened 2020-12-22
Add a MultiGet operation to RocksDB NativeClient

andrewjstone:
This is a thin wrapper around the optimized overload of the rocksdb
implementation for a single column family. It allows multiple parallel
io requests in a single thread via the use of io_uring on Linux.

The wrapper has 2 primary goals:
 * Maximum efficiency with a realistic key interface.
 * Safety - We don't want to allow passing in contiguous containers that
 are too small to receieve output data.

This implementation is not intended to be used directly by user code,
and should be wrapped by the appopriate storage code. We have the option
to either copy strings and return them from the PinnableSlices, or
return a wrapper type that allows zero copy. We could also return the
raw RocksDB type, but so far we have not made those public outside
storage layer code.

### DIFF
--- a/storage/include/rocksdb/native_client.h
+++ b/storage/include/rocksdb/native_client.h
@@ -101,6 +101,23 @@ class NativeClient : public std::enable_shared_from_this<NativeClient> {
   NativeWriteBatch getBatch() const;
   void write(NativeWriteBatch &&);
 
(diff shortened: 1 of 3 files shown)


dartdart26 (approved):
Looks great, thanks @andrewjstone ! Good interface!

PR #1105, opened 2020-12-22
ECS S3 doesn't support batch operations there's a problem storing met…

toly-kournik:
…adata as an atomic transaction which may lead to inconsistencies in State Transfer metadata if RO Replica stops ungracefully.

Since Reserved Pages are not currently used by a RO Replica for Object Store Replication, State Transfer persistence is only a optimization and can be omitted.
The only drawback can be a repeaded transfer of at most 150 blocks in case of such a crash, which we can tolerate.


teoparvanov (approved):
LGTM, thanks Toly!

PR #1106, opened 2020-12-22
Implement common category read iface for Immutable

dartdart26:
Implement the common category read interface for the
ImmutableKeyValueCategory. That includes:
 * get the value of a key at a specific version
 * get the value of a key at the latest version
 * get multiple keys at specific versions
 * get multiple keys at their latest versions
 * get the latest version of a key
 * get the latest versions of multiple keys

Introduce value types for the different categories.

Add an ability to partially deserialize immutable values by expoiting
CMF's canonical format. Use the `ImmutableDbVersion` CMF message that
contains the version only and deserialize it from a `ImmutableDbValue`
buffer that contains both version and value. That way we don't have to
copy the value when we only care about the version.

### DIFF
--- a/kvbc/include/categorization/immutable_kv_category.h
+++ b/kvbc/include/categorization/immutable_kv_category.h
@@ -18,11 +18,16 @@
 
 #include "base_types.h"
 #include "categorized_kvbc_msgs.cmf.hpp"
-#include "details.h"
+
+#include <rocksdb/slice.h>
+#include <rocksdb/status.h>
 
 #include <memory>
-#include <set>
+#include <optional>
 #include <string>
+#include <utility>
+#include <variant>
+#include <vector>
 
 namespace concord::kvbc::categorization::detail {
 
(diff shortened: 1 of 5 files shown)


kvbc/src/categorization/immutable_kv_category.cpp
```diff
@@ -120,14 +131,100 @@ void ImmutableKeyValueCategory::deleteBlock(const ImmutableUpdatesInfo &updates_
...
+std::optional<Value> ImmutableKeyValueCategory::getLatest(const std::string &key) const {
+  const auto ser = db_->getSlice(cf_, key);
   if (!ser) {
     return std::nullopt;
   }
-  auto value = ImmutableValue{};
-  deserialize(*ser, value);
-  return value;
+  return value(*ser);
+}
```
cloudnoize, line 151: very nit and personal opinion: I started disliking using `*` for optional may cause confusion when reading and might be more error-prone I think

kvbc/src/categorization/immutable_kv_category.cpp
```diff
@@ -120,14 +131,100 @@ void ImmutableKeyValueCategory::deleteBlock(const ImmutableUpdatesInfo &updates_
...
+      if (v.block_id == version) {
+        values.push_back(v);
+      } else {
+        values.push_back(std::nullopt);
+      }
+    } else if (status.IsNotFound()) {
+      values.push_back(std::nullopt);
+    } else {
+      throw std::runtime_error{"ImmutableKeyValueCategory multiGet() failure: " + status.ToString()};
+    }
```
cloudnoize, line 178: nit : handling non-ok fist and reduce nesting seems clearer to me
```
if(status.IsNotFound()){
     values.push_back(std::nullopt);
     continue;
}

if(!status.ok()){
    throw std::runtime_error{"ImmutableKeyValueCategory multiGet() failure: " + status.ToString()};
}

const auto v = value(slice);
if (v.block_id != version) {
  values.push_back(std::nullopt);
  continue;
} 

values.push_back(v);
}

File: kvbc/src/categorization/immutable_kv_category.cpp [hunk 3]
```diff
@@ -120,14 +131,100 @@ void ImmutableKeyValueCategory::deleteBlock(const ImmutableUpdatesInfo &updates_
...
+  for (auto i = 0ull; i < slices.size(); ++i) {
+    const auto &status = statuses[i];
+    const auto &slice = slices[i];
+    if (status.ok()) {
+      values.push_back(value(slice));
+    } else if (status.IsNotFound()) {
+      values.push_back(std::nullopt);
+    } else {
+      throw std::runtime_error{"ImmutableKeyValueCategory multiGetLatest() failure: " + status.ToString()};
+    }
```
cloudnoize, line 198: same as above

kvbc/src/categorization/immutable_kv_category.cpp
```diff
@@ -120,14 +131,100 @@ void ImmutableKeyValueCategory::deleteBlock(const ImmutableUpdatesInfo &updates_
...
+  versions.clear();
+  for (auto i = 0ull; i < slices.size(); ++i) {
+    const auto &status = statuses[i];
+    const auto &slice = slices[i];
+    if (status.ok()) {
+      versions.push_back(version(slice));
+    } else if (status.IsNotFound()) {
+      versions.push_back(std::nullopt);
+    } else {
+      throw std::runtime_error{"ImmutableKeyValueCategory multiGetLatestVersion() failure: " + status.ToString()};
```
cloudnoize, line 225: same

cloudnoize (approved):
Looks great @dartdart26 !!

kvbc/src/categorization/immutable_kv_category.cpp
dartdart26: I use it explicitly here, because I first check if there is a value in `ser`. Then, using `*` doesn't do any checks at runtime. In contrast, `value()` does one more check and throws if value is not there. I want to avoid the second check.

kvbc/src/categorization/immutable_kv_category.cpp
dartdart26: Not sure if that matters for modern compilers and CPUs, but I think the `ok()` status is most common and we'd like to check for it first. I also wanted to avoid explicit `continue`. But I agree that nesting makes it a bit more complex. I will leave as is for now, if you are OK.

kvbc/src/categorization/immutable_kv_category.cpp
andrewjstone: I did not know that! Thanks for the info @dartdart26  I avoided it in the past for clarity reasons similar to @cloudnoize. But if we can avoid unnecessary checks I think it makes sense to use it.

kvbc/src/categorization/immutable_kv_category.cpp
andrewjstone: I also dislike using continue, as it makes the control flow non-linear and harder to understand IMO.

kvbc/include/categorization/immutable_kv_category.h
```diff
@@ -37,39 +42,65 @@ namespace concord::kvbc::categorization::detail {
...
+  // Return std::nullopt if `key` doesn't exist in `block_id`.
+  std::optional<Value> get(const std::string &key, BlockId block_id) const;
 
   // Get the value of an immutable key.
   // Return std::nullopt if the key doesn't exist.
-  std::optional<ImmutableValue> get(const std::string &key) const;
+  std::optional<Value> getLatest(const std::string &key) const;
+
+  // Get values for keys at specific versions.
+  // `keys` and `versions` must of the same size.
```
andrewjstone, line 70: typo: s/of/be/

kvbc/src/categorization/immutable_kv_category.cpp
```diff
@@ -120,14 +131,100 @@ void ImmutableKeyValueCategory::deleteBlock(const ImmutableUpdatesInfo &updates_
...
 }
 
-std::optional<ImmutableValue> ImmutableKeyValueCategory::get(const std::string &key) const {
-  const auto ser = db_->get(cf_, key);
+std::optional<Value> ImmutableKeyValueCategory::get(const std::string &key, BlockId block_id) const {
+  auto val = getLatest(key);
+  if (!val) {
+    return std::nullopt;
+  }
+  if (asImmutable(val).block_id != block_id) {
```
andrewjstone, line 139: When would this occur? Can this happen outside of a bug or db corruption? Should we at least log it?

andrewjstone (approved):
Looks great @dartdart26 ! I only have one question. If what you are doing is safe, feel free to merge.

kvbc/src/categorization/immutable_kv_category.cpp
cloudnoize: yeah, didn't know that. cool

kvbc/src/categorization/immutable_kv_category.cpp
cloudnoize: sure, just a nit offer

kvbc/include/categorization/immutable_kv_category.h
dartdart26: Will fix, thanks.

kvbc/src/categorization/immutable_kv_category.cpp
dartdart26: This is just to conform to a more generic interface that gets the key at a specific block version. However, since immutable category keys can only have one version, if that version is not the requested one, then we report this key is not found.

andrewjstone: Makes sense. Thanks!

andrewjstone:
Merging so I can use it :)

andrewjstone:
> Merging so I can use it :)

Whoops. Should've squashed. Oh well.

PR #1107, opened 2020-12-22
Add num of RO replicas to the preprocessor

Liorgal28:
When initiating Preprocessor we initiate an array of ongoing requests.
This array is being initiated according to the client id’s, but we detected a bug since we didn’t calculate the RO replicas.
Since the RO replicas ID’s are right after the regular replicas and before the clients we need to calculate them also.
I’ve added a member variable for the number of RO replicas and use this variable in all the places that calculate/uses the number of replicas.


HristoStaykov (approved):
LGTM! Thanks @Liorgal28 !

teoparvanov (approved):
LGTM!

PR #1108, opened 2020-12-22
Add histograms for checkpoint and state transfer duration

yontyon:
Add multiple histograms:
1. To measure the time that passes between a creation of a checkpoint to the time it became stable
2. To measure the time that state transfer takes from starting to its end

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1683,6 +1683,7 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
 
   if (askForStateTransfer && !stateTransfer->isCollectingState()) {
     LOG_INFO(GL, "Call to startCollectingState()");
+    time_in_state_transfer_.start();
     clientsManager->clearAllPendingRequests();  // to avoid entering a new view on old request timeout
     stateTransfer->startCollectingState();
   } else if (msgSeqNum > lastStableSeqNum + kWorkWindowSize) {
@@ -2508,7 +2509,7 @@ void ReplicaImp::sendCheckpointIfNeeded() {
 
 void ReplicaImp::onTransferringCompleteImp(uint64_t newStateCheckpoint) {
   TimeRecorder scoped_timer(*histograms_.onTransferringCompleteImp);
-
+  time_in_state_transfer_.end();
   LOG_INFO(GL, KVLOG(newStateCheckpoint));
 
   if (ps_) {
@@ -2617,6 +2618,7 @@ void ReplicaImp::onSeqNumIsStable(SeqNum newStableSeqNum, bool hasStateInformati
   ConcordAssertEQ(newStableSeqNum % checkpointWindowSize, 0);
 
   if (newStableSeqNum <= lastStableSeqNum) return;
+  checkpoint_times_.end(newStableSeqNum);
   TimeRecorder scoped_timer(*histograms_.onSeqNumIsStable);
 
   LOG_INFO(GL,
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -463,7 +463,8 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
                                         {"onSeqNumIsStable", onSeqNumIsStable},
                                         {"onTransferringCompleteImp", onTransferringCompleteImp},
                                         {"consensus", consensus},
-                                        {"timeInActiveView", timeInActiveView}});
+                                        {"timeInActiveView", timeInActiveView},
+                                        {"checkpointFromCreationToStable", checkpoint_creation_to_stable}});
```
andrewjstone, line 467: It looks like you missed adding `timeInStateTransfer` here.

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -490,16 +491,23 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
 
     std::shared_ptr<Recorder> timeInActiveView =
         std::make_shared<Recorder>(1, MAX_VALUE_SECONDS, 3, concord::diagnostics::Unit::SECONDS);
+
+    std::shared_ptr<Recorder> timeInStateTransfer =
```
andrewjstone, line 495: I'm not sure 5 minutes is a long enough duration here. Maybe make it a few hours for max? I wouldn't change the constant above though, as making the duration longer makes the histograms larger and for most histograms we don't need hours.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3775,6 +3779,7 @@ void ReplicaImp::executeRequestsInPrePrepareMsg(concordUtils::SpanWrapper &paren
   if ((lastExecutedSeqNum + 1) % checkpointWindowSize == 0) {
     checkpointNum = (lastExecutedSeqNum + 1) / checkpointWindowSize;
     stateTransfer->createCheckpointOfCurrentState(checkpointNum);
+    checkpoint_times_.start(checkpointNum);
```
andrewjstone, line 3782: It appears that you are using checkpoint numbers here which are sequence numbers divided by window size. However, in the end method you are using sequence numbers directly. These will not match up.

andrewjstone (changes requested):
Hi @yontyon thanks for the additions. Unfortunately there are a few errors here that will prevent this code from working as intended. After fixing I suggest you run a cluster in a docker environment, induce state transfer via stopping and starting a node, and then sanity check with `./concord-ctl perf get`.

bftengine/src/bftengine/ReplicaImp.hpp
yontyon: MAX_VALUE_SECONDS is set to 5 years, Am I missing something?

bftengine/src/bftengine/ReplicaImp.hpp
yontyon: fixed

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: fixed

bftengine/src/bftengine/ReplicaImp.hpp
andrewjstone: Whoops. I misread. Sorry.

andrewjstone (approved):
Looks good @yontyon. Thanks for the fixes and the new histograms! 

I only had one minor comment about using declarations. We aren't supposed to have them in header files unless encapsulated in some manner so they don't infect users.

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
```diff
@@ -27,6 +27,9 @@
 #include "Logger.hpp"
 #include "kvstream.h"
 #include "demangle.hpp"
+#include "diagnostics.h"
+
+using Recorder = concord::diagnostics::Recorder;
```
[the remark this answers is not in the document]
yontyon: done

yontyon:
After some sanity checks, it looks like the digital signatures operation histograms are not accurate.
Thus, we reset the PR to have only checkpoint and state transfer metrics.
The DS histograms will be added in a different PR

PR #1111, opened 2020-12-23
Implement read interface and infra for category type mapping

cloudnoize:
- Prepare the infrastructure for the read interface, which will be wired when all categories will be implemented.
- Implement the mapping of categories to their corresponding type, and use it on Blockchain construction when loading the categories to memory.
-  Adding logger
- Optimization: creation of a raw block on add block, in order to use it for the next add block digest computation.
- ```RawBlock``` and ```Updates``` share the same CMF struct for seamless conversions.

### DIFF
--- a/kvbc/include/categorization/column_families.h
+++ b/kvbc/include/categorization/column_families.h
@@ -22,4 +22,6 @@ inline const auto IMMUTABLE_KV_CF_SUFFIX = std::string{"_immutable"};
 inline const auto BLOCKS_CF = std::string{"blocks"};
 inline const auto ST_CHAIN_CF = std::string{"st_chain"};
 
+inline const auto CAT_ID_TYPE_CF = std::string{"cat_id_type"};
+
 }  // namespace concord::kvbc::categorization::detail
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -18,16 +18,21 @@
 #include <memory>
 #include "blocks.h"
 #include "blockchain.h"
+#include "immutable_kv_category.h"
 
 #include "kv_types.hpp"
 
 namespace concord::kvbc::categorization {
 
-class KeyValueBlockchain {
- public:
-  KeyValueBlockchain(const std::shared_ptr<concord::storage::rocksdb::NativeClient>& native_client)
-      : native_client_{native_client}, block_chain_{native_client_}, state_transfer_block_chain_{native_client_} {}
+// Temp forward declearations
+struct MerkleCategory {};
+struct KVHashCategory {};
 
+class KeyValueBlockchain {
+  using VersionedRawBlock = std::pair<BlockId, std::optional<categorization::RawBlock>>;
+
+ public:
+  KeyValueBlockchain(const std::shared_ptr<concord::storage::rocksdb::NativeClient>& native_client, bool link_st_chain);
   /////////////////////// Add Block ///////////////////////
 
   BlockId addBlock(Updates&& updates);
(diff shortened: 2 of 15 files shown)


kvbc/include/categorization/column_families.h
```diff
@@ -22,4 +22,6 @@ inline const auto IMMUTABLE_KV_CF_SUFFIX = std::string{"_immutable"};
 inline const auto BLOCKS_CF = std::string{"blocks"};
 inline const auto ST_CHAIN_CF = std::string{"st_chain"};
 
+inline const auto CAT_ID_TYPE = std::string{"cat_id_type"};
```
dartdart26, line 25: Nit: Maybe can add `_CF` to designate it is a column family.

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -478,6 +476,159 @@ TEST_F(categorized_kvbc, link_state_transfer_chain) {
   }
 }
 
+TEST_F(categorized_kvbc, creation_of_category_type_cf) {
+  KeyValueBlockchain block_chain{db, true};
+  ASSERT_TRUE(db->hasColumnFamily(detail::ST_CHAIN_CF));
```
dartdart26, line 481: Should we check for the `CAT_ID_TYPE` instead?

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -478,6 +476,159 @@ TEST_F(categorized_kvbc, link_state_transfer_chain) {
...
+}
+
+TEST_F(categorized_kvbc, instantiation_of_categories) {
```
dartdart26, line 484: Nice test!

kvbc/include/categorization/kv_blockchain.h
```diff
@@ -90,25 +133,56 @@ class KeyValueBlockchain {
...
   detail::Blockchain block_chain_;
   detail::Blockchain::StateTransfer state_transfer_block_chain_;
+
+  // Holds the last raw block of the last corresponding block that was added.
+  // Used to save construction of a raw block per addBlock method.
+  // E.L - compare this with getRawBlock to see they are equal
+  VersionedRawBlock last_raw_block_;
+
+ public:
+  struct KeyValueBlockchain_tester {
```
dartdart26, line 170: Nit: Maybe we can make the members we need for testing `public` and just use them. Rationale is that we are in the `detail` namespace anyway, i.e. we care about encapsulation a bit less. Just a suggestion, no need to change.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -12,9 +12,57 @@
...
+  // is getValue() that returns keys from the blockchain only and ignores keys in the temporary state
+  // transfer chain.
+  linkSTChainFrom(getLastReachableBlockId() + 1);
+}
+
+void KeyValueBlockchain::instantiateCategories() {
+  auto itr = native_client_->getIterator(detail::CAT_ID_TYPE);
+  itr.first();
+  while (itr) {
+    auto cat_type = static_cast<detail::CATEGORY_TYPE>(itr.valueView()[0]);
```
dartdart26, line 41: Maybe we can assert that the value size is 1.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -234,17 +441,25 @@ KeyValueUpdatesInfo KeyValueBlockchain::handleCategoryUpdates(
...
-  for (auto& [k, v] : updates.kv) {
-    (void)v;
-    skvui.tagged_keys[k] = std::move(v.tags);
+ImmutableUpdatesInfo KeyValueBlockchain::handleCategoryUpdates(BlockId block_id,
+                                                               const std::string& category_id,
+                                                               ImmutableUpdatesData&& updates,
+                                                               concord::storage::rocksdb::NativeWriteBatch& write_batch,
+                                                               categorization::RawBlock& raw_block) {
+  // if true means that category is new and we should create an instance of it.
+  if (insertCategoryMapping(category_id, detail::CATEGORY_TYPE::immutable, write_batch)) {
```
dartdart26, line 450: Is it possible to first lookup in `categories_`, in order to avoid the DB lookup? Once we instantiate categories in the constructor, I suppose we can assume that the memory map is in sync with the DB and only update them both when a new category appears. Please let me know if I miss something.

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -478,6 +476,159 @@ TEST_F(categorized_kvbc, link_state_transfer_chain) {
...
+
+  ASSERT_EQ(cats.count("kvhash"), 1);
+  ASSERT_THROW(std::get<MerkleCategory>(cats["kvhash"]), std::bad_variant_access);
+  ASSERT_NO_THROW(std::get<KVHashCategory>(cats["kvhash"]));
+}
+
+TEST_F(categorized_kvbc, throw_on_non_exist_category_type) {
+  KeyValueBlockchain block_chain{db, true};
+  auto wb = db->getBatch();
+  wb.put(detail::CAT_ID_TYPE, std::string("merkle"), std::string(1, static_cast<char>(4)));
```
dartdart26, line 517: Maybe we can change the hardcoded value of 4 to some special enum value (i.e. last + 1, invalid cat ID or something like that) in order make it more explicit and reduce the risk of this test failing at some point.

kvbc/test/categorization/kv_blockchain_test.cpp
cloudnoize: copy paste err probably :)

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: good catch, missed it, thanks!

kvbc/test/categorization/kv_blockchain_test.cpp
cloudnoize: true

kvbc/src/categorization/blockchain.cpp
```diff
@@ -12,18 +12,24 @@
 // file.
 
 #include "categorization/blockchain.h"
+#include "Logger.hpp"
 
 namespace concord::kvbc::categorization::detail {
 
 Blockchain::Blockchain(const std::shared_ptr<concord::storage::rocksdb::NativeClient>& native_client)
     : native_client_{native_client} {
+  if (detail::createColumnFamilyIfNotExisting(detail::BLOCKS_CF, *native_client_.get())) {
```
dartdart26, line 21: Nit: Can simplify by doing `*native_client` only.

kvbc/src/categorization/kv_blockchain.cpp
dartdart26: Thanks! I guess if we assume that we are creating categories rarely, we can even first look up in `categories_` (and not in `categories_types_`) and expect that most of the time that will be enough to know it is an existing category. Doing so can avoid the 2 lookups we need now (both in `categories_` and `categories_types_`).

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -478,6 +476,170 @@ TEST_F(categorized_kvbc, link_state_transfer_chain) {
...
+  ASSERT_THROW(block_chain.getBlockData(889), std::runtime_error);
+  ASSERT_THROW(block_chain.getBlockData(887), std::runtime_error);
+  auto updates = block_chain.getBlockData(888);
+  auto variant = updates.kv[cf];
+  auto merkle_updates = std::get<MerkleUpdatesData>(variant);
+  // check reconstruction of original kv
+  ASSERT_EQ(merkle_updates.kv[key], value);
+}
+
+TEST_F(categorized_kvbc, validate_category_creation_on_add) {
```
dartdart26, line 600: Maybe we can add a test that updates keys for category `C` as one type and then, in the next block, tries to do the same, but for another type. It might be that is covered by the `instantiation_of_categories` test and is not needed. Please let me know if you think it makes sense.

dartdart26 (approved):
Great job @cloudnoize, looks great, thanks!

PR #1114, opened 2020-12-24
Add context to vc logs

oded282:
In this PR I added context to View Change logs for the usage of the causal order parser. 

- Added the view number to VC logs.

- Added a log that implies that the second phase of VC ended.


HristoStaykov (approved):
LGTM! Thanks @oded282 !

PR #1115, opened 2020-12-30
Do not crash diagnostic server on failure (#1101)

yontyon:
Now, when the diagnostic server is crashing, it also crashes the whole replica.
We would like not to crash the whole replica when the diagnosing server is not working


Liorgal28 (approved):
LGTM!

PR #1117, opened 2021-01-05
Storage Categorization: Add BlockMerkle category

andrewjstone:
The block merkle category operates by only storing a single key in the
sparse merkle tree per block. The key is the block id, and the value is
the hash of the concatenation of the hashes of all the keys and values
(including deleted keys), that were updated in the merkle tree for the
given block.

Direct lookup is enabled by versioned key lookup on hashed values.
Values can be looked up directly if the block id is known for the value,
or the latest value can be retrieved. MultiGet is also supported. In
order to avoid the computational overhead of RocksDB seeks, point
lookups of the latest value are done by maintaining an index for the
latst block id of each key in a separate column family.

Unit tests were added that mainly verify the behavior required by direct
gets and puts.

A follow up commit will add support for deletes and pruning behavior.
Unit tests will be added for this behavior, and to cover the block
merkle nodes and hashes appropriately.

Lastly, category type suffixes were changed from *UpdatesData* and
*UpdatesInfo* to *Input* and *Output*, respectively.

### DIFF
--- a/kvbc/cmf/categorized_kvbc_msgs.cmf
+++ b/kvbc/cmf/categorized_kvbc_msgs.cmf
@@ -1,6 +1,6 @@
-# Updates
+# Input for category `add`
 
-Msg MerkleUpdatesData 1 {
+Msg BlockMerkleInput 1 {
     map string string kv
     list string deletes
 }
@@ -11,7 +11,7 @@ Msg ValueWithExpirationData 2 {
     bool stale_on_update
 }
 
-Msg KeyValueUpdatesData 3 {
+Msg KeyValueInput 3 {
     map string ValueWithExpirationData kv
     list string deletes
     bool calculate_root_hash
@@ -22,27 +22,27 @@ Msg ImmutableValueUpdate 4 {
     list string tags
 }
 
-Msg ImmutableUpdatesData 5 {
+Msg ImmutableInput 5 {
     # key -> value, [tag1, tag2...]
     map string ImmutableValueUpdate kv
     bool calculate_root_hash
 }
 
-Msg CategoryUpdatesData 6 {
+Msg CategoryInput 6 {
     map string oneof{
-        MerkleUpdatesData
-        KeyValueUpdatesData
-        ImmutableUpdatesData
+        BlockMerkleInput
+        KeyValueInput
+        ImmutableInput
     } kv
 }
 
-# Updates info
+# Output for category `add`
 
 Msg MerkleKeyFlag 1000 {
     bool deleted
 }
 
-Msg MerkleUpdatesInfo 1001 {
+Msg BlockMerkleOutput 1001 {
     map string MerkleKeyFlag keys
     fixedlist uint8 32 root_hash
     uint64 state_root_version
@@ -54,12 +54,12 @@ Msg KVKeyFlag 1002 {
     bool stale_on_update
 }
 
-Msg KeyValueUpdatesInfo 1003 {
+Msg KeyValueOutput 1003 {
     map string KVKeyFlag keys
     optional fixedlist uint8 32 root_hash
 }
 
-Msg ImmutableUpdatesInfo 1004 {
+Msg ImmutableOutput 1004 {
     # key -> [tag1, tag2...]
     map string list string tagged_keys
     optional map string fixedlist uint8 32 tag_root_hashes
@@ -75,16 +75,16 @@ Msg BlockData 2001 {
     uint64 block_id
     fixedlist uint8 32 parent_digest
     map string oneof{
-        MerkleUpdatesInfo
-        KeyValueUpdatesInfo
-        ImmutableUpdatesInfo
+        BlockMerkleOutput
+        KeyValueOutput
+        ImmutableOutput
     } categories_updates_info
 }
 
 
 Msg RawBlockData 2005 {
     fixedlist uint8 32 parent_digest
-    CategoryUpdatesData updates
+    CategoryInput updates
     map string optional fixedlist uint8 32 category_root_hash
 }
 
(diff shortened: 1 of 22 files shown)


kvbc/include/sparse_merkle/base_types.h
```diff
@@ -329,6 +329,9 @@ class NibblePath {
   // Return the internal representation of the path.
   const std::vector<uint8_t>& data() const { return path_; }
 
+  // Allow moving the data out of the path.
+  std::vector<uint8_t> move_data() { return path_; }
```
dartdart26, line 333: Do we need `std::move()` here?

kvbc/include/sparse_merkle/tree.h
```diff
@@ -40,6 +42,7 @@ namespace sparse_merkle {
 // can be written to the DB atomically.
 class Tree {
  public:
+  Tree() {}
```
dartdart26, line 45: Nit: = default?

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -0,0 +1,398 @@
...
+  auto hashed_keys = hashedKeys(keys);
+  multiGetLatestVersion(hashed_keys, versions);
+}
+
+void BlockMerkleCategory::multiGetLatestVersion(const std::vector<Hash>& hashed_keys,
+                                                std::vector<std::optional<BlockId>>& versions) const {
+  auto slices = std::vector<::rocksdb::PinnableSlice>{};
+  slices.reserve(hashed_keys.size());
+  auto statuses = std::vector<::rocksdb::Status>{};
+  statuses.reserve(hashed_keys.size());
```
dartdart26, line 270: Nit: I think the NativeClient will resize the slices and statuses internally, so these might be redundant.

dartdart26 (approved):

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    // Put the latest version of the key
+    batch.put(BLOCK_MERKLE_LATEST_KEY_VERSION, key_it->value, serialize(LatestKeyVersion{block_id}));
+
+    kv_it++;
+  }
+
+  for (auto key_it = hashed_deleted_keys.begin(); key_it != hashed_deleted_keys.end(); key_it++) {
+    // Add the new tombstone to the key versions. We use 0 here to avoid the serialization overhead of a
+    // variant/optional.
```
dartdart26, line 353: I was thinking whether deleting a key at some block should treat it as a new key version, but without a value. On the other hand, though, the API would look a bit weird in that `getLatest()` will return nullopt, whereas `getLatestVersion()` will return some version. That's what I did in the versioned kv category, but I think let's decide what is better and unify them.

What do you think @andrewjstone @cloudnoize ?

kvbc/cmf/categorized_kvbc_msgs.cmf
```diff
@@ -116,3 +116,55 @@ Msg ImmutableDbValue 4002 {
...
+    # Contains a 1 where a child node is present, 0 otherwise
+    uint32 bitmask
+    list BatchedInternalNodeChild children
+}
+
+# An index to the latest version of a key.
+#
+# A sentinel of 0 is used to mark the key as deleted. We don't need tombstones as we never rewrite
+# the same key after it has been deleted. The use of a sentinel is to optimize the size and
+# serialization performance of this index.
```
dartdart26, line 166: Does that imply that once a key is deleted, the user can no longer set a new value for it in a future block?

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -0,0 +1,398 @@
...
+  }
+}
+
+void BlockMerkleCategory::putMerkleNodes(NativeWriteBatch& batch,
+                                         sparse_merkle::UpdateBatch&& update_batch,
+                                         uint64_t tree_version) {
+  for (const auto& [leaf_key, leaf_val] : update_batch.leaf_nodes) {
+    auto ser_key = serialize(leafKeyToVersionedKey(leaf_key));
+    batch.put(BLOCK_MERKLE_LEAF_NODES_CF, ser_key, leaf_val.value.string_view());
+  }
```
cloudnoize, line 364: If I understand correctly we should get only one leaf in ```update_batch.leaf_nodes``` if so does it worth to assert?

kvbc/src/categorization/block_merkle_category.cpp
cloudnoize: I think that intuitively once a key is deleted we should return a nullopt on both cases.
And will allow redefining the key as a new key if it will be added in the future. 
If we need the information when the key was deleted maybe it's worth adding ```deletedAt``` method.
wdyt @dartdart26

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -0,0 +1,398 @@
...
+      ++value_index;
+    } else {
+      values.push_back(std::nullopt);
+    }
+  }
+  ConcordAssertEQ(values.size(), keys.size());
+}
+
+void BlockMerkleCategory::putKeys(NativeWriteBatch& batch,
+                                  uint64_t block_id,
```
cloudnoize, line 336: No treatment to the stale updates since deletion is not in place yet?

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -0,0 +1,398 @@
...
+  return output;
+}
+
+std::optional<MerkleValue> BlockMerkleCategory::get(const std::string& key, BlockId block_id) const {
+  auto hasher = Hasher{};
+  auto hashed_key = hasher.digest(key.data(), key.size());
+  return get(hashed_key, block_id);
+}
+
+std::optional<MerkleValue> BlockMerkleCategory::get(const Hash& hashed_key, BlockId block_id) const {
```
dartdart26, line 186: Would that method return a value even if it has been deleted at `block_id`? I guess it is a tradeoff as you've pointed out, i.e. using a variant (with an overhead) to have a tombstone in the values table. Please let me know if I miss something.

kvbc/src/categorization/block_merkle_category.cpp
dartdart26: Yes, that's a good point.

I was also thinking of a case that key K is set some value at block 1 and then the key is deleted at block 5. I assume we need to give some sort of an indication to users that the key was deleted in block 5.

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: Ah you are right. Will remove to get rid of the clutter. Thanks!

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: I don't believe we have to store them any more with changes to pruning related to block merkle. I need to update our doc. This will also be a good opportunity to open source the document describing block merkle and deletes. Will work on that today along with the delete code.

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: Sure. It's at least good documentation at a minimum.

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: It cannot have been deleted **at** `block_id` if the key exists, but only earlier or later. Since the user is asking specifically for the value at `block_id` we return it if one exists.

kvbc/cmf/categorized_kvbc_msgs.cmf
andrewjstone: I actually don't believe that comment is correct. I don't see any problem with rewriting a key after its been deleted. I will change it.

kvbc/include/sparse_merkle/base_types.h
andrewjstone: lol. yes.

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: If a key has been deleted, `getLatest` should always return `std::nullopt`. However, if a user asks for a key from a prior version, we return it if it exists. They likely want to see the key as they know it exists at a given block version and want to see the value for that block. The user can tell if it has been deleted by asking for the latest version if they care. `getLatestVersion` will return `nullopt` also in the case that the key has been deleted. It replaces the 0 tombstone for the user reply. 

We cannot implement a `deletedAt` method without using a std::variant to encode a tombstone. I originally had that, but it adds overhead to every latest key lookup for minimum gain. I also don't see a use case for it at the moment. I think we should only implement such a perf impacting change when we have a compelling use case.

andrewjstone: I updated the code in https://github.com/vmware/concord-bft/pull/1120 so that we store a tombstone with a single bit.

PR #1119, opened 2021-01-05
Check the complaints placed inside the View Change message for higher…

HristoStaykov:
… View than the Replica's current.

This handles the case where we have a Replica behind more than 1 View from the others.
This is an improvement, the behind Replica can catch-up by the F+1 View Change messages that it will
receive from peers. We can though move to the agreed view earlier based on the F+1 signed Complaints.


teoparvanov (approved):
LGTM!

PR #1120, opened 2021-01-06
BlockMerkle category improvements

andrewjstone:
Indexes for the latest version of keys now utilize the high order bit to
indicate whether the last write was a delete. This translates to an
API change that returns a `TaggedVersion` for latest version retrievals.
This is useful for pre-execution conflict detection.

Other fixes include:
 * Properly move the internal data from a NibblePath
 * Reuse the hash function in details.h rather than calling
 Hasher.digest directly.
 * Not writing added and deleted keys to the merkle value on initial
 block creation.
 * Using the root hash of the block tree rather than just the root hash
 of the block for output to category `add`.
 * Removing unnecessary reserve calls to vectors for multiGet calls.
 * Adding a test to ensure the new key delete/tombstone behavior works

### DIFF
--- a/kvbc/include/categorization/details.h
+++ b/kvbc/include/categorization/details.h
@@ -27,6 +27,31 @@ namespace concord::kvbc::categorization::detail {
 
 using Buffer = std::vector<std::uint8_t>;
 
+struct TaggedVersion {
+  // The high bit contains a flag indicating whether the key was deleted or not.
+  TaggedVersion(uint64_t masked_version) {
+    deleted = 0x80000000 & masked_version;
+    version = 0x7FFFFFFF & masked_version;
+  }
+
+  TaggedVersion(bool deleted, BlockId version) : deleted(deleted), version(version) {}
+
+  // Return a BlockId that sets the high bit to 1 if the version is deleted
+  BlockId encode() const {
+    if (deleted) {
+      return version | 0x80000000;
+    }
+    return version;
+  }
+
+  bool deleted;
+  BlockId version;
+};
+
+inline bool operator==(const TaggedVersion &lhs, const TaggedVersion &rhs) {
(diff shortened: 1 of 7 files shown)


dartdart26 (approved):
Looks great, thanks for the idea and for addressing the comments from before!

@cloudnoize and I think that it is worthwhile to make all categories return `TaggedVersion` instead of just the version and even expose it to users. Since the test failed, I am thinking of cloning this branch and:

- move TaggedVersion to base_types.h (make it public and not part of the detail namespace) and use it on other categories
- fix the test so that we can merge

kvbc/include/categorization/details.h
```diff
@@ -27,6 +27,31 @@ namespace concord::kvbc::categorization::detail {
 
 using Buffer = std::vector<std::uint8_t>;
 
+struct TaggedVersion {
+  // The high bit contains a flag indicating whether the key was deleted or not.
+  TaggedVersion(uint64_t masked_version) {
+    deleted = 0x80000000 & masked_version;
+    version = 0x7FFFFFFF & masked_version;
```
cloudnoize, line 34: The masks are 32 bits :) 
I think we just need to use ```TaggedVersion``` as CMF message without using the upper bit as flag.

dartdart26: Great catch, thanks @cloudnoize , I've changed to 64 bits in my commit!

In terms of TaggedVersion, as discussed offline, having 1 byte more is not optimal, but it doesn't encode any logic in the message itself and other tools need not be aware of it. Since this byte will be mostly zeroes, maybe it won't have a big impact on storage. Just some thoughts.

dartdart26:
Opened another PR that adds a second commit fixing minor issues: https://github.com/vmware/concord-bft/pull/1122

PR #1121, opened 2021-01-06
Block accumulation memory leak fix

Liorgal28:
This MR fixing 3 issues:

1. Memory leak - I allocated memory for responses for each request but forgot to release the memory once I copied it to the reply message.
2. RO replica id's - inserted RO replica id when calculating the clients set inside the constructor of the replica imp.
3. Parameters sizing - fixed simple client parameters sizing we got from the configuration manager as uint16_t so we need to adhere to that size.


yontyon (approved):
LGTM!

tdimitrov (approved):

HristoStaykov (approved):

PR #1123, opened 2021-01-06
Support TaggedVersion in ImmutableKeyValueCategory

dartdart26:
Add support for TaggedVersion in ImmutableKeyValueCategory in order to
unify the interface across different categories. Always return
`deleted = false` for keys in the ImmutableKeyValueCategory.


cloudnoize (approved):
LGTM

PR #1124, opened 2021-01-06
Calculate client set id's with RO replicas

Liorgal28:
This PR is fixing 2 issues:

RO replica id's - inserted RO replica id when calculating the clients set inside the constructor of the replica imp.
Parameters sizing - fixed simple client parameters sizing we got from the configuration manager as uint16_t so we need to adhere to that size.


yontyon (approved):
The memory leak problem is not presented in 0.9 and 1.1.
make sure it is OK to merge this changes

PR #1125, opened 2021-01-07
Support TaggedVersion in VersionedKeyValueCategory

dartdart26:
Unify categories by supporting TaggedVersion in
VersionedKeyValueCategory.

Optimize tombstone values by not using a variant type. Instead, use a
simple flag to designate a value is a tombstone. The value of a
tombstone is the empty buffer. Rationale is that the variant adds an
overhead of 4 bytes for its type.

Handle deleted keys in deleteLastReachableBlock() by preserving the
deleted flag when seeking back to a previously deleted key version.

Optimize multiGetLatest() by not including deleted (or not found) keys
in the values column family lookup. Implementation is borrowed from the
BlockMerkleCategory.

Add unit tests to verify the new behavior.

### DIFF
--- a/kvbc/cmf/categorized_kvbc_msgs.cmf
+++ b/kvbc/cmf/categorized_kvbc_msgs.cmf
@@ -119,13 +119,23 @@ Msg VersionedRawKey 4003 {
     uint64 version
 }
 
-Msg ImmutableDbValue 4004 {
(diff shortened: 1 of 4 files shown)


cloudnoize (approved):
LGTM!!

PR #1126, opened 2021-01-07
Introduce a categorized KVBC interface

dartdart26:
Expose KVBC key categories as an interface that application layers can
use. Split into 3 parts:
 * IBlockAdder - allows adding a block from given categorized updates
 * IReader - allows reading keys and blocks
 * IBlocksDeleter - allows deleting the genesis block. Used for pruning.
   Unchanged.

The categorized DB adapter can implement these interfaces. The Replica
can contain an instance of the DB adapter and have methods to get each
interface and pass it around to command handlers and other parts of
the system that require KVBC storage access.

### DIFF
--- a/kvbc/include/db_interfaces.h
+++ b/kvbc/include/db_interfaces.h
@@ -6,6 +6,13 @@
 #include "kv_types.hpp"
 #include "status.hpp"
 
+#include "categorization/base_types.h"
+#include "categorization/updates.h"
+
+#include <optional>
+#include <string>
+#include <vector>
+
 namespace concord::kvbc {
 /**
  *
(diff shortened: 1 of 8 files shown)


kvbc/include/db_interfaces.h
```diff
@@ -35,13 +42,85 @@ class ILocalKeyValueStorageReadOnly {
...
+  virtual void multiGetLatestVersion(const std::string &category_id,
+                                     const std::vector<std::string> &keys,
+                                     std::vector<std::optional<categorization::TaggedVersion>> &versions) const;
+
+  // Get the value of `key` and a proof for it at `block_id`.
+  // Return std::nullopt if the key doesn't exist.
+  virtual std::optional<categorization::Proof> getProof(const std::string &category_id,
+                                                        const std::string &key,
+                                                        BlockId block_id) const = 0;
+
```
cloudnoize, line 111: Maybe this should be defined not as part of the read interface? maybe IProof ?

kvbc/include/db_interfaces.h
```diff
@@ -35,13 +42,85 @@ class ILocalKeyValueStorageReadOnly {
...
+  // Return std::nullopt if the key doesn't exist or is deleted.
+  virtual std::optional<categorization::TaggedVersion> getLatestVersion(const std::string &category_id,
+                                                                        const std::string &key) const = 0;
+
+  // Get the latest versions of the given keys in `category_id`.
+  // If a key is missing, then std::nullopt is returned for its version.
+  virtual void multiGetLatestVersion(const std::string &category_id,
+                                     const std::vector<std::string> &keys,
+                                     std::vector<std::optional<categorization::TaggedVersion>> &versions) const;
+
```
yuliasherman, line 105: Looks good to me as far as we know where to take category_id from :-).

kvbc/include/db_interfaces.h
dartdart26: I don't have a strong preference. My assumption was that users of read will also be the ones that want proofs.

kvbc/include/db_interfaces.h
dartdart26: It is application-specific and can be a mapping from key type to a category ID.

kvbc/include/db_interfaces.h
cloudnoize: I'm in favor of small interfaces that are easy to implement, moreover it proof is not really a read operation, it actually calculates stuff.

dartdart26: Actually, I am thinking of removing the proof part for now, because there are some differences (there are tags for the immutable category, possibly differences in the block merkle proof) and maybe this is not thought through yet. Will just leave the read part and handle proofs separately.

dartdart26: Removed proofs altogether.

andrewjstone (approved):
Looks good @dartdart26

robem (approved):
Lgtm

kvbc/include/db_interfaces.h
robem: @dartdart26 If `category_id` is application-specific then who is going to create those IDs and how are they generated?

dartdart26: Yep :)

cloudnoize (approved):
LGTM!

kvbc/include/db_interfaces.h
teoparvanov: Hi @dartdart26, can this one day be used, so that we store different types of Ethereum K/Vs in separate Merkle trees, each with its own root hash?

dartdart26: @teoparvanov Yes, you can have multiple categories of the same type by just using different IDs.

PR #1127, opened 2021-01-07
Fix metrics preProcInFlyRequests num

yontyon:
It turns out that a non-primary replica may update this metric twice - once if it gets a client request, and when it gets the preProcess request from primary.

We fix it such that each replica will update the metric only once


yuliasherman (approved):
LGTM

PR #1128, opened 2021-01-07
Wire categories to the addBlock operation

cloudnoize:
Wire the ```ImmutableKeyValueCategory, VersionedKeyValueCategory, BlockMerkleCategory``` 
add(BlockId block_id, category update, storage::rocksdb::NativeWriteBatch&) method to the adapter upon
adding a new block.

Now that those methods actually write to storage, the  ```raw block``` can be reconstructed
from a block, by reconstructing the updates from the block struct.

Changing the ```BlockMerkleCategory``` to return a ```Value``` variant instead of the category specific 
```MerkleValue```.

### DIFF
--- a/kvbc/include/categorization/block_merkle_category.h
+++ b/kvbc/include/categorization/block_merkle_category.h
@@ -41,12 +41,12 @@ class BlockMerkleCategory {
 
   // Return the value of `key` at `block_id`.
   // Return std::nullopt if the key doesn't exist at `block_id`.
-  std::optional<MerkleValue> get(const std::string& key, BlockId block_id) const;
-  std::optional<MerkleValue> get(const Hash& hashed_key, BlockId block_id) const;
+  std::optional<Value> get(const std::string& key, BlockId block_id) const;
+  std::optional<Value> get(const Hash& hashed_key, BlockId block_id) const;
 
   // Return the value of `key` at its most recent block version.
   // Return std::nullopt if the key doesn't exist.
-  std::optional<MerkleValue> getLatest(const std::string& key) const;
+  std::optional<Value> getLatest(const std::string& key) const;
 
   // Returns the latest *block* version of a key.
   // Returns std::nullopt if the key doesn't exist.
@@ -58,11 +58,11 @@ class BlockMerkleCategory {
   // If a key is missing at the specified version, std::nullopt is returned for it.
   void multiGet(const std::vector<std::string>& keys,
                 const std::vector<BlockId>& versions,
-                std::vector<std::optional<MerkleValue>>& values) const;
+                std::vector<std::optional<Value>>& values) const;
 
   // Get the latest values of a list of keys.
   // If a key is missing, std::nullopt is returned for it.
-  void multiGetLatest(const std::vector<std::string>& keys, std::vector<std::optional<MerkleValue>>& values) const;
+  void multiGetLatest(const std::vector<std::string>& keys, std::vector<std::optional<Value>>& values) const;
 
   // Get the latest versions of the given keys.
   // If a key is missing, std::nullopt is returned for its version.
@@ -73,7 +73,7 @@ class BlockMerkleCategory {
  private:
   void multiGet(const std::vector<Buffer>& versioned_keys,
                 const std::vector<BlockId>& versions,
-                std::vector<std::optional<MerkleValue>>& values) const;
+                std::vector<std::optional<Value>>& values) const;
 
   void putKeys(storage::rocksdb::NativeWriteBatch& batch,
                uint64_t block_id,
@@ -110,6 +110,9 @@ class BlockMerkleCategory {
   logging::Logger logger_;
   sparse_merkle::Tree tree_;
 };
+
+inline const MerkleValue& asMerkle(const Value& v) { return std::get<MerkleValue>(v); }
+inline MerkleValue& asMerkle(Value& v) { return std::get<MerkleValue>(v); }
 
 }  // namespace concord::kvbc::categorization::detail
 
--- a/kvbc/include/categorization/details.h
+++ b/kvbc/include/categorization/details.h
@@ -27,7 +27,7 @@ namespace concord::kvbc::categorization::detail {
 
 using Buffer = std::vector<std::uint8_t>;
 
-enum class CATEGORY_TYPE : char { merkle = 0, immutable = 1, kv_hash = 2, end_of_types };
+enum class CATEGORY_TYPE : char { block_merkle = 0, immutable = 1, versioned_kv = 2, end_of_types };
 
 template <typename Span>
 Hash hash(const Span &span) {
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -19,17 +19,16 @@
 #include "blocks.h"
 #include "blockchain.h"
 #include "immutable_kv_category.h"
-
+#include "block_merkle_category.h"
+#include "versioned_kv_category.h"
 #include "kv_types.hpp"
+#include "categorization/types.h"
+#include "thread_pool.hpp"
 
 namespace concord::kvbc::categorization {
 
-// Temp forward declearations
-struct BlockMerkleCategory {};
-struct KVHashCategory {};
-
 class KeyValueBlockchain {
-  using VersionedRawBlock = std::pair<BlockId, std::optional<categorization::RawBlock>>;
+  using VersionedRawBlock = std::pair<BlockId, std::optional<categorization::RawBlockData>>;
 
  public:
   KeyValueBlockchain(const std::shared_ptr<concord::storage::rocksdb::NativeClient>& native_client, bool link_st_chain);
@@ -94,8 +93,8 @@ class KeyValueBlockchain {
                              const detail::CATEGORY_TYPE type,
                              concord::storage::rocksdb::NativeWriteBatch& write_batch);
 
-  const std::variant<detail::ImmutableKeyValueCategory, BlockMerkleCategory, KVHashCategory>& getCategory(
-      const std::string& cat_id) const;
+  const std::variant<detail::ImmutableKeyValueCategory, detail::BlockMerkleCategory, detail::VersionedKeyValueCategory>&
+  getCategory(const std::string& cat_id) const;
 
   /////////////////////// deletes ///////////////////////
 
@@ -139,25 +138,21 @@ class KeyValueBlockchain {
   BlockMerkleOutput handleCategoryUpdates(BlockId block_id,
                                           const std::string& category_id,
                                           BlockMerkleInput&& updates,
-                                          concord::storage::rocksdb::NativeWriteBatch& write_batch,
-                                          categorization::RawBlock& raw_block);
+                                          concord::storage::rocksdb::NativeWriteBatch& write_batch);
 
   VersionedOutput handleCategoryUpdates(BlockId block_id,
                                         const std::string& category_id,
                                         VersionedInput&& updates,
-                                        concord::storage::rocksdb::NativeWriteBatch& write_batch,
-                                        categorization::RawBlock& raw_block);
+                                        concord::storage::rocksdb::NativeWriteBatch& write_batch);
   ImmutableOutput handleCategoryUpdates(BlockId block_id,
                                         const std::string& category_id,
                                         ImmutableInput&& updates,
-                                        concord::storage::rocksdb::NativeWriteBatch& write_batch,
-                                        categorization::RawBlock& raw_block);
+                                        concord::storage::rocksdb::NativeWriteBatch& write_batch);
 
   /////////////////////// Members ///////////////////////
 
   std::shared_ptr<concord::storage::rocksdb::NativeClient> native_client_;
-  std::map<std::string, std::variant<detail::ImmutableKeyValueCategory, BlockMerkleCategory, KVHashCategory>>
-      categorires_;
+  CategoriesMap categorires_;
   std::map<std::string, detail::CATEGORY_TYPE> categorires_types_;
   detail::Blockchain block_chain_;
   detail::Blockchain::StateTransfer state_transfer_block_chain_;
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: kvbc/include/categorization/kv_blockchain.h [hunk 1]
```diff
@@ -170,19 +169,24 @@ class KeyValueBlockchain {
  public:
   struct KeyValueBlockchain_tester {
     void instantiateCategories(KeyValueBlockchain& kvbc) { kvbc.instantiateCategories(); }
-    const std::map<std::string, std::variant<detail::ImmutableKeyValueCategory, BlockMerkleCategory, KVHashCategory>>&
+    const std::map<std::string,
```
dartdart26, line 172: Nit: Can use `auto` to save some typing.

kvbc/src/categorization/blocks.cpp
```diff
@@ -41,31 +43,21 @@ BlockMerkleInput RawBlock::getUpdates(const std::string& category_id,
                                       const BlockMerkleOutput& update_info,
                                       const BlockId& block_id,
                                       const std::shared_ptr<storage::rocksdb::NativeClient>& native_client) {
```
dartdart26, line 45: I was wondering if it makes sense to make the categories (and the native client for that matter) RawBlock members. That way we can avoid ref count updates in the categories and simplify the interface a bit.

kvbc/src/categorization/blocks.cpp
```diff
@@ -41,31 +43,21 @@ BlockMerkleInput RawBlock::getUpdates(const std::string& category_id,
...
     }
-    // E.L see how we can optimize the sliver temporary allocation
-    auto db_key =
-        v2MerkleTree::detail::DBKeyManipulator::genDataDbKey(Key(std::string(key)), update_info.state_root_version);
-    auto val = native_client->get(category_id, db_key);
+    // get value of the key for a version from storage via the category
+    auto val = cat.get(key, block_id);
     if (!val.has_value()) {
-      throw std::runtime_error("Couldn't find value for key");
+      auto msg = std::string("Couldn't find value for key [") + key + std::string("] (Merkle category)");
```
dartdart26, line 56: I think this is an assert situation.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -103,20 +102,27 @@ BlockId KeyValueBlockchain::addBlock(CategoryInput&& category_updates,
   new_block.data.parent_digest = parent_digest_future.get();
   last_raw_block.data.parent_digest = new_block.data.parent_digest;
   block_chain_.addBlock(new_block, write_batch);
+  LOG_DEBUG(CAT_BLOCK_LOG, "Writing block [" << new_block.id() << "] to the blocks cf");
   write_batch.put(detail::BLOCKS_CF, Block::generateKey(new_block.id()), Block::serialize(new_block));
   return new_block.id();
 }
 
 std::future<BlockDigest> KeyValueBlockchain::computeParentBlockDigest(const BlockId block_id,
```
dartdart26, line 110: I think now's a good time to unify and change the parent block digest calculation from CryptoPP SHA2 to OpenSSL SHA3 (using our EVP wrapper).

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -411,38 +416,42 @@ BlockMerkleOutput KeyValueBlockchain::handleCategoryUpdates(BlockId block_id,
...
-  for (auto& [k, v] : updates.kv) {
-    (void)v;
-    out.keys[k] = VersionedKeyFlags{false, v.stale_on_update};
+  // if true means that category is new and we should create an instance of it.
+  if (insertCategoryMapping(category_id, detail::CATEGORY_TYPE::versioned_kv, write_batch)) {
+    if (const auto [_, inserted] =
+            categorires_.try_emplace(category_id, detail::VersionedKeyValueCategory{category_id, native_client_});
+        !inserted) {
+      (void)_;
+      throw std::runtime_error{"Category already exists = " + category_id};
```
dartdart26, line 445: I think we should assert in this case as this should never happen. Please correct me if I am wrong.

kvbc/test/categorization/block_merkle_category_unit_test.cpp
```diff
@@ -97,12 +97,12 @@ TEST_F(block_merkle_category, put_and_get) {
   auto expected = MerkleValue{{block_id, val1}};
 
   // Get by key works
-  ASSERT_EQ(expected, cat.getLatest(key1).value());
+  ASSERT_EQ(expected, asMerkle(cat.getLatest(key1).value()));
   // Get by specific block works
-  ASSERT_EQ(expected, cat.get(key1, 1));
+  ASSERT_EQ(expected, asMerkle(cat.get(key1, 1)));
```
dartdart26, line 102: I know it is not related to this particular PR, but calling `asMerkle(cat.get())` has a different effect to calling `asMerkle(cat.get().value())` if the optional is empty as the first one will be undefined behavior and the second one will throw an exception. We might miss some issues or have the test fail for strange reasons if we use the first format.

kvbc/test/categorization/blocks_test.cpp
```diff
@@ -71,40 +71,65 @@ TEST_F(categorized_kvbc, serialization_and_desirialization_of_block) {
...
-  ASSERT_EQ(db_get_val.value(), db_val);
-
-  categorization::RawBlock rw(block, db);
-  ASSERT_EQ(rw.data.parent_digest, block.data.parent_digest);
-  auto variant = rw.data.updates.kv[cf];
-  auto merkle_updates = std::get<BlockMerkleInput>(variant);
-  // check reconstruction of original kv
-  ASSERT_EQ(merkle_updates.kv[key], value);
+    detail::Buffer in{block1_db_val.value().begin(), block1_db_val.value().end()};
+    auto block1_from_db = Block::deserialize(in);
```
dartdart26, line 104: Nit: I think `detail::deserialize(Span, T)` can be used for simplicity.

kvbc/test/categorization/blocks_test.cpp
```diff
@@ -133,6 +158,183 @@ TEST_F(categorized_kvbc, fail_reconstruct_merkle_updates) {
...
+    auto variant2 = rw.data.updates.kv["imm2"];
+    auto imm_updates2 = std::get<ImmutableInput>(variant2);
+    // check reconstruction of original kv
+    ASSERT_EQ(imm_updates2.kv.size(), 1);
+    auto ex2 = ImmutableValueUpdate{"imm_value32", {"32"}};
+    ASSERT_EQ(imm_updates2.kv["imm_key12"], ex2);
+  }
+}
+
+TEST_F(categorized_kvbc, fail_reconstruct_immutabele_updates) {
```
dartdart26, line 233: Nit: Typo, immutabele -> immutable.

kvbc/test/categorization/blocks_test.cpp
```diff
@@ -133,6 +158,183 @@ TEST_F(categorized_kvbc, fail_reconstruct_merkle_updates) {
...
+  detail::Buffer in{block1_db_val.value().begin(), block1_db_val.value().end()};
+  auto block1_from_db = Block::deserialize(in);
+
+  detail::ImmutableKeyValueCategory cat{"imm", db};
+  auto wb = db->getBatch();
+  auto out = std::get<ImmutableOutput>(block1_from_db.data.categories_updates_info["imm"]);
+  cat.deleteBlock(out, wb);
+  db->write(std::move(wb));
+
+  ASSERT_THROW(categorization::RawBlock rw(block1_from_db, db), std::runtime_error);
```
dartdart26, line 259: Can this ever happen?

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -491,29 +553,30 @@ TEST_F(categorized_kvbc, instantiation_of_categories) {
       detail::CAT_ID_TYPE_CF, std::string("merkle2"), std::string(1, static_cast<char>(detail::CATEGORY_TYPE::merkle)));
   wb.put(
       detail::CAT_ID_TYPE_CF, std::string("imm"), std::string(1, static_cast<char>(detail::CATEGORY_TYPE::immutable)));
-  wb.put(
-      detail::CAT_ID_TYPE_CF, std::string("kvhash"), std::string(1, static_cast<char>(detail::CATEGORY_TYPE::kv_hash)));
+  wb.put(detail::CAT_ID_TYPE_CF,
+         std::string("kvhash"),
```
dartdart26, line 557: Nit: versioned instead of kvhash?

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -545,57 +608,57 @@ TEST_F(categorized_kvbc, get_category) {
...
-  auto variant = updates.kv[cf];
-  auto merkle_updates = std::get<BlockMerkleInput>(variant);
+  auto updates = block_chain.getBlockData(1);
+  auto variant = updates.kv["versioned"];
+  auto versioned_updates = std::get<VersionedInput>(variant);
   // check reconstruction of original kv
-  ASSERT_EQ(merkle_updates.kv[key], value);
+  ASSERT_TRUE(versioned_updates.calculate_root_hash);
+  auto expected = ValueWithFlags{"ver_val2", true};
+  ASSERT_EQ(versioned_updates.kv["ver_key2"], expected);
```
dartdart26, line 661: Nit: I think using the generated `operator==()` for the VersionedInput  CMF type can simplify checks here.

dartdart26 (approved):
Looks great, thanks @cloudnoize ! I have put some comments that you might choose to address or not.

kvbc/include/categorization/block_merkle_category.h
```diff
@@ -111,6 +111,13 @@ class BlockMerkleCategory {
   sparse_merkle::Tree tree_;
 };
 
+inline const MerkleValue& asMerkle(const Value& v) { return std::get<MerkleValue>(v); }
+inline MerkleValue& asMerkle(Value& v) { return std::get<MerkleValue>(v); }
+
+// Optional overloads assume the optional contains a value.
+inline const MerkleValue& asMerkle(const std::optional<Value>& v) { return asMerkle(*v); }
```
andrewjstone, line 118: As @dartdart26 pointed out in another comment, this is incredibly unsafe. IMO We shouldn't wrap unrestricted calls that assume an optional contains a value. The use of the `*` operator is best left to only being used immediately after a check that the optional contains a value, in inline code. Otherwise, any modification of the code can break the implicit assumption since the user may not look up the `asMerkle` function and notice that it uses `*`. 

I also personally, am not a fan of wrapping `std::get`, but at least that code will throw an exception and not lead to silent corruption.

kvbc/include/categorization/details.h
```diff
@@ -27,7 +27,7 @@ namespace concord::kvbc::categorization::detail {
 
 using Buffer = std::vector<std::uint8_t>;
 
-enum class CATEGORY_TYPE : char { merkle = 0, immutable = 1, kv_hash = 2, end_of_types };
+enum class CATEGORY_TYPE : char { merkle = 0, immutable = 1, versioned_kv = 2, end_of_types };
```
andrewjstone, line 30: Nit: We may want to call this block_merkle, to leave open the option for other merkle tree types.

kvbc/src/categorization/blocks.cpp
andrewjstone: Good catch. I agree. A key must exist if its in an existing block.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -91,7 +91,6 @@ BlockId KeyValueBlockchain::addBlock(CategoryInput&& category_updates,
   last_raw_block.data.updates = category_updates;
   // Per category updates
   for (auto&& [category_id, update] : category_updates.kv) {
-    // https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding
     std::visit(
         [&new_block, category_id = category_id, &write_batch, &last_raw_block, this](auto& update) {
```
andrewjstone, line 95: I know this isn't part of this PR, but I just noticed it. Shouldn't the `update` parameter be taken by universal reference (`&&`) to signal a move. 

Similarly, the `update` argument should be wrapped in `std::move`, since it's an l-value reference.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -103,20 +102,27 @@ BlockId KeyValueBlockchain::addBlock(CategoryInput&& category_updates,
...
+  // cached raw block is unusable, get the raw block from storage
+  else if (block_id > INITIAL_GENESIS_BLOCK_ID) {
     cached_raw_block.second = getRawBlock(parent_block_id);
-  } else {
+  }
+  // it's the first block, we don't have a parent
+  else {
     cached_raw_block.second.reset();
   }
+  // pass the versioned raw block bu value (thread safe), for the async operation to calculate its digest
```
andrewjstone, line 125: typo: `bu` -> `by`

andrewjstone (changes requested):
Looks really good @cloudnoize. I think the few issues that were raised are significant enough to fix before merge though. Thanks!

kvbc/src/categorization/kv_blockchain.cpp
dartdart26: We might also use ThreadPool::async instead of std::async in order to avoid thread creation on every block.

kvbc/include/categorization/kv_blockchain.h
cloudnoize: I like it :)

kvbc/test/categorization/blocks_test.cpp
cloudnoize: I hope not, but the test can be seen as to whether the raw block asserts when it fails to get any key that is present at the block.

kvbc/test/categorization/block_merkle_category_unit_test.cpp
cloudnoize: I copy-pasted it from the other categories.
I removed the functions the accepts optional, but for the other categories, I guess it should be done as well.

kvbc/include/categorization/block_merkle_category.h
cloudnoize: I copy-pasted it from the other categories.
I removed the functions the accept optional, but for the other categories, I guess it should be done as well.

kvbc/include/categorization/details.h
cloudnoize: good point

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: wasn't aware of the ThreadPool class. great job btw @dartdart26

cloudnoize: Regarding the digest, I'll do it as part of the next PR since I have some question on it

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: I looked at assertions and exceptions as having the same behavior eventually.
I'll change it.

kvbc/test/categorization/kv_blockchain_test.cpp
cloudnoize: I think that until we add default initializers the generated operators are worthless

kvbc/src/categorization/blocks.cpp
cloudnoize: I took your advice but didn't make it as a member in order to keep the stateless property and avoid coping.
Instead, I pass the categories as const pointer (since I need also a default constructor).
Please review it

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: done

PR #1129, opened 2021-01-07
Diagnostics: Prevent out of bound memory access

andrewjstone:
It is not safe to take the address of a `std::array` using `&` and then
do pointer arithmetic with that address. The problem is that the
calculations may end out of bounds because the pointer appears to advances by a
pointer size and not a multiple of the size of the template
argument of the array. This is probably undefined behavior since
std::array doesn't provide a `&` operator.

The proper way to do this is to use the `.data()`
member function to get a properly typed address.


dartdart26 (approved):
Great catch!

PR #1133, opened 2021-01-08
Update log levels of several messages

nkgeorgiev:
This is done so we can more easily track request executions


teoparvanov (approved):
LGTM!

PR #1135, opened 2021-01-08
Update log levels of several messages

nkgeorgiev:
This is done so we can more easily track request executions

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1025,7 +1025,7 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
   SCOPED_MDC_SEQ_NUM(std::to_string(msg->seqNumber()));
   SCOPED_MDC_PATH(CommitPathToMDCString(CommitPath::OPTIMISTIC_FAST));
 
-  LOG_DEBUG(CNSUS, "Reached consensus, Received FullCommitProofMsg message");
+  LOG_INFO(CNSUS, "Consensus reached, Received FullCommitProofMsg message");
 
   if (relevantMsgForActiveView(msg)) {
     SeqNumInfo &seqNumInfo = mainLog->get(msgSeqNum);
@@ -1336,7 +1336,7 @@ void ReplicaImp::onMessage<CommitFullMsg>(CommitFullMsg *msg) {
   if (relevantMsgForActiveView(msg)) {
     sendAckIfNeeded(msg, msgSender, msgSeqNum);
 
-    LOG_DEBUG(GL, "Received CommitFullMsg" << KVLOG(msgSender));
+    LOG_INFO(GL, "Consensus reached, Received CommitFullMsg" << KVLOG(msgSender));
 
     SeqNumInfo &seqNumInfo = mainLog->get(msgSeqNum);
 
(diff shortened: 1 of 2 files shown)


teoparvanov (approved):
LGTM, thanks!

andrewjstone (commented):
This is going to generate a *lot* of extra log messages which are going to cause log rotations to happen much faster, and push out any other data that may help us track down bugs. Are we really sure we want to do this in production, considering we can change log levels dynamically as needed for CNSUS only?

teoparvanov:
> This is going to generate a _lot_ of extra log messages which are going to cause log rotations to happen much faster, and push out any other data that may help us track down bugs. Are we really sure we want to do this in production, considering we can change log levels dynamically as needed for CNSUS only?

Hi Andrew, I don’t actually think those would spam the logs too badly. Consensus is reached once per batch, not even per individual request. Am I missing something?

andrewjstone:
> > This is going to generate a _lot_ of extra log messages which are going to cause log rotations to happen much faster, and push out any other data that may help us track down bugs. Are we really sure we want to do this in production, considering we can change log levels dynamically as needed for CNSUS only?
> 
> Hi Andrew, I don’t actually think those would spam the logs too badly. Consensus is reached once per batch, not even per individual request. Am I missing something?
> 
> PS: notice how we have an info log per preprocessed request at INFO level and it isn’t causing any issues.

Even with batching, at high rates we are going to see several log messages per second. You can go ahead and merge this in, I'm just cautioning that it may limit our visibility. I don't think logging should be used for ordinary operations, and so I'd say the preprocessor ones are also even more spammy and should be turned off when possible. If we have no other way currently to record the information you are looking for this is fine, but why not just add a histogram to track the latencies if that's what's needed?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3828,10 +3828,10 @@ void ReplicaImp::executeRequestsInPrePrepareMsg(concordUtils::SpanWrapper &paren
     auto dur = controller->durationSincePrePrepare(lastExecutedSeqNum + 1);
     if (dur > 0) {
       // Primary
-      LOG_DEBUG(CNSUS, "Consensus reached, duration [" << dur << "ms]");
+      LOG_INFO(CNSUS, "Consensus reached, duration [" << dur << "ms]");
```
teoparvanov, line 3831: We probably don’t need this - we already know consensus has been reached.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3828,10 +3828,10 @@ void ReplicaImp::executeRequestsInPrePrepareMsg(concordUtils::SpanWrapper &paren
...
 
     } else {
-      LOG_DEBUG(CNSUS, "Consensus reached");
+      LOG_INFO(CNSUS, "Consensus reached");
```
teoparvanov, line 3834: Same here, let’s restore to DEBUG level.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1025,7 +1025,7 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
   SCOPED_MDC_SEQ_NUM(std::to_string(msg->seqNumber()));
   SCOPED_MDC_PATH(CommitPathToMDCString(CommitPath::OPTIMISTIC_FAST));
 
-  LOG_DEBUG(CNSUS, "Reached consensus, Received FullCommitProofMsg message");
+  LOG_INFO(CNSUS, "Consensus reached, Received FullCommitProofMsg message");
```
teoparvanov, line 1028: Consensus reached (fast path)

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1336,7 +1336,7 @@ void ReplicaImp::onMessage<CommitFullMsg>(CommitFullMsg *msg) {
   if (relevantMsgForActiveView(msg)) {
     sendAckIfNeeded(msg, msgSender, msgSeqNum);
 
-    LOG_DEBUG(GL, "Received CommitFullMsg" << KVLOG(msgSender));
+    LOG_INFO(GL, "Consensus reached, Received CommitFullMsg" << KVLOG(msgSender));
```
teoparvanov, line 1339: Consensus reached (slow path)

teoparvanov:
On second thought, I actually agree with @andrewjstone about the noise this change is going to create in the logs.

I suggest to close this PR and reassess on Monday. We also have to take into account the upcoming block accumulation feature, which will impact the way we do the E2E correlation.

PR #1136, opened 2021-01-08
Initial implementation of a Byzantine replica

sergiosagu:
The idea of this change is to propose an initial implementation of a Byzantine replica, or in
other words, an extension of a regular replica where it's possible to inject different
Byzantine behavior by extending the default logic.

What this change proposes tries to achieve is basically:
- Create an extension of bftengine::ReplicaImp to expose some interesting methods that can be
overridden in order to inject some Byzantine behavior.
- Create an extension of kvbc::ReplicaImp to override some methods in order to instantiate the
new ByzantineBftEngineReplicaImp.
- Add an initial test case that makes use of the Byzantine classes.

See the README file and the comments inside the ByzantineReplicaImp.cpp for some more details.

Small disclaimer: I'm not sure whether this set of changes could or should be merged "as it is"
or not, probably not :) But with the proper feedback, addressing the comments and perhaps
splitting the changes in different reviews if needed, it could be a good enough starting point.

### DIFF
--- a/kvbc/include/ReplicaImp.h
+++ b/kvbc/include/ReplicaImp.h
@@ -126,22 +126,26 @@ class ReplicaImp : public IReplica,
     }
   };
 
- private:
+  /**
+   * (Byzantine replica work) The following attributes shouldn't be "relaxed" from private to protected, if there was a
+   * small refactoring to expose them to potential subclasses.
+   */
+ protected:
   logging::Logger logger;
   RepStatus m_currentRepStatus;
-
-  std::unique_ptr<IDbAdapter> m_bcDbAdapter;
-  std::shared_ptr<storage::IDBClient> m_metadataDBClient;
   bft::communication::ICommunication *m_ptrComm = nullptr;
   const bftEngine::ReplicaConfig &replicaConfig_;
   bftEngine::IReplica::IReplicaPtr m_replicaPtr = nullptr;
-  ICommandsHandler *m_cmdHandler = nullptr;
   bftEngine::IStateTransfer *m_stateTransfer = nullptr;
-  concord::storage::DBMetadataStorage *m_metadataStorage = nullptr;
-  std::unique_ptr<ReplicaStateSync> replicaStateSync_;
   std::shared_ptr<concordMetrics::Aggregator> aggregator_;
   std::shared_ptr<bftEngine::ControlStateManager> controlStateManager_;
+  ICommandsHandler *m_cmdHandler = nullptr;
+  concord::storage::DBMetadataStorage *m_metadataStorage = nullptr;
+  std::unique_ptr<IDbAdapter> m_bcDbAdapter;
+  std::shared_ptr<storage::IDBClient> m_metadataDBClient;
+  std::unique_ptr<ReplicaStateSync> replicaStateSync_;
 
+ private:
   // 5 Minutes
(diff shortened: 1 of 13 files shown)


tests/simpleKVBC/ByzantineReplica/ByzantineReplicaImp.hpp
```diff
@@ -0,0 +1,64 @@
...
+  ByzantineBftEngineReplicaImp(const LoadedReplicaData &ld,
+                               IRequestsHandler *requestsHandler,
+                               IStateTransfer *stateTrans,
+                               shared_ptr<MsgsCommunicator> msgsCommunicator,
+                               shared_ptr<PersistentStorage> persistentStorage,
+                               shared_ptr<MsgHandlersRegistrator> msgHandlers,
+                               concordUtil::Timers &timers);
+};
+
+class ByzantineKvbcReplicaImp : public concord::kvbc::ReplicaImp {
```
teoparvanov, line 52: Can’t we avoid having a dedicated `ByzantineKvbcReplicaImp` by injecting a pointer to a `ByzantineBftEngineReplicaImp`?

kvbc/include/ReplicaImp.h
```diff
@@ -126,22 +126,26 @@ class ReplicaImp : public IReplica,
     }
   };
 
- private:
+  /**
+   * (Byzantine replica work) The following attributes shouldn't be "relaxed" from private to protected, if there was a
```
teoparvanov, line 130: I don’t understand the comment here...

tests/simpleKVBC/ByzantineReplica/README.md
```diff
@@ -0,0 +1,20 @@
...
+```
+Those contain the main interfaces and classes to start a Byzantine replica with the proper settings in the Apollo tests context and to extend the default replica logic in order to add different Byzantine behavior.
+
+```
+- setup.hpp
+- setup.cpp
+- internalCommandsHandler.hpp
+- internalCommandsHandler.cpp
+```
+Those contain helper interfaces and clases for replica initialization purposes in the Apollo tests context. They are actually pretty similar to the ones within the `../TesterReplica` folder, and with a bit of refactoring they could be shared and reused.
```
teoparvanov, line 18: As a first milestone, I suggest to reuse the same commands handler layer, to avoid the duplication here. Down the road, As you point out in your comment, there are cleaner ways to reuse and extend this code.

tests/simpleKVBC/ByzantineReplica/main.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "rocksdb/key_comparator.h"
+#endif
+
+#include <memory>
+
+namespace concord::kvbc::test {
+
+std::atomic_bool timeToExit = false;
+
+void run_replica(int argc, char** argv) {
```
teoparvanov, line 36: As suggested above, maybe we could start simple by just having a main.cpp in order to instantiate `ByzantineKvbcReplicaImp` and reuse all the rest from the normal replica (incl. the internal commands handler).

teoparvanov (changes requested):
Hi Sergio, thanks for producing this PR! I have a couple of comments, that mainly have to do with the significant code duplication.

I’d like to see it reduced as much as possible in favour of an object oriented approach, “a la open/closed principle”.

PR #1137, opened 2021-01-11
WIP: Add deleteGenesisBlock for BlockMerkle category

andrewjstone:
We delete keys when we prune a block and the version of the key is not
the latest version. If the key is the latest version we call it the
`active` key.

When all keys in a block are inactive we remove the block from the
merkle tree. Otherwise we write a new version of the block in the merkle
tree that contains all the active keys. This is necessary for proving
the still active keys.

When we update the merkle tree, we now save any stale keys so we can
delete them when they are no longer referenced. This is necessary for
the safety of the sparse merkle tree.

A serious safety issue was discovered in testing, due to the use of
thread local buffer references in `detail::serialize`. When a reference
to the buffer is used in a batch.put operation for both a key and a
value, the value overwrites the key. This also occurs in any put
operations with multiple key references. This is extremely dangerous
and can result in different keys being written if a temporary without a
reference is used in a serialized key vesion. For example, the
difference between `auto x = serialize(key)` and
`batch.put(serialize(key), serialize(val))` is significant. Therefore
the `serialize` wrapper was changed to always allocate a new buffer. The
old behavior is preserved when users opt into it knowingly with
`serialiizeThreadLocal`.

There are 2 problems with this code that must be resolved before merge

Problem 1:

 After we prune a block that contains active keys, those keys are not
 deleted. However, when we delete the keys or overwrite them in a new
 block addition after the block has been pruned, we overwrite the latest
 version. We do not, however, go back and see if we overwrote a key from
 a pruned block. This means that in the current code, the keys are never
 pruned, nor are the blocks pruned from the merkle tree. This is the
 cause of the test failure.

 If we overwote a key from an already pruned block, we must:
  1. Remove the key.
  2. Rewrite the block in the merkle tree by removing the key from the
  `active_keys` list, and updating the root hash. If this was the last
  active key, we must delete the block from the merkle tree.

 There are two strategies we can use for this:
  1. On any overwritten (or deleted) active key, we can immediately do
  the two steps above inline with the add block code. This has the
  drawback of possibly requiring multiple block rewrites for a single
  addition, thus slowing down the hotpath. It has the advantage of
  immediately freeing unused space, and not requiring storing anymore
  metadata.
  2. We add an extra field to the `MerkleBlockValue` for the newly added
  block that contains the keys from deleted blocks that were
  overwritten. Then when we prune the newly added block we also go back
  and cleanup the stale data.

 I prefer strategy 2, as it requires less work on the hot path, and it
 keeps the addBlock code simple.

Problem 2:

 When we prune a block, we create new tree versions, but do not create
 corresponding blocks for those tree versions. These tree versions have
 corresponding stale keys that must be deleted when blocks generated
 after the tree version are created (with a new tree version). The code
 currently does this correctly as shown by the unit tests.

 The problem arises when many blocks are pruned before a block is added.
 This is a typical case when pruning is performed during a maintenance
 window. In this case, when the first newly added block is pruned the
 next time, all the stale keys must be removed from the last pruning.
 This is fine from a computational sense, as enough time should be
 allocated to do the pruning. The problem is that we currently add all
 these keys to a single writeBatch so the main category code can
 atomically delete them. This means that a write batch can have millions
 of entries!. Luckily, this is completely unnecessary, and we
 can just delete these nodes in a non-atomic manner as they are unused
 by the time the current block is being pruned. All blocks that would
 have referenced them (in between the two pruning sessions) have already
 been pruned since we prune blocks in ascending order.

 There are a few strategies we can use, but a simple proposal is to just
 do the deletes inline via the db before returning from the call to
 `deleteGenesisBlock`. Direct deletes performed in batches with
 batched multigets for N stale versions, will be an efficient way to
 delete, since RocksDB works best with a single writer anyway. Then the
 batch we return only contains the actual updates for the block and not
 the intermediate tree versions and their stale keys.

### DIFF
--- a/kvbc/test/categorization/block_merkle_category_unit_test.cpp
+++ b/kvbc/test/categorization/block_merkle_category_unit_test.cpp
@@ -18,6 +18,7 @@
 #include "kv_types.hpp"
 #include "rocksdb/native_client.h"
 #include "storage/test/storage_test_common.h"
+#include "categorization/column_families.h"
 
 #include <memory>
 #include <optional>
@@ -64,11 +65,11 @@ class block_merkle_category : public Test {
   std::string val3 = "val3"s;
   std::string val4 = "val4"s;
   std::string val5 = "val5"s;
-  Hash hashed_key1 = Hasher{}.digest(key1.data(), key1.size());
-  Hash hashed_key2 = Hasher{}.digest(key2.data(), key2.size());
-  Hash hashed_key3 = Hasher{}.digest(key3.data(), key3.size());
-  Hash hashed_key4 = Hasher{}.digest(key4.data(), key4.size());
-  Hash hashed_key5 = Hasher{}.digest(key5.data(), key5.size());
+  Hash hashed_key1 = hash(key1);
+  Hash hashed_key2 = hash(key2);
+  Hash hashed_key3 = hash(key3);
+  Hash hashed_key4 = hash(key4);
+  Hash hashed_key5 = hash(key5);
 };
 
 TEST_F(block_merkle_category, empty_updates) {
@@ -76,8 +77,8 @@ TEST_F(block_merkle_category, empty_updates) {
   auto batch = db->getBatch();
   const auto output = cat.add(1, std::move(update), batch);
 
-  // A new root index, an internal node, and a leaf node are created.
-  ASSERT_EQ(batch.count(), 3);
+  // A new root index, an internal node, a leaf node, and stale index node are created.
+  ASSERT_EQ(batch.count(), 4);
 }
 
 TEST_F(block_merkle_category, put_and_get) {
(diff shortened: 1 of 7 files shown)


kvbc/test/categorization/block_merkle_category_unit_test.cpp
```diff
@@ -257,6 +258,113 @@ TEST_F(block_merkle_category, updates_with_deleted_keys) {
...
+  // Adding a new block adds stale internal keys, but no stale leaf keys, as blocks aren't deleted.
+  ASSERT_LT(0, stale.internal_keys.size());
+  ASSERT_EQ(0, stale.leaf_keys.size());
+
+  // There have been no pruned blocks yet.
+  ASSERT_EQ(0, cat.getLastDeletedTreeVersion());
+  ASSERT_EQ(2, cat.getLatestTreeVersion());
+
+  // Pruning the first block causes key2, key3, and key5 from version 1 to be deleted.
+  // Keys 1, 4, 5 are still active at version 1 and so remain in the database.
```
dartdart26, line 288: I think this comment applies to key 1 and key 4 only.

andrewjstone: Good catch!

andrewjstone:
Closed in favor of #1143

PR #1138, opened 2021-01-11
Value-initialize CMF C++ struct members

dartdart26:
Value-initialize CMF C++ struct members for clarity and safety. Do it
via `{}`, effectively:
 * initializing integer types to 0
 * initializing booleans to false
 * value-initializing std::pair members
 * value-initializing std::array members

Do not change Python code generation as all message members are
initialized to None, preventing indeterminate values.


cloudnoize (approved):
This is so awesome @dartdart26

andrewjstone (approved):
Excellent! Thanks @dartdart26

PR #1139, opened 2021-01-11
Wire the read interface to the categories implementation

cloudnoize:
Most of the work done by the adapter is to delegate the calls to the corresponding category.
Most of the code in this PR is unit testing.

### DIFF
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -55,24 +55,28 @@ class KeyValueBlockchain {
   /////////////////////// Read interface ///////////////////////
 
   // Gets the value of a key by the exact blockVersion.
-  std::optional<Value> get(const std::string& cat_id, const std::string& key, BlockId block_id) const;
-  std::optional<Value> getLatest(const std::string& cat_id, const std::string& key) const;
+  std::optional<Value> get(const std::string& category_id, const std::string& key, BlockId block_id) const;
 
-  void multiGet(const std::string& cat_id,
+  std::optional<Value> getLatest(const std::string& category_id, const std::string& key) const;
+
+  void multiGet(const std::string& category_id,
                 const std::vector<std::string>& keys,
                 const std::vector<BlockId>& versions,
                 std::vector<std::optional<Value>>& values) const;
 
-  void multiGetLatest(const std::string& cat_id,
+  void multiGetLatest(const std::string& category_id,
                       const std::vector<std::string>& keys,
                       std::vector<std::optional<Value>>& values) const;
 
-  std::optional<BlockId> getLatestVersion(const std::string& cat_id, const std::string& key) const;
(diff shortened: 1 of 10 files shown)


kvbc/include/categorization/kv_blockchain.h
```diff
@@ -55,24 +55,29 @@ class KeyValueBlockchain {
...
+
+  void multiGetLatestVersion(const std::string& category_id,
                              const std::vector<std::string>& keys,
-                             std::vector<std::optional<BlockId>>& versions) const;
+                             std::vector<std::optional<categorization::TaggedVersion>>& versions) const;
 
-  CategoryInput getBlockData(BlockId block_id);
+  // Get the updates that were used to create `block_id`.
+  Updates getBlockUpdates(BlockId block_id) const;
+  CategoryInput getCategoryInput(const Updates& updates) const;
```
dartdart26, line 80: Nit: Does **getCategoryInput()** need to be public? Can we make it private?

kvbc/src/categorization/blocks.cpp
```diff
@@ -108,6 +108,7 @@ ImmutableInput RawBlock::getUpdates(const std::string& category_id,
   ConcordAssert(categorires != nullptr);
   ConcordAssert(categorires->count(category_id) == 1);
   ImmutableInput data;
+  data.calculate_root_hash = update_info.tag_root_hashes.has_value();
```
dartdart26, line 111: Nit: Not related to this PR, but we might change the names of these functions from updates to something related to input/output. Just came to mind, we can totally think about it later.

kvbc/include/categorization/kv_blockchain.h
dartdart26: I think the answer is testing, if I am not mistaken.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -143,101 +143,63 @@ std::future<BlockDigest> KeyValueBlockchain::computeParentBlockDigest(const Bloc
...
-        // cat.multiGetLatestVersion(keys,values);
-        return;
-      },
-      category);
+Updates KeyValueBlockchain::getBlockUpdates(BlockId block_id) const {
+  return Updates{std::move(getRawBlock(block_id).data.updates)};
 }
 
-CategoryInput KeyValueBlockchain::getBlockData(BlockId block_id) { return getRawBlock(block_id).data.updates; }
+CategoryInput KeyValueBlockchain::getCategoryInput(const Updates& updates) const { return updates.category_updates_; }
```
dartdart26, line 202: Do we essentially need this method to use **CategoryInput**'s operator==() from CMF? If yes, for convenience, we might define operator==() for `Updates` too.

dartdart26 (approved):
Nice @cloudnoize, thanks! It's starting to shape up :)

PR #1140, opened 2021-01-11
Fix call to abi::__cxa_demangle() with invalid buf

dartdart26:
abi::__cxa_demangle() requires that the output buffer parameter is
either nullptr or allocated via malloc(). Therefore, remove the funcName
stack array and, instead, pass in a nullptr. Then, always free the
returned buffer from abi::__cxa_demangle().
References:
* cxxabi.h
* https://gcc.gnu.org/onlinedocs/gcc-7.5.0/libstdc++/api/a00023_source.html

Do above so that valgrind doesn't complain when we assert.


andrewjstone (approved):
Thanks for the fix @dartdart26

PR #1142, opened 2021-01-11
Added new View Change Apollo test.

HristoStaykov:
In this test we set things up to get a Replica 2 Views behind
its peers and verify it correctly manages to enter the View
and participate in quorum for execution and the system reaches
Fast Path. This test fails if we don't do a tryToEnterView() after
State Transfer.


yontyon (approved):
LGTM! 
Thanks!

teoparvanov:
Hi @HristoStaykov , is a code change needed, in order to make this test succeed?

HristoStaykov:
Yes, the fix in the code is required, otherwise we do not recover the Fast Path in 100% of the runs.

HristoStaykov:
I will close this MR and split it. The Apollo test is not passing consistently even after the fix. This requires further investigation of the instability as a separate effort.

PR #1143, opened 2021-01-13
Add deleteGenesisBlock for BlockMerkle category

andrewjstone:
We delete keys when we prune a block and the version of the key is not
the latest version. If the key is the latest version we call it the
`active` key.

When all keys in a block are inactive we remove the block from the
merkle tree. Otherwise we write a new version of the block in the merkle
tree, as well as metadata indexes that contain the active keys by block
id (`PrunedBlock` msg), and the pruned block version by hashed key. This
metadata is necessary for proving the still active keys as well as
allowing garbage collection of prior active keys and
block-rewriting/deletion in the merkle tree to be postponed until prune
operations. Essentially, we make `add` faster at the cost of
`deleteGenesisBlock` and some small storage overhead. This work is
performed in the `writePrunedBlock` and `rewriteAlreadyPrunedBlocks`
methods.

When we update the merkle tree, we now save any stale keys so we can
delete them when they are no longer referenced. This is necessary for
the safety of the sparse copy-on-write merkle tree. Nodes in un-pruned
tree versions must be able to retrieve data from nodes that may have
been generated in blocks that were already pruned. We keep these nodes
available until the last referencing tree version gets pruned.

During pruning, new tree versions are created and nodes are marked
stale. These nodes are created because the merkle tree is modified by
either inserting an updated version of a block id and hash for a pruned
block with active keys, or the pruned block has no active keys, in which
case it is removed from the tree. These tree versions are not tied to
any block, since we don't create new blocks when we prune a block.
However, the stale nodes that get generated as a result of deleting
blocks must be garbage collected at some point. After pruning, when a
new block is added, it will be at a version greater than those stale
nodes. When this new block is eventually pruned, all prior stale
versions are finally deleted from the database. This work is performed
by the `deleteStaleData` method and tested in `prune_many_nodes`.

In order to make the deletion of stale nodes as a result of pruning many
blocks in a row before adding a new block efficient, we do it in atomic
batches in the `deleteStaleBatch` method. These batches are deleted
directly and not returned as one giant batch to the caller because:
 1. The stale keys do not need to all be deleted in one atomic batch.
 2. There can be upwards of a million versions to delete for large,
 uninterrupted pruning sessions done during maintenance windows.

A serious safety issue was discovered in testing, due to the use of
thread local buffer references in `detail::serialize`. When a reference
to the buffer is used in a `batch.put` operation for both a key and a
value of the same type, the value overwrites the key. This also occurs
in any put operations with multiple key references. This is extremely
dangerous and can result in different keys being written if a temporary
without a reference is used in a serialized key vesion. For example, the
difference between `auto x = serialize(key)` and
`batch.put(serialize(key), serialize(val))` is significant. Therefore
the `serialize` wrapper was changed to always allocate a new buffer. The
old behavior is preserved when users opt into it knowingly with
`serialiizeThreadLocal`.

### DIFF
--- a/kvbc/include/categorization/block_merkle_category.h
+++ b/kvbc/include/categorization/block_merkle_category.h
@@ -70,20 +70,107 @@ class BlockMerkleCategory {
                              std::vector<std::optional<TaggedVersion>>& versions) const;
   void multiGetLatestVersion(const std::vector<Hash>& keys, std::vector<std::optional<TaggedVersion>>& versions) const;
 
+  // Delete the given block ID as a genesis one.
+  // Precondition: The given block ID must be the genesis one.
+  void deleteGenesisBlock(BlockId, const BlockMerkleOutput&, storage::rocksdb::NativeWriteBatch&);
+
+  // Delete the given block ID as a last reachable one.
+  // Precondition: The given block ID must be the last reachable one.
+  void deleteLastReachableBlock(BlockId, const BlockMerkleOutput&, storage::rocksdb::NativeWriteBatch&);
+
+  uint64_t getLatestTreeVersion() const;
+  uint64_t getLastDeletedTreeVersion() const;
+
  private:
   void multiGet(const std::vector<Buffer>& versioned_keys,
                 const std::vector<BlockId>& versions,
                 std::vector<std::optional<Value>>& values) const;
 
-  void putKeys(storage::rocksdb::NativeWriteBatch& batch,
-               uint64_t block_id,
-               std::vector<KeyHash>&& hashed_added_keys,
-               std::vector<KeyHash>&& hashed_deleted_keys,
-               BlockMerkleInput& updates);
+  // The last deleted tree version is stored at key `0` in BLOCK_MERKLE_STALE_CF
+  void putLastDeletedTreeVersion(uint64_t tree_version, storage::rocksdb::NativeWriteBatch&);
 
-  void putMerkleNodes(storage::rocksdb::NativeWriteBatch& batch,
-                      sparse_merkle::UpdateBatch&& update_batch,
-                      uint64_t tree_version);
+  // During initial pruning of a genesis block, rewrite the merkle tree value if some keys are still active.
+  // This is necessary for proofs.
+  //
+  // The data written in this function is the data that gets rewritten or removed in
+  // `rewriteAlreadyPrunedBlocks`.
+  //
+  // Retrieve the value of all active keys so we can recalculate the root hash for the modified block.
+  // Write the updated block to the merkle tree.
+  // Also write corrsponding active key indexes and pruned block index to their respective column families.
+  void writePrunedBlock(BlockId, std::vector<KeyHash>&& active_keys, storage::rocksdb::NativeWriteBatch&);
+
+  // When a block gets pruned, any keys that are still `active` (latest version of a key) are
+  // tracked in a `PrunedBlock`. PrunedBlocks are kept in their own column family. Additionally, a
+  // new root hash is calculated for all remaining active keys in the block, and a new version of
+  // the `MerkleBlockValue` is written into the block merkle tree. Lastly, each active key hash is
+  // written into a separate column family with the pruned block_id as the value. These latter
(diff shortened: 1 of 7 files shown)


andrewjstone:
This PR supercedes #1137

kvbc/test/categorization/block_merkle_category_unit_test.cpp
```diff
@@ -257,6 +258,214 @@ TEST_F(block_merkle_category, updates_with_deleted_keys) {
...
+  const auto block_out3 = add(3, BlockMerkleInput{{}, {{key1, key2, key4}}});
+  ASSERT_FALSE(cat.getLatest(key1));
+  ASSERT_FALSE(cat.getLatest(key2));
+  ASSERT_FALSE(cat.getLatest(key3));
+  ASSERT_FALSE(cat.getLatest(key4));
+  ASSERT_FALSE(cat.getLatest(key5));
+  ASSERT_EQ(2, cat.getLastDeletedTreeVersion());
+  ASSERT_EQ(5, cat.getLatestTreeVersion());
+
+  // We still see tombstones for keys 1, 2, 3
```
dartdart26, line 335: Nit: Typo -> for keys 1, 2, 4

dartdart26:
Excellent PR description, thank you!

kvbc/include/categorization/block_merkle_category.h
```diff
@@ -70,20 +70,107 @@ class BlockMerkleCategory {
...
+  // garbage collection and update process is what is performed by this function.
+  //
+  // To reiterate: pruning (calling `deleteGenesisBlock`) does not always fully delete all data in a
+  // block. If a key is at the latest version when the block is deleted, it is still `active` and
+  // users can retrieve its data and prove that it is part of the blockchain. Any active keys become
+  // `deactivated` when a new version of the key is written in a later block, or the key is deleted
+  // in a later block. However, even though the key is no longer the latest version, it is not
+  // deleted in this step. Final deletion from the database only occurs when the overwriting block
+  // is itself pruned. This tradeoff is made to defer work to the pruning process and not cause
+  // block addition to slow down.
```
dartdart26, line 128: Great explanation!

kvbc/include/categorization/block_merkle_category.h
```diff
@@ -70,20 +70,107 @@ class BlockMerkleCategory {
...
+                                  storage::rocksdb::NativeWriteBatch& batch);
+
+  // When a genesis block is pruned, we must delete all data that is stale as of `tree_version`.
+  //
+  // This includes data that may be written in prior tree versions as a result of prior pruning
+  // operations that generate new tree versions and stale data, but not new blocks. There can be a
+  // large amount of these prior versions, as each pruned block generates a new tree version. If we
+  // prune X blocks in a row before we add a new block, Y, then when we prune Block Y, we will have
+  // to lookup all the indexes for those pruned X tree versions and delete the internal and leaf
+  // nodes in the indexes.
```
dartdart26, line 139: Nit: Maybe we can call the number of blocks we prune in a row `N` and the newly-added block `X`.

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -18,38 +18,38 @@
 #include "assertUtils.hpp"
 #include "kv_types.hpp"
 #include "sha_hash.hpp"
+#include <iostream>
```
dartdart26, line 21: Nit: Do we need it?

dartdart26 (approved):
Great implementation, clean and understandable code, thanks @andrewjstone!

andrewjstone:
Thank you very much for the review @dartdart26. In the interest of saving sanity and time while I'm working temporarily on other code, I'm going to just merge this in. I'll fix the nits in the follow up commit where I implement `deleteLastReachableBlock`.

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: nope. We don't. I was printf debugging. I need to add a proper logger.

PR #1144, opened 2021-01-13
Integrate the categorized KVBC interfaces

dartdart26:
Implement the categorized KVBC interfaces in the KVBC replica and
integrate them into the whole codebase. Depreate the non-categorized
interfaces.

Define a Concord and Concord-BFT internal category of type
VersionedKeyValueCategory. It can be used for various kinds of metadata.

Serialize the last sequence number in BlockMetadata as an 8 byte
big-endian integer. Make sure it is used accordingly in Concord.

Use the categorized KVBC for normal read-write ReplicaImps. Use the
IDbAdapter instance for read-only ones.

Make sure the v1DirectKeyValue::DBAdapter::saveKvPairsSeparately_ flag
is honored and blocks with the new (and incompatible) categorization
format are not deserialized during addRawBlock().

Make IReplica::addBlockToIdleReplica() compatible with the categorized
KVBC. Also, throw an exception in case it fails instead of returning a
Status.

TODOs in subsequent commits:

- [ ] Use getSlice() in the categorized blockchain/kv_blockchain code where
   possible (for performance)
- [ ] Either remove or handle the USE_ROCKSDB compile flag correctly
- [ ] Simplify Updates::categoryUpdates()
- [x] Support both VersionedKeyValue and BlockMerkle categories in
   TesterReplica
- [x] Consider running with different category types (or a mix) in Apollo
   tests
- [x] Consider making IReader::getBlockUpdates() return an optional
- [ ] Consider an assert in ReplicaImp::putBlock() when an existing and
   different block is found in storage during ST
- [ ] If possible, change the way we test for linearizability in Apollo and
   get rid of the linear InternalCommandsHandler::getAtMost() - consider
   at a future stage
- [ ] Support categorization in the slowdown framework and its unit test
- [x] Consider implementing the persistence-mode TesterReplica option via
   an in-memory RocksDB instance - removed support for the "-p" option
   altogether
- [ ] Consider removing addBlockToIdleReplica() - left for future commits
- [ ] Consider removing the IBlockMetadata interface - left for future commits

### DIFF
--- a/kvbc/include/KVBCInterfaces.h
+++ b/kvbc/include/KVBCInterfaces.h
@@ -100,11 +100,11 @@ class IReplica {
 
   // Used to read from storage, only when a replica is Idle. Useful for
   // initialization and maintenance.
-  virtual const concord::kvbc::ILocalKeyValueStorageReadOnly& getReadOnlyStorage() = 0;
+  virtual const concord::kvbc::IReader& getReadOnlyStorage() const = 0;
 
   // Used to append blocks to storage, only when a replica is Idle. Useful
   // for initialization and maintenance.
-  virtual Status addBlockToIdleReplica(const concord::kvbc::SetOfKeyValuePairs& updates) = 0;
+  virtual concord::kvbc::BlockId addBlockToIdleReplica(concord::kvbc::categorization::Updates&& updates) = 0;
 
   /// TODO(IG) the following methods are probably temp solution,
   /// need to split interfaces implementations to differrent modules
--- a/kvbc/include/db_interfaces.h
+++ b/kvbc/include/db_interfaces.h
@@ -74,7 +74,8 @@ class IReader {
                                      std::vector<std::optional<categorization::TaggedVersion>> &versions) const = 0;
 
   // Get the updates that were used to create `block_id`.
-  virtual categorization::Updates getBlockUpdates(BlockId block_id) const = 0;
+  // Return std::nullopt if this block doesn't exist.
+  virtual std::optional<categorization::Updates> getBlockUpdates(BlockId block_id) const = 0;
 
   // Get the current genesis block ID in the system.
   virtual BlockId getGenesisBlockId() const = 0;
--- a/kvbc/include/db_interfaces.h
+++ b/kvbc/include/db_interfaces.h
@@ -101,34 +101,4 @@ class IBlocksDeleter {
   virtual ~IBlocksDeleter() = default;
 };
 
-// Deprecated interfaces. Will be removed after integration.
-class ILocalKeyValueStorageReadOnly {
- public:
-  // convenience where readVersion==latest, and block is not needed?
-  virtual concordUtils::Status get(const Key &key, Value &outValue) const = 0;
-  virtual concordUtils::Status get(BlockId readVersion, const Key &key, Value &outValue, BlockId &outBlock) const = 0;
-
-  // Returns the genesis block ID. If the blockchain is empty, 0 is returned.
-  // Throws on errors.
-  virtual BlockId getGenesisBlock() const = 0;
-  virtual BlockId getLastBlock() const = 0;
-  virtual concordUtils::Status getBlockData(BlockId blockId, SetOfKeyValuePairs &outBlockData) const = 0;
-  // TODO(GG): explain motivation
-  virtual concordUtils::Status mayHaveConflictBetween(const Key &key,
-                                                      BlockId fromBlock,
-                                                      BlockId toBlock,
-                                                      bool &outRes) const = 0;
-
-  virtual ~ILocalKeyValueStorageReadOnly() = default;
-};
-
-class IBlocksAppender {
- public:
-  virtual concordUtils::Status addBlock(const SetOfKeyValuePairs &updates,
-                                        BlockId &outBlockId,
-                                        const concordUtils::SpanWrapper &parent_span = concordUtils::SpanWrapper{}) = 0;
-
-  virtual ~IBlocksAppender() = default;
-};
-
 }  // namespace concord::kvbc
--- a/kvbc/include/db_interfaces.h
+++ b/kvbc/include/db_interfaces.h
@@ -14,49 +14,16 @@
 #include <vector>
 
 namespace concord::kvbc {
-/**
- *
- */
-class ILocalKeyValueStorageReadOnly {
- public:
-  // convenience where readVersion==latest, and block is not needed?
-  virtual concordUtils::Status get(const Key &key, Value &outValue) const = 0;
-  virtual concordUtils::Status get(BlockId readVersion, const Key &key, Value &outValue, BlockId &outBlock) const = 0;
 
-  // Returns the genesis block ID. If the blockchain is empty, 0 is returned.
-  // Throws on errors.
-  virtual BlockId getGenesisBlock() const = 0;
-  virtual BlockId getLastBlock() const = 0;
-  virtual concordUtils::Status getBlockData(BlockId blockId, SetOfKeyValuePairs &outBlockData) const = 0;
-  // TODO(GG): explain motivation
-  virtual concordUtils::Status mayHaveConflictBetween(const Key &key,
-                                                      BlockId fromBlock,
-                                                      BlockId toBlock,
-                                                      bool &outRes) const = 0;
-
-  virtual ~ILocalKeyValueStorageReadOnly() = default;
-};
-
-/**
- *
- */
-class IBlocksAppender {
- public:
-  virtual concordUtils::Status addBlock(const SetOfKeyValuePairs &updates,
-                                        BlockId &outBlockId,
-                                        const concordUtils::SpanWrapper &parent_span = concordUtils::SpanWrapper{}) = 0;
-
-  virtual ~IBlocksAppender() = default;
-};
-
-// Categorized interfaces follow. Non-categorized interfaces will be deprecated after full integration of categorized
-// ones.
+// Concord and Concord-BFT internal category that is used for various kinds of metadata.
+// The type of the internal category is VersionedKeyValueCategory.
+inline const auto kConcordInternalCategoryId = "concord_internal";
 
 // Add blocks to the key-value blockchain.
 class IBlockAdder {
  public:
   // Add a block from the given categorized updates and return its ID.
-  virtual BlockId add(categorization::Updates &&, const concordUtils::SpanWrapper &parent_span) = 0;
+  virtual BlockId add(categorization::Updates &&) = 0;
 
   virtual ~IBlockAdder() = default;
 };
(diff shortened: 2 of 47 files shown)


tests/apollo/CMakeLists.txt
```diff
@@ -29,74 +29,74 @@ if (OMIT_TEST_OUTPUT)
...
-            "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} BUILD_COMM_TCP_TLS=${BUILD_COMM_TCP_TLS} TEST_NAME=skvbc_persistence_tests_${STORAGE_TYPE} python3 -m unittest test_skvbc_persistence ${TEST_OUTPUT}"
-            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
-
-    add_test(NAME skvbc_ro_replica_tests_${STORAGE_TYPE} COMMAND sh -c
-            "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} BUILD_COMM_TCP_TLS=${BUILD_COMM_TCP_TLS} TEST_NAME=skvbc_ro_replica_tests_${STORAGE_TYPE} python3 -m unittest test_skvbc_ro_replica ${TEST_OUTPUT}"
-            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
-  endif()
-endforeach()
+add_test(NAME skvbc_basic_tests COMMAND sh -c
+        "env ${APOLLO_TEST_ENV} STORAGE_TYPE=${STORAGE_TYPE} BUILD_COMM_TCP_TLS=${BUILD_COMM_TCP_TLS} TEST_NAME=skvbc_basic_tests python3 -m unittest test_skvbc ${TEST_OUTPUT}"
```
teoparvanov, line 33: Given that you remove the loop, where is the `$STORAGE_TYPE` variable set?

dartdart26: Good points. Will handle them in future commits. Ultimately, we will reuse STORAGE_TYPE and set to either in-memory RocksDB or persistent RocksDB. Or we can rename it to persistent / in-memory.

kvbc/include/KVBCInterfaces.h
```diff
@@ -100,11 +100,11 @@ class IReplica {
 
   // Used to read from storage, only when a replica is Idle. Useful for
   // initialization and maintenance.
-  virtual const concord::kvbc::ILocalKeyValueStorageReadOnly& getReadOnlyStorage() = 0;
+  virtual const concord::kvbc::IReader& getReadOnlyStorage() const = 0;
 
   // Used to append blocks to storage, only when a replica is Idle. Useful
   // for initialization and maintenance.
-  virtual Status addBlockToIdleReplica(const concord::kvbc::SetOfKeyValuePairs& updates) = 0;
+  virtual concord::kvbc::BlockId addBlockToIdleReplica(concord::kvbc::categorization::Updates&& updates) = 0;
```
andrewjstone, line 107: Do we even use this method? I don't see how it's any different from regular add block. We should remove it altogether if we can.

kvbc/include/db_interfaces.h
```diff
@@ -14,49 +14,16 @@
...
-                                        const concordUtils::SpanWrapper &parent_span = concordUtils::SpanWrapper{}) = 0;
-
-  virtual ~IBlocksAppender() = default;
-};
-
-// Categorized interfaces follow. Non-categorized interfaces will be deprecated after full integration of categorized
-// ones.
+// Concord and Concord-BFT internal category that is used for various kinds of metadata.
+// The type of the internal category is VersionedKeyValueCategory.
+inline const auto kConcordInternalCategoryId = "concord_internal";
```
andrewjstone, line 20: The category choice is already showing its value :)

kvbc/src/block_metadata.cpp
```diff
@@ -5,27 +5,28 @@
...
+#include "assertUtils.hpp"
+#include "endianness.hpp"
 
 namespace concord {
 namespace kvbc {
 
-Sliver BlockMetadata::serialize(uint64_t bft_sequence_num) const {
-  return Sliver::copy((char*)&bft_sequence_num, sizeof(uint64_t));
+std::string BlockMetadata::serialize(uint64_t bft_sequence_num) const {
+  return concordUtils::toBigEndianStringBuffer(bft_sequence_num);
```
andrewjstone, line 15: Yassssssss. So much cleaner.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -18,6 +18,14 @@ namespace concord::kvbc::categorization {
 
 using ::bftEngine::bcst::computeBlockDigest;
 
+template <typename T>
+void nullopts(std::vector<std::optional<T>>& vec, std::size_t count) {
+  vec.clear();
```
andrewjstone, line 23: Maybe add a call to `vec.resize(count);` here to minimize allocations?

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
```diff
@@ -141,27 +159,27 @@ bool InternalCommandsHandler::executeWriteCommand(uint32_t requestSize,
...
+  BlockId currBlock = m_storage->getLastBlockId();
 
   // Look for conflicts
   bool hasConflict = false;
   for (size_t i = 0; !hasConflict && i < writeReq->numOfKeysInReadSet; i++) {
-    m_storage->mayHaveConflictBetween(
-        buildSliverFromStaticBuf(readSetArray[i].key), writeReq->readVersion + 1, currBlock, hasConflict);
+    const auto key = std::string(readSetArray[i].key, KV_LEN);
+    const auto latest = m_storage->getLatestVersion(CAT_ID, key);
+    hasConflict = (latest && (latest->deleted || latest->version > writeReq->readVersion));
```
andrewjstone, line 169: Why don't we have to check the version for `latest->deleted`? What if the initial transaction read from a prior version that wasn't deleted?

kvbc/src/ReplicaImp.cpp
```diff
@@ -263,39 +255,38 @@ ReplicaImp::~ReplicaImp() {
...
+  auto view = std::string_view{blockData, blockSize};
+  const auto rawBlock = categorization::RawBlock::deserialize(view);
+  if (m_kvBlockchain->hasBlock(blockId)) {
+    const auto existingRawBlock = m_kvBlockchain->getRawBlock(blockId);
+    if (rawBlock != existingRawBlock) {
+      LOG_ERROR(logger,
+                "found existing (and different) block ID[" << blockId << "] when receiving from state transfer");
+
+      // TODO consider assert?
+      m_kvBlockchain->deleteBlock(blockId);
```
andrewjstone, line 276: Are we sure we want to delete the existing block? Is it better to just crash the node?

I think this can happen in 2 ways:
 * More than F replicas are malicious and therefore the received block is bad.
 * The disk state is corrupt.

In both cases it may be better to preserve the evidence.

kvbc/src/ReplicaImp.cpp
```diff
@@ -319,13 +310,42 @@ bool ReplicaImp::putBlock(const uint64_t blockId, const char *block_data, const
...
  * The caller is the owner of the memory
  */
 bool ReplicaImp::getBlock(uint64_t blockId, char *outBlock, uint32_t *outBlockSize) {
+  if (replicaConfig_.isReadOnly) {
+    return getBlockFromObjectStore(blockId, outBlock, outBlockSize);
+  }
+  const auto rawBlock = m_kvBlockchain->getRawBlock(blockId);
+  const auto &ser = categorization::RawBlock::serialize(rawBlock);
+  *outBlockSize = ser.size();
+  std::memcpy(outBlock, ser.data(), *outBlockSize);
```
andrewjstone, line 344: 😢

andrewjstone (commented):
This looks really good @dartdart26. Mostly just questions from my side.

kvbc/include/KVBCInterfaces.h
dartdart26: It is used in Concord and is part of IReplica. It might be possible to remove it.

kvbc/src/categorization/kv_blockchain.cpp
dartdart26: Nice idea, will do!

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
dartdart26: Actually, TesterReplica doesn't delete any keys and, AFAIK, this check is redundant anyway.

kvbc/src/ReplicaImp.cpp
dartdart26: I just wanted to preserve the existing behaviour in this PR. I agree with you, though.

kvbc/src/ReplicaImp.cpp
dartdart26: Yeah, the API of this method is problematic. Leaving for a future commit.

kvbc/include/KVBCInterfaces.h
andrewjstone: No need to do it in this commit.

kvbc/src/ReplicaImp.cpp
andrewjstone: Ah, I didn't realize it was existing behavior.

andrewjstone (approved):
Great work Petar! Just remove that redundant check and I think it's good to go.

dartdart26:
Added more TODOs in the list. They can be handled in a separate PR, after this one gets merged.

Next step is to use the BlockMerkleCategory in TesterReplica and, consequently, in Apollo tests.

tests/apollo/CMakeLists.txt
dartdart26: Since I've added support for a mix of categories in TesterReplica by default, I've removed `STORAGE_TYPE` altogether.

As discussed offline, I will also remove TesterReplica's persistence option (`-p`) and always use RocksDB. If we need in-memory operation, we can use RocksDB in a `tmpfs` or `ramfs` location.

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
dartdart26: I've changed that check to an assert as we never delete keys in TesterReplica at that stage.

dartdart26:
> Great work Petar! Just remove that redundant check and I think it's good to go.

I've replaced the `deleted` check with an assert as we never delete keys in TesterReplica.

dartdart26:
No longer work in progress, ready for final review.

kvbc/src/ReplicaImp.cpp
```diff
@@ -89,103 +91,89 @@ Status ReplicaImp::stop() {
...
-  const auto lastDeletedBlock = std::min(lastBlock, until - 1);
-  for (auto i = genesisBlock; i <= lastDeletedBlock; ++i) {
-    m_bcDbAdapter->deleteBlock(i);
+BlockId ReplicaImp::getLastBlockId() const {
+  const auto lastSt = m_kvBlockchain->getLastStatetransferBlockId();
+  if (lastSt) {
+    return *lastSt;
   }
-  return lastDeletedBlock;
+  return m_kvBlockchain->getLastReachableBlockId();
```
cloudnoize, line 176: As discussed offline, I think we should return the last reachable, we can do it after we merge

PR #1146, opened 2021-01-14
Adding size information to updates

cloudnoize:


dartdart26 (approved):
Thanks!

One minor thing - can we make these methods `const`?

PR #1147, opened 2021-01-15
Fix active keys from pruned blocks in versioned cat

dartdart26:
Delete active keys from pruned blocks in the VersionedKeyValueCategory.
Introduce an `active keys` column family that contains a mapping from
key to pruned block ID.

When deleting (pruning) genesis block 1, if a key `K` in block 1 remains
active, insert it into the `active keys` column family with an active
block version of 1.

If a future block 2 overwrites `K` and is itself pruned, query the
`active keys` column family and delete the active key at block version 1
from the `values` column family. If `K` from block 2 is itself active,
update key `K` in the `active keys` column family with an active block
version of 2. If `K` is no longer active, delete it from the `active
keys` column family.

Add unit tests to verify that behavior.

Use serializeThreadLocal() in places that are safe to do so.

Idea borrowed from: https://github.com/vmware/concord-bft/pull/1143
Thanks to @andrewjstone!

### DIFF
--- a/kvbc/include/categorization/column_families.h
+++ b/kvbc/include/categorization/column_families.h
@@ -17,16 +17,20 @@
 
 namespace concord::kvbc::categorization::detail {
 
-inline const auto IMMUTABLE_KV_CF_SUFFIX = std::string{"_immutable"};
-
-inline const auto VERSIONED_KV_VALUES_CF_SUFFIX = std::string{"_ver_values"};
-inline const auto VERSIONED_KV_LATEST_VER_CF_SUFFIX = std::string{"_ver_latest"};
-
+// Blockchain
 inline const auto BLOCKS_CF = std::string{"blocks"};
 inline const auto ST_CHAIN_CF = std::string{"st_chain"};
-
 inline const auto CAT_ID_TYPE_CF = std::string{"cat_id_type"};
 
+// ImmutableKeyValueCategory
+inline const auto IMMUTABLE_KV_CF_SUFFIX = std::string{"_immutable"};
+
+// VersionedKeyValueCategory
+inline const auto VERSIONED_KV_VALUES_CF_SUFFIX = std::string{"_ver_values"};
+inline const auto VERSIONED_KV_LATEST_VER_CF_SUFFIX = std::string{"_ver_latest"};
+inline const auto VERSIONED_KV_ACTIVE_KEYS_FROM_PRUNED_BLOCKS_CF_SUFFIX = std::string{"_ver_active"};
+
+// BlockMerkleCategory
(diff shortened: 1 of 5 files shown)


andrewjstone (commented):
This looks really good @dartdart26. Eventually we're going to have to get together and unify our algorithms that are used in multiple categories. We'll have something like an algorithms.h from the STL :)

I'm hesitant to click approve on this however. My concern is with `serializeThreadLocal`. Even though the types are different in keys and values, I worry that writing multiple keys of the same type in the same RocksDB `WriteBatch` will cause silent corruption. It's totally possible that a WriteBatch copies that data out of the local buffer, and that appears to be the case, as the tests pass. Nonetheless, I think `serializeThreadLocal` is extremely dangerous and I"m concerned about its use across the codebase. IMO it should only be used in read operations, and not writes for this reason. We know once we read a single large value and use it we are done with it.

My primary concern is not that a bug will be created, but that it will be silent corruption and very hard to even detect. I think we should rethink our strategy toward thread local storage or you should convince me that I have nothing to worry about here.

As an alternative we could keep a limited number of thread local buffers, and then when we exhaust them per operation, we use non-thread local ones, or increase the existing number of thread local ones. 

Another concern is that while thread local may be safe in isolation, the same buffer may be reused by other categories. That isn't the case in this code due to the strong types, but I could certainly see it happening in the future by accident.

dartdart26:
Hey @andrewjstone, I share most of your concerns. However, I think if we don’t store references to the internal buffer it should be clear enough, especially from the name.

In terms of a function that accepts a const ref (as the batch does), I think it is expected  a copy is done internally. Otherwise, the user should have been explicitly notified, in order to extend the lifetime of the parameters. 

IMO, the part that is really tricky is that serialize() is a template.

Could you please share an example that is confusing?

Bottom line is I feel it is safe enough for the detail namespace. However, I don’t have a really strong preference and if you feel it is risky, I can change it. Also, it is very late here and I might be wrong :)

andrewjstone:
> Hey @andrewjstone, I share most of your concerns. However, I think if we don’t store references to the internal buffer it should be clear enough, especially from the name.
> 
> In terms of a function that accepts a const ref (as the batch does), I think it is expected a copy is done internally. Otherwise, the user should have been explicitly notified, in order to extend the lifetime of the parameters.
> 
> IMO, the part that is really tricky is that serializeThreadLocal() is a template.
> 
> Could you please share an example that is confusing?
> 
> Bottom line is I feel it is safe enough for the detail namespace. However, I don’t have a really strong preference and if you feel it is risky, I can change it. Also, it is very late here and I might be wrong :)

Hey Petar, that's a great point about the const reference and lifetime. That argument, plus the tests and prior testing shows that this is safe. I think we should merge as is.

andrewjstone (approved):

PR #1148, opened 2021-01-18
Fix incorrect message size validation when 'REJECT' status is set

yuliasherman:
When non-primary replicas send a PreProcessReply message with the status "REJECT" to the primary, it does not contain a pre-execution result hash. So, the validate() function should expect to receive a short-sized message.

### DIFF
--- a/bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
+++ b/bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
@@ -42,19 +42,25 @@ void PreProcessReplyMsg::validate(const ReplicasInfo& repInfo) const {
   auto& msgHeader = *msgBody();
   ConcordAssert(msgHeader.senderId != repInfo.myId());
 
-  if (sigManager_ == nullptr) throw runtime_error(__PRETTY_FUNCTION__ + string(": verifySig"));
+  if (sigManager_ == nullptr) throw runtime_error(__PRETTY_FUNCTION__ + string(": sigManager_ is null"));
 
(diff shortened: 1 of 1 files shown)


teoparvanov:
Hi @HristoStaykov could you please take a look?

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
```diff
@@ -42,19 +42,25 @@ void PreProcessReplyMsg::validate(const ReplicasInfo& repInfo) const {
...
-    throw runtime_error(__PRETTY_FUNCTION__ + string(": verifySig"));
-}
+  if (msgHeader.status == STATUS_GOOD) {
+    if (size() < (sizeof(Header) + sigLen)) {
+      LOG_ERROR(logger(),
+                "Message size is too small" << KVLOG(
+                    msgHeader.senderId, msgHeader.clientId, msgHeader.reqSeqNum, size(), sizeof(Header) + sigLen));
+      throw runtime_error(__PRETTY_FUNCTION__ + string(": Message size is too small"));
+    }
+    concord::diagnostics::TimeRecorder scoped_timer(*preProcessorHistograms_->validateMessage);
```
teoparvanov, line 55: shouldn't this be inside the if ?

teoparvanov (approved):
LGTM!

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
yuliasherman: It should not be inside "if" as it measures how long verifySig takes

PR #1149, opened 2021-01-18
Handle non-atomic RocksDB column family creation

dartdart26:
RocksDB creates column families in 2 non-atomic steps:
 1. Create the column family in the DB.
 2. Create the column family in the options file by persisting the
    options provided by the user.

Reference:
https://github.com/facebook/rocksdb/blob/v6.8.1/db/db_impl/db_impl.cc#L2187

If the process is stopped between these steps, there could be a mismatch
between them. That could lead to:
 1. The DB open process failing due to missing column families in the
    call to Open() as the cf_descs variable that contains column family
    names and options is loaded from the options file.
 2. Losing the column family options the user provided.

In order to solve it, track incompletely created column families (i.e.
ones that have mismatching information about them in the DB and the
options file). If such a family is empty (i.e. has no keys inside), drop
it so that user code can re-create it. If not empty, warn and continue
the DB open process.

The test causing the issue to show up is `test_skvbc_persistence.py`
when used with key categorization.

### DIFF
--- a/storage/include/rocksdb/client.h
+++ b/storage/include/rocksdb/client.h
@@ -125,6 +125,7 @@ class Client : public concord::storage::IDBClient {
   concordUtils::Status launchBatchJob(::rocksdb::WriteBatch& _batchJob);
   concordUtils::Status get(const concordUtils::Sliver& _key, std::string& _value) const;
   bool keyIsBefore(const concordUtils::Sliver& _lhs, const concordUtils::Sliver& _rhs) const;
+  bool columnFamilyIsEmpty(::rocksdb::ColumnFamilyHandle*) const;
 
   // Column family unique pointers that are managed solely by Client. This allows us to use a raw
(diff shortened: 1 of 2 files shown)


cloudnoize (approved):
LGTM! thanks @dartdart26

dartdart26:
I am having a bit of second thoughts on whether dropping a column family is safe enough (in case there's a bug somewhere).

dartdart26:
After a discussion with @cloudnoize and @andrewjstone, I've added a commit to throw if an incomplete column family has data inside, instead of just logging.

Rationale is that:
 * we want to see if that happens during testing and by just logging it
   we might miss it
 * we don't want to run the system with possibly incorrect column family
   options as that can lead to slowdowns, affecting the whole cluster

andrewjstone (approved):
LGTM. Thanks Petar!

PR #1151, opened 2021-01-19
Add cid to the earliestPendingRequest info

yontyon:
To help us debug viewchanges we would like to know which client request was timed out.
For that, we added a cid filed to the ClientInfo. Every time we take the earllest pendin grequest, we also take its cid. In addition, we log this cid if we decided on a view change.


HristoStaykov (approved):
LGTM, thanks @yontyon !

PR #1152, opened 2021-01-19
Remove pending requests after the sequence number commit (instead of …

yontyon:
…execution)

The execution is a detemenistic part in the protocol that is not depended on other replicas data. Thus it make sense not to take the execution time in account for complaining the primary.
In this PR we remove the pending request after the commit instead of after the execution.
We expect this change to stablize VC and prevent unnecessary VCs

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -4011,7 +4011,23 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
       free(req.outReply);
       send(replyMsg.get(), req.clientId);
     }
-    clientsManager->removePendingRequestOfClient(req.clientId);
+  }
+}
+
+void ReplicaImp::tryToRemovePendingRequestsForSeqNum(SeqNum seqNum) {
+  SCOPED_MDC_SEQ_NUM(std::to_string(seqNum));
+  SeqNumInfo &seqNumInfo = mainLog->get(seqNum);
+  PrePrepareMsg *prePrepareMsg = seqNumInfo.getPrePrepareMsg();
+  if (prePrepareMsg == nullptr) return;
+  LOG_DEBUG(GL, "clear pending requests" << KVLOG(seqNum));
+
+  RequestsIterator reqIter(prePrepareMsg);
+  char *requestBody = nullptr;
+  while (reqIter.getAndGoToNext(requestBody)) {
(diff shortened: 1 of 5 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4011,7 +4011,24 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
...
     }
-    clientsManager->removePendingRequestOfClient(req.clientId);
+  }
+}
+
+void ReplicaImp::tryToRemovePendingRequestsForSeqNum(SeqNum seqNum) {
+  SCOPED_MDC_SEQ_NUM(std::to_string(seqNum));
+  SeqNumInfo &seqNumInfo = mainLog->get(seqNum);
+  PrePrepareMsg *prePrepareMsg = seqNumInfo.getPrePrepareMsg();
+  if (prePrepareMsg == nullptr) return;
```
HristoStaykov, line 4021: Though we have prior checks and we shouldn't reach this situation, might be it is better to log a warning message as definite indication of a bug.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4011,7 +4011,24 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
...
+  LOG_INFO(GL, "clear pending requests" << KVLOG(seqNum));
+
+  RequestsIterator reqIter(prePrepareMsg);
+  char *requestBody = nullptr;
+  while (reqIter.getAndGoToNext(requestBody)) {
+    ClientRequestMsg req((ClientRequestMsgHeader *)requestBody);
+    if (!clientsManager->isValidClient(req.clientProxyId())) continue;
```
HristoStaykov, line 4028: Same as above, might be we should warn that something wrong has happened here.

HristoStaykov (approved):
LGTM! great work @yontyon ! Just some minor recommendations.

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: I will add this in a future PR

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: This a valid situation, for example, with NOOP commnads

PR #1158, opened 2021-01-20
tests for mini batching in preexecution

KTeofanidis:
Updated the python client to work with batch requests. Updated history tracker to work with batch requests. New tests still disabled until the functionality becomes available.

The batch message consists of one or more regular messages. The replicas respond to each of these as if it was a separate message send on it's own. For this reason in order to determine when the whole batch is done RepliesManager will now expect a list of seq_nums to get a quorum of replies instead of individual messages. It will store full reply messages of these instead of just rsi_data, so after quorum is reached they are available to be returned by the client.

### DIFF
--- a/util/pyclient/bft_client.py
+++ b/util/pyclient/bft_client.py
@@ -81,7 +81,7 @@ class BftClient(ABC):
         self.req_seq_num = ReqSeqNum()
         self.client_id = config.id
         self.primary = None
-        self.reply = None
+        self.replies = None
         self.retries = 0
         self.msgs_sent = 0
         self.replies_manager = rsi.RepliesManager()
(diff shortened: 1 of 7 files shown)


tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -142,7 +142,7 @@ async def test_wedge_command_and_specific_replica_info(self, bft_network):
                 rsi_rep = client.get_rsi_replies()
                 done = True
                 for r in rsi_rep.values():
-                    if skvbc.parse_rsi_reply(rep, r) == 0:
+                    if skvbc.parse_rsi_reply(rep, r.get_rsi_data()) == 0:
```
teoparvanov, line 145: what's the reasoning behind this change?

KTeofanidis: For singe message a reply message was returned using the last of the messages that reached quorum. Since we have parallel messages now for a single quorum I changed the RepliesManager to store the entire MsgWithReplicaSpecificInfo instead of just the rsi_data, so it can be used after quorum is reached. Looking at it now however I should change the get_rsi_replies method to still return the same as before instead of this.

teoparvanov: Thanks @KTeofanidis . Can you provide a bit more explanation about the changes in the `bft_client` ?

util/pyclient/bft_client.py
```diff
@@ -163,14 +163,51 @@ async def sendSync(self, msg, read_only, seq_num=None, cid=None, pre_process=Fal
...
             pass
 
+    async def write_batch(self, msg_batch, batch_seq_nums=None, m_of_n_quorum=None):
+        if not self.comm_prepared:
+            await self._comm_prepare()
+
+        cid = str(self.req_seq_num.next())
+
+        if batch_seq_nums is None:
+            batch_seq_nums = []
```
teoparvanov, line 182: Modifying a parameter isn't very clean. Please use a local variable instead.

util/pyclient/bft_client.py
```diff
@@ -163,14 +163,51 @@ async def sendSync(self, msg, read_only, seq_num=None, cid=None, pre_process=Fal
...
+            for n in range(len(msg_batch)):
+                batch_seq_nums.append(self.req_seq_num.next())
+        
+        msg_data = b''
+        for n in range(len(msg_batch)):
```
teoparvanov, line 187: `len(msg_batch)` repeats a lot - let's call it `batch_size` for example

util/pyclient/replica_specific_info.py
```diff
@@ -78,8 +88,28 @@ def num_matching_replies(self, matched_reply_key):
...
+    def expects_seq_num(self, seq_num):
+        return seq_num in self.seq_nums.keys()
+
+    def has_quorum_on_all(self, required_replies):
+        has_quorum = True
+        for seq_num, key in self.seq_nums.items():
+            if key is None:
+                has_quorum = False
+            elif self.num_matching_replies(key) < required_replies:
+                has_quorum = False
```
teoparvanov, line 114: For a safer implementation, I'd actually reverse the logic here. Let's start by assuming `has_quorum = False` and then set it to `True` only in the relevant cases.

util/pyclient/bft_client.py
```diff
@@ -428,24 +468,25 @@ async def _receive_from_replica(self, dest_addr, replicas_addr, required_replies
...
         If there is an error in stages 1 or 2 - exit straight (connection is closed inside _stream_recv_some)
         """
-        if dest_addr not in self.ssl_streams:
-            self.establish_ssl_stream_parklot[dest_addr].unpark()
-            return
-        data = bytearray()
-        stream = self.ssl_streams[dest_addr]
-        while len(data) < self.MSG_HEADER_SIZE:
-            if not await self._stream_recv_some(data, dest_addr, stream, self.MSG_HEADER_SIZE):
+        while True:
```
teoparvanov, line 471: why is this additional while loop needed?

teoparvanov (commented):
LGTM on first pass. I have some questions before completing the review. Thanks!

util/pyclient/bft_client.py
KTeofanidis: The _recv_data method of the TCP client implementation is supposed to receive data and call _process_received_msg until the the enclosing cancel_scope times out just like the UDP implementation. Instead it would return. With normal messages with small replies it wasn't obvious, but with batches that wait for many replies it fails to reach quorum, because when it returns it's as if the retry timeout has been reached.

util/pyclient/bft_client.py
KTeofanidis: The same approach is used for non-required parameters in bft_client. I think we should keep it like this for consistency.

teoparvanov: Agreed, makes sense.

util/pyclient/bft_client.py
teoparvanov: Thanks for clarifying.

teoparvanov (approved):
LGTM!

util/pyclient/replica_specific_info.py
KTeofanidis: We check all the expected messages if they have reached quorum and if any single one of them hasn't then we don't have quorum.

PR #1159, opened 2021-01-20
Allow updating a value

cloudnoize:
The accumulation process in the command handlers requires updating values of already existing keys.

### DIFF
--- a/kvbc/include/categorization/updates.h
+++ b/kvbc/include/categorization/updates.h
@@ -59,7 +59,9 @@ struct ImmutableUpdates {
     friend struct ImmutableUpdates;
   };
 
-  void addUpdate(std::string&& key, ImmutableValue&& val) { data_.kv.emplace(std::move(key), std::move(val.update_)); }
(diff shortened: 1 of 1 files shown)


andrewjstone:
This looks fine @cloudnoize, but why wouldn't we always allow updating a value?

cloudnoize:
> This looks fine @cloudnoize, but why wouldn't we always allow updating a value?

We assumed that keys should be unique, so they shouldn't be updated. now I'm not that sure, maybe it should be the default behavior.

dartdart26:
> > This looks fine @cloudnoize, but why wouldn't we always allow updating a value?
> 
> We assumed that keys should be unique, so they shouldn't be updated. now I'm not that sure, maybe it should be the default behavior.
> @dartdart26 wdyt ?

I was thinking if we can just use `T& operator[]( Key&& key )` instead as `kv[std::move(key)] = std::move(value)`.

dartdart26 (approved):

PR #1160, opened 2021-01-21
Clear pending request on wedge point

yontyon:
To avoid VC after wedge, we clean all pending requests once we stopped


teoparvanov (approved):
LGTM, thanks @yontyon !

PR #1161, opened 2021-01-21
Use a random key as a known key in history tracker

dartdart26:
Use an uniquely random key instead of a constant key in the history
tracker. Rationale is that using a random key allows for better test
coverage. Moreover, it makes testing different key categories easier.

In order to generate an uniquely random key, use a prefix of '1' and
then generate KV_LEN - 1 random characters. That way, if the key starts
with '1', it will always be different from the keys generated by
_create_keys().

Thanks to @tdimitrov and @teoparvanov!


teoparvanov (approved):
LGTM!

tdimitrov (approved):

PR #1163, opened 2021-01-22
Better error logging for S3 client

tdimitrov:
Use S3_get_status_name() function to convert the error code of get and
put operation to a meaningful error. Additionally print the errorMessage
for the error code received.

Add additional error message received from s3 library on client error:
When the S3 client encounters an error an error code and optionally an
error message is returned. This patch adds the error message in the logs
for easier debugging.
Additionally put status error code is converted to meaningful string
from the enum (similar to get status).


dartdart26 (approved):
Thanks!

PR #1165, opened 2021-01-25
Memory leak fixes

KTeofanidis:
Minor memory leak fixes

### DIFF
--- a/bftengine/include/bftengine/CryptoManager.hpp
+++ b/bftengine/include/bftengine/CryptoManager.hpp
@@ -23,7 +23,9 @@ namespace bftEngine {
 class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
  public:
   static CryptoManager& instance(Cryptosystem* cryptoSys = nullptr) {
-    static CryptoManager cm_(cryptoSys);
+    // ensure cryptoSys will be freed since we're expected to take ownership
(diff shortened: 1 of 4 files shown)


bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -23,7 +23,9 @@ namespace bftEngine {
 class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
  public:
   static CryptoManager& instance(const ReplicaConfig* config = nullptr, Cryptosystem* cryptoSys = nullptr) {
-    static CryptoManager cm_(config, cryptoSys);
+    // ensure cryptoSys will be freed since we're expected to take ownership
+    std::unique_ptr<Cryptosystem> cryptoSysPtr(cryptoSys);
```
teoparvanov, line 27: I'm not sure why we need this intermediate `unique_ptr` here... In the `CryptoManager`, the raw pointer is reset into the `multiSigCryptoSystem_` unique_ptr. So I don't see how the raw pointer would leak, even without your changes?

KTeofanidis: The CryptoManager construction is only called the first time. For subsequent calls the raw pointer is left dangling.

HristoStaykov (approved):
LGTM, thanks @KTeofanidis.

teoparvanov (approved):
LGTM @KTeofanidis , thanks for the clarifications!

PR #1166, opened 2021-01-25
Support put of multiple slices to RocksDB WriteBatch

andrewjstone:
RocksDB supports writing multiple slices as a key or value to remove the
need to copy into a single buffer. This is most useful for us when we
want to write large values wrapped in a CMF type. In this case we can
serialize the header and not have to serialize the value at all thus
saving time.


dartdart26 (approved):
Very useful, thanks @andrewjstone !

PR #1167, opened 2021-01-26
Support block deletion in KeyValueBlockchain

dartdart26:
Delegate block deletion calls to the respective category in
KeyValueBlockchain.

Implement KeyValueBlockchain::getCategory() with a single std::map
lookup.

If there are no blocks in the blockchain, throw an exception instead of
silently returning.

Fix genesis and last reachable cache variables handling when the last
remaining block in the blockchain is being deleted as a last reachable
one. In that case, set both to 0 instead of decrementing them.

Add unit tests for above behaviors.


cloudnoize (approved):
LGTM thanks! @dartdart26

PR #1168, opened 2021-01-26
Fix genesis and last ID cache in v2Merkle adapter

dartdart26:
Fix genesis and last reachable cache variables handling when the last
remaining block in the blockchain is being deleted as a last reachable
one. In that case, set both to 0 instead of decrementing them.

Add an unit test to verify the behavior.


cloudnoize (approved):
LGTM

PR #1169, opened 2021-01-26
Added logging to the Slowdown manager

salieri11:
This PR adds logging and improved BusyWait implementation for the Slowdown manager


performance/include/SlowdownManager.hpp
```diff
@@ -173,51 +186,39 @@ class SleepPolicy : public BasePolicy {
...
+    for (uint i = 0; i < keyCount_; ++i) {
+      char *key_data = nullptr;
+      char *value_data = nullptr;
+      if (keySize_) {
+        key_data = new char[keySize_];
+        uint64_t v = counter_.fetch_add(1);
+        memcpy(key_data, &(v), sizeof(v));
+        for (uint j = sizeof(v); j < keySize_; ++j) {
+          key_data[j] = j;
+        }
```
dartdart26, line 204: Hi @salieri11 , thanks for improving the framework! I started looking into it and noticed a couple of things, just sharing them here, we can discuss further offline.

It is not directly related to this PR, but I think generating arbitrary keys and inserting them just prior to the RocksDB multiPut() can interfere with key prefixing for v2MerkleTree adapters, leading to incorrect behaviour. I think it is somewhat related to the offline discussion we had about the `const` updates.

Another related issue I faced is that if we implement addition of keys (Delay) in the slowdown framework itself, we make it dependent on `categorization` (from the kvbc library). I think it already is, because of some basic types (SetOfKeyValuePairs), for example. I haven't thought about how we can avoid it, but maybe we can discuss.

salieri11: Thanks for the review! As for arbitrary key, do you mean that accidentally there can be generated a key similar to what Merkle tree expects? In my understanding, writing arbitrary keys outside of KVBC should not introduce a problem, since if noone knows about this key then noone should ask for it? Maybe the case is when it can be accidentally read when using range scan or seekatleast/at most? Maybe the good solution is to define the prefix that is "safe" to be added?
As for the second remark, I dont like taking dependencies on KVBC types as well, however, due to the urgency and the intend to keep all slowdown code in the framework making calls to it very clean, I didn't had too much choices. I suggest to integrate it first, and then I can think about offloading the add part to KVBC (e.g. Delay may be implemented as Delay(void *dataset, function<void(void* data, int keycount, int keysize, int valuesize)>), and it will just call the callback (which will be implemented in the KVBC layer) with all needed parameters. Then KVBC can cast the void* to any type it is aware of (of course this is brutal cast, but I don't see how do it differently)

dartdart26: Thanks for the response @salieri11 !

Regarding arbitrary keys - yes, exactly. Just incrementing an integer as a key can insert at arbitrary positions and mess up seeks and so on. In terms of prefix - yes, we can define a prefix that won't clash with anything and just use it. Final note on this topic is that v2MerkleTree is going away soon and maybe we don't need to invest too much into it anymore.

Regarding KVBC - I understand and it is not a big deal, I think this dependency is there anyway. Agreed to integrate first. In terms of std::function - I think that can work and is a good solution. I really liked the way empty template functions are removed by the compiler, though, but calling an empty std::function won't kill us :)

salieri11: Cool. We are not going to use merkle related slowdown for now anyway so the arbitrary keys problem is irrelevant. So if you are ok pls approve it, i have another PR (unrelated) that is based on this one and we need it ASAP. Thanks!

dartdart26 (approved):
Thanks @salieri11 !

PR #1170, opened 2021-01-27
Fix batch requests retry issue

KTeofanidis:
Retry timeout was set to request timeout, so no retry was attempted in time. This breaks the view_change_test in the case where a client expects primary 0 which has been stopped. Also the view_change_test will now always require a retry instead of just when by chance the same client is used as for the first request. Enough time will be spent in send_indefinite_tracked_batch_writes for retries to be sent.

### DIFF
--- a/tests/apollo/test_skvbc_batch_preexecution.py
+++ b/tests/apollo/test_skvbc_batch_preexecution.py
@@ -24,6 +24,7 @@ import util.eliot_logging as log
 
 SKVBC_INIT_GRACE_TIME = 2
 BATCH_SIZE = 4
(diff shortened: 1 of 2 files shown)


tests/apollo/test_skvbc_batch_preexecution.py
```diff
@@ -210,7 +209,7 @@ async def test_view_change(self, bft_network, tracker):
         bft_network.stop_replica(initial_primary)
 
         try:
-            with trio.move_on_after(seconds=1):
+            with trio.move_on_after(seconds=BATCH_SIZE):
```
yuliasherman, line 212: Please define additional variable for this timeout for better readability

yuliasherman (approved):
LGTM

PR #1171, opened 2021-01-27
Reserved Pages usage

toly-kournik:
- ControlStateManager to work with IReservedPages* interface, instead of IStateTransfer*.
- KeyStore use IReservedPages* instead of IReservedPages& for uniformity and ability to seamlessly replace it by a shared_ptr.


yontyon (approved):
LGTM!
Thanks

PR #1172, opened 2021-01-27
Destroy the blockchain when no blocks are left

dartdart26:
Destroy the blockchain by dropping all category-related column families
and re-creating KeyValueBlockchain-related ones (effectively deleting
all data inside) when one block is left and it is deleted as a last
reachable one. This is a scenario that can happen during replica state
sync.

A note on above is that active keys from pruned blocks will also be
deleted, meaning that all data in the blockchain is destroyed.

Rationale for above behaviors is that we cannot allow any data leftovers
in case we have deleted all blocks in the blockchain as last reachable
(during replica state sync). Moreover, there might be active keys from
pruned blocks and since the blockchain will start again from block 1,
we don't want any keys at the same block IDs from the previous
blockchain in the new one.

Introduce a special BLOCKCHAIN_DESTRUCTION_INITIATED_KEY key in the
default column family that is used as a flag to denote a destruction
operation has started. Remove the key once done. That is useful if there
is an error during destruction itself - we want to continue the process
on the next startup. Additionally, we never drop the default column
family.

Related to above, provide two modes for handling failures during
destroy:
 * call std::terminate() - used for production
 * throw an exception - used for testing only, due to the fact that
   GTest's ASSERT_DEATH is implemented via fork() and that has
   implications on RocksDB that prevent us from using the
   std::terminate() version (DB updates in the child process are not
   seen in the parent)

Add unit tests to verify the behavior.

### DIFF
--- a/kvbc/include/categorization/block_merkle_category.h
+++ b/kvbc/include/categorization/block_merkle_category.h
@@ -81,6 +81,9 @@ class BlockMerkleCategory {
   uint64_t getLatestTreeVersion() const;
   uint64_t getLastDeletedTreeVersion() const;
 
+  // Destroy the category by dropping all column families and the data inside.
+  static void destroy(std::shared_ptr<storage::rocksdb::NativeClient>& db);
+
  private:
   void multiGet(const std::vector<Buffer>& versioned_keys,
                 const std::vector<BlockId>& versions,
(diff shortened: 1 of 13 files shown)


dartdart26:
Closing in favour of https://github.com/vmware/concord-bft/pull/1174.

PR #1173, opened 2021-01-27
Implement deleteLastReachable for the BlockMerkleCategory

andrewjstone:
This is required for sync between concord-bft metadata and the
application data in the `BlockMerkleCategory` on startup after a crash.

All tree nodes and stale indexes are removed for the block being
removed. Additionally all keys and version indexes are removed. The code
to do this latter part is based on code for the
`VersionedKeyValueCategory` by @dartdart26.

One important thing to keep in mind is around writing `DbValue` CMF
data. We write it as a `DbValueHeader` and the raw value passed in by
the caller as two separate slices, rather than forcing an extra copy
during serialization. We can do this using the RocksDB API enhancements
added in #1166. When we read out the value it comes back as one
serialized `DbValue`.

I also added fixes for the review comments made in #1143 and cleaned up
some comments.

### DIFF
--- a/kvbc/cmf/categorized_kvbc_msgs.cmf
+++ b/kvbc/cmf/categorized_kvbc_msgs.cmf
@@ -129,13 +129,25 @@ Msg Deleted 4005 {
     bool value
 }
 
-Msg ImmutableDbValue 4006 {
+# Used for efficiently *writing* a DbValue to RocksDB
+# Values may be large and we want avoid having to copy the string into a vector during serialization.
+# If we just serialize the header we can pass both as slices to RocksDB to prevent an extra copy.
+# On reads we retrieve a DbValue.
+#
+# This takes advantage of the fact that CMF serializes string size as a uint32. If that ever
+# changes we must change this as well.
+Msg DbValueHeader 4006 {
+    bool deleted
+    uint32 value_size
+}
+
+Msg ImmutableDbValue 4007 {
     uint64 block_id
     string data
 }
 
 # Used to deserialize the version only from an ImmutableDbValue.
-Msg ImmutableDbVersion 4007 {
+Msg ImmutableDbVersion 4008 {
     uint64 block_id
 }
 
--- a/kvbc/src/categorization/block_merkle_category.cpp
+++ b/kvbc/src/categorization/block_merkle_category.cpp
@@ -18,9 +18,9 @@
 #include "assertUtils.hpp"
 #include "kv_types.hpp"
 #include "sha_hash.hpp"
-#include <iostream>
 
 using concord::storage::rocksdb::NativeWriteBatch;
+using concord::storage::rocksdb::detail::toSlice;
 using concordUtils::Sliver;
 
 namespace concord::kvbc::categorization::detail {
@@ -157,6 +157,10 @@ std::vector<Buffer> versionedKeys(const std::vector<std::string>& keys, const st
   return versioned_keys;
 }
 
+void putLatestKeyVersion(NativeWriteBatch& batch, const Hash& key_hash, TaggedVersion version) {
+  batch.put(BLOCK_MERKLE_LATEST_KEY_VERSION_CF, key_hash, serializeThreadLocal(LatestKeyVersion{version.encode()}));
+}
+
 void putKeys(NativeWriteBatch& batch,
              uint64_t block_id,
              std::vector<KeyHash>&& hashed_added_keys,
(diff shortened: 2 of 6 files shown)


kvbc/cmf/categorized_kvbc_msgs.cmf
```diff
@@ -129,13 +129,25 @@ Msg Deleted 4005 {
...
 
-Msg ImmutableDbValue 4006 {
+# Used for efficiently *writing* a DbValue to RocksDB
+# Values may be large and we want avoid having to copy the string into a vector during serialization.
+# If we just serialize the header we can pass both as slices to RocksDB to prevent an extra copy.
+# On reads we retrieve a DbValue.
+#
+# This takes advantage of the fact that CMF serializes string size as a uint32. If that ever
+# changes we must change this as well.
+Msg DbValueHeader 4006 {
```
dartdart26, line 139: Nice!

kvbc/test/categorization/block_merkle_category_unit_test.cpp
```diff
@@ -466,6 +466,131 @@ TEST_F(block_merkle_category, prune_many_nodes) {
...
+  auto new_out = add(5, std::move(update));
+
+  ASSERT_EQ(all_leaves_5, getAllLeaves(db));
+  ASSERT_EQ(all_stale_5, getAllStale(db));
+  ASSERT_EQ(all_internal_5, getAllInternalNodes(db));
+  ASSERT_EQ(5, cat.getLatestTreeVersion());
+  ASSERT_EQ(5, cat.getLatestVersion(key1)->version);
+  ASSERT_EQ(expected5, asMerkle(*cat.getLatest(key1)));
+
+  // Deleting again generates the same column families as the prior delete
```
dartdart26, line 562: Great checks!

kvbc/test/categorization/block_merkle_category_unit_test.cpp
```diff
@@ -466,6 +466,131 @@ TEST_F(block_merkle_category, prune_many_nodes) {
...
+  // Deleting all blocks does the right thing
+  {
+    auto batch = db->getBatch();
+    for (auto i = 4u; i > 0; i--) {
+      cat.deleteLastReachableBlock(i, out[i - 1], batch);
+    }
+    db->write(std::move(batch));
+    ASSERT_EQ(0, getAllLeaves(db).size());
+    // There's always a key pointing to the latest root
+    ASSERT_EQ(1, getAllInternalNodes(db).size());
```
dartdart26, line 584: Nit: Based on our offline discussion, we might want to be stricter and not allow deletion of the last reachable block that is also genesis and, hence, relax these checks here.

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -511,6 +555,43 @@ void BlockMerkleCategory::deleteGenesisBlock(BlockId block_id, const BlockMerkle
...
+                                                   const BlockMerkleOutput& out,
+                                                   NativeWriteBatch& batch) {
+  for (const auto& [key, _] : out.keys) {
+    (void)_;
+    const auto hashed_key = KeyHash{hash(key)};
+    const auto versioned_key = serializeThreadLocal(VersionedKey{hashed_key, block_id});
+    // Find the previous version of the key and set it as a last version. Exploit the fact that CMF
+    // serializes strings prefixed by their length (in big-endian). Therefore, VersionedKeys will be ordered by key
+    // size, key and version. If not found, then this is the only version of the key and we can remove the latest
+    // version index too.
```
dartdart26, line 568: The `Exploit the fact that CMF serializes strings prefixed by their length (in big-endian)` and ordered by `key size` parts can be removed for the hashed keys that are `fixedlist`s.

dartdart26 (approved):

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -157,15 +157,23 @@ std::vector<Buffer> versionedKeys(const std::vector<std::string>& keys, const st
...
+
 void putKeys(NativeWriteBatch& batch,
              uint64_t block_id,
              std::vector<KeyHash>&& hashed_added_keys,
              std::vector<KeyHash>&& hashed_deleted_keys,
              BlockMerkleInput& updates) {
   auto kv_it = updates.kv.begin();
   for (auto key_it = hashed_added_keys.begin(); key_it != hashed_added_keys.end(); key_it++) {
+    // Only serialize the Header of a DBValue, to prevent the need to copy a potentially large value.
+    auto header = serialize(DbValueHeader{false, static_cast<uint32_t>(kv_it->second.size())});
```
dartdart26, line 172: Nit: I know you are not a big fan of `serializeThreadLocal()`, but I think if we do `auto header = toSlice(serializeThreadLocal(DbValueHeader{...});` we might avoid an allocation.

andrewjstone: Will do! I actually came around on serializeThreadLocal even more after you convinced me it was safe by reading the RocksDB code when adding support for SliceParts. I realized all the data in a batch is copied into a single vector.

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: Fixed.

kvbc/test/categorization/block_merkle_category_unit_test.cpp
andrewjstone: Fixed.

PR #1174, opened 2021-01-28
Disallow deleting the only block as last reachable

dartdart26:
If we allow deletion of the only block as last reachable during replica
state sync, the replica will try to fetch all blocks from 1 to N, but
will always fail if some have been pruned (since we start pruning from
block 1).

The behavior in this PR assumes:
 * the last executed sequence number in the blockchain and the BFT
   metadata last executed sequence number differ by at most 1
 * when pruning, we will always leave at least 2 blocks in the
   blockchain

A corner case is if replica state sync needs to run when only one block
has been added. Will be handled in a separate commit or ignored as
an unlikely single startup event.

A final note is that deleting the only block as genesis (during pruning)
is also not allowed.

This PR is not compatible with https://github.com/vmware/concord-bft/pull/1172 - need to pick one of the two
or devise another strategy.

### DIFF
--- a/kvbc/src/categorization/kv_blockchain.cpp
+++ b/kvbc/src/categorization/kv_blockchain.cpp
@@ -14,6 +14,8 @@
 #include "categorization/kv_blockchain.h"
 #include "bcstatetransfer/SimpleBCStateTransfer.hpp"
 
+#include <stdexcept>
+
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
LGTM. Thanks Petar!

PR #1175, opened 2021-01-28
Add multi-send method to TLS layer to avoid excess copies in bft client

tdimitrov:
This patch modifies ICommunication interface for two main reasons:
- Avoid unnecessary memory copying when the bft client sends messages
to all replicas. In the current implementation the message is copied for
each replica instance.
- Remove unnecessary use of raw pointers in the code.

How this is achieved:

Add new method to ICommunication - multiSendMessage(). It receives a raw
message buffer in form of std::vector<uint8_t> and takes ownership of
it. Internally the vector is converted to shared_ptr and it is shared
between all TLS connections. This approach avoids copying the same
message multiple times.

sendAsyncMessage() from ICommunication also accepts rvalue ref to
std::vector<uint8_t>. For consistency it is internally converted to
shared_ptr.

These changes lead to some additional modifications:
1. struct OutgoingMsg accepts vector instead of raw pointer. A copy of
the vector is made, because each message consists of LEN + PAYLOAD. To
construct the correct message structure a new chunk of memory needs to
be allocated.

2. The implementation of TlsTCPCommunication is changed. Now OutgoingMsg
is created from the raw buffer passed by the upper layer. This OutgoingMsg
is passed to the implementation. The benefit is that memory for the
message is allocated only once.

3. WriteQueue works with shared_ptr<OutgoingMsg> instead of OutgoingMsg
directly. Additionally OutgoingMsg is not created in the WriteQueue
anymore. The message is created in TlsTcpImpl send functions.

4. Add implementations for sendAsyncMessage and multiSendMessage to
PlainUDPCommunication and FakeCommunication.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -11,6 +11,9 @@ jobs:
                 - "CONCORD_BFT_CONTAINER_CC=clang CONCORD_BFT_CONTAINER_CXX=clang++"
             ci_build_type:
                 - "-DCMAKE_BUILD_TYPE=DEBUG -DCI_TEST_STORAGE_TYPE=v2merkle"
+            comm_type:
+                - "-DBUILD_COMM_TCP_PLAIN=FALSE -DBUILD_COMM_TCP_TLS=FALSE" #udp
+                - "-DBUILD_COMM_TCP_PLAIN=FALSE -DBUILD_COMM_TCP_TLS=TRUE" #tls
     steps:
         - name: Checkout
           uses: actions/checkout@v2
@@ -23,8 +26,7 @@ jobs:
                             CONCORD_BFT_CMAKE_FLAGS=\"\
                             ${{ matrix.ci_build_type }} \
                             -DBUILD_TESTING=ON \
-                            -DBUILD_COMM_TCP_PLAIN=FALSE \
-                            -DBUILD_COMM_TCP_TLS=FALSE \
+                            ${{ matrix.comm_type }} \
                             -DCMAKE_CXX_FLAGS_RELEASE=-O3 -g \
                             -DUSE_LOG4CPP=TRUE \
                             -DBUILD_ROCKSDB_STORAGE=TRUE \
--- a/communication/include/communication/ICommunication.hpp
+++ b/communication/include/communication/ICommunication.hpp
@@ -13,6 +13,8 @@
 #pragma once
 
 #include <cstdint>
+#include <set>
+#include <vector>
 
 namespace bft::communication {
 
@@ -25,7 +27,7 @@ class IReceiver {
   // Invoked when a new message is received
   // Notice that the memory pointed by message may be freed immediately
   // after the execution of this method.
-  virtual void onNewMessage(NodeNum sourceNode, const char *const message, size_t messageLength) = 0;
+  virtual void onNewMessage(NodeNum sourceNode, const char* const message, size_t messageLength) = 0;
 
   // Invoked when the known status of a connection is changed.
   // For each NodeNum, this method will never be concurrently
@@ -53,9 +55,11 @@ class ICommunication {
   // Sends a message on the underlying communication layer to a given
   // destination node. Asynchronous (non-blocking) method.
   // Returns 0 on success.
-  virtual int sendAsyncMessage(NodeNum destNode, const char *const message, size_t messageLength) = 0;
+  virtual int sendAsyncMessage(NodeNum destNode, std::vector<uint8_t>&& msg) = 0;
 
-  virtual void setReceiver(NodeNum receiverNum, IReceiver *receiver) = 0;
+  virtual std::set<NodeNum> multiSendMessage(const std::set<NodeNum> dests, std::vector<uint8_t>&& msg) = 0;
+
+  virtual void setReceiver(NodeNum receiverNum, IReceiver* receiver) = 0;
 
   virtual ~ICommunication() = default;
 };
--- a/communication/include/communication/ICommunication.hpp
+++ b/communication/include/communication/ICommunication.hpp
@@ -54,10 +54,16 @@ class ICommunication {
 
   // Sends a message on the underlying communication layer to a given
   // destination node. Asynchronous (non-blocking) method.
+  // The function takes ownership of the buffer provided.
   // Returns 0 on success.
-  virtual int sendAsyncMessage(NodeNum destNode, std::vector<uint8_t>&& msg) = 0;
+  virtual int send(NodeNum destNode, std::vector<uint8_t>&& msg) = 0;
 
-  virtual std::set<NodeNum> multiSendMessage(const std::set<NodeNum> dests, std::vector<uint8_t>&& msg) = 0;
+  // Sends a message to all nodes in dests set.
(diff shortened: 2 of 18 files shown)


communication/src/TlsWriteQueue.h
```diff
@@ -36,14 +36,16 @@ static constexpr size_t MAX_QUEUE_SIZE_IN_BYTES = 1024 * 1024 * 1024;  // 1 GB
...
 struct OutgoingMsg {
-  OutgoingMsg(const char* raw_msg, const size_t len)
-      : msg(len + MSG_HEADER_SIZE), send_time(std::chrono::steady_clock::now()) {
-    uint32_t msg_size = htonl(static_cast<uint32_t>(len));
+  OutgoingMsg(std::vector<uint8_t>&& raw_msg)
+      : msg(raw_msg.size() + MSG_HEADER_SIZE), send_time(std::chrono::steady_clock::now()) {
+    uint32_t msg_size = htonl(static_cast<uint32_t>(raw_msg.size()));
     std::memcpy(msg.data(), &msg_size, MSG_HEADER_SIZE);
-    std::memcpy(msg.data() + MSG_HEADER_SIZE, raw_msg, len);
+    std::memcpy(msg.data() + MSG_HEADER_SIZE, raw_msg.data(), raw_msg.size());
```
andrewjstone, line 43: Just a note here for posterity, and because it's mentioned in the commit message as well: We can get rid of this allocation and copy by using a [ConstBufferSequence](https://think-async.com/Asio/boost_asio_1_18_1/doc/html/boost_asio/reference/ConstBufferSequence.html) in asio. This is similar to the `writev` syscall and allows writing consecutively from multiple buffers. This should be done in a follow up PR though. No need to do it now.

andrewjstone (commented):
This looks great @tdimitrov! Just like we discussed. 

My only comment is since we are changing the` ICommunication` API we may as well shorten the names of the function as well. I personally prefer just calling them both `send` and overloading. But `send` and `multiSend` also works.

communication/src/PlainUDPCommunication.cpp
```diff
@@ -251,25 +251,24 @@ class PlainUDPCommunication::PlainUdpImpl {
...
                           << ntohs(to->sin_port));
 
-    error = sendto(udpSockFd, message, messageLength, 0, (struct sockaddr *)to, sizeof(Addr));
+    error = sendto(udpSockFd, msg->data(), msg->size(), 0, (struct sockaddr *)to, sizeof(Addr));
 
     if (error < 0) {
       /** -1 return value means underlying socket error. */
       LOG_INFO(_logger, "Error while sending: " << concordUtils::errnoString(errno));
-    } else if (error < (int)messageLength) {
+    } else if (error < (int)msg->size()) {
```
f-squirrel, line 265: I know, this is not a new code, but would you mind to cast to `ssize_t`.

communication/src/TlsTcpImpl.cpp
```diff
@@ -133,16 +133,16 @@ void TlsTCPCommunication::TlsTcpImpl::setReceiver(NodeNum _, IReceiver* receiver
   receiver_ = receiver;
 }
 
-int TlsTCPCommunication::TlsTcpImpl::sendAsyncMessage(const NodeNum destination, const char* msg, const size_t len) {
+int TlsTCPCommunication::TlsTcpImpl::sendAsyncMessage(const NodeNum destination, std::shared_ptr<OutgoingMsg> msg) {
```
f-squirrel, line 134: I think that `const std::shared_ptr<OutgoingMsg>& msg` is more appropriate.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -384,7 +382,7 @@ bool AsyncTlsConnection::verifyCertificateServer(bool preverified, boost::asio::
 
 std::pair<bool, NodeNum> AsyncTlsConnection::checkCertificate(X509* receivedCert,
                                                               std::string connectionType,
-                                                              std::string subject,
+                                                              const std::string& subject,
```
f-squirrel, line 385: I guess the changes in the CPP code are related to the updated in the clang, right?

f-squirrel:
@tdimitrov , it looks great, IMO it is a very elegant code, and thank you for extending the static analysis!

communication/src/PlainUDPCommunication.cpp
tdimitrov: Actually the size() returns size_t, but error is int. Without the cast the compilation fails:
`
/home/tdimitrov/concord-bft/communication/src/PlainUDPCommunication.cpp:265:22: error: comparison of integers of different signs: 'int' and 'std::vector::size_type' (aka 'unsigned long')
      [-Werror,-Wsign-compare]
    } else if (error < msg->size()) {
               ~~~~~ ^ ~~~~~~~~~~~
`

communication/src/TlsTcpImpl.cpp
tdimitrov: I think you mean std::shared_ptr<const OutgoingMsg>?

The idea here is that we take ownership of the memory in the interface (ICommunication). The buffer there is passed as rvalue reference and for the implementation we convert it to shared_ptr, so that we can use it for all sends.
Since we own this memory I think making it const is unnecessary limitation.

communication/src/AsyncTlsConnection.cpp
tdimitrov: Yes this change is make clang-tidy happy :)

communication/include/communication/ICommunication.hpp
```diff
@@ -52,10 +54,15 @@ class ICommunication {
...
   // Sends a message on the underlying communication layer to a given
   // destination node. Asynchronous (non-blocking) method.
+  // The function takes ownership of the buffer provided.
   // Returns 0 on success.
-  virtual int sendAsyncMessage(NodeNum destNode, const char *const message, size_t messageLength) = 0;
+  virtual int send(NodeNum destNode, std::vector<uint8_t>&& msg) = 0;
 
-  virtual void setReceiver(NodeNum receiverNum, IReceiver *receiver) = 0;
+  // Sends a message to all nodes in dests set.
+  // The function takes ownership of the buffer provided.
```
andrewjstone, line 62: Can you please add another comment that explains the return value?

andrewjstone (approved):
Looks great @tdimitrov!

.github/workflows/clang-tidy.yml
```diff
@@ -11,6 +11,10 @@ jobs:
                 - "CONCORD_BFT_CONTAINER_CC=clang CONCORD_BFT_CONTAINER_CXX=clang++"
             ci_build_type:
                 - "-DCMAKE_BUILD_TYPE=DEBUG -DCI_TEST_STORAGE_TYPE=v2merkle"
+            comm_type:
+                - "-DBUILD_COMM_TCP_PLAIN=FALSE -DBUILD_COMM_TCP_TLS=FALSE" #udp
+                #- "-DBUILD_COMM_TCP_PLAIN=TRUE -DBUILD_COMM_TCP_TLS=FALSE" #tcp
```
andrewjstone, line 16: We should jsut remove this line instead of commenting it out.

teoparvanov (approved):
LGTM!

PR #1176, opened 2021-01-29
Use RocksDB SliceParts for all category values

dartdart26:
Eliminate value copying during serialization of DbValue and
ImmutableDbValue CMF types by using RocksDB slice parts for the
VersionedKeyValueCategory and the ImmutableKeyValueCategory.

Idea borrowed from #1173 by @andrewjstone.


andrewjstone (approved):
Looks great. Thanks!

PR #1177, opened 2021-01-29
Fix for stuck inactive View after State Transfer.

HristoStaykov:
The fix covers the situation we get when we are 1 Checkpoint
(currently 150 SeqNo-s) behind our peers and our View is also behind.
In this case it happens that we receive all necessary View Change messages
and the New View message from the Primary, and being behind we are unable to
activate the View and therefore we initiate State Transfer to get up to date.
After State Transfer completes there is nothing to trigger us trying to
activate the View again, because we previously gathered all, so now after ST
completes we check if our View is inactive and try to enter it.


andrewjstone (approved):
LGTM! Thanks for the fix and great explanation @HristoStaykov!

teoparvanov (approved):
LGTM, thanks Hristo!

PR #1178, opened 2021-01-29
Apollo test replica misses 2 view changes

HristoStaykov:
Added new View Change Apollo test.
In this test we set things up to get a Replica 2 Views behind
its peers and verify it correctly manages to enter the View
and participate in quorum for execution and the system reaches
Fast Path.


teoparvanov (approved):
LGTM!

PR #1179, opened 2021-01-30
Optimize BlockMerkleCategory::deleteGenesisBlock

andrewjstone:
Before this change we sometimes created two new merkle tree versions for
a single deleted block. With this change we now always only create 1.

Also, enhance the tests for `deleteGenesisBlock`. We explicitly check
that the stale leaf nodes and actual leaf nodes are what we expect after
each call to `add` and `deleteGenesisBlock`.

Lastly, add `deleteGenesisBlock` to the test fixture to reduce
boilerplate.

### DIFF
--- a/kvbc/test/categorization/block_merkle_category_unit_test.cpp
+++ b/kvbc/test/categorization/block_merkle_category_unit_test.cpp
@@ -23,6 +23,7 @@
 #include <memory>
 #include <optional>
 #include <ostream>
+#include <stdexcept>
 #include <string>
 #include <utility>
 #include <variant>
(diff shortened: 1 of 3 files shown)


kvbc/test/categorization/block_merkle_category_unit_test.cpp
```diff
@@ -36,6 +37,103 @@ using namespace concord::kvbc::categorization;
...
+
+StaleKeys getStale(std::shared_ptr<concord::storage::rocksdb::NativeClient> &db, TreeVersion version) {
+  auto ser_stale = db->get(BLOCK_MERKLE_STALE_CF, serialize(version));
+  auto stale = StaleKeys{};
+  deserialize(*ser_stale, stale);
+  return stale;
+}
+
+// Keys returned by a sparse_merkle::UpdateBatch are always hashed versions of string keys. However,
+// the keys in the block merkle are simply BlockIds. We therefore want to brute force reverse the
```
dartdart26, line 90: Nit: reverse -> traverse?

dartdart26: Actually, now I get it :D

dartdart26 (approved):

PR #1180, opened 2021-02-01
Fix for concurrent access to AsyncTimeRecorderMap in PreProcessor.

HristoStaykov:
We no longer use a map to store the measurements. Instead we directly
move them in the AsyncPreProcessJob, and rely on the lifetime of this
job to finalize the measurement. Since we have a 1:1 mapping of Client
Requests to AsyncPreProcessJob-s and we delete the jobs after we execute
them this has to be equivalent.


yuliasherman (approved):
LGTM

yontyon (approved):
LGTM

andrewjstone (approved):
LGTM. Thanks for the fix @HristoStaykov

KTeofanidis:
LGTM

PR #1182, opened 2021-02-02
Fix for crash during View Change.

HristoStaykov:
We had situations where the Primary was asserting in the check
whether it can send PrePrepares. If the preconditions for sending
PrePrepares are not met we can return false and not take any action.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -280,14 +280,13 @@ void ReplicaImp::onMessage<ReplicaAsksToLeaveViewMsg>(ReplicaAsksToLeaveViewMsg 
 }
 
 bool ReplicaImp::checkSendPrePrepareMsgPrerequisites() {
-  if (!isCurrentPrimary()) {
-    LOG_WARN(GL, "Calling checkSendPrePrepareMsgPrerequisites in non primary replica!");
-    return false;
-  }
+  ConcordAssert(isCurrentPrimary());
+
   if (!currentViewIsActive()) {
     LOG_INFO(GL, "View " << getCurrentView() << " is not active yet. Won't send PrePrepareMsg-s.");
     return false;
   }
+
   if (isSeqNumToStopAt(lastExecutedSeqNum)) {
     LOG_INFO(GL,
              "Not sending PrePrepareMsg because system is stopped at checkpoint pending control state operation "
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -280,16 +280,21 @@ void ReplicaImp::onMessage<ReplicaAsksToLeaveViewMsg>(ReplicaAsksToLeaveViewMsg
 }
 
 bool ReplicaImp::checkSendPrePrepareMsgPrerequisites() {
+  if (!isCurrentPrimary()) {
+    LOG_WARN(GL, "Calling checkSendPrePrepareMsgPrerequisites in non primary replica!");
```
yuliasherman, line 284: -> "Called in a non-primary replica; won't send PrePrepareMsgs"

yuliasherman (approved):
LGTM

bftengine/src/bftengine/ReplicaImp.cpp
teoparvanov, line 284: Let's make this an assert. I think the statement "a non-primary will never send a PrePrepare" is a system invariant.

teoparvanov (approved):
LGTM, thanks @HristoStaykov !

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -280,14 +280,13 @@ void ReplicaImp::onMessage<ReplicaAsksToLeaveViewMsg>(ReplicaAsksToLeaveViewMsg
 }
 
 bool ReplicaImp::checkSendPrePrepareMsgPrerequisites() {
-  if (!isCurrentPrimary()) {
-    LOG_WARN(GL, "Calling checkSendPrePrepareMsgPrerequisites in non primary replica!");
-    return false;
-  }
+  ConcordAssert(isCurrentPrimary());
```
yuliasherman, line 283: I cannot agree with leaving this as an assertion. There could be a race between PreProcessor that still thinks it is a primary and ReplicaImp that have changed a primary a moment later. To be on the safe side we need to just ignore the request.

teoparvanov: but Yulia, in what case would a non-primary send a pre-prepare?

yuliasherman: I can explain to you offline

teoparvanov: ok @HristoStaykov , sorry for confusing you... Let's replace the assert with an if statement, as you had done originally. I'll talk to Yulia offline to better understand the reasoning behind this.

teoparvanov (approved):
Looks great, thanks!

PR #1183, opened 2021-02-02
Send noops in fast path

yontyon:
Currently, we send Noops in a slow path (by default).
There is no real reason for this.
In this PR we change the default to fast path


Liorgal28 (approved):
LGTM

PR #1184, opened 2021-02-03
Change ControlStateManager to be a singleton

toly-kournik:


bftengine/src/bftengine/ReplicaBase.hpp
```diff
@@ -126,8 +121,7 @@ class ReplicaBase {
 
   ///////////////////////////////////////////////////
   // ControlStateManger
-  std::shared_ptr<ControlStateManager> controlStateManager_;
-  bool stopAtNextCheckpoint_ = false;
+  bool stopAtNextCheckpoint_ = false;  // TODO [TK] remove, use ControlStateManager's state
```
yontyon, line 124: I think we already have a cache variable in ControlStateMnager, so we can remove this variable and its uses

kvbc/include/KVBCInterfaces.h
```diff
@@ -121,7 +121,6 @@ class ICommandsHandler : public bftEngine::IRequestsHandler {
   void execute(ExecutionRequestsQueue& requestList,
                const std::string& batchCid,
                concordUtils::SpanWrapper& parent_span) override = 0;
-  virtual void setControlStateManager(std::shared_ptr<bftEngine::ControlStateManager> controlStateManager) = 0;
```
yontyon, line 124: Notice that this breaks API in concord also

yontyon (approved):
LGTM
Only 2 minor comments

kvbc/include/KVBCInterfaces.h
toly-kournik: Yes, we'll do it during bft version bump

bftengine/src/bftengine/ReplicaBase.hpp
toly-kournik: Removed, just wanted your feedback

PR #1185, opened 2021-02-03
Fix preexecution tests in cases where replicas aren't reset between individual tests

KTeofanidis:
The tests can fail if the preexecution metrics used don't start at 0 for each new test.

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -1133,15 +1133,24 @@ class BftTestNetwork:
                 while pre_proc_req < num_requests or \
                         total_pre_exec_requests_executed != num_requests:
                     key1 = ["preProcessor", "Counters", "preProcReqSentForFurtherProcessing"]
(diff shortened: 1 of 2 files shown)


tests/apollo/util/bft.py
```diff
@@ -1133,15 +1133,24 @@ async def assert_successful_pre_executions_count(self, replica_id, num_requests)
...
                     await trio.sleep(0.1)
         except trio.TooSlowError:
             assert False, "Preprocessor requests " + \
                       f'(expected={num_requests} ' \
                       f'executed={total_pre_exec_requests_executed} ' \
                       f'preexecsent={pre_proc_req})'
+        except AttributeError:
+            assert False, "Preprocessor counter not initialized"
+
+    async def init_preexec_count(self, replica_id):
```
teoparvanov, line 1148: let's make `replica_id=0` the default, to avoid specifying it every time

teoparvanov (approved):
LGTM, thanks @KTeofanidis !

PR #1188, opened 2021-02-04
Beginnings of KVBC storage benchmarking tool

andrewjstone:
This tool is intended to allow benchmarking and tuning the new
kvbc_blockchain code. We generate a bunch of random data up front, and
then add blocks containing that data. To simulate pre-execution we also
perform conflict detection before each block add.

This very early start of the implementation does not use the actual
kvbc_blockchain API yet. It instead puts all keys into the BlockMerkle
category. The only reasons it was done this way for ease of getting the
scaffolding for the tool in place and because the full kvbc_blockchain
code is not merged in yet on master.

The code currently:
 * Generates random BlockMerkleInput up to a configurable memory limit
 * Performs simulated conflict detection checks on a configurable
   subset of keys added in the prior block so that we know the keys exist.
 * Provides tunables for number of blocks, batch size, etc...
 * Adds a histogram for the time taken to add a block
 * Runs a diagnostic server so stats can be retreived with concord-ctl
   during runs

There's still a lot to do, including:
 * Using the real kvbc_blockchain interface
 * Adding support for concurrent reads for the duration of block adds to
   simulate pre-execution
 * Allowing disabling of pre-execution simulation, and inlining reads
   into post-execution.
 * More complex random input generation. Instead of reusing keys,
   generate input concurrently in a separate thread and use them when the
   current thread adding blocks runs out.
 * Ability to dump all histograms to a file at the end of the run.


andrewjstone:
Note that this PR includes the code for #1187, since it is not merged in yet, and without it a build wouldn't succeed.

andrewjstone:
Closing this as nobody has reviewed it and the code has changed quite a bit in the last few days. I'm going to open a new one targeting the integration branch today or tomorrow.

PR #1190, opened 2021-02-05
Add sequence number to key exchange logs

nkgeorgiev:
The motivation is to be able to causally order key-exchange events by sequence number


teoparvanov (approved):
LGTM, but to be sure, let's wait for Eran's review too (CC: @cloudnoize )

Thanks!
-- Teo

PR #1191, opened 2021-02-08
SKVBC startup deterministic check

bhavsi:


tests/apollo/util/skvbc.py
```diff
@@ -221,6 +221,10 @@ async def assert_kv_write_executed(self, key, val):
             assert {key: val} == kv_reply, \
                 f'Could not read original key-value in the case of n={config.n}, f={config.f}, c={config.c}.'
 
+    async def wait_for_liveness(self):
+        key, value = await self.write_known_kv()
```
teoparvanov, line 225: Hi Harsh, I think we also need to add retries to this, in case the system takes a bit longer to start.
you can check some of the existing wait_for_X() methods in bft.py for inspiration :)

PR #1192, opened 2021-02-08
SKVBC startup deterministic check

bhavsi:
Currently, we don't have a deterministic way to know whether the blockchain can process requests (are at least n-f replicas started, did key exchange complete, are their thread pools ready to process requests?). We use hardcoded timeouts, e.g. in the pre-execution tests, before starting the actual test.

To make tests more robust and deterministic, this PR introduces `wait_for_liveness()` method in `SimpleKVBCProtocol` and `SkvbcTracker` and replaces the non-deterministic timeout waits with deterministic checks.

### DIFF
--- a/tests/apollo/util/skvbc.py
+++ b/tests/apollo/util/skvbc.py
@@ -221,6 +221,20 @@ class SimpleKVBCProtocol:
             assert {key: val} == kv_reply, \
                 f'Could not read original key-value in the case of n={config.n}, f={config.f}, c={config.c}.'
 
+    async def wait_for_liveness(self):
+        with trio.fail_after(seconds=30):
+            while True:
+                with trio.move_on_after(seconds=5):
+                    try:
+                        key, value = await self.write_known_kv()
+                        await self.assert_kv_write_executed(key, value)
+                    except (trio.TooSlowError, AssertionError) as e:
+                        pass
+                    else:
+                        # success
+                        return
+                    await trio.sleep(0.1)
+
     async def prime_for_state_transfer(
             self, stale_nodes,
             checkpoints_num=2,
--- a/tests/apollo/util/skvbc_history_tracker.py
+++ b/tests/apollo/util/skvbc_history_tracker.py
@@ -1039,6 +1039,30 @@ class SkvbcTracker:
             self.status.record_client_timeout(client_id)
             return
 
+    async def wait_for_liveness(self):
+        with trio.fail_after(seconds=30):
+            while True:
+                with trio.move_on_after(seconds=5):
+                    try:
+                        config = self.bft_network.config
+                        # Write a KV pair with a known value
+                        known_key = self.skvbc.unique_random_key()
+                        known_val = self.skvbc.random_value()
+                        known_kv = [(known_key, known_val)]
+                        client = self.bft_network.random_client()
+                        await self.write_and_track_known_kv(known_kv, client)
+                        kv_reply = await self.read_and_track_known_kv(known_key, client)
+                        print("READ REPLY: {}".format(kv_reply))
+                        assert {known_key: known_val} == kv_reply, \
+                            f'Expected key-value: {known_kv}; Actual key-value: {kv_reply} - ' \
+                            f'in the case of n={config.n}, f={config.f}, c={config.c}.'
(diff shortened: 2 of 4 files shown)


tests/apollo/util/skvbc.py
```diff
@@ -221,6 +221,10 @@ async def assert_kv_write_executed(self, key, val):
             assert {key: val} == kv_reply, \
                 f'Could not read original key-value in the case of n={config.n}, f={config.f}, c={config.c}.'
 
+    async def wait_for_liveness(self):
+        key, value = await self.write_known_kv()
+        await self.assert_kv_write_executed(key, value)
```
teoparvanov, line 226: To make this deterministic, you'll need to add retries. For example, if the system is not yet ready to process requests, the `await self.write_known_kv()` is going to fail.
See an example of how we can add a retrying `while True` loop here:
https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft.py#L972

teoparvanov (commented):
Hi @bhavsi , in addition to my comment above, please also add a detailed description, so that the reviewers understand the purpose of the PR.

tests/apollo/util/skvbc.py
bhavsi: Thanks will add this!

teoparvanov:
Hi @KTeofanidis , could you please review and take a look at the intermittent failures? Thanks!

teoparvanov:
@bhavsi , please add a description to the PR - currently your intention is unclear for an external reviewer. Thank you.

KTeofanidis:
The failures are caused by the fact that the metrics aren't updated in real time. If you call init_preexec_count right after executing requests through preexecution the metrics won't be updated yet so the total count will be off. When running tests locally I can see it's generally either 1 or 2, but it has to be accounted for exactly when recording the initial state.

teoparvanov:
> 
> 
> The failures are caused by the fact that the metrics aren't updated in real time. If you call init_preexec_count right after executing requests through preexecution the metrics won't be updated yet so the total count will be off. When running the tests locally I can see it's generally either 1 or 2, but it has to be accounted for exactly when recording the initial state.

@KTeofanidis what is your recommendation to solve this?

KTeofanidis:
First the timeout of .5 seconds is too short and will cause us to almost always send more than one request. It should be 5 seconds. Second since we have no way of determining exactly how many requests were executed and updated in the metrics when init_preexec_count is called, we can't count exact numbers. A change to assert_successful_pre_executions_count will be needed so "total_pre_exec_requests_executed != num_requests" is changed to "total_pre_exec_requests_executed **<** num_requests". We want to wait until the total executions exceed the number of writes in the specific test since they will also contain at least 1 or potentially more requests from the wait_for_liveness phase that may or may not be updated in the metrics at the start.

tests/apollo/util/skvbc.py
```diff
@@ -221,6 +221,20 @@ async def assert_kv_write_executed(self, key, val):
             assert {key: val} == kv_reply, \
                 f'Could not read original key-value in the case of n={config.n}, f={config.f}, c={config.c}.'
 
+    async def wait_for_liveness(self):
+        with trio.fail_after(seconds=30):
+            while True:
+                with trio.move_on_after(seconds=.5):
```
teoparvanov, line 227: As suggested by @KTeofanidis , it should be `seconds=5` here (otherwise it is very likely to interrupt the await for the client write before it has a chance to complete).

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -1039,6 +1039,30 @@ async def read_and_track_known_kv(self, key, client):
             self.status.record_client_timeout(client_id)
             return
 
+    async def wait_for_liveness(self):
+        with trio.fail_after(seconds=30):
+            while True:
+                with trio.move_on_after(seconds=.5):
```
teoparvanov, line 1045: same here, `seconds=5`

bhavsi:
Works well now, thanks @teoparvanov and @KTeofanidis !

teoparvanov (approved):
LGTM, thanks @bhavsi !

PR #1194, opened 2021-02-09
Enhance Replica Status for concord-ctl

saiakhil2012:
Introducing following new members in the replica status for `concord-ctl`,
- Number of Invalid Clients
- Number of Valid No Ops
- Last Stable Checkpoint


bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -104,6 +104,8 @@ void ClientsManager::init(IStateTransfer* stateTransfer) {
 
 uint32_t ClientsManager::numberOfRequiredReservedPages() const { return requiredNumberOfPages_; }
 
+std::set<NodeIdType> ClientsManager::getInvalidClients() { return invalidClients; }
```
andrewjstone, line 107: We don't need to keep track of invalid clients.  In fact, this might actually be an attack vector, as we would grow memory for each invalid client and we never clear it. We only need a global counter like for noops below. Then we will be able to monitor it for changes.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -176,6 +176,10 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
   const bool invalidClient = !isValidClient(clientId);
   const bool sentFromReplicaToNonPrimary = repsInfo->isIdOfReplica(senderId) && !isCurrentPrimary();
 
+  if (invalidClient) {
+    clientsManager->getInvalidClients().insert(clientId);
```
andrewjstone, line 180: I would just add to a counter member variable as done for noops. No need to maintain a set. The goal is to count how frequently invalid clients arrive to track possible bugs or attacks, not to know what the IDs are. We could always log them if we wanted to.

andrewjstone (changes requested):
Looks great @saiakhil2012 ! It's just a bit overkill. Please just change the invalid client tracking to a counter and I think we are good to go.

andrewjstone (approved):
Looks great. Thanks for the contribution @saiakhil2012 !

bftengine/src/bftengine/ClientsManager.cpp
saiakhil2012: Yeah, I added a set to have the count unique. But yeah, agree that it could be dangerous in the long run, will keep in mind of such aspects for future PRs. Thanks. 

Changed it to using a numerical counter

bftengine/src/bftengine/ReplicaImp.cpp
saiakhil2012: Yes, changed it to numerical counter. If we need actual IDs, we can log them later. Thanks

saiakhil2012:
Thanks @andrewjstone !

teoparvanov (approved):
LGTM, thanks @saiakhil2012 !

PR #1196, opened 2021-02-10
Beginnings of KVBC storage benchmarking tool

andrewjstone:
This tool is intended to allow benchmarking and tuning the new
`kvbc_blockchain` code. We generate a bunch of random data up front, and
then add blocks containing that data. To simulate pre-execution we also
perform conflict detection before each block addition. Right now
conflict detection is performed via a single multiget, since it is
possible to batch all conflict reads by taking the union of the keys in
each transaction in a batch. This is easily changed as needed.

This preliminary implementation does not use the actual
`kvbc_blockchain` API yet. It instead puts all keys into the
`BlockMerkle` category. The implementation was done this way for ease
of getting the scaffolding for the tool in place and because the full
`kvbc_blockchain` code is not merged in yet on master.

The code currently:
 * Generates random `BlockMerkleInput` up to a configurable memory limit
 * Generates a configurable set of read keys for use in pre-execution
 and conflict detection
 * Performs simulated conflict detection checks
 * Provides tunables for number of blocks, batch size, etc...
 * Adds a histogram for the time taken to add a block, and one for
 conflict detection
 * Runs a diagnostic server so stats can be retreived with concord-ctl
   during benchmark runs
 * Dumps the rocksdb counters at the end of a run

There's still more to do, including:
 * Using the real `kvbc_blockchain` interface
 * Adding support for concurrent reads for the duration of block adds to
   simulate pre-execution
 * Allowing disabling of pre-execution simulation, and inlining reads
   into post-execution.

This commit also includes a significant change to the RocksDB
`NativeClient`, such that configuration of the database can be performed
largely via a rocksdb configuration file. This is a change to the
`UserOptions` mechanism, which now takes a filepath for a RocksDB
configuration file, as well as a callback to run to apply any
configuration that cannnot be performed by a RocksDB initialization
file. This latter configuration is applied **after** the configuration
file is loaded. A helper function, `openRocksDB`, was added to the
`IDBClient` based `rocksdb_client` where shared code is possible among
file based and non-file based configuration.

`kvbcbench` runs using the included configuration file, and further
configuration is applied via the `completeRocksdbConfiguration` callback
in `main.cpp`. A 4GB LRU block cache is shared among all column families
for simplicity, and as a starting point for tuning. Additionally, each
column family has a bloom filter, and the `block_merkle_key_versions`
column family is optimized for point lookup, since we will never do
range scans on it. All of this configuration is only possible in code,
and so it is performed in the callback.

Only column families for the `BlockMerkleCategory` are
included in the configuration file since `kvbcbench` only uses that
category currently. Column families are created if they do not already
exist. As the `block_merkle_keys` column family contains the bulk of the
data, it is tuned to have a 512MB
write buffer, which creates an 8GB level 0 and level
1(`max_bytes_for_level_base`) based on the
multipliers: `level0_file_num_compaction_trigger` and
`min_write_buffer_number_to_merge`. Additionally, we limit files to 10 in
level 0 and level 1 with a `target_file_size_base` 1/10 the size
of `max_bytes_for_level_base`. We also cache index and filter blocks and
pin them in memory. We do this for the key versions as well, as we want
reads optimized.

It should be noted, that this configuration is *only* used by the
benchmarking tool, and is a starting point for further experiments.

More information about why these settings were changed can be found in the [RocksDB
Tuning
Guide](https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide).


storage/include/rocksdb/native_client.ipp
```diff
@@ -62,7 +62,7 @@ inline NativeClient::NativeClient(const std::string &path, bool readOnly, const
     : client_{std::make_shared<Client>(path)} {
   auto options = Client::Options{};
```
dartdart26, line 63: Nit: I guess we can remove the `options` variable.

storage/src/rocksdb_client.cpp
```diff
@@ -114,6 +117,82 @@ void Client::Options::applyOptimizations() {
...
+    throw std::invalid_argument{msg};
+  }
+  // Add specific global options
+  db_options.IncreaseParallelism(static_cast<int>(std::thread::hardware_concurrency()));
+  db_options.sst_file_manager.reset(::rocksdb::NewSstFileManager(::rocksdb::Env::Default()));
+  db_options.statistics = ::rocksdb::CreateDBStatistics();
+  db_options.statistics->set_stats_level(::rocksdb::StatsLevel::kExceptTimeForMutex);
+
+  // Some options, notably pointers, are not configurable via the config file. We set them in code here.
+  user_options.completeInit(db_options, cf_descs);
```
dartdart26, line 140: Great idea!

kvbc/benchmark/kvbcbench/main.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+  // Each thread fills in their subset of keys in a pre-allocated vector. There is no race here, and
+  // no need for locks as the buffer segments in each thread are disjoint.
+  auto total_read_keys = std::min(num_adds * num_blocks, max_read_keys);
+  auto read_keys_per_thread = total_read_keys / num_cpus;
+  auto read_keys = ReadKeys(total_read_keys);
+
+  auto gen = [&](size_t thread_id) -> std::vector<categorization::BlockMerkleInput> {
+    static thread_local std::mt19937 generator;
+    auto distribution = std::uniform_int_distribution<uint8_t>(1, 255);
```
dartdart26, line 151: Based on https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution, it seems if the template parameter is `uint8_t`, result is undefined. Maybe we should use `uint16_t` and cast to `uint8_t`.

```

IntType | - | The result type generated by the generator. The effect is undefined if this is not one
of short, int, long, long long, unsigned short, unsigned int, unsigned long, or unsigned long long
```

dartdart26: One more thing - why is 0 excluded?

dartdart26 (approved):

One minor thing I posted is the `std::uniform_int_distribution` template parameter which I think we might want to fix.

kvbc/benchmark/kvbcbench/main.cpp
andrewjstone: Great catch on both! I don't recall why I didn't include 0. I know in some cases it was to prevent divide by zero, but I don't think that's relevant here.

Issue #1195, opened 2021-02-09
Encountered TooSlowError in pyclient_tests_udp as part of CI Checks 


saiakhil2012:
Encountered the `TooSlowError` error in the test, `pyclient_tests_udp` in `Build and Test` check of CI checks for the PR
Link to run where it failed - https://github.com/vmware/concord-bft/pull/1194/checks?check_run_id=1865803092

Steps to reproduce the behavior: Most probably this is happening because of slow startup and there was nothing in the changes introduced by PR to make it slow, so it would be a transient error

Proposed Solution: Increase in the timeout, so that new PRs would not face this problem

andrewjstone:
@glevkovich @yontyon We should discuss this. I'm actually surprised it doesn't fail more frequently. In this case the primary comes up only 43ms before the test fails. I think the line where the failure is occuring: ` await udp_client.sendSync(self.writeRequest(1), False)` is essentially just a check to see that the cluster is up and running. This should have a longer timeout for the client request, that we can wrap in a helper  function. Then it should be just 1 second for the` MofN` request, which is the actual target of the test. 

We can reuse the wrapper function with longer request in other tests such as `_testPrimaryWrite` also. 

Does this make sense to you guys?

glevkovich:
@andrewjstone , this happened this week to @Liorgal28.
Lior , I had the idea that you solve it? not committed yet?

Liorgal28:
I had another issue with UDP, but I saw this failure also.

teoparvanov:
Hi @saiakhil2012 @andrewjstone @glevkovich @Liorgal28 , currently @yontyon is working on a fix as part of PR https://github.com/vmware/concord-bft/pull/1197

andrewjstone:
I believe this has been fixed.

PR #1197, opened 2021-02-10
Fix pyclient_tests_udp

yontyon:
Recently we saw failures in pyclient_tests_udp.
After analyzing the logs, I was able to see that the failure happened when the primary started a bit late and the system was in slow path.
In this case, the client tries to read from the primary, but it takes time to commit the requests, and thus we get a tooSlowError message.

To fix the above I did the following:
1. Extending the udp client timeout from 1 second into 2 seconds.
2. Extend the read quorum for the bft client.
The second change may help in the case, where the primary has not got the FullCommitProof message from the collector.
In this case, other replicas may answer the read requests before the primary does.

In addition to the above, I found and fixed another small bag:
Recently we added the change for removing pending requests after they commit (instead of execution).
However, we may try to commit a message even if it already executed, in this case, the execution won't take any action on this sequence number (as it was already executed). However, since we clean the pending requests on committing, this may clear non-correct pending requests.
To solve this we added a condition in which if the committed sequence number was already executed we don't remove the pending requests of its clients.
Note, the bug above may delay a justified view change, but eventually, the view change will happen (due to some other pending request).


teoparvanov (approved):
LGTM!

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4023,6 +4023,7 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
 }
 
 void ReplicaImp::tryToRemovePendingRequestsForSeqNum(SeqNum seqNum) {
+  if (lastExecutedSeqNum >= seqNum) return;
   SCOPED_MDC_SEQ_NUM(std::to_string(seqNum));
```
HristoStaykov, line 4027: @yontyon , could you explain in what situation this might happen? Is it in some specific situation, as far as I know the common case is to first commit and then execute?

yontyon: A replica may receive a commit message even after it executed the same request (say, if the message has been delayed).
In this case, if we don't check it elsewhere we may clean again the client pending requests of this sequence number requests (maybe we do prevent this case even now, but I think it is better to make sure)

andrewjstone (changes requested):
Thanks for the fix @yontyon, but I don't see how changing the timeout to 2 seconds is going to make this test deterministically pass, even with the other fixes. I also don't see how changing the quorum size helps.

I believe the  problem is that we don't wait for the system to responding to requests before we test our actual logic. I suggested that [here](https://github.com/vmware/concord-bft/issues/1195#issuecomment-776259236). Do you disagree with my assessment?

Also, we need to remove those extra log lines in ReplicaImp.cpp and uncomment the tests you commented out before merge.

bftengine/src/bftengine/ReplicaImp.cpp
teoparvanov: @HristoStaykov do you agree with this comment?

andrewjstone (approved):
Looks great @yontyon! Thanks so much for fixing this. 👑

yontyon:
@teoparvanov , the CI run twice and passed successfully. I think it is safe to merge it

teoparvanov:
> @teoparvanov , the CI run twice and passed successfully. I think it is safe to merge it

Yoni, thanks so much for fixing this!

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: Yes I agree, it is good to have this check to make sure we have correct behaviour over all code paths.

PR #1198, opened 2021-02-10
Move KeyManager's Empty String from Loader Error to Warning

saiakhil2012:
Moving the error to warning as it happens every time during startup

### DIFF
--- a/bftengine/src/bftengine/KeyManager.cpp
+++ b/bftengine/src/bftengine/KeyManager.cpp
@@ -322,7 +322,7 @@ bool KeyManager::KeysView::load() {
   }
   auto str = secStore->load();
   if (str.empty()) {
-    LOG_ERROR(KEY_EX_LOG, "Got empty string from loader");
+    LOG_WARN(KEY_EX_LOG, "Got empty string from loader. This is expected on startup");
     return false;
   }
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/KeyManager.cpp
```diff
@@ -322,7 +322,7 @@ bool KeyManager::KeysView::load() {
   }
   auto str = secStore->load();
   if (str.empty()) {
-    LOG_ERROR(KEY_EX_LOG, "Got empty string from loader");
+    LOG_WARN(KEY_EX_LOG, "Got empty string from loader. This is expected on startup");
```
andrewjstone, line 325: This message isn't exactly right. If it was always expected on startup, we wouldn't need a warning :). It's expected on "first startup with an empty database."

andrewjstone (changes requested):

bftengine/src/bftengine/KeyManager.cpp
saiakhil2012: Yeah :), this is only expected if its a first startup with an empty database. Its better make the message elaborate, added that in the warning.

andrewjstone (approved):
LGTM. Thanks!

PR #1199, opened 2021-02-11
Docker update to version 0.12

andrewjstone:
We enable frame pointers and debug symbols for RocksDB to enable
profiling of release builds.


f-squirrel (approved):
LGTM!

PR #1200, opened 2021-02-11
Fix a race  with pending requests cleanup

yontyon:
@HristoStaykov found the following race:
1. We now clean pending requests after committing
2. Yet, we decide to ignore client requests based on these pending requests
The result is that we may accept client requests while we still executing their previous request, allowing the client to resubmit a request.

In this PR we fix this problem by adding a flag of `waitsForExecution` in the client info.
This flag is set once we get the client message and cleared only after the execution (similar to what we did before the latest change in the pending request)

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -133,6 +133,7 @@ void ClientsManager::loadInfoFromReservedPages() {
     if (ci.currentPendingRequest != 0 && (ci.currentPendingRequest <= replyHeader->reqSeqNum)) {
       ci.currentPendingRequest = 0;
       ci.timeOfCurrentPendingRequest = MinTime;
+      ci.waitsForExecution = false;
       ci.cid = std::string();
     }
   }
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -265,7 +266,7 @@ bool ClientsManager::noPendingAndRequestCanBecomePending(NodeIdType clientId, Re
   uint16_t idx = clientIdToIndex_.at(clientId);
   const ClientInfo& c = indexToClientInfo_.at(idx);
 
-  if (c.currentPendingRequest != 0) return false;  // if has pending request
+  if (c.currentPendingRequest != 0 || c.waitsForExecution == true) return false;  // if has pending request
```
HristoStaykov, line 269: // if has pending request, or the request was committed but not yet executed

HristoStaykov (approved):
LGTM, thanks @yontyon ! Just a minor comment clarification for the situation we discovered yesterday would be good to add.

bftengine/src/bftengine/ClientsManager.cpp
yontyon: I Will fix this in my next PT

PR #1201, opened 2021-02-11
Add metrics to storage

yontyon:
In this PR we add metrics to the new storage.
We started by adding to simple metrics:
1 . number of deleted keys per category

### DIFF
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -24,6 +24,7 @@
 #include "kv_types.hpp"
 #include "categorization/types.h"
 #include "thread_pool.hpp"
+#include "Metrics.hpp"
 
 namespace concord::kvbc::categorization {
 
@@ -178,6 +179,12 @@ class KeyValueBlockchain {
 
   // currently we are operating with single thread
   util::ThreadPool thread_pool_{1};
+
+  // metrics
+  std::shared_ptr<concordMetrics::Aggregator> aggregator_;
+  concordMetrics::Component metrics_;
+  concordMetrics::CounterHandle num_of_deleted_blocks_;
+  concordMetrics::CounterHandle num_of_deleted_keys_;
 
  public:
   struct KeyValueBlockchain_tester {
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -182,7 +182,7 @@ class KeyValueBlockchain {
 
   // metrics
   std::shared_ptr<concordMetrics::Aggregator> aggregator_;
-  concordMetrics::Component metrics_;
+  concordMetrics::Component delete_metrics_comp_;
   concordMetrics::CounterHandle versioned_num_of_deletes_keys_;
   concordMetrics::CounterHandle immuteable_num_of_deleted_keys_;
   concordMetrics::CounterHandle merkle_num_of_deleted_keys_;
@@ -205,7 +205,7 @@ class KeyValueBlockchain {
 
   void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) {
     aggregator_ = aggregator;
-    metrics_.UpdateAggregator();
+    delete_metrics_comp_.UpdateAggregator();
   }
   friend struct KeyValueBlockchain_tester;
 };
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -201,6 +201,11 @@ class KeyValueBlockchain {
 
     const VersionedRawBlock& getLastRawBlocked(KeyValueBlockchain& kvbc) { return kvbc.last_raw_block_; }
   };  // namespace concord::kvbc::categorization
+
+  void setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) {
+    aggregator_ = aggregator;
+    metrics_.UpdateAggregator();
+  }
   friend struct KeyValueBlockchain_tester;
 };
 
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -183,8 +183,9 @@ class KeyValueBlockchain {
   // metrics
   std::shared_ptr<concordMetrics::Aggregator> aggregator_;
   concordMetrics::Component metrics_;
-  concordMetrics::CounterHandle num_of_deleted_blocks_;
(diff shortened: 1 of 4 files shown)


kvbc/include/categorization/kv_blockchain.h
```diff
@@ -179,6 +180,13 @@ class KeyValueBlockchain {
   // currently we are operating with single thread
   util::ThreadPool thread_pool_{1};
 
+  // metrics
+  std::shared_ptr<concordMetrics::Aggregator> aggregator_;
+  concordMetrics::Component metrics_;
+  concordMetrics::CounterHandle versioned_num_of_deletes_keys_;
+  concordMetrics::CounterHandle immuteable_num_of_deleted_keys_;
+  concordMetrics::CounterHandle merkle_num_of_deleted_keys_;
+
```
cloudnoize, line 189: What does it mean to use metrics in the storage layer? I remember that in order to publish them some timer needs to lock and push the updates right? does it mean that we'll introduce locks to the flow? 
Also using those counters which are atomics have significant cost.
Personally I more and more tend to approach using log-based metrics.

yontyon: To address your concerns, I changed the component to be a delete component only.
It will update the aggregator every time we call the deleteBlock method.
Notice that update the aggregator requires grabbing a lock (as you mentioned), however, since we delete blocks only on pruning it doesn't really matter.
The counters themselves are not atomic, but again, if they were, we call delete block only on pruning.
I agree that we should consider enhancing our metrics platform (or maybe take a different approach) but currently, this is what we have

cloudnoize: cool

cloudnoize (approved):
LGTM thanks @yontyon

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -258,6 +266,9 @@ bool KeyValueBlockchain::deleteBlock(const BlockId& block_id) {
     return false;
   }
 
+  // Lets update the delete metrics component
+  delete_metrics_comp_.UpdateAggregator();
```
andrewjstone, line 270: Why are we updating the aggregator on every single block deletion? Wouldn't it be better to update after pruning is completed, or at least intermittently, every 1000 or so blocks?

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -373,6 +384,7 @@ void KeyValueBlockchain::deleteGenesisBlock(BlockId block_id,
                                             const std::string& category_id,
                                             const ImmutableOutput& updates_info,
                                             storage::rocksdb::NativeWriteBatch& batch) {
+  immuteable_num_of_deleted_keys_.Get().Inc(updates_info.tagged_keys.size());
```
andrewjstone, line 387: typo: `immutable`

kvbc/include/categorization/kv_blockchain.h
andrewjstone: @yontyon  is correct, a lock is only taken when aggregating stats. This is pretty much the case for all efficient metrics systems, including HdrHistogram. 

The key design is to make updating stats as fast as possible as it happens very frequently, and in the hot path most of the time. Therefore all counters are just local increments using non-atomic variables. They shouldn't take much more than a single cpu instruction, although I haven't checked the assembly. On the other hand, when an observer needs to view the collected metrics, they must be aggregated to a central location. This currently requires taking a lock on that central location so that the data can be copied to the central location. The only contention on that lock is from other writers who are updating simultaneously. It is this point where we may be able to change such that we don't require a lock, although in all cases some sort of synchronization will be required.

While much of the code updates the aggregator with a timer, because this is a straightforward mechanism to keep the centralized stats reasonably up to date, aggregation does not have to be done that way. As Yehonatan did here, it's just a matter of calling the `UpdateAggregator` method. 

I strongly think that all metrics should still be updates to local variables and only be made atomic when they must be updatable from multiple threads, as is the case now. The thing that we may want to change is the aggregation itself. One thing that @toly-kournik  and I discussed was getting rid of the aggregator altogether and instead using a `pull` mechanism, rather than a `push` mechanism. We would make the metrics register a callable and then have the diagnostics code (or similar) execute the callable on demand. In practice, running this callable would either require taking a lock or publishing to a promise (which may take a lock under the hood).  It's doubtful this would be much more efficient than the current code, but it would have the benefit that arbitrary timers wouldn't be used to publish metrics. The code would only pull when needed. So instead of publishing every 100ms even if stats are only updated on a dashboard every 5s, we'd pull once every 5s rather than 50 times. This also has the upside, that a single central component won't face much, if any contention for locks, and those locks will be spread across components (granular). The downside of this strategy is that now some central process decides when to take a lock on each independent component, rather than that independent component deciding when to publish when its most efficient for it to do so. We don't want a lock being taken in a hot path for storage and then blocking while the stats are copied out to the shared data structure. It also may be harder to structure the code such that you can update freely in a local variable and only take a lock an publish, as now two threads are accessing that data rather than just one.

kvbc/src/categorization/kv_blockchain.cpp
yontyon: We call deleteBlock only on pruning (for now at least). Updating the aggregator has almost zero effect on pruning performance (in the worse case we will wait for a lock that is taken by the puller of the aggregator).
Thus, in this specific case, I would prefer a better granularity of the metrics more than a very slightly performance improvement (if any).
WDYT?

kvbc/include/categorization/kv_blockchain.h
yontyon: I simply addressed Eran's comment. I think that for now, our metrics system balances the tradeoffs well.

kvbc/src/categorization/kv_blockchain.cpp
andrewjstone: I think it's a waste of resources, but probably not that important right now.I'm fine leaving it as is for now. We can always change it later and I suspect we will, but maybe not until we allow pruning to interleave with normal operation in which case the lock contention/cache coherency will matter more.

kvbc/include/categorization/kv_blockchain.h
andrewjstone: Yeah I know. I liked your explanation, but since this has come up a few times I figured I'd add a bit more detail for the record. I actually do think it's worthwhile to consider changing the metrics code as we rework our architecture and it's worth discussing in the future.

andrewjstone (approved):
LGTM. Thanks for the updates @yontyon.

PR #1202, opened 2021-02-11
New kv_blockchain_db_editor tool for db access.

KTeofanidis:
New **kv_blockchain_db_editor** tool similar to sparse_merkle_db_editor will allow for inspecting KeyValueBlockchains with categories.

```
Usage: kv_blockchain_db_editor PATH-TO-DB COMMAND [ARGUMENTS]...

Supported commands:

compareTo PATH-TO-OTHER-DB
  Compares the passed DB at PATH-TO-DB to the one in PATH-TO-OTHER-DB.
  Returns the ID of the first mismatching block, if such exists, in the overlapping 
  range found in the two databases. If there is no overlapping range in the 
  databases, no comparison is made.

getBlockInfo BLOCK-ID
  Returns information about the requested block (excluding its key-values).

getBlockKeyValues BLOCK-ID [CATEGORY]
  Returns the block's key-values. If a CATEGORY is passed only those keys/values are listed.

getCategories [BLOCK-VERSION]
  Returns all categories. If a BLOCK-VERSION is passed only categories in that block are listed.

getGenesisBlockID
  Returns the genesis block ID.

getLastBlockID
 Returns the last block ID. Either reachable or state transfer.

getLastReachableBlockID
  Returns the last reachable block ID.

getLastStateTransferBlockID
 Returns the last state transfer block ID. If there is no state transfer returns n/a.

getRawBlock BLOCK-ID
  Returns a serialized raw block (encoded in hex).

getRawBlockRange BLOCK-ID-START BLOCK-ID-END
  Returns a list of serialized raw blocks (encoded in hex) in the [BLOCK-ID-START, BLOCK-ID-END) range.

getValue CATEGORY HEX-KEY [BLOCK-VERSION]
  Gets a value by category, a hex-encoded key and (optionally) a block version.
  If no BLOCK-VERSION is passed, the value for the latest one will be returned
  (if existing). If the key doesn't exist at BLOCK-VERSION, it will not be returned.

removeMetadata
  Removes metadata and state transfer data from RocksDB.

Note:
The DB Editor is configured to use the following non-provable keys:
0x20, 0x22

Examples:
  kv_blockchain_db_editor /rocksdb-path getGenesisBlockID
  kv_blockchain_db_editor /rocksdb-path getRawBlock 42
  kv_blockchain_db_editor /rocksdb-path getValue merkle 0x0a0b0c
  kv_blockchain_db_editor /rocksdb-path getValue versioned 0x0a0b0c 42
```

Specific differences with the other tool are the new getCategories command, and the category parameter on getValue and getBlockKeyValues. When getBlockKeyValues is called without a category the output is the key/values for all categories one after the other.

Example output:
```
./kv_blockchain_db_editor ~/temp/kv getCategories
{
  "merkle": "block_merkle",
  "versioned": "versioned_kv"
}

./kv_blockchain_db_editor ~/temp/kv getBlockKeyValues 2
{
"merkle": {
  "0x0000000a": "0x00000014",
  "0x00000014": "0x00000028",
  "0x0000001e": "0x0000003c"
},
"versioned": {
  "0x766b65793130": "0x7676616c3230",
  "0x766b65793230": "0x7676616c3430",
  "0x766b65793330": "0x7676616c3630"
}
}

./kv_blockchain_db_editor ~/temp/kv/ getBlockInfo 1
{
  "parentBlockDigest": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "categoriesCount": "3",
  "categories": {
    "immutable": {
      "keyValueCount": "1",
      "type": "immutable"
    },
    "merkle": {
      "keyValueCount": "3",
      "type": "block_merkle"
    },
    "versioned": {
      "keyValueCount": "3",
      "type": "versioned_kv"
    }
  },
  "keyValueTotalCount": "7"
}

```

### DIFF
--- a/kvbc/test/sparse_merkle_storage/db_editor_tests_base.h
+++ b/kvbc/test/sparse_merkle_storage/db_editor_tests_base.h
@@ -1,0 +1,103 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include "gtest/gtest.h"
+#include "gmock/gmock.h"
+
+#include "endianness.hpp"
+#include "kvbc_storage_test_common.h"
+#include "kv_types.hpp"
+#include "merkle_tree_db_adapter.h"
+#include "PersistentStorageImp.hpp"
+#include "sliver.hpp"
+#include "storage/db_types.h"
+#include "storage/merkle_tree_key_manipulator.h"
+
+namespace {
+
+using namespace concord::kvbc::v2MerkleTree;
+using concord::kvbc::BlockId;
+using concord::kvbc::INITIAL_GENESIS_BLOCK_ID;
+using concord::kvbc::SetOfKeyValuePairs;
+using concordUtils::toBigEndianStringBuffer;
+using concordUtils::Sliver;
+
+using testing::EndsWith;
+using testing::HasSubstr;
+using testing::InitGoogleTest;
+using testing::Not;
+using testing::StartsWith;
+using testing::Test;
+
+using namespace std::string_literals;
+
+class DbEditorTestsBase : public Test {
+ public:
+  virtual void SetUp() override {
+    CleanDbs();
+    CreateBlockchain(main_path_db_id_, num_blocks_);
+  }
+
+  virtual void TearDown() override { CleanDbs(); }
+
+  virtual void CleanDbs() {
+    TestRocksDb::cleanup(main_path_db_id_);
+    TestRocksDb::cleanup(other_path_db_id_);
+    TestRocksDb::cleanup(non_existent_path_db_id_);
+    TestRocksDb::cleanup(empty_path_db_id_);
+  }
+
+  virtual void CreateBlockchain(std::size_t db_id,
+                                BlockId blocks,
+                                std::optional<BlockId> mismatch_at = std::nullopt) = 0;
+
+  virtual void DeleteBlocksUntil(std::size_t db_id, BlockId until_block_id) = 0;
+
+  Sliver getSliver(unsigned value) { return toBigEndianStringBuffer(value); }
+
+  SetOfKeyValuePairs generateMetadataAndStateTransfer() {
+    // Create a full range of metadata objects
+    constexpr auto num_metadata_object_ids = MAX_METADATA_PARAMS_NUM - INITIALIZED_FLAG;
+    auto updates = SetOfKeyValuePairs{};
+    concord::storage::v2MerkleTree::MetadataKeyManipulator manipulator;
+    for (concord::storage::ObjectId i = INITIALIZED_FLAG; i < num_metadata_object_ids; ++i) {
+      updates[manipulator.generateMetadataKey(i)] = getSliver(i);
+    }
+
+    concord::storage::v2MerkleTree::STKeyManipulator st_manipulator;
+    updates[st_manipulator.generateStateTransferKey(1)] = std::string{"val"};
+    updates[st_manipulator.generateSTPendingPageKey(1)] = std::string{"val"};
+    updates[st_manipulator.generateSTCheckpointDescriptorKey(1)] = std::string{"val"};
+    updates[st_manipulator.generateSTReservedPageDynamicKey(1, 1)] = std::string{"val"};
+    return updates;
+  }
+
+ protected:
+  static constexpr std::size_t main_path_db_id_{defaultDbId};
+  static constexpr std::size_t other_path_db_id_{main_path_db_id_ + 1};
+  static constexpr std::size_t non_existent_path_db_id_{other_path_db_id_ + 1};
+  static constexpr std::size_t empty_path_db_id_{non_existent_path_db_id_ + 1};
+  static constexpr unsigned kv_multiplier_{10};
+  static constexpr BlockId empty_block_id_{10};
+  static constexpr BlockId num_blocks_{10};
+  static constexpr BlockId more_num_blocks_{16};
+  static constexpr BlockId first_mismatch_block_id_{7};
+  static constexpr unsigned num_keys_{3};
+
+  std::stringstream out_;
+  std::stringstream err_;
+};
+
+}  // namespace
--- a/kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
+++ b/kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
@@ -1,6 +1,6 @@
 // Concord
 //
-// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
 //
 // This product is licensed to you under the Apache 2.0 license (the "License").
 // You may not use this product except in compliance with the Apache 2.0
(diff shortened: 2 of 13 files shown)


kvbc/test/sparse_merkle_storage/db_editor_tests_base.h
```diff
@@ -0,0 +1,90 @@
+#pragma once
```
dartdart26, line 1: I think we should add the standard Concord license and copyright .

kvbc/tools/db_editor/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_library(lib_db_editor INTERFACE)
+target_link_libraries(lib_db_editor INTERFACE)
+target_include_directories(lib_db_editor INTERFACE include)
```
dartdart26, line 3: Nice way to reuse code!

kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+      // Add a key to simulate a mismatch.
+      if (mismatch_at.has_value() && *mismatch_at == i) {
+        merkle_updates.addUpdate(mismatch_kv.first.toString(), mismatch_kv.second.toString());
+        ver_updates.addUpdate(mismatch_kv.first.toString(), mismatch_kv.second.toString());
+        // immutable_updates.addUpdate(mismatch_kv.first.toString() + "im"s, {"immutable_val", {"1", "2"}});
+      }
+      updates.add(kCategoryMerkle, std::move(merkle_updates));
+      updates.add(kCategoryVersioned, std::move(ver_updates));
+      // updates.add("immutable", std::move(immutable_updates));
```
dartdart26, line 59: Can we still add some immutable updates, even if a static amount of them?

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+#include "db_editor_common.hpp"
+#include "categorization/kv_blockchain.h"
+
+namespace concord::kvbc::tools::db_editor {
+
+using namespace categorization;
+
+inline const auto kToolName = "kv_blockchain_db_editor"s;
+inline KeyValueBlockchain getAdapter(const std::string &path, bool read_only = false) {
+  std::cout << "db path " << path << std::endl;
```
dartdart26, line 25: Will this line interfere with the tool output?

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+  std::string execute(const KeyValueBlockchain &adapter, const CommandArguments &) const {
+    return toJson("lastReachableBlockID", adapter.getLastReachableBlockId());
+  }
+};
+
+struct GetLastStateTransferBlockID {
+  bool read_only = true;
+  std::string description() const {
+    return "getLastStateTransferBlockID\n"
+           " Returns the last state transfer block ID. If there is no state transfer returns n/a";
```
dartdart26, line 65: Nit: maybe can put `.` at the end of the help line.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+    auto blockId = adapter.getLastStatetransferBlockId();
+    return toJson("lastStateTransferBlockID", blockId ? std::to_string(*blockId) : "n/a"s);
+  }
+};
+
+struct GetLastBlockID {
+  bool read_only = true;
+  std::string description() const {
+    return "getLastBlockID\n"
+           " Returns the last block ID. Either reachable or state transfer";
```
dartdart26, line 78: Nit: maybe can put . at the end of the help line.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+  std::string description() const {
+    return "getRawBlock BLOCK-ID\n"
+           "  Returns a serialized raw block (encoded in hex).";
+  }
+
+  std::string execute(const KeyValueBlockchain &adapter, const CommandArguments &args) const {
+    if (args.values.empty()) {
+      throw std::invalid_argument{"Missing BLOCK-ID argument"};
+    }
+    const auto raw_block = adapter.getRawBlock(toBlockId(args.values.front()));
```
dartdart26, line 97: Maybe can throw an exception if the raw_block is null. Rationale is that the user knows which blocks are available and which are not.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+    for (auto const &catkv : catUpdates.kv) {
+      keyValueCount += std::visit([](auto &&arg) { return arg.kv.size(); }, catkv.second);
+    }
+
+    return toJson(std::map<std::string, std::string>{
+        std::make_pair(
+            "parentBlockDigest",
+            parent_digest ? concordUtils::bufferToHex(parent_digest->data(), parent_digest->size()) : "n/a"s),
+        std::make_pair("keyValueCount", std::to_string(keyValueCount)),
+        std::make_pair("categoriesCount", std::to_string(catUpdates.kv.size()))});
```
dartdart26, line 164: I think what might be useful is to report key counts per category. Don't think it is critical for this PR, but just something I think people would want.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+    return "getBlockInfo BLOCK-ID\n"
+           "  Returns information about the requested block (excluding its key-values).";
+  }
+
+  std::string execute(const KeyValueBlockchain &adapter, const CommandArguments &args) const {
+    if (args.values.empty()) {
+      throw std::invalid_argument{"Missing BLOCK-ID argument"};
+    }
+    auto blockId = toBlockId(args.values.front());
+    const auto parent_digest = adapter.parentDigest(blockId);
```
dartdart26, line 148: I am thinking that if the block is found (after getBlockUpdates()), then the `parent_digest` must never be nullopt. Maybe can move it after getBlockUpdates() and assert if it is actually nullopt. That way we can avoid checking it later and setting "n\a". On the other hand, I guess this is an internal KV blockchain detail and not explicit from the interface. We can also leave as is. What do you think?

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+      val);
+
+  return kvout;
+}
+
+struct GetBlockKeyValues {
+  bool read_only = true;
+  std::string description() const {
+    return "getBlockKeyValues BLOCK-ID [CATEGORY]\n"
+           "  Returns the block's key-values. If a CATEGORY is passed only those keys/values are listed";
```
dartdart26, line 196: Nit: maybe can put a `.` at the end of the help line.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+      }
+      raw_blocks.emplace_back("rawBlock" + std::to_string(i),
+                              concordUtils::vectorToHex(categorization::RawBlock::serialize(*raw_block)));
+    }
+    return toJson(raw_blocks);
+  }
+};
+
+struct GetBlockInfo {
+  bool read_only = true;
```
dartdart26, line 137: I realise this is not part of this PR, but we can use it to change all `read_only` members in all commands to being `const`.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+          if constexpr (std::is_same_v<T, BlockMerkleInput>) {
+            kvout[concordUtils::bufferToHex(c.first.data(), c.first.size())] =
+                concordUtils::bufferToHex(c.second.data(), c.second.size());
+          } else if constexpr (std::is_same_v<T, VersionedInput>) {
+            kvout[concordUtils::bufferToHex(c.first.data(), c.first.size())] =
+                concordUtils::bufferToHex(c.second.data.data(), c.second.data.size());
+          } else if constexpr (std::is_same_v<T, ImmutableInput>) {
+            kvout[concordUtils::bufferToHex(c.first.data(), c.first.size())] =
+                concordUtils::bufferToHex(c.second.data.data(), c.second.data.size());
+          }
```
dartdart26, line 184: I guess these two `else if` statements can be combined into one as they are having the same "interface".

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+        out[c.first] = std::visit(
+            [&cat_type_str](auto &&arg) {
+              using T = std::decay_t<decltype(arg)>;
+              if constexpr (std::is_same_v<T, BlockMerkleInput>) {
+                return cat_type_str.at(CATEGORY_TYPE::block_merkle);
+              } else if constexpr (std::is_same_v<T, VersionedInput>) {
+                return cat_type_str.at(CATEGORY_TYPE::versioned_kv);
+              } else if constexpr (std::is_same_v<T, ImmutableInput>) {
+                return cat_type_str.at(CATEGORY_TYPE::immutable);
+              }
```
dartdart26, line 266: Maybe this snippet can be simplified by using something like:

```
template <typename T>
struct CategoryType;

template <>
struct CategoryType<CATEGORY_TYPE::block_merkle> {
  static const std::string string_type = "block_merkle";
};
...

Not sure if that is important or makes sense. Maybe it does in order to reuse it and not having if statements inside the logic. What do you think?
```

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -0,0 +1,486 @@
...
+           "  Gets a value by category, a hex-encoded key and (optionally) a block version.\n"
+           "  If no BLOCK-VERSION is passed, the value for the latest one will be returned\n"
+           "  (if existing). If the key doesn't exist at BLOCK-VERSION, it will not be returned.";
+  }
+
+  std::string execute(const KeyValueBlockchain &adapter, const CommandArguments &args) const {
+    if (args.values.size() < 2) {
+      throw std::invalid_argument{"Missing CATEGORY and HEX-KEY arguments"};
+    }
+    std::string category = args.values[0];
```
dartdart26, line 294: Nit: Maybe can use `const auto &` to avoid copying the value.

kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
```diff
@@ -0,0 +1,567 @@
...
+            run(CommandLineArguments{{kTestName, rocksDbPath(main_path_db_id_), "getValue", kCategoryMerkle, "abck"}},
+                out_,
+                err_));
+  ASSERT_TRUE(out_.str().empty());
+  ASSERT_THAT(err_.str(), StartsWith("Failed to execute command [getValue], reason: Invalid hex string: abck"));
+}
+
+TEST_F(DbEditorTests, get_categories) {
+  ASSERT_EQ(EXIT_SUCCESS,
+            run(CommandLineArguments{{kTestName, rocksDbPath(main_path_db_id_), "getCategories"}}, out_, err_));
```
dartdart26, line 537: Maybe we should add a test that passes a `BLOCK-VERSION` too and verify the output is valid per block too.

kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
```diff
@@ -0,0 +1,567 @@
...
+  ASSERT_EQ(
+      EXIT_SUCCESS,
+      run(CommandLineArguments{{kTestName, rocksDbPath(main_path_db_id_), "getValue", kCategoryMerkle, "0x00000028"}},
+          out_,
+          err_));
+  ASSERT_TRUE(err_.str().empty());
+  ASSERT_EQ("{\n  \"value\": \"0x00000050\"\n}\n", out_.str());
+}
+
+TEST_F(DbEditorTests, get_value_with_block_version) {
```
dartdart26, line 436: Another test that can be added is to call `getValue` with a non-existent block version. Maybe a version bigger than the latest block ID. I guess `getValue` with version 0 below handles it, but might still be a good test to add.

kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
```diff
@@ -0,0 +1,567 @@
...
+            run(
+                CommandLineArguments{
+                    {kTestName, rocksDbPath(main_path_db_id_), "getValue", kCategoryMerkle, "0x00000028", "0"}},
+                out_,
+                err_));
+  ASSERT_TRUE(out_.str().empty());
+  ASSERT_THAT(err_.str(), StartsWith("Failed to execute command [getValue], reason: "));
+}
+
+TEST_F(DbEditorTests, get_value_wrong_category) {
```
dartdart26, line 479: Another one that can be added is a test with a non-existent category. I guess that goes into the realm of testing the adapter itself, so we might not do it. Just a thought.

dartdart26 (approved):
Looks great, very clean, nice code reuse! Thanks @KTeofanidis!

I have put some minor comments that you can decide to address or not. Maybe the one related to using the Immutable category is most important (still not critical).

PR #1203, opened 2021-02-12
Kvbcbench concurrent reads

andrewjstone:
The second commit in this PR adds concurrent reads to `kvbcbench` to simulate read load on RocksDB during pre-execution.

Only the second commit should be reviewed, since the first commit is identical to #1196.


dartdart26 (approved):

PR #1206, opened 2021-02-15
Sort deletes in categorization::Updates

dartdart26:
Deletes in BlockMerkleInput and VersionedInput are serialized as a
vector. The order of elements in the vector is determined by insertion
order. However, deletes in BlockMerkleOutput and VersionedOutput are
contained in a map and are, therefore, sorted. In addition, the
KeyValueBlockchain contains a cached value of the last raw block set
during addBlock() so that it can avoid expensive calls to getRawBlock().

The effect of above is that a replica receiving State Transfer raw
blocks will construct its BlockMerkleInput and VersionedInput objects
with sorted order from the output data structures, whereas a replica
using the cached last raw block will use the insertion order coming from
BlockMerkleUpdates and VersionedUpdates. These two orders are
potentially different, leading to different parent digests and,
subsequently, different blockchains in these replicas.

In order to fix, sort and remove duplicate deletes in BlockMerkleUpdates
and VersionedUpdates. Implement by keeping track of when deletes are
added and only sort and remove duplicates when needed. Add unit tests to
verify the behavior.

Thanks to @cloudnoize for noticing this problem!

### DIFF
--- a/kvbc/benchmark/categorization_benchmark.cpp
+++ b/kvbc/benchmark/categorization_benchmark.cpp
@@ -17,7 +17,14 @@
 #include "categorization/details.h"
 #include "categorized_kvbc_msgs.cmf.hpp"
 
+#include <algorithm>
 #include <cstddef>
+#include <cstdint>
+#include <iterator>
+#include <random>
+#include <set>
+#include <unordered_set>
+#include <vector>
 
 namespace {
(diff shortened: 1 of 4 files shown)


cloudnoize (approved):
incredible @dartdart26

PR #1207, opened 2021-02-16
Start servers only once

yontyon:
Another try to stabilize the tests.
We saw that starting and stopping the servers frequently make the test unstable.
I changed the test such that we start the servers only once on setup.
Tests passed locally multiple times and it seems more stable now.


util/pyclient/test_client.py
```diff
@@ -95,33 +108,16 @@ def testTimeout(self):
         trio.run(self._testTimeout, write, False)
 
     async def _testTimeout(self, msg, read_only):
-       config = self.config._replace(req_timeout_milli=100)
+       config = self.config._replace(req_timeout_milli=0)
```
teoparvanov, line 111: what does a timeout of 0 mean?

util/pyclient/test_client.py
```diff
@@ -138,81 +134,60 @@ def testRetry(self):
...
-            raise
-        finally:
-            self.stopServers()
+        await self.writeWithRetryAssert()
 
     async def writeWithRetryAssert(self):
         """Issue a write and ensure that a retry occurs"""
-        config = self.config._replace(req_timeout_milli=5000)
-        val = 1
+        config = self.config._replace(retry_timeout_milli=0)
```
teoparvanov, line 141: same here - retry timeout = 0?

util/pyclient/test_client.py
```diff
@@ -138,81 +134,60 @@ def testRetry(self):
...
+        trio.run(self._testPrimaryWrite)
+
+    @verify_system_is_ready
     async def _testPrimaryWrite(self):
        # Try to guarantee we don't retry accidentally
        config = self.config._replace(retry_timeout_milli=500)
        with bft_client.UdpClient(config, self.replicas, None) as udp_client:
            self.assertEqual(None, udp_client.primary)
-           await udp_client.sendSync(self.writeRequest(1), False)
+           await udp_client.sendSync(self.writeRequest(3), False)
```
teoparvanov, line 162: are you sure the unittest framework will always run the test cases in the same order?

teoparvanov (commented):
Some questions ...

util/pyclient/test_client.py
yontyon: It doesn't really matter.
I just wanted to have different values for different tests

util/pyclient/test_client.py
yontyon: Here it means that the client will retry immidetly.
Before the change, we testes it by delaying the server's startup (and once they started, the request should have been accepted).
Here again, we can argue that the purpose of the test is to see that the client is able to retry (no matter why).
But I also agree that this may be considered as a degenerated test case.
WDYT?

util/pyclient/test_client.py
yontyon: it means the client will fail immediately.
Notice that before the change, we didn't even start the servers on this test (the purpose is to test the client behavior)

teoparvanov (approved):
I like the approach of starting the servers once. However, not sure you need to change the timeouts. The original issue was that replicas weren't starting, right - not that the timeouts needed improvement? Also, timeout of 0 is really a sort of non-sense.

andrewjstone (approved):
LGTM. Thanks @yontyon

PR #1208, opened 2021-02-16
Honour calculate_root_hash in all categories

dartdart26:
If root hash calculation has been requested via the flag in both
VersionedUpdates and ImmutableUpdates, make sure the root hash optional
in the output always contains a value, irrespective of input keys and
tags. In case of VersionedUpdates with no keys, the output contains the
hash of an empty string. In case of ImmutableUpdates with keys and no
tags or without any keys, the output contains an empty map of tags to
hashes.

Above is important in case of State Transfer, because the fact that an
output root hash optional has a value (even if it is the empty string
hash or an empty map) signifies that root hash calculation was
originally requested and a receiving replica can use that fact to
reconstruct the exact same update.

Put comments describing possible optional values (or lack of) in the
CMF file.

Update unit tests to reflect the changed behaviour.

### DIFF
--- a/kvbc/cmf/categorized_kvbc_msgs.cmf
+++ b/kvbc/cmf/categorized_kvbc_msgs.cmf
@@ -54,12 +54,22 @@ Msg VersionedKeyFlags 1002 {
 
 Msg VersionedOutput 1003 {
     map string VersionedKeyFlags keys
+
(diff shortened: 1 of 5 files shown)


cloudnoize (approved):
great @dartdart26

PR #1209, opened 2021-02-16
Additional logs for View Change and retrieved block size

HristoStaykov:
Added log when after View Change Replicas finish the rebuilding of the previous View's Working Window.
Added DEBUG log for size of the retrieved block.


dartdart26 (approved):
Thanks Hristo!

PR #1211, opened 2021-02-17
Add reqSeqNum to Preprocessor logs via MDC

nkgeorgiev:


yuliasherman:
Why do we need to add the seqNum as a string? All existing logs already contain it and if you found places where it is missing, just add it. Besides that, converting numbers to strings in run-time is a bad idea.

nkgeorgiev:
> Why do we need to add the seqNum as a string? All existing logs already contain it and if you found places where it is missing, just add it. Besides that, converting numbers to strings in run-time is a bad idea.

The idea is to have easier parsing of the logs. Though I guess it is pretty easy even now.

PR #1214, opened 2021-02-18
Remove read from wait_for_liveness

KTeofanidis:
To improve stability we removed the unnecessary read from the tracked wait_for_liveness


teoparvanov (approved):
LGTM!

PR #1215, opened 2021-02-18
Fix State Transfer thread safety bug.

toly-kournik:
When ST mt mode is enabled there were flows when the same local buffer was accessed by two threads, causing inconsistencies.
Use local buffers for reserved pages manipulation.
Use local buffer for block manioulation in a getBlockAndComputeDigest() wich used buffer_.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -230,8 +230,7 @@ BCStateTran::BCStateTran(const Config &config, IAppState *const stateApi, DataSt
   // Register metrics component with the default aggregator.
   metrics_component_.Register();
 
-  // TODO(GG): more asserts
-  buffer_ = reinterpret_cast<char *>(std::malloc(maxItemSize_));
+  buffer_ = new char[maxItemSize_]{};
   LOG_INFO(getLogger(), "Creating BCStateTran object: " << config_);
 
   if (config_.runInSeparateThread) {
@@ -254,7 +253,7 @@ BCStateTran::~BCStateTran() {
   ConcordAssert(cacheOfVirtualBlockForResPages.empty());
   ConcordAssert(pendingItemDataMsgs.empty());
 
-  std::free(buffer_);
+  delete[] buffer_;
 }
 
 // Load metrics that are saved on persistent storage
@@ -284,8 +283,6 @@ void BCStateTran::init(uint64_t maxNumOfRequiredStoredCheckpoints,
     numberOfReservedPages_ = numberOfRequiredReservedPages;
     metrics_.number_of_reserved_pages_.Get().Set(numberOfReservedPages_);
     metrics_.size_of_reserved_page_.Get().Set(sizeOfReservedPage);
-
-    memset(buffer_, 0, maxItemSize_);
 
     LOG_INFO(getLogger(),
              "Init BCStateTran object:" << KVLOG(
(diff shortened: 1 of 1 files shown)


teoparvanov:
Hi @toly-kournik , please rebase your PR - there is a fix on master for the pre-execution test failures.

andrewjstone (approved):
Thanks for the fix Toly!

Just curious though, why use dynamically allocated arrays and not a vector?

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -430,15 +427,14 @@ STDigest BCStateTran::checkpointReservedPages(uint64_t checkpointNumber, DataSto
   auto numberOfPagesInCheckpoint = pages.size();
   LOG_INFO(getLogger(),
            "Associating pending pages with checkpoint: " << KVLOG(numberOfPagesInCheckpoint, checkpointNumber));
-
+  std::unique_ptr<char[]> buffer(new char[config_.sizeOfReservedPage]);
```
HristoStaykov, line 430: Should we zero the newly allocated buffer?

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -1693,13 +1686,14 @@ char *BCStateTran::createVBlock(const DescOfVBlockForResPages &desc) {
   char *elements = rawVBlock + sizeof(HeaderOfVirtualBlock);
 
   uint32_t idx = 0;
+  std::unique_ptr<char[]> buffer(new char[config_.sizeOfReservedPage]);
```
HristoStaykov, line 1689: Same as above. Might be better to zero the new buffer?

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2414,20 +2398,20 @@ void BCStateTran::checkStoredCheckpoints(uint64_t firstStoredCheckpoint, uint64_
           ConcordAssertEQ(computedDigestOfResPagesDescriptor, desc.digestOfResPagesDescriptor);
         }
         // check all pages descriptors
+        std::unique_ptr<char[]> buffer(new char[config_.sizeOfReservedPage]);
```
HristoStaykov, line 2401: Same as above.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2487,11 +2471,14 @@ std::array<std::uint8_t, BLOCK_DIGEST_SIZE> BCStateTran::computeDigestOfBlock(co
 }
 
 STDigest BCStateTran::getBlockAndComputeDigest(uint64_t currBlock) {
+  // This function is called among others during checkpointing of current state,
+  // which can occur while this replica is a source replica.
+  // In order to make it thread safe, instead of using buffer_, a local buffer is allocated .
+  static std::unique_ptr<char[]> buffer(new char[maxItemSize_]);
```
HristoStaykov, line 2477: {} ? :)

HristoStaykov (approved):
Thanks @toly-kournik for the fix. I see we don't zero the buffer in some allocations, might be we have a guarantee that it will be filled with data in those places so no need to zero?

bftengine/src/bcstatetransfer/BCStateTran.cpp
toly-kournik: There's no need to zero in these cases, since the data are memcpy-ed to them, so I believe it's a bit quicker.
I only zero the buffer in zeroReservedPages().

PR #1216, opened 2021-02-18
Create categories on startup in KeyValueBlockchain

dartdart26:
Pass a map from category ID to type in KeyValueBlockchain's constructor.
Rationale is that this allows us to have concurrent reads and writes
without locking for access to internal in-memory category maps in
KeyValueBlockchain, namely `categories_` and `category_types_`. By
passing the map at construction, we no longer have the ability to create
categories on demand. However, we don't need that at the moment.

We allow addition of new categories to an existing blockchain. We also
require that either all existing categories (+ optionally new ones) are
passed on construction or no categories are passed at all, leading to
the system using the persisted ones only.

Add the category ID to type map to kvbc::ReplicaImp's constructor too.
Make sure that the `kConcordInternalCategoryId` is always created, even
if not passed explicitly by the user.

Add support for an optional RocksDB configuration file path in
RocksDBStorageFactory.

### DIFF
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -32,7 +32,15 @@ class KeyValueBlockchain {
   using VersionedRawBlock = std::pair<BlockId, std::optional<categorization::RawBlockData>>;
 
  public:
-  KeyValueBlockchain(const std::shared_ptr<concord::storage::rocksdb::NativeClient>& native_client, bool link_st_chain);
+  // Creates a key-value blockchain.
+  // If `category_types` is nullopt, the persisted categories in storage will be used.
+  // If `category_types` has a value, it should contain all persisted categories in storage at a minimum. New ones will
+  // be created and persisted.
+  // Users are required to pass a value for `category_types` on first construction (i.e. a new blockchain) in order to
+  // specify the categories in use. Failure to do so will generate an exception.
+  KeyValueBlockchain(const std::shared_ptr<concord::storage::rocksdb::NativeClient>& native_client,
+                     bool link_st_chain,
+                     const std::optional<std::map<std::string, CATEGORY_TYPE>>& category_types = std::nullopt);
   /////////////////////// Add Block ///////////////////////
 
   BlockId addBlock(Updates&& updates);
@@ -83,7 +91,7 @@ class KeyValueBlockchain {
   std::optional<Updates> getBlockUpdates(BlockId block_id) const;
 
   // Get a map from category ID -> type for all known categories in the blockchain.
-  const std::map<std::string, CATEGORY_TYPE>& blockchainCategories() const { return categorires_types_; }
+  const std::map<std::string, CATEGORY_TYPE>& blockchainCategories() const { return category_types_; }
 
   std::shared_ptr<concord::storage::rocksdb::NativeClient> db() { return native_client_; }
   std::shared_ptr<const concord::storage::rocksdb::NativeClient> db() const { return native_client_; }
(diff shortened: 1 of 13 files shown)


kvbc/include/categorization/kv_blockchain.h
```diff
@@ -100,16 +108,21 @@ class KeyValueBlockchain {
...
-  bool insertCategoryMapping(const std::string& cat_id,
-                             const CATEGORY_TYPE type,
-                             concord::storage::rocksdb::NativeWriteBatch& write_batch);
+  void insertCategoryMapping(const std::string& cat_id, const CATEGORY_TYPE type);
+  void addNewCategory(const std::string& cat_id, CATEGORY_TYPE type);
 
-  const Category& getCategory(const std::string& cat_id) const;
-  Category& getCategory(const std::string& cat_id);
-  bool hasCategory(const std::string& cat_id) const;
+  // Return nullopt if the category doesn't exist.
```
andrewjstone, line 119: typo: `nullopt` -> `nullptr`.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -160,88 +211,105 @@ std::future<BlockDigest> KeyValueBlockchain::computeParentBlockDigest(const Bloc
...
+  return *category;
+}
 
 std::optional<Value> KeyValueBlockchain::get(const std::string& category_id,
                                              const std::string& key,
                                              BlockId block_id) const {
-  if (!hasCategory(category_id)) {
+  const auto category = getCategoryPtr(category_id);
+  if (!category) {
     return std::nullopt;
```
andrewjstone, line 251: I know this is maintaining existing behavior, but we may want to consider throwing if a category isn't available. That would be a pretty unexpected error from a user point of view, and it can be confused easily with the key just not existing which is non-exceptional behavior.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -429,46 +497,54 @@ void KeyValueBlockchain::deleteLastReachableBlock(BlockId block_id,
...
-    throw std::runtime_error{"Failed to insert new category"};
+  if (!inserted) {
+    LOG_FATAL(CAT_BLOCK_LOG, "Category [" << cat_id << "] already exists in categories map");
+    ConcordAssert(false);
   }
-  write_batch.put(detail::CAT_ID_TYPE_CF, cat_id, std::string(1, static_cast<char>(type)));
-  return true;
 }
 
 BlockMerkleOutput KeyValueBlockchain::handleCategoryUpdates(BlockId block_id,
```
andrewjstone, line 541: Do we actually use these `handleCategoryUpdates` methods anywhere?

andrewjstone (approved):
Looks great. Thanks for this added functionality @dartdart26!

kvbc/src/categorization/kv_blockchain.cpp
dartdart26: I think I agree, will put as a TODO. I don't exactly remember why it returns nullopt.

kvbc/src/categorization/kv_blockchain.cpp
dartdart26: Yes, we do - when adding a block. I guess they can be rewritten as a single template method, introducing a type map from input to category.

kvbc/include/categorization/kv_blockchain.h
dartdart26: Will fix in a subsequent commit.

PR #1217, opened 2021-02-19
Revert to not using wait_for_liveness

KTeofanidis:
Due to instabilities revert back to not using the new wait_for_liveness method. Also add small delay to unstable test_wedge_command_with_state_transfer to see if it improves stability.


teoparvanov (approved):
LGTM, thanks @KTeofanidis !

PR #1219, opened 2021-02-20
kvbcbench: Use KeyValueBlockchain directly

andrewjstone:
This is a small, intermediate change that uses the `KeyValueBlockchain` to
add blocks rather than only using the `BlockMerkleCategory`. While this
code still only adds data to the BlockMerkleCategory, the usage of the
KeyValueBlockchain also results in the writing of the block nodes
themselves, and results in more accurate benchmarks. A follow up commit
will modify the code to write different key distributions to different
categories.

This code also includes an updated rocksdb configuration file containing
column family configuration for all 3 categories.

`kvbcbench --help` also displays usage information properly now.


dartdart26 (approved):
Looks good, thanks @andrewjstone !

PR #1220, opened 2021-02-21
Fix stale keys deleted metrics

yontyon:
In this PR we fix the metrics for keys deletions.
On deleteGenesisBlock we count only the deletions of stale keys

### DIFF
--- a/kvbc/include/categorization/block_merkle_category.h
+++ b/kvbc/include/categorization/block_merkle_category.h
@@ -73,7 +73,7 @@ class BlockMerkleCategory {
   std::vector<std::string> getBlockStaleKeys(BlockId, const BlockMerkleOutput&) const;
   // Delete the given block ID as a genesis one.
   // Precondition: The given block ID must be the genesis one.
-  void deleteGenesisBlock(BlockId, const BlockMerkleOutput&, storage::rocksdb::NativeWriteBatch&);
+  std::size_t deleteGenesisBlock(BlockId, const BlockMerkleOutput&, storage::rocksdb::NativeWriteBatch&);
 
   // Delete the given block ID as a last reachable one.
   // Precondition: The given block ID must be the last reachable one.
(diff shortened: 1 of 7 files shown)


kvbc/include/categorization/block_merkle_category.h
```diff
@@ -72,7 +72,7 @@ class BlockMerkleCategory {
 
   // Delete the given block ID as a genesis one.
   // Precondition: The given block ID must be the genesis one.
-  void deleteGenesisBlock(BlockId, const BlockMerkleOutput&, storage::rocksdb::NativeWriteBatch&);
```
dartdart26, line 75: Can we put a comment that these functions return the number of keys actually deleted from disk? Applies to all categories.

dartdart26: Nit: Can use `std::size_t` instead.

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -541,11 +544,17 @@ std::pair<SetOfKeyValuePairs, KeysVector> BlockMerkleCategory::rewriteAlreadyPru
...
 
-void BlockMerkleCategory::deleteGenesisBlock(BlockId block_id, const BlockMerkleOutput& out, NativeWriteBatch& batch) {
+size_t BlockMerkleCategory::deleteGenesisBlock(BlockId block_id,
+                                               const BlockMerkleOutput& out,
+                                               NativeWriteBatch& batch) {
   auto [hashed_keys, latest_versions] = getLatestVersions(out);
   auto overwritten_active_keys_from_pruned_blocks = findActiveKeysFromPrunedBlocks(hashed_keys);
+  size_t num_of_deletes = 0;
+  for (auto& kv : overwritten_active_keys_from_pruned_blocks) {
+    num_of_deletes += kv.second.size();
```
dartdart26, line 554: Are we counting key deletes from disk only or stale data too?

yontyon: stale only

dartdart26: @andrewjstone Could you please also have a look at the block merkle category?

dartdart26 (approved):
Thanks @yontyon, looks good!

kvbc/src/categorization/block_merkle_category.cpp
andrewjstone: Looks fine to me. It's somewhat strange that we have 3 ways to return data, but 🤷

andrewjstone (approved):
Thanks for adding the support for this @yontyon

PR #1226, opened 2021-02-24
Fix mini-batching issues in SimpleClientImp/PreProcessor

yuliasherman:
Add operation result to ClientReply structure. 
Set batch expiration timeout to the maximum single request timeout.
Fix reply buffer offset calculation in PreProcessor.
Verify arrival of all replies.


f-squirrel (approved):
LGTM

PR #1227, opened 2021-02-24
BCStateTran: add lastInBatch to ItemDataMsg, change fetch retransmission

glevkovich:
1. source mark last ItemDataMsg sent in a batch with true. This enables
the destination to send fetchBlockMsg immediately after and improving the
protocol efficiency. That will trigger sending FetchBlocksMsg
immediately. This will optimize St speed when there are no drops.
We might see a significant increase in throughput after that.
2. retransmissionTimeout was not used correctly. That's caused
a significant slowdown in the protocol.
Use fetchRetransmissionTimeoutMs parameter in a different way - for
every received ItemDataMsg , time is recorded. If more than
<fetchRetransmissionTimeoutMs> ms passed since last received
ItemDataMsg, or initially from sending fetch block message - we
retransmit FetchBlocksMsg.
Optimize slightly source - avoid last chosen source + avoid a primary.
The same handling is done for FetchResPagesMsg.
3. Since retransmissionTimeout = 250 is too low, change the default value to 1000.
We have noticed long delays at a heavily loaded consensus network until ST
incoming messages are processed. Until we prioritize messages, we must
increase this timeout.

### DIFF
--- a/bftengine/src/bcstatetransfer/Messages.hpp
+++ b/bftengine/src/bcstatetransfer/Messages.hpp
@@ -156,6 +156,7 @@ struct ItemDataMsg : public BCStateTranBaseMsg {
   uint16_t chunkNumber;
 
   uint32_t dataSize;
+  uint8_t lastInBatch;
   char data[1];
 
   uint32_t size() const { return sizeof(ItemDataMsg) - 1 + dataSize; }
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -963,7 +963,7 @@ void BCStateTran::sendFetchBlocksMsg(uint64_t firstRequiredBlock,
                   msg.lastRequiredBlock,
                   msg.lastKnownChunkInLastRequiredBlock));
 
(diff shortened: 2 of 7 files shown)


bftengine/src/bcstatetransfer/Messages.hpp
```diff
@@ -156,6 +156,7 @@ struct ItemDataMsg : public BCStateTranBaseMsg {
   uint16_t chunkNumber;
 
   uint32_t dataSize;
+  uint lastInBatch;
```
toly-kournik, line 159: bool?

glevkovich: yes, it's a bool, but bool is implementation-specific. can be 1 byte or 4 bytes. It's not a good habit in my idea to put a bool on messages going between machines.

toly-kournik: can you change it to a more standart uint8_t?

glevkovich: Sure, That is a typo, good that you saw it.
I meant it to be uint8_t.
Thanks.

toly-kournik (approved):
LGTM! Thanks Gil

PR #1228, opened 2021-02-24
Add perf sampling option to apollo

HristoStaykov:
Added support to select a Replica for performance analysis by collecting stack samples.
Added test duration record when KEEP_APOLLO_LOGS=true


teoparvanov (approved):
LGTM!

PR #1230, opened 2021-02-24
Enable using storage from multiple threads

cloudnoize:
Under the assumption that writes will occur only from the message-processing and state-transfer thread(s) and that those threads can't execute in parallel. 

- make last_reachable_block_id_ atomic

- make genesis_block_id_ atomic

- call setAddedBlockId() after commit to storage

- Do not use getLastStateTranferBLockId() in ReplicaImp::getLastBlockId()

- Make last_block_id_ in ST chain atomic

### DIFF
--- a/kvbc/include/ReplicaImp.h
+++ b/kvbc/include/ReplicaImp.h
@@ -98,6 +98,8 @@ class ReplicaImp : public IReplica,
   bool getPrevDigestFromBlock(uint64_t blockId, bftEngine::bcst::StateTransferDigest *) override;
   bool putBlock(const uint64_t blockId, const char *blockData, const uint32_t blockSize) override;
   uint64_t getLastReachableBlockNum() const override;
+  // This method is used by state-transfer in order to find the latest block id in either the state-transfer chain or
+  // the main blockchain
   uint64_t getLastBlockNum() const override;
   /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
--- a/kvbc/include/categorization/blockchain.h
+++ b/kvbc/include/categorization/blockchain.h
@@ -97,8 +97,8 @@ class Blockchain {
                  "Created [" << detail::ST_CHAIN_CF << "] column family for the state transfer blockchain");
       }
       loadLastBlockId();
-      if (last_block_id_) {
-        LOG_INFO(CAT_BLOCK_LOG, "State transfer last block id: " << last_block_id_.value());
+      if (last_block_id_ > 0) {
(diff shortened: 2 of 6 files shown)


andrewjstone (commented):
While this looks fine, I really dislike the idea of making a class specifically available to be used by multiple threads and then trusting those threads to "do the right thing" and not use any write methods at the "wrong time".

I think it would be better to provide an entirely separate class that only has access to the read API and then sharing that across threads. The sole thread that does writing gets access to the write methods. This prevents mistakes at compile time right now, but more importantly it prevents someone who is unaware of these constraints from doing the wrong thing in a future commit.

cloudnoize:
> While this looks fine, I really dislike the idea of making a class specifically available to be used by multiple threads and then trusting those threads to "do the right thing" and not use any write methods at the "wrong time".
> 
> I think it would be better to provide an entirely separate class that only has access to the read API and then sharing that across threads. The sole thread that does writing gets access to the write methods. This prevents mistakes at compile time right now, but more importantly it prevents someone who is unaware of these constraints from doing the wrong thing in a future commit.

@andrewjstone the usage of this class is exposed by read and write interfaces, which if I understand correctly is what you've described.
But state-transfer and block-appender are performing writes from different threads, therefore till we have proper architecture (channels ? ), in order to avoid locking, we must consider their usage and knowledge that both operations can't be performed in parallel because of the "competing" nature of state-transfer and block appender.

andrewjstone:
> > While this looks fine, I really dislike the idea of making a class specifically available to be used by multiple threads and then trusting those threads to "do the right thing" and not use any write methods at the "wrong time".
> > I think it would be better to provide an entirely separate class that only has access to the read API and then sharing that across threads. The sole thread that does writing gets access to the write methods. This prevents mistakes at compile time right now, but more importantly it prevents someone who is unaware of these constraints from doing the wrong thing in a future commit.
> 
> @andrewjstone the usage of this class is exposed by read and write interfaces, which if I understand correctly is what you've described.
> But state-transfer and block-appender are performing writes from different threads, therefore till we have proper architecture (channels ? ), in order to avoid locking, we must consider their usage and knowledge that both operations can't be performed in parallel because of the "competing" nature of state-transfer and block appender.

Ok, if I follow you are saying that state transfer and block appender are already writing from multiple threads, and this just makes the read interface thread safe. I believe you are right. In that case this is strictly an improvement and I withdraw my objections. Thanks!

andrewjstone (approved):
LGTM after discussion.

PR #1231, opened 2021-02-24
Enhance State Transfer Status for concord-ctl

saiakhil2012:
Enhancing the State Transfer Status for `concord-ctl` by,
- Making the output JSON based
- Adding more details related to General State Transfer Metrics
- Adding more details when the replica is in Collecting State

Tested the code changes, by manually inducing a replica to go into state-transfer scenario

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -675,9 +675,9 @@ void BCStateTran::onTimerImp() {
 std::string BCStateTran::getStatus() {
   std::ostringstream oss;
   std::unordered_map<std::string, std::string> result, nested_data;
-  result.insert(toPair("Fetching State", stateName(getFetchingState())));
+  result.insert(toPair("fetchingState", stateName(getFetchingState())));
   result.insert(toPair(getName(lastMsgSeqNum_), lastMsgSeqNum_));
-  result.insert(toPair("cacheOfVirtualBlockForResPages Size", cacheOfVirtualBlockForResPages.size()));
+  result.insert(toPair("cacheOfVirtualBlockForResPagesSize", cacheOfVirtualBlockForResPages.size()));
 
   auto current_source = sourceSelector_.currentReplica();
   auto preferred_replicas = sourceSelector_.preferredReplicasToString();
@@ -686,7 +686,7 @@ std::string BCStateTran::getStatus() {
     nested_data.insert(toPair(std::to_string(id), seq_num));
   }
 
-  result.insert(toPair("Last Msg Sequence Numbers (Replica ID: SeqNum)",
+  result.insert(toPair("LastMsgSequenceNumbers(ReplicaID:SeqNum)",
                        concordUtils::kvContainerToJson(nested_data, [](const auto &arg) { return arg; })));
   nested_data.clear();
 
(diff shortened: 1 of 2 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -35,7 +36,10 @@
 // TODO(GG): for debugging - remove
 // #define DEBUG_SEND_CHECKPOINTS_IN_REVERSE_ORDER (1)
 
+#define getName(var) #var
```
robem, line 39:
```suggestion
#define VARSTR(var) #var
```
or
```suggestion
#define NAME2STR(var) #var
```

robem (approved):
Lgtm. Thanks for the enhancement!

bftengine/src/bcstatetransfer/BCStateTran.cpp
andrewjstone: Looking at the rest of the code, I see that we only use this when we call `toPair(getName(var), var)`. That gets quite repetitive. I suggest turning the whole thing into a macro such that you just write `STRPAIR(var)` or something like that.

andrewjstone (approved):
This is great! Thanks @saiakhil2012. I only have one minor suggestion.

glevkovich (approved):
Looks good @saiakhil2012.
Just one comment: we should consider adding also the FetchBlockMsg <-> ItemDataMsg latencies. This is recorded by the BCStateTran::histograms_.
It is already possible to get it with concord-ctl from the command line, by calling:
> concord-ctl perf snapshot state_transfer
and then 
>concord-ctl perf get fetch_blocks_msg_latency
But maybe it can be nice to have it printed as part of the full status update?

bftengine/src/bcstatetransfer/BCStateTran.cpp
saiakhil2012: Yeah, I agree, the use of `toPair(getName(var), var)` is quite repetitive here. Sure, as suggested, I moved to using macro. Thanks

saiakhil2012:
> Looks good @saiakhil2012.
> Just one comment: we should consider adding also the FetchBlockMsg <-> ItemDataMsg latencies. This is recorded by the BCStateTran::histograms_.
> It is already possible to get it with concord-ctl from the command line, by calling:
> 
> ```
> concord-ctl perf snapshot state_transfer
> concord-ctl perf get fetch_blocks_msg_latency
> ```
> 
> But maybe it can be nice to have it printed as part of the full status update?

@glevkovich , I understand it would be good to have all together, but as the same content is already available via `perf` commands of `concord-ctl`, then instead of we repeating content across multiple commands, we can combine output of multiple commands like, `{command1 & command2;}`. In this way, we can combine any histogram with any status on need basis while performing diagnostics. Also, apart from the above thought, the latencies would be in the form of histogram which is what we are providing in `perf` commands, whereas in `status` we are focusing on status which is more of key-value pair format? Any thoughts?

teoparvanov:
> 
> 
> > Looks good @saiakhil2012.
> > Just one comment: we should consider adding also the FetchBlockMsg <-> ItemDataMsg latencies. This is recorded by the BCStateTran::histograms_.
> > It is already possible to get it with concord-ctl from the command line, by calling:
> > ```
> > concord-ctl perf snapshot state_transfer
> > concord-ctl perf get fetch_blocks_msg_latency
> > ```
> > 
> > 
> > But maybe it can be nice to have it printed as part of the full status update?
> 
> @glevkovich , I understand it would be good to have all together, but as the same content is already available via `perf` commands of `concord-ctl`, then instead of we repeating content across multiple commands, we can combine output of multiple commands like, `{command1 & command2;}`. In this way, we can combine any histogram with any status on need basis while performing diagnostics. Also, apart from the above thought, the latencies would be in the form of histogram which is what we are providing in `perf` commands, whereas in `status` we are focusing on status which is more of key-value pair format? Any thoughts?

I agree with @saiakhil2012 , let's not duplicate the information here.

teoparvanov (approved):
LGTM, thank you @saiakhil2012 !

PR #1232, opened 2021-02-24
Add root hash to raw block

cloudnoize:
Adding the root hash of each category instance to the raw block.

note:
Some category has the root hash defined as a std::optional,
if the optional is set the category root hash will be present in the raw block,
otherwise, it won't i.e. the raw block root hash is not using std::optional.

### DIFF
--- a/kvbc/cmf/categorized_kvbc_msgs.cmf
+++ b/kvbc/cmf/categorized_kvbc_msgs.cmf
@@ -92,7 +92,10 @@ Msg BlockData 2001 {
 Msg RawBlockData 2005 {
     fixedlist uint8 32 parent_digest
     CategoryInput updates
-    map string optional fixedlist uint8 32 category_root_hash
+    map string fixedlist uint8 32 block_merkle_root_hash
+    map string fixedlist uint8 32 versioned_root_hash 
+    map string map string fixedlist uint8 32 immutable_root_hashes
+
 }
 
 # Used to deserialize the parent_digest only from RawBlockData and BlockData.
--- a/kvbc/include/categorization/blocks.h
+++ b/kvbc/include/categorization/blocks.h
@@ -122,6 +122,21 @@ struct RawBlock {
   RawBlockData data;
 };
 
+inline void addRootHash(const std::string& category_id, RawBlockData& data, const BlockMerkleOutput& update_info) {
+  data.block_merkle_root_hash[category_id] = update_info.root_hash;
+}
+
+// If optional is null do not include in the raw block
+inline void addRootHash(const std::string& category_id, RawBlockData& data, const VersionedOutput& update_info) {
+  if (!update_info.root_hash) return;
+  data.versioned_root_hash[category_id] = update_info.root_hash.value();
+}
+
+inline void addRootHash(const std::string& category_id, RawBlockData& data, const ImmutableOutput& update_info) {
+  if (!update_info.tag_root_hashes) return;
+  data.immutable_root_hashes[category_id] = update_info.tag_root_hashes.value();
+}
+
 inline bool operator==(const RawBlock& l, const RawBlock& r) { return l.data == r.data; }
 inline bool operator!=(const RawBlock& l, const RawBlock& r) { return !(l == r); }
 
--- a/kvbc/test/categorization/kv_blockchain_test.cpp
+++ b/kvbc/test/categorization/kv_blockchain_test.cpp
@@ -809,7 +809,7 @@ TEST_F(categorized_kvbc, compare_raw_blocks) {
   block_chain.addBlock(std::move(up));
 
   KeyValueBlockchain::KeyValueBlockchain_tester tester;
-  auto last_raw = tester.getLastRawBlocked(block_chain);
+  auto last_raw = tester.getLastRawBlock(block_chain);
   ASSERT_EQ(last_raw.first, 1);
   ASSERT_TRUE(last_raw.second.has_value());
   auto raw_from_api = block_chain.getRawBlock(1);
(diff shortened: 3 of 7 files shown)


kvbc/cmf/categorized_kvbc_msgs.cmf
```diff
@@ -92,7 +92,10 @@ Msg BlockData 2001 {
 Msg RawBlockData 2005 {
     fixedlist uint8 32 parent_digest
     CategoryInput updates
-    map string optional fixedlist uint8 32 category_root_hash
+    map string fixedlist uint8 32 block_merkle_root_hash
+    map string fixedlist uint8 32 versioned_root_hash 
+    map string map string fixedlist uint8 32 immutable_root_hashes
```
dartdart26, line 97: I think I didn't get the comment about optional hashes. Could you please explain it?

kvbc/include/categorization/blocks.h
```diff
@@ -122,6 +122,24 @@ struct RawBlock {
...
+// If optional is null do not include in the raw block
+inline void addRootHash(const std::string& category_id, RawBlockData& data, const VersionedOutput& update_info) {
+  if (!update_info.root_hash) return;
+  data.versioned_root_hash[category_id] = update_info.root_hash.value();
+}
+
+inline void addRootHash(const std::string& category_id, RawBlockData& data, const ImmutableOutput& update_info) {
+  if (!update_info.tag_root_hashes) return;
+  auto& imm_map = data.immutable_root_hashes[category_id];
+  for (const auto& [tag, hash] : update_info.tag_root_hashes.value()) {
```
dartdart26, line 138: Can we just assign `*update_info.tag_root_hashes` to `imm_map` to avoid the loop?

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -1841,6 +1841,188 @@ TEST_F(categorized_kvbc, deletes_ordered_on_add_to_updates) {
...
+    ASSERT_EQ(rb->data.versioned_root_hash.count("versioned"), 1);
+    ASSERT_TRUE(rb->data.immutable_root_hashes.empty());
+    KeyValueBlockchain::KeyValueBlockchain_tester tester;
+    auto cached_rb = tester.getLastRawBlocked(block_chain).second;
+    ASSERT_TRUE(cached_rb.has_value());
+    const auto& ser_rb = categorization::detail::serializeThreadLocal(rb->data);
+    std::string str_rb(ser_rb.begin(), ser_rb.end());
+    const auto& ser_cached_rb = categorization::detail::serializeThreadLocal(*cached_rb);
+    std::string str_cached_rb(ser_cached_rb.begin(), ser_cached_rb.end());
+    ASSERT_EQ(std::hash<std::string>{}(str_rb), std::hash<std::string>{}(str_cached_rb));
```
dartdart26, line 1893: Can we just compare the raw blocks instead of calling std::hash?

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -1841,6 +1841,188 @@ TEST_F(categorized_kvbc, deletes_ordered_on_add_to_updates) {
...
+
+    KeyValueBlockchain::KeyValueBlockchain_tester tester;
+    auto cached_rb = tester.getLastRawBlocked(block_chain).second;
+    ASSERT_TRUE(cached_rb.has_value());
+    const auto& ser_rb = categorization::detail::serializeThreadLocal(rb->data);
+    std::string str_rb(ser_rb.begin(), ser_rb.end());
+    const auto& ser_cached_rb = categorization::detail::serializeThreadLocal(*cached_rb);
+    std::string str_cached_rb(ser_cached_rb.begin(), ser_cached_rb.end());
+    ASSERT_EQ(std::hash<std::string>{}(str_rb), std::hash<std::string>{}(str_cached_rb));
+    ASSERT_NE(std::hash<std::string>{}(str_rb), std::hash<std::string>{}(prev_digest));
```
dartdart26, line 1946: I am a bit confused by this assert. Maybe `prev_digest` is to be called `prev_block`?

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -1841,6 +1841,188 @@ TEST_F(categorized_kvbc, deletes_ordered_on_add_to_updates) {
...
+    ASSERT_TRUE(rb.has_value());
+    ASSERT_EQ(rb->data.block_merkle_root_hash.size(), 2);
+    ASSERT_EQ(rb->data.block_merkle_root_hash.count("merkle"), 1);
+    ASSERT_EQ(rb->data.block_merkle_root_hash.count("merkle2"), 1);
+
+    ASSERT_EQ(rb->data.versioned_root_hash.size(), 0);
+    ASSERT_EQ(rb->data.immutable_root_hashes.size(), 0);
+
+    KeyValueBlockchain::KeyValueBlockchain_tester tester;
+    auto cached_rb = tester.getLastRawBlocked(block_chain).second;
```
dartdart26, line 1974: Nit: I just noticed this is a typo - `Blocked`.

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -1841,6 +1841,188 @@ TEST_F(categorized_kvbc, deletes_ordered_on_add_to_updates) {
...
+    ImmutableUpdates immutable_updates3;
+    immutable_updates3.calculateRootHash(true);
+    immutable_updates3.addUpdate("immutable_key4_3", {"immutable_val1", {}});
+    updates.add("immutable3", std::move(immutable_updates3));
+
+    ASSERT_EQ(block_chain.addBlock(std::move(updates)), (BlockId)4);
+
+    auto rb = block_chain.getRawBlock(4);
+    ASSERT_EQ(rb->data.immutable_root_hashes.size(), 2);
+    ASSERT_EQ(rb->data.immutable_root_hashes.count("immutable3"), 1);
```
dartdart26, line 2005: Should we check that `immutable3` doesn't have any tags here?

kvbc/test/categorization/kv_blockchain_test.cpp
```diff
@@ -1841,6 +1841,188 @@ TEST_F(categorized_kvbc, deletes_ordered_on_add_to_updates) {
...
+
+    const auto& imm_map = rb->data.immutable_root_hashes["immutable"];
+    ASSERT_EQ(imm_map.size(), 2);
+    ASSERT_EQ(imm_map.count("1"), 1);
+    ASSERT_EQ(imm_map.count("2"), 1);
+
+    const auto& imm_map2 = rb->data.immutable_root_hashes["immutable3"];
```
dartdart26, line 2012: Ah, we do check (from a previous comment). Maybe the variable can be called `imm_map3`?

kvbc/test/categorization/kv_blockchain_test.cpp
dartdart26: We do check later...

kvbc/cmf/categorized_kvbc_msgs.cmf
cloudnoize: I think it's better to include the hashes only in case they are really calculated i.e. not nullopt.
It will save data on the wire on ST and is more compact.

kvbc/test/categorization/kv_blockchain_test.cpp
cloudnoize: compare doesn't take into account ordering, I want to detect non-determinism

kvbc/include/categorization/blocks.h
cloudnoize: I guess we can

kvbc/test/categorization/kv_blockchain_test.cpp
cloudnoize: I just wanted to make sure I really calculate something, it's a bit silly I can remove it

kvbc/cmf/categorized_kvbc_msgs.cmf
dartdart26: Thanks! Yes, I got it. I was thinking we are using these for reconstruction on ST, but we are not.

kvbc/test/categorization/kv_blockchain_test.cpp
dartdart26: I meant can't we compare the serialized blocks (the std::string objects), something like:
```
ASSERT_EQ(str_rb, str_cached_rb);
```

dartdart26 (approved):
Looks good, thanks @cloudnoize !

kvbc/test/categorization/kv_blockchain_test.cpp
cloudnoize: :)

PR #1233, opened 2021-02-25
Encrypted configuration implementation

tdimitrov:
This is a combined PR for SecretsManager and SecretsRetriever (done by @KTeofanidis).

The purpose of the PR is to add implementation for encryption/decryption of configuration files and secrets retrieval from a file on the local file system (used for development tests) or REST API.

SecretsManager is a singleton, which wraps implementation doing the actual work. It's singleton because:
1. During early initialisation of the process one needs to know if encrypted configuration is enabled or not. It's more convenient to initialise a single SecretsManager instance and use it throughout the code.
2. To avoid the need each part of the code to check if encrypted configuration is used or not when working with files.

SecretsRetriever is responsible for fetching details about the encryption algorithm which is used to encrypt the configuration, the secret key + its size, digest algorithm and so on. As already mentioned it can work with file on the local filesystem or remote REST API.

Currently openssl's OPENSSL_EVP_BytesToKey() function is used as key derivation algorithm. 
AES256 CBC is used for encryption/decryption.

For further implementation details please check the comments in the code.


secretsmanager/src/SecretRetriever.cpp
```diff
@@ -0,0 +1,91 @@
...
+    if ("http"s.compare(url_match[1]) == 0) {
+      int p = 80;
+      if (!port.empty()) p = stoi(port);
+      httplib::Client cli(host, p);
+      httpres = cli.Get(path.c_str());
+    } else {
+      int p = 443;
+      if (!port.empty()) p = stoi(port);
+      httplib::SSLClient cli(host, p);
+      cli.enable_server_certificate_verification(false);
```
andrewjstone, line 66: Why wouldn't we verify the server cert? Doesn't this open us up to a MIIM attack?

Is this because the server doesn't use PKI and we don't have a way to load the server public key locally first?

secretsmanager/src/misc.h
```diff
@@ -0,0 +1,69 @@
+// This implementation is copied from CryptoPP: https://www.cryptopp.com/wiki/OPENSSL_EVP_BytesToKey
```
andrewjstone, line 1: Why are we copying this?

secretsmanager/src/openssl_pass.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <string>
+#include <vector>
+
+#include <cryptopp/cryptlib.h>
+#include <cryptopp/sha.h>
+#include "misc.h"
+
+namespace concord::secretsmanager {
+
+KeyParams derive_key_pass(const std::string_view pass,
```
andrewjstone, line 27: nit: function names should follow `camelCase` pattern.

secretsmanager/src/secrets_manager_enc.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "secrets_manager_enc.h"
+
+#include "aes.h"
+#include "openssl_pass.h"
+#include "base64.h"
+
+namespace concord::secretsmanager {
+
+SecretsManagerEnc::SecretsManagerEnc(SecretRetriever::SecretData& secrets) {
+  if (supported_digets_.find(secrets.digest) == supported_digets_.end()) {
```
andrewjstone, line 25: typo: `supported_digets` -> `supported_digests`

andrewjstone (commented):
@tdimitrov @KTeofanidis This code looks really good! It's some of the cleanest code I've reviewed in a while for such a major feature. You guys are giving @dartdart26 a run for his money! My comments are all pretty minor.

However, it may worthwhile to discuss the solution with VMware security team and/or @ittaiab if you haven't already as I'm no expert in this stuff and it's easy to make mistakes.

tests/secrets_api_mock/service.py
```diff
@@ -0,0 +1,18 @@
+import json
+from flask import Flask, request
+
+app = Flask(__name__)
```
f-squirrel, line 4: I like it!

tests/secrets_api_mock/requirements.txt
```diff
@@ -0,0 +1,7 @@
+click==7.1.2
```
f-squirrel, line 1: Is it supposed to be installed in the docker image?

util/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ target_include_directories(util PUBLIC include)
 
 target_link_libraries(util_shared PUBLIC Threads::Threads)
 target_include_directories(util_shared PUBLIC include)
-target_include_directories(util_shared PUBLIC ${CRYPTOPP_INCLUDE_DIRS})
+target_include_directories(util_shared PUBLIC)
```
f-squirrel, line 32: I am not sure I understand it: it seems like we add an empty string to the includes.

toly-kournik:
@tdimitrov, @KTeofanidis thanks for the PR!

As far as I understand, the secret data decryption process includes three main steps:
1. Secret key retrieval using TPM (SAAS team);
2. Encrypted data retrieval (local or remote file specified by URL) - `FileSecretRetrieve` and `RemoteSecretRetrieve`;
3. Encrypted data decryption using the key from  1. - `ISecretsManagerImpl & Sons` 

There's one thing though I have to clarify.
I'm not sure where the requirement on REST API came from, I haven't seen it in a design documents.
Also I haven't seen any requirement about using JSON as an underlying format.

In order to support all possible options, including mentioned above, as well as to avoid adding dependencies to concord-bft (httplib and nlohmann/json), I propose a full separation between secret data retrieval and decryption on an interface level:
1.  `ISecretManagerImpl` with its successors are already defined, although I'd rename them into `ISecretManager`, `SecretManagerAES` and `SecretManagerPlain` correspondingly;
2. Define `ISecretRetriever` interface with an abstract method like `SecretData get(string URL);`
2. `FileSecretRetriever` will implement it in concord-bft (as per proposed PR); 
3. `RemoteSecretRetriever` will implement it in concord (including httplib deps if decided, which are already there).
4. TPM interface should also be defined in concord-bft and implemented in concord.

Secret data format is application dependent, i.e. is opaque to a SecretManager and known by its clients - `KeyManager`, etc.

andrewjstone:
> @tdimitrov, @KTeofanidis thanks for the PR!
> 
> As far as I understand, the secret data decryption process includes three main steps:
> 
> 1. Secret key retrieval using TPM (SAAS team);
> 2. Encrypted data retrieval (local or remote file specified by URL) - `FileSecretRetrieve` and `RemoteSecretRetrieve`;
> 3. Encrypted data decryption using the key from  1. - `ISecretsManagerImpl & Sons`
> 
> There's one thing though I have to clarify.
> I'm not sure where the requirement on REST API came from, I haven't seen it in a design documents.
> Also I haven't seen any requirement about using JSON as an underlying format.
> 
> In order to support all possible options, including mentioned above, as well as to avoid adding dependencies to concord-bft (httplib and nlohmann/json), I propose a full separation between secret data retrieval and decryption on an interface level:
> 
> 1. `ISecretManagerImpl` with its successors are already defined, although I'd rename them into `ISecretManager`, `SecretManagerAES` and `SecretManagerPlain` correspondingly;
> 2. Define `ISecretRetriever` interface with an abstract method like `SecretData get(string URL);`
> 3. `FileSecretRetriever` will implement it in concord-bft (as per proposed PR);
> 4. `RemoteSecretRetriever` will implement it in concord (including httplib deps if decided, which are already there).
> 5. TPM interface should also be defined in concord-bft and implemented in concord.
> 
> Secret data format is application dependent, i.e. is opaque to a SecretManager and known by its clients - `KeyManager`, etc.

The separation of the interfaces seems like a good suggestion to me. However, I'd like to avoid moving anything to closed source if we can. The goal moving forward should be to open source everything related to concord unless its proprietary and non-generic. There is no fundamental reason why we can't have a generic http retriever based on json in concord-bft. If you want to prevent it from being required, make it into it's own library that is separately linkable. Moving things into concord will only result in more coupling with closed source code and difficulty when we decide to move this code back to open source. 

Basically, unless it has to be closed source for licensing reasons, it should live here. How we structure the code to not mandate dependencies on users is up for debate. But moving things to closed source because they use code we didn't write is silly IMO.

andrewjstone:
> > @tdimitrov, @KTeofanidis thanks for the PR!
> > As far as I understand, the secret data decryption process includes three main steps:
> > 
> > 1. Secret key retrieval using TPM (SAAS team);
> > 2. Encrypted data retrieval (local or remote file specified by URL) - `FileSecretRetrieve` and `RemoteSecretRetrieve`;
> > 3. Encrypted data decryption using the key from  1. - `ISecretsManagerImpl & Sons`
> > 
> > There's one thing though I have to clarify.
> > I'm not sure where the requirement on REST API came from, I haven't seen it in a design documents.
> > Also I haven't seen any requirement about using JSON as an underlying format.
> > In order to support all possible options, including mentioned above, as well as to avoid adding dependencies to concord-bft (httplib and nlohmann/json), I propose a full separation between secret data retrieval and decryption on an interface level:
> > 
> > 1. `ISecretManagerImpl` with its successors are already defined, although I'd rename them into `ISecretManager`, `SecretManagerAES` and `SecretManagerPlain` correspondingly;
> > 2. Define `ISecretRetriever` interface with an abstract method like `SecretData get(string URL);`
> > 3. `FileSecretRetriever` will implement it in concord-bft (as per proposed PR);
> > 4. `RemoteSecretRetriever` will implement it in concord (including httplib deps if decided, which are already there).
> > 5. TPM interface should also be defined in concord-bft and implemented in concord.
> > 
> > Secret data format is application dependent, i.e. is opaque to a SecretManager and known by its clients - `KeyManager`, etc.
> 
> The separation of the interfaces seems like a good suggestion to me. However, I'd like to avoid moving anything to closed source if we can. The goal moving forward should be to open source everything related to concord unless its proprietary and non-generic. There is no fundamental reason why we can't have a generic http retriever based on json in concord-bft. If you want to prevent it from being required, make it into it's own library that is separately linkable. Moving things into concord will only result in more coupling with closed source code and difficulty when we decide to move this code back to open source.
> 
> Basically, unless it has to be closed source for licensing reasons, it should live here. How we structure the code to not mandate dependencies on users is up for debate. But moving things to closed source because they use code we didn't write is silly IMO.

Thinking a bit more about this, if the data and tpm usage is indeed tied to our closed source (SAAS) implementation, than maybe @toly-kournik is correct about keeping that part of the solution closed source, as it's not applicable otherwise. My only point in commenting was to try to prevent code that is useful to the community from being put into closed source for reasons of dependency management. Regardless, this is a good discussion to have, as Toly has pointed out, it's not exactly clear where the boundaries are.

toly-kournik:
@andrewjstone, I agree that the open/close source boundaries are unclear. As for me, a large part of concord could be an open source as well, probably in a separate repository, and probably as a reference app implementation. 
I was talking more about core/app separation. Providing proper interfaces in concord-bft will allow easy future implementations with seamless integration.

secretsmanager/src/misc.h
tdimitrov: By the initial design we use a password based key derivation algorithm from openssl which is now deprecated. I'm discussing this with Ittai and Sanchita and I hope we will remove this.

About your question - the Crypto++ folks say "This algo is depricated, so we have got no implementation. If you insist on using it - here is a header which you can put in your source tree." 

Even this warning sounds scary enough to me so we change the key derivation algorithm.

secretsmanager/src/openssl_pass.cpp
tdimitrov: My bad. I always mess up the code style :(

secretsmanager/src/base64.h
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace concord::secretsmanager {
+
+struct SaltedCipher {
+  std::vector<uint8_t> salt;
+  std::vector<uint8_t> ct;
```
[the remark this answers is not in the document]
tdimitrov: ct is cipher text. I'm a bit biased here - is a comment enough or better rename all instances?

tests/secrets_api_mock/requirements.txt
tdimitrov: No, we don't need it for now.

util/CMakeLists.txt
tdimitrov: The whole line is meant to be deleted.

secretsmanager/include/secrets_manager.h
```diff
@@ -0,0 +1,73 @@
...
+
+  std::unique_ptr<ISecretsManagerImpl> impl_;
+  SecretsManager(std::unique_ptr<ISecretsManagerImpl> impl) : impl_{std::move(impl)} {}
+
+ public:
+  static SecretsManager& get() {
+    if (instance)
+      return *instance;
+    else
+      throw std::runtime_error("SecretsManager is not initialised");
```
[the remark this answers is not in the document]
tdimitrov: Here me and @KTeofanidis  made an unfair assumption (unfair, because it is in our heads only). The SecretsManager purpose is to encrypt configuration. The process can't do much until its configuration is decrypted, so init() will be  called before get().

But I agree that we might introduce a nasty issues in the future in this code.

Considering all the feedback - we'll remove the singleton. I'll update the PR later today.

secretsmanager/include/secrets_manager.h
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  static void init(bool secrets_manager_mode, const std::string& secrets_url) {
```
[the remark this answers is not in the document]
tdimitrov: Actually this is an excellent idea.

But as I already mentioned - we'll probably remove the singleton altogether.

secretsmanager/include/secrets_manager_enc.h
```diff
@@ -0,0 +1,66 @@
...
+#include <cryptopp/osrng.h>
+
+#include "Logger.hpp"
+
+#include "SecretRetriever.hpp"
+#include "secrets_manager_impl.h"
+
+namespace concord::secretsmanager {
+
+// SecretsManagerEnc handles encryption and decryption of files.
```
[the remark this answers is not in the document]
tdimitrov: He seemed to approve the general solution on the design review, but I one way or another - I want to talk about the openssl password key derivation with him. I'll ask him about the final solution.

secretsmanager/src/SecretRetriever.cpp
```diff
@@ -0,0 +1,91 @@
...
+    json j;
+    jsonFile >> j;
+    return ParseJson(j);
+  } else {
+    return std::nullopt;
+  }
+}
+
+std::optional<SecretData> RemoteSecretRetrieve(const std::string &secret_url) {
+  std::regex url_regex(R"(^(https|http)://([a-zA-Z0-9\.-]+)(:([0-9]+))?(.*))", std::regex::extended);
```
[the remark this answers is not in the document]
tdimitrov: We decided to allow http for two main reasons:

1. We have got a mock for the Secrets API which currently doesn't support HTTPS. We want to add some apollo tests and this will make our life easier.
2. The Secrets API in production will work on localhost, so there might be a case where someone might want to use HTTP instead of HTTPS.

Or in nutshell - we deliberately decided to let our users shoot theirselves in the foot if they insist.

secretsmanager/src/SecretRetriever.cpp
tdimitrov: I've added my comments below.

> @tdimitrov, @KTeofanidis thanks for the PR!
> 
> As far as I understand, the secret data decryption process includes three main steps:
> 
> 1. Secret key retrieval using TPM (SAAS team);
> 2. Encrypted data retrieval (local or remote file specified by URL) - `FileSecretRetrieve` and `RemoteSecretRetrieve`;

FileSecretRetrieve will be used only for tests. It's not intended for production use. We do can (and will) use the mock of the service for tests, but there was a hard requirement for a file interface too.

> 3. Encrypted data decryption using the key from  1. - `ISecretsManagerImpl & Sons`
> 
> There's one thing though I have to clarify.
> I'm not sure where the requirement on REST API came from, I haven't seen it in a design documents.
> Also I haven't seen any requirement about using JSON as an underlying format.

Both requirements are in the design document. SaaS will provide REST API running as a sidecar container and it will return JSON.

> In order to support all possible options, including mentioned above, as well as to avoid adding dependencies to concord-bft (httplib and nlohmann/json), I propose a full separation between secret data retrieval and decryption on an interface level:
> 
> 1. `ISecretManagerImpl` with its successors are already defined, although I'd rename them into `ISecretManager`, `SecretManagerAES` and `SecretManagerPlain` correspondingly;

This was my initial approach, but unfortunately the decryption process is not that simple. It combines key derivation and encryption, which are two separate algorithms. For this reason there is a single class which encapsulates different encryption and digest algorithms.

> 2. Define `ISecretRetriever` interface with an abstract method like `SecretData get(string URL);`
> 3. `FileSecretRetriever` will implement it in concord-bft (as per proposed PR);
> 4. `RemoteSecretRetriever` will implement it in concord (including httplib deps if decided, which are already there).

We can do this but I think it's not needed. At the moment the Retriever and the SecretsManagerEnc work as a pipeline. One initialises the retriever, gets the secrets parameter and passes them to the SecretsManagerEnc. So the way secrets are retrieved is fully decoupled from SecretsManagerEnc.

But even if we introduce interfaces - we'll have to provide implementations for REST/File endpoint, so we will need the dependencies somewhere in the project.

> 5. TPM interface should also be defined in concord-bft and implemented in concord.
The whole TPM iface is done by the SaaS team. Accessing TPM from virtual machine is hard, so they handle this from the outside.

> Secret data format is application dependent, i.e. is opaque to a SecretManager and known by its clients - `KeyManager`, etc.

tdimitrov:
I've added my comments below.

> @tdimitrov, @KTeofanidis thanks for the PR!
> 
> As far as I understand, the secret data decryption process includes three main steps:
> 
> 1. Secret key retrieval using TPM (SAAS team);
> 2. Encrypted data retrieval (local or remote file specified by URL) - `FileSecretRetrieve` and `RemoteSecretRetrieve`;

FileSecretRetrieve will be used only for tests. It's not intended for production use. We do can (and will) use the mock of the service for tests, but there was a hard requirement for a file interface too.

> 3. Encrypted data decryption using the key from  1. - `ISecretsManagerImpl & Sons`
> 
> There's one thing though I have to clarify.
> I'm not sure where the requirement on REST API came from, I haven't seen it in a design documents.
> Also I haven't seen any requirement about using JSON as an underlying format.

Both requirements are in the design document. SaaS will provide REST API running as a sidecar container and it will return JSON.

> In order to support all possible options, including mentioned above, as well as to avoid adding dependencies to concord-bft (httplib and nlohmann/json), I propose a full separation between secret data retrieval and decryption on an interface level:
> 
> 1. `ISecretManagerImpl` with its successors are already defined, although I'd rename them into `ISecretManager`, `SecretManagerAES` and `SecretManagerPlain` correspondingly;

This was my initial approach, but unfortunately the decryption process is not that simple. It combines key derivation and encryption, which are two separate algorithms. For this reason there is a single class which encapsulates different encryption and digest algorithms.

> 2. Define `ISecretRetriever` interface with an abstract method like `SecretData get(string URL);`
> 3. `FileSecretRetriever` will implement it in concord-bft (as per proposed PR);
> 4. `RemoteSecretRetriever` will implement it in concord (including httplib deps if decided, which are already there).

We can do this but I think it's not needed. At the moment the Retriever and the SecretsManagerEnc work as a pipeline. One initialises the retriever, gets the secrets parameter and passes them to the SecretsManagerEnc. So the way secrets are retrieved is fully decoupled from SecretsManagerEnc.

But even if we introduce interfaces - we'll have to provide implementations for REST/File endpoint, so we will need the dependencies somewhere in the project.

> 5. TPM interface should also be defined in concord-bft and implemented in concord.
The whole TPM iface is done by the SaaS team. Accessing TPM from virtual machine is hard, so they handle this from the outside.

> Secret data format is application dependent, i.e. is opaque to a SecretManager and known by its clients - `KeyManager`, etc.

secretsmanager/src/base64.h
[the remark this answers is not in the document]
f-squirrel: I am biased too, I think naming is more efficient in terms of documentation than comments.
P.S. not critical, I was wondering what it was.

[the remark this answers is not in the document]
tdimitrov: I renamed it.

secretsmanager/include/secrets_manager_enc.h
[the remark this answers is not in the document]
tdimitrov: I spoke with @ittaiab. This solution is good enough for initial tests. Later the key generation part will be removed as it reduces the entropy of the symmetric key.

secretsmanager/src/SecretRetriever.cpp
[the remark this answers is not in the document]
tdimitrov: Small update here - Secrets API will listen on HTTP by design, so we need to support it.

secretsmanager/src/misc.h
tdimitrov: Also we won't need this header once we remove the key generation step.

secretsmanager/src/SecretRetriever.cpp
tdimitrov: We have no way to obtain the Secrets API certificate in secure manner. HTTPS will not even be used in real deployments.

tdimitrov:
One more clarification about the last commits I put.

@toly-kournik insists on not adding http and json libraries as dependency in concord, so SecretsRetriever implementation won't be provided here. The only thing that remains is SecretData struct which is used as glue for the Retriever and the SecretsManagerEnc (output of the former, input for the latter).

Also the singleton is removed because it is too error prone. SecretsManagerEnc will be instantiated directly.

I removed the dependencies from the cmake files and I will clean the install_deps.sh script in separate PR.

secretsmanager/src/secrets_manager_impl.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <fstream>
+
+namespace concord::secretsmanager {
+
+std::string ISecretsManagerImpl::readFile(std::string_view path) {
+  std::string data;
+
+  std::string buf(4096, '0');
+  std::ifstream in(path.data());
+  while (in.read(buf.data(), buf.length())) {
```
andrewjstone, line 28: Why are you reading into `buf` and then copying into `data` rather than just reading into `buf` repeatedly and then returning it?

andrewjstone (approved):
LGTM.

secretsmanager/src/secrets_manager_impl.cpp
tdimitrov: Fixed

secretsmanager/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+    src/openssl_pass.cpp)
+
+add_library(secretsmanager STATIC 
+    ${secretsmanager_source_files})
+add_library(secretsmanager_shared SHARED
+    ${secretsmanager_source_files})
+
+find_package(OpenSSL REQUIRED)
+if(OPENSSL_FOUND)
+    set(HTTPLIB_IS_USING_OPENSSL TRUE)
```
toly-kournik, line 20: We can remove this

secretsmanager/include/secrets_manager_enc.h
```diff
@@ -0,0 +1,66 @@
...
+// The following flow is used for encryption:
+// 1. A random salt is generated
+// 2. The salt and the password (provided by the configuration) are used to derive a secret key and iv. OpenSSL
+//   OPENSSL_EVP_BytesToKey() is used for key derivation (openssl_pass.h).
+// 3. The key is used to encrypt the payload with a supported block cipher (AES256 CBC).
+// 4. The encrypted cipher text and the salt are concatenated.
+//    The contents of the buffer is: Salted__<8 bytes salt><cipher>
+// 5. The buffer is base64 encoded and written to file.
+class SecretsManagerEnc : public ISecretsManagerImpl {
+  std::string password_;
```
toly-kournik, line 40: Why not to use SecretData struct here?

secretsmanager/include/secrets_manager_enc.h
```diff
@@ -0,0 +1,66 @@
...
+  std::string digest_;
+  std::string algo_;
+  uint32_t key_length_;
+
+  const std::set<std::string> supported_encs_{"AES/CBC/PKCS5Padding", "AES/CBC/PKCS7Padding"};
+  const std::set<std::string> supported_digests_{"SHA-256"};
+  const uint32_t SALT_SIZE = 8;
+
+  logging::Logger logger = logging::getLogger("secrets-manager-enc");
```
toly-kournik, line 49: please use logname convention: concord.bdt.secret-manager

secretsmanager/src/base64.h
```diff
@@ -0,0 +1,31 @@
+// Concord
```
toly-kournik, line 1: Consider moving to utils

secretsmanager/src/base64.cpp
```diff
@@ -0,0 +1,63 @@
+// Concord
```
toly-kournik, line 1: consider moving to utils

secretsmanager/src/aes.h
```diff
@@ -0,0 +1,42 @@
+// Concord
```
toly-kournik, line 1: consider moving to utils

secretsmanager/src/aes.cpp
```diff
@@ -0,0 +1,44 @@
+// Concord
```
toly-kournik, line 1: consider moving to utils

secretsmanager/src/misc.h
toly-kournik, line 1: consider moving to utils

secretsmanager/src/secrets_manager_enc.cpp
tdimitrov: This contains some custom steps - Salted__ and the salt itself is prepended to the cipher text. That's why it is in src folder.

secretsmanager/src/aes.h
tdimitrov: This is just a convenience wrapper around Crypto++ AES code. No benefit to reuse it.

secretsmanager/src/misc.h
tdimitrov: This algorithm is deprecated. It will be removed in future PR.

PR #1234, opened 2021-02-25
Refactor pruning

yontyon:
In this PR we added pruning in concordbft (including tests)
Note that as Time service is not part of concordbdt yet, the basic pruning does not support pruning on time based

### DIFF
--- a/reconfiguration/include/reconfiguration/dispatcher.hpp
+++ b/reconfiguration/include/reconfiguration/dispatcher.hpp
@@ -38,8 +38,12 @@ class Dispatcher {
   concord::messages::ReconfigurationResponse dispatch(const concord::messages::ReconfigurationRequest&,
                                                       uint64_t sequence_num);
 
-  void addReconfigurationHandler(std::shared_ptr<IReconfigurationHandler> h) { reconfig_handlers_.push_back(h); }
-  void addPruningHandler(std::shared_ptr<IPruningHandler> h) { pruning_handlers_.push_back(h); }
+  void addReconfigurationHandler(std::shared_ptr<IReconfigurationHandler> h) {
+    if (h != nullptr) reconfig_handlers_.push_back(h);
+  }
+  void addPruningHandler(std::shared_ptr<IPruningHandler> h) {
+    if (h != nullptr) pruning_handlers_.push_back(h);
+  }
 
  private:
   logging::Logger getLogger() {
--- a/reconfiguration/include/reconfiguration/pruning_handler.hpp
+++ b/reconfiguration/include/reconfiguration/pruning_handler.hpp
@@ -17,22 +17,112 @@
 #include "concord.cmf.hpp"
 #include "ireconfiguration.hpp"
 #include "OpenTracing.hpp"
+#include "pruning_utils.hpp"
+namespace concord::reconfiguration::pruning {
 
-namespace concord::reconfiguration {
-
-class PruningHandler : public IPruningHandler {
- public:
-  bool handle(const concord::messages::LatestPrunableBlockRequest&, concord::messages::LatestPrunableBlock&) override {
-    return true;
+class DefaultPruningHandler : public IPruningHandler {
+  bool handle(const concord::messages::LatestPrunableBlockRequest &,
+              concord::messages::LatestPrunableBlock &) override {
+    LOG_ERROR(logger_, "Not implemented");
+    return false;
+  };
+  bool handle(const concord::messages::PruneStatusRequest &, concord::messages::PruneStatus &) override {
+    LOG_ERROR(logger_, "Not implemented");
+    return false;
   }
-  bool handle(const concord::messages::PruneStatusRequest&, concord::messages::PruneStatus&) override { return true; }
-  bool handle(const concord::messages::PruneRequest&, kvbc::BlockId&) override { return true; }
-
- protected:
-  logging::Logger getLogger() {
-    static logging::Logger logger_(logging::getLogger("concord.pruning"));
-    return logger_;
+  bool handle(const concord::messages::PruneRequest &, kvbc::BlockId &) override {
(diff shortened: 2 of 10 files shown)


reconfiguration/include/reconfiguration/pruning_handler.hpp
```diff
@@ -17,22 +17,111 @@
 #include "concord.cmf.hpp"
 #include "ireconfiguration.hpp"
 #include "OpenTracing.hpp"
+#include "pruning_utils.hpp"
+namespace concord::reconfiguration::pruning {
 
-namespace concord::reconfiguration {
+class DefaultPruningHandler : public IPruningHandler {
```
toly-kournik, line 23: I believe we don't need the DefaultPruningHandler

reconfiguration/include/reconfiguration/pruning_utils.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <type_traits>
+#include <unordered_set>
+#include <bftengine/IStateTransfer.hpp>
+#include <db_interfaces.h>
+#include <future>
+#include "Crypto.hpp"
+#include "bftengine/ControlStateManager.hpp"
+
+namespace concord::reconfiguration::pruning {
+class PruningException : public std::exception {
```
toly-kournik, line 27: If there's no special treatment for the pruning exceptions, we don't need them.
Just throw std::runtime_error

reconfiguration/include/reconfiguration/pruning_handler.hpp
```diff
@@ -17,22 +17,111 @@
...
-    return true;
-  }
-  bool handle(const concord::messages::PruneStatusRequest&, concord::messages::PruneStatus&) override { return true; }
-  bool handle(const concord::messages::PruneRequest&, kvbc::BlockId&) override { return true; }
+  // Construct by providing an interface to the storage engine, state transfer,
+  // configuration and tracing facilities. Note this constructor may throw an
+  // exception if there is an issue with the configuration (for example, if the
+  // configuration enables pruning but does not provide a purning operator
+  // public key).
+  PruningHandler(const kvbc::IReader &,
```
toly-kournik, line 87: Please convert all methods to start with a lowercase letter as per code conventions

reconfiguration/include/reconfiguration/pruning_utils.hpp
```diff
@@ -0,0 +1,133 @@
...
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include "concord.cmf.hpp"
+#include <boost/endian/buffers.hpp>
```
toly-kournik, line 16: Is it possible not to use boost here?

reconfiguration/include/reconfiguration/dispatcher.hpp
```diff
@@ -38,8 +38,12 @@ class Dispatcher {
   concord::messages::ReconfigurationResponse dispatch(const concord::messages::ReconfigurationRequest&,
                                                       uint64_t sequence_num);
 
-  void addReconfigurationHandler(std::shared_ptr<IReconfigurationHandler> h) { reconfig_handlers_.push_back(h); }
-  void addPruningHandler(std::shared_ptr<IPruningHandler> h) { pruning_handlers_.push_back(h); }
+  void addReconfigurationHandler(std::shared_ptr<IReconfigurationHandler> h) {
```
toly-kournik, line 41: why check for nullptr? in this case it will be silently ignored.

reconfiguration/include/reconfiguration/pruning_handler.hpp
```diff
@@ -17,22 +17,93 @@
...
+  void persistLastAgreedPrunableBlockId(kvbc::BlockId block_id) const;
+  // Prune blocks in the [genesis, block_id] range (both inclusive).
+  // Throws on errors.
+  void pruneThroughBlockId(kvbc::BlockId block_id) const;
+  void pruneThroughLastAgreedBlockId() const;
+  void pruneOnStateTransferCompletion(uint64_t checkpoint_number) const noexcept;
+  logging::Logger logger_;
+  RSAPruningSigner signer_;
+  RSAPruningVerifier verifier_;
+  const kvbc::IReader &ro_storage_;
```
toly-kournik, line 95: I believe it's better to store shared_ptrs to interfaces rather than references

reconfiguration/include/reconfiguration/dispatcher.hpp
yontyon: if we push nullptr to the handler list and then trying to invoke them, wouldn't this cause a problem?

toly-kournik: It would, but it's a bug, therefore the earlier we catch it the better

reconfiguration/include/reconfiguration/pruning_handler.hpp
```diff
@@ -17,22 +17,93 @@
...
+  void pruneThroughLastAgreedBlockId() const;
+  void pruneOnStateTransferCompletion(uint64_t checkpoint_number) const noexcept;
+  logging::Logger logger_;
+  RSAPruningSigner signer_;
+  RSAPruningVerifier verifier_;
+  std::shared_ptr<kvbc::IReader> ro_storage_;
+  std::shared_ptr<kvbc::IBlockAdder> blocks_adder_;
+  std::shared_ptr<kvbc::IBlocksDeleter> blocks_deleter_;
+
+  bool pruning_enabled_{false};
```
toly-kournik, line 99: If we don't keep the configuration as local class variables, but always ask ReplicaConfig::instance().get() it would make future dynamic configuration impl easier.

PR #1235, opened 2021-02-26
kvbcbench: Fully utilize KeyValueBlockchain

andrewjstone:
Realistic pre-execution workloads where contention check reads and block
writes are performed as fast as possible while pre-execution
reads take place in parallel are now possible.

Random keys and values for all 3 categories: `VersionedKeyValue`, `Immutable`,
and `BlockMerkle` are generated up front such that they can fill a
maximum number of blocks up to `max-memory-for-kv-gen` bytes. If more
blocks are created in a kvbcbench run than can be generated with the
configured memory usage, then the keys and values are reused as blocks
are being written. This is a tradeoff between entropy and memory usage.
If you want 3TB of completely random data in one run, you need 3TB of
RAM. To work around this, you can reuse the existing database and
execute a sequence of shorter runs. This will show impact of a large
database on disk, but the caches will not be the same as one long run,
and those prior keys will be "cold", although they will participate in
compactions and seeks.

The current default size of a transaction is ~20KB of total keys and
values. By default pre-execution reads 100 values in a multiget per transaction,
while contention checks in the hot path read 100 key versions per
transaction. The `batch-size` parameter writes multiple transactions
into a single batch. However, contention check reads are still performed
per request. Therefore in a batch of 10 transactions, 10 multiget
operations will be run for each of the `VersionedKeyValue` category and
`BlockMerkle` category. By default, 50 key versions are read for each
category in each of the simulated contention checks. To minimize the
time taken to choose which keys to read (which would happen outside the
kvbc code in a real system), we pick a random offset in the read key
vectors for each category then grab a consecutive block. We don't just
pick 50 random keys from anywhere in the vector as that would be
additional work not representative of a real storage workload.

Since data now has to be generated for 3 categories, the `InputData`
generation was refactored. To speed up input generation we want to use
all of our cores. Therefore a `parallelInputGen` function that computes a
subset of the random data on each core given a generator, and then
combines the output in a map-reduce fashion, was created. This is used
for all three categories. For the `VersionedKeyValue` and `BlockMerkle`
categories, ReadKeys are also created from a subset of the generated key
value pairs. These are written in parallell to disjoint parts of a
pre-allocated vector during the `parallelInputGen` run. However, in
order to properly do this we actually need a double closure that
properly closes over the pre-allocated `read_keys` vector in the outer
closure and returns a generator that can fill it during the map-reduce
run. This is the reason for the added complexity in
`makeBlockMerkleInputGenerator` and `makeVersionedInputGenerator` as
opposed to `makeImmutableUpdateGenerator` which doesn't have read keys,
as that category does not participate in pre-execution in our current
workloads. In the future, we may make it possible to also have immutable
keys read during pre-execution simulation, although it would make no
sense to do conflict checks on them, since they are immutable.

Due to the constraints on memory when trying to run long benchmarks, we
must be able to copy the generated input. However, to prevent expensive
mistakes in production code, the copy assignment and copy constructors
are deleted for `ImmutableUpdates` and `VersionedUpdates`. We get around
this by adding conditional compilation that allows them only when
building kvbcbench.

Lastly, this commit enables building kvbcbench by default and also
defaults the concord-bft build to Release mode. The rationale is to
always run the fastest code when testing locally. I'm open to reverting
this, but I don't see a good reason not to change it at this point.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ CONCORD_BFT_CONTAINER_SHELL:=/bin/bash
 CONCORD_BFT_CONTAINER_CC:=clang
 CONCORD_BFT_CONTAINER_CXX:=clang++
 CONCORD_BFT_CMAKE_FLAGS:= \
-			-DCMAKE_BUILD_TYPE=Debug \
+			-DCMAKE_BUILD_TYPE=Release \
 			-DBUILD_TESTING=ON \
 			-DBUILD_COMM_TCP_PLAIN=FALSE \
 			-DBUILD_COMM_TCP_TLS=TRUE\
@@ -24,7 +24,7 @@ CONCORD_BFT_CMAKE_FLAGS:= \
 			-DOMIT_TEST_OUTPUT=OFF \
 			-DKEEP_APOLLO_LOGS=TRUE \
 			-DBUILD_SLOWDOWN=FALSE \
-			-DBUILD_KVBC_BENCH=FALSE \ # Should only run this with -DCMAKE_BUILD_TYPE=Release
+			-DBUILD_KVBC_BENCH=TRUE \ # Should only run this with -DCMAKE_BUILD_TYPE=Release
 
 # The consistency parameter makes sense only at MacOS.
 # It is ignored at all other platforms.
--- a/kvbc/benchmark/kvbcbench/input.h
+++ b/kvbc/benchmark/kvbcbench/input.h
@@ -21,6 +21,7 @@
 #include <boost/program_options/variables_map.hpp>
 
 #include "categorized_kvbc_msgs.cmf.hpp"
+#include "categorization/updates.h"
 
 const std::string kCategoryMerkle = "block_merkle";
 const std::string kCategoryVersioned = "ver_test";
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -11,7 +11,7 @@ CONCORD_BFT_CONTAINER_SHELL:=/bin/bash
 CONCORD_BFT_CONTAINER_CC:=clang
 CONCORD_BFT_CONTAINER_CXX:=clang++
 CONCORD_BFT_CMAKE_FLAGS:= \
-			-DCMAKE_BUILD_TYPE=Debug \
+			-DCMAKE_BUILD_TYPE=Release \
```
dartdart26, line 14: I agree, I don't see a reason why the default should be Debug.

dartdart26 (approved):

By the way, I ran with **THREADCHECK** set to ON and I got some races, but I am not exactly sure if there is a problem. I also rebased it to latest master that has the atomic block counters, but still get it. Will investigate further.

```
  Previous write of size 1 at 0x7f478f96e412 by main thread:
    #0 memcpy <null> (libtsan.so.0+0x32505)
    #1 memcpy /usr/include/x86_64-linux-gnu/bits/string_fortified.h:34 (librocksdb.so.6.8+0x2cc91a)
    #2 rocksdb::MemTable::Add(unsigned long, rocksdb::ValueType, rocksdb::Slice const&, rocksdb::Slice const&, bool, rocksdb::MemTablePostProcessInfo*, void**) db/memtable.cc:492 (librocksdb.so.6.8+0x2cc91a)
    #3 concord::kvbc::categorization::KeyValueBlockchain::addBlock(concord::kvbc::categorization::Updates&&) <null> (kvbcbench+0x19680c)
    #4 concord::kvbc::bench::addBlocks(boost::program_options::variables_map const&, std::shared_ptr<concord::storage::rocksdb::NativeClient>&, concord::kvbc::categorization::KeyValueBlockchain&, concord::kvbc::bench::InputData&, std::shared_ptr<concord::diagnostics::Recorder>&, std::shared_ptr<concord::diagnostics::Recorder>&) <null> (kvbcbench+0x1592aa)
    #5 main <null> (kvbcbench+0x1524e0)
```

```Read of size 1 at 0x7f478f96e412 by thread T15:
  Read of size 1 at 0x7f478f96e412 by thread T15:
    #0 memcmp <null> (libtsan.so.0+0x4f182)
    #1 rocksdb::Slice::compare(rocksdb::Slice const&) const include/rocksdb/slice.h:250 (librocksdb.so.6.8+0x4729af)
    #2 Compare util/comparator.cc:28 (librocksdb.so.6.8+0x4729af)
    #3 concord::kvbc::categorization::detail::BlockMerkleCategory::multiGet(std::vector<std::vector<unsigned char,
```

@cloudnoize just FYI.

PR #1238, opened 2021-02-26
Add more logs to TlsTcpImpl

nkgeorgiev:
These logs will be useful when investigating connection issues.

Currently we are investigating an issue where the replicas seem to stop accepting connections. These logs will be useful in seeing if the replicas really stop listening and if there are any connection leaks.


HristoStaykov (approved):
LGTM, Thanks @nkgeorgiev!

PR #1243, opened 2021-03-01
Add a new script to generate transaction signing keypairs

glevkovich:
create_concord_clients_transaction_signing_keys.sh generates rsa-2048
bit private/public keypairs (non-encrypted) in selected folder with the
next tree structure, inside the destination folder transaction_signing_keys:

├── 1
│   ├── private_key.pem
│   └── public_key.pem
├── 2
│   ├── private_key.pem
│   └── public_key.pem
└── 3
    ├── private_key.pem
    └── public_key.pem

For now, we need these keypairs for Concord Client transaction signing: the private key is used by the participant node (client) for signing transactions. All Public keys are loaded by each replica to validate signed transactions.

For more informations on all available options, see script header, or type:
> create_concord_clients_transaction_signing_keys.sh --help

### DIFF
--- a/scripts/linux/create_concord_clients_transaction_signing_keys.sh
+++ b/scripts/linux/create_concord_clients_transaction_signing_keys.sh
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+
+# This script generates sets of key pairs - private/public keys to be used for
+# concord client (participant node) transaction signing.
+#
+# Output is generated into output path output_base_path/transaction_signing_keys folder.
+# If --output_base_path is not given by user, output is generated into output
+# path script_working_path/transaction_signing_keys.
+# Each generated set resides in a separate numeric sub-folder, range from 1 to
+# num_participants.
+#
+# In general, it can be used by any 2 entities; the signer should use the
+# private_key.pem to sign, and the verifier should use public_key.pem
+# to verify the signature.
+
+# This script assumes that the output path is empty. If not, it fails with a warning.
+set -eo pipefail
+
+usage() {
+    printf "\n-----\nhelp:\n-----\n"
+    printf "%s\n\n" "-h --help, print this message"
+    printf "%s\n\n" "-n --num_participants <integer>, mandatory"
+    printf "%s\n\n" "-r --private_key_name <string>, optional, default: private_key.pem"
+    printf "%s\n\n" "-u --public_key_name <string>, optional, default: public_key.pem"
+    printf "%s\n\t\t\t\t%s\n\t\t\t\t%s\n" "-o --output_base_path <string>, optional, base path in relative/absolute format" \
+           "output is redirected to output_base_path/transaction_signing_keys folder" \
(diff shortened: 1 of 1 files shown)


f-squirrel (approved):
LGTM!

scripts/linux/create_concord_clients_transaction_signing_keys.sh
```diff
@@ -0,0 +1,118 @@
...
+# If --output_base_path is not given by user, output is generated into output
+# path script_working_path/transaction_signing_keys.
+# Each generated set resides in a separate numeric sub-folder, range from 1 to
+# num_participants.
+#
+# In general, it can be used by any 2 entities; the signer should use the
+# private_key.pem to sign, and the verifier should use public_key.pem
+# to verify the signature.
+
+# This script assumes that the output path is empty. If not, content is deleted.
```
andrewjstone, line 16: That seems dangerous

scripts/linux/create_concord_clients_transaction_signing_keys.sh
```diff
@@ -0,0 +1,118 @@
...
+parser $@
+
+
+rm -rf ${output_path}
+mkdir -p ${output_path}
+
+for ((i=1; i<=${num_participants}; i++)); do
+    current_path=${output_path}/${i}
+    mkdir ${current_path}
+    openssl genrsa -out ${current_path}/${private_key_name} 2048 > /dev/null 2>&1
```
andrewjstone, line 110: Why don't we want to see reported errors from openssl? It may alert us to a problem.

andrewjstone (changes requested):
This looks great @glevkovich. It's very clean and readable. However, I think you should seriously reconsider deleting from arbitrary paths. Even deleting when a path is passed in is pretty dangerous. Most scripts don't do things like this as it makes accidents very dangerous. I would let clearing the directory be the user's responsibility.

 A simple mechanism is to just print a warning message stating that keys are being created in a populated directory. Another variant is to just refuse to run in a non-empty directory. This latter suggestion should only be taken if an empty directory is strictly required though.

scripts/linux/create_concord_clients_transaction_signing_keys.sh
glevkovich: OK, I'm going to add an option to overwrite an existing folder.
Else, if the destination folder is not-empty, the script will fail.
Thanks!

glevkovich: @andrewjstone  thanks for the comment, I'll just fail the script if the base destination folder is not empty, and print a warning.

andrewjstone: Cool! Approved :)

andrewjstone (approved):
Approved based on @glevkovich's  latest changes to fail if the folder is not empty rather than deleting.

PR #1246, opened 2021-03-02
Add Pruning Status for concord-ctl

saiakhil2012:
Add functionality of getting Pruning Status from `concord-ctl`

### DIFF
--- a/kvbc/src/categorization/kv_blockchain.cpp
+++ b/kvbc/src/categorization/kv_blockchain.cpp
@@ -18,8 +18,6 @@
 
 #include <stdexcept>
 
-#define getName(var) #var
-
 namespace concord::kvbc::categorization {
 
 using ::bftEngine::bcst::computeBlockDigest;
@@ -607,11 +605,14 @@ std::string KeyValueBlockchain::getPruningStatus() {
   std::ostringstream oss;
   std::unordered_map<std::string, std::string> result;
 
-  result.insert(toPair(getName(versioned_num_of_deletes_keys_), versioned_num_of_deletes_keys_.Get().Get()));
-  result.insert(toPair(getName(immutable_num_of_deleted_keys_), immutable_num_of_deleted_keys_.Get().Get()));
-  result.insert(toPair(getName(merkle_num_of_deleted_keys_), merkle_num_of_deleted_keys_.Get().Get()));
-  result.insert(toPair(getName(getGenesisBlockId()), getGenesisBlockId()));
-  result.insert(toPair(getName(getLastReachableBlockId()), getLastReachableBlockId()));
+  result.insert(toPair("versioned_num_of_deletes_keys_",
+                       aggregator_->GetCounter("kv_blockchain_deletes", "numOfVersionedKeysDeleted").Get()));
+  result.insert(toPair("immutable_num_of_deleted_keys_",
+                       aggregator_->GetCounter("kv_blockchain_deletes", "numOfImmutableKeysDeleted").Get()));
+  result.insert(toPair("merkle_num_of_deleted_keys_",
+                       aggregator_->GetCounter("kv_blockchain_deletes", "numOfMerkleKeysDeleted").Get()));
+  result.insert(toPair("getGenesisBlockId()", getGenesisBlockId()));
+  result.insert(toPair("getLastReachableBlockId()", getLastReachableBlockId()));
   result.insert(toPair("isPruningInProgress", bftEngine::ControlStateManager::instance().getPruningProcessStatus()));
 
   oss << concordUtils::kContainerToJson(result);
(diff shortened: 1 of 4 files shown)


kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -13,12 +13,17 @@
 
 #include "categorization/kv_blockchain.h"
 #include "bcstatetransfer/SimpleBCStateTransfer.hpp"
+#include "bftengine/ControlStateManager.hpp"
+#include "json_output.hpp"
 
 #include <stdexcept>
 
+#define getName(var) #var
```
andrewjstone, line 21: I prefer not to litter every cpp file with this macro. Also, `getName(x)` is longer than just `"x"`. Why even use this macro at all?

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -598,4 +603,19 @@ void KeyValueBlockchain::writeSTLinkTransaction(const BlockId block_id, RawBlock
   block_chain_.setAddedBlockId(new_block_id);
 }
 
+std::string KeyValueBlockchain::getPruningStatus() {
+  std::ostringstream oss;
+  std::unordered_map<std::string, std::string> result;
+
+  result.insert(toPair(getName(versioned_num_of_deletes_keys_), versioned_num_of_deletes_keys_.Get().Get()));
+  result.insert(toPair(getName(immutable_num_of_deleted_keys_), immutable_num_of_deleted_keys_.Get().Get()));
+  result.insert(toPair(getName(merkle_num_of_deleted_keys_), merkle_num_of_deleted_keys_.Get().Get()));
```
andrewjstone, line 612: This is not thread safe. These counters are not atomic, but `getPruningStatus` is called from the diagnostic handler thread. You can either make the variables AtomicCounters  (which will slow down updates which are much more frequent than status access) or use the [aggregator api](https://github.com/vmware/concord-bft/blob/master/util/src/Metrics.cpp#L130) to access the metrics in a thread safe manner. This has no cost to updates, but it will hold a lock that may cause contention if another thread is trying to update its metrics. Either way is probably ok for this specific case.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -598,4 +603,19 @@ void KeyValueBlockchain::writeSTLinkTransaction(const BlockId block_id, RawBlock
...
+  result.insert(toPair(getName(getGenesisBlockId()), getGenesisBlockId()));
+  result.insert(toPair(getName(getLastReachableBlockId()), getLastReachableBlockId()));
```
andrewjstone, line 614: These two are already atomic, so you are fine here.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -598,4 +603,19 @@ void KeyValueBlockchain::writeSTLinkTransaction(const BlockId block_id, RawBlock
...
+  result.insert(toPair("isPruningInProgress", bftEngine::ControlStateManager::instance().getPruningProcessStatus()));
```
andrewjstone, line 615: You should make this an `atomic_bool`: https://github.com/vmware/concord-bft/blob/e3661fcd8be966969625e57bb4089fcfe9402642/bftengine/include/bftengine/ControlStateManager.hpp#L79

andrewjstone (changes requested):
@saiakhil2012 Unfortunately, the status handler interface is not automatically thread safe. Any data accessed in the callable must be made thread safe. I pointed out the specific problems and solutions in the comments. (If we were using rust this wouldn't have even compiled :))

kvbc/src/categorization/kv_blockchain.cpp
saiakhil2012: Yes, I agree, I added as the same was being used in `concord-ctl` related functionality in multiple places already. So, my understanding was that, it was being used for code readability and maybe to handle future changes in the names of variables. But sure, I changed the usage to direct name of the variable, its more precise and no need of the macro.

kvbc/src/categorization/kv_blockchain.cpp
saiakhil2012: Oh, missed that, thanks. I updated to using the Aggregator API to access the counters in a thread safe manner

kvbc/src/categorization/kv_blockchain.cpp
saiakhil2012: Yes, made it `atomic_bool` for making it thread safe

saiakhil2012:
Thanks @andrewjstone for the review. I have incorporated the suggested changes

andrewjstone (approved):
Looks great! Thanks @saiakhil2012.

PR #1253, opened 2021-03-04
Remove wrong deleted flag assert in Versioned cat

dartdart26:
Remove a wrong assert in VersionedKeyValueCategory that was checking if
a key's deleted flag in the output is equal to the latest key version's
deleted flag. There might be valid cases that violate it.

Delete and then update:
 * delete key `K` in block N
 * update key `K` with a value in block N + 1
 * prune block N

Update and then delete:
 * update key `K` in block N
 * delete key `K` in block N + 1
 * prune block N

Add unit tests to cover above valid cases.


yontyon (approved):
LGTM, thanks @dartdart26

PR #1254, opened 2021-03-04
Add view number to relevant logs

nkgeorgiev:
The view number will ease automated log analysis.

The view number is written in the field for seqNum so we need to restore the old value after logging where applicable.
That is why I have added a new get function for MDC.


HristoStaykov (approved):
LGTM, thanks @nkgeorgiev !

PR #1256, opened 2021-03-05
Docker update to version 0.16

tdimitrov:
Remove nlohman_json and httplib, because they are not used.


teoparvanov (approved):
LGTM, thanks!

f-squirrel (approved):

PR #1258, opened 2021-03-06
kvbcbench: Periodically dump rocksdb stats

andrewjstone:
Every configurable number of blocks write rocksdb stats to output. This
helps us to debug memory usage and tune the system appropriately,
especially during long running tests. It enables us to better see where
degradation is occurring.

This change requires exposing the underlying rocksdb client itself
through the NativeClient wrapper with the `rawDB` method. The
justification for exposing the underlying rocksdb client is so that we
don't have to wrap every single thing when we want to access stats or
configuration. The interfaces for these things are fine without
wrapping, and are only expected to be used for diagnostics. Future
commits should make these available through the diagnostics subsystem in
concord-bft and concord-ctl in addition to dumping all stats in
kvbcbench.

The default RocksDB block cache size for kvbcbench was increased to 16GB
and made configurable.

OptimizeForPointLookup was removed from the
BLOCK_MERKLE_LATEST_KEY_VERSION_CF column family, as it was suspected of
uncontrollable memory growth, and didn't appear to help performance. A
future change to both the BlockMerkle and VersionedKeyValue categories
is planned for an application specific LRU cache for versions to avoid
the need to hit RocksDB for this data in the common case.

### DIFF
--- a/kvbc/benchmark/kvbcbench/main.cpp
+++ b/kvbc/benchmark/kvbcbench/main.cpp
@@ -55,6 +55,8 @@ using categorization::TaggedVersion;
 // threaad and atomically swapping when needed by block addition. For now we keep it simple.
 static constexpr size_t MAX_MEMORY_SIZE_FOR_KV_DEFAULT = 1024 * 1024 * 1024 * 4ul;  // 4GB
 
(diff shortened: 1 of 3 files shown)


dartdart26 (approved):
That’s useful, thanks!

PR #1261, opened 2021-03-08
Add support for encrypted private keys on TLS comm

KTeofanidis:
The tester replica will now decrypt it's private keys used for TLS communication.

### DIFF
--- a/communication/include/communication/CommDefs.hpp
+++ b/communication/include/communication/CommDefs.hpp
@@ -24,6 +24,7 @@
 
 #include "communication/ICommunication.hpp"
 #include "communication/StatusInfo.h"
+#include "secret_data.h"
 
 namespace bft::communication {
 typedef struct sockaddr_in Addr;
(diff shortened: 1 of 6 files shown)


communication/include/communication/CommDefs.hpp
```diff
@@ -35,6 +36,7 @@ struct NodeInfo {
 };
 
 typedef std::unordered_map<NodeNum, NodeInfo> NodeMap;
+using namespace concord::secretsmanager;
```
tdimitrov, line 39: Don't put 'using namespace' in header files. It will pollute the namespace of every cpp using the header.

tdimitrov (approved):

communication/include/communication/CommDefs.hpp
```diff
@@ -108,6 +110,9 @@ struct TlsTcpConfig : PlainTcpConfig {
   // https://www.openssl.org/docs/man1.0.2/man1/ciphers.html
   std::string cipherSuite;
 
+  bool usePKDecryption;
```
tdimitrov, line 113: Maybe we can use std::optional for SecretData. I don't insist though - it's up to you.

PR #1265, opened 2021-03-10
Change log error to warning for shutdown message

shgandhi:


andrewjstone (approved):
Thanks for the fix @shgandhi

PR #1267, opened 2021-03-10
Fix reconfiguration bug (master)

yontyon:
In this PR we fix a reconfiguration bug:
Once we decide to wedge, we fill the working window with noop commands in order to get to the next-next checkpoint.
However, considering the following case:
1. The wedge command received close to checkpoint (say on 300)
2. The current latestStableCheckpoint is still 150
3.  Thus, the working window is from 0 to 300
4. But we need to send noops until 600(!)
5. The master failed to send noop and eventually, VC is happening

The fix is that once we execute the wedge command we send noops up to the near checkpoint, and once we get to a stable checkpoint, we send the rest of the noops.
Notice that now, it is possible to process some client request before we completely wedge

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -195,7 +195,7 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
     return;
   }
 
-  if (isSeqNumToStopAt(lastExecutedSeqNum)) {
+  if ((isCurrentPrimary() && isSeqNumToStopAt(primaryLastUsedSeqNum + 1)) || isSeqNumToStopAt(lastExecutedSeqNum + 1)) {
     LOG_INFO(GL,
              "Ignoring ClientRequest because system is stopped at checkpoint pending control state operation (upgrade, "
              "etc...)");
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -545,22 +545,19 @@ void ReplicaImp::sendInternalNoopPrePrepareMsg(CommitPath firstPath) {
...
+  while (primaryLastUsedSeqNum < seqNumToStopAt) {
     sendInternalNoopPrePrepareMsg(firstPath);
   }
 }
 
 bool ReplicaImp::isSeqNumToStopAt(SeqNum seq_num) {
   if (ControlStateManager::instance().getPruningProcessStatus()) return true;
-  if (seqNumToStopAt_ > 0 && seq_num == seqNumToStopAt_) return true;
-  if (seqNumToStopAt_ > 0 && seq_num > seqNumToStopAt_) return false;
   auto seq_num_to_stop_at = ControlStateManager::instance().getCheckpointToStopAt();
```
HristoStaykov, line 557: For consistency, could we use camel case for the variables' names.

HristoStaykov (approved):
Great finding and fix, @yontyon !

PR #1269, opened 2021-03-10
Fix an issue when using nested scopedMDC

nkgeorgiev:
ScopedMDC is a RAII class that removes the value associated with its key from the MDC map when it goes out of scope.
But if we have nested ScopedMDC objects on the same key, the inner one overwrites the outer. 
When the inner object is destroyed, it deletes the value from the MDC map instead of restoring the one from the outer scope.
This leads to a loss of information inside the logs.

The fix is to store the previous value inside the ScopedMDC object and restore it on destruction.

### DIFF
--- a/logging/include/Logger.hpp
+++ b/logging/include/Logger.hpp
@@ -48,6 +48,7 @@ class ScopedMdc {
 
  private:
   const std::string key_;
+  const std::string last_value_;
 };
 
 }  // namespace logging
--- a/logging/src/Logger.cpp
+++ b/logging/src/Logger.cpp
@@ -16,8 +16,15 @@
 
 namespace logging {
 
-ScopedMdc::ScopedMdc(const std::string& key, const std::string& val) : key_{key} { MDC_PUT(key, val); }
-ScopedMdc::~ScopedMdc() { MDC_REMOVE(key_); }
+ScopedMdc::ScopedMdc(const std::string& key, const std::string& val) : key_{key}, last_value_(MDC_GET(key)) {
+  MDC_PUT(key, val);
+}
+ScopedMdc::~ScopedMdc() {
+  MDC_REMOVE(key_);
+  if (!last_value_.empty()) {
+    MDC_PUT(key_, last_value_);
+  }
+}
 
 }  // namespace logging
 


toly-kournik:
What's the example of a nested MDC?
What if there are more than one nested MDC?

f-squirrel:
> What's the example of a nested MDC?
> What if there are more than one nested MDC?

From my understanding, every scoped MDC in a chain restores its previous value.
Is it correct @nkgeorgiev ?
However, I wonder why we have multiple MDC with the same value:)

nkgeorgiev:
> What's the example of a nested MDC?

One of the patterns is to create a scopedMDC at the start of onMessage methods.

Here for example:
https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L2148
onMessage<ViewChangeMsg> sets the seqNum value to the view number (which is an issue in itself but it's harder to change the format of the logs now)
After that it calls onMessage<ReplicaAsksToLeaveViewMsg> which again sets the seqNum value. I changed that call to non-scoped MDC_PUT in a previous change as a temp fix. I have forgotten to revert it in this PR. I will update it in a minute :)

> What if there are more than one nested MDC?
It will become like a stack of values with each nested MDC pushing a value and popping when it is destroyed. The logger always gets the top value. Since MDC is thread local there should be no issues.

nkgeorgiev:
> > What's the example of a nested MDC?
> > What if there are more than one nested MDC?
> 
> From my understanding, every scoped MDC in a chain restores its previous value.
> Is it correct @nkgeorgiev ?
> However, I wonder why we have multiple MDC with the same key:)

I gave an example of mutiple MDC with the same key in my previous comments.
There are also many redundant MDC that were trying to workaround this problem.
I will start removing them in a different PR.

toly-kournik:
> > What's the example of a nested MDC?
> 
> One of the patterns is to create a scopedMDC at the start of onMessage methods.
> 
> Here for example:
> https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L2148
> onMessage sets the seqNum value to the view number (which is an issue in itself but it's harder to change the format of the logs now)
> After that it calls onMessage which again sets the seqNum value. I changed that call to non-scoped MDC_PUT in a previous change as a temp fix. I have forgotten to revert it in this PR. I will update it in a minute :)
> 
> > What if there are more than one nested MDC?
> 
> It will become like a stack of values with each nested MDC pushing a value and popping when it is destroyed. The logger always gets the top value. Since MDC is thread local there should be no issues.



> > What's the example of a nested MDC?
> 
> One of the patterns is to create a scopedMDC at the start of onMessage methods.
> 
> Here for example:
> https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L2148
> onMessage sets the seqNum value to the view number (which is an issue in itself but it's harder to change the format of the logs now)
> After that it calls onMessage which again sets the seqNum value. I changed that call to non-scoped MDC_PUT in a previous change as a temp fix. I have forgotten to revert it in this PR. I will update it in a minute :)
> 
> > What if there are more than one nested MDC?
> 
> It will become like a stack of values with each nested MDC pushing a value and popping when it is destroyed. The logger always gets the top value. Since MDC is thread local there should be no issues.

Using view number as a sequence number is a misuse!

nkgeorgiev:
> > > What's the example of a nested MDC?
> > 
> > 
> > One of the patterns is to create a scopedMDC at the start of onMessage methods.
> > Here for example:
> > https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L2148
> > onMessage sets the seqNum value to the view number (which is an issue in itself but it's harder to change the format of the logs now)
> > After that it calls onMessage which again sets the seqNum value. I changed that call to non-scoped MDC_PUT in a previous change as a temp fix. I have forgotten to revert it in this PR. I will update it in a minute :)
> > > What if there are more than one nested MDC?
> > 
> > 
> > It will become like a stack of values with each nested MDC pushing a value and popping when it is destroyed. The logger always gets the top value. Since MDC is thread local there should be no issues.
> 
> > > What's the example of a nested MDC?
> > 
> > 
> > One of the patterns is to create a scopedMDC at the start of onMessage methods.
> > Here for example:
> > https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L2148
> > onMessage sets the seqNum value to the view number (which is an issue in itself but it's harder to change the format of the logs now)
> > After that it calls onMessage which again sets the seqNum value. I changed that call to non-scoped MDC_PUT in a previous change as a temp fix. I have forgotten to revert it in this PR. I will update it in a minute :)
> > > What if there are more than one nested MDC?
> > 
> > 
> > It will become like a stack of values with each nested MDC pushing a value and popping when it is destroyed. The logger always gets the top value. Since MDC is thread local there should be no issues.
> 
> Using view number as a sequence number is a misuse!

I agree, but it is a different issue for another PR

PR #1273, opened 2021-03-11
BCStateTran: SourceSelector - do not choose RO replicas as source (fix)

glevkovich:
On Constructor, BCStateTran was passed Config::numReplicas as the sum of
consensus replicas and RO replicas.
This parameter was used to construct an std::set which was passed to
sourceSelector_ ctor. This included all replicas except "myself",
and in the case where we have RO replicas, it would include RO replicas
as well.

This can cause a slowdown in ST, as a RO replica might be chosen
randomly in some cases (after setting preferredReplicas_ ), usually long
state transfers. When RO replica is chosen, it will never answer, since
probably it is in ST as well (and for other reasons). Only after some
time, the dst replica will move to another source.


toly-kournik (approved):
LGTM

PR #1278, opened 2021-03-14
Fix issues found in LR

yuliasherman:
- Fix bitmap index for skipped requests in ReplicaImp while collecting requests to be accumulated in one block.
- Fix reply add/remove issues in ClientsManager.


toly-kournik:
> Don't we like to fail in both cases?

If yes, we need to remove the `if (repMsg)` at both calls

cloudnoize (approved):
LGTM thanks @yuliasherman!

PR #1279, opened 2021-03-14
[Diagnostic][master] Remove async execution of handleRequest

KTeofanidis:
Due to the way it was written the handleRequest was synchronous anyway, so for now remove the unnecessary overhead.

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -147,9 +147,7 @@ class Server {
           return;
         }
         int sock = accept(listen_sock_, NULL, NULL);
-        // We must bind the result future or else this call blocks.
-        auto _ = std::async(std::launch::async, [&]() { handleRequest(registrar, sock); });
-        (void)_;  // unused variable hack
+        handleRequest(registrar, sock);
       }
     });
   }


dartdart26:
@KTeofanidis It makes sense to do that as it should be equivalent to the existing code. However, I think the original intent (also based on the code comment) was to support parallel requests.

I was thinking we might create a `std::thread` and call `detach` to achieve the original intent. There are some drawbacks, but I think the `run` method doesn't change any state outside this process, meaning that even if the replica is stopped while the detached thread is running, it should not corrupt anything.

I am not sure how important it is to execute in parallel, thought. @andrewjstone would know better.

andrewjstone:
> @KTeofanidis It makes sense to do that as it should be equivalent to the existing code. However, I think the original intent (also based on the code comment) was to support parallel requests.
> 
> I was thinking we might create a `std::thread` and call `detach` to achieve the original intent. There are some drawbacks, but I think the `run` method doesn't change any state outside this process, meaning that even if the replica is stopped while the detached thread is running, it should not corrupt anything. Also, we don't return anything in the promise and we don't throw in the handler.
> 
> I am not sure how important it is to execute in parallel, though. @andrewjstone would know better.

Indeed, the original intent was to handle requests concurrently. The entire design was meant to support that. In practice though, clearly it was only operating synchronously. I think it's fine to merge this. But we should also track the issue to make it concurrent for real. Petar's suggestion seems fine, although I generally dislike detached threads.

andrewjstone (approved):
LGTM. Fine to fix with this for now as in my other comment.

dartdart26:
> LGTM. Fine to fix with this for now as in my other comment.

Cool

PR #1280, opened 2021-03-14
[Diagnostic][v0.10.x] Remove async execution of handleRequest

KTeofanidis:
Due to the way it was written the handleRequest was synchronous anyway, so for now remove the unnecessary overhead.


andrewjstone (approved):
LGTM. Thanks for the fix @KTeofanidis 

 I talked to @dartdart26 about this on Friday afternoon, and we should use our built in thread-pool/async behavior to make it concurrent again, although I'm not sure it's critical.

PR #1281, opened 2021-03-15
Status exchange mechanism refactors

HristoStaykov:
1. Remove assert when peer Replica reports inconsistent stable
   sequence ID, only print an error log.
2. Refactor View Change msgs exchange mechanism in status exchange,
   now all Replicas will forward the complete set they have collected.
3. Send View Change messages even to peers that are more than a Working Window
   behind. In this case the catching Replica should be able to gather them even
   before State Transfer starts and be able to activate a View faster once ST completes.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1890,7 +1890,10 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
   const ReplicaId msgSenderId = msg->senderId();
   const SeqNum msgLastStable = msg->getLastStableSeqNum();
   const ViewNum msgViewNum = msg->getViewNumber();
-  ConcordAssertEQ(msgLastStable % checkpointWindowSize, 0);
+  LOG_ERROR(MSGS,
+            "FATAL ERROR for peer msgSenderId = "
+                << msgSenderId << ". Reported Last Stable Sequence not consistent with checkpointWindowSize "
+                << KVLOG(msgLastStable, checkpointWindowSize));
 
   LOG_DEBUG(MSGS, KVLOG(msgSenderId, msgLastStable, msgViewNum, lastStableSeqNum));
 
@@ -1908,7 +1911,6 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
     }
 
     delete msg;
-    return;
   } else if (msgLastStable > lastStableSeqNum + kWorkWindowSize) {
     tryToSendStatusReport();  // ask for help
   } else {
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1890,7 +1890,10 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
   const ReplicaId msgSenderId = msg->senderId();
   const SeqNum msgLastStable = msg->getLastStableSeqNum();
   const ViewNum msgViewNum = msg->getViewNumber();
-  ConcordAssertEQ(msgLastStable % checkpointWindowSize, 0);
+  LOG_ERROR(MSGS,
+            "FATAL ERROR for peer msgSenderId = "
+                << msgSenderId << ". Reported Last Stable Sequence not consistent with checkpointWindowSize "
```
teoparvanov, line 1895: How can we change the message, so that we emphasize the issue is with the peer replica, and not the receiving one?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1953,43 +1955,33 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
...
   else if ((msgViewNum == curView) && (!msg->currentViewIsActive())) {
-    auto sendViewChangeMsg = [&msg, &msgSenderId, this]() {
-      if (curView > 0 &&  // we only have ViewChangeMsg for View > 0
-          msg->hasListOfMissingViewChangeMsgForViewChange() &&
-          msg->isMissingViewChangeMsgForViewChange(config_.getreplicaId())) {
-        ViewChangeMsg *myVC = viewsManager->getMyLatestViewChangeMsg();
-        ConcordAssertNE(myVC, nullptr);
-        sendAndIncrementMetric(myVC, msgSenderId, metric_sent_viewchange_msg_due_to_status_);
+    auto sendViewChangeMsgs = [&msg, &msgSenderId, this]() {  // Send all View Change messages we have
+      if (curView > 0 &&                                      // we only have ViewChangeMsg for View > 0
```
teoparvanov, line 1959: let's put the comment as a single line, in the beginning of the lambda body

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1953,43 +1955,33 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
...
+        }
       }
     };
 
     if (isCurrentPrimary() || (repsInfo->primaryOfView(curView) == msgSenderId))  // if the primary is involved
     {
-      if (!isCurrentPrimary())  // I am not the primary of curView
-      {
-        sendViewChangeMsg();
-      } else  // I am the primary of curView
```
teoparvanov, line 1971: Is this case captured by the final `else` case?

teoparvanov (approved):
LGTM, thanks @HristoStaykov ! I have some minor comments/questions above.

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: The final else is for the case when the Primary is not involved. For the case where the Primary is involved we need to first send the New View message, if we are the current Primary and we have such, and then the View Change messages we have. Otherwise only send all View Change messages we have to the Primary.

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: Will change to "ERROR detected in peer msgSenderId ="

PR #1282, opened 2021-03-15
Release status exchange fixes

HristoStaykov:
1. Remove assert when peer Replica reports inconsistent stable
   sequence ID, only print an error log.
2. Refactor View Change msgs exchange mechanism in status exchange,
   now all Replicas will forward the complete set they have collected.
3. Send View Change messages even to peers that are more than a Working Window
   behind. In this case the catching Replica should be able to gather them even
   before State Transfer starts and be able to activate a View faster once ST completes.
4.  Additional logged info for status request.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1897,10 +1897,13 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
   const ReplicaId msgSenderId = msg->senderId();
   const SeqNum msgLastStable = msg->getLastStableSeqNum();
   const ViewNum msgViewNum = msg->getViewNumber();
-  LOG_ERROR(MSGS,
-            "FATAL ERROR for peer msgSenderId = "
-                << msgSenderId << ". Reported Last Stable Sequence not consistent with checkpointWindowSize "
-                << KVLOG(msgLastStable, checkpointWindowSize));
+  if (msgLastStable % checkpointWindowSize != 0) {
+    LOG_ERROR(MSGS,
+              "ERROR detected in peer msgSenderId = "
+                  << msgSenderId << ". Reported Last Stable Sequence not consistent with checkpointWindowSize "
+                  << KVLOG(msgLastStable, checkpointWindowSize));
+    return;
+  }
 
   LOG_DEBUG(MSGS, KVLOG(msgSenderId, msgLastStable, msgViewNum, lastStableSeqNum));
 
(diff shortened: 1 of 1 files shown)


teoparvanov (approved):
LGTM!

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1897,9 +1897,15 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
   const ReplicaId msgSenderId = msg->senderId();
   const SeqNum msgLastStable = msg->getLastStableSeqNum();
   const ViewNum msgViewNum = msg->getViewNumber();
-  ConcordAssertEQ(msgLastStable % checkpointWindowSize, 0);
```
andrewjstone, line 1900: What's the justification behind this change? This seems like a pretty serious error to run into. The assert was added presumably because a violation of this invariant is a protocol bug.

The rationale for changes like this should be put in the commit message, as just listing the change is not sufficient for a reviewer.

teoparvanov: Hey Andrew, the idea is - this error happens in another replica (the sender). No need to crash the receiver because of this :)

HristoStaykov: Yes, it is a serious problem, but it is most likely in the peer Replica. This leaves an open door for a Byzantine peer to bring all the Replicas down by reporting an inconsistent Stable Sequence ID. If we detect that a peer has probably gotten corrupted I think we must log an error and not shut ourself down?

teoparvanov: On a side note, this actually doesn't happen. We just noticed the vulnerability while reviewing the code in this area.

andrewjstone: Aha! This is an excellent point! Thank you for the clarification.

andrewjstone (approved):
LGTM. Thanks for the fixes @HristoStaykov!

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: Actually after researching deeper, I see that we have this check in the validate method of this message [ReplicaStatusMsg.cpp:103](https://github.com/vmware/concord-bft/blob/b81b12406a2f838cc8797bda7aa4413ccb7218e1/bftengine/src/bftengine/messages/ReplicaStatusMsg.cpp#L103). Basically this means that malformed StatusRequest messages will be discarded on an earlier stage -> [ReplicaBase.hpp:90](https://github.com/vmware/concord-bft/blob/b81b12406a2f838cc8797bda7aa4413ccb7218e1/bftengine/src/bftengine/ReplicaBase.hpp#L90).

PR #1284, opened 2021-03-16
Fix race condition reported by tsan.

HristoStaykov:
Fix reported race by Thread Sanitizer:
`WARNING: ThreadSanitizer: data race (pid=1)
  Write of size 8 at 0x7b5c00001438 by thread T137:
/concord/submodules/concord-bft/bftengine/src/bftengine/SimpleClientImp.cpp:341
...
  Previous read of size 8 at 0x7b5c00001438 by thread T129 (mutexes: write M345505431491187552):
/concord/submodules/concord-bft/bftengine/src/bftengine/SimpleClientImp.cpp:551`


yuliasherman:
Good catch! But we need to go through all the SimpleClient code and verify that pendingRequests_ is safely used everywhere.

andrewjstone (approved):
LGTM. Nice find and fix. Gotta love tsan :)

PR #1285, opened 2021-03-16
Don't remove the blockchain state in case lastExecutedSeqNum is zero

yuliasherman:
- Avoid a whole state removal after cleaning of metadata
- Don't allow removal of more than one block

### DIFF
--- a/bftengine/src/bftengine/BFTEngine.cpp
+++ b/bftengine/src/bftengine/BFTEngine.cpp
@@ -116,9 +116,7 @@ IReplica::IReplicaPtr IReplica::createNewReplica(const ReplicaConfig &replicaCon
                                                  IStateTransfer *stateTransfer,
                                                  bft::communication::ICommunication *communication,
                                                  MetadataStorage *metadataStorage,
-                                                 bool &erasedMetadata,
                                                  std::shared_ptr<concord::performance::PerformanceManager> pm) {
-  erasedMetadata = false;
   {
     std::lock_guard<std::mutex> lock(mutexForCryptoInitialization);
     if (!cryptoInitialized) {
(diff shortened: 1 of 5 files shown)


kvbc/src/ReplicaImp.cpp
```diff
@@ -63,19 +63,17 @@ Status ReplicaImp::start() {
...
-                 << ", new m_lastReachableBlock = " << getLastReachableBlockNum());
+      replicaConfig_, m_cmdHandler, m_stateTransfer, m_ptrComm, m_metadataStorage, pm_);
+  const auto lastExecutedSeqNum = m_replicaPtr->getLastExecutedSequenceNum();
+  LOG_INFO(logger, KVLOG(lastExecutedSeqNum));
+  if (!replicaConfig_.isReadOnly && !m_stateTransfer->isCollectingState()) {
+    try {
+      uint64_t removedBlocksNum = replicaStateSync_->execute(logger, *m_kvBlockchain, lastExecutedSeqNum);
+      LOG_INFO(logger, KVLOG(lastExecutedSeqNum, removedBlocksNum, getLastBlockNum(), getLastReachableBlockNum()));
+    } catch (std::exception &e) {
+      std::terminate();
```
dartdart26, line 75: Does it make sense to log before calling terminate()?

yuliasherman: "throw" has an information in it

bftengine/src/bftengine/BFTEngine.cpp
```diff
@@ -116,9 +116,7 @@ IReplica::IReplicaPtr IReplica::createNewReplica(const ReplicaConfig &replicaCon
                                                  IStateTransfer *stateTransfer,
                                                  bft::communication::ICommunication *communication,
                                                  MetadataStorage *metadataStorage,
-                                                 bool &erasedMetadata,
```
dartdart26, line 119: Thanks for removing that. It was confusing, especially having the `erasedMetaData` local variable.

dartdart26 (approved):
Thanks for improving the state sync procedure @yuliasherman !

kvbc/src/ReplicaImp.cpp
dartdart26: Yes, I see. Will that end up in the log file, though?

yuliasherman: I'll add the log just to be sure.

PR #1286, opened 2021-03-16
Add storage metrics v0.10.x

yontyon:
we add netrics for number of keys per category in the blockchain level


dartdart26 (approved):
Looks good, thanks!

PR #1289, opened 2021-03-17
Fix race condition reported by tsan. (#1284)

HristoStaykov:
* Fix race condition reported by tsan.

* Added proper locking everywhere we use msgQueue_ and pendingRequest.


yuliasherman (approved):
LGTM

teoparvanov (approved):
LGTM, thanks Hristo!

PR #1290, opened 2021-03-18
Diagnostics server fixes

andrewjstone:
* On destruction of the server join the main thread to prevent
std::terminate from being called.
* Handle errors returned from `accept`
* Add logging to better diagnose production issues
* Add a few tests

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -62,7 +62,7 @@ inline std::string readline(int sock) {
     }
     if (rv < 0 && errno == EINTR) continue;
     if (rv < 0) {
-      throw std::runtime_error("select failed: " + errnoString(rv));
+      throw std::runtime_error("diagnostics server readline select failed: " + errnoString(rv));
     }
 
     if (count == MAX_INPUT_SIZE) {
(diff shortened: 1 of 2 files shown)


andrewjstone:
This should be cherry-picked onto the 0.10.x branch as well

diagnostics/include/diagnostics_server.h
```diff
@@ -90,6 +90,7 @@ inline std::string readline(int sock) {
 
 inline void handleRequest(Registrar& registrar, int sock) {
   try {
+    LOG_INFO(logger, "Handle Diagnostics Request");
```
teoparvanov, line 93: Hi Andrew, AFAI, the diagnostics server is called by the agent every 30 seconds, so I'm afraid this might pollute the logs.
Can we make this a DEBUG log?

diagnostics/include/diagnostics_server.h
```diff
@@ -108,6 +109,7 @@ inline void handleRequest(Registrar& registrar, int sock) {
     }
     close(sock);
   }
+  LOG_INFO(logger, "Finished handling diagnostics request");
```
teoparvanov, line 112: Same here, I'd suggest LOG_DEBUG.

yontyon (approved):
LGTM!

PR #1297, opened 2021-03-18
Disallow less blocks than CPUs in kvbcbench

dartdart26:
If the `total-blocks` configuration option is set to a value that is
less than the number of CPUs in the system, the kvbcbench tool crashes
due to:
 * divide by 0 error when calculating max read offsets
 * invalid `blocks_per_thread` calculation, leading to no data being
   generated (provided the first issue is fixed)

Avoid above by disallowing `total-blocks` being less than the CPUs in
the system and by intoducing the `randomReadIter()` function that
returns cbegin() if we want to read all keys in the vector, avoiding
divide by 0.

### DIFF
--- a/kvbc/benchmark/kvbcbench/main.cpp
+++ b/kvbc/benchmark/kvbcbench/main.cpp
@@ -20,6 +20,7 @@
 #include <random>
 
 #include <boost/program_options.hpp>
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
Looks great. Thanks for the fix @dartdart26

PR #1304, opened 2021-03-22
Add batching to the new client

Liorgal28:
As part of the work on client-side batching, we need to support a batch of requests also as part of the new bft client.
on this PR:

1) Added a method that creates a ClientBatchRequestMsg from several requests.
2) Add clientRequest struct which holds the necessary part for every request.
3) Added sendBatch method which gets queue of clientRequest and returns a queue of replies for each of the requests.
4) Refactored the wait method to support several requests and return the replies through a queue of replies.
5) Add unit test for the batch scenario.

### DIFF
--- a/bftclient/include/bftclient/bft_client.h
+++ b/bftclient/include/bftclient/bft_client.h
@@ -44,7 +44,7 @@ class Client {
   // Throws a BftClientException on error.
   Reply send(const WriteConfig& config, Msg&& request);
   Reply send(const ReadConfig& config, Msg&& request);
-  std::deque<Reply> sendBatch(std::deque<ClientRequest>& clientRequests, const std::string& cid);
+  std::deque<Reply> sendBatch(std::deque<WriteRequest>& write_requests, const std::string& cid);
   bool isServing(int numOfReplicas, int requiredNumOfReplicas) const;
 
   // Useful for testing. Shouldn't be relied on in production.
@@ -58,6 +58,9 @@ class Client {
   //
   // Inserts the Replies to the input queue.
   void wait(std::deque<Reply>& replies);
+
+  // Return a Reply on quorum, or std::nullopt on timeout.
+  std::optional<Reply> wait();
 
   // Extract a matcher configurations from operational configurations
   //
@@ -74,18 +77,22 @@ class Client {
   Msg createClientMsg(const RequestConfig& req_config, Msg&& request, bool read_only, uint16_t client_id);
 
   // This function creates a ClientBatchRequestMsg.
-  Msg createClientBatchMsg(const std::deque<Msg>& clientRequests,
-                           uint32_t batchBufSize,
+  Msg createClientBatchMsg(const std::deque<Msg>& client_requests,
+                           uint32_t batch_buf_size,
                            const string& cid,
                            uint16_t client_id);
+
+  Msg initBatch(std::deque<WriteRequest>& write_requests,
+                const std::string& cid,
+                std::chrono::milliseconds& max_time_to_wait);
 
   MsgReceiver receiver_;
 
   std::unique_ptr<bft::communication::ICommunication> communication_;
   ClientConfig config_;
   logging::Logger logger_ = logging::getLogger("bftclient");
-  std::deque<Msg> pendingRequests_;
-  std::unordered_map<uint64_t, Matcher> replyCertificates_;
+  std::deque<Msg> pending_requests_;
+  std::unordered_map<uint64_t, Matcher> reply_certificates_;
 
   // The client doesn't always know the current primary.
   std::optional<ReplicaId> primary_;
--- a/bftclient/include/bftclient/bft_client.h
+++ b/bftclient/include/bftclient/bft_client.h
@@ -44,6 +44,7 @@ class Client {
   // Throws a BftClientException on error.
   Reply send(const WriteConfig& config, Msg&& request);
   Reply send(const ReadConfig& config, Msg&& request);
+  std::deque<Reply> sendBatch(std::deque<ClientRequest>& clientRequests, const std::string& cid);
   bool isServing(int numOfReplicas, int requiredNumOfReplicas) const;
 
   // Useful for testing. Shouldn't be relied on in production.
@@ -55,8 +56,8 @@ class Client {
 
   // Wait for messages until we get a quorum or a retry timeout.
   //
-  // Return a Reply on quorum, or std::nullopt on timeout.
-  std::optional<Reply> wait();
+  // Inserts the Replies to the input queue.
+  void wait(std::deque<Reply>& replies);
 
   // Extract a matcher configurations from operational configurations
   //
@@ -72,11 +73,19 @@ class Client {
   // to return the messages as vectors with proper RAII based memory management.
   Msg createClientMsg(const RequestConfig& req_config, Msg&& request, bool read_only, uint16_t client_id);
 
+  // This function creates a ClientBatchRequestMsg.
+  Msg createClientBatchMsg(const std::deque<Msg>& clientRequests,
+                           uint32_t batchBufSize,
+                           const string& cid,
+                           uint16_t client_id);
+
   MsgReceiver receiver_;
 
   std::unique_ptr<bft::communication::ICommunication> communication_;
   ClientConfig config_;
   logging::Logger logger_ = logging::getLogger("bftclient");
+  std::deque<Msg> pendingRequests_;
+  std::unordered_map<uint64_t, Matcher> replyCertificates_;
 
   // The client doesn't always know the current primary.
   std::optional<ReplicaId> primary_;
--- a/bftclient/include/bftclient/config.h
+++ b/bftclient/include/bftclient/config.h
@@ -92,7 +92,7 @@ struct ReadConfig {
   ReadQuorum quorum;
 };
 
-struct ClientRequest {
+struct WriteRequest {
   WriteConfig config;
   Msg request;
 };
--- a/bftclient/include/bftclient/config.h
+++ b/bftclient/include/bftclient/config.h
@@ -92,4 +92,9 @@ struct ReadConfig {
   ReadQuorum quorum;
 };
 
+struct ClientRequest {
+  WriteConfig config;
+  Msg request;
+};
+
 }  // namespace bft::client
--- a/bftclient/include/bftclient/exception.h
+++ b/bftclient/include/bftclient/exception.h
@@ -41,8 +41,12 @@ class TimeoutException : public BftClientException {
   TimeoutException(uint64_t seq_num, const std::string& cid)
       : BftClientException("Timeout for request sequence number: " + std::to_string(seq_num) +
                            ", and correlation id: " + cid) {}
+};
 
-  TimeoutException(const std::string& cid) : BftClientException("Timeout for correlation id: " + cid) {}
+class BatchTimeoutException : public BftClientException {
+ public:
+  BatchTimeoutException(const std::string& cid)
+      : BftClientException("Timeout for a batch request with correlation id: " + cid) {}
 };
 
 class InvalidPrivateKeyException : public BftClientException {
--- a/bftclient/include/bftclient/exception.h
+++ b/bftclient/include/bftclient/exception.h
@@ -41,6 +41,8 @@ class TimeoutException : public BftClientException {
   TimeoutException(uint64_t seq_num, const std::string& cid)
       : BftClientException("Timeout for request sequence number: " + std::to_string(seq_num) +
                            ", and correlation id: " + cid) {}
+
+  TimeoutException(const std::string& cid) : BftClientException("Timeout for correlation id: " + cid) {}
 };
 
 class InvalidPrivateKeyException : public BftClientException {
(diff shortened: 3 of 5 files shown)


bftclient/src/bft_client.cpp
```diff
@@ -153,45 +184,82 @@ Reply Client::send(const MatchConfig& match_config,
...
+    }
+
+    wait(replies);
+    metrics_.retransmissions.Get().Inc();
+  }
+  replyCertificates_.clear();
+  if (replies.size() == pendingRequests_.size()) return replies;
+
+  expected_commit_time_ms_.add(max_time_to_wait.count());
+  replyCertificates_.clear();
```
andrewjstone, line 240: This call is a duplicate of the call on line 236.

bftclient/src/bft_client.cpp
```diff
@@ -153,45 +184,82 @@ Reply Client::send(const MatchConfig& match_config,
...
+      for (const auto& d : match_config.quorum.destinations) {
+        dests.emplace(d.val);
+      }
+      communication_->send(dests, std::move(msg));
+    }
+
+    wait(replies);
+    metrics_.retransmissions.Get().Inc();
+  }
+  replyCertificates_.clear();
```
andrewjstone, line 236: You need to track the duration of this request as in the original `send` calls and add it to `expected_commit_time_ms`. This mechanism governs retries and without tracking successful calls the bft client will never send a retry for batched requests which may result in failure to trigger view change.

bftclient/test/bft_client_api_tests.cpp
```diff
@@ -362,6 +362,53 @@ TEST_F(ClientApiTestFixture, write_f_plus_one) {
   client.stop();
 }
 
+TEST_F(ClientApiTestFixture, batch_of_writes) {
```
andrewjstone, line 365: This is a great "successful" test case Lior.

However, I don't think it's sufficient to cover this new feature. We need at least one negative test as well that covers failure cases. If there was a behavior for batches similar to the RetryBehavior for single requests, we may have caught the bug with regards to dynamic timeouts in `sendBatch`. Please add a test to ensure retries get sent to all replicas so we don't delay view change. Thanks!

andrewjstone (changes requested):
Lior, I really appreciate you getting this PR out. Overall the structure and code looks good. As you can tell from my review there are some critical subtleties that need fixing and testing though.

bftclient/include/bftclient/config.h
```diff
@@ -92,4 +92,9 @@ struct ReadConfig {
   ReadQuorum quorum;
 };
 
+struct ClientRequest {
```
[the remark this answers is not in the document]
Liorgal28: Changed to WriteRequest

bftclient/src/bft_client.cpp
```diff
@@ -116,15 +116,45 @@ Msg Client::createClientMsg(const RequestConfig& config, Msg&& request, bool rea
   return msg;
 }
 
+Msg Client::createClientBatchMsg(const std::deque<Msg>& clientRequests,
+                                 uint32_t batchBufSize,
+                                 const string& cid,
+                                 uint16_t client_id) {
+  auto header_size = sizeof(ClientBatchRequestMsgHeader);
+  auto msg_size = sizeof(ClientBatchRequestMsgHeader) + cid.size() + batchBufSize;
+  const auto& numOfMessagesInBatch = clientRequests.size();
```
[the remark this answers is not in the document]
Liorgal28: I returned the wait() method as you suggested, and therefore I removed the changes in the regular send method.

bftclient/src/bft_client.cpp
```diff
@@ -134,14 +164,15 @@ Reply Client::send(const MatchConfig& match_config,
...
   metrics_.retransmissionTimer.Get().Set(expected_commit_time_ms_.upperLimit());
   metrics_.updateAggregator();
-  outstanding_request_ = Matcher(match_config);
+  replyCertificates_.insert(std::make_pair(request_config.sequence_number, Matcher(match_config)));
   receiver_.activate(request_config.max_reply_size);
   auto orig_msg = createClientMsg(request_config, std::move(request), read_only, config_.id.val);
   auto start = std::chrono::steady_clock::now();
   auto end = start + request_config.timeout;
-  while (std::chrono::steady_clock::now() < end) {
+  while (std::chrono::steady_clock::now() < end && replies.size() != 1) {
```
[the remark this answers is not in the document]
Liorgal28: I returned the wait() method as you suggested, and therefore I removed the changes in the regular send method.

bftclient/src/bft_client.cpp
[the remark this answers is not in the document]
Liorgal28: Removed this variable.

bftclient/src/bft_client.cpp
```diff
@@ -153,45 +184,82 @@ Reply Client::send(const MatchConfig& match_config,
...
+    pendingRequests_.push_back(createClientMsg(req.config.request, std::move(req.request), false, config_.id.val));
+    if (req.config.request.timeout > max_time_to_wait) max_time_to_wait = req.config.request.timeout;
+    if (req.config.request.max_reply_size > max_reply_size) max_reply_size = req.config.request.max_reply_size;
+  }
+  receiver_.activate(max_reply_size);
+  uint32_t batchBufSize = 0;
+  for (auto const& msg : pendingRequests_) batchBufSize += msg.size();
+  auto start = std::chrono::steady_clock::now();
+  auto end = start + max_time_to_wait;
+  auto batch_msg = createClientBatchMsg(pendingRequests_, batchBufSize, cid, config_.id.val);
```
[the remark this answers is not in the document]
Liorgal28: `InitBatch` has been added as you suggested, and all parts related to this method have been taken.

bftclient/include/bftclient/bft_client.h
```diff
@@ -55,8 +56,8 @@ class Client {
 
   // Wait for messages until we get a quorum or a retry timeout.
   //
-  // Return a Reply on quorum, or std::nullopt on timeout.
-  std::optional<Reply> wait();
```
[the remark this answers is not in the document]
Liorgal28: As I mentioned before, I used the wait() function as you suggested.

bftclient/include/bftclient/bft_client.h
```diff
@@ -44,6 +44,7 @@ class Client {
   // Throws a BftClientException on error.
   Reply send(const WriteConfig& config, Msg&& request);
   Reply send(const ReadConfig& config, Msg&& request);
+  std::deque<Reply> sendBatch(std::deque<ClientRequest>& clientRequests, const std::string& cid);
```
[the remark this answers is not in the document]
Liorgal28: Done.

bftclient/include/bftclient/bft_client.h
```diff
@@ -72,11 +73,19 @@ class Client {
   // to return the messages as vectors with proper RAII based memory management.
   Msg createClientMsg(const RequestConfig& req_config, Msg&& request, bool read_only, uint16_t client_id);
 
+  // This function creates a ClientBatchRequestMsg.
+  Msg createClientBatchMsg(const std::deque<Msg>& clientRequests,
```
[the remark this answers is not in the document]
Liorgal28: Done.

bftclient/include/bftclient/bft_client.h
```diff
@@ -72,11 +73,19 @@ class Client {
...
+                           uint32_t batchBufSize,
+                           const string& cid,
+                           uint16_t client_id);
+
   MsgReceiver receiver_;
 
   std::unique_ptr<bft::communication::ICommunication> communication_;
   ClientConfig config_;
   logging::Logger logger_ = logging::getLogger("bftclient");
+  std::deque<Msg> pendingRequests_;
```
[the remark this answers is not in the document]
Liorgal28: Done.

bftclient/include/bftclient/bft_client.h
```diff
@@ -72,11 +73,19 @@ class Client {
...
+  std::unordered_map<uint64_t, Matcher> replyCertificates_;
```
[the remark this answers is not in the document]
Liorgal28: Done.

bftclient/include/bftclient/exception.h
```diff
@@ -41,6 +41,8 @@ class TimeoutException : public BftClientException {
   TimeoutException(uint64_t seq_num, const std::string& cid)
       : BftClientException("Timeout for request sequence number: " + std::to_string(seq_num) +
                            ", and correlation id: " + cid) {}
+
+  TimeoutException(const std::string& cid) : BftClientException("Timeout for correlation id: " + cid) {}
```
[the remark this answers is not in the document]
Liorgal28: Added custom `BatchTimeoutException` as you suggested.

bftclient/src/bft_client.cpp
```diff
@@ -153,45 +184,82 @@ Reply Client::send(const MatchConfig& match_config,
...
+      if (auto match = request->second.onReply(std::move(reply))) {
         primary_ = match->primary;
-        outstanding_request_ = std::nullopt;
-        return match->reply;
+        replies.push_back(match->reply);
+        replyCertificates_.erase(request->first);
       }
     }
   }
-  // If there are multiple distinct replies, we may want to clear any replies to free memory. This really only matters
```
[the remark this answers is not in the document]
Liorgal28: Added this part to the single request wait() method.

bftclient/src/bft_client.cpp
Liorgal28: Done.

bftclient/src/bft_client.cpp
Liorgal28: Done.

bftclient/test/bft_client_api_tests.cpp
Liorgal28: Added failure scenario:
1) sendBatch
2)The behaviour does not return replies.
3)Timeout reached
4)catch `BatchTimeoutException`.

bftclient/src/bft_client.cpp
```diff
@@ -156,42 +204,88 @@ Reply Client::send(const MatchConfig& match_config,
...
 
+std::deque<Reply> Client::sendBatch(std::deque<WriteRequest>& write_requests, const std::string& cid) {
+  std::deque<Reply> replies;
+  std::chrono::milliseconds max_time_to_wait = 0s;
+  MatchConfig match_config = writeConfigToMatchConfig(write_requests.front().config);
+  auto batch_msg = initBatch(write_requests, cid, max_time_to_wait);
+  auto start = std::chrono::steady_clock::now();
+  auto end = start + max_time_to_wait;
+  while (std::chrono::steady_clock::now() < end && replies.size() != pending_requests_.size()) {
+    bft::client::Msg msg(batch_msg);  // create copy here due to the loop
```
yuliasherman, line 226: Why do we need to have a copy here?

bftclient/src/bft_client.cpp
```diff
@@ -156,42 +204,88 @@ Reply Client::send(const MatchConfig& match_config,
...
+    if (primary_) {
+      communication_->send(primary_.value().val, std::move(msg));
+    } else {
+      std::set<bft::communication::NodeNum> dests;
+      for (const auto& d : match_config.quorum.destinations) {
+        dests.emplace(d.val);
```
yuliasherman, line 232: Is it possible that the element already exists in dests and you just want to update it? If positive, use add_or_insert instead of emplace.

bftclient/src/bft_client.cpp
```diff
@@ -156,42 +204,88 @@ Reply Client::send(const MatchConfig& match_config,
...
+  expected_commit_time_ms_.add(max_time_to_wait.count());
+  throw BatchTimeoutException(cid);
+}
+
 std::optional<Reply> Client::wait() {
+  std::deque<Reply> replies;
+  wait(replies);
+  if (replies.empty()) {
+    static constexpr size_t CLEAR_MATCHER_REPLIES_THRESHOLD = 5;
+    // reply_certificates_ should hold just one request when using this wait method
```
yuliasherman, line 256: Irrelevant comment?

yuliasherman (commented):
LGTM

bftclient/src/bft_client.cpp
Liorgal28: Actually, this is relevant, this wait method should be called only to non-batched requests so the reply_certificates_ suppose to hold just one request.

bftclient/src/bft_client.cpp
Liorgal28: The dests set created 2 rows before so it's supposed to be clean and the match config destinations should not hold duplicate destinations.

bftclient/src/bft_client.cpp
yuliasherman: So, why do we have a 'for' loop a line below?

bftclient/src/bft_client.cpp
Liorgal28: We have like retry mechanism, we are sending the copied message to the replicas then we are going into the wait() method and waiting for replies, if replies don't arrive (we got dynamic utility for replies timeouts) and we still have time we are sending again the same message.
since we are moving the message so we need to move copies.

bftclient/src/bft_client.cpp
Liorgal28: Because I did it wrong, I'll fix it

andrewjstone (approved):
Looks great! Thanks for all the fixes @Liorgal28.

PR #1305, opened 2021-03-23
Add method to kv_editor for getting number of stale keys

yontyon:
Add a new method for kv_bolckchain_editor.
This method allows us to get the number of all stale keys (stale and purpose and override keys). This will give the testing team the ability to test pruning

### DIFF
--- a/kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
+++ b/kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
@@ -494,13 +494,76 @@ TEST_F(DbEditorTests, get_category_earliest_stale_immutable) {
   ASSERT_THAT(out_.str(), EndsWith("\n}\n"));
 }
 
-TEST_F(DbEditorTests, get_summary_stake_keys) {
+TEST_F(DbEditorTests, get_summary_stale_keys_invalid_argumets) {
+  ASSERT_EQ(EXIT_FAILURE,
+            run(CommandLineArguments{{kTestName, rocksDbPath(main_path_db_id_), "getStaleKeysSummary", "0", "5"}},
+                out_,
+                err_));
+  ASSERT_TRUE(out_.str().empty());
+  ASSERT_THAT(err_.str(), HasSubstr("BLOCK-VERSION-FROM is incorrect: current genesis: 1"));
+  ASSERT_THAT(err_.str(), HasSubstr("BLOCK-VERSION-FROM: 0"));
+
+  ASSERT_EQ(EXIT_FAILURE,
+            run(CommandLineArguments{{kTestName, rocksDbPath(main_path_db_id_), "getStaleKeysSummary", "1", "1000"}},
+                out_,
+                err_));
+  ASSERT_TRUE(out_.str().empty());
+  ASSERT_THAT(err_.str(), HasSubstr("latest reachable block: 10"));
+  ASSERT_THAT(err_.str(), HasSubstr("BLOCK-VERSION-TO: 1000"));
+
+  ASSERT_EQ(
+      EXIT_FAILURE,
+      run(CommandLineArguments{{kTestName, rocksDbPath(main_path_db_id_), "getStaleKeysSummary", "1000"}}, out_, err_));
+  ASSERT_TRUE(out_.str().empty());
+  ASSERT_THAT(err_.str(), HasSubstr("latest reachable block: 10"));
+  ASSERT_THAT(err_.str(), HasSubstr("BLOCK-VERSION-TO: 1000"));
+}
+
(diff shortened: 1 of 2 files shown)


kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -392,6 +392,40 @@ struct GetCategoryEarliestStale {
   }
 };
 
+struct GetStaleKeysSummary {
+  const bool read_only = true;
+  std::string description() const {
+    return "getStaleKeysSummary [BLOCK-VERSION-TO]\n"
+           "  return the number of stale keys per category type in the current blockchain\n"
```
dartdart26, line 399: Nit: Could you please use capital `R` and use `.t` at the end?

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -392,6 +392,40 @@ struct GetCategoryEarliestStale {
...
+           "If BLOCK-VERSION-TO has been given, the method will summerize the stale keys from current genesis to "
```
dartdart26, line 400: Nit: Could you please indent it?

Also, `summerize` is a typo. Maybe we can use `sum` instead.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -392,6 +392,40 @@ struct GetCategoryEarliestStale {
...
+           "BLOCK-VERSION-TO";
+  }
+
+  std::string execute(KeyValueBlockchain &adapter, const CommandArguments &args) const {
+    auto latestBlockID = adapter.getLastReachableBlockId();
+    if (args.values.size() >= 1) {
+      latestBlockID = toBlockId(args.values.front());
```
dartdart26, line 407: I think we need to make sure that the argument's value is the range `[getGenesisBlockId(), getLastReachableBlockId()]`. I think we can add a test for it.

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -392,6 +392,40 @@ struct GetCategoryEarliestStale {
   }
 };
 
+struct GetStaleKeysSummary {
+  const bool read_only = true;
+  std::string description() const {
+    return "getStaleKeysSummary [BLOCK-VERSION-TO]\n"
```
dartdart26, line 398: We can add a note to the description that mentions to users that this is essentially a linear search and it might take a lot of time.

kvbc/test/sparse_merkle_storage/kv_blockchain_db_editor_test.cpp
```diff
@@ -494,6 +494,16 @@ TEST_F(DbEditorTests, get_category_earliest_stale_immutable) {
   ASSERT_THAT(out_.str(), EndsWith("\n}\n"));
 }
 
+TEST_F(DbEditorTests, get_summary_stake_keys) {
```
dartdart26, line 497: Typo: -> get_stale_keys_summary

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -392,6 +392,40 @@ struct GetCategoryEarliestStale {
...
+    }
+    for (auto block = adapter.getGenesisBlockId(); block <= latestBlockID; block++) {
+      auto stale_keys = adapter.getBlockStaleKeys(block);
+      for (const auto &[cat_id, cat_type] : categories) {
+        stale_keys_per_category_type_[cat_type] += stale_keys[cat_id].size();
+      }
+    }
+    std::map<std::string, std::string> out;
+    for (auto const &[cat_type, num_of_stale] : stale_keys_per_category_type_) {
+      out[cat_type_str.at(cat_type)] = std::to_string(num_of_stale);
```
dartdart26, line 423: Now that I think of it, since we can have multiple categories of the same type, we should return a map of `category ID` -> `count` instead.

yontyon: The requirement is to have the number of stale key per category type and not per category ID. This way, the testing team can compare the number of stale keys in, for example, merkle tree keys and the number of deleted of that type

dartdart26: I understand, but, in general, I think it makes sense to expose this information to users. Currently, there is a use case where we have one category of every type, but that might change.

We can leave as is for now, but I think it won't hurt to expose every separate category and its type. Users can then just sum.

dartdart26 (approved):
Thanks @yontyon, looks good!

PR #1306, opened 2021-03-23
Add method for getting stale keys

yontyon:
same as #1305


dartdart26 (approved):
Thanks @yontyon, looks good!

PR #1309, opened 2021-03-23
Fix to ask from peers the View Change messages for the highest View we have reached.

HristoStaykov:
myLatestPendingView reflects the last View for which we have collected 2F+2C+1 ViewChange messages
and might be behind the last view we have moved to.


andrewjstone (approved):
LGTM

PR #1310, opened 2021-03-23
Fix to ask from peers the View Change messages for the highest View w…

HristoStaykov:
…e have reached.

myLatestPendingView reflects the last View for which we have collected 2F+2C+1 ViewChange messages
and might be behind the last view we have moved to.


andrewjstone (changes requested):
@HristoStaykov This change looks good. Thanks for the fix!

However, can you please reformat/reword the commit message in accordance with https://chris.beams.io/posts/git-commit/

andrewjstone (approved):
Looks great. Thanks @HristoStaykov

PR #1311, opened 2021-03-23
Fix taking in account that we might receive and exchange complaints

HristoStaykov:
when the View is active and inactive.


f-squirrel (approved):
LGTM!

PR #1312, opened 2021-03-23
Fix taking in account that we might receive and exchange complaints

HristoStaykov:
when the View is active and inactive.

### DIFF
--- a/bftengine/src/bftengine/messages/ReplicaStatusMsg.cpp
+++ b/bftengine/src/bftengine/messages/ReplicaStatusMsg.cpp
@@ -44,9 +44,9 @@ MsgSize ReplicaStatusMsg::calcSizeOfReplicaStatusMsg(bool viewIsActive,
   else if (viewIsActive)
     return sizeof(ReplicaStatusMsg::Header) + maxNumberOfReplicasBitMaskSize;
   else if (listOfMissingViewChangeMsgForViewChange)
-    return sizeof(ReplicaStatusMsg::Header) + maxNumberOfReplicasBitMaskSize;
+    return sizeof(ReplicaStatusMsg::Header) + (2 * maxNumberOfReplicasBitMaskSize);
   else if (listOfMissingPrePrepareMsgForViewChange)
-    return sizeof(ReplicaStatusMsg::Header) + kWorkWindowBitMaskSize;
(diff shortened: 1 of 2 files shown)


andrewjstone (changes requested):
Hristo,

I'm sorry but this commit message is not sufficient to describe a change to the consensus mechanism. Please describe what the current problem is as well as why this change resolves the issue. Without that context it is hard to review properly, and future developers will not be able to understand the commit either when using git history/blame.

Please use the format described in [this post](https://chris.beams.io/posts/git-commit/) as described in [CONTRIBUTING.md](https://github.com/vmware/concord-bft/blob/master/CONTRIBUTING.md).

bftengine/src/bftengine/messages/ReplicaStatusMsg.cpp
```diff
@@ -44,9 +44,9 @@ MsgSize ReplicaStatusMsg::calcSizeOfReplicaStatusMsg(bool viewIsActive,
   else if (viewIsActive)
     return sizeof(ReplicaStatusMsg::Header) + maxNumberOfReplicasBitMaskSize;
   else if (listOfMissingViewChangeMsgForViewChange)
-    return sizeof(ReplicaStatusMsg::Header) + maxNumberOfReplicasBitMaskSize;
+    return sizeof(ReplicaStatusMsg::Header) + (2 * maxNumberOfReplicasBitMaskSize);
```
andrewjstone, line 47: why `2*`?

HristoStaykov: We might receive and exchange Complaints at all times - both when the View is active and if it is not active in both stages (when we are waiting for quorum of View Change messages and for the list of PrePrepares). The highlighted part is the case when our View is inactive and we are waiting for a quorum of View Change messages to proceed. While we do this it is possible to also receive Complaints (from someone who is ahead of us). We need a bit mask to indicate both if we have missing View Change messages from each Replica and what Complaints we currently have for the View again from each peer Replica, therefore 1 maxNumberOfReplicasBitMaskSize for the View Change messages and 1 maxNumberOfReplicasBitMaskSize for the Complaints.

dartdart26 (approved):
Thanks for the fix, commit message and the offline explanations @HristoStaykov!

PR #1314, opened 2021-03-24
[Proposal]: Add support for concurrent pruning and block adds

dartdart26:
Add support for concurrent pruning and block addition. In this context,
concurrent means alternating between additions and pruning, but never
executing them trully in parallel, from different threads. Rationale is:
 * no changes to storage are required
 * simplifies the implementation
 * safer

Give block addition priority over pruning by setting an atomic flag when
an addition is pending. Once seen by the pruning thread, it will yield
and let the addition proceed. When the addition finishes, it will signal
the prune thread and it can continue pruning. Doing so gives
transactions and state transfer priority over pruning.

Another possible implementation is to have a single thread call add and
delete methods and have a queue of commands for it. However, switching
between threads for the case of no pruning proves more expensive than
what the proposal in this PR does - just set an atomic, lock an unlocked
mutex and signal a condition variable that has no threads waiting on it
(in case of no pruning).

A note on the atomic `pending_add_` flag is that it is intentionally set
before locking `write_mtx_`. That way, contention for the mutex can be
resolved and we don't have to rely on a fair mutex policy.

In temrs of methods, users can call the newly-introduced
`deleteBlocksUntil()` method from a different thread that executes block
addition. More details about multi-threading and pruning is described in
a code comment.

Implement ReplicaImp::deleteBlocksUntil() by just calling
KeyValueBlockchain::deleteBlocksUntil().

Update `kvbcbench` so that it supports pruning and add options to
configure it to run after block addition or concurrently with it.

Add unit tests for the new behavior.

### DIFF
--- a/kvbc/benchmark/kvbcbench/main.cpp
+++ b/kvbc/benchmark/kvbcbench/main.cpp
@@ -337,7 +337,11 @@ void addBlocks(const po::variables_map& config,
         kvbc.addBlock(std::move(updates));
       } else {
         auto&& merkle_input = std::move(input.block_merkle_input[i - 1]);
+        auto&& versioned_updates = std::move(input.ver_updates[i - 1]);
+        auto&& immutable_updates = std::move(input.imm_updates[i - 1]);
         updates.add(kCategoryMerkle, categorization::BlockMerkleUpdates(std::move(merkle_input)));
+        updates.add(kCategoryVersioned, std::move(versioned_updates));
+        updates.add(kCategoryImmutable, std::move(immutable_updates));
         kvbc.addBlock(std::move(updates));
       }
     }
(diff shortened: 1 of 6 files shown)


dartdart26:
Will add a commit to this PR that will return nullopt when calling get() and multiGet() methods for versions less than the one passed to `deleteBlocksUntil()` as pruning can progress on its own on separate replicas and the return of these methods can, therefore, differ.

dartdart26:
In an offline discussion, @ittaiab made a great observation that we update the merkle tree in the block merkle category on pruning, making the proposal in this PR non-deterministic in relation to the tree root hash. Reason is that separate replicas can progress with pruning at different rates and, therefore, when adding the next block, the tree's root can be different across replicas.

PR #1317, opened 2021-03-24
ReplicaImp: modify ReplicaImp::getReplicaState

glevkovich:
1)
Former ReplicaImp::getReplicaState is renamed to
ReplicaImp::getReplicaLastStableSeqNum

2) ReplicaImp::getReplicaState now returns the minimal needed info for
agent. It returns a json formated string in the format:
{
  "Sequence Numbers ": {
  "lastStableSeqNum": "0"
    }
}

The reason for the change is data races between the message processing
thread and ST thread. which potentially cause later on crashes
(probably due to caused data corruptions).
For now, we want to minimize the probability for race condition and
apply a short-term fix which greatly reduces the change for
race-condition. Later on, we will apply a different log-term fix.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1115,12 +1115,27 @@ void ReplicaImp::onInternalMsg(InternalMessage &&msg) {
   ConcordAssert(false);
 }
 
+std::string ReplicaImp::getReplicaLastStableSeqNum() const {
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
Looks great. Thanks for the fix @glevkovich. Solid teamwork with @toly-kournik and @HristoStaykov.

HristoStaykov (approved):
LGTM, thanks @glevkovich !

PR #1321, opened 2021-03-25
Pause status request during collecting state.

HristoStaykov:
For the destination replica in State Transfer we delegate the
access to state variables to the BCStateTran. We need to pause
this timer as we do for all other timers.


toly-kournik (approved):
LGTM, thanks

PR #1322, opened 2021-03-26
CMF: Add 'Enum' type

andrewjstone:
A change was made to the CMF grammar to introduce a new top-level type:
`Enum`. An Enum creates an automatically numbered set of tags starting
at 0. Similar to messsages, Enum's can be directly used in fields and
compound types. We still only allow Msgs in oneofs however.

This choice to introduce a top-level type was made to allow including
Enums directly in messages rather than declaring them in a message and
then having to refer indirectly to a container messsage. It's unclear to
me whether this was the best decision, or whether we should only have
`Msg`s as top-level types. I'm open to revisiting this if need be. It
should be noted that even if we decide to change the grammar, the on the
wire format will not change.

Support for C++ and Python was added, along with tests and updated
documentation.

### DIFF
--- a/messages/compiler/cpp/serialize.cpp
+++ b/messages/compiler/cpp/serialize.cpp
@@ -84,6 +84,26 @@ void deserialize(const uint8_t*& start, const uint8_t* end, T& t) {
 }
 
 /******************************************************************************
+ * Enums
+ *
+ * Enums are type wrappers around uint8_t
+ ******************************************************************************/
+template <typename T, typename std::enable_if<std::is_enum<T>::value>::type* = nullptr>
+void serialize(std::vector<uint8_t>& output, const T& t) {
+  serialize(output, static_cast<uint8_t>(t));
+}
+
+template <typename T, typename std::enable_if<std::is_enum<T>::value>::type* = nullptr>
+void deserialize(const uint8_t*& start, const uint8_t* end, T& t) {
+  uint8_t val;
+  deserialize(start, end, val);
+  t = static_cast<T>(val);
+  if (val >= enumSize(t)) {
+    throw BadDataError(std::string("Value < ") + std::to_string(enumSize(t)), std::to_string(val));
+  }
+}
+
+/******************************************************************************
  * Strings
  *
  * Strings are preceded by a uint32_t length
--- a/messages/README.md
+++ b/messages/README.md
@@ -19,18 +19,25 @@ generates an [Abstract Syntax Tree (AST)](https://en.wikipedia.org/wiki/Abstract
 the parsed messages according to the grammar. The parser generator finds syntax errors, and some
 basic typechecking is done via a [semantics plugin](compiler/semantics.py).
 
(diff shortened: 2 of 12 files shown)


messages/compiler/cpp/serialize.cpp
```diff
@@ -83,6 +83,23 @@ void deserialize(const uint8_t*& start, const uint8_t* end, T& t) {
...
+template <typename T, typename std::enable_if<std::is_enum<T>::value>::type* = nullptr>
+void serialize(std::vector<uint8_t>& output, const T& t) {
+  serialize(output, static_cast<uint8_t>(t));
+}
+
+template <typename T, typename std::enable_if<std::is_enum<T>::value>::type* = nullptr>
+void deserialize(const uint8_t*& start, const uint8_t* end, T& t) {
+  uint8_t val;
+  deserialize(start, end, val);
+  t = static_cast<T>(val);
```
dartdart26, line 100: Does it make sense to check if the deserialised value is valid for type T (like we do for oneof)? It can be done with some C++ trickery or adding an ugly-named last option to the enum and checking the value against the enum range.

dartdart26 (approved):
Looks great, thanks!

Personally, I prefer enums being top-level as in this PR.

I've added one comment only about verifying the deserialized enum value. I don't think it is mandatory as we don't support different messages versions. However, It might make sense to protect from a malicious party, especially when processing messages and expecting the static C++ enum class tags only. Also, depending on the standard version, it seems it can be undefined behaviour if we assign arbitrary values to enum classes.

messages/compiler/cpp/serialize.cpp
andrewjstone: It's a really good point, especially considering we do check this in python. I'm not sure we can just add an extra parameter due to recursive calls, but maybe we can. C++ trickery always an option. I'll think about it. Thanks for the review!

dartdart26: What I meant was adding the option to all enums, something like:

```
enum class Color {
  red = 0,
  green = 1,
  blue = 2,
  __UGLY_LONG_AND_UNUSED_NAME_AS_LAST_OPTION = 3
};
```

and then compare for the [0, T::__UGLY_LONG_AND_UNUSED_NAME_AS_LAST_OPTION] range.

andrewjstone: Ah, yeah i did think of that. I believe I can do it with a function template specialization that returns the max size and then using that as a check in the deserialize function. I'm attempting that now.

robem (approved):
Looks good to me. Thanks for adding the tests and updating the docs 👍

messages/compiler/cpp/serialize.cpp
andrewjstone: I added support for namespace scoped [`enumSize` functions](https://github.com/vmware/concord-bft/pull/1322/files#diff-2558f8429abbfe0d6f9d30ef7191bfc071b7edaaf043a62dd28bc85dc052794dR191-R192) and used [Argument Dependent Lookup](https://en.cppreference.com/w/cpp/language/adl) to ensure the [checks get called](https://github.com/vmware/concord-bft/pull/1322/files#diff-fdbbe49391692d69f4b4eb144885d711630f1f50ecbd5c12b3b754f2c8011e0cR101-R103). The argument itself gets ignored because the size is computed by the code generation, but it's needed for the ADL so arbitrary user namespaces can be looked in similar to our generated serialize functions for Msgs and oneofs.

PR #1323, opened 2021-03-26
Ignore all but client requests on async pruning - v0.10.x

yontyon:
Pruning can be performed in a different thread in order to allow pruning status messages.
However, when done so, we should treat pruning as blocking and ignore requests except for client requests.
In this PR we prevent all but client requests on the message handler level (we already block write client requests)

This is to prevent problems like:

ST while running pruning
Byzantine replica that tries to run the protocol while other replicas are pruning and so on

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -627,10 +627,6 @@ void BCStateTran::startCollectingState() {
 
   ConcordAssert(running_);
   ConcordAssert(!isFetching());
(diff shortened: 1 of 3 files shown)


toly-kournik (approved):
LGTM

yontyon:
Pruning can be performed in a different thread in order to allow pruning status messages.
However, when done so, we should treat pruning as blocking and ignore requests except for client requests.
In this PR we prevent all but client requests on the message handler level (we already block write client requests)

This is to prevent problems like:

ST while running pruning
Byzantine replica that tries to run the protocol while other replicas are pruning and so on

PR #1324, opened 2021-03-26
Ignore all but client requests on async pruning

yontyon:
Pruning can be performed in a different thread in order to allow pruning status messages.
However, when done so, we should treat pruning as blocking and ignore requests except for client requests.
In this PR we prevent all but client requests on the message handler level (we already block write client requests)

This is to prevent problems like:
1. ST while running pruning
2. Byzantine replica that tries to run the protocol while other replicas are pruning and so on

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -627,10 +627,6 @@ void BCStateTran::startCollectingState() {
 
   ConcordAssert(running_);
   ConcordAssert(!isFetching());
-  if (bftEngine::ControlStateManager::instance().getPruningProcessStatus()) {
-    LOG_INFO(getLogger(), "Unable to perform ST while pruning");
-    return;
-  }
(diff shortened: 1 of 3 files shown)


HristoStaykov (approved):
LGTM, thanks!

dartdart26:
Thanks @yontyon, this PR makes sense. I've discussed the following with @HristoStaykov:

1. Ignoring messages should only lead to a maximum of 1 view change after pruning completes and that seems safe.
2. Ignoring CheckpointMsg should make sure the replica doesn't start fetching blocks from ST (which is what we want as that would mean concurrent writes).
3. AFAIK, ignoring AskForCheckpoint would, however, stop a replica from being a ST source. I think it is beneficial to allow a replica to server blocks over ST while pruning. Otherwise, a replica waiting for ST while others are pruning will be further delayed, as it might have to do pruning itself when it finally gets the blocks from ST. @toly-kournik could you please share your opinion too?

All I am saying is that I am just not entirely sure the above change is 100% safe in terms of all the implications across subsystems it might have. I would propose to discuss more before merging it in.

PR #1328, opened 2021-03-29
Fix undefined behavior due to up casting.

HristoStaykov:
Everywhere we deserialize messages inheriting MessageBase we use
MessageBase::deserializeMsg, which returns a MessageBase pointer.
Up-casting it to the actual Message type is not sufficient, we need
to actually create the proper object and transfer the packed struct
containing the header and dynamic payload to it.

### DIFF
--- a/bftengine/src/bftengine/PersistentStorageDescriptors.cpp
+++ b/bftengine/src/bftengine/PersistentStorageDescriptors.cpp
@@ -277,7 +277,7 @@ void DescriptorOfLastNewView::deserializeElement(uint32_t id, char *buf, size_t 
   std::unique_ptr<MessageBase> msgBase(MessageBase::deserializeMsg(buf, bufLen, actualSize));
   auto msg = new ViewChangeMsg(msgBase.get());
   ConcordAssert(viewChangeMsgs[id] == nullptr);
-  viewChangeMsgs[id] = ((ViewChangeMsg *)msg);
+  viewChangeMsgs[id] = msg;
 }
 
 /***** DescriptorOfLastExecution *****/
--- a/bftengine/src/bftengine/PersistentStorageDescriptors.cpp
+++ b/bftengine/src/bftengine/PersistentStorageDescriptors.cpp
@@ -120,8 +120,10 @@ void DescriptorOfLastExitFromView::deserializeSimpleParams(char *buf, size_t buf
 
   size_t actualMsgSize = 0;
   std::unique_ptr<MessageBase> baseMsg(MessageBase::deserializeMsg(buf, bufLen, actualMsgSize));
-  myViewChangeMsg = new ViewChangeMsg(baseMsg.get());
-
+  myViewChangeMsg = nullptr;
+  if (baseMsg) {
+    myViewChangeMsg = new ViewChangeMsg(baseMsg.get());
+  }
   uint32_t elementsNum;
   size_t elementsNumSize = sizeof(elementsNum);
   memcpy(&elementsNum, buf, elementsNumSize);
@@ -140,11 +142,15 @@ void DescriptorOfLastExitFromView::deserializeElement(uint32_t id, char *buf, si
   PrepareFullMsg *prepareFullMsgPtr = nullptr;
   {
     std::unique_ptr<MessageBase> baseMsg(MessageBase::deserializeMsg(buf, bufLen, msgSize1));
-    prePrepareMsgPtr = new PrePrepareMsg(baseMsg.get());
+    if (baseMsg) {
+      prePrepareMsgPtr = new PrePrepareMsg(baseMsg.get());
+    }
   }
   {
     std::unique_ptr<MessageBase> baseMsg(MessageBase::deserializeMsg(buf, bufLen, msgSize2));
-    prepareFullMsgPtr = new PrepareFullMsg(baseMsg.get());
+    if (baseMsg) {
+      prepareFullMsgPtr = new PrepareFullMsg(baseMsg.get());
+    }
   }
 
   bool hasAllRequests = false;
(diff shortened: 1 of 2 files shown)


KTeofanidis (approved):
LGTM

bftengine/src/bftengine/PersistentStorageDescriptors.cpp
```diff
@@ -261,7 +274,8 @@ void DescriptorOfLastNewView::deserializeElement(uint32_t id, char *buf, size_t
   actualSize = 0;
   ConcordAssert(id < viewChangeMsgsNum);
 
-  auto *msg = MessageBase::deserializeMsg(buf, bufLen, actualSize);
+  std::unique_ptr<MessageBase> msgBase(MessageBase::deserializeMsg(buf, bufLen, actualSize));
+  auto msg = new ViewChangeMsg(msgBase.get());
   ConcordAssert(viewChangeMsgs[id] == nullptr);
   viewChangeMsgs[id] = ((ViewChangeMsg *)msg);
```
dartdart26, line 280: Nit: We can remove the cast now as `msg` is of the proper type.

dartdart26 (approved):
Brilliant, great find, thanks @HristoStaykov ! That is definitely undefined behaviour as the vtable is messed up, sizeof is different, etc. and could lead to who knows what.

bftengine/src/bftengine/PersistentStorageDescriptors.cpp
HristoStaykov: Thanks @dartdart26 ! Will remove it.

tdimitrov:
Really great catch @HristoStaykov

andrewjstone:
> Really great catch @HristoStaykov

+1 Fantastic catch and fix.

f-squirrel (approved):
Amazing catch!

andrewjstone (approved):

andrewjstone:
This fix is buggy because it doesn't copy data out of the allocated base class into the allocated derived class. The base class containing the data goes out of scope. Fixing this would require at least one memcpy AFAICT. 

More problematic, when we allocate a `NewViewMsg` we actually have no idea how big its supposed to be. Elements are added dynamically, but the allocation only allocates the size of a `NewViewMsg` packed struct, and then the macro defined constructor initializes the base class fields. It sets the size of the object in the base class, but that doesn't inform malloc how big the struct is. This is a major problem and I believe the cause of segfaults in our tests, including `testSerialization`. 

I'm actually not sure this fix is necessary at all after thinking more about it. The structs are packed, and we don't use virtual functions to manipulate them, although `validate` is virtual. I believe we should leave things as is here or come up with a better fix. Ideally we'd transition these messages to a standard format ASAP.

dartdart26:
> This fix is buggy because it doesn't copy data out of the allocated base class into the allocated derived class. The base class containing the data goes out of scope. Fixing this would require at least one memcpy AFAICT.
> 
> More problematic, when we allocate a `NewViewMsg` we actually have no idea how big it's supposed to be. Elements are added dynamically, but the allocation only allocates the size of a `NewViewMsg` packed struct, and then the macro defined constructor initializes the base class fields. It sets the size of the object in the base class, but that doesn't inform malloc how big the struct is. This is a major problem and I believe the cause of segfaults in our tests, including `testSerialization`.
> 
> I'm actually not sure this fix is necessary at all after thinking more about it. The structs are packed, and we don't use virtual functions to manipulate them, although `validate` is virtual. I believe we should leave things as is here or come up with a better fix. Ideally we'd transition these messages to a standard format ASAP.

I think the macro constructors "steal" the internal pointer by calling `releaseOwnership()`. It should be correct, as far as I can see.

HristoStaykov:
> > This fix is buggy because it doesn't copy data out of the allocated base class into the allocated derived class. The base class containing the data goes out of scope. Fixing this would require at least one memcpy AFAICT.
> > More problematic, when we allocate a `NewViewMsg` we actually have no idea how big it's supposed to be. Elements are added dynamically, but the allocation only allocates the size of a `NewViewMsg` packed struct, and then the macro defined constructor initializes the base class fields. It sets the size of the object in the base class, but that doesn't inform malloc how big the struct is. This is a major problem and I believe the cause of segfaults in our tests, including `testSerialization`.
> > I'm actually not sure this fix is necessary at all after thinking more about it. The structs are packed, and we don't use virtual functions to manipulate them, although `validate` is virtual. I believe we should leave things as is here or come up with a better fix. Ideally we'd transition these messages to a standard format ASAP.
> 
> About the first issue - I think the macro constructors "steal" the internal pointer by calling `releaseOwnership()`. It should be correct, as far as I can see.
> 
> Definitely agree we should change the way we handle messages.

Yes, we should not copy, we only steal the pointer from the MessageBase object and set the ownership accordingly. MessageBase is designed to be able to store the complete packed struct (both the static header and the dynamic payload) of all its Sub Class objects. Up-casting an object of a Super Class to a Sub Class (what used to happen with the C++ wrapper object) is undefined behavior and might cause problems.
I agree the messaging system might be improved as a long term effort. In the meantime it is best to clear all undefined behavior.

andrewjstone:
> > > This fix is buggy because it doesn't copy data out of the allocated base class into the allocated derived class. The base class containing the data goes out of scope. Fixing this would require at least one memcpy AFAICT.
> > > More problematic, when we allocate a `NewViewMsg` we actually have no idea how big it's supposed to be. Elements are added dynamically, but the allocation only allocates the size of a `NewViewMsg` packed struct, and then the macro defined constructor initializes the base class fields. It sets the size of the object in the base class, but that doesn't inform malloc how big the struct is. This is a major problem and I believe the cause of segfaults in our tests, including `testSerialization`.
> > > I'm actually not sure this fix is necessary at all after thinking more about it. The structs are packed, and we don't use virtual functions to manipulate them, although `validate` is virtual. I believe we should leave things as is here or come up with a better fix. Ideally we'd transition these messages to a standard format ASAP.
> > 
> > 
> > About the first issue - I think the macro constructors "steal" the internal pointer by calling `releaseOwnership()`. It should be correct, as far as I can see.
> > Definitely agree we should change the way we handle messages.
> 
> Yes, we should not copy, we only steal the pointer from the MessageBase object and set the ownership accordingly. MessageBase is designed to be able to store the complete packed struct (both the static header and the dynamic payload) of all its Sub Class objects. Up-casting an object of a Super Class to a Sub Class (what used to happen with the C++ wrapper object) is undefined behavior and might cause problems.
> I agree the messaging system might be improved as a long term effort. In the meantime it is best to clear all undefined behavior.

Thanks for the great explanation @HristoStaykov. You are correct. I totally missed the steal behavior. Great find and fix.

PR #1330, opened 2021-03-29
Sparse Merkle Proptests: Skip precondition failure

andrewjstone:
Rapidcheck allows runtime preconditions for generators for cases where
it's difficult or not possible to generate exactly the right test input
during compile time. Generation is run by default 100 times maximum to
try to meet the preconditions. If the preconditions are not met an
exception is thrown. However, this isn't actually an error as the test
case didn't fail. The generator just didn't generate the proper input to
the test case. We fix these rare failures by catching the exception and
forcing success.

This fix was tested by generating a precondition,
`RC_PRE(key_hashes.size() == 10000ul)` that could never be met and
caused failure on every run. The failing precondition was wrapped in a
try/catch clause allowing the test to pass.


robem (approved):
We need to keep an eye out if skipping happens too often but seems to be a rare occurrence as of now.

PR #1331, opened 2021-03-30
Remove old replies before getting in sync with other replicas after ST

yuliasherman:
- In order not to overload a number of supported requests per client, old replies need to be cleaned from requestsInfo before adding ones received through state transfer.
- While deleting the oldest replies, take into account that replies received via ST don't contain timestamps.


HristoStaykov (approved):
LGTM, thanks @yuliasherman!

andrewjstone:
Awesome!

PR #1333, opened 2021-03-30
Proper error handling for file read/write functions in SecretsManager

tdimitrov:
ISecretsManagerImpl::readFile() doesn't handle errors like missing file
for example. This patch adds proper checks.

Courtesy to @glevkovich


secretsmanager/test/secrets_manager_test.cpp
```diff
@@ -116,4 +117,22 @@ TEST(SecretsManagerEnc, EmptyInput) {
...
+  SecretsManagerEnc sm(ret);
+  auto res = sm.decryptFile("/path/which/doesnt/exist");
+  ASSERT_FALSE(res.has_value());
+}
+
+TEST(SecretsManagerPlain, NonExistantPath) {
+  SecretsManagerPlain sm;
+  auto res = sm.decryptFile("/path/which/doesnt/exist");
+  ASSERT_FALSE(res.has_value());
 }
```
glevkovich, line 138: can we add here EOL? :)

glevkovich (approved):
Thanks for the fix :)

secretsmanager/test/secrets_manager_test.cpp
tdimitrov: You mean new line for the file?

PR #1336, opened 2021-03-30
Fix undefined behavior due to up casting. (#1328)

HristoStaykov:
* Fix undefined behavior due to up casting.

Everywhere we deserialize messages inheriting MessageBase we use
MessageBase::deserializeMsg, which returns a MessageBase pointer.
Up-casting it to the actual Message type is not sufficient, we need
to actually create the proper object and transfer the packed struct
containing the header and dynamic payload to it.

* Remove unnecessary C-cast.

* Take in consideration that MessageBase::deserializeMsg might return nullptr.

If the storage is empty we might get a nullptr for a given message.
In this case we must not construct anything and simply set to nullptr.


yuliasherman (approved):
LGTM

f-squirrel (approved):
Looks great, thank you @HristoStaykov for fixing it!

PR #1338, opened 2021-03-30
Docker Build: Expose all CMAKE flags to CLI build

andrewjstone:
Every CMAKE flag is now directly editable for commandline builds and can
be used in the following manner:

`CONCORD_BFT_CMAKE_USE_LOG4CPP=FALSE make`

To make it easier to configure transport, a conditional was added such
that it can be explicitly enabled by string type:

`CONCORD_BFT_CMAKE_TRANSPORT=UDP make`

The README was also updated to reflect the latest build recommendations.

Lastly, CMakeLists.txt was updated to make `BUILD_ROCKSDB_STORAGE` a default, since
it's required for KVBC now. In the future we can probably flip this
back as we allow building each library individually and enable the user
to not build KVBC at all.


f-squirrel (approved):

PR #1340, opened 2021-03-31
Write a block on every write reconfiguration request

yontyon:
We add a new dedicated reconfiguration handler that writes a block for every new write reconfiguration request

### DIFF
--- a/kvbc/include/reconfiguration_add_block_handler.hpp
+++ b/kvbc/include/reconfiguration_add_block_handler.hpp
@@ -16,6 +16,7 @@
 #include "db_interfaces.h"
 #include "hex_tools.h"
 #include "block_metadata.hpp"
+#include "kvbc_key_types.hpp"
 
 namespace concord::kvbc::reconfiguration {
 class ReconfigurationHandler : public concord::reconfiguration::IReconfigurationHandler {
@@ -25,7 +26,8 @@ class ReconfigurationHandler : public concord::reconfiguration::IReconfiguration
   bool handle(const concord::messages::WedgeCommand& command, concord::messages::ReconfigurationErrorMsg&) override {
     std::vector<uint8_t> serialized_command;
     concord::messages::serialize(serialized_command, command);
-    auto blockId = persistReconfigurationBlock(serialized_command, command.bft_seq_num, reconfiguration_key_prefix + 1);
+    auto blockId =
+        persistReconfigurationBlock(serialized_command, command.bft_seq_num, kvbc::keyTypes::reconfiguration_wedge_key);
     LOG_INFO(getLogger(), "WedgeCommand block is " << blockId);
     return true;
   }
@@ -42,7 +44,8 @@ class ReconfigurationHandler : public concord::reconfiguration::IReconfiguration
   bool handle(const concord::messages::DownloadCommand& command, concord::messages::ReconfigurationErrorMsg&) override {
     std::vector<uint8_t> serialized_command;
     concord::messages::serialize(serialized_command, command);
-    auto blockId = persistReconfigurationBlock(serialized_command, command.bft_seq_num, reconfiguration_key_prefix + 2);
+    auto blockId = persistReconfigurationBlock(
+        serialized_command, command.bft_seq_num, kvbc::keyTypes::reconfiguration_download_key);
     LOG_INFO(getLogger(), "DownloadCommand command block is " << blockId);
     return true;
   }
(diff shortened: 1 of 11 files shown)


dartdart26 (commented):
Thanks @yontyon, code looks good! I was looking at an issue that is somewhat related and I am concerned about two things:

1. TimeContract operates under the assumption that all blocks have the summarized time key. If it isn't there, we assume epoch. That, however, will break another assumption for TimeContract - time only goes forward.
2. When we use block accumulation, we prefix events with the block ID. If we have pre-execution mixed in with Pruning and now any reconfiguration, we will add some blocks without updating the accumulated block ID in memory.

I would say let's have an offline discussion and find a solution for both issues above. I think we should have a central place for adding blocks. Maybe we get the updates from all submodules (reconf, time, etc.) and then add them at once.

dartdart26:
> Thanks @yontyon, code looks good! I was looking at an issue that is somewhat related and I am concerned about two things:
> 
> 1. TimeContract operates under the assumption that all blocks have the summarized time key. If it isn't there, we assume epoch. That, however, will break another assumption for TimeContract - time only goes forward.
> 2. When we use block accumulation, we prefix events with the block ID. If we have pre-execution mixed in with Pruning and now any reconfiguration, we will add some blocks without updating the accumulated block ID in memory.
> 
> I would say let's have an offline discussion and find a solution for both issues above. I think we should have a central place for adding blocks. Maybe we get the updates from all submodules (reconf, time, etc.) and then add them at once.

Had an offline discussion with @yontyon and the conclusions are:

1. We will do a linear search backwards in the Pruning State Machine if a block reached via `std::lower_bound()` doesn't have the summarized time key until we find a block with that key and use its time. We don't expect this to happen often and we don't expect to have to search many blocks.
2. We came to a conclusion that it is not an issue as nobody reads events, we just strip the block ID prefix. A minor thing that we need to consider, though, is `AccumulatedBlock::find()` that is used for conflict detection. It seems it doesn't need to lookup the immutable events map. Will be resolved in a separate commit.

Thank you so much for the clarifications and considering all these implications @yontyon!

@cloudnoize FYI

kvbc/include/reconfiguration_add_block_handler.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+namespace concord::kvbc::reconfiguration {
+class ReconfigurationHandler : public concord::reconfiguration::IReconfigurationHandler {
+ public:
+  ReconfigurationHandler(kvbc::IBlockAdder& block_adder, kvbc::IReader& ro_storage)
+      : blocks_adder_{block_adder}, block_metadata_{ro_storage} {}
+  bool handle(const concord::messages::WedgeCommand& command, concord::messages::ReconfigurationErrorMsg&) override {
+    std::vector<uint8_t> serialized_command;
+    concord::messages::serialize(serialized_command, command);
+    auto blockId = persistReconfigurationBlock(serialized_command, command.bft_seq_num, reconfiguration_key_prefix + 1);
```
dartdart26, line 28: I think it will be easier to reason about and also when refactoring/editing if we use constants for the whole key per command.

kvbc/include/reconfiguration_add_block_handler.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+    // All blocks are expected to have the BFT sequence number as a key.
+    ver_updates.addUpdate(std::string{bft_seq_num_key_}, block_metadata_.serialize(bft_seq_num));
+
+    concord::kvbc::categorization::Updates updates;
+    updates.add(kvbc::kConcordInternalCategoryId, std::move(ver_updates));
+    try {
+      return blocks_adder_.add(std::move(updates));
+    } catch (...) {
+      throw std::runtime_error{"Reconfiguration Handler failed to persist last agreed prunable block ID"};
```
dartdart26, line 94: Nit: I think that changing the original exception might hide information about the real issue. Maybe we can log that the reconfiguration handler failed and and just call re-throw as `throw;` .

kvbc/include/reconfiguration_add_block_handler.hpp
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  logging::Logger getLogger() {
```
dartdart26, line 98: This method can be either `const` or `static`.

tests/apollo/util/concord_msgs.py
```diff
@@ -4,7 +4,7 @@
 
 # Concord
 #
-# Copyright (c) 2020 VMware, Inc. All Rights Reserved.
```
dartdart26, line 7: Just curious - is there a way to avoid this file altogether?

dartdart26 (approved):
Thanks @yontyon, looks good! I have added a few minor comments. I was also not sure about the `concord_msgs.py` file and how it is being used and generated.

kvbc/include/reconfiguration_add_block_handler.hpp
yontyon: done

kvbc/include/reconfiguration_add_block_handler.hpp
yontyon: done

tests/apollo/util/concord_msgs.py
yontyon: I removed the file and added it to .gitignore

PR #1341, opened 2021-03-31
Update encryption flow for SecretsManager

tdimitrov:
Change the encryption/decryption flow in SecretsManagerEnc

SecretsManagerEnc uses new flow for encryption and decryption. The
reason for the change is that the old flow used a password for key
generation, which is not as secure as generating a random 256bit secret
key. The new flow addresses this by removing the depricated password
based secret key generation algorithm from OpenSSL.

The new encryption flow:
1. SecretData are the input parameters for SecretsManagerEnc. They
   contain an algorithm name, symmetric key and IV.
2. The key and iv are used to encrypt the payload with a supported
   block cipher (AES256 CBC).
3. The buffer is base64 encoded and written to file.

The decryption flow is the same as encyption but in reverse order.

@glevkovich Could you please check the modifications I made on bft_client_api_tests.cpp? (1382c6b3cbf71b2b800973dee8ef861d4002d732)
@KTeofanidis Could you please share if the new struct SecretsData works for you?


secretsmanager/include/secret_data.h
```diff
@@ -20,10 +20,12 @@ namespace concord::secretsmanager {
 
 // This struct contains the secret parameters needed to initialise the SecretsManagerEnc
 struct SecretData {
-  std::string password;  // Secret used for key generation
-  std::string digest;    // Digest algorithm used for key generaetion
-  std::string algo;      // Encryption algorithm used by SecretsManagerEnc
-  uint32_t key_length;   // The size of the key in bytes
+  std::string key;      // symmetric key
```
andrewjstone, line 23: Nit: Maybe mention that `key` and `iv` are hex strings in comments as opposed to something like Base64.

secretsmanager/include/secret_data.h
```diff
@@ -20,10 +20,12 @@ namespace concord::secretsmanager {
...
+  std::string iv;       // iv to use with the key
+  std::string algo;     // Encryption algorithm used by SecretsManagerEnc
+  uint32_t key_length;  // The size of the key in bytes
+  std::string
```
andrewjstone, line 27: Nit: This looks weird. Maybe put the comment above the field since it's so long.

secretsmanager/include/secrets_manager_enc.h
```diff
@@ -27,34 +27,31 @@
...
-  std::string algo_;
-  uint32_t key_length_;
-
   const std::set<std::string> supported_encs_{"AES/CBC/PKCS5Padding", "AES/CBC/PKCS7Padding"};
-  const std::set<std::string> supported_digests_{"SHA-256"};
-  const uint32_t SALT_SIZE = 8;
 
-  logging::Logger logger = logging::getLogger("concord.bft.secrets-manager-enc");
-  CryptoPP::BlockingRng rand;
+  logging::Logger logger_ = logging::getLogger("concord.bft.secrets-manager-enc");
```
andrewjstone, line 41: Nit: The files in our code base, by convention,  almost always put the private data fields below the public methods with an explicit `private:` section.

secretsmanager/src/base64.cpp
```diff
@@ -20,50 +20,21 @@
...
 }
 
-SaltedCipher base64Dec(const std::string& input) {
-  std::vector<uint8_t> decoded;
-  CryptoPP::StringSource ss(input, true, new CryptoPP::Base64Decoder(new CryptoPP::VectorSink(decoded)));
-
-  SaltedCipher result;
+std::vector<uint8_t> base64Dec(const std::string& input) {
+  std::vector<uint8_t> dec;
+  CryptoPP::StringSource ss(input, true, new CryptoPP::Base64Decoder(new CryptoPP::VectorSink(dec)));
```
andrewjstone, line 36: How is the decoder destroyed after the explicit new?
Does `StringSource` take ownership of the `Base64Decoder?`?

secretsmanager/src/key_params.cpp
```diff
@@ -0,0 +1,12 @@
+#include "key_params.h"
+
+#include <cryptopp/cryptlib.h>
+#include <cryptopp/filters.h>
+#include <cryptopp/hex.h>
+
+namespace concord::secretsmanager {
+KeyParams::KeyParams(const std::string& pkey, const std::string& piv) {
+  CryptoPP::StringSource sskey(pkey, true, new CryptoPP::HexDecoder(new CryptoPP::VectorSink(key)));
```
andrewjstone, line 9: Same question about destruction as above.

andrewjstone (approved):
Less code, more secure! Looks great @tdimitrov!. 

Mostly just nits from me. Please ensure we aren't actually leaking data with Cryptopp where I pointed out also.

One last question: Is an IV typically generated with a key? It may be worth running a test to generate them or at least provide a comment with instructions in the existing tests about how the key and iv were generated.

secretsmanager/src/base64.cpp
tdimitrov: This is one of the weirdnesses of crypto++. Their [README](https://github.com/weidai11/cryptopp/blob/b72007386d0fdb817d95a299f7368646ffd50afd/Readme.txt#L105) states this:

> 1. If a constructor for A takes a pointer to an object B (except primitive
> types such as int and char), then A owns B and will delete B at A's
> destruction.  If a constructor for A takes a reference to an object B,
> then the caller retains ownership of B and should not destroy it until
> A no longer needs it.

So yes - StringSource takes ownership and releases it. It was confusing for me too.

secretsmanager/src/key_params.cpp
tdimitrov: Ditto

tdimitrov:
> Less code, more secure! Looks great @tdimitrov!.

Thanks Andrew.

> One last question: Is an IV typically generated with a key? It may be worth running a test to generate them or at least provide a comment with instructions in the existing tests about how the key and iv were generated.

IV might or might not be generated with a key. You can use one key with different IVs for security reasons.

In our case I used /dev/random to generate them :
```
dd if=/dev/random of=./key bs=1 count=32 bs=1
dd if=/dev/random of=./iv bs=1 count=16 bs=1
```
I'll put this as a comment in the unit test.

andrewjstone:
> > Less code, more secure! Looks great @tdimitrov!.
> 
> Thanks Andrew.
> 
> > One last question: Is an IV typically generated with a key? It may be worth running a test to generate them or at least provide a comment with instructions in the existing tests about how the key and iv were generated.
> 
> IV might or might not be generated with a key. You can use one key with different IVs for security reasons.
> 
> In our case I used /dev/random to generate them :
> 
> ```
> dd if=/dev/random of=./key bs=1 count=32 bs=1
> dd if=/dev/random of=./iv bs=1 count=16 bs=1
> ```
> 
> I'll put this as a comment in the unit test.

Great. Thanks for the info!

secretsmanager/src/base64.cpp
andrewjstone: Thanks for the info !

PR #1342, opened 2021-03-31
Fix for deletion of messages that are being used.

HristoStaykov:
In case we recover from descriptorOfLastExitFromView we pass
the messages inside it to ViewsManager. Those messages currently
are raw pointers and we cannot delete them in the end of loadViewInfo()
We only delete them if we have not used them - the case in which we
use descriptorOfLastNewView.


andrewjstone (approved):
LGTM

HristoStaykov:
> Thanks @HristoStaykov !
> 
> Just for my understanding - basically, for all code paths in which we call `p->getAndAllocateDescriptorOfLastExitFromView()` to allocate elements in `descriptorOfLastExitFromView` and actually use the `.elements` field, we should not call `clean()`. And for cases where we don't use this field, we call `clean()` on line 180.

Yes, those are raw pointers that ViewsManager does not copy, the same holds for descriptorOfLastExitFromView.myViewChangeMsg. We need to make those std::shared_ptr-s in the long turn. Passing raw pointers and tracking ownership this way is error prone.

PR #1344, opened 2021-04-01
ReplicaConfig.hpp: remove const from publicKeysOfClients

glevkovich:
This const creates issues when trying to emplace or insert items to a
map.
Instead of adding copy/move functions, I prefer to remove it.


HristoStaykov (approved):
LGTM, thanks @glevkovich .

PR #1345, opened 2021-04-01
RocksDB: Use OptimisticTransactionDB

andrewjstone:
Our metadata storage makes use of RocksDB transactions. Unfortunately,
by default RocksDB transactions have a timeout of 1s. In *very* rare
cases we hit this timeout and an exception is thrown. However, the
calling code using transactions does not take this exception into
account, and it winds up being caught at a higher level triggering a
call to `std::terminate`. It would be a relatively significant change at
this point to the behavior to try to catch these timeouts, especially
when we are considering writing a new implementation of state transfer
anyway.

As we don't expect any conflicts in these transactions, and the crash is
occurring during state collection, which is basically single threaded,
the easiest change we can make is to use an `OptimisticTransactionDB`
instead of a `TransactionDB`. An `OptimisticTransactionDB` does conflict
detection checking at commit time and never takes locks, thus
sidestepping the problem of lock timeouts altogether. Additionally, it
likely better suits this specific use case and may actually be faster.

### DIFF
--- a/storage/include/rocksdb/client.h
+++ b/storage/include/rocksdb/client.h
@@ -20,7 +20,7 @@
 #ifdef USE_ROCKSDB
 #include "Logger.hpp"
 #include <rocksdb/db.h>
-#include <rocksdb/utilities/transaction_db.h>
+#include <rocksdb/utilities/optimistic_transaction_db.h>
 #include <rocksdb/sst_file_manager.h>
 #include "storage/db_interface.h"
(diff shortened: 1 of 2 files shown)


andrewjstone:
We've done a long running test and see no perf degradation. We also have not seen a reoccurrence of the lock issue. System test approves of this commit.

robem (approved):

PR #1350, opened 2021-04-04
BFT client return sequence number

Liorgal28:
When working with PerfTool, we need to match the reply with the sequence number, so instead of just returning a reply, we need to return a map.

### DIFF
--- a/bftclient/include/bftclient/bft_client.h
+++ b/bftclient/include/bftclient/bft_client.h
@@ -28,6 +28,8 @@
 
 namespace bft::client {
 
+typedef std::unordered_map<uint64_t, Reply> SeqNumToReplyMap;
+
 class Client {
  public:
   Client(std::unique_ptr<bft::communication::ICommunication> comm, const ClientConfig& config);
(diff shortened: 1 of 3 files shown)


bftclient/include/bftclient/bft_client.h
```diff
@@ -44,7 +44,7 @@ class Client {
   // Throws a BftClientException on error.
   Reply send(const WriteConfig& config, Msg&& request);
   Reply send(const ReadConfig& config, Msg&& request);
-  std::deque<Reply> sendBatch(std::deque<WriteRequest>& write_requests, const std::string& cid);
+  std::unordered_map<uint64_t, Reply> sendBatch(std::deque<WriteRequest>& write_requests, const std::string& cid);
```
yuliasherman, line 47: std::unordered_map<uint64_t, Reply> is too long, create define for it.

yuliasherman, line 47: unordered_map does not have a specific order of elements. Will the client pool know to match replies to requests?

Liorgal28: Yes, before we send the batch, we create a dictionary between the sequence number and the CID.

Liorgal28: Added typedef called ReplySeqNumMap

bftclient/include/bftclient/bft_client.h
```diff
@@ -28,6 +28,8 @@
 
 namespace bft::client {
 
+typedef std::unordered_map<uint64_t, Reply> ReplySeqNumMap;
```
yuliasherman, line 31: ReplySeqNumMap -> SeqNumToReplyMap

Liorgal28: Done

yuliasherman (approved):
LGTM

PR #1351, opened 2021-04-05
Add pruning handler to read only replica

yontyon:
When pruning, we need to verify that read-only replicas already have all blocks until the pruning point. For this, we add a pruning handler to the read-only replica.
This pruning handler knows to handler only latestPruneableBlock request and it returns the lastReachableBlock of the read-only replica


bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -83,6 +88,10 @@ void ReadOnlyReplica::sendAskForCheckpointMsg() {
 
 template <>
 void ReadOnlyReplica::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
+  if (isCollectingState()) {
```
dartdart26, line 91: Could you please explain the rationale for this change? I assume it is not related to this PR, but I might be wrong here.

bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -128,4 +137,81 @@ void ReadOnlyReplica::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
...
+  const NodeIdType senderId = m->senderId();
+  const NodeIdType clientId = m->clientProxyId();
+  const bool reconfig_flag = (m->flags() & MsgFlag::RECONFIG_FLAG) != 0;
+  const ReqId reqSeqNum = m->requestSeqNum();
+  const uint8_t flags = m->flags();
+
+  SCOPED_MDC_CID(m->getCid());
+  LOG_DEBUG(MSGS, KVLOG(clientId, reqSeqNum, senderId) << " flags: " << std::bitset<8>(flags));
+
+  const auto &span_context = m->spanContext<std::remove_pointer<decltype(m)>::type>();
```
dartdart26, line 151: Nit: Since we are in the `ClientRequestMsg` handler anyway, maybe it would be simpler to just do `spanContext<ClientRequestMsg>()`. Just a thought.

bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -128,4 +137,81 @@ void ReadOnlyReplica::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
...
+  span.setTag("rid", config_.getreplicaId());
+  span.setTag("cid", m->getCid());
+  span.setTag("seq_num", reqSeqNum);
+
+  // A read only replica can handle only reconfiguration requests. Those requests are signed by the operator and
+  // the validation is done in the reconfiguration engine. Thus, we don't need to check the client validity as in
+  // the committers
+
+  if (reconfig_flag) {
+    LOG_INFO(GL, "ro replica has received a reconfiguration reqeust");
```
dartdart26, line 162: typo: reqeust -> request

bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -128,4 +137,81 @@ void ReadOnlyReplica::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
...
+    executeReadOnlyRequest(span, m);
+    delete m;
+    return;
+  }
+
+  delete m;
+}
+
+void ReadOnlyReplica::executeReadOnlyRequest(concordUtils::SpanWrapper &parent_span, ClientRequestMsg *request) {
```
dartdart26, line 171: Nit: I think we can use a `const ClientRequestMsg&` type as we know it can't be nullptr and we don't change it. Speaking of which, we might also change:

```
char* ClientRequestMsg::requestBuf() const;
```
to return `const char*`.

bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -128,4 +137,81 @@ void ReadOnlyReplica::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
...
+  auto span = concordUtils::startChildSpan("bft_execute_read_only_request", parent_span);
+  // Read only replica does not know who is the primary, so it always return 0. It is the client responsibility to treat
+  // the replies accordingly.
+  ClientReplyMsg reply(0, request->requestSeqNum(), config_.getreplicaId());
+
+  uint16_t clientId = request->clientProxyId();
```
dartdart26, line 177: Nit: can be made `const`.

bftengine/src/bftengine/ReadOnlyReplica.hpp
```diff
@@ -41,7 +42,7 @@ class ReadOnlyReplica : public ReplicaForStateTransfer {
 
   template <typename T>
   void messageHandler(MessageBase* msg) {
-    if (validateMessage(msg) && !isCollectingState())
```
dartdart26, line 44: Could you please explain that change (as above)?

kvbc/include/pruning_handler.hpp
```diff
@@ -189,4 +189,47 @@ class PruningHandler : public reconfiguration::IPruningHandler {
   mutable std::mutex pruning_status_lock_;
   mutable std::future<void> async_pruning_res_;
 };
+
+/*
+ * The read only pruning handler knows to answer only getLatestPruneableBlock.
+ * As it doesn't participant the consensus, the only thing we want to know is what is the latest reachable block.
```
dartdart26, line 195: Nit: Maybe we want to say "... what the latest **prunable** block is".

kvbc/include/pruning_handler.hpp
```diff
@@ -189,4 +189,47 @@ class PruningHandler : public reconfiguration::IPruningHandler {
...
+ public:
+  ReadOnlyReplicaPruningHandler(IReader &ro_storage)
+      : ro_storage_{ro_storage},
+        signer_{bftEngine::ReplicaConfig::instance().replicaPrivateKey},
+        pruning_enabled_{bftEngine::ReplicaConfig::instance().pruningEnabled_},
+        replica_id_{bftEngine::ReplicaConfig::instance().replicaId} {}
+  bool handle(const concord::messages::LatestPrunableBlockRequest &,
+              concord::messages::LatestPrunableBlock &latest_prunable_block,
+              concord::messages::ReconfigurationErrorMsg &) override {
+    if (!pruning_enabled_) return true;
```
dartdart26, line 207: I guess this might not be related to this PR, but I was wondering if we should return 0 as the last reachable block in case pruning is disabled. Does it make sense?

dartdart26: Actually, now that I think of it, maybe a read-only replica doesn't even care if pruning is enabled. What do you think?

util/pyclient/bft_client.py
```diff
@@ -284,7 +284,7 @@ def _process_received_msg(self, data, sender, replicas_addr, required_replies, c
             self.replies_manager.add_reply(rsi_msg)
             if self.replies_manager.has_quorum_on_all(required_replies):
                 self.replies = self.replies_manager.get_all_replies()
-                self.rsi_replies = self.replies_manager.get_rsi_replies(rsi_msg.get_matched_reply_key())
```
dartdart26, line 287: Could you please explain the Apollo RSI changes?

kvbc/include/pruning_handler.hpp
```diff
@@ -189,4 +189,47 @@ class PruningHandler : public reconfiguration::IPruningHandler {
...
+ * As it doesn't participant the consensus, the only thing we want to know is what is the latest reachable block.
+ */
+class ReadOnlyReplicaPruningHandler : public reconfiguration::IPruningHandler {
+ public:
+  ReadOnlyReplicaPruningHandler(IReader &ro_storage)
+      : ro_storage_{ro_storage},
+        signer_{bftEngine::ReplicaConfig::instance().replicaPrivateKey},
+        pruning_enabled_{bftEngine::ReplicaConfig::instance().pruningEnabled_},
+        replica_id_{bftEngine::ReplicaConfig::instance().replicaId} {}
+  bool handle(const concord::messages::LatestPrunableBlockRequest &,
```
dartdart26, line 204: Now the reconfiguration engine needs to verify the `LatestPrunableBlock` message from the read-only replica too. We need to change the `PublicKeys()` function in the pruning handler so that it includes read-only replicas, correct?

dartdart26:
Looks great, thanks @yontyon! I have added a few minor comments and asked a couple of questions. In general, I think it is good to go.

bftengine/src/bftengine/ReadOnlyReplica.cpp
yontyon: Previously we prevented any message from being processed while ST. now we want to allow client request and deny any other message (Checkpoint message)

bftengine/src/bftengine/ReadOnlyReplica.hpp
yontyon: We want to allow client requests (because the replica is mostly on ST and we want to allow the latestPruneableBlock request even then), but still to deny checkpoint messages

kvbc/include/pruning_handler.hpp
yontyon: We want not to touch anything in case pruning is disabled, because there might be other handlers that will be called (currently it is only concord's pruning handler, we want only one pruning handler (concord's) and for that we simply deactivate the internal command handler)

util/pyclient/bft_client.py
yontyon: Currently, the common data contain also the primary. However, we want to enable read requests to read only replicas, and ro replica doesn't know who the primary is

kvbc/include/pruning_handler.hpp
yontyon: yes

bftengine/src/bftengine/ReadOnlyReplica.cpp
dartdart26: Thanks!

kvbc/include/pruning_handler.hpp
dartdart26: I see, thanks. I was thinking that we might want to always return the last block for read-only replica, irrespective of `pruning_enabled_`.

Also, wanted to ask something else. What will the value be for `latest_prunable_block` if `pruning_enabled_ = false`?

util/pyclient/bft_client.py
dartdart26: Does the primary now become part of RSI?

dartdart26 (approved):
Thanks for the explanations @yontyon!

util/pyclient/bft_client.py
andrewjstone: I'm not sure this change is correct. The knowledge of the primary is required for normal operation of the client. The client should only send to the primary for write operations, unless there is a timeout. Sending to all replicas should only be done on timeout to ensure a primary is not malicious, otherwise we are sending more traffic than necessary. It shouldn't induce view change, but it's still different from the behavior of the real C++ client, and doesn't seem right to me. 

Excluding the primary is probably fine for read-only operations, but not write operations.

yontyon: @andrewjstone, the replicas still send the write requests to the primary only. This is just an implementation change. We collect the results from the replicas for a specific sequence number under a common key (as the reply may contain RSI). This change means that the primary number is not part of this common data, but it is still part of the reply itself, once the client decides the replies can be delivered up, it delivers also the primary ID.
We removed it from the common data, because the RORs don't have the knowledge on who the primary is, and we would like to collect their responses together with the committers reply.

kvbc/include/pruning_handler.hpp
yontyon: The point is that we may want to return something else from another pruning handler. So if the basic pruning (or any other pruning handler) is not enabled, we want not to touch the latest pruneable block as we may override a previous value of it. The value will be the one that set to the reference by either another pruning handler or by the dispatcher.

util/pyclient/bft_client.py
yontyon: @andrewjstone in any case, I revreted this change. I will perform vast changes once I will merge the Apollo tests for this

PR #1354, opened 2021-04-05
Docker update to version 0.20

robem:
This change updates boost to 1.75 and installs all boost libraries.
The main motivation is to use Boost.Beast (introduced in 1.66) at some point.
However, we don't want to miss out on 3 years worth of improvements either.
See https://www.boost.org/users/history/


f-squirrel (approved):
LGTM!

robem:
There are some significant issues at runtime that needs more investigation. Closing for now.

PR #1355, opened 2021-04-06
WIP: Bft client and Tester Replica changes for transaction signing

bhavsi:


glevkovich:
Is it possible to have 2 separate commits?
Although we do a single merge from branch->master for technical purposes (workflow via github pull requests), it is still very beneficial to separate commits with appropriate comments. That way we can easily trace the changes.
Since the commits are modifying separate areas, I would separate to:
1) An Apollo infra commit
2) tester Replica commit.

I would also suggest adding a detailed commit message per each commit, explaining the main changes and the reason for the change. Currently, your single commit has only a header, without a commit message body.

tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -151,6 +164,20 @@ std::unique_ptr<TestSetup> TestSetup::ParseArgs(int argc, char** argv) {
...
+    if ( !principalsMapping.empty() ) {
+      if ( !txnSigningKeysPath.empty() ) {
+        replicaConfig.clientTransactionSigningEnabled = true;
+      }
+      else {
+        throw std::runtime_error("Expected param txn-signing-key-path to be set when principals-mapping is set");
+      }
+    }
+
+    if (replicaConfig.clientBatchingEnabled) {
```
glevkovich, line 177: I assumed you wanted to check replicaConfig.clientTransactionSigningEnabled?
Why not adding this 1 liner after line 170? (after fixes you make).

tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -243,4 +270,50 @@ std::unique_ptr<IStorageFactory> TestSetup::GetStorageFactory() {
   return std::make_unique<v2MerkleTree::RocksDBStorageFactory>(dbPath.str());
 }
 
+std::vector<std::string> getKeyDirectories(const std::string& keysRootPath)
```
glevkovich, line 273: nit: You don't check the strict folder names 1,2,...
This is fine for Apollo but not for production :)

tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -243,4 +270,50 @@ std::unique_ptr<IStorageFactory> TestSetup::GetStorageFactory() {
...
+  }
+  return result;
+}
+
+void setPublicKeysOfClients(const std::string& principalsMapping,
+                            const std::string& keysRootPath,std::map<const std::string,
+                            std::set<const uint16_t>>& publicKeysOfClients) {
+  std::vector<std::string> keysDirectories = getKeyDirectories(keysRootPath);
+  std::vector<std::string> clientIdChunks;
+  boost::split(clientIdChunks, principalsMapping, boost::is_any_of(","));
```
glevkovich, line 290: nice usage..

tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -243,4 +270,50 @@ std::unique_ptr<IStorageFactory> TestSetup::GetStorageFactory() {
...
+  for(int i = 0; i < clientIdChunks.size(); ++i) {
+    std::string keyFilePath = keysDirectories.at(i) + "/public_key.pem"; // Will be "../1/public_key.pem" etc.
+
+    SecretsManagerPlain smp;
+    std::optional<std::string> keyPlaintext;
+    keyPlaintext = smp.decryptFile(keyFilePath);
+    if (keyPlaintext.has_value()) {
+      // Each clientIdChunk would look like "11 12 13 14"
+      std::vector<std::string> clientIds;
+      boost::split(clientIds, clientIdChunks.at(i), boost::is_any_of(" "));
```
glevkovich, line 309: 2 comments:
1) the next code will work, why use a vector?
`
    std::set<std::string> clientIds;
    boost::split(clientIds, "11 12 13", boost::is_any_of(" "));
`

2) Definition of publicKeysOfClients:
`std::set<std::pair<const std::string, std::set<uint16_t>>> publicKeysOfClients;`

Sorry for the bad sync.
That means, you need to transform this set into another set of uint16_t

glevkovich:
I suggest performing a manual temprary test, that way you could have tracked down some of the small issues commented:
 add a printout on loadConfig (see ReplicaLoader) to print the size of publicKeysOfClients and check the parameter. Run an arbitrary Apollo test with transaction signing enabled.
That way we can be sure all keys are loaded.

tests/apollo/util/bft.py
```diff
@@ -410,6 +419,32 @@ def random_client(self):
...
+        def split(a, n):
+            """ Splits list a into n chunks """
+            k, m = divmod(len(a), n)
+            return [a[i * k + min(i, m):(i + 1) * k + min(i + 1, m)] for i in range(n)]
+
+        principals = ""
+        client_ids = self.clients.keys().sort()
+        client_ids_chunks = split(client_ids, NUM_PARTICIPANTS)
+        reserved_client_ids = self.reserved_clients.keys().sort()
+        reserved_client_ids_chunks = split(reserved_client_ids, NUM_PARTICIPANTS)
```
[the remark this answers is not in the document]
bhavsi: I think merging the two lists might make it messier to compute how to distribute the reserved clients. Right now we want to distribute them in a round robin fashion based on the number of participants (and not all in a single participant in the end). Merged list size check (that it is the sum of two lists that we merged) might seem redundant, no?

tests/simpleKVBC/TesterReplica/setup.cpp
bhavsi: I think boost::split expects vector input, so I am collecting it as a vector and then converting it into a set. 
No your sync was fine, it was a mistake on my end. The set needed `uint16_t` not string. I have corrected that.

tests/apollo/util/bft.py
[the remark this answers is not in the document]
glevkovich: Merging the list is done only for "sanity check" for the computation.
We want to check 2 things:
L = merge(client_ids_chunks, reserved_client_ids_chunks)
You check that all elements in L are distinct + size(L) == number of clients + reserved.
Just an  suggestion.

tests/simpleKVBC/TesterReplica/setup.cpp
glevkovich: boost::split works with other containers as well..
This code compiles and work:

std::set<std::string> clientIds; 
boost::split(clientIds, "11 12 13", boost::is_any_of(" "));

bhavsi: I see. However, it will result in a collection of string, which we will again have to convert into a set of `uint16_t`. So I am wondering that we just leave the vector as is, and add the string to int conversion in a loop. Here's the updated code:

```
      std::vector<std::string> clientIds;
      boost::split(clientIds, clientIdChunks.at(i), boost::is_any_of(" "));
      std::set<uint16_t> clientIdsSet;
      for (const std::string clientId : clientIds) {
        clientIdsSet.insert(std::stoi(clientId));
      }
```

glevkovich: Yeah, I guess it's fine that way :)

glevkovich:
@bhavsi, please pay attention to you pipeline. you have some typo error:
`
self.txn_signing_enabled = True if os.environ.ge('TXN_SIGNING_ENABLED', "").lower() == "true" else False
AttributeError: '_Environ' object has no attribute 'ge'
`

PR #1356, opened 2021-04-06
Fix assertion on !isFetching() in BCStateTran::getDigestOfCheckpoint()

toly-kournik:
This method is called during both fetching and states. So it should be removed.
For some reason, probably related to the transaction policy, which has recently changed,
this long sleeping bug has been recently revealed.


f-squirrel (approved):
LGTM!

PR #1357, opened 2021-04-06
Fix intermittent failure of SkvbcPersistenceTest.test_st_while_crashing_primary_with_vc

nkgeorgiev:
In the test we get the current view by looking at the 'lastAgreedView' metric from a random replica and then wait for n-f replicas to enter that view.
But if we are (un)lucky enough, the random replica is the one that was stopped before and if we are even more (un)lucky, it may not have updated its 'lastAgreedView' metric. So we try to wait for n-f replicas to enter the old view and timeout.

I have refactored the "get_current_view" to work as follows:
We get the lastAgreedView from every live replica.
We then see if there are 2f + 2c + 1 replicas with the same view.
If there is, then that is the current view.
Else we sleep for 1 sec and try again until timeout.

Testing: I have run the test locally more than 200 times with no failure

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -678,7 +678,7 @@ class BftTestNetwork:
                             for r in self.get_live_replicas():
                                 nursery.start_soon(get_view, r)
                         view = Counter(replica_views).most_common(1)[0]
-                        # wait for n-f = 2f+2c+1 replicas to be in the expected view
+                        # wait for n-f = 2f+2c+1 replicas to have agreed the expected view
                         if view[1] >= 2 * self.config.f + 2 * self.config.c + 1:
                             matching_view = view[0]
                             nb_replicas_in_matching_view = view[1]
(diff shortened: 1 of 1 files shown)


HristoStaykov (approved):
LGTM, great finding and fix @nkgeorgiev, thanks!

tests/apollo/util/bft.py
```diff
@@ -662,11 +662,40 @@ async def get_current_view(self):
...
-            return current_view
+            matching_view = None
+            nb_replicas_in_matching_view = 0
+
+            async def get_view(replica_id):
+                def expected(_): return True
+                replica_view = await self._wait_for_matching_agreed_view(replica_id, expected)
+                replica_views.append(replica_view)
+            try:
+                while True:
```
KTeofanidis, line 673: I think we should have a timeout here.

nkgeorgiev: Fixed. Thanks for catching this.

KTeofanidis:
LGTM

KTeofanidis (approved):
LGTM!

PR #1359, opened 2021-04-07
Transaction signing - Load client public keys, refactor signature manager

glevkovich:
commits:

1. ReplicaConfig: change publicKeysOfClients into set
2. Crypto: refactor RSAVerifier and RSASigner
3. SigManager: Refactor, transform to singleton, and add functionality
4. gtests: fix relevant tests to adjust to recent SigManager changes
5. preprocessor_test: disable until re-designed
6. SigManager_test: Introduce a new gtest test for SigManager


bftclient/test/bft_client_api_tests.cpp
```diff
@@ -38,6 +38,7 @@ using namespace bftEngine::impl;
 using namespace bftEngine;
 using namespace placeholders;
 using namespace concord::secretsmanager;
+using ReplicaId_t = bft::client::ReplicaId;
```
f-squirrel, line 41: Why not just `using bft::client::ReplicaId`?

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -234,16 +234,18 @@ class RSAVerifier::Impl {
...
-
-RSASigner::RSASigner(const string& private_key_pem) {
-  StringSource ss(private_key_pem, true);
-  RSA::PrivateKey priv_key;
-  PEM_Load(ss, priv_key);
-  impl = std::make_unique<Impl>(priv_key);
+RSASigner::RSASigner(const char* privateKey, KeyFormat format) {
+  if (format == KeyFormat::HexaDecimalStrippedFormat) {
+    StringSource s(privateKey, true, new HexDecoder);
+    impl = std::make_unique<Impl>(s);
```
f-squirrel, line 240: I know that it is not a new code but could you please change `impl` to `impl_`?

f-squirrel (commented):
Thank you, @glevkovich.
What is the reasoning behind transforming sig manager to a singleton?
It has certain disadvantages from the unit tests perspective and since we already pass it everywhere, it seems more work to change it, does it not?

bftclient/test/bft_client_api_tests.cpp
glevkovich: Possible, but this structure repeats 16 times in the file.
My idea is that it might make the code less readable.

glevkovich:
There are few reasons to transform Sigmanager into a singleton.
1st, this is not the main reason usually, but there are multiple SigManagers before my change, which is not a good idea.
2nd, the current use of signature manager is somehow "patchy', without a well-defined API. SigManager pointers are transferred between modules to messages and modules to modules, static vars are defined, additional class functions which set and get locally. Soon, I need to add more functionality for transaction signing, which creates a bigger complications.
If you look at SigManager, besides initialization which is a bit complex in current implementation - it is just a registry / database which holds keys and maps them according to a principal id. 
So it makes sense to give it the same status as configManager -
1) we need only one
2) we call it from many places and want to avoid transferring this object around.

So I do know about claims saying that singleton is very bad - but here we don't misuse it. There are no public vars, only APi is public. 
The only complication I currently have is the need to re-design one test (which actually is not a unit test, but an attempt to be a "test apollo test" . All other tests adapt well to the new design.

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -192,6 +192,7 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
     serialize(outStream, batchingFactorCoefficient);
 
     serialize(outStream, publicKeysOfReplicas);
+    // serialize(outStream, publicKeysOfClients);
```
tdimitrov, line 195: Why publicKeysOfClients can't be serialized/deserialized?

bftclient/test/bft_client_api_tests.cpp
tdimitrov: Just curious - what changed so that we need to refer ReplicaId with the whole namespace>

tdimitrov:
I think converting SigManager to a singleton will bite us in the future. The positive side is that we get rid of the set methods, which are a bit error prone. But there are negatives:
1. Passing pointers via constructors may seem as tedious and too much work, but it creates a clear separation between the modules and makes the dependencies between them clear. Calling getInstance() and using the functionality is useful, but creates a hidden coupling between the modules. 
2. If someone decides to unit test any module which uses SigManager it will be very hard to mock for SigManager.
3. The instantiation of the SigManager is not trivial. How can we be sure there won't be a race condition between init() and getInstance() in the future?

My main concern is point 1. Points 2 and 3 are a bit artificial, but not impossible.

bftengine/include/bftengine/ReplicaConfig.hpp
glevkovich: Actually, This one is removed on my local branch but didn't push yet. I want to perform it in another task.
When uncommenting it just doesn't work.
The current serilize/deserialie functions do not handle a type:
std::set<std::pair<const std::string, std::set<uint16_t>>>

So I want to implement this in another task, since It will take at least a day or more to do, and from my conversations with people, it isn't a high priority task for now.

glevkovich:
`Passing pointers via constructors may seem as tedious and too much work, but it creates a clear separation between the modules and makes the dependencies between them clear. Calling getInstance() and using the functionality is useful, but creates a hidden coupling between the modules.`

In this case, I do not see the coupling.
SigManager is stateless, you. cannot change it after initializing it.
It only acts as a registry of a private key for signing, and a group of verifiers (public keys), each one mapped to 1 more principal ID which sign with a matching private key.
I don't see a big difference between SigManager and Configuration Manager.

glevkovich:
`If someone decides to unit test any module which uses SigManager it will be very hard to mock for SigManager.`

I have successfully converted with a helper function  I wrote in helper.cpp 3 out of 4 tests and added one test of my own. It wasn't very hard after writing the helper function.
One test is problematic since it is not defined to be a real unittest. I have opened a ticket for it and disabled it, and will refactor it seperately. It tries to simulate multiple SigManager, which in my opinion out of scope for unittests - this should be checked in Apollo. In my new tests, I once checked SigManager as a signer, and once as verifier. To simulate signers and verifiers without the need for SigManager, once can just use a Bare array of RSASigners/Verifiers.

The bottom line - I do tend to be flexible on function/class declaration to allow easier unit testing, but I don't think it should be the main design considiration.

glevkovich:
`The instantiation of the SigManager is not trivial. How can we be sure there won't be a race condition between init() and getInstance() in the future?

No one can call getInstance before it is created since it is created in createNewReplica. preprocessor and view change modules are created after.  I guess it's also a matter of testing and unit-testing.
If issues raise, It would be possible to create the SigManager much earlier, even on replica loader.
`

glevkovich:
Comment abut Thread safety:

From CryptoPP main Readme:
"Crypto++ is thread-safe at the class level. This means you can use
Crypto++ safely in a multithreaded application, but you must provide
synchronization when multiple threads access a common Crypto++ object."

Preprocessor does not protect objects, and luckily do not have data races, or at least non that we are aware about.
The preprocessor didn’t have data races (or maybe it is very rare) probably since  1:1 relation between replicas, and the fact that at any given time 1 thread is using RSAVerifier / RSASigner.
I don’t think the same case will hold when handling transaction signing from multiple clients, since in some cases the same verifier is used (I do not want to hold 1000-2000 RSAVerifiers objects, 1 per bft-client).
Another possibility is that we do have races in Preprocessor which we've never observed.

I am now researching whether I'll have to add lock guards all over the RSAVerifiers. 
This will be part of another PR, anyway, since I prefer to test 1st.

tdimitrov (approved):

f-squirrel (approved):

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
```diff
@@ -90,7 +90,8 @@ void PreProcessReplyMsg::setupMsgBody(const char* preProcessResultBuf,
     }
     {
       concord::diagnostics::TimeRecorder scoped_timer(*preProcessorHistograms_->signHash);
-      sigManager_->sign((char*)hash.data(), SHA3_256::SIZE_IN_BYTES, body() + headerSize, sigSize);
+      auto sigManager = SigManager::getInstance();
```
yuliasherman, line 93: Why take sigManager again?  It has been already taken at line 82. 
Or you can use it directly, like: sigSize = SigManager::getInstance()->getMySigLength();

bftengine/src/preprocessor/messages/PreProcessReplyMsg.hpp
```diff
@@ -77,10 +75,8 @@ class PreProcessReplyMsg : public MessageBase {
       NodeIdType senderId, uint16_t clientId, uint16_t reqOffsetInBatch, ReqId reqSeqNum, uint64_t reqRetryId);
   Header* msgBody() const { return ((Header*)msgBody_); }
 
- private:
```
yuliasherman, line 80: I prefer to have separate 'private', 'public', etc. sections for functions and data members.

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
glevkovich: That's redundant. Thanks for seeing that.
Removed.

bftengine/src/preprocessor/messages/PreProcessReplyMsg.hpp
glevkovich: ok. will return.

PR #1361, opened 2021-04-07
Makefile: add test-range to run a range of tests

glevkovich:
1) To get test numbers to use 'make list-tests'. This gives for each test
an index number;
2) This target allows running tests in a range, instead of running all
tests (e.g 'make test')).


f-squirrel (approved):
LGTM!

andrewjstone:
Cool. Thanks @glevkovich

PR #1363, opened 2021-04-07
Add Key-Exchange Status for concord-ctl

saiakhil2012:
Add functionality of getting Key-Exchange status from `concord-ctl`


teoparvanov (approved):
LGTM, thanks Akhil!

PR #1365, opened 2021-04-08
Extending fake_comm for unit testing of BFT client pool

irisk29:
The delayed behavior is used for testing the overload case in the BFT client pool unit tests.

Workflow:

1. Created an additional type of behavior - delayed behavior.
2. Extracted common functionality from immediate behavior - to replayConfig function.
3. In the delayed behavior function, I created the delay in the response of the client using the sleep method for 100ms.
4. In the BFT client pool unit test, using a static field(which was used as a flag), we can know when to use the delayed and immediate behavior accordingly.

### DIFF
--- a/bftclient/include/bftclient/fake_comm.h
+++ b/bftclient/include/bftclient/fake_comm.h
@@ -133,8 +133,20 @@ class FakeCommunication : public bft::communication::ICommunication {
   BehaviorThreadRunner runner_;
   std::thread fakeCommThread_;
 };
+std::vector<uint8_t> replayConfig(const MsgFromClient &msg);
 
 inline void immideateBehaviour(const MsgFromClient& msg, IReceiver* client_receiver) {
+  std::vector<uint8_t> reply = replayConfig(msg);
+  client_receiver->onNewMessage(msg.destination.val, reinterpret_cast<const char*>(reply.data()), reply.size());
+}
+
+inline void delayedBehaviour(const MsgFromClient& msg, IReceiver* client_receiver) {
+  std::vector<uint8_t> reply = replayConfig(msg);
+  std::this_thread::sleep_for(100ms);
+  client_receiver->onNewMessage(msg.destination.val, reinterpret_cast<const char*>(reply.data()), reply.size());
+}
+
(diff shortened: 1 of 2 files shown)


bftclient/include/bftclient/fake_comm.h
```diff
@@ -133,8 +133,20 @@ class FakeCommunication : public bft::communication::ICommunication {
   BehaviorThreadRunner runner_;
   std::thread fakeCommThread_;
 };
+std::vector<uint8_t> replayConfig(const MsgFromClient& msg);
```
andrewjstone, line 136: I think a better name for this would be `createReply`.

bftclient/include/bftclient/fake_comm.h
```diff
@@ -133,8 +133,20 @@ class FakeCommunication : public bft::communication::ICommunication {
...
+  client_receiver->onNewMessage(msg.destination.val, reinterpret_cast<const char*>(reply.data()), reply.size());
+}
+
+inline void delayedBehaviour(const MsgFromClient& msg, IReceiver* client_receiver) {
+  std::vector<uint8_t> reply = replayConfig(msg);
+  std::this_thread::sleep_for(100ms);
+  client_receiver->onNewMessage(msg.destination.val, reinterpret_cast<const char*>(reply.data()), reply.size());
+}
+
+std::vector<uint8_t> replayConfig(const MsgFromClient& msg) {
```
andrewjstone, line 149: This needs to be marked `inline` to prevent ODR violations. This will fix the clang-tidy error as well.

andrewjstone (approved):
Looks great @irisk29!

Please just make the requested fixes then, all good to merge.

bftclient/include/bftclient/fake_comm.h
irisk29: changed it. Thank you.

bftclient/include/bftclient/fake_comm.h
irisk29: Fixed it! Thank you!

Liorgal28 (approved):
Thanks @irisk29 nice changes!

PR #1366, opened 2021-04-08
Fix issues with new client batching

Liorgal28:
In this PR I'm fixing issues with new client batching.
1. pending_requests queue is holding all the requests that arrive in batch requests, before this PR I didn't reset the pending_requests that caused an issue in which we got more and more requests inside this queue.
2. After each wait() if we didn't get enough replies for the request and its still part of reply_certificates_ we should reset the replies in the matcher because if we don't in the next round we will ask all the other replicas for replies and we will get other RSI for the same request which causes an error.

A unit test has also been added to test several batch requests that are sent through the same client.

### DIFF
--- a/bftclient/src/bft_client.cpp
+++ b/bftclient/src/bft_client.cpp
@@ -247,7 +247,7 @@ SeqNumToReplyMap Client::sendBatch(std::deque<WriteRequest>& write_requests, con
     pending_requests_.clear();
     return replies;
   }
-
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):
Looks great. Thanks for walking me through the code and adding a test Lior! You found a bug in my Error message code which makes me happy :)

f-squirrel (approved):
Looks great, thank you @Liorgal28

PR #1369, opened 2021-04-11
On-demand key exchange

toly-kournik:
Change CryptoManager signers/verifiers access methods to receive sequence number as a parameter. No actual functionality change yet.
Rename KeyManager -> KeyExchangeManager.


yontyon (approved):
LGTM

PR #1370, opened 2021-04-11
DBDataStore Transaction

toly-kournik:
Copy on write inconsistency.
Updating the InMemoryDataStore took place immediately, while DBDataStore update occurred only on transaction commit.
Solution: call to InMemoryDataStore updates as callback functions to DataStore transactions commit(), which are invoked AFTER updates are committed to DB.


yontyon (approved):
LGTM! 
Thanks!

PR #1371, opened 2021-04-12
ReplicaStatusHandlers: register replica state (bug fix)

glevkovich:
On commit 66005bb1874a2ac8e1fe8c59be76c7a8c3a0e71c we have added a new
handler, but I forgot to register the old one withit's new name.
Here I rename the handler name:
replica_state -> replica-state
and register it in ReplicaStatusHandlers (forgot to do so).


HristoStaykov (approved):
LGTM, thanks @glevkovich !

PR #1372, opened 2021-04-12
State Transfer transaction inconsistency.

toly-kournik:
Set txn->isFetching(false) after calling to onTransferringComplete callbacks.


yontyon (approved):
This solves the problem we found.
I wonder if there might be other problems: 
Basically when the ST thread is working the main thread does not get mew messages, but is it possible that there are other race conditions? 
what about concord bft timers? if they still work in parallel to ST, does this race condition can create inconsistency in their flows?

PR #1373, opened 2021-04-12
State Transfer transaction inconsistency. (#1372)

toly-kournik:
Set txn->isFetching(false) after calling to onTransferringComplete callbacks.


yontyon (approved):
As always, LGTM

PR #1376, opened 2021-04-13
Fix memory leak in ClientsManager

f-squirrel:
Problem description:
The method `allocateReplyFromSavedOne` in case of failure returns `nullptr` and does not delete `ClientReplyMsg* const r`.

Fix:
In order to avoid such problems in the future, methods `allocateReplyFromSavedOne` and `allocateNewReplyMsgAndWriteToStorage` create and return unique pointer.


yuliasherman (approved):
LGFM, thanks for fixing this.

Liorgal28 (approved):

PR #1377, opened 2021-04-13
send empty client request on pre process time out

yontyon:
In order to avoid executing a request that has been timed out on pre-processing, we now send an empty request.
Correspondingly, empty requests will be ignored in the execution loop.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -4054,8 +4054,9 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
       continue;
     }
     ClientRequestMsg req((ClientRequestMsgHeader *)requestBody);
-    if (req.requestLength() == 0) continue;
-    if (req.flags() & EMPTY_CLIENT_FLAG) {
+    if (req.requestLength() == 0 || (req.flags() & EMPTY_CLIENT_FLAG)) {
+      if (clientsManager->isValidClient(req.clientProxyId()))
+        clientsManager->removePendingForExecutionRequest(req.clientProxyId(), req.requestSeqNum());
       continue;
     }
     SCOPED_MDC_CID(req.getCid());
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -4054,6 +4054,7 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
       continue;
     }
     ClientRequestMsg req((ClientRequestMsgHeader *)requestBody);
+    if (req.requestLength() == 0) continue;
     if (req.flags() & EMPTY_CLIENT_FLAG) {
       continue;
     }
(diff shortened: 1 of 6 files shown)


bftengine/src/preprocessor/messages/ClientPreProcessRequestMsg.cpp
```diff
@@ -28,12 +28,12 @@ ClientPreProcessRequestMsg::ClientPreProcessRequestMsg(NodeIdType sender,
...
 
-unique_ptr<MessageBase> ClientPreProcessRequestMsg::convertToClientRequestMsg(bool resetPreProcessFlag) {
+unique_ptr<MessageBase> ClientPreProcessRequestMsg::convertToClientRequestMsg(bool resetPreProcessFlag, bool empty) {
   if (resetPreProcessFlag) msgBody()->flags &= ~(1 << 1);
   unique_ptr<MessageBase> clientRequestMsg = make_unique<ClientRequestMsg>(clientProxyId(),
                                                                            flags(),
                                                                            requestSeqNum(),
-                                                                           requestLength(),
+                                                                           empty ? 0 : requestLength(),
                                                                            requestBuf(),
```
yuliasherman, line 37: I guess we don't need to pass and then copy request buffer either.

bftengine/src/preprocessor/RequestProcessingState.cpp
```diff
@@ -193,8 +193,8 @@ PreProcessingResult RequestProcessingState::definePreProcessingConsensusResult()
   return CONTINUE;
 }
 
-unique_ptr<MessageBase> RequestProcessingState::buildClientRequestMsg(bool resetPreProcessFlag) {
-  return clientPreProcessReqMsg_->convertToClientRequestMsg(resetPreProcessFlag);
+unique_ptr<MessageBase> RequestProcessingState::buildClientRequestMsg(bool resetPreProcessFlag, bool empty) {
```
yuliasherman, line 196: empty -> emptyReq

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4054,6 +4054,7 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
       continue;
     }
     ClientRequestMsg req((ClientRequestMsgHeader *)requestBody);
+    if (req.requestLength() == 0) continue;
```
yuliasherman, line 4057: We need to call to clientsManager->removePendingForExecutionRequest() here, otherwise, the problem will not be resolved.

yuliasherman (changes requested):

bftengine/src/preprocessor/messages/ClientPreProcessRequestMsg.cpp
yontyon: if the size is 0, then nothing will be copied

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: we call it on the commit itself (before the execution)

yuliasherman: I see that in tryToRemovePendingRequestsForSeqNum we just mark them as committed. The call to removePendingForExecutionRequest is only one.

yontyon: right. 
Done.

yuliasherman (approved):
LGFM! Thanks for fixing that.

PR #1379, opened 2021-04-13
Add source selector unit tests

StefanPalashev:


bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  ASSERT_TRUE(source_selector.shouldReplaceSource(kSmallDeltaTime, true));
+}
+
+TEST(source_selector_test, should_change_source_when_the_replacement_time_has_elapsed) {
+  auto source_selector = SourceSelector(replicas, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  source_selector.updateSource(kSampleCurrentTimeMs);
+  ASSERT_TRUE(source_selector.shouldReplaceSource(kBigDeltaTime, false));
+}
+
+TEST(source_selector_test, source_should_NOT_be_changed_when_working_and_replacement_time_has_NOT_elapsed_yet) {
```
dartdart26, line 85: Nit: Maybe we can just use lower case for `NOT`. Not exactly sure what the coding standard for test names is, though. My personal preference is lower case.

dartdart26: Just came to mind that we can shorten it by saying something like `source_should_not_be_changed_if_replacement_time_has_not_elapsed`.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  source_selector.updateSource(kSampleCurrentTimeMs);
+  ASSERT_NE(source_selector.currentReplica(), NO_REPLICA);
+  source_selector.removeCurrentReplica();
+  ASSERT_EQ(source_selector.currentReplica(), NO_REPLICA);
+}
+
+TEST(source_selector_test, removing_the_current_replica_makes_it_non_preferred_as_well) {
+  auto source_selector = SourceSelector(replicas, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  source_selector.updateSource(kSampleCurrentTimeMs);
+  const auto currentReplica = source_selector.currentReplica();
```
dartdart26, line 104: Nit: currentReplica -> current_replica

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  ASSERT_FALSE(source_selector.isPreferred(currentReplica));
+}
+
+// Reset tests
+TEST(source_selector_test, on_construction_is_reset_reports_true) {
+  auto source_selector = SourceSelector(replicas, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  ASSERT_TRUE(source_selector.isReset());
+}
+
+TEST(source_selector_test, leave_initial_state_when_there_are_any_preferred_replicas) {
```
dartdart26, line 116: Nit: Maybe we can say `not_reset_when_there_are_preferred_replicas`.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  ASSERT_TRUE(source_selector.isReset());
+  source_selector.setFetchingTimeStamp(GL, kSampleCurrentTimeMs);
+  ASSERT_FALSE(source_selector.isReset());
+}
+
+TEST(source_selector_test, reset_removes_preferred_replicas) {
+  auto source_selector = SourceSelector(replicas, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  ASSERT_EQ(source_selector.numberOfPreferredReplicas(), 0);
+  source_selector.setAllReplicasAsPreferred();
+  ASSERT_NE(source_selector.numberOfPreferredReplicas(), 0);
```
dartdart26, line 141: Should we assert for equal to `replicas.size()` instead?

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+constexpr auto kRetransmissionTimeoutMs = 100;
+constexpr auto kReplicaReplacementTimeoutMs = 300;
+
+constexpr uint64_t kSampleCurrentTimeMs = 1000;
+constexpr auto kSmallDeltaTime = kSampleCurrentTimeMs + 1;
+constexpr auto kBigDeltaTime = 10 * kSampleCurrentTimeMs;
+constexpr uint16_t kSampleReplicaId = 1;
+
+const auto replicas = std::set<uint16_t>{1, 2, 3, 4};
```
dartdart26, line 32: A thing that came to mind is that we use `replicas` as `other` replicas, meaning that for a cluster of 4 replicas, a more meaningful count of others is 3. Maybe we can change that set to 3 or 6 elements and call it `other_replicas`. What do you think?

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  ASSERT_EQ(source_selector.numberOfPreferredReplicas(), 0);
+  source_selector.setAllReplicasAsPreferred();
+  ASSERT_NE(source_selector.numberOfPreferredReplicas(), 0);
+  source_selector.reset();
+  ASSERT_EQ(source_selector.numberOfPreferredReplicas(), 0);
+}
+
+TEST(source_selector_test, reset_removes_current_replica) {
+  auto source_selector = SourceSelector(replicas, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  ASSERT_TRUE(source_selector.currentReplica() == NO_REPLICA && source_selector.isReset());
```
dartdart26, line 148: I think it would be better to use separate asserts, because if one fails, it will be clearer which one it was.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  ASSERT_FALSE(source_selector.hasPreferredReplicas());
+  source_selector.setAllReplicasAsPreferred();
+  ASSERT_TRUE(source_selector.hasPreferredReplicas());
+
+  for (const auto& replica : replicas) {
+    ASSERT_TRUE(source_selector.isPreferred(replica));
+  }
+}
+
+TEST(source_selector_test, select_an_unique_replica_on_each_selection) {
```
dartdart26, line 189: Nice test!

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  auto source_selector = SourceSelector(replicas, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  ASSERT_TRUE(source_selector.currentReplica() == NO_REPLICA && source_selector.isReset());
+  source_selector.updateSource(kSampleCurrentTimeMs);
+  // Set the selection and fetching times to zero in order to ensure that the reset function's
+  // result is solely based on the current replica.
+  source_selector.setSourceSelectionTime(0);
+  source_selector.setFetchingTimeStamp(GL, 0);
+  ASSERT_FALSE(source_selector.isReset());
+  source_selector.reset();
+  ASSERT_TRUE(source_selector.currentReplica() == NO_REPLICA && source_selector.isReset());
```
dartdart26, line 156: I think separate asserts would be better.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+  auto source_selector = SourceSelector({}, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  source_selector.addPreferredReplica(kSampleReplicaId);
+  ASSERT_TRUE(source_selector.isPreferred(kSampleReplicaId));
+}
+
+TEST(source_selector_test, select_the_only_preferred_replica) {
+  auto source_selector = SourceSelector({}, kRetransmissionTimeoutMs, kReplicaReplacementTimeoutMs);
+  source_selector.addPreferredReplica(kSampleReplicaId);
+  source_selector.updateSource(kSampleCurrentTimeMs);
+  ASSERT_TRUE(source_selector.isPreferred(kSampleReplicaId) && source_selector.currentReplica() == kSampleReplicaId);
```
dartdart26, line 222: I think separate asserts would be better.

dartdart26 (approved):
Looks great, thanks @StefanPalashev! I have added a few minor comments.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
StefanPalashev: The assertion for `replicas.size()` is more accurate and it has been applied. Thanks!

bftengine/tests/bcstatetransfer/source_selector_test.cpp
StefanPalashev: The asserts have been separated.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
StefanPalashev: The asserts have been separated.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
StefanPalashev: This is a good point, thanks.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
StefanPalashev: The casing has been taken into consideration and applied, however I think the name should still suggest that the source has to be working when looking for a timeout, otherwise it has to be changed regardless of the replacement time.

bftengine/tests/bcstatetransfer/source_selector_test.cpp
StefanPalashev: The set now includes only three replicas, considering that the cluster contains four replicas in total.

PR #1381, opened 2021-04-13
Fix memory leak in onMessage<AskForCheckpointMsg>

f-squirrel:


tdimitrov (approved):
Good catch

HristoStaykov (commented):
Good catch, @f-squirrel , thanks!

PR #1382, opened 2021-04-13
Always start important traces

bayarcher:
Add root traces in the absences of span context in the client request messages.


f-squirrel:
We've discussed the PR offline with @bayarcher and agreed to try another approach.
Stay tuned! :)

bayarcher:
Thanks for the quick review and discussion @f-squirrel.

PR #1383, opened 2021-04-14
Allow primary replica rejecting client requests arrived through non-primaries

yuliasherman:
To avoid request timeouts on non-primary replicas, the primary should send a 'reject' message in case it cannot handle the request.


salieri11 (approved):
LGTM

PR #1384, opened 2021-04-14
Fix the trigger for sending a complaint for VC

yontyon:
Now with the new client manager, consider the following scenario:
1. Request X has beeen received by the primary
2. The primary adds X to the pending requests and starts consensus for a pp message that contains X
3. A nonprimary replica R commits the pp message, as it has never got the client request, it has nothing to remove from its pending requests
4. Now, the client retries X and sends X to all replicas.
5. R gets the client request and puts it in its pending requests
6. X will be never removed from R's pending requests, as it will never be committed again

We fix it by having every replica tracking the client requests that received either by the client or in the pre-prepare.
The later solution was decided in a discussion between @ittaiab @HristoStaykov and @yontyon

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -702,6 +702,15 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
       }
       msgAdded = true;
 
+      // Start tracking all client requests with in this pp message
+      RequestsIterator reqIter(msg);
+      char *requestBody = nullptr;
+      while (reqIter.getAndGoToNext(requestBody)) {
+        ClientRequestMsg req((ClientRequestMsgHeader *)requestBody);
+        if (!clientsManager->isValidClient(req.clientProxyId())) continue;
+        if (clientsManager->canBecomePending(req.clientProxyId(), req.requestSeqNum()))
+          clientsManager->addPendingRequest(req.clientProxyId(), req.requestSeqNum(), req.getCid());
+      }
       if (ps_) {
         ps_->beginWriteTran();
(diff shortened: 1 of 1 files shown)


Liorgal28 (approved):
LGTM, Thanks @yontyon !

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -336,8 +336,15 @@ Time ClientsManager::infoOfEarliestPendingRequest(std::string& cid) const {
...
+        earliestPendingReqInfo = {MaxTime, std::string()};
+        earliestTime = MaxTime;
+        break;
+      }
       // Don't take into account already committed requests
-      if ((req.second.time != MinTime) && (earliestTime > req.second.time) && (!req.second.committed)) {
+      if ((req.second.time != MinTime) && (earliestTime > req.second.time)) {
         earliestPendingReqInfo = req.second;
         earliestTime = earliestPendingReqInfo.time;
       }
```
yuliasherman, line 350: Maybe it worth changing the name of this function to reflect what it actually does now.

yuliasherman (approved):
LGTM

f-squirrel (commented):
Great catch!

HristoStaykov (approved):
Great catch and fix @yontyon. Thanks!

andrewjstone (approved):
Great find, fix, and commit message @yontyon!

andrewjstone:
Actually, @yontyon the great description is only in the MR, but not the commit message. Can you possibly amend the commit to include the description?

yontyon:
In a test that we did, it seems to improve the stability of the system (even that we don't know that this bug does not fully explain the VC, as discussed with @HristoStaykov).
@HristoStaykov I think we should merge it anyway, and track the runs to see if the VC storm is happening again.

PR #1389, opened 2021-04-16
Using a static secrets_manager can cause thread issues

KTeofanidis:


tdimitrov (approved):

communication/src/AsyncTlsConnection.cpp
```diff
@@ -466,25 +466,20 @@ std::pair<bool, NodeNum> AsyncTlsConnection::checkCertificate(X509* receivedCert
...
   namespace fs = boost::filesystem;
   std::string pkpath;
 
+  std::unique_ptr<ISecretsManagerImpl> secrets_manager;
   if (tlsTcpImpl_.config_.secretData) {
     pkpath = (path / fs::path("pk.pem.enc")).string();
-    if (secrets_manager_ == nullptr) {
-      secrets_manager_.reset(new SecretsManagerEnc(tlsTcpImpl_.config_.secretData.value()));
-    }
+    secrets_manager.reset(new SecretsManagerEnc(tlsTcpImpl_.config_.secretData.value()));
```
f-squirrel, line 476: Do we actually want to create the secrets manager for each `AsyncTlsConnection`?
Maybe we should create it somewhere else and then provide it to the `AsyncTlsConnection`?
WDYT?

andrewjstone: I'm in the middle of rewriting the TLS code to remove all locks and be multithreaded via the use of asio strands. We absolutely should not have a secrets manager per connection. We should talk more about this next week. I'll look at the current implementation, but we really should not be using the secrets at all outside of establishing a connection and that can be done in the connection manager strand, or better yet, once at startup.

tdimitrov: The solution might not be perfect, but I want to make a small clarification. 
If the main concern is performance, copying the SecretsManager on each new connection is not that expensive - only a copy of iv and secret key is made. This is 42 bytes.
The manager itself is a wrapper around crypto++ code.

Speaking more generally I see mix of two styles in the code - "initialize once (in main) and pass pointers to wherever needed" and "do it singleton style". Maybe we should agree on a default approach and follow it unless there is a good reason?

P.S. I'm in the pointer camp.

andrewjstone: Sorry, this was my mistake. This change is totally fine. It's just due to the way keys are read that the code for reading is in the `AsyncTlsConnection` class. This code is still only run during connection establishment, and this is where it must live. With the multithreading changes it won't even block reads and writes for established connections. There is no issue here.

andrewjstone (approved):
LGTM. Sorry for the confusion @KTeofanidis.

communication/src/AsyncTlsConnection.cpp
toly-kournik: Pointer/singleton approach depends on the usage of the specific class.
In our case `SecretsManager` is a _service_ which may be used by different modules, e.g. by `KeyExchangeManager`'s internals to securely store the Replica's private key. I find it highly inconvenient to pass it all the way down.

tdimitrov: It's actually an utility class. You can have as many instances as you wish.

PR #1390, opened 2021-04-16
Add name for histogram related to State Transfer

saiakhil2012:
Add name for the histogram corresponding to the State Transfer for use by `concord-ctl`


teoparvanov (approved):
LGTM, thanks @saiakhil2012 !

PR #1395, opened 2021-04-20
AddRemoveCommand for reconfiguration engine

yontyon:
We add the AddRemoveCommand to the reconfiguration engine.
The main benefit of this PR is to persist the new configuration in the BC

### DIFF
--- a/kvbc/include/reconfiguration_add_block_handler.hpp
+++ b/kvbc/include/reconfiguration_add_block_handler.hpp
@@ -89,6 +89,17 @@ class ReconfigurationHandler : public concord::reconfiguration::IReconfiguration
     return true;
   }
 
+  bool handle(const concord::messages::AddRemoveCommand& command,
+              concord::messages::ReconfigurationErrorMsg&,
+              uint64_t sequence_number) override {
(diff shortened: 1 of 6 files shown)


kvbc/include/reconfiguration_add_block_handler.hpp
```diff
@@ -89,6 +89,17 @@ class ReconfigurationHandler : public concord::reconfiguration::IReconfiguration
...
 
+  bool handle(const concord::messages::AddRemoveCommand& command,
+              concord::messages::ReconfigurationErrorMsg&,
+              uint64_t sequence_number) override {
+    std::vector<uint8_t> serialized_command;
+    concord::messages::serialize(serialized_command, command);
+    auto blockId =
+        persistReconfigurationBlock(serialized_command, sequence_number, kvbc::keyTypes::reconfiguration_add_remove);
+    LOG_INFO(getLogger(), "AddRemoveCommand command block is " << blockId);
+    return true;
```
Liorgal28, line 100: If there's no way to get false maybe the method should not return anything

Liorgal28 (approved):
LGTM

PR #1397, opened 2021-04-20
SigManager: (temporary patch) - define instance_ in header

glevkovich:
This is a temporary patch. Task BC-8242 should revert after solving the
the issue in the right way.

Since some external modules, such as concord external client, include
all concord bftengine (core) header files - we must move Signature
Manager static instance_ definition into the header file. External
modules look for a definition and must find it.

To avoid multiple definitions the flag _SIG_MANAGER_CPP_ must be
defined to allow defining instance_ only once. I have chosen
SimpleClientImp.cpp to define this flag on the top since this is a
common file for both concord-bft core library, and external client.

### DIFF
--- a/bftengine/src/bftengine/SigManager.cpp
+++ b/bftengine/src/bftengine/SigManager.cpp
@@ -21,8 +21,6 @@ using namespace std;
 namespace bftEngine {
 namespace impl {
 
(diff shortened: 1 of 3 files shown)


yontyon (approved):
Thanks!

PR #1398, opened 2021-04-20
Makefile: Make all variables configurable via CLI and env variables

f-squirrel:
This PR is a continuation of #1338.

Reason:
Before the change the only way to configure Makefile variables was
```
make VARIABLE_NAME=value
```

After the change Makefile will support the old way and more canonical configuration:
```
VARIABLE_NAME=value make
```

or
```
export VARIABLE_NAME=value
make
```

Thus variable can be set not only via the CLI but also as a regular environment variable.


andrewjstone (approved):
Looks great.  Thanks @f-squirrel .

PR #1399, opened 2021-04-20
View change with f replicas collected stable checkpoint

HristoStaykov:
This test is a recreation of a infinite View Change scenario, where we have only F Replicas that have gathered a Stable Checkpoint. This causes disagreement in the computed Restrictions for the previous View's Working Window and View Change is never finalized.

### DIFF
--- a/tests/apollo/test_skvbc_chaotic_startup.py
+++ b/tests/apollo/test_skvbc_chaotic_startup.py
@@ -22,6 +22,8 @@ from util.bft import with_trio, with_bft_network, with_constant_load, KEY_FILE_P
 from util import bft_network_partitioning as net
 from util import eliot_logging as log
 
+from util import skvbc as kvbc
+
 def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="10000"):
     """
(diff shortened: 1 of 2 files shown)


f-squirrel:
@HristoStaykov , what an amazing way to document issues that are hard to reproduce!

teoparvanov (approved):
Awesome, thanks @HristoStaykov !

tests/apollo/test_skvbc_chaotic_startup.py
```diff
@@ -48,6 +50,99 @@ class SkvbcChaoticStartupTest(unittest.TestCase):
 
     __test__ = False  # so that PyTest ignores this test scenario
 
+    @unittest.skip("Edge case scenario - not part of CI")
```
teoparvanov, line 53: can you also specify the BC-XXXX ticket number that this test allows to reproduce?

f-squirrel: AFAIK, the approach is to avoid mentioning ticket numbers since they are not public.

PR #1400, opened 2021-04-20
PreProcessor pushes a wrong message to ReplicaImp

yuliasherman:
Before pushing a message to ReplicaImp (e.g when it has already been executed), PreProcessor has to convert ClientPreProcessRequest to ClientRequest, otherwise, it will return again to PreProcessor as it has a registered message handler for it.


teoparvanov (approved):
LGTM, thanks Yulia!

PR #1401, opened 2021-04-21
WIP: Remove not relevant requests from pending

yontyon:
This change is to reduce the number of VC we experienced. In this commit we want to enforce the following interval: A client cannot have more pending requests than the max client batch size.
How to force it:
In a former commit (212114f) we made a nonprimary replica to start tracking the client messages that are in the primary pre-prepare message.
Now consider the following scenario,
1. Assume the client batch size is 1
2. Due to some networks delays, data races, or Byzantine client, the primary replica started to handle request X while a nonprimary replica started to handle request Y where Y> X
3. The nonprimary will tries to send Y to the primary, however, the primary is already busy with X
4. The nonprimary also will not be able to add X to its pending requests as it is already taken by Y
5. X will be committed and removed from the primary pending requests, while Y will never be committed
6. Y will never be committed and eventually, VC will happen

To solve the above scenario, before adding a specific request in pre-prepare to the pending requests, we first try to see if there is an out-boundaries request. In the example above, we got X in the
pre-prepare and we know that the client should have only one ongoing request then we can remove Y from the pending requests

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -278,6 +278,15 @@ bool ClientsManager::isClientRequestInProcess(NodeIdType clientId, ReqId reqSeqN
   return false;
 }
 
+bool ClientsManager::isPending(NodeIdType clientId, ReqId reqSeqNum) const {
+  uint16_t idx = clientIdToIndex_.at(clientId);
+  const auto& requestsInfo = indexToClientInfo_.at(idx).requestsInfo;
+  const auto& reqIt = requestsInfo.find(reqSeqNum);
+  if (reqIt != requestsInfo.end() && !reqIt->second.committed) {
+    return true;
+  }
+  return false;
+}
 // Check that:
 // * max number of pending requests not reached for that client.
 // * request seq number is bigger than the last reply seq number.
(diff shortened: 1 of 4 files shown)


bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -316,6 +325,31 @@ void ClientsManager::markRequestAsCommitted(NodeIdType clientId, ReqId reqSeqNum
...
+ * The client manager cannot hold request that are out of the bounds of a committed sequence number +
+ * maxNumOfRequestsInBatch We know that the client sequence number are always ascending. In order to keep this invariant
+ * we do the following: Every time we commit or execute a sequence number, we order all of our existing tracked sequence
+ * numbers. Then, we count how many bigger sequence number than the given reqSequenceNumber we have. We know for sure
+ * that we shouldn't have more than maxNumOfRequestsInBatch. Thus, we can safely remove them from the client manager.
+ */
+bool ClientsManager::removeRequestsOutsideBoundsOfBatch(NodeIdType clientId, ReqId reqSequenceNum) {
+  uint16_t idx = clientIdToIndex_.at(clientId);
+  auto& requestsInfo = indexToClientInfo_.at(idx).requestsInfo;
+  if (requestsInfo.find(reqSequenceNum) != requestsInfo.end()) return false;
```
yuliasherman, line 339: I don't understand the logic. In case we found reqSequenceNum, we exit the function?
It doesn't look that this function could serve a batch size of > 1.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -316,6 +326,30 @@ void ClientsManager::markRequestAsCommitted(NodeIdType clientId, ReqId reqSeqNum
...
 
+/*
+ * We have to keep the following invariant:
+ * The client manager cannot hold request that are out of the bounds of a committed sequence number +
+ * maxNumOfRequestsInBatch We know that the client sequence number are always ascending. In order to keep this invariant
+ * we do the following: Every time we commit or execute a sequence number, we order all of our existing tracked sequence
+ * numbers. Then, we count how many bigger sequence number than the given reqSequenceNumber we have. We know for sure
+ * that we shouldn't have more than maxNumOfRequestsInBatch. Thus, we can safely remove them from the client manager.
+ */
+void ClientsManager::removeRequestsOutsideBoundsOfBatch(NodeIdType clientId, ReqId reqSequenceNum) {
```
yuliasherman, line 337: removeRequestsOutsideBoundsOfBatch -> removeRequestsOutOfBatchBounds

yuliasherman (approved):
LGFM

PR #1403, opened 2021-04-21
ConcordClient: Service interface

robem:
With this change, we publish the future interface for the Concord Client
service. In short, Concord Client service is a wrapper around the Concord
Client library which includes a pool of bft clients and a subscription
mechanism. Concord Client and its components will be introduced and documented
in future changes.


andrewjstone (approved):
Looks great. This is the result of a lot of teamwork. The simplicity hides the effort that went in here, and I think that's exactly what we wanted :)

PR #1404, opened 2021-04-22
SigManager: (temporary patch) - move instance definition to another file

glevkovich:
This is a temporary patch. Task BC-8242 should revert after solving the
the issue in the right way.

Since some external modules, such as concord external client, include
all concord bftengine (core) header files - linker will look to the
instance definition, which is in SigManager.cpp before the change.
Here we must move Signature Manager static instance_ definition a new
commonplace - SimpleClientimp.cpp. This file is also compiled
by the external client.


Liorgal28 (approved):
LGTM

PR #1405, opened 2021-04-22
Move static declaration to getInstance

yontyon:


bftengine/src/bftengine/SigManager.cpp
```diff
@@ -88,23 +86,17 @@ SigManager* SigManager::init(ReplicaId myId,
...
+                            mySigPrivateKey,
+                            publicKeysOfReplicas,
+                            replicasKeysFormat,
+                            publicKeysOfClients,
+                            clientsKeysFormat,
+                            numReplicas,
+                            numRoReplicas,
+                            numOfClientProxies,
+                            numOfExternalClients);
+  return SigManager::getInstance(sm);
```
toly-kournik, line 99: We use the `instance()` everywhere

glevkovich: @toly-kournik , I will rename it in another dedicated PR.
I suggest keeping this name as-is for now, this modification is urgent.

glevkovich (approved):

PR #1407, opened 2021-04-22
Several fixes related to the client batching feature

yuliasherman:
1. Use isClientRequestInProcess instead of canBecomePending to define if a request is in process by ReplicaImp
2. As batched client requests don't have a correct senderId, use ClientBatchRequestMsg senderId to identify that a message was forwarded by a non-primary replica.
3. Don't resend replies to the client for messages forwarded by non-primary replicas.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -267,6 +267,7 @@ void PreProcessor::updateAggregatorAndDumpMetrics() {
 }
 
 void PreProcessor::sendCancelPreProcessRequestMsg(const ClientPreProcessReqMsgUniquePtr &clientReqMsg,
+                                                  NodeIdType destId,
                                                   uint16_t reqOffsetInBatch,
                                                   uint64_t reqRetryId) {
   const auto clientId = clientReqMsg->clientProxyId();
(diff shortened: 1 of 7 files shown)


teoparvanov (approved):
LGTM, thanks Yulia!

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -471,6 +477,7 @@ void PreProcessor::onMessage<ClientBatchRequestMsg>(ClientBatchRequestMsg *msg)
 // Non-primary replica request handling
 template <>
 void PreProcessor::onMessage<PreProcessRequestMsg>(PreProcessRequestMsg *msg) {
+  concord::diagnostics::TimeRecorder scoped_timer(*histograms_.onPreProcessRequestMsg);
```
andrewjstone, line 480: Is this histogram updated from multiple threads? If so, this should be changed to use the atomic variant with `TimeRecorder<true>`.

yuliasherman: It is always called from IncomingMsgsStorage context, which is alone. Thanks for the reminder!

andrewjstone: Cool. Thanks :)

PR #1408, opened 2021-04-22
Fix span context extraction from ClientPreProcessRequestMsg

f-squirrel:
The calculation of the position of span context is based on the size of
`MessageType::Header`.

Most of the message headers look like this:
```
struct Header {
MessageBase::Header header;
MessageSpecific field1;
...
MessageSpecific fieldN;
};
```
However, `Client*Request` messages have a special declaration of headers where
`MessageBase:Header` is embedded into a `Client*Request`' message struct.  This
requires a template specialization for the function `sizeOfHeader`.

Since `ClientPreProcessRequestMsg` did not have it,
`sizeOfHeader<ClientPreProcessRequestMsg>` would return 6 (the size of
`MessageBase::Header`) instead of 35 (the size of `ClientRequestMsgHeader`).

Changes:
- Added the specialization for `ClientPreProcessRequestMsg`.
- Added unit tests.


yuliasherman (approved):
LGTM! Thanks for fixing that.

teoparvanov (approved):
Awesome, thanks Dima!

PR #1409, opened 2021-04-22
Docker image update to version 0.22

andrewjstone:
This MR adds support for the latest version of ASIO. ASIO comes in two
flavors: standalone, and packaged with boost. We include the standalone
here to allow independent upgrade of ASIO without upgrading boost. Boost
touches not only things in concord-bft, but also upstream, and so
updgrading it is a bigger effort.

The immediate use case for ASIO is to support multi-threaded TLS which will come
in a follow up commit.


f-squirrel:
The migration to the standalone version of ASIO seems to be a serious change. I'd expect it to be discussed in a wider forum.

* Why can we not just update the boost? The upgrade of boost is something that we'll have to do anyway. So why postpone it?
* Are the two versions of ASIO interchangeable?
* Is the standalone version as popular and frequently reviewed as the one from boost?

f-squirrel (approved):
Thank you @andrewjstone for the clarification!

andrewjstone:
> Thank you @andrewjstone for the clarification!

Thanks for listening @f-squirrel :)

PR #1411, opened 2021-04-25
Read only replica pruning Apollo test

yontyon:
In this PR we introduce a full RO replica pruning support.
The pruning is done using a bft client and the new reconfiguration engine

### DIFF
--- a/kvbc/include/pruning_handler.hpp
+++ b/kvbc/include/pruning_handler.hpp
@@ -210,6 +210,11 @@ class ReadOnlyReplicaPruningHandler : public reconfiguration::IPruningHandler {
     latest_prunable_block.replica = replica_id_;
     latest_prunable_block.block_id = latest_prunable_block_id;
     signer_.sign(latest_prunable_block);
+    LOG_INFO(GL,
+             KVLOG(latest_prunable_block.bft_sequence_number,
+                   latest_prunable_block.replica = replica_id_,
+                   latest_prunable_block.block_id,
+                   std::string(latest_prunable_block.signature.begin(), latest_prunable_block.signature.end())));
     return true;
   }
   // Read only replica doesn't perform the actual pruning
--- a/kvbc/include/pruning_handler.hpp
+++ b/kvbc/include/pruning_handler.hpp
@@ -213,8 +213,7 @@ class ReadOnlyReplicaPruningHandler : public reconfiguration::IPruningHandler {
     LOG_INFO(GL,
              KVLOG(latest_prunable_block.bft_sequence_number,
                    latest_prunable_block.replica = replica_id_,
-                   latest_prunable_block.block_id,
-                   std::string(latest_prunable_block.signature.begin(), latest_prunable_block.signature.end())));
+                   latest_prunable_block.block_id));
     return true;
   }
   // Read only replica doesn't perform the actual pruning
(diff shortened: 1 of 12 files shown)


kvbc/include/pruning_handler.hpp
```diff
@@ -210,6 +210,11 @@ class ReadOnlyReplicaPruningHandler : public reconfiguration::IPruningHandler {
     latest_prunable_block.replica = replica_id_;
     latest_prunable_block.block_id = latest_prunable_block_id;
     signer_.sign(latest_prunable_block);
+    LOG_INFO(GL,
+             KVLOG(latest_prunable_block.bft_sequence_number,
+                   latest_prunable_block.replica = replica_id_,
+                   latest_prunable_block.block_id,
+                   std::string(latest_prunable_block.signature.begin(), latest_prunable_block.signature.end())));
```
dartdart26, line 217: Nit: I would assume that printing this buffer as a string would possibly produce garbage, because of non-printable chars. Maybe we should hexify it.

tests/apollo/CMakeLists.txt
```diff
@@ -97,9 +97,11 @@ add_test(NAME skvbc_backup_restore COMMAND sh -c
         "env ${APOLLO_TEST_ENV} BUILD_COMM_TCP_TLS=${BUILD_COMM_TCP_TLS} TEST_NAME=skvbc_backup_restore python3 -m unittest test_skvbc_backup_restore 2>&1 > /dev/null"
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
-add_test(NAME skvbc_reconfiguration COMMAND sh -c
-        "env ${APOLLO_TEST_ENV} BUILD_COMM_TCP_TLS=${BUILD_COMM_TCP_TLS} TEST_NAME=skvbc_reconfiguration python3 -m unittest test_skvbc_reconfiguration ${TEST_OUTPUT}"
-        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+if (USE_S3_OBJECT_STORE)
```
dartdart26, line 100: Why do we only do it if **USE_S3_OBJECT_STORE** is set?

yontyon: Now the reconfiguration tests contain also pruning with RO replica which requires S3

kvbc/include/pruning_handler.hpp
yontyon: fixed

tdimitrov (approved):
Nice work, Yoni. RO replica testing in apollo looks much more integrated now in comparison with the old ad-hoc approach.

tests/apollo/CMakeLists.txt
dartdart26: Is there a way to lift this restriction so that whoever builds without can still test reconfiguration?

dartdart26 (approved):
Looks good, thanks @yontyon !

kvbc/src/ReplicaImp.cpp
```diff
@@ -175,7 +175,12 @@ std::optional<categorization::Updates> ReplicaImp::getBlockUpdates(BlockId block
 
 BlockId ReplicaImp::getGenesisBlockId() const { return m_kvBlockchain->getGenesisBlockId(); }
 
-BlockId ReplicaImp::getLastBlockId() const { return m_kvBlockchain->getLastReachableBlockId(); }
+BlockId ReplicaImp::getLastBlockId() const {
+  if (replicaConfig_.isReadOnly) {
```
toly-kournik, line 179: what a beautiful if

tests/apollo/CMakeLists.txt
yontyon: we can separate it into a different test file. For now I will leave it as it is, if we see that it causes issue I will consider to separate it

PR #1413, opened 2021-04-27
Update log messages to indicate fast/slow path

shgandhi:
This commit updates log messages to include commit path, this is to
allow easy debugging without having to correlate a specific protocol
with the commit path being used. This closes BC-7983.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -531,11 +531,12 @@ void ReplicaImp::startConsensusProcess(PrePrepareMsg *pp, bool isInternalNoop) {
   SCOPED_MDC_PATH(CommitPathToMDCString(firstPath));
 
   if (isInternalNoop) {
-    LOG_INFO(CNSUS, "Sending PrePrepare message containing internal NOOP");
+    LOG_INFO(CNSUS, "Sending PrePrepare message containing internal NOOP, commit path: " << CommitPathToStr(firstPath));
   } else {
     LOG_INFO(CNSUS,
              "Sending PrePrepare message" << KVLOG(pp->numberOfRequests())
-                                          << " correlation ids: " << pp->getBatchCorrelationIdAsString());
+                                          << " correlation ids: " << pp->getBatchCorrelationIdAsString()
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -968,7 +974,9 @@ void ReplicaImp::sendPreparePartial(SeqNumInfo &seqNumInfo) {
 
     ConcordAssertNE(pp, nullptr);
 
-    LOG_DEBUG(MSGS, "Sending PreparePartialMsg. " << KVLOG(pp->seqNumber()));
+    LOG_INFO(MSGS,
+             "Sending PreparePartialMsg, sequence number:" << pp->seqNumber()
+                                                           << ", commit path: " << CommitPathToStr(pp->firstPath()));
```
teoparvanov, line 979: I think this will flood the logs... It should remain at DEBUG level IMO.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1000,7 +1008,9 @@ void ReplicaImp::sendCommitPartial(const SeqNum s) {
 
   if (seqNumInfo.committedOrHasCommitPartialFromReplica(config_.getreplicaId())) return;  // not needed
 
-  LOG_DEBUG(CNSUS, "Sending CommitPartialMsg");
+  LOG_INFO(CNSUS,
+           "Sending CommitPartialMsg, sequence number:" << pp->seqNumber()
+                                                        << ", commit path: " << CommitPathToStr(pp->firstPath()));
```
teoparvanov, line 1013: same here, should be DEBUG

teoparvanov (changes requested):
Hi Shruti, thanks for attempting a fix. However, I wouldn't suggest making those logs INFO, as they will pollute the log output. We are already looking for ways to reduce the overall log size.

CC: @HristoStaykov

shgandhi:
Thanks for the review @teoparvanov, updated the PR as per your comments.

HristoStaykov (approved):
LGTM, thanks @shgandhi for the added thorough information on the Commit paths.

teoparvanov (approved):
LGTM, thanks Shruti!

PR #1417, opened 2021-04-27
Add test for verifying catching up from Inactive Window, and a fix for the situation the test recreates. 

HristoStaykov:
The goal of this test is to verify full catch up of a Replica
only from the Inactive Window by isolating it from all others
except the Primary, so that State Transfer will not be able to
start.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2114,7 +2114,7 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
         }
         // help from Inactive Window
         if (mainLog->isPressentInHistory(msg->getLastExecutedSeqNum() + 1)) {
-          for (SeqNum i = msg->getLastExecutedSeqNum() + 1; i < lastStableSeqNum + 1; i++) {
+          for (SeqNum i = msg->getLastExecutedSeqNum() + 1; i <= lastStableSeqNum; i++) {
             PrePrepareMsg *prePrepareMsg = mainLog->getFromHistory(i).getSelfPrePrepareMsg();
             if (prePrepareMsg != nullptr && !msg->isPrePrepareInActiveWindow(i)) {
               sendAndIncrementMetric(prePrepareMsg, msgSenderId, metric_sent_preprepare_msg_due_to_status_);
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2112,6 +2112,15 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
             sendAndIncrementMetric(prePrepareMsg, msgSenderId, metric_sent_preprepare_msg_due_to_status_);
           }
         }
+        // help from Inactive Window
+        if (mainLog->isPressentInHistory(msg->getLastExecutedSeqNum() + 1)) {
+          for (SeqNum i = msg->getLastExecutedSeqNum() + 1; i < lastStableSeqNum + 1; i++) {
+            PrePrepareMsg *prePrepareMsg = mainLog->getFromHistory(i).getSelfPrePrepareMsg();
+            if (prePrepareMsg != nullptr && !msg->isPrePrepareInActiveWindow(i)) {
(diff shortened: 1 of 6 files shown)


tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -1085,6 +1085,9 @@ async def tracked_prime_for_state_transfer(
             kv = [(key, val)]
             await self.write_and_track_known_kv(kv, client1)
 
+        await self.bft_network.wait_for_replicas_to_collect_stable_checkpoint(self.bft_network.get_live_replicas(),
```
toly-kournik, line 1088: where is it used?

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -40,6 +40,13 @@ SeqNumInfo::~SeqNumInfo() {
 }
 
 void SeqNumInfo::acquire(SeqNumInfo& rhs) {
+  primary = rhs.primary;
```
toly-kournik, line 43: how could it work before?

HristoStaykov: It was not working correctly. It was only working in the case where a Replica has actually received the PrePrepare from the Primary. With the fix the Primary will resend the PrePrepares on Status exchange.

tests/apollo/util/skvbc_history_tracker.py
HristoStaykov: This synchronisation point makes sure all live Replicas will collect a Stable Checkpoint. Without it the test test_st_when_fetcher_crashes in SkvbcPersistenceTest suite was consistently failing, because the late Replica (stale_node) catches up from the inactive Window, while enough Replicas collect a Stable Checkpoint in order for State Transfer to start.

toly-kournik (approved):
LGTM!
@HristoStaykov do you want some Apollo expert to review as well?

tests/apollo/util/skvbc_history_tracker.py
KTeofanidis, line 1088: Maybe use network_wait_for_checkpoint?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2112,6 +2112,15 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
             sendAndIncrementMetric(prePrepareMsg, msgSenderId, metric_sent_preprepare_msg_due_to_status_);
           }
         }
+        // help from Inactive Window
+        if (mainLog->isPressentInHistory(msg->getLastExecutedSeqNum() + 1)) {
+          for (SeqNum i = msg->getLastExecutedSeqNum() + 1; i < lastStableSeqNum + 1; i++) {
```
KTeofanidis, line 2117: <= is probably easier to read

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2112,6 +2112,15 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(ReplicaStatusMsg *msg) {
...
+            PrePrepareMsg *prePrepareMsg = mainLog->getFromHistory(i).getSelfPrePrepareMsg();
+            if (prePrepareMsg != nullptr && !msg->isPrePrepareInActiveWindow(i)) {
```
KTeofanidis, line 2119: !msg->isPrePrepareInActiveWindow(i) should probably be an assert since we're getting the messages from the history.

KTeofanidis (changes requested):

tests/apollo/util/skvbc_history_tracker.py
HristoStaykov: I checked the code and **network_wait_for_checkpoint** waits for the **last_stored_checkpoint** metric in Replicas to get to the needed value. This metric is updated when we generate a Checkpoint Message. In order to shift the Working Window we need the Replicas to collect a Stable Checkpoint.

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: This check is in order not to re-transmit unnecessarily. The Replica that sends Status Exchange request will state for which SeqNo-s it already has PrePrepares, which will be in its Active Window. For the Replica that is helping in this case the SeqNo will be in the Inactive Window.

KTeofanidis (approved):
LGTM!

PR #1420, opened 2021-04-28
Add ro replicas support in new bft client

yontyon:
We want to enable the reconfiguration requests in the read-only replica. For this, we need to enable the client to send requests to the read-only replica.
In this PR we introduce basic support for this in the new bft client. 
This support does not break anything with the previous behavior. In order to issue a bft request to the read-only replicas, the client has to explicitly specify them in the quorum.


andrewjstone (approved):
LGTM. Nice backwards compatible change @yontyon!

PR #1422, opened 2021-04-28
Implement non-blocking messages handling in the PreProcessor

salieri11:
This PR implememts asynchronous PreProcessor.
The idea is to release dispatching thread, which is main replica's thread, ASAP when any PreProcessor message is handled by replica.
This implementation uses lockfree SPSC queue from Boost, since it perfectly fits this scenario

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -130,7 +130,7 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
       numOfThreads = myReplica.getReplicaConfig().numOfClientProxies / numOfReplicas_;
   }
   threadPool_.start(numOfThreads);
-  msg_proc_thread_ = std::thread{&PreProcessor::loop, this};
+  msgLoopThread_ = std::thread{&PreProcessor::loop, this};
   LOG_INFO(logger(),
            KVLOG(numOfReplicas_,
                  numOfExternalClients,
@@ -147,11 +147,11 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
 }
 
 PreProcessor::~PreProcessor() {
-  done_ = true;
-  msg_cv_.notify_all();
+  msgLoopDone_ = true;
+  msgLoopSignal_.notify_all();
   cancelTimers();
   threadPool_.stop();
-  if (msg_proc_thread_.joinable()) msg_proc_thread_.join();
+  if (msgLoopThread_.joinable()) msgLoopThread_.join();
 }
 
 void PreProcessor::addTimers() {
@@ -271,6 +271,7 @@ void PreProcessor::updateAggregatorAndDumpMetrics() {
 }
 
 void PreProcessor::sendCancelPreProcessRequestMsg(const ClientPreProcessReqMsgUniquePtr &clientReqMsg,
+                                                  NodeIdType destId,
                                                   uint16_t reqOffsetInBatch,
                                                   uint64_t reqRetryId) {
   const auto clientId = clientReqMsg->clientProxyId();
@@ -286,7 +287,6 @@ void PreProcessor::sendCancelPreProcessRequestMsg(const ClientPreProcessReqMsgUn
                                         nullptr,
                                         clientReqMsg->getCid(),
                                         clientReqMsg->spanContext<ClientPreProcessReqMsgUniquePtr::element_type>());
-  const auto destId = clientReqMsg->senderId();
   SCOPED_MDC_CID(preProcessReqMsg->getCid());
   LOG_DEBUG(logger(), "Sending PreProcessRequestMsg with REQ_TYPE_CANCEL" << KVLOG(clientId, reqSeqNum, destId));
   sendMsg(preProcessReqMsg->body(), destId, preProcessReqMsg->type(), preProcessReqMsg->size());
@@ -314,6 +314,7 @@ void PreProcessor::sendRejectPreProcessReplyMsg(NodeIdType clientId,
 
 template <>
 void PreProcessor::onMessage<ClientPreProcessRequestMsg>(ClientPreProcessRequestMsg *msg) {
+  concord::diagnostics::TimeRecorder scoped_timer(*histograms_.onClientPreProcessRequestMsg);
   preProcessorMetrics_.preProcReqReceived.Get().Inc();
   ClientPreProcessReqMsgUniquePtr clientMsg(msg);
   const string &cid = clientMsg->getCid();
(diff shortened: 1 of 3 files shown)


bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -185,6 +189,14 @@ class PreProcessor {
   }
 
  private:
+  void loop();
+
+  boost::lockfree::spsc_queue<MessageBase *> msgs_{10000};
```
yuliasherman, line 194: Please define the const variable instead of a hard-coded number.

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -185,6 +189,14 @@ class PreProcessor {
   }
 
  private:
+  void loop();
```
yuliasherman, line 192: Consider giving a more meaningful name to this function.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -589,24 +593,58 @@ template <typename T>
...
+    int count = 0;
+    {
+      std::unique_lock<std::mutex> l(msgLock_);
+      while (!msgLoopDone_ && msgs_.empty()) msgLoopSignal_.wait(l);
+      if (msgLoopDone_) break;
+      count = msgs_.read_available();
+    }
+
+    while (!msgLoopDone_ && count--) {
+      auto m = msgs_.front();
```
yuliasherman, line 620: m -> msg

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -589,24 +593,58 @@ template <typename T>
...
+            onMessage<PreProcessReplyMsg>(static_cast<PreProcessReplyMsg *>(m));
+            break;
+          }
+          default:
+            LOG_ERROR(logger(), "Unknown message" << KVLOG(m->type()));
+        }
+
+      } else {
+        preProcessorMetrics_.preProcReqInvalid.Get().Inc();
+      }
```
yuliasherman, line 645: + delete msg

salieri11: I think it used internall after that (the same as before, because it is already copy of the original message). Deleting message here leads to crash.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -593,6 +593,10 @@ template <typename T>
 void PreProcessor::messageHandler(MessageBase *msg) {
   T *trueTypeObj = new T(msg);
   delete msg;
+  if (!msgs_.write_available()) {
+    LOG_ERROR(logger(), "PreProcessor queue is full, returning message");
+    incomingMsgsStorage_->pushExternalMsg(std::unique_ptr<MessageBase>(msg));
```
yuliasherman, line 598: You have already deleted the message in line 595 :-)

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -602,11 +606,15 @@ void PreProcessor::messageHandler<PreProcessReplyMsg>(MessageBase *msg) {
   PreProcessReplyMsg *trueTypeObj = new PreProcessReplyMsg(msg);
   trueTypeObj->setPreProcessorHistograms(&histograms_);
   delete msg;
+  if (!msgs_.write_available()) {
+    LOG_ERROR(logger(), "PreProcessor queue is full, returning message");
+    incomingMsgsStorage_->pushExternalMsg(std::unique_ptr<MessageBase>(msg));
```
yuliasherman, line 611: The same as before

yuliasherman (changes requested):

yuliasherman (approved):
LGTM. Thanks for fixing that.

bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -185,6 +189,15 @@ class PreProcessor {
...
 
  private:
+  const uint32_t MAX_MSGS = 10000;
+  void msgProcessingLoop();
+
+  boost::lockfree::spsc_queue<MessageBase *> msgs_{MAX_MSGS};
+  std::thread msgLoopThread_;
+  std::mutex msgLock_;
+  std::condition_variable msgLoopSignal_;
+  bool msgLoopDone_ = false;
```
andrewjstone, line 199: Doesn't this need to be atomic?

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -587,26 +591,71 @@ void PreProcessor::onMessage<PreProcessReplyMsg>(PreProcessReplyMsg *msg) {
...
+  msgs_.push(trueTypeObj);
+  msgLoopSignal_.notify_one();
+}
+
+void PreProcessor::msgProcessingLoop() {
+  while (!msgLoopDone_) {
+    int count = 0;
+    {
+      std::unique_lock<std::mutex> l(msgLock_);
+      while (!msgLoopDone_ && msgs_.empty()) msgLoopSignal_.wait(l);
```
andrewjstone, line 624: I believe there's a race condition here that could result in indefinite waiting. It's due to not holding the lock when `msgs_.push` is called. I only recognized this because I've made a similar mistake before.

The problem is that `msgs_.empty()` can be checked, then the push thread runs, and signal is triggered. Then the wait is called, which misses the signal. 

This isn't likely a huge problem here, as the next push will trigger the signal, but the code is still technically not correct. This is the problem with using a lock-free queue with a condition-variable.

There are 2 ways to fix this:
 1. Just use a regular queue and hold the lock.
 2. Add a timeout to the wait call and recheck the condition.

bftengine/src/preprocessor/PreProcessor.hpp
salieri11: Yes, good point!

bftengine/src/preprocessor/PreProcessor.cpp
andrewjstone: It's not the `empty` that is the problem, although as you pointed out it's not thread safe and should be replace.. The problem is the point in time between the while condition and the `wait` call. There's no way to guarantee that the signal isn't called from another thread in between those 2 points, resulting in missing the signal. 

With a lock around the push method, it wouldn't be possible to miss the signal because you would have to hold the lock to check empty. If you got the lock and it was empty, then you would wait, but the push could only happen after the wait when the lock was released.

But since you likely don't want an extra lock around the pusher, since you are purposefully using a lock free queue, you can just timeout and re-check periodically in case you miss a signal. In practice this will not happen frequently, as throughput will be high.

salieri11: I dont care about the lock around the pusher, the lock from the consumer is very quick, just to take the number of elements, and taking non locked mutex by the producer is not too bad. But in the sake of safety, I prefer to switch to regular queue. Will post an update to this branch.

PR #1423, opened 2021-04-29
Enable ignoring the primary in the matched part of bft replies

yontyon:
The read-only replica does not know who the primary is. However, when matching a reply we also look at the primary from the replica.
To enable reading from read-only replicas, we added another configuration option to the matcher that instructs him to ignore the primary. This way, only if the user gives a configuration that contains ro replicas, we ignore from the primary in the replies

Notice that in this case, the current primary of the client won't be affected

### DIFF
--- a/bftclient/src/matcher.cpp
+++ b/bftclient/src/matcher.cpp
@@ -15,7 +15,7 @@ namespace bft::client {
 
 std::optional<Match> Matcher::onReply(UnmatchedReply&& reply) {
   if (!valid(reply)) return std::nullopt;
-
+  if (config_.ignore_primary_) reply.metadata.primary = ReplicaId{0};
   auto key = MatchKey{reply.metadata, std::move(reply.data)};
   if (matches_[key].count(reply.rsi.from)) {
     if (matches_[key][reply.rsi.from] != reply.rsi.data) {
@@ -34,6 +34,7 @@ std::optional<Match> Matcher::match() {
     return match.second.size() == config_.quorum.wait_for;
   });
   if (result == matches_.end()) return std::nullopt;
+  primary_ = result->first.metadata.primary;
   return Match{Reply{result->first.data, std::move(result->second)}, result->first.metadata.primary};
 }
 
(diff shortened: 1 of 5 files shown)


bftclient/src/matcher.h
```diff
@@ -26,6 +26,8 @@ struct MatchConfig {
   // All quorums can be distilled into an MofN quorum.
   MofN quorum;
   uint64_t sequence_number;
+  bool ignore_primary_ =
```
andrewjstone, line 29: I think negative booleans are hard to reason about. Would you mind changing this to `include_primary_` and defaulting it to true. Otherwise it reads like 'don't ignore primary' which is kind of a confusing statement and less straightforward than something like 'don't include primary'.

bftclient/src/matcher.h
```diff
@@ -66,6 +68,11 @@ class Matcher {
 
   void clearReplies() { matches_.clear(); }
 
+  void getPrimary(std::optional<ReplicaId>& primary) {
```
andrewjstone, line 71: I think the code would read better by returning the value rather than passing a mutable reference as an out parameter. It also wouldn't rely on the default value being set for that parameter.

You would just return `std::nullopt` for `config_.ignore_primary_` otherwise return `primary_`. No need to check the value of primary_.

bftclient/src/matcher.cpp
```diff
@@ -15,7 +15,7 @@ namespace bft::client {
 
 std::optional<Match> Matcher::onReply(UnmatchedReply&& reply) {
   if (!valid(reply)) return std::nullopt;
-
+  if (config_.ignore_primary_) reply.metadata.primary = ReplicaId{0};
```
andrewjstone, line 18: I realize you are using the type of primary as is, but this is somewhat misleading.  ReplicaId{0} is completely arbitrary, and other code may assume that it actually matched as ReplicaId{0}. I don't think the value is used outside the matcher, but it may be worth changing the type of `reply.metadata.primary` to be a `std::optional<ReplicaId>`.

bftclient/src/matcher.h
yontyon: fixed

bftclient/src/matcher.h
yontyon: fixed

bftclient/src/matcher.cpp
yontyon: fixed

andrewjstone (approved):
Looks great. Thanks for the update @yontyon

PR #1425, opened 2021-04-29
Transaction signing apollo tests integration

glevkovich:
All Apollo tests pass when the feature is enabled/disable.
For now the feature is disabled  (CONCORD_BFT_CMAKE_TRANSACTION_SIGNING_ENABLED is FALSE).

Major changes by commits:
1) ReplicasInfo: fix bug and add functionality
2) SigManager: add Metrics counters and replicasInfo
3) ClientMsgs: add some comments for struct members
4) ClientsManager: fix bug
5) ReplicaImp,preprocessor: important modification for transaction signing
6) tests: fix bug, adjust other tests to all recent modifications
7) Makefile, CMakeLists, tests/apollo/CMakeLists: tx signing support


yontyon (commented):
LGTM!
However, due to the size of these PR and the variety of code components it touches, I would suggest that other team members review it before merging it

Thanks

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -447,16 +448,22 @@ void PreProcessor::onMessage<ClientBatchRequestMsg>(ClientBatchRequestMsg *msg)
                 << KVLOG(senderId, clientId, cid, clientBatch->numOfMessagesInBatch()));
   if (!checkClientBatchMsgCorrectness(clientBatch)) {
     preProcessorMetrics_.preProcReqIgnored.Get().Inc();
+    LOG_ERROR(logger(),
+              "ClientBatchPreProcessRequestMsg failed correctness check"
+                  << KVLOG(senderId, clientId, cid, clientBatch->numOfMessagesInBatch()));
```
yuliasherman, line 453: There are a lot of cases in which checkClientMsgCorrectness fails and they have specific logs. Not all of them are errors. Please remove this log as redundant.

bftengine/src/preprocessor/RequestProcessingState.hpp
```diff
@@ -96,6 +96,8 @@ class RequestProcessingState {
   const std::string cid_;
   const ReqId reqSeqNum_;
   const uint64_t entryTime_;
+  const std::string clientRequestSignature_;
+  const uint32_t clientRequestSignatureLen_;
```
yuliasherman, line 100: clientRequestSignature_ is a string now, why do you need clientRequestSignatureLen_?

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
```diff
@@ -12,6 +12,7 @@
 #include "PreProcessReplyMsg.hpp"
 #include "ReplicaConfig.hpp"
 #include "assertUtils.hpp"
+#include "SigManager.hpp"
```
yuliasherman, line 15: Is in use?

bftengine/src/preprocessor/RequestProcessingState.hpp
glevkovich: I was first thinking to have it intentionally explicit since all other messages hold the same sig/siglen.
Rethinking, I'll remove that..

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
glevkovich: Removed, thanks for pointing that out..

yontyon (approved):

PR #1428, opened 2021-04-29
Remove PassThroughTracker from Apollo tests

mdarade:
Removed PassThroughTracker which has redundant functionality as
that of history tracker. Also removed functionality to disable
linearizability check.


teoparvanov (approved):
LGTM, thanks Mahaveer! Before merging this, please first disable the test we talked about in Hermes. Otherwise it will fail when concord-bft gets bumped into Concord.

f-squirrel (approved):
LGTM, @mdarade!

PR #1430, opened 2021-04-30
Pass the received span context to next stage after PreProcess

bayarcher:
After PreProcess completes the work, transfer the span context received to the next stage.


f-squirrel (approved):
LGTM!

PR #1431, opened 2021-05-02
Improve noop mechanism on wedge command

yontyon:
The former implementation of the noop mechanism was as follows:
Once a primary replica has detected that we need to wedge on seq num X it sent empty prepreapres up to X.
This created several problems:
1. It violates the concurrency level - creating many slow path commits
2. No new blocks were written for these sequence numbers, this can damage the sync mechanism in case of a crash

In this PR we introduce a new mechanism:
1. Once the primary done executing a request, and it detects that we need to wedge on seqNum X, it uses the internal client to send a special noop reconfiguration request. This noop request is executed by the reconfiguration engine which means that it also creates a new block.
2. In addition, now, we send a new prepreapre (containing the noop) only after executing one - which means we don't violate the concurrency level principle.

In the future, we plan to add a verification step to validate this noop request. However, even now, it does not violate anything, if a malicious client wants to launch a DOS attack, it can also send a regular heavy computed requests (rather than noops)

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -581,14 +581,6 @@ void ReplicaImp::sendInternalNoopPrePrepareMsg(CommitPath firstPath) {
   startConsensusProcess(pp, isInternalNoop);
 }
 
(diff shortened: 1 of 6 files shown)


Liorgal28 (approved):
LGTM

PR #1432, opened 2021-05-03
Periodic keys generation and exchange

toly-kournik:


toly-kournik:
On-demand KeyExchange
    - CryptoManager to support multiple crypto systems accessed by a seqnum;
    - Hence removed key rotation logic. Instead old keys are garbage collected on checkpoint;
    - ExecutionRequest to include CID;
    - ISecureStore + hierarchiy deleted in favour of existing ISecretsManagerImpl;
    - KeyExchangeMgr
            - Moved to include/bftengine as it needs to be accessed by other components;
            - Holds replica's private keys collection per seqnum, as well as newly generated key pair;
            - Initialization logic changed to use RAII;
    - Serializable to has a default impl of getVersion() in order not to implement it subclasses if not needed;

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -17,8 +17,11 @@
 #include "threshsign/IThresholdVerifier.h"
 #include "ReplicaConfig.hpp"
 #include "IKeyExchanger.hpp"
+#include "Logger.hpp"
 
 namespace bftEngine {
+typedef std::int64_t SeqNum;                    // TODO [TK] redefinition
```
cloudnoize, line 23: should it be signed?

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -30,73 +33,134 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
-    init();
+  // onPrivateKeyExchange and onPublicKeyExchange callbacks for a given checkpoint may be called in a different order.
+  // Therefore the first called will create a CryptoSys
+  void onPrivateKeyExchange(const std::string& secretKey,
+                            const std::string& verificationKey,
+                            const SeqNum& sn) override {
+    LOG_INFO(logger(), KVLOG(sn, verificationKey));
+    auto sys = create(sn);
+    sys->sys_->updateKeys(secretKey, verificationKey);
+    sys->init();
```
cloudnoize, line 69: what is the meaning of ```sys```, with ```auto``` it's hard to understand what is this variable.
(also will eliminate the sys->sys_)

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -30,73 +33,134 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
+    LOG_INFO(logger(), "chckp: " << checkpoint);
+    clearOldKeys();
   }
 
  private:
-  CryptoManager(Cryptosystem* cryptoSys) : multiSigCryptoSystem_{cryptoSys} { init(); }
+  /**
+   *  Holds Cryptosystem, signers and verifiers per checkpoint
+   */
+  struct CryptoSys {
```
cloudnoize, line 90: can we rename it to a more meaningful name?

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -30,73 +33,134 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
+    }
+    ConcordAssert(false && "should never reach here");
+  }
+  // create CryptoSys for sn if still doesn't exist
+  std::shared_ptr<CryptoSys> create(const SeqNum& sn) {
+    // Cryptosystem for this sn will be activated upon reaching a second checkpoint from now
+    uint64_t chckp = sn / checkpointWindowSize + 2;
+    if (auto it = cryptoSystems_.find(chckp); it != cryptoSystems_.end()) return it->second;
+    // copy construct new Cryptosystem from a last one as we want it to include all the existing keys
+    Cryptosystem* cs = new Cryptosystem(*cryptoSystems_.rbegin()->second->sys_.get());
```
cloudnoize, line 136: nit: why not creating ```std::make_shared<CryptoSys>(new Cryptosystem(*cryptoSystems_.rbegin()->second->sys_.get()))```

toly-kournik: I pass raw pointer to a CryptoSustemWrapper(), which owns it by creating a sp

bftengine/include/bftengine/CryptoManager.hpp
toly-kournik: agree

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -76,181 +58,95 @@ std::string KeyExchangeManager::generateCid() {
...
   KeyExchangeMsg msg;
+  auto cid = generateCid();
   auto [prv, pub] = multiSigKeyHdlr_->generateMultisigKeyPair();
-  keysView_.keys().generatedPrivateKey = prv;
-  msg.key = pub;
+  private_keys_.key_data().generated.priv = prv;
+  private_keys_.key_data().generated.pub = pub;
+  private_keys_.key_data().generated.cid = cid;
+  private_keys_.key_data().generated.sn = sn;
+  private_keys_.save();
```
cloudnoize, line 126: Do we have an issue if for some reason we fail to send but save the file, in the prev imp I saved the file after sending, not sure

toly-kournik: In this case the system won't be able to generate and exchange keys for two checkpoints, but will continue to operate normally with the old keys.
In your case there could be a problem in case of failure to persist generated keys after sending them to consensus.

PR #1434, opened 2021-05-03
Fix SPSC queue usage by adding timeout

salieri11:
This MR fixes usage of spsc queue in the PreProcessor, by introducing timeout to wait when the queue is empty.
In the worst case, only first message will wait for the timeout. Under heavy loads, it will be probably negligible.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -606,15 +606,19 @@ void PreProcessor::messageHandler<PreProcessReplyMsg>(MessageBase *msg) {
 
 void PreProcessor::msgProcessingLoop() {
   while (!msgLoopDone_) {
-    int count = 0;
     {
       std::unique_lock<std::mutex> l(msgLock_);
-      while (!msgLoopDone_ && msgs_.empty()) msgLoopSignal_.wait(l);
+      while (!msgLoopDone_ && msgs_.empty()) {
+        msgLoopSignal_.wait_until(l, chrono::steady_clock::now() + std::chrono::milliseconds(WAIT_TIMEOUT_MILLI));
+      }
       if (msgLoopDone_) break;
-      count = msgs_.read_available();
+      if (!msgs_.read_available()) {
+        LOG_FATAL(logger(), "queue empty on wakeup");
+        ConcordAssert(false);
+      }
     }
 
-    while (!msgLoopDone_ && count--) {
+    while (!msgLoopDone_ && msgs_.read_available()) {
       auto msg = msgs_.front();
       msgs_.pop();
       if (validateMessage(msg)) {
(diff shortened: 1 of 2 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -606,15 +606,19 @@ void PreProcessor::messageHandler<PreProcessReplyMsg>(MessageBase *msg) {
...
       std::unique_lock<std::mutex> l(msgLock_);
-      while (!msgLoopDone_ && msgs_.empty()) msgLoopSignal_.wait(l);
+      while (!msgLoopDone_ && msgs_.empty()) {
+        msgLoopSignal_.wait_until(l, chrono::steady_clock::now() + std::chrono::milliseconds(WAIT_TIMEOUT_MILLI));
+      }
       if (msgLoopDone_) break;
-      count = msgs_.read_available();
+      if (!msgs_.read_available()) {
+        LOG_FATAL(logger(), "queue empty on wakeup");
+        ConcordAssert(false);
```
yuliasherman, line 617: Couldn't it be some sporadic wake-up? I would ignore it and continue.

salieri11: the condition to reach this point is that if the msgs is empty - from the loop. If it waits, wakes up for some reason - but if the queue is empy, it should go sleep again. To be on the safe side you are right, but maybe then we will hide some bug?

andrewjstone: I actually don't think you need the check at all. The code is pretty clear. It will loop until there is a message or an exit.  The only way this would be a problem is if multiple threads were pulling off the queue. I don't believe that's the case. 

But you should probably change the loop above to use read_available() instead of empty() for thread safety as you mentioned in a prior comment.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -606,15 +606,19 @@ void PreProcessor::messageHandler<PreProcessReplyMsg>(MessageBase *msg) {
 
 void PreProcessor::msgProcessingLoop() {
   while (!msgLoopDone_) {
-    int count = 0;
     {
       std::unique_lock<std::mutex> l(msgLock_);
-      while (!msgLoopDone_ && msgs_.empty()) msgLoopSignal_.wait(l);
+      while (!msgLoopDone_ && msgs_.empty()) {
```
yuliasherman, line 611: Don't you like to replace msgs_.empty() by !msgs_. read_available()

salieri11: Yes, its on the way

yuliasherman (approved):
LGTM

andrewjstone:
Thanks @salieri11. Solid fix!

PR #1438, opened 2021-05-04
Multi thread only on replicas

Liorgal28:
With the case of the client pool, we want multithreaded connections to be created only on replicas since if each client creates 8 threads that will create a resource problem.

### DIFF
--- a/communication/src/TlsTCPCommunication.cpp
+++ b/communication/src/TlsTCPCommunication.cpp
@@ -23,7 +23,10 @@ TlsTCPCommunication::TlsTCPCommunication(const TlsTcpConfig &config) : config_(c
   // Runners can actually support multiple principals. The Communication interface does not though. Currently we are
   // focused on backwards compatibility, and will use the future runner functionality to replace client thread pools.
   const auto configs = std::vector<TlsTcpConfig>{config};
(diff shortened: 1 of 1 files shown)


communication/src/TlsTCPCommunication.cpp
```diff
@@ -23,7 +23,10 @@ TlsTCPCommunication::TlsTCPCommunication(const TlsTcpConfig &config) : config_(c
   // Runners can actually support multiple principals. The Communication interface does not though. Currently we are
   // focused on backwards compatibility, and will use the future runner functionality to replace client thread pools.
   const auto configs = std::vector<TlsTcpConfig>{config};
-  runner_.reset(new tls::Runner(configs, NUM_THREADS));
+  if ((uint32_t)config.selfId > (uint32_t)config.maxServerId)
```
andrewjstone, line 26: You only need to cast one of these values. Also, you should really use `static_cast<uint32>` rather than c-style casts.

andrewjstone (approved):
Thanks for the important fix @Liorgal28. Just one minor suggestion from me.

communication/src/TlsTCPCommunication.cpp
Liorgal28: Done, thanks :)

PR #1443, opened 2021-05-05
Client proxy shutdown on concord termination

nkumar04:
Conenction objects for client proxies sould also be closed for graceful termination of concord

### DIFF
--- a/kvbc/src/ClientImp.cpp
+++ b/kvbc/src/ClientImp.cpp
@@ -46,7 +46,10 @@ Status ClientImp::start() {
 
 Status ClientImp::stop() {
   // TODO: implement
-  return Status::IllegalOperation("Not implemented");
+  if (0 != comm_->Stop()) {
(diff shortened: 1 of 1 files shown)


kvbc/src/ClientImp.cpp
```diff
@@ -46,7 +46,8 @@ Status ClientImp::start() {
 
 Status ClientImp::stop() {
   // TODO: implement
-  return Status::IllegalOperation("Not implemented");
+  comm_->Stop();
+  return Status::OK();
```
dartdart26, line 50: I see that ICommunicatio::stop() returns non-zero on failure. Maybe we should accommodate that return code here. What do you think?

nkumar04: agreed. updated the PR

andrewjstone (approved):
LGTM! Thanks for the fix @nkumar04

PR #1445, opened 2021-05-05
Add internal verification and signer to reconfiguration engine

yontyon:
There are several internal reconfiguration commands such as wedge noops (to bring the system to the next checkpoint) and key exchange.
We would like to have a verification mechanism for them as well.
In this PR we added verifiers to the reconfiguration engine and signers in the replicas.
They based on the already given RSA keys of the replicas.
In addition,  If a replica wants to mark a request as internal, it should put in the additinal_data field a specific prefix.


robem (approved):
Lgtm, thanks!

PR #1447, opened 2021-05-06
Docker update to version 0.23

yontyon:
Update concord-bft image to include ecdsa library in python.
We need ecdsa library in python in order to enable the apollo operator to issue signed reconfiguration requests


f-squirrel (changes requested):
Please change the commit summary so that it follows the pattern described in the Readme: Docker update to version <new version>. 
Please explain what is added and why in the commit message.

f-squirrel (approved):
Thank you @yontyon for addressing the comments!
LGTM!

PR #1448, opened 2021-05-06
Add ecdsa support in Apollo operator

yontyon:
In this PR we add ECDSA to the Apollo operator. 
This way, we will be able to verify the signatures and better simulate the real world scenario


Liorgal28 (approved):
LGTM

PR #1453, opened 2021-05-09
Increase max primary queue size

salieri11:
This MR increases default primary queue size to 1500.
For achieving maximum performance, we boost the concurrency both in terms of concurrent consensus operations and in terms of concurrent clients. Default value of 700 is quite low for maximim primary queue size, all messages beyond this size are just ignored.
This parameter should be moved to the configuration in the future.


yuliasherman (approved):
LGTM.

yuliasherman:
Don't we like to make this parameter configurable?

PR #1455, opened 2021-05-10
Enabling State Transfer in Separate thread by default in tester replica

ukg19:
We have the option to run state transfer mechanism in a separate thread from BFT message processing. Running ST in a separate thread has become an established way for the system in real deployments. So enabling option to run ST in separate thread.


glevkovich (approved):
LGTM!

teoparvanov (approved):
LGTM, thanks @ukg19 !

PR #1457, opened 2021-05-10
Add target replicas to hey exchange reconfiguration request

yontyon:
We add a list of target replicas to the key exchange reconfiguration message such that the operator will be able to trigger KE for only part of the replicas


toly-kournik (approved):
LGTM

PR #1461, opened 2021-05-12
Implement framework for n-f reconfiguration actions + specific implementation for n-f wedge

yontyon:
In this PR we present a framework for supporting n-f reconfiguration actions.
Until now, the reconfiguration engine has worked under the assumption that we have a n/n quorum. However, in many cases, this assumption does not hold.
To support n-f reconfiguration actions we add a dedicated reconfiguration SM that implements a callback for state transfer.
Once state transfer is completed, the call back is being called and we can change the reconfiguration state accordingly.
Notice, that the above is supported by the fact that we write every reconfiguration command to the blockchain.

In addition, in this PR we present an n-f wedge implementation.
To implement n-f wedge, we did the following:
1. A replica reports on a wedge on a stable checkpoint already (instead of n/n checkpoint)
2. In addition, once a replica is starting, it unwedge if necessary (this is a temporary solution until we implement a real unwedge request)
3. Thus, a replica that is in ST has two options: (a) it done with ST before the replicas have restarted. In this case, once ST is done, the data is in the reserved pages and the replica simply behaves like all other replicas. (b) The replica completed the ST after the other replicas have unwedged, in this case, we need to do nothing and simply continue with the quorum.

To validate the n-f wedge, we added an Apollo test for this case.
Note that it is still possible to wait for n/n wedge.

### DIFF
--- a/kvbc/include/ReplicaImp.h
+++ b/kvbc/include/ReplicaImp.h
@@ -9,6 +9,7 @@
 #include <string>
 #include <atomic>
 
+#include "st_reconfiguraion_sm.hpp"
 #include "OpenTracing.hpp"
 #include "categorization/kv_blockchain.h"
 #include "communication/ICommunication.hpp"
@@ -26,7 +27,6 @@
 #include "bftengine/DbMetadataStorage.hpp"
 #include "storage_factory_interface.h"
 #include "ControlStateManager.hpp"
-
 namespace concord::kvbc {
 
 class ReplicaImp : public IReplica,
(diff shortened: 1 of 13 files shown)


kvbc/include/ReplicaImp.h
```diff
@@ -176,6 +176,7 @@ class ReplicaImp : public IReplica,
   // secretsManager_ can be nullptr. This means that encrypted configuration is not enabled
   // and there is no instance of SecretsManagerEnc available
   const std::shared_ptr<concord::secretsmanager::ISecretsManagerImpl> secretsManager_;
+  std::unique_ptr<concord::kvbc::StReconfigurationHandler> stReconfigurationSM_;
```
toly-kournik, line 179: Is it in use yet?

yontyon: currently all the implementation is empty (but the callback is being called when ST is done)
This object exists only for the state transfer callback

toly-kournik (approved):
LGTM

PR #1462, opened 2021-05-12
Transaction signing integration fixes

glevkovich:
1. ClientRequestMsg: do not verify reconfiguration (operatore) requests
2. SigManager: transform assert explciit error logging
3. bft.py: redirect errs to devnull when gen transaction signing keys


yontyon (commented):
LGTM

yontyon (approved):

PR #1464, opened 2021-05-12
Reserved Pages convenience improvements

toly-kournik:
- Initialize ReservedPagesCLient only once instead of initializing every client with ReservedPages;
- ReservedPagesClient to implemet IReservedPages;
- Make res pages offset implicit making each RP client to use relative page id instead of explicitly adding its resPagesOffset()

### DIFF
--- a/bftengine/src/bftengine/BFTEngine.cpp
+++ b/bftengine/src/bftengine/BFTEngine.cpp
@@ -22,9 +22,12 @@
 #include "PreProcessor.hpp"
 #include "MsgReceiver.hpp"
 #include "RequestHandler.h"
+#include "ReservedPagesClient.hpp"
 #include <condition_variable>
 #include <memory>
 #include <mutex>
+
+bftEngine::IReservedPages *bftEngine::ReservedPagesClientBase::res_pages_ = nullptr;
 
 namespace bftEngine::impl {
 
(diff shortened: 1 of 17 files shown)


bftengine/src/bftengine/BFTEngine.cpp
```diff
@@ -22,10 +22,13 @@
 #include "PreProcessor.hpp"
 #include "MsgReceiver.hpp"
 #include "RequestHandler.h"
+#include "ReservedPagesClient.hpp"
 #include <condition_variable>
 #include <memory>
 #include <mutex>
 
+bftEngine::IReservedPages *bftEngine::ReservedPagesClientBase::res_pages_ = nullptr;
```
yontyon, line 30: Make sure it doesn't create problems with the product. A few weeks ago we had issues with global static variables and eventually, we forced to define them on the stack

bftengine/src/bftengine/ControlStateManager.cpp
```diff
@@ -89,7 +85,7 @@ void ControlStateManager::clearCheckpointToStopAt() {
   std::ostringstream outStream;
   concord::serialize::Serializable::serialize(outStream, tmpPage);
   auto data = outStream.str();
-  reserved_pages_->saveReservedPage(resPageOffset(), data.size(), data.data());
+  saveReservedPage(0, data.size(), data.data());
```
yontyon, line 88: Why do we have a hardcoded 0?

yontyon (approved):
LGTM!
Can you have a dummy MR with this branch before merging it? 
I just to be sure that we don't run into static variables problem in the product again

bftengine/src/bftengine/BFTEngine.cpp
toly-kournik: it's a class static variable

bftengine/src/bftengine/ControlStateManager.cpp
toly-kournik: It's a relative offset, meaning page 0

PR #1465, opened 2021-05-13
Copy openssl_crypto_utils source and header files to concord-bft

shgandhi:
This commit copies openssl_crypto_utils source and header files to
concord-bft from close-source, with the objective of ultimately
removing these files from close-source codebase. The key value hashes
in thin replica server are computed using `ComputeSHA256Hash`
function from `openssl_crypto_utils`. Since the dependency can't
be refactored out of the TRS, for the move of TRS to concord-bft,
the move of openssl_crypto_utils to opensource becomes a pre-requisite.


robem (approved):
Lgtm. Do we have any tests for this? Otherwise, this might be a good task to add.

shgandhi:
Thanks for catching that @robem. I have added the corresponding tests.

robem (approved):
Thanks!

upshaw-alex:
You mention you are adding `openssl_crypto_utils` to Concord-BFT because we need `ComputeSHA256Hash`, but I see you are also adding the asymmetric-key stuff in `openssly_crypto_utils` (i.e. `AsymmetricPrivateKey`, `AsymmetricPublicKey`, `kPermittedAsymmetricCryptoSchemes`, `generateAsymmetricCryptoKeyPair`, `deserializePrivateKey`, `deserializePrivateKeyFromPem`, `deserializePrivateKeyFromPemString`, `deserializePublicKeyFromPem`, and `deserializePublicKey`) to Concord-BFT with this PR. Do we also need that asymmetric-key-related code for anything in Concord-BFT at the moment?

shgandhi:
@upshaw-alex no we don't need any asymmetric-key related code in concord-BFT at the moment. The reason I moved the `openssl_crypto_utils` in its entirety was to avoid splitting the utility between `vmwathena` and `concord-bft`. At the moment I don't see anything problematic in  moving the utility to `concord-bft`, but I may have missed something, and would be happy to reconsider if that is the case.

PR #1467, opened 2021-05-13
Clear output in versioned cat multiGetLatestVersion

dartdart26:
The VersionedKeyValueCategory::multiGetLatestVersion() method accepts a
vector as an output parameter. If the vector already contains data,
adding to it will be the wrong thing to do as it will contain more data
at the end than requested. Therefore, clear it before adding.

Add unit tests for other multi* methods in all categories, showing
that reusing the same vectors without being cleared by callers works
properly.


andrewjstone (approved):
LGTM. Thanks @dartdart26

PR #1471, opened 2021-05-13
On-demand KeyExchange

toly-kournik:
- CryptoManager to support multiple crypto systems accessed by a seqnum;
- Hence removed key rotation logic. Instead old keys are garbage collected on checkpoint;
- ExecutionRequest to include CID;
- ISecureStore + hierarchiy deleted in favour of existing ISecretsManagerImpl;
- KeyExchangeMgr
-- Moved to include/bftengine as it needs to be accessed by other components;
-- Holds replica's private keys collection per seqnum, as well as newly generated key pair;
-- Initialization logic changed to use RAII;
- Serializable to has a default impl of getVersion() in order not to implement it subclasses if not needed;


bftengine/tests/keyManager/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
...
-    GTest::Main
-    corebft)
+#find_package(GTest REQUIRED)
+#
+#add_executable(KeyManager_test KeyManager_test.cpp )
+#add_test(KeyManager_test KeyManager_test)
+#
+#target_link_libraries(KeyManager_test PUBLIC
+#    GTest::Main
+#    corebft)
```
yontyon, line 8: Why not removing this?

yontyon (approved):
LGTM!

cloudnoize (approved):
Looks good thanks @toly-kournik

bftengine/tests/keyManager/CMakeLists.txt
toly-kournik: I still need to change the unittest

PR #1472, opened 2021-05-13
Fix inactive window catchup up with gap

HristoStaykov:
Fix for searching in Inactive Window on Status exchange and Apollo test recreating the problematic situation.
Replicas helping from Inactive Window can only send information up to the end of the peer's Active Working Window.

### DIFF
--- a/tests/apollo/test_skvbc_chaotic_startup.py
+++ b/tests/apollo/test_skvbc_chaotic_startup.py
@@ -53,6 +53,94 @@ class SkvbcChaoticStartupTest(unittest.TestCase):
     @unittest.skipIf(environ.get('BUILD_COMM_TCP_TLS', "").lower() == "true", "Unstable on CI (TCP/TLS only)")
     @with_trio
     @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_inactive_window_catchup_up_to_gap(self, bft_network):
+        """
+        In this test we check the catchup from Inactive Window when we have a gap related to the Peers.
+        The situation can happen if the catching up Replica's last Stable SeqNo is 3 Checkpoints behind its Peers, but
+        its Last Executed is only 2 Checkpoints behind.
+        Steps to recreate:
+        1) Start all replicas.
+        2) Isolate 1 Replica from all but the Primary. We will call it Late Replica.
+        3) Advance all replicas beyond the first Stable Checkpoint. The Late Replica won't be able to collect a
+           Stable Checkpoint.
+        4) Stop the Late Replica and advance all others 2 more Checkpoints.
+        5) Start the late Replica and verify it catches up to the end of its Working Window from the Inactive Windows of
+           its Peers.
+        """
+
+        late_replica = 1
+        primary = 0
+
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        first_stable_checkpoint_to_reach = 1
+        checkpoints_to_advance_after_first = 2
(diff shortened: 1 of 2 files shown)


tests/apollo/test_skvbc_chaotic_startup.py
```diff
@@ -50,6 +50,94 @@ class SkvbcChaoticStartupTest(unittest.TestCase):
...
+            await bft_network.wait_for_replicas_to_collect_stable_checkpoint(
+                bft_network.all_replicas(without={late_replica}),
+                first_stable_checkpoint_to_reach + checkpoints_to_advance_after_first)
+
+            bft_network.start_replica(late_replica)
+            with trio.fail_after(seconds=30):
+            
+                late_replica_catch_up = False
+                while not late_replica_catch_up:
+                    for replica_id in bft_network.get_live_replicas():
```
andrewjstone, line 131: Why loop over all replicas? Why not just get metrics from `late_replica` ?

andrewjstone (approved):
@HristoStaykov 
First off, great test. It's truly amazing to see how far apollo has come in writing these types of tests.

If I understand correctly, the problem was that the replica sending the PrePrepares was sending messages outside the window of the requesting replica, right? This occurs because the last stable checkpoint is outside the window of the requesting replica which missed 3 stable checkpoints.

If this is correct, the code looks correct and you are good to merge.

If you can though, can you please modify the commit message to explain the problem more clearly: i.e. Messages sent outside the window are ignored by the requester and may cause state transfer. I think you are just more close to the protocol code now and so the simple explanations make sense to you, but for those of us that haven't looked at this stuff in a while the more information about the exact problem being fixed the better. Thanks!

KTeofanidis (approved):
LGTM!

teoparvanov (approved):
LGTM!

tests/apollo/test_skvbc_chaotic_startup.py
HristoStaykov: It is sufficient to check only the late_replica, indeed. During the analysis it was convenient for me to log the status of all replicas though, this is why I decided to leave the additional logs in case of a test failure to have more thorough information about the system.

HristoStaykov:
> @HristoStaykov
> First off, great test. It's truly amazing to see how far apollo has come in writing these types of tests.
> 
> If I understand correctly, the problem was that the replica sending the PrePrepares was sending messages outside the window of the requesting replica, right? This occurs because the last stable checkpoint is outside the window of the requesting replica which missed 3 stable checkpoints.
> 
> If this is correct, the code looks correct and you are good to merge.
> 
> If you can though, can you please modify the commit message to explain the problem more clearly: i.e. Messages sent outside the window are ignored by the requester and may cause state transfer. I think you are just more close to the protocol code now and so the simple explanations make sense to you, but for those of us that haven't looked at this stuff in a while the more information about the exact problem being fixed the better. Thanks!

Yes this is exactly the situation, we can only help up to the end of the requesting replica's Working Window. If we go beyond, which was the case, we assert, because in the ReplicaStatusMsg there is a bit mask specifying for each SeqNo of the requestor's Working Window whether it has a PrePrepareMsg for it in order for the Primary not to resend unnecessary msg-s to honest Replicas. Before the Primary sends the PrePrepareMsg, he checks if the peer replica indicated that it is missing, and if the Primary checks beyond the Working Window of the peer we have recreation of the problem.

tests/apollo/test_skvbc_chaotic_startup.py
andrewjstone: Makes sense. Thanks!

PR #1475, opened 2021-05-14
Add view change recovery test

StefanPalashev:
Add a test that checks the recovery of fast path of replicas after they have
initiated a view change.

### DIFF
--- a/tests/apollo/test_skvbc_view_change.py
+++ b/tests/apollo/test_skvbc_view_change.py
@@ -452,6 +452,54 @@ class SkvbcViewChangeTest(unittest.TestCase):
         self.assertEqual(initial_primary, current_primary,
                          "Make sure we are still on the initial view.")
 
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: c == 0, rotate_keys=True)
+    @verify_linearizability()
+    async def test_recovering_of_replicas_with_initiated_view_change(self, bft_network, tracker):
+        """
+        The goal of this test is to validate fast path's recovery after stopping all replicas during an ongoing view change.
+        The test is valid only when c = 0.
+        Initially N - (F + 1) replicas are started. In order to initiate a view change (because of 3f + 2 * 0 + 1 - f - 1 = 2f),
+        the clients run concurrent operations, which trigger the view change, however it will not be completed (because of 2f < 2f + 1).
+
+        1) Given a BFT network start N - (F + 1) replicas.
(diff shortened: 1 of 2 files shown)


tests/apollo/test_skvbc_view_change.py
```diff
@@ -452,6 +452,52 @@ async def write_req():
...
                          "Make sure we are still on the initial view.")
 
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: c == 0, rotate_keys=True)
+    @verify_linearizability()
+    async def test_recovering_of_replicas_with_initiated_view_change(self, bft_network, tracker):
+        """
+        The goal of this test is to validate recovery to fast path after stopping all replicas during view change.
+        Test is valid only when c = 0.
+        To trigger view change we start N - (F+1) replicas. This causes fast view change because 3f+2*0+1-f-1=2f
```
HristoStaykov, line 462: "This causes fast view change" -> "This causes View Change to be initiated, but will not be completed"

HristoStaykov (approved):
LGTM, thanks, @StefanPalashev ! Just a minor description correction.

tests/apollo/util/bft.py
```diff
@@ -1321,4 +1321,29 @@ async def the_replica_to_ask_for_view_change():
...
+                    for replica_id in replicas_ids:
+                        nursery.start_soon(self.wait_for_replica_to_reach_view, replica_id, expected_view)
+            
+    async def wait_for_replica_to_reach_view(self, replica_id, expected_view):
+        """
+        Wait for a single replica to reach the expected view.
+        """
+        with log.start_action(action_type="wait_for_replica_to_reach_view", replica=replica_id, expected_view=expected_view) as action:
+            async def expected_view_to_be_reached():
+                key = ['replica', 'Gauges', 'view']
```
teoparvanov, line 1342: Are we sure this metric is incremented when the replica does `MoveToHigherView`?

teoparvanov (approved):
LGTM, thanks!

tests/apollo/util/bft.py
StefanPalashev: Yes, this is the correct metric. The following LOC (Currently 2431) in `MoveToHigherView`'s body sets the value of the metric:
`metric_view_.Get().Set(nextView);`

tests/apollo/test_skvbc_view_change.py
StefanPalashev: The description has been corrected. Thanks!

PR #1477, opened 2021-05-18
Limit the number of non-synchronised blocks to be deleted from the state

yuliasherman:
Limit it to maxNumOfRequestsInBatch instead of 1, as we can have as many blocks with the same seqNum as requests pass the consensus (without block accumulation).

### DIFF
--- a/kvbc/src/ReplicaImp.cpp
+++ b/kvbc/src/ReplicaImp.cpp
@@ -80,8 +80,15 @@ void ReplicaImp::createReplicaAndSyncState() {
   LOG_INFO(logger, KVLOG(lastExecutedSeqNum));
   if (!replicaConfig_.isReadOnly && !m_stateTransfer->isCollectingState()) {
     try {
-      uint64_t removedBlocksNum = replicaStateSync_->execute(logger, *m_kvBlockchain, lastExecutedSeqNum);
-      LOG_INFO(logger, KVLOG(lastExecutedSeqNum, removedBlocksNum, getLastBlockNum(), getLastReachableBlockNum()));
+      const auto maxNumOfBlocksToDelete = replicaConfig_.maxNumOfRequestsInBatch;
+      const auto removedBlocksNum =
+          replicaStateSync_->execute(logger, *m_kvBlockchain, lastExecutedSeqNum, maxNumOfBlocksToDelete);
+      LOG_INFO(logger,
+               KVLOG(lastExecutedSeqNum,
+                     removedBlocksNum,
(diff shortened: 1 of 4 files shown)


kvbc/src/ReplicaImp.cpp
```diff
@@ -80,8 +80,15 @@ void ReplicaImp::createReplicaAndSyncState() {
   LOG_INFO(logger, KVLOG(lastExecutedSeqNum));
   if (!replicaConfig_.isReadOnly && !m_stateTransfer->isCollectingState()) {
     try {
-      uint64_t removedBlocksNum = replicaStateSync_->execute(logger, *m_kvBlockchain, lastExecutedSeqNum);
-      LOG_INFO(logger, KVLOG(lastExecutedSeqNum, removedBlocksNum, getLastBlockNum(), getLastReachableBlockNum()));
+      uint32_t maxNumOfBlocksToDelete = replicaConfig_.maxNumOfRequestsInBatch;
+      uint64_t removedBlocksNum =
```
teoparvanov, line 84: nit - `const auto`?

kvbc/src/replica_state_sync_imp.cpp
```diff
@@ -45,8 +46,8 @@ uint64_t ReplicaStateSyncImp::execute(logging::Logger& logger,
     }
     // SBFT State Metadata is not in sync with the Blockchain State.
     // Remove blocks which sequence number is greater than lastExecutedSeqNum.
-    if (removedBlocksNum == 1) {
-      std::string error = " Detected more than one block needs to be deleted from the blockchain - unsupported";
+    if (removedBlocksNum >= maxNumOfBlocksToDelete) {
```
teoparvanov, line 49: shouldn't we have strict inequality `>` here?

teoparvanov (approved):
LGTM, thanks Yulia!

PS: some minor comments/questions below.

kvbc/src/replica_state_sync_imp.cpp
yuliasherman: We are checking how many blocks we have ALREADY removed, so it should be not greater than maxNumOfBlocksToDelete.

kvbc/src/ReplicaImp.cpp
yuliasherman: ok

kvbc/src/replica_state_sync_imp.cpp
teoparvanov: thanks for the clarification!

PR #1479, opened 2021-05-19
Improve and refactor the reconfiguration handler

yontyon:
In this PR we present a couple of refactoring to the reconfiguration mechanism:
1. We united the IPruningHandler and the IReconifugrationHandler. Now we have only IReconfigurationHandler
2. The above means that pruning handlers are now considered as reconfiguration handlers
3. We simplified the dispatcher code to use visitor instead of std::holds_alternative
4. Each reconfiguration Handler handles only the requests it can verify. This allows us to have more flexibility with the responsibilities chain of reconfiguration
5. We added a basic bftReconfigurationHandler that is responsible for verifying the operator requests
6. We separated the internal reconfiguration handler (that is now responsible for writing the wedge noops to the blockchain) from the regular reconfiguration handlers, giving it its own validation mechainsm


reconfiguration/include/reconfiguration/ireconfiguration.hpp
```diff
@@ -27,51 +27,43 @@ class IReconfigurationHandler {
   // Message handlers
```
toly-kournik, line 27: In order not to have trivial implementations in subclasses, since not all the handlers handle all commands, I would not make these methods pure virtual, but rather added a trivial impl `{return true;}`

yontyon: fixed

toly-kournik (approved):
LVGTM!!!

yontyon:
> Is it a valid situation? Consider assert.

Yes, it is. It simply means that reconfiguration requests won't be executed. The bft engine is still valid and works for everything else

PR #1481, opened 2021-05-19
Extend the timeout of VC with f replicas down

StefanPalashev:
This PR extends the timeout period of the test
"test_single_vc_with_f_replicas_down".

* the timeout is extended to 45 seconds to ensure that the replicas have
enough time to agree to the same view.
* the test utilizes the new generic wait_for method.


HristoStaykov (approved):
LGTM, thanks @StefanPalashev.

PR #1483, opened 2021-05-20
Initialize secrets manager to SecretsManagerPlain in case none is pro…

toly-kournik:
…vided.

This is needed as there are modules like KeyExchangeManager which rely on secrets manager's existence.


yontyon (approved):
LGTM!

PR #1485, opened 2021-05-20
Allocate preexecution result buffers on first use

KTeofanidis:
Under certain configuration we have gaps in the sequence of client_ids. In those cases we will allocate unnecessary buffers that are never used. We can avoid that by allocating the buffers on demand.


Liorgal28 (approved):
Great change, thanks.

yuliasherman (approved):
LGTM

teoparvanov:
After discussion with @yuliasherman , closing in favor of a simpler approach:
https://github.com/vmware/concord-bft/pull/1492/

PR #1488, opened 2021-05-21
State transfer in separate thread

ukg19:
Enabling State Transfer in Separate thread. Mechanism for state transfer is already there, just turning on the flag.

### DIFF
--- a/kvbc/src/ReplicaImp.cpp
+++ b/kvbc/src/ReplicaImp.cpp
@@ -228,7 +228,7 @@ ReplicaImp::ReplicaImp(ICommunication *comm,
     replicaConfig_.get<uint32_t>("concord.bft.st.sourceReplicaReplacementTimeoutMs", 15000),
     replicaConfig_.get<uint32_t>("concord.bft.st.fetchRetransmissionTimeoutMs", 1000),
     replicaConfig_.get<uint32_t>("concord.bft.st.metricsDumpIntervalSec", 5),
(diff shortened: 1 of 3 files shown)


kvbc/src/ReplicaImp.cpp
```diff
@@ -228,7 +228,7 @@ ReplicaImp::ReplicaImp(ICommunication *comm,
     replicaConfig_.get<uint32_t>("concord.bft.st.sourceReplicaReplacementTimeoutMs", 15000),
     replicaConfig_.get<uint32_t>("concord.bft.st.fetchRetransmissionTimeoutMs", 1000),
     replicaConfig_.get<uint32_t>("concord.bft.st.metricsDumpIntervalSec", 5),
-    replicaConfig_.get("concord.bft.st.runInSeparateThread", replicaConfig_.isReadOnly),
+    replicaConfig_.get<bool>("concord.bft.st.runInSeparateThread", true),
```
tdimitrov, line 231: I prefer not to change the default behaviour of ReplicaImp. I'd keep replicaConfig_.isReadOnly here.

Handle the change in tests/simpleKVBC/TesterReplica/main.cpp. ReplicaImp is initialised in run_replica(). Replica config is generated in the beginning of the function:
`  const auto setup = TestSetup::ParseArgs(argc, argv);
`
Add new command line parameter and set its default value to true.

ukg19: Makes sense. Resolving as per comments.

tdimitrov (approved):

PR #1490, opened 2021-05-22
CMake: Fix package name warnings

robem:
With the recent update of CMake we now see warnings for mismatching package
names (upper vs lower case). This change aligns the names and thereby keeps
stdout clean.


glevkovich (approved):
LGTM, thanks!

PR #1495, opened 2021-05-24
TS: Add timestamp to ClientRequestMsg

f-squirrel:
Changes:
* Added flag `TIME_SERVICE_FLAG = 0x40` to distinguish `ClientRequestMsg`
with timestamp from others
* Added serialize/deserialize methods for integral types (do not take
        endiannes into consideration)
* Unit tests for ClientRequestMsg

### DIFF
--- a/util/include/serialize.hpp
+++ b/util/include/serialize.hpp
@@ -1,0 +1,50 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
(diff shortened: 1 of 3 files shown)


util/include/serialize.hpp
```diff
@@ -0,0 +1,50 @@
...
+template <typename T>
+T deserialize(const char* begin, const char* end) {
+  static_assert(std::is_integral_v<T>);
+  ConcordAssertEQ((end - begin), sizeof(T));
+  auto value = T{};
+  std::memcpy(&value, begin, sizeof(T));
+  return value;
+}
+
+}  // namespace concord::util
```
yuliasherman, line 50: Couldn't existing "Serializable.h" be used for this purpose?

f-squirrel: It seemed a bit overhead for me because it serializes/deserializes to/from `iosteam` while `ClientRequestMsg` uses raw pointers/strings.

yuliasherman: OK

yuliasherman (approved):
LGTM

PR #1497, opened 2021-05-24
Add addRemove status command and apollo test

nkumar04:
*Add new reconfiguration concord message to handle addRemove status response
*Add a handler to read addRemove configuration from storage
*Add apollo test for addRemove command and status

### DIFF
--- a/kvbc/include/reconfiguration_add_block_handler.hpp
+++ b/kvbc/include/reconfiguration_add_block_handler.hpp
@@ -17,12 +17,13 @@
 #include "hex_tools.h"
 #include "block_metadata.hpp"
 #include "kvbc_key_types.hpp"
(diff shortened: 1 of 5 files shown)


kvbc/include/reconfiguration_add_block_handler.hpp
```diff
@@ -17,12 +17,13 @@
 #include "hex_tools.h"
 #include "block_metadata.hpp"
 #include "kvbc_key_types.hpp"
+#include "st_reconfiguraion_sm.hpp"
```
yontyon, line 20: Why do we need this include?

yontyon (approved):
LGTM!
I have only a small comment
Thanks!

PR #1498, opened 2021-05-24
Remove askForAnotherStateTransfer optimization.

toly-kournik:
This scenario happens rarely causing startCollectingState() to be called from onTransferringCompleteImp() which is called at the end of ST.
In this case the state is still "fetching" while startCollectingState() expects a "not fetching" state and asserts.
ST will resume in any case using a regular Checkpoints messages mechanism.

(Other way to solve this problem was to send an internal message to itself with request for another ST)


yontyon (approved):
LGTM!

PR #1500, opened 2021-05-25
More specific histograms for IncomingMsgsStorage

andrewjstone:
Instead of measuring the overall time of `getMsgForProcessing`, which
doesn't give us enough information, and includes internal queue time as well as
queue waiting and swapping, we now measure lock time and wait time
specifically.

This commit also removes unnecessary callbacks in TLS code, as they are
no longer used upstream. Eventually we will remove support for status
callbacks altogether.


yuliasherman (approved):
LGTM

PR #1502, opened 2021-05-26
Decrease TLS threads

Liorgal28:
Changed the default from 8 to 2.
In addition i've set the concurrency level metric to be updated in run time.


yuliasherman (approved):
LGTM

PR #1504, opened 2021-05-26
Return valid error message for addRemoveStatus

nkumar04:
AddRemove status command should return a valid error message if key is not found in the block chain


yontyon (approved):
LGTM!
Thanks

PR #1505, opened 2021-05-26
TS: Add Reserved Pages Client for Time Service

f-squirrel:
This commit introduces a reserved pages client for time service which is used to store the last used timestamp.

Changes:
* TimeServiceResPageClient - the client
* Unit tests


bftengine/src/bftengine/TimeServiceResPageClient.cpp
```diff
@@ -0,0 +1,40 @@
...
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the sub-component's license, as noted in the LICENSE
+// file.
+
+#include "TimeServiceResPageClient.hpp"
+
+namespace {
+constexpr auto RESERVED_PAGE_ID = uint32_t{0};
```
yuliasherman, line 15: Funny, you could define it just as constexpr uint32_t RESERVED_PAGE_ID = 0;

yuliasherman (approved):
LGTM

bftengine/src/bftengine/TimeServiceResPageClient.cpp
f-squirrel: @yuliasherman, but of course!
I kinda like the AAA (almost always auto) approach introduced by Herb Sutter:
https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/ :)

f-squirrel:
Update: changed the inheritance to private because I want to isolate the low-level methods related to reserved pages client.

PR #1506, opened 2021-05-26
Wrap the calls of map::at() in a try-catch block

StefanPalashev:


f-squirrel:
Well, now I see where the PR comes from.

I think that wrapping map::at calls with a try-except block is not a good practice:
If the flow expects the keys to be absent, then we need to look for a value, and in case of its absence, log/throw an exception with a meaningful message.

If the flow does not expect the keys to be absent, then leave the `out_of_range` exception for higher-level methods.

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -54,8 +54,18 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
+    auto log = logger();
+    LOG_INFO(log, "Generating new multisig key pair");
+
+    try {
+      auto cryptoSystem = cryptoSystems_.at(0);
+      return cryptoSystem->cryptosys_->generateNewKeyPair();
+    }
+    catch(const std::out_of_range& e) {
+      LOG_FATAL(log, e.what());
+      LOG_FATAL(log, "cryptoSystems_.at(0) has failed!");
```
yuliasherman, line 66: Actually, you can combine those two in one log: LOG_FATAL(log, "cryptoSystems_.at(0) has failed! " << e.what);

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -82,7 +82,7 @@ class ClientsManager : public ResPagesClient<ClientsManager, 0> {
   // General
   static uint32_t reservedPagesPerClient(const uint32_t& sizeOfReservedPage, const uint32_t& maxReplySize);
   int getIndexOfClient(const NodeIdType& id) const;
-
+
```
yuliasherman, line 85: I guess you touched this file occasionally?

bftengine/src/bftengine/KeyStore.cpp
```diff
@@ -47,8 +47,19 @@ std::optional<ClusterKeyStore::PublicKeys> ClusterKeyStore::loadReplicaKeyStoreF
 }
 
 void ClusterKeyStore::saveReplicaKeyStoreToReserevedPages(const uint16_t& repID) {
```
yuliasherman, line 49: The function name is misleading.

communication/src/TlsConnectionManager.cpp
```diff
@@ -128,6 +128,13 @@ void ConnectionManager::handleSend(const NodeNum destination, std::shared_ptr<Ou
   if (it != connections_.end()) {
     it->second->send(std::move(msg));
     status_->total_messages_sent++;
+    if (config_.statusCallback && isReplica() && (status_->total_messages_sent % 1000 == 1)) {
```
yuliasherman, line 131: How does this change relate to .at() issue?

kvbc/test/sparse_merkle/test_db.h
```diff
@@ -34,10 +34,19 @@ class TestDB : public concord::kvbc::sparse_merkle::IDBReader {
       return BatchedInternalNode();
     }
     auto root_key = InternalNodeKey::root(latest_version_);
-    return internal_nodes_.at(root_key);
+    return get_internal(root_key);
   }
 
-  BatchedInternalNode get_internal(const InternalNodeKey& key) const override { return internal_nodes_.at(key); }
```
yuliasherman, line 40: We don't need to do this change for tests

bftengine/src/bftengine/ClientsManager.hpp
StefanPalashev: I was about to make a new function that had to do the same thing as the one above. It seems that I've left a blank space unintentionally.

bftengine/include/bftengine/CryptoManager.hpp
StefanPalashev: Thanks for pointing that out, I'll make the necessary changes.

teoparvanov (approved):
LTGM, thanks @StefanPalashev !

PR #1507, opened 2021-05-27
Add perf logs + histograms to find wait time

andrewjstone:
Add trace logs for messages being put on and pulled off the
`InternalMsgsQueue` and log messages relating to the queue condition
variable.  The goal here is to see what message is received after
waiting on a condition variable. We are attempting to find what message
is being delayed if there is nothing in the external queue for a full
20ms (default condition variablel timeout).

Also, it was noticed on the primary that the 99th+ percentiles of the
`evaluate_timers` histogram in test runs was taking over 30ms. This time
correlates almost exactly with the wait time for a `PrePrepare` to be
received on a backup after no messages were received for a while. This
indicates that the primary sending of PrePrepares were getting delayed
due to a timer callback. Therefore a few new histograms to measure timer
callbacks were added. It turns out that `onBatchFlushTimer` matches this
time, and all percentages are pretty expensive. We need to see if we can
take this out of the hot path.

This change just adds the tools to debug this stuff. It doesn't add the
fix.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -644,6 +644,7 @@ void BCStateTran::startCollectingState() {
 // this function can be executed in context of another thread.
 void BCStateTran::onTimerImp() {
   if (!running_) return;
+  concord::diagnostics::TimeRecorder scoped_timer(*histograms_.on_timer);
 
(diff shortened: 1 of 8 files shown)


yuliasherman (approved):
LGTM

Liorgal28 (approved):
Looks good, thank you

PR #1509, opened 2021-05-27
TS: Add TimeServiceManager

f-squirrel:
This commit introduces TimeServiceManager, a class implementing the following functionality:
* `compareAndSwap` of the new and old timestamps
* `createClientRequestMsg` creates a corresponding `ClientRequestMsg`
* `isPrimarysTimeWithinBounds` check the time in the `ClientRequestMsg` against soft and hard limits from the configuration

Additional changes:
* template specializations for chrono types
* unit tests


bftengine/src/bftengine/TimeServiceManager.hpp
```diff
@@ -0,0 +1,98 @@
...
+               "Current primary's time reached soft limit, please synchronize local clocks! "
+                   << "Primary's time: " << t.count() << ", local time: " << now.count() << ", time limits: +/-"
+                   << config_.timeServiceSoftLimitMillis.count() << ". Time is presented as ms since epoch");
+    }
+    return true;
+  }
+
+ private:
+  TimeServiceResPageClient client_;
+  const ReplicaConfig& config_;
```
yuliasherman, line 96: ReplicaConfig is a singleton now, no need to pass.

bftengine/src/bftengine/TimeServiceManager.hpp
```diff
@@ -0,0 +1,98 @@
...
+  // if this is the case, returns reserved pages time + epsilon
+  // otherwise, returns the new time
+  [[nodiscard]] ConsensusTime compareAndSwap(ConsensusTime new_time) {
+    auto last_timestamp = client_.getLastTimestamp();
+    if (new_time > last_timestamp) {
+      client_.setLastTimestamp(new_time);
+      return new_time;
+    }
+
+    LOG_INFO(TS_MNGR,
```
yuliasherman, line 44: Should be at DEBUG level IMHO.

yuliasherman (approved):
LGTM

bftengine/src/bftengine/TimeServiceManager.hpp
f-squirrel: Discussed offline: the situation is not supposed to happen very frequently, if it does happen then action should be taken.

bftengine/src/bftengine/TimeServiceManager.hpp
f-squirrel: Done

PR #1510, opened 2021-05-28
Persist all Checkpoint msgs that serve as proof for Replica's lastStableSeqNum

HristoStaykov:
After adding signatures to checkpoint messages in order for Replicas to be
able to forward them we need to also persist those messages. This way after
restart we still will be able to prove to the Peers that our lastStableSeqNum
really is stable.

### DIFF
--- a/bftengine/src/bftengine/PersistentStorageImp.cpp
+++ b/bftengine/src/bftengine/PersistentStorageImp.cpp
@@ -23,8 +23,8 @@ namespace impl {
 
 const string METADATA_PARAMS_VERSION = "1.1";
 
-PersistentStorageImp::PersistentStorageImp(uint16_t fVal, uint16_t cVal)
-    : fVal_(fVal), cVal_(cVal), version_(METADATA_PARAMS_VERSION) {
+PersistentStorageImp::PersistentStorageImp(uint16_t numReplicas, uint16_t fVal, uint16_t cVal)
+    : numReplicas_(numReplicas), fVal_(fVal), cVal_(cVal), version_(METADATA_PARAMS_VERSION) {
   DescriptorOfLastNewView::setViewChangeMsgsNum(fVal, cVal);
 }
 
@@ -75,6 +75,7 @@ void PersistentStorageImp::setDefaultsInMetadataStorage() {
   initDescriptorOfLastExitFromView();
   initDescriptorOfLastNewView();
   initDescriptorOfLastExecution();
+  initDescriptorOfLastStableCheckpoint();
   endWriteTran();
 }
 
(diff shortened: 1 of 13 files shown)


bftengine/src/bftengine/PersistentStorageImp.hpp
```diff
@@ -92,14 +92,15 @@ const uint16_t numOfLastExitFromViewDescObjs = kWorkWindowSize + 1;
 enum DescMetadataParameterIds {
   LAST_EXIT_FROM_VIEW_DESC = WIN_PARAMETERS_NUM + reservedWindowParamsNum,
   LAST_EXEC_DESC = LAST_EXIT_FROM_VIEW_DESC + numOfLastExitFromViewDescObjs,
+  LAST_STABLE_CHECKPOINT_DESC,
```
yuliasherman, line 95: Are you going to save/retrieve all stable checkpoints at once without having the ability to update/get them for a specific replica only?

yuliasherman (approved):
LGTM

bftengine/src/bftengine/PersistentStorageImp.hpp
HristoStaykov: Yes, we need all the quorum of CheckpointMsg-s that serve as proof for our latest Stable Sequence Number. We store them in the transaction that we open in **onSeqNumIsStable** and in case the Replica has been restarted we recover them from persistent storage and insert all of them in **checkpointsLog**. Once the Replica moves to higher Sequence Numbers and collects Stable Checkpoint it will call **advanceActiveWindow** to clean up the old CheckpointMsg-s and will persist the new ones.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -290,11 +293,27 @@ void PersistentStorageImp::setDescriptorOfLastExecution(const DescriptorOfLastEx
...
+void PersistentStorageImp::setDescriptorOfLastStableCheckpoint(
+    const DescriptorOfLastStableCheckpoint &stableCheckDesc) {
+  const size_t bufLen = DescriptorOfLastStableCheckpoint::maxSize(numReplicas_);
+  UniquePtrToChar descBuf(new char[bufLen]);
+  char *descBufPtr = descBuf.get();
+  size_t actualSize = 0;
+  stableCheckDesc.serialize(descBufPtr, bufLen, actualSize);
+  ConcordAssert(actualSize != 0);
+  metadataStorage_->writeInBatch(LAST_STABLE_CHECKPOINT_DESC, descBuf.get(), actualSize);
+}
```
teoparvanov, line 305: Hi @HristoStaykov , does it make sense to add unit tests for the new descriptor in [TestSerialization.cpp](https://github.com/vmware/concord-bft/blob/master/bftengine/tests/testSerialization/TestSerialization.cpp)?

teoparvanov (approved):
Looks great, thanks @HristoStaykov !

bftengine/src/bftengine/PersistentStorageImp.cpp
HristoStaykov: Yes, definitely. Will add. Thanks!

glevkovich (approved):

PR #1511, opened 2021-05-28
Test refactor for consistent recreation of infinite View Change.

HristoStaykov:
Refactored the test to recreate consistently infinite VC with only 1
Replica down. This bug will be fixed by PR-s 1478, 1489 and 1510.


KTeofanidis (approved):
LGTM

PR #1512, opened 2021-05-28
ThinReplicaClient: Prepend namespace with client::

robem:
In order to keep folder struture, include paths, and namespaces in sync, this
change moves the existing thin_replica_client namespace to
client::thin_replica_client.


shgandhi (approved):

PR #1513, opened 2021-05-28
ThinReplicaClient: Tabs vs spaces cleanup

robem:
This change removes all tabs and re-aligns lines.


shgandhi (approved):
Thanks for the cleanup @robem! My bad, I missed this. Lgtm!

PR #1518, opened 2021-05-31
bft_client: snapshot signing duration into log every 3 minutes

glevkovich:
We do not use concord_ctl on client. In order to have an idea on how
long it takes to sign, we will take a snapshot every 3 minutes and log
log the results (debug level) every 3 minutes.


Liorgal28 (approved):
LGTM!

PR #1520, opened 2021-05-31
Don't flush consensus batch on specific policies

yuliasherman:
Forced consensus batch flush is relevant only for BATCH_SELF_ADJUSTED; don't launch it for other policies.


Liorgal28 (approved):
LGTM

PR #1521, opened 2021-05-31
ConcordClient Interface

robem:
This change introduces the Concord Client C++ interface.
See commits for details.


shgandhi (approved):

Liorgal28 (approved):
Looks good @robem!

andrewjstone (approved):
Looks great @robem. This is a huge step forward for the product.

One suggestion: I feel that the concord-client api subscription mechanism should be agnostic to implementation. Currently it's coupled to thin replica, but only by types. We may want to move those types into a separate library to enable decoupling.

robem:
> Looks great @robem. This is a huge step forward for the product.
> 
> One suggestion: I feel that the concord-client api subscription mechanism should be agnostic to implementation. Currently it's coupled to thin replica, but only by types. We may want to move those types into a separate library to enable decoupling.

We are doing the same for the BFT client. However, you can argue that the BFT client **is** Concord Client and maybe the thin replica functionality will become part of it.
Are you suggesting we should copy `TrsConnectionConfig`, to e.g. `SubscriptionConfig`, and combine it with the client id?

andrewjstone:
> > Looks great @robem. This is a huge step forward for the product.
> > One suggestion: I feel that the concord-client api subscription mechanism should be agnostic to implementation. Currently it's coupled to thin replica, but only by types. We may want to move those types into a separate library to enable decoupling.
> 
> We are doing the same for the BFT client. However, you can argue that the BFT client **is** Concord Client and maybe the thin replica functionality will become part of it.
> Are you suggesting we should copy `TrsConnectionConfig`, to e.g. `SubscriptionConfig`, and combine it with the client id?

Yep. That's basically what I'm suggesting. It's certainly not necessary in this commit though. It's just something to think about.

andrewjstone (approved):
Looks great! Ship it!

PR #1522, opened 2021-06-01
updateAggregatorCounter is used from multiple threads

KTeofanidis:
updateAggregatorCounter should be atomic since it's being touched from multiple threads


teoparvanov (approved):
LGTM, thanks @KTeofanidis !

glevkovich (approved):
LGTM!

PR #1523, opened 2021-06-01
Add pre, post and regular reconfiguration handlers

yontyon:
We want to have the ability to define an order on the reconfiguration actions that we take.
For example, we would like to persist the reconfiguration request to the blockchain prior to the actual execution, such that if we crash, we have the record in the blockchain.
This PR introduces the notion of pre, post, and regular reconfiguration handlers.
The dispatcher on its side first executes the pre, then the regular and only then the post.
Notice, that currently we don't have post-pre-execution handlers but we may have them in the future.

### DIFF
--- a/bftengine/include/bftengine/IRequestHandler.hpp
+++ b/bftengine/include/bftengine/IRequestHandler.hpp
@@ -23,6 +23,7 @@
 
 namespace concord::reconfiguration {
 class IReconfigurationHandler;
+enum ReconfigurationHandlerType : unsigned int;
 }  // namespace concord::reconfiguration
 
 namespace bftEngine {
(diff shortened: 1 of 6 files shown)


bftengine/include/bftengine/IRequestHandler.hpp
```diff
@@ -55,7 +56,9 @@ class IRequestsHandler {
   std::shared_ptr<concord::reconfiguration::IReconfigurationHandler> getReconfigurationHandler() const {
     return reconfig_handler_;
   }
-  virtual void setReconfigurationHandler(std::shared_ptr<concord::reconfiguration::IReconfigurationHandler> rh) {
+  virtual void setReconfigurationHandler(std::shared_ptr<concord::reconfiguration::IReconfigurationHandler> rh,
+                                         concord::reconfiguration::ReconfigurationHandlerType type =
+                                             static_cast<concord::reconfiguration::ReconfigurationHandlerType>(1)) {
```
nkumar04, line 61: Should we use REGULAR here for better readability

nkumar04 (approved):
LGTM!!

bftengine/include/bftengine/IRequestHandler.hpp
yontyon: We can't. Here we use forward declaration of the enum.

PR #1526, opened 2021-06-01
bft_client: Add a histogram to record transaction size in bytes

glevkovich:
snapshot of this histogram is taken once in 3 minutes with the other
historgram which belongs to the same component.


yontyon (approved):
LGTM!
Thanks

PR #1527, opened 2021-06-01
Fix potential race with metrics aggregator while starting replicas

nkgeorgiev:
In ReplicaInternal::start() we first start the replica and then update the preprocessor aggregator with the replica one.
The last step of ReplicaInternal::start() is to async sendInitialKey().
In rare conditions we may get a race where we try to update some metrics in key-exchange while we are updating the aggregator, which may lead to map.at() throwing std::out_of_range
The fix is to update the aggregator before starting the replica.


yuliasherman (approved):
LGTM, thanks for fixing this!

HristoStaykov (approved):
LGTM, thanks @nkgeorgiev !

PR #1529, opened 2021-06-01
AddRemove reconfiguration with wedge and erase metadata flag

nkumar04:
* Add new command to do addremove with wedge and set flag to erase metadata on restart
* Add apollo test to remove node with and without f failures
* Add apollo test to add new nodes to the cluster


kvbc/include/reconfiguration_add_block_handler.hpp
```diff
@@ -77,6 +77,23 @@ class ReconfigurationHandler : public concord::reconfiguration::BftReconfigurati
...
+  bool handle(const concord::messages::AddRemoveWithWedgeCommand& command,
+              uint64_t sequence_number,
+              concord::messages::ReconfigurationResponse&) override {
+    std::vector<uint8_t> serialized_command;
+    concord::messages::serialize(serialized_command, command);
+    auto blockId = persistReconfigurationBlock(
+        serialized_command, sequence_number, std::string{kvbc::keyTypes::reconfiguration_add_remove});
+    LOG_INFO(getLogger(), "AddRemove configuration command block is " << blockId);
+
+    concord::messages::WedgeCommand wedgeCommand{command.id, false};
```
yontyon, line 89: The add remove command should be interpreted as "wedge included". No need to write a wedge command to the blockchain (and respectively, after ST this command should be interpreted as "wedge included"

reconfiguration/cmf/concord.cmf
```diff
@@ -94,6 +94,11 @@ Msg AddRemoveStatusResponse 28 {
     string reconfiguration
 }
 
+Msg AddRemoveWithWedgeCommand 29 {
+    uint64 sender_id
+    string reconfiguration
```
yontyon, line 99: I would change the name to "config_descriptor" (as this is what it is going to be)

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+             1. A client sends a remove command which will also wedge the system on next next checkpoint
+             2. Validate that all replicas have stopped
+             3. Load  a new configuration to the bft network
+             4. Rerun the cluster with only 4 nodes and make sure they succeed to perform transactions in fast path
+         """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        for i in range(100):
+            await skvbc.write_known_kv()
+        key, val = await skvbc.write_known_kv()
```
yontyon, line 574: Why another one?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+        client = bft_network.random_client()
+        client.config._replace(req_timeout_milli=10000)
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=0)
+        op = operator.Operator(bft_network.config, client,  bft_network.builddir)
+        test_config = 'new_configuration_n_4_f_1_c_0'
+        await op.add_remove_with_wedge(test_config)
+        await self.verify_replicas_are_in_wedged_checkpoint(bft_network, checkpoint_before, range(bft_network.config.n))
+        await self.verify_last_executed_seq_num(bft_network, checkpoint_before)
+        await self.validate_stop_on_super_stable_checkpoint(bft_network, skvbc)
```
yontyon, line 583: We assume here only n-f wedge, so mp need to test the wedge on a super stable checkpoint (super stable means that n/n replicas are in the wedge point)

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+        bft_network.stop_all_replicas()
+        # We now expect the replicas to start with a fresh new configuration which means that we
+        # need to see in the logs that isNewStorage() = true. Also,
```
yontyon, line 586: not sure this comment is correct. Please check

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_remove_nodes_with_f_failures(self, bft_network):
+        """
+        In this test we show how a system operator can remove nodes (and thus reduce the cluster) from 7 nodes cluster
+        to 4 nodes cluster even when f nodes are not responding
+        For that the operator performs the following steps:
+        1. Stop 2 nodes (f=2)
+        2. Send a remove_node command - this command also wedges the system
+        3. Verify that all live (including the removed candidates) nodes have stopped
```
yontyon, line 619: If we have f crashes, why do we need to check that they have stopped?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+        4. Load  a new configuration to the bft network
+        5. Rerun the cluster with only 4 nodes and make sure they succeed to perform transactions in fast path
+        """
+        bft_network.start_all_replicas()
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        client = bft_network.random_client()
+        for i in range(100):
+            await skvbc.write_known_kv()
+        # choose two replicas to crash and crash them
+        crashed_replicas = {5, 6} # For simplicity, we crash the last two replicas
```
yontyon, line 629: would it work also with random replicas? (I guess it will, but it will be a much stronger test if we choose the the crashed replicas randomly)

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+        live_replicas = bft_network.all_replicas(without=crashed_replicas)
+        client = bft_network.random_client()
+        client.config._replace(req_timeout_milli=10000)
+        checkpoint_before = await bft_network.wait_for_checkpoint(replica_id=0)
+        op = operator.Operator(bft_network.config, client,  bft_network.builddir)
+        test_config = 'new_configuration_n_4_f_1_c_0'
+        await op.add_remove_with_wedge(test_config)
+        await self.verify_replicas_are_in_wedged_checkpoint(bft_network, checkpoint_before, range(len(live_replicas)))
+        expectedSeqNum = (checkpoint_before  + 2) * 150
+        for r in bft_network.all_replicas(without=crashed_replicas):
```
yontyon, line 644: use live_replicas instead of `bft_network.all_replicas(without=crashed_replicas)`

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+            lastExecSn = await bft_network.get_metric(r, bft_network, "Gauges", "lastExecutedSeqNum")
+            self.assertEqual(expectedSeqNum, lastExecSn)
+        bft_network.stop_all_replicas()
+        # We now expect the replicas to start with a fresh new configuration which means that we
+        # need to see in the logs that isNewStorage() = true. Also,
```
yontyon, line 649: same as before

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+        for r in bft_network.all_replicas():
+            assert (r < 4)
+            nb_fast_path = await bft_network.get_metric(r, bft_network, "Counters", "totalFastPaths")
+            self.assertGreater(nb_fast_path, 0)
+    
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_add_nodes(self, bft_network):
+        """
+             Sends a addRemove command and checks that new configuration is written to blockchain.
```
yontyon, line 676: Where the test check that the command is written to the blockchain?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+             Note that in this test we assume no failures and synchronized network.
+             The test does the following:
+             1. A client sends a add node command which will also wedge the system on next next checkpoint
+             2. Validate that all replicas have stopped
+             3. Load a new configuration to the bft network
+             4. Add node is done in phases, (n=4,f=1,c=0)->(n=6,f=1,c=0)->(n=7,f=2,c=0)
```
yontyon, line 682: I would like to have another test (maybe in a different PR) in which we use the state of one replica for the new 3 replicas such that we move directly from 4->7 replicas.
This can be done by copying the rocksdb data to the directories of the new replicas.
Once the new replicas are starting they will remove the old metadata and they should start correctly.
Note that this may require changes in the bft module

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -555,6 +555,228 @@ async def test_addRemove_command(self, bft_network):
...
+                   start_replica_cmd=start_replica_cmd,
+                   stop_replica_cmd=None,
+                   num_ro_replicas=0)
+        await bft_network.change_configuration(conf)
+        initial_prim = 0
+        new_replicas = {4, 5}
+        on_time_replicas = bft_network.all_replicas(without=new_replicas)
+        bft_network.start_replicas(on_time_replicas)
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        for i in range(151):
```
yontyon, line 728: We have to issue 151 requests after the restart to initiate state transfer, right? (because we remove the metadata)

yontyon (commented):
Great work!
I have only few comments

reconfiguration/cmf/concord.cmf
yontyon: Also, I would add a bft filed (like in the wedge command), this way we give the operator the control if to wait for n/n wedge or n-f wedge

reconfiguration/src/reconfiguration_handler.cpp
```diff
@@ -57,6 +57,15 @@ bool ReconfigurationHandler::handle(const KeyExchangeCommand& command,
 
   return true;
 }
+bool ReconfigurationHandler::handle(const concord::messages::AddRemoveWithWedgeCommand&,
+                                    uint64_t bft_seq_num,
+                                    concord::messages::ReconfigurationResponse&) {
+  LOG_INFO(getLogger(), "AddRemoveWithWedgeCommand instructs replica to stop at sequence number " << bft_seq_num);
+  bftEngine::ControlStateManager::instance().setStopAtNextCheckpoint(bft_seq_num);
+  bftEngine::IControlHandler::instance()->addOnStableCheckpointCallBack(
```
yontyon, line 65: Following the "adding bft field to the command", if we have bft=false, then the callback should be called on superStableCheckpoint (see the wedge handler)

nkumar04: fixed

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: yes 151 requests are issues to initiate state transfer

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: Fixed

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: Fixed

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: Fixed

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: Fixed

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: Fixed

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: This key val is checked later to validate state consistency after cluster is restarted with new configuration

reconfiguration/cmf/concord.cmf
nkumar04: fixed

kvbc/include/reconfiguration_add_block_handler.hpp
nkumar04: fixed

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: Fixed

kvbc/src/st_reconfiguration_sm.cpp
```diff
@@ -44,16 +44,14 @@ void StReconfigurationHandler::stCallBack(uint64_t current_cp_num) {
...
 template <typename T>
 bool StReconfigurationHandler::handlerStoredCommand(const std::string &key, uint64_t current_cp_num) {
   auto res = ro_storage_.getLatest(kvbc::kConcordInternalCategoryId, key);
   if (res.has_value()) {
-    auto blockid =
-        ro_storage_
-            .getLatestVersion(kvbc::kConcordInternalCategoryId, std::string{kvbc::keyTypes::reconfiguration_wedge_key})
-            .value()
-            .version;
+    auto blockid = ro_storage_.getLatestVersion(kvbc::kConcordInternalCategoryId, key).value().version;
```
yontyon, line 54: good catch!
Thanks!

yontyon (approved):
LGTM!

tests/apollo/test_skvbc_reconfiguration.py
nkumar04: It's bit complicated to test with random replicas failure instead of last two i.e. {5,6}. Let say, test chooses random nodes in {0,1,2,3} - for example 2,3. Then after i restart the cluster with (n=4,f=1), there are two nodes with empty state. So, they won't be able to catch up via state transfer. At least we should have 3 non-faulty nodes

PR #1530, opened 2021-06-01
Publish clients public keys

cloudnoize:
This feature will publish the public keys of the external clients on start whenever a change is detected.
The flow is as follows:

- On start, the public keys of the client are added to a serializable map.
- On start, the primary compares the version above with the version persisted in the reserved pages.
- If versions not equal, the publish flag will be set to false and replicas won't accept msgs as in key-exchage.
- if publish is false, the primary will send the serialized keys via consensus.
- Upon execution, there replicas will store the digest of the published keys in reserved pages and will set the publish flag to true.


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2641,6 +2643,8 @@ void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNew
   LOG_INFO(VC_LOG, "Start working in new view: " << KVLOG(curView));
 
   controller->onNewView(curView, primaryLastUsedSeqNum);
+  // pubish clients public keys, in case prev primar failed to publish
+  if (primaryIsMe) KeyExchangeManager::instance().publishClientsPublicKeys(SigManager::instance()->ClientsPublicKeys());
```
cloudnoize, line 2647: @HristoStaykov can you confirm this code calls ```publishClientsPublicKeys``` on VC complete ?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -133,14 +133,20 @@ SigManager::SigManager(PrincipalId myId,
...
+    } else {
       verifiers_[p.first] = iter->second;
+    }
+    if (replicasInfo_.isIdOfExternalClient(p.first)) {
+      clientsPublicKeys_.ids_to_keys[p.first] = concord::messages::keys_and_signatures::PublicKey{key, (uint8_t)format};
+      LOG_DEBUG(ON_CHAIN_LOG, "Adding key of client " << p.first);
+    }
   }
 
+  LOG_DEBUG(ON_CHAIN_LOG, "Map contains " << clientsPublicKeys_.ids_to_keys.size() << " public clients keys");
```
cloudnoize, line 149: @glevkovich can you confirm that this code, inserts correctly the public keys of the external clients?

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: This is the point in the View Change process where the beginning of the rebuilding of the previous View's Working Window begins. After this point we have 300 Sequence ID-s for the Replicas to reach consensus on (as for now), but I think posting the Client Operation here should be ok. The Primary never complains against himself, and he is always the first to enter the View.

glevkovich (approved):
LGTM!

bftengine/src/bftengine/SigManager.cpp
glevkovich: @cloudnoize , looks fine to me..

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -131,19 +133,39 @@ void KeyExchangeManager::sendKeyExchange(const SeqNum& sn) {
...
+  LOG_DEBUG(ON_CHAIN_LOG, "Sends clients keys, cid: " << cid << " payload size " << keys.size());
+  client_->sendRequest(
+      bftEngine::CLIENTS_PUB_KEYS_FLAG | bftEngine::PUBLISH_ON_CHAIN_OBJECT_FLAG, keys.size(), keys.c_str(), cid);
+}
+
+void KeyExchangeManager::sendInitialKey(std::optional<std::string> clientsPublicKeys) {
   LOG_INFO(KEY_EX_LOG, "");
   if (private_keys_.hasGeneratedKeys()) {
     LOG_INFO(KEY_EX_LOG, "Replica has already generated keys");
+    if (clientsPublicKeys) publishClientsPublicKeys(*clientsPublicKeys);
```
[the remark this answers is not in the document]
cloudnoize: it can be std::nullopt if it's not the primary, that why I've chosen to use std::optional

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -131,19 +133,39 @@ void KeyExchangeManager::sendKeyExchange(const SeqNum& sn) {
...
+  auto ret = std::async(std::launch::async, [this, clientsPublicKeys]() {
     SCOPED_MDC(MDC_REPLICA_ID_KEY, std::to_string(ReplicaConfig::instance().replicaId));
     waitForFullCommunication();
     sendKeyExchange(0);
     metrics_->sent_key_exchange_on_start_status.Get().Set("True");
+    if (!clientsPublicKeys) return;
+    // In order to prevent race on the client, we'll wait untill we know, that all replicas finished processing the key
+    // exchange requests.
+    while (!initial_exchange_) {
+      std::this_thread::sleep_for(std::chrono::seconds(1));
```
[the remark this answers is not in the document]
cloudnoize: You're right, but on the other hand if ```initial_exchange_``` is not true the replica is out of order anyway, and it occurs in a different thread

bftengine/src/bftengine/SigManager.cpp
cloudnoize: Thanks for the verification @glevkovich !

bftengine/include/bftengine/KeyExchangeManager.hpp
```diff
@@ -42,6 +42,22 @@ class KeyExchangeManager {
     return ReplicaConfig::instance().getkeyExchangeOnStart() ? (publicKeys_.numOfExchangedReplicas() == clusterSize_)
                                                              : true;
   }
+  const std::string kInitialKeyExchangeCid = "KEY-EXCHANGE-";
+  const std::string kInitialClientsKeysCid = "INITIAL-CLIENTS-PUB-KEYS-";
```
toly-kournik, line 46: Thinking about periodic client keys exchange do we need th INITIAL- prefix?

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -131,6 +133,27 @@ void KeyExchangeManager::sendKeyExchange(const SeqNum& sn) {
...
+    }
+    auto cid = generateCid(kInitialClientsKeysCid);
+    LOG_DEBUG(ON_CHAIN_LOG, "Sends clients keys, cid: " << cid << " payload size " << keys.size());
+    client_->sendRequest(
+        bftEngine::CLIENTS_PUB_KEYS_FLAG | bftEngine::PUBLISH_ON_CHAIN_OBJECT_FLAG, keys.size(), keys.c_str(), cid);
+  });
+}
+
+bool KeyExchangeManager::isInitialClientKeysReq(const std::string& cid) {
+  return cid.find(kInitialClientsKeysCid) != std::string::npos;
```
toly-kournik, line 154: Can't we have another indication except the string comparison?

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -66,6 +67,33 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
...
     }
+    // Replicas can publish an object e.g public_keys, configuration file, etc
+    // this object pass consensus, and replicas can perform action against is as:
+    // - validated that is equal to the object that is stored in memory of the replica.
+    // - save it to reserved pages.
+    // - proxy it to the application command handler.
+    if (req.flags & bftEngine::MsgFlag::PUBLISH_ON_CHAIN_OBJECT_FLAG) {
+      auto save = true;
+      if (req.flags & bftEngine::MsgFlag::CLIENTS_PUB_KEYS_FLAG) {
+        std::string recieved_keys(req.request, req.requestSize);
```
toly-kournik, line 78: I think this logic should be internal to KeyExchangeManager and not visible in the RequestHandler

bftengine/src/bftengine/SigManager.hpp
```diff
@@ -72,6 +73,13 @@ class SigManager {
   SigManager(SigManager&&) = delete;
   SigManager& operator=(SigManager&&) = delete;
 
+  concord::messages::keys_and_signatures::ClientsPublicKeys clientsPublicKeys_;
+  std::string ClientsPublicKeys() {
```
toly-kournik, line 77: confusing - function name looks like constructor because it starts from a capital letter

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -147,7 +147,7 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
   // Keys Management
   CONFIG_PARAM(keyExchangeOnStart, bool, false, "whether to perform initial key exchange");
   CONFIG_PARAM(keyViewFilePath, std::string, ".", "TODO");
-
+  CONFIG_PARAM(publishClientsKeysOnStart, bool, false, "whether to publish client keys");
```
toly-kournik, line 150: Is it a rollout flag which eventually will be always true, or can it be disabled in future?

bftengine/include/bftengine/KeyExchangeManager.hpp
cloudnoize: the periodic will be from an external client, and it won't get this CID, I want a way to differentiate an initial request because the initial request verifies against the bootstrap content.

bftengine/src/bftengine/KeyExchangeManager.cpp
cloudnoize: the cid is the indication

bftengine/src/bftengine/RequestHandler.cpp
cloudnoize: you're right, I had that doubt

bftengine/include/bftengine/ReplicaConfig.hpp
cloudnoize: I don't want this feature to be enabled by default in Apollo since it has side effects i.e. start sn, and it makes other tests like VC flaky

logging/src/Logger.cpp
```diff
@@ -34,3 +34,4 @@ logging::Logger ST_SRC_LOG = logging::getLogger("concord.bft.st.src");
 logging::Logger MSGS = logging::getLogger("concord.bft.msgs");
 logging::Logger CL_MNGR = logging::getLogger("concord.bft.client-manager");
 logging::Logger TS_MNGR = logging::getLogger("concord.bft.time-service-manager");
+logging::Logger ON_CHAIN_LOG = logging::getLogger("persist.on-chain");
```
toly-kournik, line 37: Isn't it logically belongs to a "concord" logging namespace?

cloudnoize: it's cross layers feature from bft to concord, what do you think?

bftengine/src/bftengine/RequestHandler.cpp
toly-kournik: are you going to change this?

logging/src/Logger.cpp
toly-kournik: all our logs start from "concord".
besides, why key-exchange log is not suitable here?

cloudnoize: easier to analyse, but I can reuse it

bftengine/src/bftengine/RequestHandler.cpp
cloudnoize: sure, I consider your comments as the bible

bftengine/src/bftengine/KeyExchangeManager.cpp
cloudnoize: actually you're right, I have the validation flag on the request I'm using

bftengine/include/bftengine/KeyExchangeManager.hpp
cloudnoize: I'll remove the initial, see the comment on the comment below

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -66,6 +67,25 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
...
     }
+    // Replicas can publish an object e.g public_keys, configuration file, etc
+    // this object pass consensus, and replicas can perform action against is as:
+    // - validated that is equal to the object that is stored in memory of the replica.
+    // - save it to reserved pages.
+    // - proxy it to the application command handler.
+    if (req.flags & bftEngine::MsgFlag::CLIENTS_PUB_KEYS_FLAG) {
+      std::string recieved_keys(req.request, req.requestSize);
+      std::optional<std::string> bootstrap_keys;
+      if (req.flags & bftEngine::MsgFlag::PUBLISH_ON_CHAIN_OBJECT_FLAG) {
```
toly-kournik, line 78: is PUBLISH_ON_CHAIN_OBJECT_FLAG for initial exchange only?

cloudnoize: for publish keys, yes

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -131,6 +133,37 @@ void KeyExchangeManager::sendKeyExchange(const SeqNum& sn) {
...
 }
 
+// sends the clients public keys via the internal client, if keys weren't published or outdated.
+void KeyExchangeManager::sendInitialClientsKeys(const std::string& keys) {
+  if (clientsPublicKeys_.published_) {
+    LOG_INFO(KEY_EX_LOG, "Clients public keys were already published");
+    return;
+  }
+  auto ret = std::async(std::launch::async, [this, keys]() {
+    while (!initial_exchange_) {
```
toly-kournik, line 143: why not to use an atomic bool and not exchanged() only?

bftengine/src/bftengine/KeyStore.h
```diff
@@ -98,4 +100,28 @@ class ClusterKeyStore : public ResPagesClient<ClusterKeyStore> {
 
   std::string buffer_;
 };
+
+// Manages the BFT state of the clients public keys:
+// It stores the digest of the keys in the reserved pages and sets the published_ accordingly
+class ClientKeyStore : public ResPagesClient<ClientKeyStore, 1> {
+ public:
+  bool published_{false};
```
toly-kournik, line 108: I guess `published_` variable will be removed when we'll load actual keys rather than a hash, then we will be able to check a key container size to understand whether keys are published.

bftengine/src/bftengine/KeyExchangeManager.cpp
cloudnoize: exchanged is not thread-safe no?

bftengine/src/bftengine/KeyStore.h
cloudnoize: but then you'll do it each time you query whether keys were published?

toly-kournik (approved):
LGTM!

PR #1531, opened 2021-06-01
Rename ReplicaImp to Replica

tdimitrov:
There are two classes named ReplicaImp in the codebase:
(1) concord::kvbc::ReplicaImp
(2) bftEngine::impl::ReplicaImp

They are in different namespaces but the name collision causes a lot of
confusion.

(1) looks like more general representation of a replica. It
has got generic methods as start, stop, get, add, etc.

(2) has got far more specifics and ReplicaImp is a good name for it.

This patch renames (1) to just Replica so that there is less confusion
in terms of reading the code.


andrewjstone (approved):
LGTM.

PR #1532, opened 2021-06-01
Making commands and keys uniform for concord-ctl

saiakhil2012:
- Changing key listing command to `list` for `status` command to be uniform with `perf` command of `concord-ctl`
- Making keys of the output of commands uniform


diagnostics/include/protocol.h
```diff
@@ -95,7 +95,7 @@ inline std::string run(const std::vector<std::string>& tokens, Registrar& regist
         }
         return usage();
       }
-      if (command == "list") {
+      if (command == "list-keys") {
```
andrewjstone, line 98: My preference would be to change `list-keys` in status to `list`, rather than the other way around.. I think the shorter name is better for a cli.

saiakhil2012: sure, agreed, changed the command to `list` for both `status` and `perf` commands

andrewjstone (approved):
LGTM. Thanks for the cleanup @saiakhil2012

PR #1535, opened 2021-06-02
Race Condition Fixes-1

ukg19:
In past we have seen crashed in Replica due to different race conditions. During analysis , we ran thread sanitizer to find different race conditions in our code base and have found 9 such cases.

This PR fixes 4 of those race conditions where
1. fetching is a non-atomic bool variable fetched and set by two different threads
2. Class member variable val_ of Template type T fetched and set by two different threads for class BasicGauge
3. Class member variable val_ of Template type T fetched and set by two different threads for class BasicCounter
4. msgLoopDone_ is being accessed and set in different threads.


util/include/Metrics.hpp
```diff
@@ -29,8 +29,8 @@ class BasicGauge;
 template <class T>
 class BasicCounter;
 
-using Gauge = BasicGauge<uint64_t>;
-using Counter = BasicCounter<uint64_t>;
+using Gauge = BasicGauge<std::atomic_uint64_t>;
+using Counter = BasicCounter<std::atomic_uint64_t>;
```
HristoStaykov, line 33: I think if we make all Counters/Gauges atomic, we might have a bit of performance degradation in the cases where they are only manipulated from a single thread. Even though by making this change we will solve all potential problems I think it might be better to change the types of the ones that are reported by TSAN to the atomic counterpart.

ukg19: @HristoStaykov : Both of them were reported by TSAN. That's why I have made changes in both.
Screenshot 2021-06-02 at 3 29 32 PM

util/include/Metrics.hpp
```diff
@@ -29,8 +29,8 @@ class BasicGauge;
...
 using AtomicGauge = BasicGauge<std::atomic_uint64_t>;
 using AtomicCounter = BasicCounter<std::atomic_uint64_t>;
```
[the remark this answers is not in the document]
ukg19: That's right , let me fix this.

util/include/Metrics.hpp
ukg19: Following were the Race conditions regarding Metric , so I have made them atomic accordingly

1. Write: BCStateTran.cpp:402: metrics_.last_block_.Get().Set(lastBlock);

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

2. Write : BCStateTran.cpp:570 : metrics_.load_reserved_page_from_checkpoint_.Get().Inc();

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

3. Write : BCStateTran.cpp:496 : metrics_.create_checkpoint_.Get().Inc();

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

4. Write : BCStateTran.cpp:587, metrics_.save_reserved_page_.Get().Inc();

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

util/include/Metrics.hpp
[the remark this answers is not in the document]
ukg19: 1. Write: BCStateTran.cpp:402: metrics_.last_block_.Get().Set(lastBlock);

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

2. Write : BCStateTran.cpp:570 : metrics_.load_reserved_page_from_checkpoint_.Get().Inc();

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

3. Write : BCStateTran.cpp:496 : metrics_.create_checkpoint_.Get().Inc();

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

4. Write : BCStateTran.cpp:587, metrics_.save_reserved_page_.Get().Inc();

Read : BCStateTran.cpp:647, metrics_component_.UpdateAggregator();

teoparvanov (approved):
This is much better, thank you @ukg19 !

HristoStaykov (approved):
LGTM, thanks @ukg19 !

PR #1536, opened 2021-06-02
CI stability fixes.

HristoStaykov:
Test is only valid for N=7 nodes.
Removed unnecessary write of Client request.


teoparvanov (approved):
LGTM, thanks Hristo!

PR #1537, opened 2021-06-02
bft_client: snapshot signing stats every fixed amount of signing

glevkovich:
till now we relay on time, every 3 minutes.
From now on snapshot to log every 200 signs by a single client.


HristoStaykov (approved):
LGTM, thanks @glevkovich !

PR #1540, opened 2021-06-02
Fix SimpleTest

salieri11:
This PR fixes SimpleTest client to wait for the communication layer to connect to at least N - F replicas.


andrewjstone (approved):
LGTM. Thanks for the fix Igor!

andrewjstone (changes requested):
Sorry @salieri11. I didn't look close enough in my prior approval.

This can double count the same replicas. We should either use a std::set for ready replicas and check the size as loop condition or reset `readyReplicas = 0` inside the while loop.

PR #1543, opened 2021-06-03
Remove config generated as metadata

yontyon:
When we remove the metadata on startup, we also need to remove the gen-sec key that contains the rotated private key.
This PR handles it as well as change the Apollo tests to support key exchange and test this functionality


nkumar04 (approved):
LGTM!!

PR #1544, opened 2021-06-03
Keep flushing consensus's in noop scenario

Liorgal28:
In this scenario, we are trying to flush consensus to fill the 300 window as fast as possible; that's why we need to remove the check if it self adjust policy.


yontyon (approved):
LGTM!

PR #1545, opened 2021-06-03
Maintain reserved pages client indices in one place

dartdart26:
Introduce the `ReservedPagesClientIndex` enum that maintains the
reserved pages client indices in one place. Additionally, making it an
`enum class` ensures one cannot instantiate the `ResPagesClient` class
with an arbitrary integer.


dartdart26:
Closing this PR, because as discussed offline with @toly-kournik, there is a way to order clients by their type hash and not have to bother users with indices.

dartdart26:
Closing this PR, because as discussed offline with @toly-kournik, there is a way to order clients by their type hash and not have to bother users with indices.

PR #1547, opened 2021-06-03
Migrate pyclient_tests_udp to Apollo

upshaw-alex:
This PR adds a test suite to Apollo, specifically `skvbc_pyclient_tests` (in `tests/apollo/test_skvbc_pyclient.py`), re-implementing the test suite `pyclient_tests_udp` (in `util/pyclient/test_client.py`); this PR also removes that original implementation of these tests.

Re-implementing this test suite in Apollo is a clean-up to make these tests more consistent with the other system tests already in Apollo, and we also hope moving this suite to Apollo will improve its stability and facilitate running these tests against against different cluster configurations.


tests/apollo/test_skvbc_pyclient.py
```diff
@@ -0,0 +1,191 @@
...
+        key = protocol.random_key()
+        value = protocol.random_value()
+        kv_pair = [(key, value)]
+        await client.write(protocol.write_req([], kv_pair, 0))
+
+        quorum = MofNQuorum(bft_network.all_replicas(), 1)
+        read_result = await client.read(protocol.read_req([key]), \
+                                        m_of_n_quorum=quorum)
+        value_read = (protocol.parse_reply(read_result))[key]
+        self.assertEqual(value, value_read, "Using an M of N Quorum of M==1, " \
```
upshaw-alex, line 185: For reasons I do not currently fully understand, I have found this assertion to be flaky in my testing. Here is an example of the output I saw from its failure:

```
self.assertEqual(value, value_read, "Using an M of N Quorum of " \
AssertionError: b'MZW86OD9IVKP5LCS2NFYQ' != b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\[39 chars]\x00' : Using an M of N Quorum of M==(F + 1), A BFT Client failed to read a key-value pair from a SimpleKVBC cluster matching the key-value pair it wrote immediately prior to the read.
```

It would appear it is somehow getting a blank value from the server, if I interpret this correctly. I do not know why this would be the case, as I have not seen `test_read_written_value` above fail, and `test_m_of_n_quorum` should be basically the same as that test case, except that it passes an `m_of_n_quorum` parameter to the `read` call. I think the only difference this should make is that the `read` call will only get a response from one replica before returning it rather than validating that response with a larger quorum.

Is there a subtlety of using the `m_of_n_quorum` here that I am missing? Is it possible there are Byzantine-faulty (or at least out-of-sync) replicas I should be worried about in the test network, even though this test case has made no effort to destabilize the cluster?

KTeofanidis: Maybe you can try adding a couple second delay after start_all_replicas similar to what is done in test_skvbc_preexecution.py

KTeofanidis (approved):
LGTM!

tests/apollo/test_skvbc_pyclient.py
andrewjstone: Please don't add sleeps to make tests "pass".

upshaw-alex: What consensus and actions I need to take can I infer from your comments? Is it safe to assume we agree this issue needs to be debugged further and fixed? Also, can I assume neither of you have previous experience with issues of this particular character or speculations as to what could be going wrong here other than that the issue is sensitive to timing?

andrewjstone: Yes, this may be a real bug and should be fixed. It just shouldn't be fixed in a non-deterministic manner by adding sleeps.

andrewjstone:
May I ask why we are moving this to apollo? The test was intended to be a unit test for our python client, which is independent of apollo. If we are moving this test, we may as well move the whole client. What this says to me is that we never intend to use the python client outside apollo. That wasn't the original intention of the code, but I'm fine if that's the decision we are making now.

tests/apollo/test_skvbc_pyclient.py
teoparvanov: Hi @upshaw-alex , I haven't seen this test fail in CI so far (I reran CI on your PR twice, just in case). If the test fails intermittently, we can disable it and investigate separately. For now, let's merge this PR as is. I agree with @andrewjstone that adding a sleep is not the right approach here.

PR #1548, opened 2021-06-04
Wire Concord Cron into Concord-BFT

dartdart26:
Wire Concord Cron (ccron) so that users can take advantage of it. In
general, there are 2 main components that users need to operate ccron
and they are both exposed by kvbc::Replica:
 * TicksGenerator
 * CronTableRegistry

Add the `ticksGeneratorPollPeriod` replica configuration option with a
default value of 1 second. It controls how often the ticks generator
wakes up and, consequently, fires pending ticks. This option specifies
the resolution of cron entry triggers and not how often an entry is
executed - that is dependent on the entry's rule.

Dispatch TickInternalMsg messages to the ticks generator in
ReplicaImp.

Call the relevant cron table from the cron table registry on a tick
ClientRequestMsg in RequestHandler.

Add a ccron reserved pages client and use it directly. Remove the
IReservedPages parameter from periodic action functions.

Add the `skvbc_cron` Apollo test with a single basic test that,
nevertheless, tests Concord Cron end to end. More tests will be added in
the future.

For the time being, the `ccron` library is removed and ccron is part of
the `corbeft` library. Rationale is explained in the
ccron/CMakeLists.txt file. To sum it up here, doing it doesn't break
encapsulation and doesn't require huge amounts of refactoring.

### DIFF
--- a/bftengine/src/bftengine/RequestHandler.cpp
+++ b/bftengine/src/bftengine/RequestHandler.cpp
@@ -15,6 +15,9 @@
 
 #include "bftengine/KeyExchangeManager.hpp"
 
+#include <ccron/cron_table_registry.hpp>
+#include "ccron_msgs.cmf.hpp"
+
 using concord::messages::ReconfigurationRequest;
 using concord::messages::ReconfigurationResponse;
 
@@ -61,7 +64,32 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
         req.outActualReplySize = 0;
       }
       req.outExecutionStatus = 0;  // stop further processing of this request
+    } else if (req.flags & TICK_FLAG) {
+      // Make sure the reply always contains one dummy 0 byte. Needed as empty replies are not supported at that stage.
+      // Also, set replica specific information size to 0.
+      req.outActualReplySize = 1;
+      req.outReply[0] = '\0';
+      req.outReplicaSpecificInfoSize = 0;
+
+      // Default is success (0).
+      req.outExecutionStatus = 0;
+
+      if (req.flags & READ_ONLY_FLAG) {
+        LOG_ERROR(GL, "Received a read-only Tick, ignoring");
+        req.outExecutionStatus = 1;
+      } else if (cron_table_registry_) {
+        using namespace concord::cron;
+        auto payload = ClientReqMsgTickPayload{};
+        auto req_ptr = reinterpret_cast<const uint8_t*>(req.request);
+        deserialize(req_ptr, req_ptr + req.requestSize, payload);
+        const auto tick = Tick{payload.component_id, req.executionSequenceNum};
+        cron_table_registry_->operator[](payload.component_id).evaluate(tick);
+      } else {
+        LOG_ERROR(GL, "Received a Tick, but the cron table registry is not initialized");
+        req.outExecutionStatus = 2;
+      }
     }
+
     if (req.flags & READ_ONLY_FLAG) {
       // Backward compatible with read only flag prior BC-5126
       req.flags = READ_ONLY_FLAG;
--- a/bftengine/src/bftengine/RequestHandler.cpp
+++ b/bftengine/src/bftengine/RequestHandler.cpp
@@ -83,7 +83,7 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
         auto req_ptr = reinterpret_cast<const uint8_t*>(req.request);
         deserialize(req_ptr, req_ptr + req.requestSize, payload);
         const auto tick = Tick{payload.component_id, req.executionSequenceNum};
-        cron_table_registry_->operator[](payload.component_id).evaluate(tick);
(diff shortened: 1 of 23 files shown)


bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -61,7 +64,32 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
...
+      if (req.flags & READ_ONLY_FLAG) {
+        LOG_ERROR(GL, "Received a read-only Tick, ignoring");
+        req.outExecutionStatus = 1;
+      } else if (cron_table_registry_) {
+        using namespace concord::cron;
+        auto payload = ClientReqMsgTickPayload{};
+        auto req_ptr = reinterpret_cast<const uint8_t*>(req.request);
+        deserialize(req_ptr, req_ptr + req.requestSize, payload);
+        const auto tick = Tick{payload.component_id, req.executionSequenceNum};
+        cron_table_registry_->operator[](payload.component_id).evaluate(tick);
```
andrewjstone, line 86: Why explicitly call `operator[]` instead of just using `[]`?

tests/apollo/test_skvbc_cron.py
```diff
@@ -0,0 +1,73 @@
...
+
+        Instruct TesterReplica to add a cron entry that is executed a specific
+        number of times. Verify that number after an additional period of time,
+        ensuring that the cron table doesn't execute the entry any more, even
+        though ticks generation is proceeding.
+        """
+        bft_network.start_all_replicas()
+
+        # Since the cron entry is called every second, wait for the number of expected executes + a period after them.
+        await trio.sleep(seconds=expected_number_of_executes + self.period_after_executes_seconds)
```
andrewjstone, line 61: This is non-deterministic. The [wait_for](https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft.py#L1194-L1217) pattern should be used instead of a sleep, otherwise the metric checks below may fail.

andrewjstone (approved):
Hi @dartdart26,

Great job @dartdart26! Integration is as clean as it can be for the current structure.

I'm conditionally approving this. Please remove the sleep in the apollo test in favor of a wrapped or unwrapped `wait_for`. Then you are good to merge.

tests/apollo/test_skvbc_cron.py
dartdart26: Oh, cool, I wasn't aware of that function! Will use it, thanks.

bftengine/src/bftengine/RequestHandler.cpp
dartdart26: I think the only reason is that it is a pointer and not a reference. I can dereference it instead.

andrewjstone: It's fine like this. I was just curious :)

tests/apollo/test_skvbc_cron.py
dartdart26: Actually, `get_metric()` seems to be doing the same thing, just with hardcoded parameters. Still, I agree `wait_for()` is better as it takes the test case into consideration, allowing for custom timeout and interval. The `trio.sleep()` I've put seems to be redundant anyway.

Will redo the code with `wait_for()`

andrewjstone: `get_metric` doesn't do the same thing. I thought someone added that at first glance also. It only checks if no request has been received or the metric isn't yet published. In essence it helps with slow startup issues. `wait_for` allows you to wait specifically for a metric to reach an exact value.

dartdart26: I see, but I was assuming I should be calling `wait_for()` with `retrieve_metric()` as a task. As an example:
https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft.py#L1225

In that case, it seems to be doing the same thing, just with hardcoded time values.

I might be misunderstanding you or I might be missing something obvious, just haven't used Apollo a lot.

andrewjstone: Instead of using asserts below in the for loop, you pass a closure to `wait_for` that runs the loop and returns true only if `num_executes == expected_number_of_executes` and `component_id == self.expected_component_id` for all replicas.

andrewjstone: Actually, the task predicate relies on not being `None` for true, but the principle is the same. You loop checking that the conditions you want are true until they are or there is a timeout.

andrewjstone: Ah I see. That test looks buggy to me. A better example is this one: https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft.py#L1124-L1133

dartdart26: Thanks for the comment. I have removed the sleep and made the test independent of timing. I am waiting until the correct number of executes happens and I am using the last executed BFT sequence number to know the system has progressed.

PR #1549, opened 2021-06-06
Reserved pages client usage convenience simplification to rely on typ…

toly-kournik:
…e_index instead of unique hardcoded index provided by a developer.


dartdart26 (approved):
Thanks @toly-kournik, looks great, much easier for users!

PR #1551, opened 2021-06-06
Use appropriate asserts to get more informative logs

yuliasherman:
- The basic ConcordAssert does not print the required information in case a condition contains more than one expression.
- Fix requests bitmap index for skpped requests.


yontyon (approved):
lgtm

PR #1552, opened 2021-06-06
Add/remove integration with read only replica

yontyon:
In this PR we present add/Remove functionality in read-only replica via state transfer mechanism.
For this we added the following:
1. priorities in state-transfer callbacks (we need the reconfiguration call back to be called first)
2. Handler for addRemove command in reconfiguration state transfer handler
3. The handler is invoked only if the state transfer checkpoint is equal to the wedge point, which prevents multiple invocations for the same reconfiguration action once we continue beyond it.
Note that this solution also solves the scenario of late replicas.

In addition, we fixed some more issues:
1. We mark directly to remove the metadata (instead of raising a flag to mark it)
2. We add an Apollo test for removing nodes with a read-only replica
3. We add an Apollo test for removing nodes with late replica 
4. We fixed the key exchange feature in the tests


nkumar04 (approved):
LGTM!!

PR #1553, opened 2021-06-07
Changes to upgrade TLS to 1.3

cjn:
Added Changes to upgrade TLS to version 1.3 by default for inter-replica communication and client-replica communication.


salieri11:
I would consider to replace default cipher suite if we move to 1.3. Today we use ECDHE-ECDSA-AES256-GCM-SHA384, which is pretty strong one, but maybe in 1.3 there are better cipher suites?

andrewjstone:
> I would consider to replace default cipher suite if we move to 1.3. Today we use ECDHE-ECDSA-AES256-GCM-SHA384, which is pretty strong one, but maybe in 1.3 there are better cipher suites?

Great idea @salieri11 ! I just found a [page discussing this](https://www.thesslstore.com/blog/tls-1-3-everything-possibly-needed-know/) and it looks like the closest to our current suite is `TLS_AES_256_GCM_SHA384` (Key exchange and signature algos are no longer included).

andrewjstone (approved):
This looks great @cjn. Please just change the cipher suite as pointed out in the comments. Then you are good to merge.

cjn:
> I would consider to replace default cipher suite if we move to 1.3. Today we use ECDHE-ECDSA-AES256-GCM-SHA384, which is pretty strong one, but maybe in 1.3 there are better cipher suites?

Thanks. I have updated to TLS_AES_256_GCM_SHA384 as default cipher suite.

cjn:
> > I would consider to replace default cipher suite if we move to 1.3. Today we use ECDHE-ECDSA-AES256-GCM-SHA384, which is pretty strong one, but maybe in 1.3 there are better cipher suites?
> 
> Great idea @salieri11 ! I just found a [page discussing this](https://www.thesslstore.com/blog/tls-1-3-everything-possibly-needed-know/) and it looks like the closest to our current suite is `TLS_AES_256_GCM_SHA384` (Key exchange and signature algos are no longer included).

Thanks. I have updated to TLS_AES_256_GCM_SHA384 since as suggested that looks the closest to current default cipher suite

PR #1554, opened 2021-06-07
fix new reconfiguration tests

yontyon:
We fix the following in the reconfiguration tests:
1. Wait for the right number of positive wedge status messages
2. Crash replica (in test_remove_nodes_with_failures) from the very beginning


nkumar04 (approved):
LGTM!!

PR #1555, opened 2021-06-07
Introduce Orrery: A Concord msg passing framework

andrewjstone:
This is a proposal for a message passing framework to be used in
Concord. There is a design document, as well as initial implementation
in the `orrery` folder. Some things, like the queue implementation
underlying mailboxes, are not optimized. However, the framework should
be fully usable as is.

The intention is to get people reviewing and playing with orrery and
suggesting changes. This PR will not be  immediately merged to give
reviewers time to do a thorough job and engage in discussion. As the
orrery code is entirely independent of any other code, a merge
would not be harmful. However, a merge would implicitly signify
acceptance of the proposal and so this PR will remain open for at least
a week or two.

There has been general high level consensus on building a message
passing system for about a year internally on the team. This is a
concrete design and implementation as a result of those discussions.

Please see the README in orerry/README.md for full details.


f-squirrel (commented):
Thank you @andrewjstone for this amazing document, it is very well structured and clearly written!
I have a few questions. 
* The `handle` function consumes a message, I am afraid we might have messages that can be processed by several components.
* Is View Change not expected to be another component?

Since we're talking about general design, the components are in fact state machines, so why not implement them as a state-machine design pattern with a clear transition table and relations between them? We can develop our own state machine framework or just use the one from the boost (it is efficient and has multiple features).

Another thing that came to my mind is that the `Consensus` component would fit better if it was representing a consensus(seq number) rather than the whole replica(`ReplicaImp`).
The consensus can be a state machine representing a single consensus per sequence number accepting messages related only to it and handling them. While `ReplicaImp` can hold all the consensus objects and routing a message to the corresponding consensus object. Obviously, the consensus object can serializable so that it can be stored in the reserved pages or whatever.

andrewjstone:
Thanks for the great feedback @f-squirrel. I think we are largely in agreement. My purpose in designing things the way I did was to ensure we didn't bite off more than we could chew and end up trying to rewrite all of concord at once. I have specific answers below that I hope make sense.

> Thank you @andrewjstone for this amazing document, it is very well structured and clearly written!
> I have a few questions.
> 
> * The `handle` function consumes a message, I am afraid we might have messages that can be processed by several components.

Yes, this is normal and expected. Messages are passed in an `Envelope` and routed appropriately. One example of this being useful is that we can send "system messages" to all components to do things like tell them to shutdown, or that a replica is malicious.

> * Is View Change not expected to be another component?
> 
> Since we're talking about general design, the components are in fact state machines, so why not implement them as a state-machine design pattern with a clear transition table and relations between them? We can develop our own state machine framework or just use the one from the boost (it is efficient and has multiple features).
> 

I see where you are going, but I view orrery at a higher level, and a somewhat intermediate step. We need to be able to fix our problems with multithreading and make it easier to separate logic out to their own threads. Importantly, we want to use our existing code without a complete rewrite. This very loose, high level structure avoids the need to structure the entire code base at once. I don't see how you could easily do that within the constraints of something like the boost FSM library.

> Another thing that came to my mind is that the `Consensus` component would fit better if it was representing a consensus(seq number) rather than the whole replica(`ReplicaImp`).
> The consensus can be a state machine representing a single consensus per sequence number accepting messages related only to it and handling them. While `ReplicaImp` can hold all the consensus objects and routing a message to the corresponding consensus object. Obviously, the consensus object can serializable so that it can be stored in the reserved pages or whatever.

This is essentially my longer term plan, and I definitely had our discussion from months ago in my head when building orrery. But without pulling out all the non-consensus logic, storage, etc... this will be very hard to do. So first we make consensus one component, then we split out things like clientManager and persistence into their own components. Then we can take the step of rewriting consensus. We can start that by having the consensus component become a manager of all the slot FSMs and driving them and distributing the messages that get output. But if we tried to do this right away, we would essentially be doing a complete rewrite of concord from scratch.

IMO we should use the boost fsm thing, or write our own that allows us to have per slot FSMs, but that is still just a subcomponent of orrery. Ideally, all state in those FSMs is in memory, and there is a shared, minimal context, and messages in a log that can be persisted by forwarding to another component responsible for storage. In this environment, we have orrery, and then the FSMs for complex logic underneath. Coarse and fine grained systems, where the fsm messages are simply a subset of the orrery messages.

tdimitrov:
Great work, Andrew. I like the approach because it will force decoupling of the different concord elements and hopefully will make the whole project easier to maintain.

One question - what's the motivation for separating World and Environment? They are a bit confusing for me. If I understand correctly world represents all components while environment creates mapping between component and mailbox. Can this happen in the same place?

orrery/README.md
```diff
@@ -0,0 +1,426 @@
...
+Like all distributed algorithms, the [sbft](https://arxiv.org/abs/1804.01626) algorithm implemented
+in concord is structured around message passing. The main replica thread contains a set of message
+handlers that farm work out to other parts of the system, including thread pools for signing and
+networking. Responses come back as messages to the main thread. Pre-execution, whereby execution
+runs concurrently before consensus, in an optimistic fashion, also relies on messages to communicate
+with the main replica thread and other parts of the system.
+
+When originally written, there was no pre-execution, and all storage access was single threaded. Due
+to performance pressures and feature additions, much single threaded code was incrementally made
+multi-threaded. In many cases this was done through thea addition of locks, and in some cases reuse
```
dartdart26, line 17: Typo: thea -> the

orrery/include/orrery/component.h
```diff
@@ -0,0 +1,69 @@
...
+#include "orrery_msgs.cmf.hpp"
+#include "Logger.hpp"
+#include <iostream>
+
+namespace concord::orrery {
+
+// A component provides an interface to handle `AllMsgs`
+class IComponent {
+ public:
+  virtual ~IComponent(){};
```
dartdart26, line 29: Nit: could use `= default` instead.

orrery/include/orrery/component.h
```diff
@@ -0,0 +1,69 @@
...
+  // Trait to ensure that a ComponentImpl has a `handle` method that takes a message of type `Msg`.
+  template <typename Impl, typename Msg, typename = std::void_t<>>
+  struct CanHandleMsgT : std::false_type {};
+
+  template <typename Impl, typename Msg>
+  struct CanHandleMsgT<
+      Impl,
+      Msg,
+      std::void_t<decltype(std::declval<Impl>().handle(std::declval<ComponentId>(), std::declval<Msg>()))>>
+      : std::true_type {};
```
dartdart26, line 45: I was wondering if the `CanHandleMsg` type should be out of `Component` or we should drop the `Impl` template parameter. I guess we can drop it as every `Component` for a specific `ComponentImpl` will have its own `CanHandleMsg`.

I only thought about it, because I wanted to find out if `Impl` is the same as `ComponentImpl`. What do you think?

orrery/include/orrery/component.h
```diff
@@ -0,0 +1,69 @@
...
+
+// A component provides an interface to handle `AllMsgs`
+class IComponent {
+ public:
+  virtual ~IComponent(){};
+  virtual void handle(ComponentId from, AllMsgs&&) = 0;
+};
+
+template <typename ComponentImpl>
+class Component : public IComponent {
```
dartdart26, line 34: I love this pattern - having runtime polymorphic behaviour and not forcing users to have virtual methods and allocate dynamically themselves!

orrery/include/orrery/component.h
```diff
@@ -0,0 +1,69 @@
...
+
+  // TODO: Recursive visitor to also check for non-top-level messages?
+  void handle(ComponentId from_id, AllMsgs&& msg_variant) override {
+    std::visit(
+        [from_id, this](auto&& msg) {
+          if constexpr (CanHandleMsgT<ComponentImpl, decltype(msg)>::value) {
+            this->impl_.handle(from_id, std::forward<decltype(msg)>(msg));
+          } else {
+            size_t to = static_cast<uint8_t>(impl_.id);
+            size_t from = static_cast<uint8_t>(from_id);
```
dartdart26, line 57: Nit: Can we maybe use `auto` and cast to `size_t` directly here, in order to avoid 2 casts?

orrery/include/orrery/component.h
```diff
@@ -0,0 +1,69 @@
...
+            LOG_ERROR(logger_, "Component cannot handle message: " << KVLOG(from, to, msg.id));
+          }
+        },
+        std::move(msg_variant.msg));
+  }
+
+ private:
+  ComponentImpl impl_;
+  logging::Logger logger_ = logging::getLogger("orrery.component");
```
dartdart26, line 66: Just came to mind. Does it make sense to somehow identify the component in the logger? Maybe making a static enum member for all `ComponentImpl`s or even a string name.

orrery/include/orrery/environment.h
```diff
@@ -0,0 +1,68 @@
...
+// An Environment is a static map of components to their executors.
+//
+// An Environment is meant to be created at startup and remain immutable for the lifetime of the
+// process.
+//
+// Executors communicate by placing envelopes in each other's mailboxes. Since executors can handle
+// messages for multiple components, multiple component ids may map to the same executor mailbox.
+//
+// As there are always a static number of components and executors for a given orrery world, we can
+// use a a std::array for the mapping.
```
dartdart26, line 35: Typo: redundant `a`.

orrery/include/orrery/environment.h
```diff
@@ -0,0 +1,68 @@
...
+class Environment {
+ public:
+  Environment() {
+    // The value doesn't matter. `enumSize` is a generated function that ignores the parameter.
+    ComponentId dummy = ComponentId::broadcast;
+    mailboxes_.resize(enumSize(dummy));
```
dartdart26, line 41: Nit: Not related to this PR, but I propose to name the function `enumCount()` as size can be mistaken to something like "size in bytes".

orrery/include/orrery/environment.h
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  void add(ComponentId id, Mailbox& mailbox) {
+    ConcordAssertNE(id, ComponentId::broadcast);
+    size_t index = static_cast<uint8_t>(id);
```
dartdart26, line 46: As above, I think we can avoid 2 casts.

orrery/include/orrery/mailbox.h
```diff
@@ -0,0 +1,39 @@
...
+#include "queue.h"
+
+#include "orrery_msgs.cmf.hpp"
+
+namespace concord::orrery {
+class Mailbox {
+ public:
+  void put(Envelope&& envelope);
+
+  const std::string& executorName() { return name_; }
```
dartdart26, line 25: Can be made `const`.

orrery/include/orrery/environment.h
```diff
@@ -0,0 +1,68 @@
...
+// use a a std::array for the mapping.
+class Environment {
+ public:
+  Environment() {
+    // The value doesn't matter. `enumSize` is a generated function that ignores the parameter.
+    ComponentId dummy = ComponentId::broadcast;
+    mailboxes_.resize(enumSize(dummy));
+  }
+
+  void add(ComponentId id, Mailbox& mailbox) {
```
dartdart26, line 44: Since we copy `mailbox`, I guess we can accept it as a const ref instead.

orrery/include/orrery/executor.h
```diff
@@ -0,0 +1,67 @@
...
+  Executor(const std::string& name) : name_(name), queue_(std::make_shared<detail::Queue>()), mailbox_(name, queue_) {}
+
+  Mailbox& mailbox() { return mailbox_; }
+
+  void init(const Environment& environment) {
+    // The value doesn't matter. `enumSize` is a generated function that ignores the parameter.
+    ComponentId dummy = ComponentId::broadcast;
+    components_.resize(enumSize(dummy));
+    environment_ = environment;
+  }
```
dartdart26, line 45: Since the queue is shared anyway, I guess who "owns" it doesn't really matter. Maybe it can be a member of `Mailbox` and then we can construct a `Mailbox` separately, passing it  and `Environment` to `Executor`, in order to avoid `init()`. Just a thought. Let me know if it makes sense.

```
 auto mailbox1 = Mailbox("replica_executor");
 auto mailbox2 = Mailbox("state_transfer_executor");

 auto env = Environment();
 env.add(ComponentId::replica, mailbox1);
 env.add(ComponentId::state_transfer, exec2);

 auto exec1 = Executor(env , mailbox1);
 auto exec2 = Executor(env , mailbox2);

```

orrery/include/orrery/executor.h
```diff
@@ -0,0 +1,67 @@
...
+
+  void add(ComponentId id, std::unique_ptr<IComponent>&& component) {
```
dartdart26, line 47: I think using an rvalue reference makes it a bit confusing. Can we just take it by value? Should be the same as unique_ptr only has a move ctor anyway.

orrery/include/orrery/executor.h
dartdart26: I guess the problem now is that one can pass the same mailbox to two or more executors. However, the user can also mess up the call to Environment::add() by passing the wrong mailbox.

orrery/src/executor.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "orrery_msgs.cmf.hpp"
+
+using namespace std::chrono_literals;
+
+namespace concord::orrery {
+
+static constexpr std::chrono::milliseconds TIMEOUT = 100ms;
+
+std::thread Executor::start() && {
+  return std::thread([executor{std::move(*this)}]() mutable {
```
dartdart26, line 26: Would it make sense to call `executor.join()` instead in main(), avoiding the moved-from executor? And having the thread as a member of `Executor`.

dartdart26 (approved):
As we've already discussed, I think this is a great framework for its simplicity and potential to make our code easier to reason about, make adding new features easier and make debugging simpler! Great work and thank you for it!

I have added some minor code comments that you can choose to address or not. I think we can merge as is and from there on add more unit tests, more examples and start experimenting with it.

andrewjstone:
> One question - what's the motivation for separating World and Environment? They are a bit confusing for me. If I understand correctly world represents all components while environment creates mapping between component and mailbox. Can this happen in the same place?

This separation is mainly for access control. The term 'World' is probably a terrible choice of name. The only thing a world is useful for is sending messages to other components. It probably should be called `Sender`. I named it `World`, because I originally envisioned other mechanisms for components to interact with the system like "raise alarms about corruption or overflow" that could bypass the message sending somehow (setting an atomic, polled by each executor?). However, I don't think that's the right thing to do. Those can all just be messages sent via `broadcast`. A rename to `Sender` would thus make the intent clearer.

Environments are setup at start time and they must be static when components start interacting. If components had access to the Environment API they could perform actions they shouldn't be allowed to inside the system.

orrery/include/orrery/component.h
andrewjstone: Ah yes, I think you are right and we can drop it. IIRC `Impl` should be the same as `ComponentImpl`.

The more important thing that I didn't implement and I think is important is to allow nested visitation for subvariants. That would be super powerful. There's a comment in one of the examples.

orrery/include/orrery/component.h
andrewjstone: IIRC I used size_t explicitly because uint8_t is the proper type to cast to, but it doesn't print correctly.

orrery/include/orrery/component.h
andrewjstone: Yes! Good idea. Maybe using the `MDB` macro?

orrery/include/orrery/environment.h
andrewjstone: Sure.

orrery/include/orrery/executor.h
andrewjstone: I'm open either way. I don't remember the exact reason I did it this way.

orrery/include/orrery/executor.h
andrewjstone: You are correct!

orrery/src/executor.cpp
andrewjstone: Yes, I thought about this afterwards as well. I believe this is the correct thing to do. This pattern is a bit confusing. I will leave this as follow up to whoever takes ownership of orrery 😛

PR #1556, opened 2021-06-07
Fix a few bugs in the diagnostics_server

andrewjstone:
It's possible that if the readline function, including read itself takes
longer than a second, that the remaining time can become negative and
cause select to fail in the next iteration of the loop. We prevent that
with a check at the bottom of the readline function.

Pass `errno` to `errnoString` instead of the return value from
failing libc calls.

Use the correct type for the return value of `read` and also for `count`.

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -45,7 +45,7 @@ static logging::Logger logger = logging::getLogger("concord.diagnostics");
 inline std::string readline(int sock) {
   std::array<char, MAX_INPUT_SIZE> buf;
   buf.fill(0);
-  int count = 0;
+  size_t count = 0;
   auto start = std::chrono::steady_clock::now();
   auto timeout = std::chrono::microseconds(999999);
   auto remaining = timeout;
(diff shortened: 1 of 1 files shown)


diagnostics/include/diagnostics_server.h
```diff
@@ -62,16 +62,16 @@ inline std::string readline(int sock) {
     }
     if (rv < 0 && errno == EINTR) continue;
     if (rv < 0) {
-      throw std::runtime_error("diagnostics server readline select failed: " + errnoString(rv));
+      throw std::runtime_error("diagnostics server readline select failed: " + errnoString(errno));
```
Liorgal28, line 65: I'm not sure if it has any difference but why here and in line 74 we use '+' between the two parts and not '<<'?

Liorgal28 (approved):

diagnostics/include/diagnostics_server.h
andrewjstone: It's because in the other lines we are using the logging API which uses `<<`. Here and in line 74 we are building a string through concatenation. We aren't "streaming".

PR #1559, opened 2021-06-08
Unordered_map race condition fix

ukg19:
This PR fixes a race-condition observed in reading and writing an unordered_map, while running thread sanitiser on back of crash observed in Concord.

In favour of performance, have avoided using locks. Instead, I have made value part of unordered_map as atomic so that reader thread doesn't read a stale value while writer has updated the value for that key.

### DIFF
--- a/bftengine/src/bcstatetransfer/MsgsCertificate.hpp
+++ b/bftengine/src/bcstatetransfer/MsgsCertificate.hpp
@@ -18,6 +18,7 @@
 #include <cstring>
 #include <unordered_map>
 #include <forward_list>
+#include <atomic>
 
 namespace bftEngine {
 namespace bcst {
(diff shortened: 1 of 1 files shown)


f-squirrel (commented):
@ukg19 , is there a possibility that threads add new key-values to the map while another thread(s) is reading/updating the values? I am asking because if it is true, then I think we do need to secure the map with a lock.
WDYT?

ukg19:
@f-squirrel yeah one thread is writing/updating into map while other is reading. I guess lock is not necessary because reading from unordered_map is thread-safe. Iterators to elements must stay valid when inserting or deleting other elements. So I think lock is not required. Am I missing something here? Looking for opinions.

f-squirrel:
Looks legit!

f-squirrel (approved):

PR #1563, opened 2021-06-09
Fix count calculation in diagnostics readline()

dartdart26:
Add the `read_rv` return value from read() to `count` instead of the
`rv` return value from select().

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -69,11 +69,11 @@ inline std::string readline(int sock) {
       throw std::runtime_error("Request exceeded max size: " + std::to_string(MAX_INPUT_SIZE));
     }
 
-    ssize_t read_rv = read(sock, buf.data() + count, buf.size() - count);
+    const auto read_rv = read(sock, buf.data() + count, buf.size() - count);
     if (read_rv <= 0) {
       throw std::runtime_error("diagnostics server read failed: " + errnoString(errno));
     }
-    count += rv;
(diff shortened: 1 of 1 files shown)


diagnostics/include/diagnostics_server.h
```diff
@@ -69,11 +69,11 @@ inline std::string readline(int sock) {
       throw std::runtime_error("Request exceeded max size: " + std::to_string(MAX_INPUT_SIZE));
     }
 
-    ssize_t read_rv = read(sock, buf.data() + count, buf.size() - count);
+    const auto read_rv = read(sock, buf.data() + count, buf.size() - count);
```
yuliasherman, line 72: How is it different from the previous definition? It is still ssize_t.

andrewjstone, line 72: I specifically put `ssize_t` there, because auto creates an `int` which is a narrowing conversion in some cases.

andrewjstone (approved):
Great find and fix @dartdart26. This was introduced with my latest change.

diagnostics/include/diagnostics_server.h
dartdart26: Do you refer to `read_rv` being `int`? It should be the return type of read(), i.e. `ssize_t`. I suspect I didn't get what you meant, though. Please correct me if I am wrong.

andrewjstone: You are correct. Sorry, I was thinking of something else.

PR #1564, opened 2021-06-09
Reconfiguration scale-up with f failure apollo test

nkumar04:
Added apollo test to add node with f failures


yontyon (approved):
LGTM!
I would also like us to have a test where we start the crashed replica after the reconfiguration action, but this is not possible as long as we rely on state transfer to get the new state

PR #1570, opened 2021-06-10
Fix skvbc_pyclient_tests.test_m_of_n_quorum 

nkgeorgiev:
The test creates a write request and waits for 2f + c + 1 quorum.
After that it creates a read requests but it waits for only 1 quorum.
In rare cases we may get a "split brain" and read from a slower replica that hasn't written the first request yet.
With this change, the read quorum is changed from 1 to f+1

### DIFF
--- a/tests/apollo/test_skvbc_pyclient.py
+++ b/tests/apollo/test_skvbc_pyclient.py
@@ -178,11 +178,12 @@ class SkvbcPyclientTest(unittest.TestCase):
         kv_pair = [(key, value)]
         await client.write(protocol.write_req([], kv_pair, 0))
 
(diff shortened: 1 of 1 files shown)


teoparvanov (approved):
LGTM, thanks @nkgeorgiev !

upshaw-alex:
BTW, thank you for this change. I recently migrated this test case from an implementation outside of Apollo and in PR #1547, and I had some concerns about this stability of this particular test case (see the discussion at https://github.com/vmware/concord-bft/pull/1547#discussion_r645194708), but that PR somehow got merged anyway. Thank you very much for your analysis of the instability here and for proactively revising this test case.

PR #1573, opened 2021-06-10
Logging: Change default format

robem:
This change aligns with the closed-source logging format. The main motivation
is to print the timestamp first so that messages can be sorted by the order in
which they appear. Having the replcia ID in the front makes sense when all
replicas print to the same destination. However, that is not the case and when
used in a docker-compose envrionment then it gets prefixed with the container
id anyways.

### DIFF
--- a/logging/src/Logging4cplus.cpp
+++ b/logging/src/Logging4cplus.cpp
@@ -24,7 +24,7 @@ using namespace log4cplus;
 
 namespace logging {
 
-static const char* logPattern = "%X{rid}|%d{%m-%d-%Y %H:%M:%S.%q}|%-5p|%c|%X{cid}|%X{sn}|%X{thread}|%b:%L|%M|%m%n";
+static const char* logPattern = "%d{%Y-%m-%dT%H:%M:%S,%qZ}|%-5p|%X{rid}|%c|%X{thread}|%X{cid}|%X{sn}|%b:%L|%M|%m%n";
 
 void initLogger(const std::string& configFileName) {
   std::ifstream infile(configFileName);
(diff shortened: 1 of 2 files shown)


logging/src/Logging4cplus.cpp
```diff
@@ -24,7 +24,7 @@ using namespace log4cplus;
 
 namespace logging {
 
-static const char* logPattern = "%X{rid}|%d{%m-%d-%Y %H:%M:%S.%q}|%-5p|%c|%X{cid}|%X{sn}|%X{thread}|%b:%L|%M|%m%n";
+static const char* logPattern = "%d{%Y-%m-%dT%H:%M:%S,%qZ}|%-5p|%X{rid}|%c|%X{thread}|%X{cid}|%X{sn}|%b:%L|%M|%m%n";
```
toly-kournik, line 27: Note: this only a default format for ConsoleAppender.
Actual runtime configuration is in the configuration file.

toly-kournik (approved):
LGTM

teoparvanov:
Hi @robem , can you please double-check with Harsh (@bhavsi ), to make sure this change won't affect how the logs are being parsed by fluentd ? Thanks!

logging/src/Logging4cplus.cpp
robem: Thanks for pointing this out. Yes, this is intentional but when I was playing around with logging w/o a config file I wanted to see the same output :)

bhavsi:
> Hi @robem , can you please double-check with Harsh (@bhavsi ), to make sure this change won't affect how the logs are being parsed by fluentd ? Thanks!

Hi @robem, since the order of the keys are being changed, we'll need a change in the fluentd config parsing order as well.

bhavsi:
Hi @teoparvanov, had a chat with @robem and checked that the internal properties file format aligns with Fluentd configuration, and we should be good!

PR #1574, opened 2021-06-11
Cherry picking Cryptosystems rebuild on startup fix. (#1568)

toly-kournik:
By design every subsequent cryptosystem is based on the previous one to include all its keys for starters.
Upon replica start from an existing state "replay" cryptosystems creation by adding public keys in an order of appearence, i.e. by seqnum.


teoparvanov (approved):
LGTM, thanks Toly!

PR #1575, opened 2021-06-11
Views manager refactoring

StefanPalashev:
The purpose of this Pull Request is to make a step towards making the ViewsManager module unit testable. In order to do so, the mentioned module's logic has to be pulled apart from the one that resides in ReplicaImp.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -594,7 +594,7 @@ void ReplicaImp::sendInternalNoopPrePrepareMsg(CommitPath firstPath) {
     return;
   }
   PrePrepareMsg *pp = new PrePrepareMsg(
-      config_.getreplicaId(), curView, (primaryLastUsedSeqNum + 1), firstPath, sizeof(ClientRequestMsgHeader));
+      config_.getreplicaId(), getCurrentView(), (primaryLastUsedSeqNum + 1), firstPath, sizeof(ClientRequestMsgHeader));
   ClientRequestMsg emptyClientRequest(config_.getreplicaId());
   pp->addRequest(emptyClientRequest.body(), emptyClientRequest.size());
   pp->finishAddingRequests();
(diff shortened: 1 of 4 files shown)


bftengine/src/bftengine/ViewsManager.hpp
```diff
@@ -71,6 +71,9 @@ class ViewsManager {
                                                                           // in viewChangeMsgs
                                         std::vector<ViewChangeMsg *> viewChangeMsgs);
 
+  ViewNum getCurrentView() const { return myCurrentView; }
+  void setHigherView(ViewNum higherViewNum);
+  void setViewFromRecovery(ViewNum explicitViewNum) { myCurrentView = explicitViewNum; }
```
teoparvanov, line 76: `setCurrentView()` ?

bftengine/src/bftengine/ViewsManager.hpp
```diff
@@ -71,6 +71,9 @@ class ViewsManager {
                                                                           // in viewChangeMsgs
                                         std::vector<ViewChangeMsg *> viewChangeMsgs);
 
+  ViewNum getCurrentView() const { return myCurrentView; }
+  void setHigherView(ViewNum higherViewNum);
```
teoparvanov, line 75: `setHigerCurrentView()` maybe ?

teoparvanov (approved):
LGTM, thanks @StefanPalashev !

bftengine/src/bftengine/ViewsManager.hpp
StefanPalashev: The name of this function aims to state that it is called only when the replica is recovering. In this certain scenario the view number might be set to a specific value, which is not strictly higher than the one before (i.e. it might be the same value).

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -835,13 +836,14 @@ void ReplicaImp::tryToAskForMissingInfo() {
 
   SeqNum minSeqNum = 0;
   SeqNum maxSeqNum = 0;
+  const int16_t searchWindow = 32;  // TODO(GG): TBD - read from configuration
 
   if (!recentViewChange) {
-    const int16_t searchWindow = 32;  // TODO(GG): TBD - read from configuration
+    // const int16_t searchWindow = 32;  // TODO(GG): TBD - read from configuration
```
HristoStaykov, line 842: I'd recommend to remove the unused code.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -835,13 +836,14 @@ void ReplicaImp::tryToAskForMissingInfo() {
...
     minSeqNum = lastExecutedSeqNum + 1;
     maxSeqNum = std::min(minSeqNum + searchWindow - 1, lastStableSeqNum + kWorkWindowSize);
   } else {
-    const int16_t searchWindow = 32;  // TODO(GG): TBD - read from configuration
+    // const int16_t searchWindow = 32;  // TODO(GG): TBD - read from configuration
```
HristoStaykov, line 846: Same as above.

HristoStaykov (approved):
LGTM, thanks @StefanPalashev !

PR #1576, opened 2021-06-11
Change apollo skvbc_pyclient_tests.test_m_of_n_quorum to wait for 2f+c+1 read quorum

nkgeorgiev:
It turns out that the quorum change in https://github.com/vmware/concord-bft/pull/1570 was not enough to guarantee no stale reads.
Linearizable reads need to be 2f+1, otherwise we may get a stale read


teoparvanov (approved):
LGTM, thank you @nkgeorgiev !

PR #1577, opened 2021-06-11
Skip test_key_exchange_command_with_restart

nkgeorgiev:
This test is experiencing intermittent failures so skip it until it is fixed.


teoparvanov (approved):
LGTM, thanks!

PR #1579, opened 2021-06-13
Increase maxNumOfRequestsInBatch parameter

yuliasherman:
As an internal ReplicaImp queue size has been increased from 700 to 1500, maxNumOfRequestsInBatch needs to be larger, too. The number of requests in the consensus batch is controlled by maxExternalMessageSize and the maxNumOfRequestsInBatch parameter is required only to limit a memory allocation for the bitmap.


Liorgal28 (approved):
LGTM

PR #1581, opened 2021-06-13
Define an execution request related CMF msgs

cloudnoize:
1. Define the CMF mags for an execution request and its container.
2. Propogate the msg signature to the execution as well, in order to persist it on-chain.

### DIFF
--- a/bftengine/cmf/execution_data.cmf
+++ b/bftengine/cmf/execution_data.cmf
@@ -1,0 +1,24 @@
+
+Msg RequestInput 1 {
+    uint16 clientId
+    uint64 executionSequenceNum
+    string cid
+    uint64 flags
+    string request
+    string signature
+    uint64 requestSequenceNum
+}
+
+Msg TimeService 2 {
+    int64 time_since_epoch
+    # If the execution mode is not accumulated i.e. execution of a single request per execute, the command handler 
+    # needs to know the request_position of the request in the list.
+    uint64 request_position
+}
+
+Msg RequestsRecord 3 {
+    list RequestInput requests
+    optional TimeService timestamp
+    uint64 keys_version
+    bool keys_rotated_at_this_block 
+}
--- a/bftengine/cmf/CMakeLists.txt
+++ b/bftengine/cmf/CMakeLists.txt
@@ -2,3 +2,9 @@ cmf_generate_cpp(header cpp concord::messages::keys_and_signatures keys_and_sign
 add_library(keys_and_signatures_cmf ${cpp})
 set_target_properties(keys_and_signatures_cmf PROPERTIES LINKER_LANGUAGE CXX)
 target_include_directories(keys_and_signatures_cmf PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
+
+
+cmf_generate_cpp(header cpp concord::messages::execution_data execution_data.cmf)
+add_library(execution_data_cmf ${cpp})
+set_target_properties(execution_data_cmf PROPERTIES LINKER_LANGUAGE CXX)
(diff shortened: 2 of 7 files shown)


bftengine/cmf/execution_data.cmf
```diff
@@ -0,0 +1,22 @@
...
+}
+
+Msg Timestamp 2 {
+    int64 time_since_epoch
+    uint64 request_position
+}
+
+Msg ExecutedRequests 2 {
+    list RequestInput Requests
+    Timestamp timestamp
```
f-squirrel, line 19: The `Timestamp` has to be optional because the Time Service is optional itself.

bftengine/cmf/execution_data.cmf
```diff
@@ -0,0 +1,22 @@
+
+Msg RequestInput 1 {
```
f-squirrel, line 2: Nit: maybe we can simplify`RequestInput` => just `Request` ?

f-squirrel (changes requested):

bftengine/cmf/execution_data.cmf
cloudnoize: right, thanks!

bftengine/cmf/execution_data.cmf
cloudnoize: I want to emphasize that it's the input data part of the execution request since the struct itself ```struct ExecutionRequest``` contains also output data.

f-squirrel (approved):
LGTM, @cloudnoize !

bftengine/cmf/execution_data.cmf
```diff
@@ -0,0 +1,22 @@
...
+    string cid
+    uint64 flags
+    string request
+    string signature
+    uint64 requestSequenceNum
+}
+
+Msg Timestamp 2 {
+    int64 time_since_epoch
+    uint64 request_position
```
robem, line 14: What is request_position in the context of a timestamp? Btw, CMF allows comments :)

cloudnoize: The new time may add a small amount of time to the given ```time_since_epoch``` per request i.e. if a request is the 3rd in the queue the command handler will calculate its execution time as ```time_since_epoch``` + 3*epsilon.

If the execution mode is not accumulated i.e. execution of a single request per execute, the command handler needs to know the ```request_position``` of the request in the list.   
@f-squirrel

robem: That makes sense to me. However, the message organization doesn't make sense to me yet.

The `ExecutedREquests` message contains a list of requests and a single timestamp with a `request_position`. How does the single request_position relate to the list of requests? This might just be a terminology issue.

bftengine/cmf/execution_data.cmf
```diff
@@ -0,0 +1,22 @@
...
+}
+
+Msg ExecutedRequests 3 {
+    list RequestInput Requests
```
robem, line 18: `requests` (lower case)

bftengine/cmf/execution_data.cmf
f-squirrel: @robem , `request_position` represents the position of `ExecutedREquests` in the PrePrepare.
I agree that the name does not fully explain the purpose of the field but I could not find a better one.

cloudnoize: I will add a comment as it's not a trivial field

bftengine/cmf/execution_data.cmf
cloudnoize: thanks

PR #1582, opened 2021-06-14
Pruning via reconfiguration

yontyon:
In this PR we complete the implementation of pruning via reconfiguration. The main purpose of this PR is to move the functionality of pruning after state transfer and on startup to the reconfiguration mechanism:
1. Writing the prune request under the following key std::string{0x24, 0x1} (for backward compatibility)
2. Adding pruning handler to the state transfer reconfiguration handler
3. Remove the functionality for recovering and state transfer from the default pruning handler
4. Adapting the default pruning tests to these changes
5. Add Apollo test to test the recovery mechanism with state transfer

### DIFF
--- a/kvbc/src/Replica.cpp
+++ b/kvbc/src/Replica.cpp
@@ -70,14 +70,17 @@ Status Replica::start() {
 
 void Replica::createReplicaAndSyncState() {
   auto requestHandler = bftEngine::IRequestsHandler::createRequestsHandler(m_cmdHandler, cronTableRegistry_);
+  stReconfigurationSM_->registerHandler(requestHandler->getReconfigurationHandler());
   stReconfigurationSM_->registerHandler(m_cmdHandler->getReconfigurationHandler());
   requestHandler->setReconfigurationHandler(
       std::make_shared<kvbc::reconfiguration::ReconfigurationHandler>(*this, *this));
   requestHandler->setReconfigurationHandler(
       std::make_shared<kvbc::reconfiguration::InternalKvReconfigurationHandler>(*this, *this),
       concord::reconfiguration::ReconfigurationHandlerType::PRE);
-  requestHandler->setReconfigurationHandler(std::shared_ptr<kvbc::pruning::PruningHandler>(
-      new concord::kvbc::pruning::PruningHandler(*this, *this, *this, *m_stateTransfer, true)));
+  auto pruning_handler = std::shared_ptr<kvbc::pruning::PruningHandler>(
+      new concord::kvbc::pruning::PruningHandler(*this, *this, *this, *m_stateTransfer, true));
+  requestHandler->setReconfigurationHandler(pruning_handler);
+  stReconfigurationSM_->registerHandler(pruning_handler);
   m_replicaPtr = bftEngine::IReplica::createNewReplica(
       replicaConfig_, requestHandler, m_stateTransfer, m_ptrComm, m_metadataStorage, pm_, secretsManager_);
   const auto lastExecutedSeqNum = m_replicaPtr->getLastExecutedSequenceNum();
--- a/kvbc/src/Replica.cpp
+++ b/kvbc/src/Replica.cpp
@@ -100,10 +100,6 @@ void Replica::createReplicaAndSyncState() {
       std::terminate();
     }
   }
-  // the stReconfigurationSM_ also serves as a recovering mechanism if the replica crashed before/after the
-  // reconfiguration request. So once we done initialize the replica, we try to execute its main handler
-  uint64_t lastCheckpointNum = lastExecutedSeqNum / checkpointWindowSize;
-  stReconfigurationSM_->stCallBack(lastCheckpointNum);
 }
 
 /**
@@ -286,12 +282,12 @@ Replica::Replica(ICommunication *comm,
   auto stKeyManipulator = std::shared_ptr<storage::ISTKeyManipulator>{storageFactory->newSTKeyManipulator()};
   m_stateTransfer = bftEngine::bcst::create(stConfig, this, m_metadataDBClient, stKeyManipulator, aggregator_);
   m_metadataStorage = new DBMetadataStorage(m_metadataDBClient.get(), storageFactory->newMetadataKeyManipulator());
+  bftEngine::IControlHandler::instance(new bftEngine::ControlHandler());
   if (!replicaConfig.isReadOnly) {
     stReconfigurationSM_ = std::make_unique<concord::kvbc::StReconfigurationHandler>(*m_stateTransfer, *this);
   }
   // Instantiate IControlHandler.
   // If an application instantiation has already taken a place this will have no effect.
-  bftEngine::IControlHandler::instance(new bftEngine::ControlHandler());
 }
 
 Replica::~Replica() {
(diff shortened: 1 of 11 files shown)


kvbc/src/Replica.cpp
```diff
@@ -279,12 +282,12 @@ Replica::Replica(ICommunication *comm,
   auto stKeyManipulator = std::shared_ptr<storage::ISTKeyManipulator>{storageFactory->newSTKeyManipulator()};
   m_stateTransfer = bftEngine::bcst::create(stConfig, this, m_metadataDBClient, stKeyManipulator, aggregator_);
   m_metadataStorage = new DBMetadataStorage(m_metadataDBClient.get(), storageFactory->newMetadataKeyManipulator());
+  bftEngine::IControlHandler::instance(new bftEngine::ControlHandler());
   if (!replicaConfig.isReadOnly) {
     stReconfigurationSM_ = std::make_unique<concord::kvbc::StReconfigurationHandler>(*m_stateTransfer, *this);
   }
   // Instantiate IControlHandler.
   // If an application instantiation has already taken a place this will have no effect.
```
dartdart26, line 290: I think we should move the comment too. Btw, just out of curiosity, what is the reason for moving it?

kvbc/src/pruning_handler.cpp
```diff
@@ -192,9 +180,7 @@ bool PruningHandler::handle(const concord::messages::PruneRequest& request,
   }
 
   const auto latest_prunable_block_id = agreedPrunableBlockId(request);
-  // Make sure we have persisted the agreed prunable block ID before proceeding.
-  // Rationale is that we want to be able to pick up in case of a crash.
-  persistLastAgreedPrunableBlockId(latest_prunable_block_id, bftSeqNum);
```
dartdart26, line 197: Is that replaced by persisting the `PruneRequest` reconfiguration command?

kvbc/src/st_reconfiguration_sm.cpp
```diff
@@ -48,6 +47,13 @@ void StReconfigurationHandler::stCallBack(uint64_t current_cp_num) {
                                                             current_cp_num);
   handlerStoredCommand<concord::messages::AddRemoveWithWedgeCommand>(
       std::string{kvbc::keyTypes::reconfiguration_add_remove, 0x1}, current_cp_num);
+  handlerStoredCommand<concord::messages::PruneRequest>(std::string{kvbc::keyTypes::reconfiguration_pruning_key, 0x1},
+                                                        current_cp_num);
+}
+
+void StReconfigurationHandler::pruneOnStartup() {
```
dartdart26, line 54: I was wondering who calls this function. Could you please elaborate?

kvbc/src/st_reconfiguration_sm.cpp
```diff
@@ -82,4 +88,17 @@ bool StReconfigurationHandler::handle(const concord::messages::AddRemoveWithWedg
   return true;
 }
 
+bool StReconfigurationHandler::handle(const concord::messages::PruneRequest &command, uint64_t bft_seq_num, uint64_t) {
+  // Actual pruning will be done from the lowest latestPruneableBlock returned by the replicas. It means, that even
+  // on every state transfer there might be at most one relevant pruning command. Hence it is enough to take the latest
+  // saved command and try to execute it
```
dartdart26, line 94: This method will be called when ST completes and it will get the latest version of the key that contains the prune command. Then, it will prune up to the last agreed block ID in that command. Is my understanding correct?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -489,6 +489,66 @@ async def test_pruning_command(self, bft_network):
...
+            op = operator.Operator(bft_network.config, client,  bft_network.builddir)
+            await op.latest_pruneable_block()
+
+            latest_pruneable_blocks = []
+            rsi_rep = client.get_rsi_replies()
+            for r in rsi_rep.values():
+                lpab = cmf_msgs.ReconfigurationResponse.deserialize(r)[0]
+                latest_pruneable_blocks += [lpab.response]
+
+            # Now, crash on of the non-primary replicas
```
dartdart26, line 516: Nit: Typo, on -> one

kvbc/src/Replica.cpp
yontyon: there is no real reason.
I will return it to where it was

kvbc/src/pruning_handler.cpp
yontyon: yes, but currently we save the command itself under a different key to prevent issues with concord pruing handler

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -489,6 +489,66 @@ async def test_pruning_command(self, bft_network):
...
+            crashed_replica = 3
+            bft_network.stop_replica(crashed_replica)
+            await op.prune(latest_pruneable_blocks)
+            rsi_rep = client.get_rsi_replies()
+            # we expect to have at least 2f + 1 replies
+            for rep in rsi_rep:
+                r = rsi_rep[rep]
+                data = cmf_msgs.ReconfigurationResponse.deserialize(r)[0]
+                pruned_block = int(data.additional_data.decode('utf-8'))
+                assert pruned_block <= 90
```
dartdart26, line 526: Why do we assert for less than or equal to 90?

kvbc/src/st_reconfiguration_sm.cpp
yontyon: The replica calls this method once it is starting. 
This is for the case where the replica is restarting during the pruning

kvbc/src/st_reconfiguration_sm.cpp
yontyon: exactly

tests/apollo/test_skvbc_reconfiguration.py
yontyon: because we create at most 90 blocks at the beginning of the test

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -489,6 +489,66 @@ async def test_pruning_command(self, bft_network):
...
+            # creates 100 new blocks
+            for i in range(100):
+                v = skvbc.random_value()
+                await client.write(skvbc.write_req([], [(k, v)], 0))
+
+            # now, return the crashed replica and wait for it to done with state transfer
+            bft_network.start_replica(crashed_replica)
+            await self._wait_for_st(bft_network, crashed_replica, 150)
+
+            # We expect the late replica to catch up with the state and to perform pruning
```
dartdart26, line 537: Nice test.

kvbc/src/st_reconfiguration_sm.cpp
dartdart26: Yep :)

tests/apollo/test_skvbc_reconfiguration.py
dartdart26: Don't we create 100 blocks in the beginning? Also, how many are expected to be pruned?

kvbc/src/st_reconfiguration_sm.cpp
yontyon: fixed

tests/apollo/test_skvbc_reconfiguration.py
yontyon: fixed

tests/apollo/test_skvbc_reconfiguration.py
yontyon: Oh, sorry. you right,
We create 100 and the default blocks to keep is 10 (in https://github.com/vmware/concord-bft/blob/master/tests/simpleKVBC/TesterReplica/setup.cpp#L60)

dartdart26: Thanks!

dartdart26 (approved):
Looks great, thanks @yontyon!

PR #1583, opened 2021-06-14
Test disabled because of intermittent failures.

HristoStaykov:
This test covers an edge case scenario, which prior to the fix for the situation used to fail consistently.
Now we have intermittent failures only, which require further analysis.
Until the intermittent failures are understood and resolved the test won't be part of CI.


teoparvanov (approved):
LGTM, thanks Hristo!

PR #1586, opened 2021-06-15
On ST's consistency check, take pruning under consideration

yontyon:
When we check the consistency of ST we must consider the case of pruning. In this case, the genesis block has changed, hence we need to consider this.


toly-kournik (approved):
LGTM!

PR #1587, opened 2021-06-15
do not check blocks that are smaller than the current genesis

yontyon:
cherry-pick https://github.com/vmware/concord-bft/pull/1586


toly-kournik (approved):
LGTM!

PR #1592, opened 2021-06-16
Persist the last KVBC block ID in metadata

dartdart26:
Perist the last KVBC block ID in metadata as part of the effort to
change replica state sync so that we no longer persist the last executed
BFT sequence number in KVBC blocks.

Introduce an user data field in metadata as a
`ConstMetadataParameterIds` type. Reduce `reservedSimpleParamsNum` to
ensure adding user data is backwards compatible.

Persist the last block ID in metadata in two cases:
 * once request processing has completed, in
   `IRequestsHandler::onFinishExecutingReadWriteRequests()`.
 * once state transfer completes, potentially adding new blocks

A subsequent commit will change replica state sync so that it uses the
block ID instead of the BFT sequence number.

Improve const-correctness in MetadataStorage as part of this commit.

### DIFF
--- a/bftengine/src/bftengine/PersistentStorage.hpp
+++ b/bftengine/src/bftengine/PersistentStorage.hpp
@@ -110,6 +110,7 @@ class PersistentStorage {
   virtual void setCheckpointMsgInCheckWindow(SeqNum seqNum, CheckpointMsg *msg) = 0;
   virtual void setCompletedMarkInCheckWindow(SeqNum seqNum, bool mark) = 0;
 
+  // User data to be persisted, e.g. application-specific local data, scratchpad data, etc.
   virtual void setUserData(const void *data, std::size_t numberOfBytes) = 0;
 
   virtual void setEraseMetadataStorageFlag() = 0;
--- a/bftengine/src/bftengine/PersistentStorage.hpp
+++ b/bftengine/src/bftengine/PersistentStorage.hpp
@@ -19,6 +19,7 @@
 #include "ReplicaConfig.hpp"
 #include "PersistentStorageDescriptors.hpp"
 
+#include <cstdint>
 #include <vector>
 
 namespace bftEngine {
@@ -109,6 +110,8 @@ class PersistentStorage {
   virtual void setCheckpointMsgInCheckWindow(SeqNum seqNum, CheckpointMsg *msg) = 0;
   virtual void setCompletedMarkInCheckWindow(SeqNum seqNum, bool mark) = 0;
 
+  virtual void setUserData(const void *data, std::size_t numberOfBytes) = 0;
+
   virtual void setEraseMetadataStorageFlag() = 0;
   virtual bool getEraseMetadataStorageFlag() = 0;
   virtual void eraseMetadata() = 0;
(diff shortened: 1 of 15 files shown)


bftengine/src/bftengine/PersistentStorage.hpp
```diff
@@ -109,6 +110,8 @@ class PersistentStorage {
   virtual void setCheckpointMsgInCheckWindow(SeqNum seqNum, CheckpointMsg *msg) = 0;
   virtual void setCompletedMarkInCheckWindow(SeqNum seqNum, bool mark) = 0;
 
+  virtual void setUserData(const void *data, std::size_t numberOfBytes) = 0;
```
yuliasherman, line 113: Please add an explanation what is the purpose of this data.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -812,6 +816,14 @@ bool PersistentStorageImp::getCompletedMarkInCheckWindow(SeqNum seqNum) {
   return readCompletedMarkFromDisk(seqNum);
 }
 
+std::vector<std::uint8_t> PersistentStorageImp::getUserData() const {
+  auto buf = std::vector<std::uint8_t>(2048);
```
yuliasherman, line 820: Define 2048 as a constant variable

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -812,6 +816,14 @@ bool PersistentStorageImp::getCompletedMarkInCheckWindow(SeqNum seqNum) {
...
+  auto actualSize = std::uint32_t{0};
+  metadataStorage_->read(USER_DATA, buf.size(), reinterpret_cast<char *>(buf.data()), actualSize);
+  buf.resize(actualSize);
```
yuliasherman, line 823: Consider replacing vector by queue/deque to avoid resizing

bftengine/src/bftengine/PersistentStorageImp.hpp
```diff
@@ -64,6 +64,7 @@ enum ConstMetadataParameterIds : uint32_t {
   LAST_VIEW_TRANSFERRED_SEQ_NUM = 6,
   LAST_STABLE_SEQ_NUM = 7,
   ERASE_METADATA_ON_STARTUP = 9,
+  USER_DATA = 10,
```
yuliasherman, line 67: USER_DATA -> USER_SPECIFIC_DATA?

yuliasherman (approved):
LGTM with some minor comments

bftengine/src/bftengine/PersistentStorage.hpp
dartdart26: I've added a general description of what that data might be, e.g. application-specific local data, scratchpad data, etc.

bftengine/src/bftengine/PersistentStorageImp.cpp
dartdart26: Done.

bftengine/src/bftengine/PersistentStorageImp.cpp
dartdart26: I am allocating a buffer, because `read()`'s API requires a `char*` buffer to a contiguous sequence and a length. The resize() should just reduce it for user's convenience. Not sure I got what you meant and I might have misunderstood you, please correct me if I am wrong.

bftengine/src/bftengine/PersistentStorageImp.hpp
dartdart26: I prefer the shorter one, but I don't have a strong opinion. I can change if you feel strongly about it.

PR #1596, opened 2021-06-16
Fix data race on ClientsInfo's requestsInfo and repliesInfo maps.

HristoStaykov:
We need to sync access to those maps, because they get modified/queried
by multiple threads. We introduce 2 mutexes that guard each and
make sure we never hold them both at the same time for a given
ClientInfo object.repliesInfo.

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -97,14 +97,18 @@ void ClientsManager::loadInfoFromReservedPages() {
     ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
 
     // YS TBD: Multiple replies for client batching should be sorted by incoming time
-    auto& repliesInfo = indexToClientInfo_.at(clientIdx).repliesInfo;
-    if (repliesInfo.size() >= maxNumOfReqsPerClient_) deleteOldestReply(clientId);
-    const auto& res = repliesInfo.insert_or_assign(replyHeader->reqSeqNum, MinTime);
-    const bool added = res.second;
-    LOG_INFO(CL_MNGR, "Added/updated reply message" << KVLOG(clientId, replyHeader->reqSeqNum, added));
-
+    auto& clientInfo = indexToClientInfo_.at(clientIdx);
+    {
+      std::unique_lock<std::mutex> lock(clientInfo.repliesInfoGuard);
+      auto& repliesInfo = clientInfo.repliesInfo;
+      if (repliesInfo.size() >= maxNumOfReqsPerClient_) deleteOldestReply(clientId);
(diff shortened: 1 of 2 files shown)


KTeofanidis (commented):
Maybe we can use a shared_mutex for read/write access to improve performance.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -97,14 +97,18 @@ void ClientsManager::loadInfoFromReservedPages() {
...
-    LOG_INFO(CL_MNGR, "Added/updated reply message" << KVLOG(clientId, replyHeader->reqSeqNum, added));
-
+    auto& clientInfo = indexToClientInfo_.at(clientIdx);
+    {
+      std::unique_lock<std::mutex> lock(clientInfo.repliesInfoGuard);
+      auto& repliesInfo = clientInfo.repliesInfo;
+      if (repliesInfo.size() >= maxNumOfReqsPerClient_) deleteOldestReply(clientId);
+      const auto& res = repliesInfo.insert_or_assign(replyHeader->reqSeqNum, MinTime);
+      const bool added = res.second;
+      LOG_INFO(CL_MNGR, "Added/updated reply message" << KVLOG(clientId, replyHeader->reqSeqNum, added));
```
yuliasherman, line 107: Nit: the log could be put out of the lock. Actually, "const bool added = res.second;", too.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -124,13 +128,16 @@ bool ClientsManager::hasReply(NodeIdType clientId, ReqId reqSeqNum) {
     throw;
   }
 
-  const auto& repliesInfo = indexToClientInfo_.at(idx).repliesInfo;
+  auto& clientInfo = indexToClientInfo_.at(idx);
+  std::unique_lock<std::mutex> lock(clientInfo.repliesInfoGuard);
```
yuliasherman, line 132: Shared lock could be used here and for all reads

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -108,11 +108,13 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
...
 
   struct ClientInfo {
+    std::mutex requestsInfoGuard;
+    std::mutex repliesInfoGuard;
     std::map<ReqId, RequestInfo> requestsInfo;
     std::map<ReqId, Time> repliesInfo;  // replyId to replyTime
   };
 
-  std::vector<ClientInfo> indexToClientInfo_;
+  mutable std::deque<ClientInfo> indexToClientInfo_;
```
yuliasherman, line 117: I prefer deque over the vector, but why did you change it here?

yuliasherman (changes requested):

PR #1597, opened 2021-06-16
Fix data race on ClientsInfo's requestsInfo and repliesInfo maps.

HristoStaykov:
We need to sync access to those maps, because they get modified/queried
by multiple threads. We introduce 2 mutexes that guard each and
make sure we never hold them both at the same time for a given
ClientInfo object.repliesInfo.

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -97,14 +97,18 @@ void ClientsManager::loadInfoFromReservedPages() {
     ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
 
     // YS TBD: Multiple replies for client batching should be sorted by incoming time
-    auto& repliesInfo = indexToClientInfo_.at(clientIdx).repliesInfo;
-    if (repliesInfo.size() >= maxNumOfReqsPerClient_) deleteOldestReply(clientId);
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -124,13 +128,16 @@ bool ClientsManager::hasReply(NodeIdType clientId, ReqId reqSeqNum) {
     throw;
   }
 
-  const auto& repliesInfo = indexToClientInfo_.at(idx).repliesInfo;
+  auto& clientInfo = indexToClientInfo_.at(idx);
+  std::unique_lock<std::mutex> lock(clientInfo.repliesInfoGuard);
```
robem, line 132: Do you mind wrapping this block in `{ ... }` as well? This will make it visually clear which lines need to be guarded and which don't. Thereby, it will be easier to change/add code later. Same for other places where the lock is used at the end of the function.

robem (approved):
Lgtm! Thanks for the change, only one nit :)

dartdart26:
> We need to sync access to those maps, because they get modified/queried
> by multiple threads. We introduce 2 mutexes that guard each and
> make sure we never hold them both at the same time for a given
> ClientInfo object.repliesInfo

Thanks for spotting that!

Out of curiosity, which threads are accessing the clients manager concurrently? I had the same need and, if you remember, we introduced an internal message to only access it from the messaging thread. Can we use the same approach here as well, to avoid locks?

yuliasherman (approved):
LGTM. Have you launched a performance check through Maestro?

PR #1598, opened 2021-06-17
Refactor to remove functionality duplication from SkvbcTracker

KTeofanidis:
Some SkvbcTracker methods are essentially an instrumented version of similar methods in skvbc.py

We can remove code duplication by adding tracking in skvbc.py. All test functionality is the same. The goal is only cleaning up some code duplication.

Removed methods:

- write_and_track_known_kv
- read_and_track_known_kv
- tracked_prime_for_state_transfer
- wait_for_liveness
- tracked_read_your_writes


tests/apollo/util/skvbc.py
```diff
@@ -176,18 +177,65 @@ async def send_indefinite_write_requests(self, client=None, delay=.1):
...
+                client = self.bft_network.random_client()
+            if kv == None:
+                kv_input = False
+                key = self.random_key()
+                val = self.random_value()
+                kv = [(key, val)]
+            else:
+                kv_input = True
+            read_version = 0
+            if self.tracker != None:
```
teoparvanov, line 192: `if self.tracker is not None` sounds more like Python. Same comment throughout.

tests/apollo/test_skvbc_auto_view_change.py
```diff
@@ -129,4 +127,4 @@ async def test_auto_vc_all_nodes_up_fast_path(self, bft_network, tracker):
         await skvbc.assert_kv_write_executed(key, val)
         await bft_network.assert_fast_path_prevalent()
 
-        await tracker.tracked_read_your_writes()
+        await skvbc.read_your_writes(self)
```
teoparvanov, line 130: why do we need to pass the test class instance as a parameter ?

teoparvanov (approved):
LGTM overall, I just have a couple of minor comments/questions.

tests/apollo/test_skvbc_auto_view_change.py
KTeofanidis: I didn't change the original code for read_your_writes which uses test_class.assertEqual.

tests/apollo/util/skvbc.py
KTeofanidis: Agreed it's better style. I'll change it.

tests/apollo/test_skvbc_auto_view_change.py
teoparvanov: why can't we just use `assert` instead of `test_class.assertEqual` ?

teoparvanov (approved):
Awesome, thank you @KTeofanidis !

PR #1600, opened 2021-06-17
Ensure replicas are up on write ops in VC tests

dartdart26:
Make sure that replicas are up when sending write ops in view change
tests. Do that via nested `move_on_after()` calls in
`_send_random_writes()`, because the BFT client that sends the tracked
ops will terminate the test if it times out. Therefore, make sure we
never let it timeout by calling it for 1 second at a time and assuming
its timeout is more than 1 second.

Above is useful, because if replicas haven't started yet and we give the
BFT client 1 second to send a request, some replicas might not even
observe it and never trigger view change. Doing that in a loop for
`retry_for_seconds` makes the race condition less likely.

TODO: We might want to rewrite the tests so that they are not dependent
on timing.

Thanks to @HristoStaykov for finding the issue and suggesting this
improvement.

### DIFF
--- a/tests/apollo/test_skvbc_view_change.py
+++ b/tests/apollo/test_skvbc_view_change.py
@@ -549,10 +549,26 @@ class SkvbcViewChangeTest(unittest.TestCase):
                     else:
                         break
 
-    async def _send_random_writes(self, tracker):
-        with trio.move_on_after(seconds=1):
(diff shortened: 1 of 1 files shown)


HristoStaykov (approved):
LGTM, thanks @dartdart26!

PR #1601, opened 2021-06-17
Persist metadata user data in a transaction

dartdart26:
Persist the metadata user data in an already created transaction. That
ensures it is consistent with other metadata.

Move the call to `endWriteTran()` after the call to
`onFinishExecutingReadWriteRequests()`, allowing users to persist
metadata safely and consistently.


yuliasherman (approved):
LGTM

PR #1604, opened 2021-06-18
Move complaints for higher view to views manager

StefanPalashev:
This Pull Request is yet another step towards making the ViewsManager module unit testable. The changes in it are related to the complaints for higher views and their extraction from "ReplicaImp".


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2342,21 +2341,20 @@ void ReplicaImp::onMessage<ViewChangeMsg>(ViewChangeMsg *msg) {
...
-    LOG_INFO(
-        VC_LOG,
-        "Got quorum of Replicas complaining for a higher View in VCMsg: " << KVLOG(msg->newView(), getCurrentView()));
+
+  if (viewsManager->shouldJumpToHigherViewBasedOnQuorumOfComplaints(msg)) {
     MoveToHigherView(msg->newView());
   }
 
+  // The container has to be empty for the next View change message.
+  viewsManager->getComplaintsForHigherView().clear();
```
teoparvanov, line 2356: can we have a separate `cleanComplaintsForHigherView` method, to avoid calling `.clear()` on a private member?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3109,6 +3092,16 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
   delete msg;
 }
 
+void ReplicaImp::askToLeaveView(ReplicaAsksToLeaveViewMsg::Reason reasonToLeave) {
```
teoparvanov, line 3095: can we move this method to the `ViewsManager`, instead of having it as a private method in `ReplicaImp` ?

bftengine/src/bftengine/ViewsManager.hpp
```diff
@@ -134,6 +136,19 @@ class ViewsManager {
 
   bool hasViewChangeMessageForFutureView(uint16_t repId);
 
+  ReplicasAskedToLeaveViewInfo &getComplainedReplicas() { return complainedReplicas; }
```
teoparvanov, line 139: Instead of exposing the `complainedReplicas` object, can we instead expose several methods in `ViewsManager` with the supported operations:
* `getComplaintFromReplica()`
* `addComplaint()`
* `hasQuorumToLeaveView()`
* ...

@HristoStaykov what are your thoughts on such an approach?

teoparvanov (commented):
LGTM overall. I have some comments from my first read-through... More may follow later.

bftengine/src/bftengine/ReplicaImp.cpp
StefanPalashev: This method was created in order to remove some of the code duplication in `ReplicaImp`. It wraps a call to `tryToGotoNextView()`, which eventually boils down to a call to `MoveToHigherView()` (another member function of `ReplicaImp`). Unfortunately the call to 'MoveToHigherView()' is a major drawback, which prevents a lot of the code from being moved to `ViewsManager`.

bftengine/src/bftengine/ViewsManager.hpp
HristoStaykov: I agree. The only place where we need to iterate over all complaints is during Status Exchange to see what the Peer requesting Status Exchange is missing, compared to what we have and help accordingly. [ReplicaImp.cpp:2220](https://github.com/vmware/concord-bft/blob/c197dc3b8703e0b859d847e0dbb24248eafb4c55/bftengine/src/bftengine/ReplicaImp.cpp#L2220). The filling of this information to Status Request messages I see is correctly moved to ViewsManager.

HristoStaykov (approved):
LGTM, thanks @StefanPalashev !

teoparvanov (approved):
This is great, thank you @StefanPalashev !

PR #1606, opened 2021-06-18
Introduce UnwedgeStatusRequest and UnwedgeCommand in reconfiguration handler

nkgeorgiev:
This is the first part of the unwedge command implementation.

First is the unwedgeStatus command in the reconfiguration commands handler.
If the replica can unwedge, we return a signed answer by the replica's private key.

The unwedge command receives the signatures collected with the unwedgeStatus command.
Each replica verifies the signatures and if a quorum of them are valid, it unwedges.

### DIFF
--- a/bftengine/src/bftengine/ControlStateManager.cpp
+++ b/bftengine/src/bftengine/ControlStateManager.cpp
@@ -12,7 +12,9 @@
 // file.
 
 #include "ControlStateManager.hpp"
+#include "SigManager.hpp"
 #include "Logger.hpp"
+#include "ReplicaConfig.hpp"
 namespace bftEngine {
 
 /*
@@ -96,6 +98,47 @@ void ControlStateManager::onRestartProof() {
   for (const auto& kv : onRestartProofCbRegistery_) {
     kv.second.invokeAll();
   }
+}
+
+std::optional<std::string> ControlStateManager::canUnwedge() {
+  std::optional<std::string> result;
+  if (!getPruningProcessStatus()) {
+    std::string replica_id = std::to_string(ReplicaConfig::instance().getreplicaId());
+    auto sigManager = impl::SigManager::instance();
+    std::string sig(sigManager->getMySigLength(), '\0');
+    sigManager->sign(replica_id.c_str(), replica_id.size(), sig.data(), sig.size());
+    result = sig;
+  }
+  return result;
+}
+
+bool ControlStateManager::verifyUnwedgeSignatures(
+    std::vector<std::pair<uint64_t, std::vector<uint8_t>>> const& signatures) {
+  auto sigManager = impl::SigManager::instance();
+
+  size_t quorum = ReplicaConfig::instance().numReplicas - ReplicaConfig::instance().fVal;
+  if (signatures.size() < quorum) {
+    LOG_INFO(GL, "Not enough signatures for verification");
+    return false;
+  }
+  size_t verified_sigs = 0;
+  for (auto const& sig : signatures) {
+    std::string data = std::to_string(sig.first);
+    std::string signature(sig.second.begin(), sig.second.end());
+    bool valid = sigManager->verifySig(sig.first, data.c_str(), data.size(), signature.data(), signature.size());
+    if (!valid) {
+      LOG_INFO(GL, "Invalid signature for principal id " << sig.first);
+    } else {
+      verified_sigs++;
+    }
+  }
+
+  if (verified_sigs < quorum) {
+    LOG_INFO(GL, "Not enough valid signatures for unwedge");
+    return false;
+  }
+  LOG_INFO(GL, "Successfully validated unwedge signatures");
+  return true;
 }
 
 }  // namespace bftEngine
--- a/bftengine/src/bftengine/ControlStateManager.cpp
+++ b/bftengine/src/bftengine/ControlStateManager.cpp
@@ -102,17 +102,19 @@ void ControlStateManager::onRestartProof() {
 }
 
 std::pair<bool, std::string> ControlStateManager::canUnwedge() {
-  if (!enabled_) return {false, "ControlStateManager is disabled"};
-  std::optional<std::string> result;
+  if (!enabled_) {
+    return {false, "ControlStateManager is disabled"};
+  }
+
   if (!bftEngine::IControlHandler::instance()->isOnNOutOfNCheckpoint()) {
     return {false, "Replica has not reached the wedge point yet"};
   }
+
   auto last_checkpoint = getCheckpointToStopAt();
   ConcordAssert(last_checkpoint.has_value());
 
   std::string sig_data =
       std::to_string(ReplicaConfig::instance().getreplicaId()) + std::to_string(last_checkpoint.value());
-
   auto sig_manager = impl::SigManager::instance();
   std::string sig(sig_manager->getMySigLength(), '\0');
   sig_manager->sign(sig_data.c_str(), sig_data.size(), sig.data(), sig.size());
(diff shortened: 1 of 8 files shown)


reconfiguration/cmf/concord.cmf
```diff
@@ -106,6 +106,21 @@ Msg AddRemoveWithWedgeStatusResponse 31 {
     string config_descriptor
 }
 
+Msg UnwedgeStatusRequest 32{
+    uint64 sender
+}
+
+Msg UnwedgeStatusResponse 33{
```
yontyon, line 113: I would also add here a "reason" field that describes why the replica is unable to wedge

reconfiguration/src/reconfiguration_handler.cpp
```diff
@@ -100,4 +100,35 @@ bool BftReconfigurationHandler::verifySignature(const std::string& data, const s
   if (pub_key_ == nullptr && verifier_ == nullptr) return false;
   return pub_key_->verify(data, signature) || verifier_->verify(data, signature);
 }
+
+bool ReconfigurationHandler::handle(const UnwedgeCommand& cmd,
```
yontyon, line 104: This is the tricky part.
The command handler is being called **after** consensus. However, on wedge, the consensus itself is blocked.
This is why we need to verify this request when getting it, prior to consensus.
On the primary replica this will on **onMessage<ClientRequestMsg>** (see https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L223).
On the follower replica, this would be when they check the conditions to perPrepare (see https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L668).
We need to define a special case for these two spots in which, if we get an unwedge command and it is valid, it can pass consensus.
Once this is done, the handler for unwedge is simply to clear the wedge checkpoint (no need to verify the signatures again)

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -220,7 +220,47 @@ async def test_wedge_command(self, bft_network):
         await self.verify_replicas_are_in_wedged_checkpoint(bft_network, checkpoint_before, range(bft_network.config.n))
         await self.verify_last_executed_seq_num(bft_network, checkpoint_before)
         await self.validate_stop_on_super_stable_checkpoint(bft_network, skvbc)
-    
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_unwedge_command(self, bft_network):
```
yontyon, line 226: The reason why this test is working (despite my comment from above) is because the operator sends a read command (which is not blocked on wedge).
While this is a valid solution, we want the unwedge command to be a write command due to 2 reasons:
1. We want to be able to record it on the blockchain (see https://github.com/vmware/concord-bft/blob/master/kvbc/include/reconfiguration_add_block_handler.hpp)
2. We want all replicas to get it together, and unlike write requests, read requests do not have this guarantee.

tests/apollo/util/operator.py
```diff
@@ -149,6 +171,24 @@ async def wedge_status(self, quorum=None, fullWedge=True):
...
+    async def unwedge(self):
+        await self.can_unwedge()
+        signatures = []
+        for r in self.client.get_rsi_replies().values():
+            res = cmf_msgs.ReconfigurationResponse.deserialize(r)[0]
+            signatures.append(
+                (res.response.replica_id, bytes(res.response.signature)))
+
+        msg = self._construct_reconfiguration_wedge_command(signatures)
+        return await self.client.read(msg.serialize(), reconfiguration=True)
```
yontyon, line 183: See my comment above

yontyon (changes requested):
Thanks @nkgeorgiev for this great PR.
I have only few comments

reconfiguration/src/reconfiguration_handler.cpp
yontyon: After discussing with @ittaiab, we decided to leave it as it is under the assumption that we will have n/n replies for the unwedge status.
Once the work on epochs will be done (see https://github.com/vmware/concord-bft/pull/1608) the unwedge read request will cause the replica to start a new epoch. Eventually, wedge will be tied to an epoch.
Meaning, when issuing a wedge, it means that wedge all write messages until we start a new epoch

bftengine/src/bftengine/ControlStateManager.cpp
```diff
@@ -87,4 +89,45 @@ void ControlStateManager::clearCheckpointToStopAt() {
...
+    result = sig;
+  }
+  return result;
+}
+
+bool ControlStateManager::verifyUnwedgeSignatures(
+    std::vector<std::pair<uint64_t, std::vector<uint8_t>>> const& signatures) {
+  auto sigManager = impl::SigManager::instance();
+
+  size_t quorum = ReplicaConfig::instance().numReplicas - ReplicaConfig::instance().fVal;
```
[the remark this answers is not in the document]
yontyon: For now lets assume that we have n/n quorum for unwedge

bftengine/src/bftengine/ControlStateManager.cpp
```diff
@@ -87,4 +90,61 @@ void ControlStateManager::clearCheckpointToStopAt() {
...
+std::pair<bool, std::string> ControlStateManager::canUnwedge() {
+  std::optional<std::string> result;
+  if (!bftEngine::IControlHandler::instance()->isOnNOutOfNCheckpoint()) {
+    return {false, "Replica has not reached the wedge point yet"};
+  }
+
+  std::string replica_id = std::to_string(ReplicaConfig::instance().getreplicaId());
+  auto sigManager = impl::SigManager::instance();
+  std::string sig(sigManager->getMySigLength(), '\0');
+  sigManager->sign(replica_id.c_str(), replica_id.size(), sig.data(), sig.size());
```
glevkovich, line 102: This is not a good practice in my opinion: The signature will always look the same. I am not sure how many times this signature will occur, but there should be a "salt". I would recommend adding a counter (wedge/unwedge cycle) so that eavesdrops attacker see a different signature every time.
Sign: replica id + cycle number. I think it would be more secure.

bftengine/src/bftengine/ControlStateManager.cpp
```diff
@@ -87,4 +90,61 @@ void ControlStateManager::clearCheckpointToStopAt() {
...
+  size_t verified_sigs = 0;
+  for (auto const& sig : signatures) {
+    // SigManager cannot verify signature coming from same replica
+    // Generate the signature again and compare them.
+    std::string data = std::to_string(sig.first);
+    std::string signature(sig.second.begin(), sig.second.end());
+    bool valid = false;
+    if (sig.first == ReplicaConfig::instance().replicaId) {
+      verified_sigs++;
+      std::string sign(sigManager->getMySigLength(), '\0');
```
glevkovich, line 125: Why do you need to sign every time again, instead of keeping the signature and generate it once on ctor?
Signing is expansive.

teoparvanov:
@nkgeorgiev and @glevkovich please agree on a final approach, so that we can move forward with this PR.

bftengine/src/bftengine/ControlStateManager.cpp
glevkovich: Well, I am aware we are using open source and the attacker maybe read the code. But in general - having access to the exact data that was signed can open the get to a sophisticated attacker to get the signature private key.
So it is better to use "salt" before signing.

bftengine/src/bftengine/ControlStateManager.cpp
glevkovich: This comment is valid only if you decided to sign the replica id over and over without "salt"

bftengine/src/bftengine/ControlStateManager.cpp
nkgeorgiev: @glevkovich  What do you think about using the last sequence number as salt. It is guaranteed to be the same in all replicas since they are wedged.

yontyon (approved):
LGTM!
Thanks!

bftengine/src/bftengine/ControlStateManager.cpp
glevkovich: good idea!

PR #1608, opened 2021-06-20
Introducing Epoch manager

yontyon:
**An epoch** is a life cycle of the system between two events that define the system differently. For example, adding or removing nodes, updating software, and other reconfiguration events.
We need the ability to differentiate between commands of different epochs. 
For example, if a replica gets (via state transfer) a reconfiguration command that is not relevant to its current epoch, it should ignore it.
In this PR we introduce concord-bft's epoch manager.
it preserves several principles:
1. To update its epoch, a replica issues a special reconfiguration request to the consensus mechanism, the execution result is that all on-time replicas update the epoch number of the replica on the chain
2. The epochs are also written to the reserved pages - this is to let read-only replica get updated
3. Because we save the epochs on-chain, we don't remove them when we remove the metadata
4. When we write a command that changes the epoch, we also write its epoch number

The power of knowing the epochs is demonstrated in the state transfer call back for reconfiguration https://github.com/vmware/concord-bft/blob/ee708c78f0b7256d66573e4c2ae00b8fdce80871/kvbc/src/st_reconfiguration_sm.cpp#L85

On a replica state transfer, if we found an addRemove command in the blockchain, we compute the next 3 numbers:
* The replica epoch (read from the blockchain)
* The command epoch (read from the blockchain)
* The max known by quorum epoch (read from reserved pages)

Based on these 3 numbers we cover all cases:
1. If the replica epoch is greater than the command epoch, it means that we already executed this command and we don't need to do anything
2. If the replica epoch is the same as the command epoch, but the checkpoint is smaller than the wedge point, it means that we are not in the wedge point yet. Thus, we return without doing anything and wait for the next checkpoint (notice that this case can only happen if we start state transfer for the last checkpoint before the wedge)
3. If the replica checkpoint is (a) the same as the command epoch and (b) the same as the max known epoch and (c) the checkpoint number is the same as the wedge point, it means that we are in the same state as all other replicas (as far as we know). Thus, we need to run the relevant command handler, wedge and remove the metadata.
Notice, that in this case, if we started state transfer for a checkpoint that does not reflect the reality (for example,  all other replicas are in an advanced epoch, but we started the state transfer for this checkpoint), then eventually another state transfer will happen and we will fall under the next case
4. If the replica epoch is less than the command epoch or less than the max known epoch, it means that we are behind and we want to jump to the max known epoch as fast as we can. 

**Note** that as for (4.) the jump is done to the highest recorded global epoch (tied to the latest addRemove command) + 1. In case that the other replicas have advanced to a newer epoch, it means that eventually, we will start a state transfer and achieve this epoch as well. 

**Note**  that even when we will have other epoch changers (such as unwedge), we will also have every state transfer reconfiguration handler doing its own relevant part, such that once we restart and jump to the new epoch, the replica will be aligned with the state and epochs

**Note** that to complete this feature, we need to add an epoch number to the messages. However, this will be part of a future PR
In addition, we updated the reconfiguration tests to test also that we change epochs

### DIFF
--- a/bftengine/CMakeLists.txt
+++ b/bftengine/CMakeLists.txt
@@ -70,6 +70,7 @@ set(corebft_source_files
     ../ccron/src/ticks_generator.cpp
     ../ccron/src/periodic_action.cpp
     ../ccron/src/cron_table.cpp
+    src/bftengine/EpochsManager.cpp
 )
 
 #
--- a/bftengine/include/bftengine/ControlStateManager.hpp
+++ b/bftengine/include/bftengine/ControlStateManager.hpp
@@ -50,7 +50,7 @@ class ControlStateManager : public ResPagesClient<ControlStateManager, ControlHa
     return instance_;
   }
   ~ControlStateManager() = default;
-  void setStopAtNextCheckpoint(int64_t currentSeqNum);
+  void setStopAtNextCheckpoint(int64_t currentSeqNum, bool saveToReservedPages = true);
   std::optional<int64_t> getCheckpointToStopAt();
 
   void setEraseMetadataFlag(int64_t currentSeqNum);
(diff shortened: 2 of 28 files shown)


nkumar04 (approved):
Great PR!
LGTM!!

PR #1609, opened 2021-06-21
Fix shared signature verification failure processing.

toly-kournik:
Threshsign library uses signers index starting at 1 rathrer than at 0 as replica ids.
This fact wasn't taken into consideration when mapped failed signer id into replica id.
In turn this could cause the system to lose the liveness (n=4, f=1) since the combination
of 3 good sig shares was never verified.
Propper mapping is decrementing 1 from signer index to get replica id.

Appropriate test test_key_exchange_with_file_backup was added to verify the right behavior.
In the test one replica was restarted with an old private keys, which caused it to send improper signatures.


yontyon (approved):
LGTM!

PR #1610, opened 2021-06-21
ConcordClient: Header-only, Namespace, Ownership

robem:
This PR introduces ConcordClient to the ClientService. Thereby, we correct the library to become an interface library and adjust the namespace so that it start with `concord` instead of `client`. See commits for details.


client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -9,6 +9,8 @@
 // terms and conditions of the sub-component's license, as noted in the LICENSE
 // file.
 
+#pragma once
```
upshaw-alex, line 12: I was of the impression the Google C++ style guide called for define guards (https://google.github.io/styleguide/cppguide.html#The__define_Guard) over `#pragma once` (https://google.github.io/styleguide/cppguide.html#Windows_Code). Do we officially diverge from the style guide on this particular matter?

andrewjstone: Pretty much everyone in concord-bft just uses `#pragma once` for simplicity. This works due to the recent, and limited version of compilers we support.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -18,7 +20,7 @@
 #include "bftclient/base_types.h"
 #include "bftclient/bft_client.h"
 
-namespace client::concordclient {
+namespace concord::client::concordclient {
```
andrewjstone, line 23: The repetition here makes me itchy... Same for the path. I'm not sure exactly what to do about it though.

andrewjstone (approved):
Looks good @robem. My only comment would be to see if we can do something about the namespace/name/directory repetition over time.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: I feel similar but the repetition comes from the directory structure. I couldn't think of a better name.
However, I feel strongly that `client` shouldn't be the root namespace but the project name instead.
`concord::ConcordClient` would be better but I'm trying to be consistent with the directory structure and include paths.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: Good point @upshaw-alex, I went with the convenient way as most do and as @andrewjstone pointed out. "Rules are rules" but in this case I think we should allow `#pragma once` specifically in our `CONTRIBUTING.md` given that I cannot think of or haven't found any issues with it that would be solved by include guards. I will prepare a change.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: I feel similar. I was guided by two principals. First, the root namespace should be the project name or something more meaningful than `client`. Second, the namespace hierarchy should follow the directory structure, similar to include paths. I don't have a better suggestion but I'm feeling the itch too :/

PR #1612, opened 2021-06-21
Skip test_skvbc_pyclient.test_primary_write

nkgeorgiev:
The python bft_client has a retry timeout of 250 miliseconds.
If the client request has not finished in 250 miliseconds, the client retries the request by sending to all replicas.
In the case of this test, we sometimes hit this timeout, making the client send more requests than expected in the test.

Skip this test until a proper fix is implemented so it does not block other PRs


teoparvanov (approved):
LGTM, thanks.

PR #1613, opened 2021-06-21
Code style: Allow #pragma once explicitly

robem:
We already use `#pragma once` all over the code base. This change explicitly
calls it out due to the difference to the Google C++ Style Guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,6 +84,7 @@ Our code style has the following exceptions that differ from the google C++ styl
 
  * We use `camelCase` function and method names, rather than `PascalCase`.
  * We allow, and encourage the use of exceptions.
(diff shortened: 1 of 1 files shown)


upshaw-alex:
For the record and for consideration of other reviewers, I see Wikipedia has an article on [`#pragma once`](https://en.wikipedia.org/wiki/Pragma_once) which includes discussions of its pros and cons. FWIW, Wikipedia also describes `#pragma once` as "non-standard but widely supported", which I guess I will admit is acceptable if no one else has objections to it as I am sure we already rely on a lot of other non-standard but widely supported things in C++.

andrewjstone (approved):
Thanks Rob!

f-squirrel (approved):
@robem , I'd even say that `#pragma once` is preferable :)

PR #1614, opened 2021-06-21
Define CMF messages for SKVBC

upshaw-alex:
This PR defines CMF messages intended for use with SKVBC (i.e. simpleKVBC, implemented in `tests/simpleKVBC`) for communication between the SKVBC clients and replicas.

The intention is to eventually move SKVBC clients and replicas (including those implemented in `tests/simpleKVBC/TesterClient` and `tests/simpleKVBC/TesterReplica` and also Apollo's usage of SKVBC implemented in `tests/apollo/util/skvbc.py`) to use these messages defined in CMF rather than their current ad-hoc message format rooted in manually packing C++ struct and object data. We hope the switch to CMF in this area will improve the code cleanliness, readability, and long-term maintainability.

If this commit is accepted, it will be followed up at a later time with changes to actually switch the SKVBC clients and replicas (as outlined above) to use the messages newly defined in this commit.

Note this PR also includes changes to add the step of compiling the CMF file in C++ and python for SKVBC and Apollo, which, other than being eventually needed when we hopefully switch SKVBC to use these CMF messages, will cause `tests/simpleKVBC/cmf/skvbc_messages.cmf` to be compiled to C++ and Python as part of the main Concord-BFT build, which should ensure that `make build` working will imply that `tests/simpleKVBC/cmf/skvbc_messages.cmf` is proper and buildable CMF message schemata.


tests/simpleKVBC/cmf/skvbc_messages.cmf
```diff
@@ -0,0 +1,86 @@
+# Definition of client-server interface for the SimpleKVBC test application.
+#
+# Note SimpleKVBC effectively implements a versioned key-value store supporting
+# only keys and values that are byte strings of exactly 21 bytes.
+#
+# Note that, when key-value pairs (of 21-byte byte strings) need to be handled
```
andrewjstone, line 6: I understand the goal here is to maintain backwards compatibility, but I'm not sure it's useful in this case. I believe the 21 byte form only existed because it was easier to pick a static allocation size for whoever wrote the code initially.  I think it would be beneficial to the system as a whole if instead the key value pairs could be arbitrary sizes (with some max limits). I don't think there's many specific cases where we rely on 21 byte keys and values for any reason other than "sizing". Since we get that sizing for free with CMF, we can probably do away with the limitation altogether.

tests/simpleKVBC/cmf/skvbc_messages.cmf
```diff
@@ -0,0 +1,86 @@
...
+    # workloads in testing.
+    bool long_exec
+
+    list fixedlist uint8 21 readset
+    list fixedlist fixedlist uint8 21 2 writeset
+}
+
+Msg SKVBCGetLastBlockRequest 5 {
+
+    # Though this message does not need to convey any information beyond that
```
andrewjstone, line 39: This is definitely a limitation of CMF right now, although it's likely easily solved. I believe you may only need to change a `+` to a `*` in the [grammar](https://github.com/vmware/concord-bft/blob/570bd6fad4e54f89ce4910107b3e689be4adf869/messages/compiler/grammar.ebnf#L8). It may also need some small fixes for code generation, and at a minimum tests for both C++ and Python.

andrewjstone (commented):
This looks great @upshaw-alex! Thanks so much for doing this. Other than the suggested 21 byte change, I see no other issues.

upshaw-alex:
I have force-pushed a rebase of my branch for this PR off of `master` in the interest of keeping this PR up-to-date.

upshaw-alex:
I have force-pushed a rebase of my branch for this PR off of `master` in the interest of keeping this PR up-to-date.

upshaw-alex:
I have force-pushed an amendment to my commit for this PR making revisions in light of review comments.

tests/simpleKVBC/cmf/skvbc_messages.cmf
upshaw-alex: I have decided to adopt this suggestion in my most recent revisions to this PR; the bytestrings in the CMF messages this PR adds for SKVBC should now be arbitrary-length and represented with the CMF primitive `bytes` type. I have considered this a bit, and I think this may increase the scope of my upcoming changes to actually switch SKVBC and Apollo to actually use the CMF messages I am defining in this PR, as more of the existing SKVBC implementation will need to be refactored to accommodate arbitrary-length byte strings, but I guess now is a good a time as any to make this revision and I think the SKVBC code will be cleaner and more modern for it.

tests/simpleKVBC/cmf/skvbc_messages.cmf
upshaw-alex: I have attempted to remove this limitation as you suggested in my most recent revisions to this PR; I have changed the `+` to an `*`, removed this unused field from the `SKVBCGetLastBlockRequest`. I have also attempted to revise the CMF code generation to fix it so that it runs, even with empty messages, so that the generated 
C++ compiles, and so that a generated Python file can be run without being rejected for syntax errors or other such malformations. I have also validated that `test_cppgen.py` still appears to work. Are there any other revisions I ought to add to CMF to accommodate and test this support for empty messages?

upshaw-alex:
I have noticed the `clang-tidy` check is failing on my branch for this PR. I have not diagnosed why that is, but when I began the process of trying to diagnose this failure, I noticed there are some lines specifically referencing the directories containing existing CMF files in the target for `clang-tidy` in the `Makefile`. I have force-pushed an amendment to my commit for this PR mirroring those lines for `tests/simpleKVBC/cmf` to speculatively see if this fixes the `clang-tidy` issue.

upshaw-alex:
I have force-pushed a rebase of my branch for this PR off of `master` in the interest of keeping this PR up to date.

upshaw-alex:
I have force-pushed an ammendment to my commit for this PR to revise line `178` that I am adding to `Makefile` so that it is indented with tab characters rather than spaces for consistency with the surrounding lines.

tests/simpleKVBC/cmf/skvbc_messages.cmf
andrewjstone: That should be good enough! Thanks Alex.

andrewjstone (approved):
@upshaw-alex This is fantastic. Thanks for the enhancement to CMF!

PR #1615, opened 2021-06-22
Implement replica state sync based on block ID

dartdart26:
Persist the last reachable KVBC block ID in metadata for replica state
sync. We need that, because we write the KVBC block and metadata in two
separate storage transactions - first the KVBC block and then the
metadata. On startup, we compare the last reachable KVBC block ID to the
block ID stored in metadata and delete blocks until they match.

Previous implementation uses the BFT sequence number in the other way
around - persisting it into KVBC blocks. The drawback is that we need to
add KVBC blocks on BFT commands, incurring unnecessary overhead. The
solution in this commit avoids that as we persist the last reachable
KVBC block ID in metadata instead.

Bridge the two approaches by taking software upgrade into consideration,
i.e. when this commit goes live on an existing database storage. First,
make sure the system is in sync based on the BFT sequence number. From
that point on, rely on the new block ID based approach. See comments in
replica_state_sync_imp.cpp and replica_state_sync_test.cpp for more
details.

Note:
We **assume** that the first block after software upgrade is
successfully added to both KVBC and metadata by the replica. If that is
not the case, we don't have a way of knowing how many blocks to delete
and, therefore, we don't delete any as the last block ID in metadata
would just be persisted with the last reachable block ID from KVBC,
making them equal.

### DIFF
--- a/bftengine/src/bftengine/DebugPersistentStorage.cpp
+++ b/bftengine/src/bftengine/DebugPersistentStorage.cpp
@@ -266,7 +266,7 @@ void DebugPersistentStorage::setCheckpointMsgInCheckWindow(SeqNum s, CheckpointM
 
 void DebugPersistentStorage::setUserDataAtomically(const void *data, std::size_t numberOfBytes) {
   const auto p = static_cast<const char *>(data);
-  userData_.assign(p, p + numberOfBytes);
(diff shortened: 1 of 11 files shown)


yuliasherman (approved):
LGTM

tdimitrov (approved):

PR #1620, opened 2021-06-23
TRS: Remove unnecessary for loop in GetClientIdFromClientCert

shgandhi:
The client cert-chain obtained from the property
GRPC_X509_PEM_CERT_PROPERTY_NAME is the only value added to the vector
returned from the `FindPropertyValues` function, hence, only the first
element is required from the vector.

### DIFF
--- a/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
+++ b/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
@@ -293,43 +293,47 @@ class ThinReplicaImpl {
   std::string getClientIdFromClientCert(ServerContextT* context) {
     std::string client_id;
     // get certificate from the client
-    for (auto& temp : context->auth_context()->FindPropertyValues(GRPC_X509_PEM_CERT_PROPERTY_NAME)) {
-      std::string cert_str = temp.data();
-      const char* data = (const char*)cert_str.c_str();
-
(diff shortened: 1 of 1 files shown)


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -293,43 +293,47 @@ class ThinReplicaImpl {
...
-      size_t end = result.find('/', start);
-      client_id = result.substr(start, end - start);
+    // on the client side `GRPC_X509_PEM_CERT_PROPERTY_NAME` returns the full certificate chain
+    auto cert_prop_vec = context->auth_context()->FindPropertyValues(GRPC_X509_PEM_CERT_PROPERTY_NAME);
+    if (cert_prop_vec.empty()) {
+      LOG_FATAL(logger_, "No certificate chain received from the client");
       return client_id;
     }
+    // cert_prop_vec is of type std::vector<grpc::string_ref>, the cert is the first element in the vector
+    std::string cert_str = cert_prop_vec.front().data();
```
robem, line 303: As you mentioned in the description, we assume that the vector has either no or exactly one element. Let's put this assumption in the code and add an assert here or add another if condition.

robem (approved):
Lgtm, thanks for the fix!

PR #1622, opened 2021-06-23
BC-9566: PrePrepareMsg creation and validation performance improvement

arc-vmware:
The creation of preprepare message in Primary replica and validation of the same in secondary replica was taking 21 ms (approx). This PR is to reduce the process. This required around 8 experiments to finalise the fix.
below are some facts based on experiments:
1) The size of prepreparemessage can touch upto 5 MB+. It consist of many client messages, batched together and the order of the client messages is important. So its an n-ordered tuple.
2) PrePrepareMsg involves creation of digest using sha 256 algorithm. Majority of the creation time is consumed in digest calculation.
3) The digest creation of small message is fast. 
The digest is essentially a summarisation process of messages. Based on the above facts (established through experiments) following scatter-gather approach is taken:
1) If signature is available then use that, else calculate the digest of each of the client message in separate threads. 
2) Some finite number of threads are in a thread pool and all the digest calculating tasks are scheduled
3) Wait for the threadpool to finish the calculation of digest of smaller sized client messages
4) Gather all the digests in the original order and calculate the digest of these ordered digests.
This digest of digest finally gives the digest of request. 
This reduced the time from 21 ms to 1.5 to 1.9 ms. This is the upper bound when there are no signatures. If client sends signatures then we will use that directly and the time will reduce further in such case.

### DIFF
--- a/bftengine/src/bftengine/messages/PrePrepareMsg.cpp
+++ b/bftengine/src/bftengine/messages/PrePrepareMsg.cpp
@@ -9,6 +9,7 @@
 // these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
 // file.
 
+#include <utility>
 #include <bftengine/ClientMsgs.hpp>
 #include "OpenTracing.hpp"
 #include "PrePrepareMsg.hpp"
@@ -16,6 +17,7 @@
 #include "Crypto.hpp"
 #include "ClientRequestMsg.hpp"
 #include "SigManager.hpp"
+#include "RequestThreadPool.hpp"
 
 namespace bftEngine {
 namespace impl {
@@ -29,6 +31,46 @@ static Digest nullDigest(0x18);
 ///////////////////////////////////////////////////////////////////////////////
 
 const Digest& PrePrepareMsg::digestOfNullPrePrepareMsg() { return nullDigest; }
+
+void PrePrepareMsg::calculateDigestOfRequests(Digest& digest) const {
+  std::vector<std::pair<char*, size_t>> sigOrDigestOfRequest(b()->numberOfRequests, std::make_pair(nullptr, 0));
+  auto digestBuffer = std::make_unique<char[]>(b()->numberOfRequests * sizeof(Digest));
+
+  std::vector<std::future<void>> tasks;
+  auto it = RequestsIterator(this);
+  char* requestBody = nullptr;
+  size_t local_id = 0;
+
+  while (it.getAndGoToNext(requestBody)) {
+    ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
+    char* sig = req.requestSignature();
+    if (sig != nullptr) {
+      sigOrDigestOfRequest[local_id].first = sig;
+      sigOrDigestOfRequest[local_id].second = req.requestSignatureLength();
+    } else {
+      tasks.push_back(RequestThreadPool::getThreadPool().async(
+          [&sigOrDigestOfRequest, &digestBuffer, local_id](auto* request, auto requestLength) {
+            DigestUtil::compute(request, requestLength, digestBuffer.get() + local_id * sizeof(Digest), sizeof(Digest));
+            sigOrDigestOfRequest[local_id].first = digestBuffer.get() + local_id * sizeof(Digest);
+            sigOrDigestOfRequest[local_id].second = sizeof(Digest);
+          },
+          req.body(),
+          req.size()));
+    }
+    local_id++;
+  }
+  for (const auto& t : tasks) {
+    t.wait();
+  }
+
+  std::string sigOrDig;
+  for (const auto& sod : sigOrDigestOfRequest) {
+    sigOrDig.append(sod.first, sod.second);
+  }
+
+  // compute and set digest
+  DigestUtil::compute(sigOrDig.c_str(), sigOrDig.size(), (char*)&digest, sizeof(Digest));
+}
 
 void PrePrepareMsg::validate(const ReplicasInfo& repInfo) const {
   ConcordAssert(senderId() != repInfo.myId());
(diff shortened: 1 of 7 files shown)


teoparvanov:
Hi @arc-vmware , could you please add a description to the PR, where you describe the goal of this effort, as well as outline the main choices in this approach. Thank you! This will help with the review process.

util/include/thread_pool.hpp
```diff
@@ -109,6 +128,16 @@ class ThreadPool {
 
     // A task queue stop flag.
     bool stop{false};
+
+    // Task count
+    unsigned int tcount{0u};
```
f-squirrel, line 133: `size_t`

util/include/thread_pool.hpp
```diff
@@ -73,11 +74,24 @@ class ThreadPool {
...
+      task_queue_.tcount++;
     }
     task_queue_.cv.notify_one();
     return future;
   }
 
+  // Call this only after putting all jobs or tasks
+  // This is a special function, which wait till all the
+  // tasks are done
+  void finalWaitForAll() {
```
f-squirrel, line 87: Nit:
As far as I understand, this approach is designed for a single producer case:
Thread A may lock on `finalWaitForAll` (`task_queue_.done_check_mutex`) while thread B would continue adding tasks via `auto async(F&& func, Args&&... args)`.

Is it expected behavior?

@dartdart26 , would mind having a look?

f-squirrel (commented):
Great PR with amazing improvements! I have a few suggestions and questions...

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -55,25 +55,45 @@ void PrePrepareMsg::validate(const ReplicasInfo& repInfo) const {
...
 
-  // digest
-  Digest d;
-  const char* requestBuffer = (char*)&(b()->numberOfRequests);
-  const uint32_t requestSize = (b()->endLocationOfLastRequest - prePrepareHeaderPrefix);
-
-  DigestUtil::compute(requestBuffer, requestSize, (char*)&d, sizeof(Digest));
-
-  if (d != b()->digestOfRequests) throw std::runtime_error(__PRETTY_FUNCTION__ + std::string(": digest"));
+  std::vector<std::string> sodofrequests(b()->numberOfRequests);
```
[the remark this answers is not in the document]
arc-vmware: its actually sigOrDigestOfRequest. I will change this name.

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
```diff
@@ -141,6 +144,24 @@ class RequestsIterator {
   uint32_t currLoc;
 };
 
+class RequestThreadPool final {
```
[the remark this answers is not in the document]
arc-vmware: The threadpool that it encapsulates is only needed for PrePrepareMsg. This is similar to RequestIterator so its inside the  same header file.

util/include/thread_pool.hpp
arc-vmware: Here this ability will provide a Thread bag facility. So consider that we create a thread pool of 4 threads. Now consider that we have 50 tasks [lambda] to run. So thread A will put all the tasks in the bag / pool of threads in any order. After that thread A will wait until the 4 threads of the thread pool / bag finishes those 50 tasks. So if we have a finite tasks in hand that we want to process, then we will use this to wait for all threads to complete, if we don't have this kind of scenario then we don't have to wait.

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
```diff
@@ -141,6 +144,24 @@ class RequestsIterator {
...
+ public:
+  static auto& getThreadPool() {
+    static concord::util::ThreadPool threadBag{
```
cloudnoize, line 150: maybe worth defining it as ```static thread_local```

util/include/thread_pool.hpp
cloudnoize: it seems logically correct but if a bug is hidden then we can encounter a deadlock, since we must have a symmetry between the task increment and decrement.
I also suggest @dartdart26 to review this code

cloudnoize (commented):
@arc-vmware brilliant work, before approving I would like @dartdart26 to take a look on the thread pool code

util/include/thread_pool.hpp
```diff
@@ -73,11 +74,24 @@ class ThreadPool {
     {
       auto lock = std::lock_guard{task_queue_.mutex};
       task_queue_.tasks.push(std::move(task));
+      auto done_lock = std::lock_guard{task_queue_.done_check_mutex};
```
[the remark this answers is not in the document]
arc-vmware: Yes I will change this and make it consistent.

util/include/thread_pool.hpp
f-squirrel: @arc-vmware , I am afraid that if there are multiple producer threads, the method `finalWaitForAll` will exit before all the tasks are completed. It seems to me that the method covers a very specific use case and may leave the user with a false impression that it will work for all the cases.

The method `async` returns future so if a user wants to wait until all their tasks are done, they need to save the futures and wait for them, i.e.:

```
std::vector<future> results;
for (...) {
   results.push_back(pool.async( my_func );
 }
 
 for(auto& r: results) {
    f.wait();
}
```

WDYT?

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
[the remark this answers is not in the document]
f-squirrel: Well, It does not justify the definition in the message's class.

f-squirrel (changes requested):

util/include/thread_pool.hpp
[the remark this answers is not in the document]
arc-vmware: I reviewed the threadpool again, its actually consistent. Also auto lock = locktype{mutex} is a syntactic sugar for locktype lock(mutex) since both are initialisation. I will discuss and then change this one.

bftengine/tests/messages/PrePrepareMsg_test.cpp
```diff
@@ -42,6 +46,47 @@ ClientRequestMsg create_client_request() {
...
+}
+
+void create_random_client_requests(std::vector<std::shared_ptr<ClientRequestMsg>>& crmv, size_t numMsgs) {
+  crmv.clear();
+  std::mt19937_64 eng{std::random_device{}()};
+  std::uniform_int_distribution<> dist{12, 10000};
+  std::uniform_int_distribution<> msdist{0, 1000};
+
+  while (numMsgs > 0) {
+    uint64_t reqSeqNum = 100u + numMsgs;
```
andrewjstone, line 72: This `100u` is an implicit limit not exposed to the caller. If a caller passes numMsgs > 100, the size of the batch will be silently truncated. Why not instead remove the limit and count upwards?

bftengine/tests/messages/PrePrepareMsg_test.cpp
```diff
@@ -58,6 +103,53 @@ class PrePrepareMsgTestFixture : public ::testing::Test {
...
+  EXPECT_NO_THROW(msg.finishAddingRequests());  // create the digest
+  EXPECT_EQ(msg.numberOfRequests(), 50u);
+
+  std::string dod;
+  for (const auto& s : dv) {
+    dod.append(s);
+  }
+  Digest d;
+  DigestUtil::compute(dod.c_str(), dod.size(), (char*)&d, sizeof(Digest));
+  EXPECT_EQ(d == msg.digestOfRequests(), true);
```
andrewjstone, line 149: This test should be `EXPECT_EQ(d, msg.digestOfRequests())`. Otherwise, you will just get an error of "false == true" rather than a precise printing of what went wrong.

andrewjstone (commented):
Hi @arc-vmware. This looks very good. This should give us a substantial performance improvement!

We still want to offload all the hashing off the main thread and not block at all in the manner we discussed. Even blocking the main event loop for 2ms is expensive. But as we've talked about, we have other more expensive spots now that this was shrunk so much. Nice work!

f-squirrel (approved):
Looks good to me, thank you @arc-vmware for addressing the comments!

andrewjstone:
Hey @arc-vmware this PR now seems to include several other commits unrelated to this change. Can you please rebase this branch to include only your relevant commits and repush? Thanks!

arc-vmware:
@andrewjstone Apologies for the other picks. I have rebased with the right commits.

andrewjstone (approved):
Looks great @arc-vmware! Ship it!

PR #1628, opened 2021-06-24
ClientService: Integrate ConcordClient into EventService

robem:
This change connect the gRPC service with the concord client subscribe() backend. For the purpose of basic manual testing, this change returns empty data but the actual integration with TRC would look similar. You can test out the backend by starting clientservice and issue the following grpcli request:
`./grpc_cli call localhost:1337 StreamEventGroups "event_group_id: 1337"`

See commit messages for details.

### DIFF
--- a/client/clientservice/src/event_service.cpp
+++ b/client/clientservice/src/event_service.cpp
@@ -9,9 +9,11 @@
 // these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
 // file.
 
+#include <chrono>
 #include <iostream>
 #include <google/protobuf/util/time_util.h>
 #include <opentracing/tracer.h>
+#include <thread>
 
 #include "event_service.hpp"
 #include "client/concordclient/concord_client.hpp"
@@ -25,6 +27,8 @@ using google::protobuf::util::TimeUtil;
 using vmware::concord::client::v1::StreamEventGroupsRequest;
 using vmware::concord::client::v1::EventGroup;
 
+using namespace std::chrono_literals;
+
 namespace cc = concord::client::concordclient;
 
 namespace concord::client::clientservice {
@@ -32,34 +36,33 @@ namespace concord::client::clientservice {
 Status EventServiceImpl::StreamEventGroups(ServerContext* context,
                                            const StreamEventGroupsRequest* proto_request,
                                            ServerWriter<EventGroup>* stream) {
-  LOG_INFO(logger_, "EventServiceImpl::StreamEventGroups called");
-
   cc::SubscribeRequest request;
   request.event_group_id = proto_request->event_group_id();
 
-  auto callback = [this, stream](cc::SubscribeResult event) {
-    if (not std::holds_alternative<cc::EventGroup>(event)) {
+  auto callback = [this, stream](cc::SubscribeResult&& subscribe_result) {
+    if (not std::holds_alternative<cc::EventGroup>(subscribe_result)) {
       LOG_INFO(logger_, "Subscribe returned error");
       return;
     }
-    auto eg = std::get<cc::EventGroup>(event);
-    EventGroup event_group;
-    event_group.set_id(eg.id);
-    for (cc::Event e : eg.events) {
-      *event_group.add_events() = std::string(e.begin(), e.end());
+    auto event_group = std::get<cc::EventGroup>(subscribe_result);
+    EventGroup proto_event_group;
+    proto_event_group.set_id(event_group.id);
+    for (cc::Event e : event_group.events) {
+      *proto_event_group.add_events() = std::string(e.begin(), e.end());
     }
-    *event_group.mutable_record_time() = TimeUtil::MicrosecondsToTimestamp(eg.record_time.count());
-    *event_group.mutable_trace_context() = {eg.trace_context.begin(), eg.trace_context.end()};
+    *proto_event_group.mutable_record_time() = TimeUtil::MicrosecondsToTimestamp(event_group.record_time.count());
+    *proto_event_group.mutable_trace_context() = {event_group.trace_context.begin(), event_group.trace_context.end()};
 
-    stream->Write(event_group);
+    stream->Write(proto_event_group);
   };
 
   auto span = opentracing::Tracer::Global()->StartSpan("stream_event_groups", {});
   client_->subscribe(request, span, callback);
 
   // TODO: Consider all gRPC return error codes as described in concord_client.proto
-  while (!context->IsCancelled())
-    ;
+  while (!context->IsCancelled()) {
+    std::this_thread::sleep_for(10ms);
+  }
 
   client_->unsubscribe();
   return grpc::Status::OK;
--- a/client/clientservice/src/event_service.cpp
+++ b/client/clientservice/src/event_service.cpp
@@ -37,7 +37,12 @@ Status EventServiceImpl::StreamEventGroups(ServerContext* context,
   cc::SubscribeRequest request;
   request.event_group_id = proto_request->event_group_id();
 
-  auto callback = [stream](cc::EventGroup eg) {
+  auto callback = [this, stream](cc::SubscribeResult event) {
+    if (not std::holds_alternative<cc::EventGroup>(event)) {
+      LOG_INFO(logger_, "Subscribe returned error");
+      return;
+    }
+    auto eg = std::get<cc::EventGroup>(event);
     EventGroup event_group;
     event_group.set_id(eg.id);
     for (cc::Event e : eg.events) {
--- a/client/clientservice/src/event_service.cpp
+++ b/client/clientservice/src/event_service.cpp
@@ -10,22 +10,51 @@
 // file.
 
 #include <iostream>
+#include <google/protobuf/util/time_util.h>
+#include <opentracing/tracer.h>
 
 #include "event_service.hpp"
+#include "client/concordclient/concord_client.hpp"
 
 using grpc::Status;
 using grpc::ServerContext;
 using grpc::ServerWriter;
 
+using google::protobuf::util::TimeUtil;
+
 using vmware::concord::client::v1::StreamEventGroupsRequest;
 using vmware::concord::client::v1::EventGroup;
+
+namespace cc = concord::client::concordclient;
 
 namespace concord::client::clientservice {
 
 Status EventServiceImpl::StreamEventGroups(ServerContext* context,
-                                           const StreamEventGroupsRequest* request,
+                                           const StreamEventGroupsRequest* proto_request,
                                            ServerWriter<EventGroup>* stream) {
   LOG_INFO(logger_, "EventServiceImpl::StreamEventGroups called");
+
+  cc::SubscribeRequest request;
+  request.event_group_id = proto_request->event_group_id();
+
+  auto callback = [stream](cc::EventGroup eg) {
+    EventGroup event_group;
+    event_group.set_id(eg.id);
+    for (cc::Event e : eg.events) {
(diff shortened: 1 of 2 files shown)


client/clientservice/src/event_service.cpp
```diff
@@ -52,9 +52,11 @@ Status EventServiceImpl::StreamEventGroups(ServerContext* context,
   auto span = opentracing::Tracer::Global()->StartSpan("stream_event_groups", {});
   client_->subscribe(request, span, callback);
 
-  while (!context->IsCancelled()) {
-  }
+  // TODO: Consider all gRPC return error codes as described in concord_client.proto
+  while (!context->IsCancelled())
```
f-squirrel, line 61: If `IsCancelled` does not use some sort of blocking, this code may potentially become a huge busy loop.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -130,14 +132,59 @@ class ConcordClient {
...
+  // TODO: Allow multiple subscriptions
+  std::atomic_bool stop_subscriber_{true};
+  std::thread subscriber_;
 };
 
+template <class CallbackT>
+void ConcordClient::subscribe(const SubscribeRequest& request,
+                              const std::unique_ptr<opentracing::Span>& parent_span,
+                              CallbackT callback) {
+  if (not stop_subscriber_) {
```
f-squirrel, line 163: Let us use the classic C++ not - `!`

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -130,14 +132,59 @@ class ConcordClient {
...
+    LOG_ERROR(logger_, "subscription already in progress - unsubscribe first");
+    return;
+  }
+
+  stop_subscriber_ = false;
+  std::atomic_bool* stop = &stop_subscriber_;
```
f-squirrel, line 170: The pointer is redundant here, a regular reference will be enough because accessing an atomic variable through a reference doesn't introduce any additional accesses or modifications, because references do not occupy memory locations. Therefore, performing potentially concurrent atomic operations is still safe when those operations occur through references.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -130,14 +132,59 @@ class ConcordClient {
...
+      // The thread is in place to simulate the asynchronous subscription.
+      EventGroup eg;
+      eg.id = request.event_group_id;
+      std::string event = std::to_string(request.event_group_id);
+      eg.events.push_back({event.begin(), event.end()});
+      std::chrono::duration time_now = std::chrono::system_clock::now().time_since_epoch();
+      eg.record_time = std::chrono::duration_cast<std::chrono::microseconds>(time_now);
+      eg.trace_context = {};
+
+      callback(eg);
```
f-squirrel: I am not aware of the size of `EventGroup` but I would suggest changing the signature of the `callback` to accept an rvalue reference to `EventGroup`. Especially, because it is never reused in the thread.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -130,14 +132,59 @@ class ConcordClient {
...
+    }
+  });
+}
+
+inline void ConcordClient::unsubscribe() {
```
f-squirrel, line 189: Please remove `inline`:
1. `inline` only recommends to a compiler to make the function inlined, i.e. it is not guaranteed.
2. functions defined in headers are implicitly treated by the C++ compiler as candidates for inlining
2. a "heavy" function has no chance to become inline. This function is super-heavy because it calls `join`(interaction with the kernel).

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -130,14 +132,59 @@ class ConcordClient {
 
   // Register a callback that gets invoked for every validated event received.
   // void callback(EventGroup);
+  // Return subscriber ID used to unsubscribe.
   template <class CallbackT>
   void subscribe(const SubscribeRequest& request,
```
f-squirrel, line 141: The callback seems to have a very specific signature, it accepts EventGroups. I do not see the point of making the method templated. I suggest passing callback as `std::function` with a clear list of arguments.

Additionally, once we get rid of templates, we may move the implementation to a CPP file to decrease the compilation time and improve incapslation.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -130,14 +132,59 @@ class ConcordClient {
...
                  CallbackT callback);
+
+  // Unsubscribe with the subscriber ID retrieved from a prior call to subscribe.
+  // Note, if the caller doesn't unsubscribe and no runtime error occurs then resources
+  // will be occupied forever.
   void unsubscribe();
 
  private:
+  logging::Logger logger_;
   ConcordClientConfig config_;
```
f-squirrel, line 152: `const ConcordClientConfig& config_;` ?

f-squirrel (changes requested):
Thank you @robem for this PR, I have added a few suggestions below.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: `inline` isn't only to tell the compiler about possible optimizations (which we don't need to do anyways) but also to work around the "one-definition-rule". Meaning, if we define a function in a header file (which we have to do because we are using templates) then multiple includes of this file in different compilation units can lead to double-definition when the linker links those objects together. This is exactly what happened here. We will need to make the other functions `inline` as well.
This ties back to your API remark above, if we would use `std::function` instead of templates then we wouldn't face this issue. However, so far, I've valued the benefits of templates higher then the issues I faced.
@andrewjstone might have insightful opinions too.

robem (commented):
Many thanks for review @f-squirrel!

client/clientservice/src/event_service.cpp
robem: This won't be an empty loop. It will be filled with the TRC integration.

client/concordclient/include/client/concordclient/concord_client.hpp
f-squirrel: Well, from my experience, this is a very rarely used feature and most C++ devs are confused when they see it.

client/clientservice/src/event_service.cpp
f-squirrel: I am afraid to leave such code as-is: if you say that it will be filled later then put a sleep in the while for now.

client/concordclient/include/client/concordclient/concord_client.hpp
f-squirrel: Well, `inline` is still redundant: you could have written the method definition inside the class. In this case, the compiler will handle "one-definition-rule" for you implicitly. So now, I suggest putting this method to the class declaration.

client/concordclient/include/client/concordclient/concord_client.hpp
f-squirrel: The OP at SO is concerned with performance, which is not the problem in our case: subscription does not happen too often. Regarding the compile-time guarantees, `std::function` provides them: if you try to pass a function with the wrong signature, the compiler will fail.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: I think I misunderstood your comment to get rid of the extra variable. Nvm. That works. I will change it. Thanks!

robem: I think I misunderstood your comment to remove the extra variable. Nvm. That works. Thanks!

client/concordclient/include/client/concordclient/concord_client.hpp
robem: I don't like that because then you lose the overview of the class declaration. At a first glance, I don't care about the implementation but the API and the documentation. Therefore, I prefer for it to be as concise as possible.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: Good catch - thanks!

client/concordclient/include/client/concordclient/concord_client.hpp
robem: I don't mind changing it but we need a good reason now and for later (we can put it in our CONTRIBUTING md). I'm not sure if confused devs really counts because it won't be constant confusion and it is still proper C++. I couldn't find any guidelines around that.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: Performance is probably no issue for `subscribe` but might be for `send` and the APIs should look the same IMHO. Is your only objection that the callback parameter are defined in the documentation and not in the code or is there more?

client/concordclient/include/client/concordclient/concord_client.hpp
f-squirrel: It bothers me only because from my experience most devs are surprised to see it. If you really like it, keep it, no problem.

client/clientservice/src/event_service.cpp
```diff
@@ -37,7 +37,12 @@ Status EventServiceImpl::StreamEventGroups(ServerContext* context,
   cc::SubscribeRequest request;
   request.event_group_id = proto_request->event_group_id();
 
-  auto callback = [stream](cc::EventGroup eg) {
+  auto callback = [this, stream](cc::SubscribeResult event) {
```
shgandhi: Nit: since `event` here corresponds to event group, naming the `cc::SubscribeResult` type object as `event` is confusing to me. Can we rename it to - `subscribe_result` maybe?

shgandhi (approved):
Thanks for the change @robem, I just have a few minor comments. LGTM!

client/concordclient/include/client/concordclient/concord_client.hpp
robem: We talked offline. TLDR: We believe that the potential performance boost of templates will be negligible at runtime and hence we choose `std::function` for practicality. 

Our conclusion is that we don't see the callback invocation itself as a performance bottleneck for now. For both APIs ,`send` and `subscribe`, concord client is going to wait on multiple replicas to respond and compare the results. Invoking the callback via templates or std::function afterwards seems negligible in comparison. However, we also have no data (we should do this once concord client is fully implemented). We agreed that `std::function` makes the interface straightforward and easier to maintain. Templates require all functions to be defined in header files and callback types to be "defined" in comments. If performance is not an issue then we prefer a clean interface via `std::function`.

client/clientservice/src/event_service.cpp
```diff
@@ -10,22 +10,51 @@
...
 namespace concord::client::clientservice {
 
 Status EventServiceImpl::StreamEventGroups(ServerContext* context,
-                                           const StreamEventGroupsRequest* request,
+                                           const StreamEventGroupsRequest* proto_request,
                                            ServerWriter<EventGroup>* stream) {
   LOG_INFO(logger_, "EventServiceImpl::StreamEventGroups called");
+
+  cc::SubscribeRequest request;
+  request.event_group_id = proto_request->event_group_id();
```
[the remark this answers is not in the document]
robem: Probably not but you need both to keep concord client independent from gRPC.

f-squirrel (approved):
LGTM, thank you @robem for a healthy discussion!

PR #1630, opened 2021-06-25
Remove key prefix from TRC and TRS

shgandhi:
This PR contains commits that remove key_prefix from TRC and TRS. The reason this change is required is because TRS only streams events and every event is already prefixed with “F” (F = fragment), hence the client application doesn't need to filter for fragment kv-pairs in the read-service.


f-squirrel:
What is the motivation behind removing the key prefix?

shgandhi:
> What is the motivation behind removing the key prefix?

@f-squirrel  this is because TRS now only streams events and every event is already prefixed with “F” (F = fragment), Daml doesn't need to filter for fragment kv-pairs in the read-service.

f-squirrel:
Thank you for the explanation, @shgandhi !

f-squirrel (approved):

shgandhi:
Thanks for the review @robem, @f-squirrel.
Marking this as WIP to avoid accidental merge, waiting for the corresponding closed-source change to get approved.

PR #1631, opened 2021-06-27
Transaction signing enable test

glevkovich:
1) SigManager: modify the way we update metrics
2) CMakeLists: enable transaction signing for Apollo tests


cloudnoize (approved):
LGTM!

PR #1634, opened 2021-06-28
Removed select usage diagnostics server master

HristoStaykov:
We handle only 1 request at a time, therefore we don't need to use IO Multiplexing. As it turns out the **select** system call is causing problems, because of the macros it uses to setup the set of sockets to monitor. If they are passed a socket with higher value than 1024, they cause stack corruption. This we found is described in the man page for select:
`POSIX allows an implementation to define an upper limit, advertised via the constant FD_SETSIZE, on the range of file descriptors that can be specified in a file descriptor set. The Linux kernel imposes no fixed limit, but the glibc implementation makes fd_set a fixed-size type, with FD_SETSIZE defined as 1024, and the FD_*() macros operating according to that limit. To monitor file descriptors greater than 1023, use poll(2) instead.`
In this PR we preserve the same handling of 1 request at a time, since this interface is meant for debugging/monitoring purposes and we don't need to support heavy load. We do this by simply setting up timeouts on the sockets.

### DIFF
--- a/diagnostics/include/diagnostics_server.h
+++ b/diagnostics/include/diagnostics_server.h
@@ -144,7 +144,6 @@ class Server {
         int sock = accept(listen_sock_, NULL, NULL);
         if (sock < 0) {
           if (errno == EAGAIN || errno == EINTR) {
-            LOG_WARN(logger, "DEBUG: TIMEOUT: " << errnoString(errno));
             continue;
           } else {
             LOG_WARN(logger, "Failed to accept connection: " << errnoString(errno));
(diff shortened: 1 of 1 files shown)


diagnostics/include/diagnostics_server.h
```diff
@@ -133,6 +104,22 @@ inline void handleRequest(Registrar& registrar, int sock) {
 // async connections via boost ASIO if necessary, although this seems extremely heavy handed for the
 // use case.
 class Server {
+  bool setTimeOut(int fd, time_t seconds = 1) {
```
f-squirrel, line 107: Nit: `chrono::seconds` ?

f-squirrel (approved):
LGTM!

diagnostics/include/diagnostics_server.h
HristoStaykov: In order to set socket timeout and preserve the logic as it was before (handling 1 Client Request at a time) we use **setsockopt**. This is a Linux interface call, which takes a structure **timeval**. In this structure the seconds are represented as **time_t**.

PR #1637, opened 2021-06-29
Introduce a reserved pages client for pruning

dartdart26:
Provide persistence of pruning agreement and progress information in
reserved pages. Pseudo code for how it can be used is provided as an
inline comment.

Introduce the `ReservedPagesMock` class and reuse it in ccron and
pruning reserved pages client unit tests. In order to make it work, add
the static `getNumResPages()` method and make the `my_offset()` and
`calc_my_offset()` methods static instead of const in `ResPagesClient`.
That allows for a simpler init procedure in unit tests as we can now
call them without an instance.

Move the `INITIAL_GENESIS_BLOCK_ID` constant to kv_types.hpp and reuse
it across the codebase.

Add an unit test to verify the pruning reserved pages client's
functionality.


bftengine/include/bftengine/ReservedPagesClient.hpp
```diff
@@ -94,14 +104,14 @@ class ResPagesClient : public ReservedPagesClientBase, public IReservedPages {
   }
 
   /** is called when calculating absolute pageId */
-  uint32_t my_offset() const {
+  static uint32_t my_offset() {
```
toly-kournik, line 107: my_offset() was intentionally declares private in order not to allow clients to intervene the res pages calc math

kvbc/include/pruning_reserved_pages_client.hpp
```diff
@@ -0,0 +1,103 @@
...
+//
+// * On Startup and State Transfer:
+//   auto agreement = client.lastAgreement();
+//   const auto to = client.latestBatchBlockIdTo();
+//   if (to && *to + 1 > genesis) blockchain.deleteUntil(*to + 1);
+//   if (genesis < agreement->last_agreed_prunable_block_id) ticks_gen.start();
+//   else ticks_gen.stop();
+class ReservedPagesClient : private bftEngine::ResPagesClient<ReservedPagesClient, kPruningResPageCount> {
+ public:
+  using bftEngine::ResPagesClient<ReservedPagesClient, kPruningResPageCount>::getNumResPages;
```
toly-kournik, line 62: I didn't find its usage

bftengine/include/bftengine/ReservedPagesClient.hpp
```diff
@@ -81,6 +81,16 @@ class ResPagesClient : public ReservedPagesClientBase, public IReservedPages {
     }
   }
 
+  static uint32_t getNumResPages() {
```
toly-kournik, line 84: a bit confusing name, since the it should return client's num of pages

bftengine/include/bftengine/ReservedPagesClient.hpp
dartdart26: I will make it private and make `ReservedPagesMock` a friend.

kvbc/include/pruning_reserved_pages_client.hpp
dartdart26: It is used by `ReservedPagesMock`.

bftengine/include/bftengine/ReservedPagesClient.hpp
dartdart26: Any suggestion for a better one?

kvbc/include/pruning_reserved_pages_client.hpp
toly-kournik: I mean this particular line in pruning ReservedPagesClient is not in use. Mock is a non-related class

bftengine/include/bftengine/ReservedPagesClient.hpp
toly-kournik: there is already a not very successful `numberOfReservedPages()` name...
maybe getMyNumResPaged() :-/

kvbc/include/pruning_reserved_pages_client.hpp
dartdart26: Ah, yes. What I meant is that the inheritance is `private` as I don't think the pruning client needs any of the ResPagesClient methods. That's why I "lift" this method here so that `ReservedPagesMock` can use it. Does that make sense?

bftengine/include/bftengine/ReservedPagesClient.hpp
dartdart26: I was thinking that `ResPagesClient::numberOfReservedPages()`'s return value is misleading in the sense that it returns the total number of reserved pages, if I am not mistaken (because it will have a pointer to the one initialised in ST). Maybe it should return the "client/my" value instead?

Maybe we can name the static function something like `numberOfReservedPagesForClient()` or go with `getMyNumResPages()`, I don' have a preference.

kvbc/include/pruning_reserved_pages_client.hpp
dartdart26: Or I can just use composition and implement that method. No actual reason for using `private` inheritance here, as far as I remember.

bftengine/include/bftengine/ReservedPagesClient.hpp
toly-kournik: just choose

kvbc/include/pruning_reserved_pages_client.hpp
toly-kournik: yeah, I never found a private inheritance to be really useful

bftengine/include/bftengine/ReservedPagesClient.hpp
dartdart26: Done.

toly-kournik (approved):
LGTM!

bftengine/include/bftengine/ReservedPagesClient.hpp
dartdart26: Since we are on the topic. What should `IRservervedPages:: numberOfReservedPages()` return, in general>?

toly-kournik: the total number

dartdart26: In that case, I guess we can remove it from the interface and use the information from the registry (separately from this PR).

PR #1639, opened 2021-06-30
PreExecution non-repudiation implementation

tdimitrov:
This PR implements PreExecution result verification by non-primary replicas (aka PreExecution non-repudiation).

What it does:
1. Adds flag to the PreProcessor indicating if the feature should be enabled or not (feature flag).
2. Adds new parameter to ClientRequestMessage, which allows allocating extra buffer space by MessageBase class. This buffer can be used by messages inheriting from ClientRequestMessage to store additional fields (see next point).
3. Adds new PreExecution message - PreProcessedResult message. It is carries the result from the PreExecution and the signatures of all f+1 replicas which has preexecuted the concrete request. This message is implemented by inheriting from ClientRequestMessage. The feature described in point 2 is used to allocate additional buffer space for the signatures.
4. Result verification logic handling.

How the result verification logic changes the PreExecution flow?
================================================
The flow remains unchanged up to receiving a PreProcessReply message on the primary. Now instead of adding a ClientRequestMsg to its internal queue the primary uses PreProcessedResult for this purpose. A clear benefit from this is responsibility separation - ClientRequest messages are used for client requests and PreProcessedResult - for pre-executed requests. All hash signatures from the non-primary replicas are added in the PreProcessedResult message. All its other fields are 1:1 compared to ClientRequest.

On finalise preprocessing phase (finalizePreprocessing method) the Primary bundles some ClientRequest and PreProcessedResult messages in a PrePrepare message. There is another benefit from the inheritance here - PrePrepare handling logic doesn't care about the new message. From its point of view it is just another ClientRequest message.

When a non-primary receives a PrePrepare message it unbundles all ClientRequest and PreProcessedResult messages from it. They are discriminated by the message type in the message header. If the non-primary find a PreProcessedResult message it verifies the validity of all signatures in it and additionally if f+1 signatures are assigned. If both requirements are satisfied the message is processed as usual.

However if a non-primary finds a PreProcessedResult message with invalid signature or less than f+1 signatures it assumes that the primary is malicious and triggers a view change. New reason is added to ReplicaAsksToLeaveViewMsg -  PrimarySentBadPreProcessedResult.

### DIFF
--- a/bftengine/include/bftengine/ClientMsgs.hpp
+++ b/bftengine/include/bftengine/ClientMsgs.hpp
@@ -40,6 +40,7 @@ struct ClientRequestMsgHeader {
   uint64_t timeoutMilli;
   uint32_t cidLength = 0;
   uint16_t reqSignatureLength = 0;
+  uint32_t extraDataLength = 0;
 
   // followed by the request (security information, such as signatures, should be part of the request)
 
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -42,6 +42,7 @@
 #include "messages/ReplicaAsksToLeaveViewMsg.hpp"
 #include "messages/ReplicaRestartReadyMsg.hpp"
 #include "messages/ReplicasRestartReadyProofMsg.hpp"
+#include "messages/PreProcessResultMsg.hpp"
 #include "CryptoManager.hpp"
 #include "ControlHandler.hpp"
 #include "bftengine/KeyExchangeManager.hpp"
@@ -99,6 +100,9 @@ void ReplicaImp::registerMsgHandlers() {
 
   msgHandlers_->registerMsgHandler(MsgCode::ClientRequest,
                                    bind(&ReplicaImp::messageHandler<ClientRequestMsg>, this, _1));
+
+  msgHandlers_->registerMsgHandler(MsgCode::PreProcessResult,
+                                   bind(&ReplicaImp::messageHandler<preprocessor::PreProcessResultMsg>, this, _1));
 
   msgHandlers_->registerMsgHandler(MsgCode::ReplicaStatus,
                                    bind(&ReplicaImp::messageHandler<ReplicaStatusMsg>, this, _1));
@@ -300,6 +304,14 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
     }
   }
   delete m;
+}
+
+template <>
+void ReplicaImp::onMessage<preprocessor::PreProcessResultMsg>(preprocessor::PreProcessResultMsg *m) {
+  LOG_DEBUG(GL,
+            "Handling PreProcessResultMsg via ClientRequestMsg handler "
+                << KVLOG(m->clientProxyId(), m->getCid(), m->requestSeqNum()));
+  return onMessage<ClientRequestMsg>(m);
 }
 
 template <>
@@ -673,6 +685,32 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
   }
 }
 
+bool ReplicaImp::validatePreProcessedResults(const PrePrepareMsg *msg,
+                                             const bftEngine::impl::ReplicasInfo &replicasInfo) {
+  RequestsIterator reqIter(msg);
+  char *requestBody = nullptr;
+  while (reqIter.getAndGoToNext(requestBody)) {
+    const MessageBase::Header *hdr = (MessageBase::Header *)requestBody;
+    if (hdr->msgType == MsgCode::PreProcessResult) {
+      preprocessor::PreProcessResultMsg req((ClientRequestMsgHeader *)requestBody);
+
+      if (auto err = req.validatePreProcessResultSignatures(getReplicaConfig().replicaId, getReplicaConfig().fVal);
+          err) {
+        // trigger view change
+        LOG_INFO(VC_LOG,
+                 "Received PrePrepareMsg containing PreProcessResult with invalid signature: "
+                     << *err << ". Ask to leave view " << getCurrentView());
+        askToLeaveView(ReplicaAsksToLeaveViewMsg::Reason::PrimarySentBadPreProcessResult);
+        return false;
+      }
+    } else {
+      // ClientRequest messages are handled in PrePrepare::validate()
+      continue;
+    }
+  }
+  return true;
+}
+
 template <>
 void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
   if (isSeqNumToStopAt(msg->seqNumber())) {
@@ -681,6 +719,12 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
              "etc...)");
     return;
   }
+
+  if (!validatePreProcessedResults(msg, getReplicasInfo())) {
+    // validatePreProcessedResults() logs the error
+    return;
+  }
+
   metric_received_pre_prepares_.Get().Inc();
   const SeqNum msgSeqNum = msg->seqNumber();
 
--- a/bftengine/src/bftengine/ReplicaImp.hpp
+++ b/bftengine/src/bftengine/ReplicaImp.hpp
@@ -42,6 +42,9 @@
 
 #include <ccron/ticks_generator.hpp>
 
+namespace preprocessor {
+class PreProcessResultMsg;
+}
 namespace bftEngine::impl {
 
 class ClientRequestMsg;
@@ -479,6 +482,9 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
   void startConsensusProcess(PrePrepareMsg* pp, bool isInternalNoop);
   void startConsensusProcess(PrePrepareMsg* pp);
   bool isSeqNumToStopAt(SeqNum seq_num);
+
+  bool validatePreProcessedResults(const PrePrepareMsg* msg, const bftEngine::impl::ReplicasInfo& replicasInfo);
+  bool validatePreProcessResultSignatures(preprocessor::PreProcessResultMsg& msg);
 
   // 5 years
   static constexpr int64_t MAX_VALUE_SECONDS = 60 * 60 * 24 * 365 * 5;
--- a/bftengine/src/bftengine/messages/ClientRequestMsg.hpp
+++ b/bftengine/src/bftengine/messages/ClientRequestMsg.hpp
@@ -26,7 +26,7 @@ class ClientRequestMsg : public MessageBase {
   static_assert(sizeof(ClientRequestMsgHeader::msgType) == sizeof(MessageBase::Header::msgType), "");
   static_assert(sizeof(ClientRequestMsgHeader::idOfClientProxy) == sizeof(NodeIdType), "");
   static_assert(sizeof(ClientRequestMsgHeader::reqSeqNum) == sizeof(ReqId), "");
-  static_assert(sizeof(ClientRequestMsgHeader) == 42, "ClientRequestMsgHeader size is 42B");
+  static_assert(sizeof(ClientRequestMsgHeader) == 46, "ClientRequestMsgHeader size is 46B");
   static concord::diagnostics::Recorder sigNatureVerificationRecorder;
   // TODO(GG): more asserts
 
@@ -40,7 +40,8 @@ class ClientRequestMsg : public MessageBase {
                    const std::string& cid = "",
                    const concordUtils::SpanContext& spanContext = concordUtils::SpanContext{},
                    const char* requestSignature = nullptr,
-                   uint32_t requestSignatureLen = 0);
+                   uint32_t requestSignatureLen = 0,
+                   const uint32_t extraBufSize = 0);
 
   ClientRequestMsg(NodeIdType sender);
 
@@ -75,6 +76,11 @@ class ClientRequestMsg : public MessageBase {
 
   void validateImp(const ReplicasInfo& repInfo) const;
 
+  // Returns a pair of pointer and size to the extra buffer which was allocated during initialisation
+  std::pair<char*, uint32_t> getExtraBufPtr() {
+    return std::make_pair(body() + internalStorageSize() - msgBody()->extraDataLength, msgBody()->extraDataLength);
+  }
+
  private:
   void setParams(NodeIdType sender,
                  ReqId reqSeqNum,
@@ -82,7 +88,8 @@ class ClientRequestMsg : public MessageBase {
                  uint64_t flags,
                  uint64_t reqTimeoutMilli,
                  const std::string& cid,
-                 uint32_t requestSignatureLen);
+                 uint32_t requestSignatureLen,
+                 uint32_t extraBufSize);
 };
 
 template <>
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -288,7 +288,8 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
                  clientMaxBatchSize_,
                  maxPreExecResultSize_,
                  preExecReqStatusCheckPeriodMilli_,
-                 numOfThreads));
+                 numOfThreads,
+                 ReplicaConfig::instance().preExecutionResultAuthEnabled));
   RequestProcessingState::init(numOfRequiredReplies(), &histograms_);
   PreProcessReplyMsg::setPreProcessorHistograms(&histograms_);
   addTimers();
@@ -1032,7 +1033,7 @@ void PreProcessor::cancelPreProcessing(NodeIdType clientId, uint16_t reqOffsetIn
 }
 
 void PreProcessor::finalizePreProcessing(NodeIdType clientId, uint16_t reqOffsetInBatch, const string &batchCid) {
-  std::unique_ptr<ClientRequestMsg> clientRequestMsg;
+  std::unique_ptr<impl::MessageBase> clientRequestMsg;
   const auto &batchEntry = ongoingReqBatches_[clientId];
   const auto &reqEntry = batchEntry->getRequestState(reqOffsetInBatch);
   {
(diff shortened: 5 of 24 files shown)


tdimitrov:
@yuliasherman @arc-vmware this is still work in progress. I suggest to delay you review with a few days.

bftengine/src/preprocessor/RequestProcessingState.cpp
```diff
@@ -103,7 +103,10 @@ void RequestProcessingState::handlePreProcessReplyMsg(const PreProcessReplyMsgSh
   if (preProcessReplyMsg->status() == STATUS_GOOD) {
     numOfReceivedReplies_++;
     const auto &newHashArray = convertToArray(preProcessReplyMsg->resultsHash());
-    preProcessingResultHashes_[newHashArray]++;  // Count equal hashes
+    // Counts equal hashes and saves the signatures with the replica ID. They will be used as a proof that the primary
+    // is sending correct preexecution result to the rest of the replicas.
+    preProcessingResultHashes_[newHashArray].emplace_back(preProcessReplyMsg->getResultHashSignature(),
+                                                          preProcessReplyMsg->senderId());
```
yuliasherman, line 109: Didn't you lose "++"?

bftengine/src/preprocessor/RequestProcessingState.cpp
```diff
@@ -199,4 +202,11 @@ unique_ptr<MessageBase> RequestProcessingState::buildClientRequestMsg(bool empty
   return clientPreProcessReqMsg_->convertToClientRequestMsg(emptyReq);
 }
 
+// Returns const reference to internal state. NOT THREAD SAFE!
```
yuliasherman, line 205: Actually, none of RequestProcessingState functions is thread-safe. Maybe worth adding a general comment that they are guarded by RequestState::mutex from PreProcessor.

bftengine/src/preprocessor/messages/PreProcessedResultMsg.cpp
```diff
@@ -0,0 +1,106 @@
+// Concord
+//
+// Copyright (c) 2018-2021 VMware, Inc. All Rights Reserved.
```
yuliasherman, line 3: Nit: completely new file: 2021

bftengine/src/preprocessor/messages/PreProcessedResultMsg.cpp
```diff
@@ -0,0 +1,106 @@
...
+PreProcessedResultMsg::PreProcessedResultMsg(NodeIdType sender,
+                                             uint64_t reqSeqNum,
+                                             uint32_t resultLength,
+                                             const char* result,
+                                             uint64_t reqTimeoutMilli,
+                                             const std::string& cid,
+                                             const concordUtils::SpanContext& spanContext,
+                                             const char* messageSignature,
+                                             uint32_t messageSignatureLen,
+                                             const char* resultSignaturesBuf,
```
yuliasherman, line 26: Consider passing a string instead.

bftengine/src/preprocessor/messages/PreProcessedResultMsg.cpp
```diff
@@ -0,0 +1,106 @@
...
+          "PreProcessResultSignature deserialisation error - remaining buffer length less than fixed size values size");
+    }
+
+    // Read fixed size values
+    sender_id = concordUtils::fromBigEndianBuffer<bftEngine::impl::NodeIdType>(buf + pos);
+    pos += sizeof(bftEngine::impl::NodeIdType);
+    signature_size = concordUtils::fromBigEndianBuffer<uint32_t>(buf + pos);
+    pos += sizeof(uint32_t);
+
+    if (signature_size > len - pos) {
```
yuliasherman, line 90: This should be verified in validate() function, which looks not implemented.

bftengine/src/preprocessor/messages/PreProcessedResultMsg.cpp
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  return output;
+}
+
+std::list<PreProcessResultSignature> PreProcessResultSignature::deserializeResultSignatureList(const char* buf,
+                                                                                               size_t len) {
+  size_t pos = 0;
+  std::list<PreProcessResultSignature> ret;
+  while (1) {
```
yuliasherman, line 75: We should know ahead how long is the signatures list. The signature size is also constant.

bftengine/src/preprocessor/messages/PreProcessedResultMsg.hpp
```diff
@@ -0,0 +1,73 @@
+// Concord
+//
+// Copyright (c) 2018-2021 VMware, Inc. All Rights Reserved.
```
yuliasherman, line 3: ->2021

yuliasherman (commented):
Great job!

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -668,6 +681,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
   }
 }
 
+bool ReplicaImp::validatePreProcessedResultSignatures(preprocessor::PreProcessedResultMsg &msg) {
+  auto [buf, buf_len] = msg.getResultSignaturesBuf();
+  auto sigs = preprocessor::PreProcessResultSignature::deserializeResultSignatureList(buf, buf_len);
```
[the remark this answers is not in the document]
tdimitrov: RVO kicks in here so no copy is made. We can't capture the return by reference, because it's a local variable inside deserializeResultSignatureList().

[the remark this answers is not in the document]
yuliasherman: OK, I've missed that we return a local variable.

bftengine/src/preprocessor/RequestProcessingState.cpp
tdimitrov: Sorry, but I don't follow. Which value should be incremented?

yuliasherman: Forget about it, it's OK.

bftengine/include/bftengine/ClientMsgs.hpp
```diff
@@ -40,6 +40,7 @@ struct ClientRequestMsgHeader {
   uint64_t timeoutMilli;
   uint32_t cidLength = 0;
   uint16_t reqSignatureLength = 0;
+  uint32_t extraDataLength = 0;
```
arc-vmware, line 43: Will this make Client request message backward incompatible. We should check if this addition effects serialisation and deserialisation of this message.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -673,6 +686,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+  const auto expectedSignatureCount = getReplicaConfig().fVal;
+  if (sigs.size() < expectedSignatureCount) {
+    LOG_ERROR(GL,
+              "PreProcessedResult signatures validation failure - not enough signatures. Expected "
+                  << expectedSignatureCount << " got " << sigs.size());
+    return false;
+  }
+
+  for (const auto &s : sigs) {
+    auto hash = concord::util::SHA3_256().digest(msg.requestBuf(), msg.requestLength());
```
arc-vmware, line 703: We should create the hashes in parallel. Its done in PrePrepareMsg creation.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -673,6 +686,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+    bool verificationResult = false;
+    if (getReplicaConfig().replicaId == s.sender_replica) {
+      std::vector<char> mySignature(sigManager_->getMySigLength(), 0);
```
arc-vmware, line 706: its better to pass '\0' instead of 0 as the second parameter is a char.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -673,6 +686,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+      sigManager_->sign((const char *)hash.data(), hash.size(), mySignature.data(), mySignature.size());
```
arc-vmware, line 707: we should static_cast instead of the c-style cast.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -673,6 +686,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+      verificationResult = mySignature == s.signature;
```
arc-vmware, line 708: if replica himself gets his own message, then will there be any chance of changes? If no then we can just avoid checking the signature for same replica.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -673,6 +686,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+    }
+  }
+
+  return true;
+}
+
+bool ReplicaImp::validatePrePrepareMsg(const PrePrepareMsg *msg, const bftEngine::impl::ReplicasInfo &replicasInfo) {
+  RequestsIterator reqIter(msg);
+  char *requestBody = nullptr;
+  while (reqIter.getAndGoToNext(requestBody)) {
```
arc-vmware, line 728: The validate() method of PrePrepareMsg is doing the same. We can call that function here.

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -480,6 +483,9 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
   void startConsensusProcess(PrePrepareMsg* pp);
   bool isSeqNumToStopAt(SeqNum seq_num);
 
+  bool validatePrePrepareMsg(const PrePrepareMsg* msg, const bftEngine::impl::ReplicasInfo& replicasInfo);
+  bool validatePreProcessedResultSignatures(preprocessor::PreProcessedResultMsg& msg);
```
arc-vmware, line 487: As this is validate [read only] method, should msg be const& ?

bftengine/src/bftengine/messages/ClientRequestMsg.hpp
```diff
@@ -75,14 +76,20 @@ class ClientRequestMsg : public MessageBase {
 
   void validateImp(const ReplicasInfo& repInfo) const;
 
+  // Returns a pair of pointer and size to the extra buffer which was allocated during initialisation
+  std::pair<char*, uint32_t> getExtraBufPtr() {
```
arc-vmware, line 80: nit: If we need only length then also this will result into buffer pointer calculation.

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
```diff
@@ -87,6 +87,15 @@ void PreProcessReplyMsg::validate(const ReplicasInfo& repInfo) const {
   }
 }  // namespace preprocessor
 
+std::vector<char> PreProcessReplyMsg::getResultHashSignature() const {
```
arc-vmware, line 90: nit: [very minor comment] void PreProcessReplyMsg::getResultHashSignature(std::vector<char> & result_hash) -- This will avoid result vector and copy ctors.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -800,7 +811,7 @@ void PreProcessor::cancelPreProcessing(NodeIdType clientId, uint16_t reqOffsetIn
 }
 
 void PreProcessor::finalizePreProcessing(NodeIdType clientId, uint16_t reqOffsetInBatch) {
-  std::unique_ptr<ClientRequestMsg> clientRequestMsg;
+  std::unique_ptr<impl::MessageBase> preProcessedMsg;
```
arc-vmware, line 814: nit: Nothing to change, just as assertion : if we keep this ClientRequestMsg then also it should be fine, since preprocess result msg is a type of ClientRequestMsg.

arc-vmware (commented):
Great work, enjoyed and learned quite a lot. I have few minor comments, feel free to defer if they are not relevant.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -812,21 +823,42 @@ void PreProcessor::finalizePreProcessing(NodeIdType clientId, uint16_t reqOffset
...
+        preProcessedMsg = make_unique<PreProcessedResultMsg>(clientId,
+                                                             reqSeqNum,
+                                                             reqProcessingStatePtr->getPrimaryPreProcessedResultLen(),
+                                                             reqProcessingStatePtr->getPrimaryPreProcessedResult(),
+                                                             reqProcessingStatePtr->getReqTimeoutMilli(),
+                                                             cid,
+                                                             span_context,
+                                                             reqProcessingStatePtr->getReqSignature(),
+                                                             reqProcessingStatePtr->getReqSignatureLength(),
+                                                             sigsBuf.data(),
```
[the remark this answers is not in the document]
tdimitrov: I agree, will do it for sigsBuf too

bftengine/src/preprocessor/messages/PreProcessedResultMsg.cpp
tdimitrov: I agree that we should know that there will be f signatures and its size is constant, but I have a reason to do it in such a way.

I put the signature lengths so that the message encoding/decoding is decoupled from the SIgManager. I think it's better to pass value sizes in the message unless they are a hard constant. In this case I don't consider them hard constant, because if we change the hashing algorithm this size will change. Replicas will easily figure this out, but the message will be unparsable outside Concord.

My motivation about the list size is similar - we can do for loop from 0 to F, but I wanted to implement more independant message format. I could have put the list size in the message too, but it seemed redundant for me.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -42,10 +42,12 @@
 #include "messages/ReplicaAsksToLeaveViewMsg.hpp"
 #include "messages/ReplicaRestartReadyMsg.hpp"
 #include "messages/ReplicasRestartReadyProofMsg.hpp"
+#include "messages/PreProcessedResultMsg.hpp"
```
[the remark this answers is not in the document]
tdimitrov: It was painful, but I did it ;)

bftengine/include/bftengine/ClientMsgs.hpp
tdimitrov: Actually this change doesn't affect the ClientRequestMsg. It's a mechanism to allocate extra buffer space if you are inheriting from the message (as it's done for PreProcess(ed)Result message).

The ClientRequestMsg message itself is not affected by this change because the extraDataLength is always set to 0 for it.

I also had concerns for serialisation/deserialisation so I added additional test cases in the message unit test. Check my changes in ClientRequestMsg_test.cpp/

bftengine/src/bftengine/ReplicaImp.cpp
tdimitrov: Good point, I'll do it.

bftengine/src/preprocessor/PreProcessor.cpp
tdimitrov: Initially we will keep both logics and switch between them via configuration. But later we have to remove the old approach (with ClientRequestMsg reuse), because it is not optimal.

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
tdimitrov: Actually no copy happens here too. RVO is applied for return-by-value vector. 

This function is called in RequestProcessingState::handlePreProcessReplyMsg(), as a parameter for emplace_back() which accepts universal reference. This is one of the cases where copy elision is applied so we are fine I think.

bftengine/src/bftengine/messages/ClientRequestMsg.hpp
tdimitrov: True, but we usually need both. Even if the pointer is not used I think it's a small price to pay.

bftengine/src/bftengine/ReplicaImp.hpp
tdimitrov: Good point. I tried this while writing the function. There is a call to msg.getResultSignaturesBuf() in it:
`const auto [buf, buf_len] = msg.getResultSignaturesBuf();`
getResultSignaturesBuf returns pointer so it's not correct to be const. This was the only reason to  get non-const reference in validatePreProcessedResultSignatures().

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -668,6 +681,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+  // f because the primary also executes the request (check PreProcessor::numOfRequiredReplies())
+  const auto expectedSignatureCount = getReplicaConfig().fVal;
+  if (sigs.size() < expectedSignatureCount) {
+    LOG_ERROR(GL,
+              "PreProcessedResult signatures validation failure - not enough signatures. Expected "
+                  << expectedSignatureCount << " got " << sigs.size());
+    return false;
+  }
+
+  for (const auto &s : sigs) {
```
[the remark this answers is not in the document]
tdimitrov: My idea about PreProcess(ed)ResultMsg is to be just a carrier for the signatures without actually validating them.

Initially I wanted to have a validate() method which does the signature checks. However this validate is called very early in the message processing code and if it fails the message is discarded. This doesn't work for us because we want to initiate a view change if signature verification fails. 

After considering pros and cons I decided that it's better to handle this in ReplicaImp because (1) it makes more sense to have a view change triggering logic there and (2) I didn't want to couple PreProcess(ed)ResultMsg with SigManager too much. I think this will bite us in the future when we decide to rework the message encoding/decoding code.

Does this make sense?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -668,6 +681,83 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+                "PreProcessedResult signatures validation failure - invalid signature "
+                    << KVLOG(s.sender_replica, msg.clientProxyId(), msg.getCid(), msg.requestSeqNum()));
+      return false;
+    }
+  }
+
+  return true;
+}
+
+bool ReplicaImp::validatePrePrepareMsg(const PrePrepareMsg *msg, const bftEngine::impl::ReplicasInfo &replicasInfo) {
```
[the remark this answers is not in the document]
tdimitrov: Please check my previous comment: https://github.com/vmware/concord-bft/pull/1639#discussion_r665335187

bftengine/src/preprocessor/messages/PreProcessReplyMsg.cpp
arc-vmware: yeah, i missed the RVO. this is absolutely fine. Thanks @tdimitrov

bftengine/src/bftengine/ReplicaImp.cpp
tdimitrov: We can't static_cast from unsigned to signed, so I'll replace it with reinterpret_cast.

bftengine/include/bftengine/ClientMsgs.hpp
arc-vmware: Checked the unit test, I think this should work.

bftengine/src/bftengine/ReplicaImp.cpp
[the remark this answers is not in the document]
yuliasherman: We all the time trying to offload the ReplicaImp code from a non-essential functionally, but the functions you added break this approach. I think it's less evil to have PreProcessResultMsg be aware of SigManager than adding all this functionality to ReplicaImp. In case a VC is required, the corresponding PreProcessResultMsg function needs just to return a flag.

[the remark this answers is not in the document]
tdimitrov: Yes, message validation can be separated from ViewChange triggering. I've fixed this.

bftengine/src/bftengine/ReplicaImp.cpp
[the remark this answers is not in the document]
tdimitrov: This is also fixed.

bftengine/src/preprocessor/messages/PreProcessedResultMsg.cpp
tdimitrov: This check is "deferred" due to the VC trigger we talked about. This is the only reason not to put it in validate().

Is this acceptable after the validation code is removed from ReplicaImp?

yuliasherman: Yes, thanks.

bftengine/src/bftengine/ReplicaImp.cpp
tdimitrov: Actually the hash is calculated only once and should be outside the for loop. Good catch!

tdimitrov:
Thanks for the feedback @yuliasherman @arc-vmware . I have addressed your comments. There are a bunch of commits which I will squash before merging.
Could you please approve the PR if you are satisfied with the changes?

yuliasherman (approved):
LGTM, 10x!

arc-vmware (approved):
Awesome, great feature getting in for concord.

PR #1640, opened 2021-06-30
Update value of reconfiguration_client_key_exchange

shgandhi:
This change is required to avoid collision with client app key type
value.


toly-kournik (approved):
LGTM!

PR #1562, opened 2021-06-09
Implement new PreProcessBatchRequestMsg

yuliasherman:
+ Add configuration parameter
+ Add spanContextSize to PreProcessRequestMsg header

PR #1641, opened 2021-06-30
Make use of PreProcessBatchReplyMsg

yuliasherman:
This MR implements the last part of PreProcessor full batch awareness functionality.
Now all PreProcessor messages have a batched version used in inter-replica communication for batched client requests, which should significantly reduce the IncomingStorage queue size.
The functionality is currently disabled by default using batchedPreProcessEnabled=false configuration parameter. 
Previous parts are https://github.com/vmware/concord-bft/pull/1562, https://github.com/vmware/concord-bft/pull/1602, and https://github.com/vmware/concord-bft/pull/1585.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.hpp
+++ b/bftengine/src/preprocessor/PreProcessor.hpp
@@ -14,6 +14,7 @@
 #include "OpenTracing.hpp"
 #include "messages/ClientBatchRequestMsg.hpp"
 #include "messages/PreProcessBatchRequestMsg.hpp"
+#include "messages/PreProcessBatchReplyMsg.hpp"
 #include "MsgsCommunicator.hpp"
 #include "MsgHandlersRegistrator.hpp"
 #include "SimpleThreadPool.hpp"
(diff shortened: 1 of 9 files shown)


bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -47,16 +48,56 @@ struct RequestState {
...
+ private:
+  PreProcessor &preProcessor_;
+  const uint16_t clientId_;
+  std::string cid_;
+  std::atomic_uint32_t batchSize_ = 0;
+  std::atomic_bool batchRegistered_ = false;
+  std::atomic_bool batchInProcess_ = false;
+  std::atomic_uint32_t numOfCompletedReqs_ = 0;
+  // pre-allocated map; request offset in batch -> request state
+  std::map<uint16_t, RequestStateSharedPtr> requestsMap_;
```
tdimitrov, line 87: As far as I understand this map is initialised once and later it is used only for reading. Maybe it can be initialised in the constructor and made const? 
Or at least a comment that it is intended for RO use and it is not protected with a mutex.

tdimitrov (approved):
I'm not very familiar with the feature but the code looks good. Just one minor comment.

bftengine/src/preprocessor/PreProcessor.hpp
yuliasherman: NP, will add in the next MR.

Liorgal28 (approved):
LGTM!

PR #1642, opened 2021-06-30
ConcordClient: Use std::function over templates

robem:
The main change in this PR is switching from templates to std::function. This is the result of a discussion with @f-squirrel. See https://github.com/vmware/concord-bft/pull/1628#discussion_r659796019 for more information. In addition, this PR fixes a bug with multiple subscriptions. Tests are still missing and will be introduced in the future. Please take a look at the commit messages for more details.


f-squirrel (approved):
Looks great, thank you @robem!

PR #1644, opened 2021-07-01
ClientService: Integrate ConcordClient into RequestService

robem:
This change is similar to the previous integration change for the EventService.
It fills the empty RequestService with concord client's send() method. The
latter will echo the request until the client pool moves to concord client.

Manual testing:
$> ./grpc_cli call localhost:1337 Send 'request:"Hello",read_only:false'
connecting to localhost:1337
response: "Hello"
Rpc succeeded with OK status


Liorgal28 (approved):
This looks really good @robem!

PR #1646, opened 2021-07-02
addRemove automated replica restart bug fix

nkumar04:
some replicas can get restart proof before reaching stable checkpoint,
in case of addRemove with (n-f) chorum. Replcias should not restart
before reaching the stable checkpoint. This change is to fix this issue


yontyon (approved):
I think we should associate the restart sequence number with the restart request.
This way, a replica will be able to differentiate between two different restart proofs (although, this is a rare situation to have).
in addition, we should define what is the behavior in case the replica failed **after** it got the restart proof. In this case, if we don't save in persistent storage the received message, we can endup without the ability to restart. However, in my opinion, we can rely in this case on state transfer and the epoch mechanism to catch up with the "right" epoch.
WDYT?

nkumar04:
> I think we should associate the restart sequence number with the restart request.
> This way, a replica will be able to differentiate between two different restart proofs (although, this is a rare situation to have).
> in addition, we should define what is the behavior in case the replica failed **after** it got the restart proof. In this case, if we don't save in persistent storage the received message, we can endup without the ability to restart. However, in my opinion, we can rely in this case on state transfer and the epoch mechanism to catch up with the "right" epoch.
> WDYT?

Good point, we should associate restartproof with the seq number. Regarding the second point, yes we can rely on state transfer and epoch mechanism i think.

yontyon (approved):
LGTM!
Thanks!

PR #1647, opened 2021-07-02
Add concurrent pruning parameters to PruneRequest

dartdart26:
Add the `tick_period_seconds` and `batch_blocks_num` parameters to
PruneRequest. Make them mandatory for the following reasons:
 * simplicity
 * make sure all replicas use the same values
 * if they were optional, replicas would need to use locally configured
   default ones and risk misconfiguration and/or different values on
   different replicas
 * flexibility, allowing users of the Operator to set different values
   in different periods (e.g. busy hours, low-traffic hours)

For now, these parameters are ignored. Will be used when concurrent
pruning is fully implemented.

Make sure verification takes these parameters into account and that
tests are aware of and using them.

### DIFF
--- a/kvbc/src/pruning_handler.cpp
+++ b/kvbc/src/pruning_handler.cpp
@@ -82,6 +82,11 @@ bool RSAPruningVerifier::verify(const concord::messages::PruneRequest& request) 
     return false;
   }
 
+  // Make sure pruning parameters are in range.
+  if (request.tick_period_seconds <= 0 || request.batch_blocks_num <= 0) {
+    return false;
+  }
+
   // Note RSAPruningVerifier does not handle verification of the operator's
   // signature authorizing this pruning order, as the operator's signature is a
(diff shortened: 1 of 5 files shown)


tdimitrov (approved):

teoparvanov (approved):
LGTM, thanks @dartdart26 !

kvbc/src/pruning_handler.cpp
```diff
@@ -82,6 +82,11 @@ bool RSAPruningVerifier::verify(const concord::messages::PruneRequest& request)
     return false;
   }
 
+  // Make sure pruning parameters are in range.
+  if (request.tick_period_seconds <= 0 || request.batch_blocks_num <= 0) {
```
f-squirrel, line 86: Nit: If I am not mistaken, both `tick_period_seconds` and `batch_blocks_num` are unsigned, I suggest using `==`.

f-squirrel (approved):
LGTM!

PR #1648, opened 2021-07-02
ConcordClient: Offer metrics API

robem:
In order to collect metrics in concord client and its subcomponents, we are
using our own basic metrics framework. We didn't want to depend on another 3rd
party dependency but the user can still choose to use Prometheus,
OpenTelemetry, ... on top of it.
The actual usage of the Aggregator will come with the integration of the client
pool and the thin replica client.

### DIFF
--- a/client/clientservice/src/main.cpp
+++ b/client/clientservice/src/main.cpp
@@ -16,6 +16,7 @@
 #include "Logger.hpp"
 #include "client_service.hpp"
 #include "client/concordclient/concord_client.hpp"
+#include "Metrics.hpp"
 
 using concord::client::clientservice::ClientService;
 using concord::client::concordclient::ConcordClient;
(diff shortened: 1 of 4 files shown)


client/clientservice/src/main.cpp
```diff
@@ -31,6 +32,8 @@ int main(int argc, char** argv) {
   // TODO: Setup ConcordClientConfig and read from config file
   ConcordClientConfig config{};
   auto concord_client = std::make_unique<ConcordClient>(config);
+  auto metrics = std::make_shared<concordMetrics::Aggregator>();
+  concord_client->setMetricsAggregator(metrics);
```
f-squirrel, line 36: Nit: I assume there is no scenario when we'd like to reset the aggregator in the runtime.
If my assumption is correct, we can pass the metrics aggregator right to the client's ctor without adding a new method.
WDYT?

f-squirrel (approved):
LGTM

Liorgal28 (approved):
LGTM!

client/clientservice/src/main.cpp
robem: The idea is that metrics are optional. If we have an aggregator than we use it. Otherwise, metrics are ignored.

shgandhi (approved):

PR #1652, opened 2021-07-04
TS: Harden TS requirements

f-squirrel:
This commit introduces additional checks and asserts for PrePrepare
messages with Time Service feature is ON. The checks are performed only
for non-empty PrePrepare messages; empty messages are created during
View Change.

The list of checks:
* PrePrepare contains a TS message
* PrePrepare contains only one TS message
* The TS message is the first in PrePrepare

 Additional changes:
* Add a method to check timestamp from PrePrepare
* Unit tests


yuliasherman (approved):
LGTM

PR #1654, opened 2021-07-05
Reading requests from storage using the new category

cloudnoize:
Enhancing the DB viewer to:
-  list requests that were executed as part of a block.
- Validate the signatures of the requests that were executed as part of a block.


Liorgal28 (approved):
LGTM!!!

PR #1655, opened 2021-07-05
TS: Add timestamp to IRequestHandler

f-squirrel:
* Updated all implementation of IRequestHandler::execute.
* Added logic to calculate the time offset for single requests within a
batch

Don't be afraid to review it, most of the changes is just an update for all the classes deriving from `IRequestHandler` :)

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -10,6 +10,7 @@
 // file.
 
 #include "PreProcessor.hpp"
+#include <optional>
 #include "InternalReplicaApi.hpp"
 #include "Logger.hpp"
 #include "MsgHandlersRegistrator.hpp"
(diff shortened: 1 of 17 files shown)


tdimitrov (approved):
Looks good to me. I've got one question only (as comment).
Good job @f-squirrel

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1120,7 +1121,7 @@ uint32_t PreProcessor::launchReqPreProcessing(uint16_t clientId,
       std::move(signature),
       maxPreExecResultSize_,
       (char *)getPreProcessResultBuffer(clientId, reqSeqNum, reqOffsetInBatch)});
-  requestsHandler_.execute(accumulatedRequests, cid, span);
+  requestsHandler_.execute(accumulatedRequests, std::nullopt, cid, span);
```
tdimitrov, line 1124: Just curious - why the timestamp is not set in preexecution?

f-squirrel: The reason is because in pre-execution we cannot guarantee full consensus on the timestamp.

glevkovich (approved):
LGTM

bftengine/src/preprocessor/PreProcessor.cpp
tdimitrov: Makes sense. Thanks!

PR #1657, opened 2021-07-06
A Fake clock to end to end test Time Service. 

ukg19:
Based on the command present in file, now() function will adjust the current time accordingly.


f-squirrel (approved):
Great job, @ukg19!

PR #1659, opened 2021-07-06
Reset control handler state after unwedge

nkgeorgiev:
Some checks like the 'wedge status' command rely on the control handler state.
Until now, we restarted the replicas to 'unwedge' and that reset the state.
Now that we have a proper unwedge mechanism, we should reset the state as part of it.


yontyon (approved):
LGTM!
Thanks

teoparvanov (approved):
LGTM, thanks!

PR #1664, opened 2021-07-06
Changed size of Primary queue for Client Requests.

HristoStaykov:
Because of the frequent report for dropping messages due to full queue
under heavy load we need to increase.


Liorgal28 (approved):
LGTM

PR #1665, opened 2021-07-07
refactored network-related methods from SkvbcTracker to SimpleKVBCProtocol

ananpal:
refactored network-related methods from SkvbcTracker to SimpleKVBCProtocol


teoparvanov (approved):
LGTM, thank you @ananpal ! 

You have cleaned-up the SkvbcTracker from "non-core" methods. The tracker now focuses on its main task - to track and linearize SKVBC read/write requests.

I have an idea for additional improvement and code dup reduction, but it can be done in a separate PR. This one can be merged as is IMO.

PR #1666, opened 2021-07-07
State transfer fix bug in calcMaxNumOfChunksInBlock

glevkovich:
calcMaxNumOfChunksInBlock() returns and uses uint16_t while the value should be uint32_t.
This may cause a wrong returned result in some input configuration (returned value of 0).


yontyon (approved):
LGTM!

PR #1668, opened 2021-07-07
Remove duplicates in requestsQueueOfPrimary.

HristoStaykov:
Because of the frequent dropping of Client Requests due to full queue under high load we need to clean duplicates more frequently. Prior to this change the actual filtration happened only when we create the PrePrepare. This is a quick fix. We might consider changing the data structure holding the Client Requests to something more suitable.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -361,14 +361,14 @@ void ReplicaImp::removeDuplicatedRequestsFromRequestsQueue() {
 
   LOG_INFO(GL, "" << KVLOG(requestsQueueOfPrimary.size()));
 
-  struct ClientsReqCmparator {
+  struct ClientsReqComparator {
     bool operator()(const ClientRequestMsg *const lhs, const ClientRequestMsg *const rhs) {
       return (lhs->clientProxyId() < rhs->clientProxyId() ||
               (lhs->clientProxyId() == rhs->clientProxyId() && lhs->requestSeqNum() < rhs->requestSeqNum()));
     }
   };
 
-  std::set<ClientRequestMsg *, ClientsReqCmparator> uniqueReqs;
+  std::set<ClientRequestMsg *, ClientsReqComparator> uniqueReqs;
   for (ClientRequestMsg *c : requestsQueueOfPrimary) {
     auto res = uniqueReqs.insert(c);
     if (res.second == false) {
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -363,7 +363,8 @@ void ReplicaImp::removeDuplicatedRequestsFromRequestsQueue() {
 
   struct ClientsReqCmparator {
     bool operator()(const ClientRequestMsg *const lhs, const ClientRequestMsg *const rhs) {
-      return (lhs->clientProxyId() < rhs->clientProxyId() || lhs->requestSeqNum() < rhs->requestSeqNum());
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -363,7 +363,8 @@ void ReplicaImp::removeDuplicatedRequestsFromRequestsQueue() {
 
   struct ClientsReqCmparator {
```
yuliasherman, line 364: Tipo: Cmparator -> Comparator

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -351,9 +355,31 @@ void ReplicaImp::removeDuplicatedRequestsFromRequestsQueue() {
   while (first != nullptr && !clientsManager->canBecomePending(first->clientProxyId(), first->requestSeqNum())) {
```
Liorgal28, line 355: Maybe I'm missing something, but still, this loop will stop at the first request that can become pending.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -351,9 +355,31 @@ void ReplicaImp::removeDuplicatedRequestsFromRequestsQueue() {
...
-    requestsQueueOfPrimary.pop();
+    requestsQueueOfPrimary.pop_front();
     first = (!requestsQueueOfPrimary.empty() ? requestsQueueOfPrimary.front() : nullptr);
   }
+
+  LOG_INFO(GL, "" << KVLOG(requestsQueueOfPrimary.size()));
+
+  struct ClientsReqCmparator {
+    bool operator()(const ClientRequestMsg *const lhs, const ClientRequestMsg *const rhs) {
+      return (lhs->clientProxyId() < rhs->clientProxyId() ||
```
Liorgal28, line 366: why not just <=?

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -104,8 +104,8 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
   SeqNum maxSeqNumTransferredFromPrevViews = 0;
 
   // requests queue (used by the primary)
-  std::queue<ClientRequestMsg*> requestsQueueOfPrimary;  // only used by the primary
-  size_t primaryCombinedReqSize = 0;                     // only used by the primary
+  std::list<ClientRequestMsg*> requestsQueueOfPrimary;  // only used by the primary
```
yuliasherman, line 107: Does a list container serve us better than a queue?

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: I did not change this loop, because I saw that we check the condition prior to insertion [ReplicaImp.cpp:261](https://github.com/vmware/concord-bft/blob/ff6416fa3d0f31c961a8d377460da61e1c466867/bftengine/src/bftengine/ReplicaImp.cpp#L261)

bftengine/src/bftengine/ReplicaImp.cpp
HristoStaykov: The std::set requires us to implement strict lower comparison on the data type, I think. We compare first the **clientProxyId**-s and if those are equal we further check the **requestSeqNum**-s.

bftengine/src/bftengine/ReplicaImp.hpp
HristoStaykov: It makes it easier (less code) when we populate with the filtered contents from the set.

PR #1669, opened 2021-07-07
BCStateTran: add new config param enableSourceBlocksPreFetch

glevkovich:
enableSourceBlocksPreFetch is set to true by default.
When true, pre-fetch is enabled - source try to fetch future batch(es)
blocks for improved performance.


toly-kournik (approved):
LGTM!

PR #1670, opened 2021-07-08
Define ++/--/+= operators for metrics instead of Get().Inc() pattern.

toly-kournik:


yontyon (approved):
LGTM!
Thanks!

util/include/Metrics.hpp
```diff
@@ -76,16 +78,19 @@ class Aggregator {
...
   explicit BasicGauge(const uint64_t val) : val_(val) {}
   BasicGauge(const BasicGauge& gauge) { val_ = (unsigned long)gauge.val_; }
   BasicGauge& operator=(const BasicGauge& gauge) {
     val_ = (unsigned long)gauge.val_;
     return *this;
   }
-  void Inc() { ++val_; }
-  void Dec() { --val_; }
+  // postfix
+  BasicGauge operator++(int) { return BasicGauge(val_++); }
```
arc-vmware, line 89: nit: [very small comment, feel free to ignore] : implementing a prefix will avoid creation of the temp object.
BasicGauge& operator++() { return ++val_; return *this;}
Also we will not keep any post fix operator so that no one can use x++ and thus improve the overall perf by reducing the object creation.

toly-kournik: I agree but it's against C++ semantics:
https://en.cppreference.com/w/cpp/language/operator_incdec
postfix operator returns the value before incrementing.

util/include/Metrics.hpp
```diff
@@ -76,16 +78,19 @@ class Aggregator {
...
+  // postfix
+  BasicGauge operator--(int) { return BasicGauge(val_--); }
   void Set(const uint64_t val) { val_ = val; }
```
f-squirrel, line 92: Nit: I know that this is an old code but I think it will be correct to change the argumant `const uint64_t val` to `const T& val`.

f-squirrel (approved):
Great CR, super logical, thank you @toly-kournik !

util/include/Metrics.hpp
arc-vmware: True. In all the scenarios where the inc() and dec() is used, pre and post fix can be used interchangeably. So for our purpose prefix and postfix are equivalent. So the suggestion here is not to use postfix but use prefix operator everywhere where inc and dec is needed. I agree that one of the test case need to be removed which is doing check(0, x++). For metric scenarios explicit postfix will never be needed, so its better to keep a prefix version and use that only.

arc-vmware (approved):
Great idea!!!

util/include/Metrics.hpp
toly-kournik: I'm not arguing about advantages of the prefix operations, but then the usage in the code should change from `metrics_++` to `++metrics_;

arc-vmware: Yes thats what I am saying. Since except for one unit test case, for all our scenarios ++metrics_ and metrics_++ are equivalent and we donot have any scenario where we specifically need postincrement. We need increment (either way, post or pre , is fine)/ decrement.

util/include/Metrics.hpp
toly-kournik: Just missed it, will do it on another occasion - there are other things to refactor there

PR #1671, opened 2021-07-08
Fix batched inter-replicas issues

yuliasherman:
- Improve batched inter-replicas tracking through the logs
- Enable batched inter-replica communication in Apollo tests by default
- Don't cancel the whole batch when some of its messages failed to pass sanity checks


Liorgal28 (approved):
LGTM! thanks

PR #1673, opened 2021-07-09
Time service apollo tests

ukg19:
This PR adds apollo tests for Time Service by using a fake clock to drift the current time of Replicas by given amount
It tests Time Service for following scenarios :
1. Drifting time in container of primary replica
2. Drifting time in container of non-primary replica
3. Both 1) and 2) by disabling time service.

Macro USE_FAKE_CLOCK_IN_TIME_SERVICE is defined to enable the test cases and fake clock functionality which is not set by Default. 

This PR also adds these test cases to CI.


ukg19:
Got issue with merge. closing this one, raised another.

PR #1674, opened 2021-07-09
Use Metrics.hpp to update metrics in TRC/S

shgandhi:
This PR updates TRC/S such that they use concord-bft metrics.hpp to update metrics instead of supporting third-party metric frameworks such as prometheus via the use of callbacks. The main motivation of this change is to avoid potential performance degradation and to reduce complexity.

### DIFF
--- a/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
+++ b/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
@@ -88,8 +88,9 @@ class ThinReplicaImpl {
   const std::string kCorrelationIdTag = "cid";
 
  public:
-  ThinReplicaImpl(std::unique_ptr<ThinReplicaServerConfig> config, std::unique_ptr<ThinReplicaServerMetrics> metrics)
-      : logger_(logging::getLogger("concord.thin_replica")), config_(std::move(config)), metrics_(std::move(metrics)) {}
+  ThinReplicaImpl(std::unique_ptr<ThinReplicaServerConfig> config,
+                  std::shared_ptr<concordMetrics::Aggregator> aggregator)
+      : logger_(logging::getLogger("concord.thin_replica")), config_(std::move(config)), aggregator_(aggregator) {}
 
   ThinReplicaImpl(const ThinReplicaImpl&) = delete;
   ThinReplicaImpl(ThinReplicaImpl&&) = delete;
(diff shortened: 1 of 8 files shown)


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -196,22 +197,24 @@ class ThinReplicaImpl {
     if (!subscribe_status.ok()) {
       return subscribe_status;
     }
-
-    metrics_->setSubscriberListSize(config_->subscriber_list.Size());
+    // TRS metrics
+    ThinReplicaServerMetrics metrics_(stream_type, getClientId(context));
+    metrics_.setAggregator(aggregator_);
+    metrics_.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
```
robem, line 203: `.Get().Set()` Can't we just call `.Set()` instead?

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -253,11 +256,12 @@ class ThinReplicaImpl {
...
         break;
       }
-      metrics_->setLastSentBlockId(stream_type, getClientId(context), update.block_id);
+      metrics_.last_sent_block_id.Get().Set(update.block_id);
     }
     config_->subscriber_list.removeBuffer(live_updates);
     live_updates->removeAllUpdates();
-    metrics_->setSubscriberListSize(config_->subscriber_list.Size());
+    metrics_.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
+    metrics_.updateAggregator();
```
robem, line 264: When do we need to update? Given that we should expect that this function never returns because it is an infinite stream of data, maybe we need to call it inside the while-loop too? Not sure, let's check `Metrics.hpp`.

robem (approved):
Thanks for changing metrics again :) Much appreciated 💪
Nice commits and description. Thank you!

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: Thanks for catching this, updated.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -59,17 +59,21 @@ struct ThinReplicaServerConfig {
...
+  // the threshold after which metric aggregator is updated
+  uint16_t update_metrics_aggregator_thresh;
 
   ThinReplicaServerConfig(const bool is_insecure_trs_,
                           const std::string& tls_trs_cert_path_,
                           const concord::kvbc::IReader* rostorage_,
                           SubBufferList& subscriber_list_,
-                          std::unordered_set<std::string>& client_id_set_)
+                          std::unordered_set<std::string>& client_id_set_,
+                          uint16_t update_metrics_aggregator_thresh_)
```
robem, line 70: Let's use a default value `100` here. This parameter is not too important to the caller. Only if the default doesn't work then we are interested in modifying it.

PR #1676, opened 2021-07-10
Clientservice: Command line options and basic config

robem:
With this change we introduce command line options to the clientservice. One of the options expects a YAML configuration file. We changed [install_deps.sh in a separate PR](https://github.com/vmware/concord-bft/pull/1680).
Once the config file is parsed we fill `ConcordClientConfig`. There are still some holes and not all information is present yet but this change gives a good starting point to add more later instead of bloating this PR. Note: ConcordClientConfig is actively used yet. Its users will be the thin replica client and the client pool once they integrate.

### DIFF
--- a/client/clientservice/CMakeLists.txt
+++ b/client/clientservice/CMakeLists.txt
@@ -1,15 +1,18 @@
 add_subdirectory("proto")
 
 find_package(GRPC REQUIRED)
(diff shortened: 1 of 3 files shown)


Liorgal28 (approved):
LGTM, as we talked this could be the solution for now but its tricky to track changes on configurations with this solution.

PR #1678, opened 2021-07-11
Improve checkpoint summary consensus time

mdarade:
While comparing checkpoint summary messages received from 2 different
replicas there is no need to compare sequence numbers as some of
the checkpoint summary messages could have been received in earlier
window before retransmission happens.


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -1106,6 +1107,7 @@ bool BCStateTran::onMessage(const CheckpointSummaryMsg *m, uint32_t msgLen, uint
   LOG_DEBUG(getLogger(), "");
 
   FetchingState fs = getFetchingState();
+  LOG_DEBUG(getLogger(), "Fetching state is " << stateName(fs));
```
glevkovich, line 1218: This line is redundant, we have an assert one line after, so the fetching state is obvious.
Also, we don't need to write fetch state all the time.
now we a new log which prints every time state change (info level):
`FetchingState changed from`

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -659,6 +659,7 @@ void BCStateTran::onTimerImp() {
   FetchingState fs = getFetchingState();
   if (fs == FetchingState::GettingCheckpointSummaries) {
     if ((currTime - lastTimeSentAskForCheckpointSummariesMsg) > config_.checkpointSummariesRetransmissionTimeoutMs) {
+      LOG_DEBUG(getLogger(), "Retransmitting AskForCheckpointSummaries");
```
glevkovich, line 701: The main reason I would add a log here and not inside sendAskForCheckpointSummariesMsg (there is one there already, but just printing vars), is to print values of retransmissionNumberOfAskForCheckpointSummariesMsg, kResetCount_AskForCheckpointSummaries.
It's easy to find a send without this by searching `sendAskForCheckpointSummariesMsg`

bftengine/src/bcstatetransfer/Messages.hpp
```diff
@@ -66,11 +66,18 @@ struct CheckpointSummaryMsg : public BCStateTranBaseMsg {
   uint64_t requestMsgSeqNum;
 
   static bool equivalent(const CheckpointSummaryMsg* a, const CheckpointSummaryMsg* b) {
-    return (memcmp(a, b, sizeof(CheckpointSummaryMsg)) == 0);
+    return (a->lastBlock == b->lastBlock and
+      a->checkpointNum == b->checkpointNum and
+      (memcmp(&a->digestOfLastBlock, &b->digestOfLastBlock, sizeof(STDigest)) == 0) and
+      (memcmp(&a->digestOfResPagesDescriptor, &b->digestOfResPagesDescriptor, sizeof(STDigest)) == 0));
   }
```
glevkovich, line 72: I see that for some expressions you put brackets and some not. I recommend having brackets around all. same for the next change.

bftengine/src/bcstatetransfer/Messages.hpp
```diff
@@ -66,11 +66,18 @@ struct CheckpointSummaryMsg : public BCStateTranBaseMsg {
...
 
   static bool equivalent(const CheckpointSummaryMsg* a, uint16_t a_id, const CheckpointSummaryMsg* b, uint16_t b_id) {
-    if (memcmp(a, b, sizeof(CheckpointSummaryMsg)) != 0) {
+    if (a->lastBlock != b->lastBlock ||
```
glevkovich: If someone adds fields to this message in the future, and they should be compared too, we might have a bug.
I suggest adding a static assert here with this bool expression(pseudo-code):
`sizeof(CheckpointSummaryMsg) - sizeof(CheckpointSummaryMsg::requestMsgSeqNum == size of all compared elements`
Also, at the end add a warning. That way if class change the developer can decide if to compare the new field or  not.

PR #1679, opened 2021-07-11
Introducing epoch manager

yontyon:
In this PR we introduce the epoch manager and demonstrate its power via add/remove node command with lagged replicas.
**Epoch** is a life cycle of the system. by changing epoch we indicate that we actually started a new system on top of an old state.
All replicas have to decide on starting a new epoch exactly at the same state to preserve consistency. Thus, starting a new epoch after wedge makes great sense.
Practically, when a replica decides to start a new epoch, it writes a new "genesis" block that contains this new epoch under a pre-known key.

### DIFF
--- a/bftengine/include/bftengine/EpochManager.hpp
+++ b/bftengine/include/bftengine/EpochManager.hpp
@@ -24,8 +24,6 @@ class EpochManager : public bftEngine::ResPagesClient<EpochManager, 1> {
     EpochData() = default;
     EpochData(uint64_t epochNumber) : epochNumber_{epochNumber} {}
 
-    const std::string getVersion() const override { return "1"; }
-
     void serializeDataMembers(std::ostream& outStream) const override { serialize(outStream, epochNumber_); }
     void deserializeDataMembers(std::istream& inStream) override { deserialize(inStream, epochNumber_); }
   };
@@ -42,22 +40,22 @@ class EpochManager : public bftEngine::ResPagesClient<EpochManager, 1> {
     static EpochManager instance_;
     return instance_;
   }
-  uint64_t getEpochNumber(bool fromPages) {
-    if (!fromPages) return epochNumber_;
+  uint64_t getSelfEpochNumber() { return epochNumber_; }
+  uint64_t getGlobalEpochNumber() {
     if (!loadReservedPage(0, sizeOfReservedPage(), page.data())) return 0;
     EpochData edata;
     std::istringstream inStream;
     inStream.str(page);
     concord::serialize::Serializable::deserialize(inStream, edata);
-    epochNumber_ = edata.epochNumber_;
     return edata.epochNumber_;
   }
-  void setEpochNumber(uint64_t newEopch, bool toPages) {
-    epoch_number_gauge_.Get().Set(newEopch);
+  void setSelfEpochNumber(uint64_t newEpoch) {
+    epochNumber_ = newEpoch;
+    epoch_number_gauge_.Get().Set(newEpoch);
     metrics_.UpdateAggregator();
-    epochNumber_ = newEopch;
-    if (!toPages) return;
-    EpochData edata{newEopch};
+  }
+  void setGlobalEpochNumber(uint64_t newEpoch) {
+    EpochData edata{newEpoch};
     std::ostringstream outStream;
     concord::serialize::Serializable::serialize(outStream, edata);
     auto data = outStream.str();
--- a/bftengine/include/bftengine/EpochManager.hpp
+++ b/bftengine/include/bftengine/EpochManager.hpp
@@ -1,0 +1,84 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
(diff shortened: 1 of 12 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -346,6 +341,14 @@ void BCStateTran::init(uint64_t maxNumOfRequiredStoredCheckpoints,
 
       ConcordAssertEQ(getFetchingState(), FetchingState::NotFetching);
     }
+    {
```
toly-kournik, line 344: All the initialization was a single transaction.

bftengine/include/bftengine/EpochManager.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <functional>
+
+namespace bftEngine {
+class EpochManager : public bftEngine::ResPagesClient<EpochManager, 1> {
+  struct EpochData : public concord::serialize::SerializableFactory<EpochData> {
+    uint64_t epochNumber_;
+    EpochData() = default;
+    EpochData(uint64_t epochNumber) : epochNumber_{epochNumber} {}
+
+    const std::string getVersion() const override { return "1"; }
```
toly-kournik, line 27: no need in `getVersion()` anymore

bftengine/include/bftengine/EpochManager.hpp
```diff
@@ -0,0 +1,84 @@
...
+    page.resize(sizeOfReservedPage());
+    metrics_.Register();
+  }
+
+ public:
+  static EpochManager& instance() {
+    static EpochManager instance_;
+    return instance_;
+  }
+  uint64_t getEpochNumber(bool fromPages) {
```
toly-kournik, line 45: what's use case when you don't want to get the most updated version?

bftengine/include/bftengine/EpochManager.hpp
```diff
@@ -0,0 +1,84 @@
...
+    if (!fromPages) return epochNumber_;
+    if (!loadReservedPage(0, sizeOfReservedPage(), page.data())) return 0;
+    EpochData edata;
+    std::istringstream inStream;
+    inStream.str(page);
+    concord::serialize::Serializable::deserialize(inStream, edata);
+    epochNumber_ = edata.epochNumber_;
+    return edata.epochNumber_;
+  }
+  void setEpochNumber(uint64_t newEopch, bool toPages) {
```
toly-kournik, line 55: Same here, why not to always persist epoch num?

bftengine/include/bftengine/EpochManager.hpp
```diff
@@ -0,0 +1,84 @@
...
+    std::ostringstream outStream;
+    concord::serialize::Serializable::serialize(outStream, edata);
+    auto data = outStream.str();
+    saveReservedPage(0, data.size(), data.data());
+  }
+
+  void startNewEpoch() { startNewEpoch_ = true; }
+  void markEpochAsStarted() { startNewEpoch_ = false; }
+  bool isNewEpoch() { return startNewEpoch_; }
+  void setNewEpochFlagHandler(const std::function<void(bool)>& cb) { setNewEpochFlagCallBack_ = cb; }
```
toly-kournik, line 70: consider using callback subscription mechanism as in ST callbacks if suitable

bftengine/include/bftengine/EpochManager.hpp
yontyon: in some cases, the replica wants to get its own epoch number and not the one that was updated in the reserved pages from other replicas.
actually, the only reason to use reserved pages is for read only replicas.

bftengine/include/bftengine/EpochManager.hpp
yontyon: we want to persist the epoch number only when we sure that other replicas do that again. If we catch up from a previous epoch, we want to update only our local epoch and not in the reserved pages

bftengine/include/bftengine/EpochManager.hpp
yontyon: I have only a single callback here, if in the future there will be more, I will change it

bftengine/include/bftengine/EpochManager.hpp
toly-kournik: it complicates the logic, hard to understand

bftengine/include/bftengine/EpochManager.hpp
toly-kournik: why to use a flag?

kvbc/src/Replica.cpp
```diff
@@ -158,6 +159,40 @@ void Replica::createReplicaAndSyncState() {
       std::terminate();
     }
   }
+
+  uint64_t epoch = 0;
```
toly-kournik, line 163: `createReplicaAndSyncState()` becomes a beast, maybe make a separate function?

bftengine/include/bftengine/EpochManager.hpp
toly-kournik: `setNewEpochHandler()` ?

yontyon: we sometimes (after state transfer) want to cancel the decision to start a new epoch

toly-kournik (approved):
LGTM!

PR #1684, opened 2021-07-13
Fix some critical bugs in reconfiguration mechanism

yontyon:
in this PR we fix some critical bugs with the reconfiguration mechanism, mostly on remove metadata from state transfer persistent storage
1. removing the whole metadata of state transfer from storage (including reserved pages)
2. mark the bft metadata as initialized after removing metadata on startup


dartdart26:
Thanks for fixing them @yontyon! Could you please briefly describe what these were and their effects?

PR #1688, opened 2021-07-14
Initialise with zeros the array for primary result hash in PreProcessing

tdimitrov:
primaryPreProcessResultHash_ in RequestProcessingState is of type
concord::util::SHA3_256::Digest which actually is std::array.

std::array's default constructor default constructs each element of the
array which for POD types effectively means nothing (whatever was in the
memory).

This was causing the preprocessor test failure in
primaryReplicaDidNotCompletePreProcessingWhileNonPrimariesDid test case.

The patch initialises the array with 0 on construction.


tdimitrov:
Courtesy to @pkthapa for finding the issue!

dartdart26 (approved):

PR #1689, opened 2021-07-14
Implement ClientExchangePublicKey handling.

toly-kournik:
It's the second phase of an on-demand client key exchange sequence
after client key exchange command issued by an operator.

### DIFF
--- a/bftengine/src/bftengine/KeyExchangeManager.cpp
+++ b/bftengine/src/bftengine/KeyExchangeManager.cpp
@@ -14,6 +14,7 @@
 #include "InternalBFTClient.hpp"
 #include "kvstream.h"
 #include "json_output.hpp"
+#include "SigManager.hpp"
 #include <thread>
 
 namespace bftEngine::impl {
@@ -161,11 +162,15 @@ void KeyExchangeManager::onPublishClientsKeys(const std::string& keys, std::opti
       LOG_FATAL(KEY_EX_LOG, "Initial Published Clients keys and replica client keys do not match");
       ConcordAssertEQ(keys, *bootstrap_keys);
     }
-    if (impl::KeyExchangeManager::instance().clientKeysPublished()) save = false;
+    if (clientKeysPublished()) save = false;
   }
-  if (save) {
-    impl::KeyExchangeManager::instance().saveClientsPublicKeys(keys);
-  }
+  if (save) saveClientsPublicKeys(keys);
+}
+
+void KeyExchangeManager::onClientPublicKeyExchange(const std::string& key, std::uint16_t clientId) {
+  LOG_INFO(KEY_EX_LOG, "key: " << key << " client: " << clientId);
+  SigManager::instance()->setClientPublicKey(key, clientId, KeyFormat::PemFormat);
+  saveClientsPublicKeys(SigManager::instance()->getClientsPublicKeys());
 }
 
 void KeyExchangeManager::sendInitialKey() {
--- a/bftengine/src/bftengine/SigManager.cpp
+++ b/bftengine/src/bftengine/SigManager.cpp
@@ -118,11 +118,11 @@ SigManager::SigManager(PrincipalId myId,
           metrics_component_.RegisterAtomicCounter("peer_replicas_signature_verification_failed"),
           metrics_component_.RegisterAtomicCounter("peer_replicas_signatures_verified"),
           metrics_component_.RegisterAtomicCounter("signature_verification_failed_on_unrecognized_participant_id")} {
-  map<KeyIndex, RSAVerifier*> publicKeyIndexToVerifier;
+  map<KeyIndex, std::shared_ptr<RSAVerifier>> publicKeyIndexToVerifier;
   size_t numPublickeys = publickeys.size();
 
   ConcordAssert(publicKeysMapping.size() >= numPublickeys);
-  mySigner_ = new RSASigner(mySigPrivateKey.first.c_str(), mySigPrivateKey.second);
+  mySigner_.reset(new RSASigner(mySigPrivateKey.first.c_str(), mySigPrivateKey.second));
   for (const auto& p : publicKeysMapping) {
     ConcordAssert(verifiers_.count(p.first) == 0);
     ConcordAssert(p.second < numPublickeys);
@@ -130,7 +130,7 @@ SigManager::SigManager(PrincipalId myId,
     auto iter = publicKeyIndexToVerifier.find(p.second);
     const auto& [key, format] = publickeys[p.second];
     if (iter == publicKeyIndexToVerifier.end()) {
-      verifiers_[p.first] = new RSAVerifier(key.c_str(), format);
+      verifiers_[p.first] = std::make_shared<RSAVerifier>(key.c_str(), format);
       publicKeyIndexToVerifier[p.second] = verifiers_[p.first];
     } else {
       verifiers_[p.first] = iter->second;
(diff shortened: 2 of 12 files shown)


toly-kournik:
Still missing SigManager initialization with keys from reserved pages if the was a key exchange.
Work in progress

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -281,12 +281,27 @@ async def test_client_key_exchange_command(self, bft_network):
...
-            rep = await op.client_key_exchange(all_client_ids)
+            rep = await op.client_key_exchange_command(all_client_ids)
             rep = cmf_msgs.ReconfigurationResponse.deserialize(rep)[0]
             assert rep.success is True        
             log.log_message(message_type=f"block_id {rep.response.block_id}")
             assert rep.response.block_id == 1
     
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7)
+    async def test_client_exchange_public_key(self, bft_network):
```
yontyon, line 292: Currently, we are not validating the client reconfiguration commands (it will be done in the near future), once it is done, this test won't work as the operator key won't match the expected signature (unless we sign with the actual bft client id). Do u think we should enable this kind of operation from the operator as well?

toly-kournik: We'll change the tests

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -161,11 +162,15 @@ void KeyExchangeManager::onPublishClientsKeys(const std::string& keys, std::opti
...
-    if (impl::KeyExchangeManager::instance().clientKeysPublished()) save = false;
-  }
-  if (save) {
-    impl::KeyExchangeManager::instance().saveClientsPublicKeys(keys);
+    if (clientKeysPublished()) save = false;
   }
+  if (save) saveClientsPublicKeys(keys);
+}
+
+void KeyExchangeManager::onClientPublicKeyExchange(const std::string& key, std::uint16_t clientId) {
```
cloudnoize, line 170: On client key exchange we need to set ```ClientKeyStore::published_``` to false and I think we need to call ```sendInitialClientsKeys``` with ```SigManager::instance()->getClientsPublicKeys()``` .
This should publish the new key set

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -167,46 +167,35 @@ SigManager::SigManager(PrincipalId myId,
...
-}
-
 uint16_t SigManager::getSigLength(PrincipalId pid) const {
   if (pid == myId_) {
     return (uint16_t)mySigner_->signatureLength();
   } else {
-    auto pos = verifiers_.find(pid);
-    if (pos == verifiers_.end()) {
+    std::shared_lock lock(mutex_);
+    if (auto pos = verifiers_.find(pid); pos != verifiers_.end()) {
```
glevkovich, line 175: nit: this way is similar to a for loop initialization, and is new in c++17, right?

Also, a question - I'm not familiar  with the client key exchange, but in case you have an old request signed with the old key, we assume retransmissions will work? Is the client's private key exchanged exactly at the same time? I see that retransmission enforces a new creation (and signing) of a message (throws TimeoutException to the caller), so the old message is not kept.

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -167,46 +167,35 @@ SigManager::SigManager(PrincipalId myId,
...
-    if (alreadyDeleted.find(v.second) == alreadyDeleted.end()) {
-      delete v.second;
-      alreadyDeleted.insert(v.second);
-    }
-  }
-}
-
 uint16_t SigManager::getSigLength(PrincipalId pid) const {
   if (pid == myId_) {
     return (uint16_t)mySigner_->signatureLength();
```
glevkovich, line 172: nit: Can you please onvert this to static cast?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -240,5 +229,16 @@ void SigManager::sign(const char* data, size_t dataLength, char* outSig, uint16_
 
 uint16_t SigManager::getMySigLength() const { return (uint16_t)mySigner_->signatureLength(); }
 
+void SigManager::setClientPublicKey(const std::string& key, PrincipalId id, KeyFormat format) {
+  if (replicasInfo_.isIdOfExternalClient(id)) {
+    std::unique_lock lock(mutex_);
+    verifiers_.insert_or_assign(id, std::make_shared<RSAVerifier>(key.c_str(), format));
```
glevkovich, line 235: I don't that insert_or_assign should be used here, at least not like that. You are interested at this stage to replace keys right?
That means if a key was not in the map, something is wrong with the specific operation you want to perform.

I agree that we might want in the future to have an option to insert a new key, but the 2 operations should be distinct, at least in terms of catching errors and warning the user about errors.

For now, the only operation we want to perform is to "replace the key with an already existing one".
So if the key was not there and inserted, something very wrong happened, and the caller won't know about it.

I suggest adding some flag "replaceOnly" on the intention of the user and. according to that perform the operation.
If replaceOnly is true, only accept assign (throw an exception if a key wasn't there). else use insert_and_assign.

glevkovich (changes requested):

bftengine/src/bftengine/SigManager.hpp
```diff
@@ -68,13 +67,16 @@ class SigManager {
...
+  void setClientPublicKey(const std::string& key, PrincipalId, KeyFormat);
+
   SigManager(const SigManager&) = delete;
   SigManager& operator=(const SigManager&) = delete;
   SigManager(SigManager&&) = delete;
   SigManager& operator=(SigManager&&) = delete;
 
   concord::messages::keys_and_signatures::ClientsPublicKeys clientsPublicKeys_;
   std::string getClientsPublicKeys() {
+    std::shared_lock lock(mutex_);
```
glevkovich, line 79: I don't think it's atomic.
you are replacing key by key. 
So if I want to change 10 keys, and call setClientPublicKey 5 times, then call getClientsPublicKeys, I will get 5 old keys and 5 new keys.
I think we should change the API above to setClientPublicKeys to replace all keys at once.

glevkovich (changes requested):

bftengine/src/bftengine/SigManager.hpp
glevkovich: Hi, after our zoom conversation I'm not sure if this comment is correct. I mean, the solution implements the interface definition to replace key by key. But on the other hand, I'm not sure what will happen if someone calls getClientsPublicKeys while replacing key by key (in the middle of the replacement). Need to think if it's a valid well-defined scenario.

bftengine/src/bftengine/SigManager.cpp
toly-kournik: If a client has changed its key, it should sign with a new one

bftengine/src/bftengine/SigManager.hpp
toly-kournik: `getClientsPublicKeys()` is used in the initial client public keys publishing scenario only, so I don't see how it can create a problem.

bftengine/src/bftengine/SigManager.cpp
toly-kournik: `replicasInfo_.isIdOfExternalClient(id)` is checking the validity of the client, why to restrict ourselves? It can be a new client after reconfiguration for example. After all, SigManager is responsible for holding signer and verifiers by client id and all the key exchange logic is out of its scope.

cloudnoize (approved):
LGTM!

bftengine/src/bftengine/SigManager.hpp
glevkovich: OK, if you take only this use case, I agree. 
But in the future, if used in other use-cases it can be a problem.
I suggest adding a comment above, stating that this might be not atomic.

bftengine/src/bftengine/SigManager.cpp
glevkovich: But if there is only one use case (see the other comment https://github.com/vmware/concord-bft/pull/1689#discussion_r672129719 ), then any other case is an error (a bug).
I didn't say to restrict. I suggested making the interface more clear, to reflect the user intention and catch bugs:
`void SigManager::setClientPublicKey(const std::string& key, PrincipalId id, KeyFormat format, boolkeyMustExist)`

if boolkeyMustExist is true then only insert.
if boolkeyMustExist is false,  insert_or_assign.

bftengine/src/bftengine/SigManager.hpp
toly-kournik: To tell the truth I don't see neither atomicity nor consistency problem here. `getClientPublicKeys()` returns keys updated to the specific moment.

bftengine/src/bftengine/SigManager.cpp
toly-kournik: Sorry, from my point of view if `KeyExchangeMgr` has decided that key exchange was legal, it just sets the key for the client. I don't see any case when I use the `boolKeyMustExist`.

bftengine/src/bftengine/SigManager.hpp
glevkovich: If someone calls getClientPublicKeys during an update (which is a use-case you've stated is impossible for now), it will get an output: mix of old a new keys. If this will be used later (whatever use) there will probably be a bug.

It is the responsibility of the internal implementation to prevent such inconsistencies.

A caller is not interested in a mix of old and new keys, and might be unaware of the background key-exchange process that randomly happened during the call.

toly-kournik: that's why there's a published_ variable

toly-kournik: I think we are stuck on the esoteric scenario

glevkovich: I agree, this became a less practical discussion. 
In at worst case, will have to deal with a bug years/months from now and/or re-design this interface  :)

glevkovich (approved):

PR #1690, opened 2021-07-14
ConcordClient: TRC integration

robem:
This PR integrates the thin replica client into concord client. There is more to be done but with this change we got to play around with TRC initialization, using ConcordClientConfig, and subscribe/unsubscribe.
See commit messages for detail.

$> ./grpc_cli call localhost:1337 StreamEventGroups 'event_group_id:12'
connecting to localhost:1337

### DIFF
--- a/client/concordclient/src/concord_client.cpp
+++ b/client/concordclient/src/concord_client.cpp
@@ -35,13 +35,13 @@ void ConcordClient::send(const bft::client::WriteConfig& config,
 void ConcordClient::subscribe(const SubscribeRequest& request,
                               const std::unique_ptr<opentracing::Span>& parent_span,
                               const std::function<void(SubscribeResult&&)>& callback) {
-  if (not stop_subscriber_) {
+  if (subscriber_) {
     LOG_ERROR(logger_, "subscription already in progress - unsubscribe first");
     throw SubscriptionExists();
   }
 
   stop_subscriber_ = false;
-  subscriber_ = std::thread([&] {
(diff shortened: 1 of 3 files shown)


client/concordclient/src/concord_client.cpp
```diff
@@ -35,13 +35,13 @@ void ConcordClient::send(const bft::client::WriteConfig& config,
...
-  if (not stop_subscriber_) {
+  if (subscriber_) {
     LOG_ERROR(logger_, "subscription already in progress - unsubscribe first");
     throw SubscriptionExists();
   }
 
   stop_subscriber_ = false;
-  subscriber_ = std::thread([&] {
+  subscriber_ = std::make_unique<std::thread>([&] {
     while (not stop_subscriber_) {
```
shgandhi, line 68: Why are we still checking the flag `stop_subscriber_` here, instead of checking the pointer like we do above?

shgandhi (approved):
Thanks for the change @robem! LGTM! I just have one minor comment.

client/concordclient/src/concord_client.cpp
robem: The pointer represents the thread. The while-loop is within the thread so it would always be true. In this while-loop we specifically want to check if the "outside" (the gRPC server) wants us to stop.

shgandhi: Got it, thanks for the explanation.

PR #1691, opened 2021-07-15
Fix a race condition in RSASigner::Impl::sign()

tdimitrov:
The sign() method of RSASigner::Impl is const and thread-safe by itself
however it uses an instance of RandomPool which is not thread safe. As a
result concurrent calls to sign() lead to data races inside the
RandomPool.

This patch fixes this by making the RandomPool instance thread local.
This way each thread works with its own object and there are no races.

RandomPool is wrapped in a new type because an adequate default
constructor needs to be provided. Otherwise the object initialisation
becomes complicated.

Big  thanks to @dartdart26 for helping with the fix.

### DIFF
--- a/bftengine/src/bftengine/Crypto.cpp
+++ b/bftengine/src/bftengine/Crypto.cpp
@@ -58,23 +58,18 @@ namespace impl {
 //#define RSA_STANDARD PKCS1v15
 //#define RSA_STANDARD OAEP<SHA>
 
-static RandomPool sGlobalRandGen;  // not thread-safe !!
+struct RandomPoolWrapper : RandomPool {
+  RandomPoolWrapper() {
+    SecByteBlock seed(32);
+    OS_GenerateRandomBlock(false, seed, seed.size());
+    IncorporateEntropy(seed, seed.size());  // method of RandomPool
+  }
+};
 
-void CryptographyWrapper::init(const char* randomSeed) {
-  string s(randomSeed);
-  if (s.length() < 16) s.resize(16, ' ');
-  sGlobalRandGen.IncorporateEntropy((CryptoPP::byte*)s.c_str(), s.length());
-
-  VERIFY(DigestUtil::digestLength() == DIGEST_SIZE);
-
-  // Initialize RELIC library for BLS threshold signatures
-  BLS::Relic::Library::Get();
-}
-
-void CryptographyWrapper::init() {
-  std::string seed = IntToString(time(NULL));
-  CryptographyWrapper::init(seed.c_str());
-}
+// RandomPoolWrapper wraps RandomPool in a struct which initialises it. The wrapper is used as static thread local
+// variable so that each thread has got its own instance. The reason for this complication is thread safety. This
+// RandomPool is used by the SigManager which is a singleton used by many threads concurrently.
+static thread_local RandomPoolWrapper sGlobalRandGen;
 
 void convert(const Integer& in, string& out) {
   out.clear();
@@ -194,8 +189,8 @@ bool ECDSAVerifier::verify(const std::string& data_to_verify, const std::string&
 ECDSAVerifier::~ECDSAVerifier() = default;
 class RSASigner::Impl {
  public:
-  Impl(BufferedTransformation& privateKey) : rand(sGlobalRandGen), priv(privateKey){};
-  Impl(RSA::PrivateKey& privateKey) : rand(sGlobalRandGen), priv(privateKey){};
+  Impl(BufferedTransformation& privateKey) : priv(privateKey){};
+  Impl(RSA::PrivateKey& privateKey) : priv(privateKey){};
 
   size_t signatureLength() const { return priv.SignatureLength(); }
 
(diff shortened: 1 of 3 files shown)


KTeofanidis (approved):
Great fix!

glevkovich (approved):
Looks good! Thanks for coming with this fix.

Comment: I think if we reach hundreds of thread count in the process, thread_local won't be an optimal solution anymore (we have that in a few more other places).
Also a question: if there was an actual data race - how come the system works? this race is here for ages and is unrelated to the recent changes in SigManager, even before there used to be pre-processor calling with multiple threads.

tdimitrov:
The declaration of RandomPool is here: https://www.cryptopp.com/docs/ref/randpool_8h_source.html
The size is 48 bytes + (smart) pointer + bool. I think it won't be an issue even with 1000 threads.

Not all data races lead to a crash, but we can rely on this. In the current case threads were writing to a buffer simultaneously and the random data generated was mangled. This can't lead to a crash but also the generated data is no longer correct.

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -58,7 +58,18 @@ namespace impl {
 //#define RSA_STANDARD PKCS1v15
 //#define RSA_STANDARD OAEP<SHA>
 
-static RandomPool sGlobalRandGen;  // not thread-safe !!
+// Wrap RandomPool in a struct which initialises it. The wrapper is used as static thread local variable so that each
+// thread has got its own instance. The reason for this complication is thread safety. This RandomPool is used by the
+// SigManager which is a singleton used by many threads concurrently.
+struct RandomPoolWrapper : RandomPool {
+  RandomPoolWrapper() {
+    static std::atomic<uint32_t> cnt = 0;
```
dartdart26, line 66: Nit: Maybe it is worth describing what `cnt` is used for and why.

dartdart26: Actually, what worries me is the following (especially the last sentence):

_[basic.start.dynamic] It is implementation-defined whether the dynamic initialization of a non-local non-inline variable with thread storage duration is sequenced before the first statement of the initial function of a thread or is deferred. If it is deferred, the initialization associated with the entity for thread t is sequenced before the first non-initialization odr-use by t of any non-inline variable with thread storage duration defined in the same translation unit as the variable to be initialized. It is implementation-defined in which threads and at which points in the program such deferred dynamic initialization occurs._

I might be misunderstanding above, but I guess `cnt` will not work as intended in cases where two or more random number generators are initialised from the same thread. Maybe we can use another random seed? You were mentioning /dev/urandom or something like that.

dartdart26 (approved):
Great find!

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -58,24 +58,27 @@ namespace impl {
...
+void CryptographyWrapper::init() {
+  SecByteBlock seed(32);
+  // this might block for a while - man 4 random for details
+  OS_GenerateRandomBlock(false, seed, seed.size());
+  sGlobalRandGen.IncorporateEntropy(seed, seed.size());
 
   VERIFY(DigestUtil::digestLength() == DIGEST_SIZE);
 
   // Initialize RELIC library for BLS threshold signatures
   BLS::Relic::Library::Get();
```
tdimitrov, line 79: I wonder if we need this here... If not the whole init() can be moved inside RandomPoolWrapper constructor.

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -58,23 +58,19 @@ namespace impl {
...
-  BLS::Relic::Library::Get();
-}
+// Wrap RandomPool in a struct which initialises it. The wrapper is used as static thread local variable so that each
+// thread has got its own instance. The reason for this complication is thread safety. This RandomPool is used by the
+// SigManager which is a singleton used by many threads concurrently.
+struct RandomPoolWrapper : RandomPool {
+  RandomPoolWrapper() {
+    SecByteBlock seed(32);
+    // this might block for a while - 'man 4 random' for details
+    OS_GenerateRandomBlock(true, seed, seed.size());
```
robem, line 68: You don't need to block here and read from `/dev/urandom` instead. All modern systems have a HWRNG.

robem (approved):

bftengine/src/bftengine/Crypto.cpp
```diff
@@ -58,23 +58,19 @@ namespace impl {
...
+    IncorporateEntropy(seed, seed.size());  // method of RandomPool
+  }
+};
 
-void CryptographyWrapper::init() {
-  std::string seed = IntToString(time(NULL));
-  CryptographyWrapper::init(seed.c_str());
-}
+static thread_local RandomPoolWrapper sGlobalRandGen;
```
robem, line 73: Nit: The comment about "why" this variable is thread_local should be right here.

bftengine/src/bftengine/Crypto.cpp
tdimitrov: Doesn't /dev/random use the same HWRNG? If yes - I think it's safer this way.

robem: Why is it safer? (It is a loaded question, no need to answer) :)
https://www.2uo.de/myths-about-urandom/#experts

robem: We can talk more offline about /dev/random vs /dev/urandom. Either is fine if blocking doesn't become an issue in this case.

tdimitrov: Great read, @robem . And thanks for the feedback. I'll use the non-blocking version (/dev/urandom).

PR #1693, opened 2021-07-15
Clientservice: YAML test

robem:
This PR introduces the first test for clientservice. For convenience and to keep tests simple, this PR refactors the existing parsing code. Please see commit messages for details.

$> TEST_NAME=clientservice-test-yaml_parsing make single-test


shgandhi (approved):
Thanks for the change and the cleanup @robem (especially the use of YAML node in the parsing function, I learned something new today)! 
I just have one minor comment, that can be addressed separately as well.

PR #1694, opened 2021-07-16
Threshsign: CMake refactoring

robem:
Understanding the internal dependencies of the threshsign library, tests, and benchmarks is not easy to grasp and the existing CMake files don't really help given the dependency circle. This PR is an attempt to clean up the obvious but there is certainly room for more improvement.
Please refer to the commit messages for details.


toly-kournik (approved):
Thanks Rob,
LGTM!

PR #1696, opened 2021-07-19
Destroy RocksDB object before cleanup in tests

dartdart26:
Make sure the RocksDB object is destroyed before removing the RocksDB
directory. This ensures RocksDB is shutdown properly.

In order to achieve it, make sure no references to the DB are kept as
test members.


f-squirrel (approved):
LGTM!
Thank you @dartdart26

PR #1700, opened 2021-07-20
ThinReplica: Introduce EventGroups

robem:
We already have EventGroups defined in `concord_client.proto`. This PR adds EventGroups to the thin replica protocol so that we can use them in subsequent changes. Note, we make sure that this PR doesn't change any behavior. From now on, all tests use "legacy" events and not EventGroups. We have to re-evaluate our tests to maybe find a generic solution.


shgandhi (approved):

PR #1703, opened 2021-07-20
TRS memory leaks in getClientIdFromClientCert

Liorgal28:
In the method getClientIdFromClientCert, we observed a memory leak, which was caused in several places:
1) Calling BIO_new without freeing the allocated memory.
2) By utilizing PEM_read_bio_X509, the memory that has been allocated is not released.
3) not releasing the memory which has been allocated with X509_NAME_oneline.

I ran address sanitizer after fixing these issues, and none of the leaks were observed.


HristoStaykov (approved):
Great catch @Liorgal28 , Thanks!

robem (approved):
Lgtm. Thanks! 💪

shgandhi (approved):
Thanks a lot for fixing and testing this @Liorgal28! LGTM!

PR #1705, opened 2021-07-20
Clientservice: Introduce legacy events

robem:
This PR introduces legacy events to `concord_client.proto`.
A later change will handle those events properly.
See commits for details.


shgandhi (approved):

PR #1706, opened 2021-07-20
Move Client Pool To Open Source

irisk29:
The project currently uses the facade which resides in the close-source.
The client pool is used by the facade and with this PR the client pool will be located within the open-source instead of the close-source.
In order to do that, I transferred the client pool directory and refactored its structure.
Then, I modified the CMake to include the client pool directory.
This PR is part of building a standalone client in the open-source.


robem (approved):

robem:
I fixed the build issues but unfortunately, I couldn't push them to your branch because it was disabled. I created a new branch which takes @irisk29 commits and adds a fix on top #1713

PR #1707, opened 2021-07-20
State transfer performance improvments put block phase1

glevkovich:
This is the 1st phase of the stage 2 series of commits for state transfer improvements.
The series of commits here concentrate on API changes + minor comments and API refactoring.
The implemented changes are done to support the 2nd phase of this stage, which will be submitted imminently after. It will include additional performance improvements. The main improvement: committing blocks concurrently.

3 new APIS:

1. IAppState::putBlockAsync
2. IAppState::getPrevDigestFromBlock (overloaded)
3. IStateTransfer::addOneShotTimer

1 API modified:
- IAppState::putBlock - added a new argument

I suggest reviewing this PR by separate commits.

### DIFF
--- a/bftengine/tests/bcstatetransfer/test_replica.hpp
+++ b/bftengine/tests/bcstatetransfer/test_replica.hpp
@@ -52,6 +52,7 @@ class TestReplica : public IReplicaForStateTransfer {
 
   void changeStateTransferTimerPeriod(uint32_t timerPeriodMilli) override{};
 
+  concordUtil::Timers::Handle addOneShotTimer(uint32_t timeoutMilli) override { return concordUtil::Timers::Handle(); }
   ///////////////////////////////////////////////////////////////////////////
   // Data - All public on purpose, so that it can be accessed by tests
   ///////////////////////////////////////////////////////////////////////////
--- a/kvbc/src/Replica.cpp
+++ b/kvbc/src/Replica.cpp
@@ -552,59 +552,58 @@ RawBlock Replica::getBlockInternal(BlockId blockId) const { return m_bcDbAdapter
  * This method assumes that *outBlock is big enough to hold block content
  * The caller is the owner of the memory
  */
-bool Replica::getBlock(uint64_t blockId, char *outBlock, uint32_t *outBlockSize) {
+bool Replica::getBlock(uint64_t blockId, char *outBlock, uint32_t outBlockMaxSize, uint32_t *outBlockActualSize) {
   if (replicaConfig_.isReadOnly) {
-    return getBlockFromObjectStore(blockId, outBlock, outBlockSize);
+    return getBlockFromObjectStore(blockId, outBlock, outBlockMaxSize, outBlockActualSize);
   }
   const auto rawBlock = m_kvBlockchain->getRawBlock(blockId);
   if (!rawBlock) {
     throw NotFoundException{"Raw block not found: " + std::to_string(blockId)};
   }
   const auto &ser = categorization::RawBlock::serialize(*rawBlock);
-  *outBlockSize = ser.size();
-  LOG_DEBUG(logger, KVLOG(blockId, *outBlockSize));
-  std::memcpy(outBlock, ser.data(), *outBlockSize);
+  if (ser.size() > outBlockMaxSize) {
+    LOG_ERROR(logger, KVLOG(ser.size(), outBlockMaxSize));
+    throw std::runtime_error("not enough space to copy block!");
+  }
+  *outBlockActualSize = ser.size();
+  LOG_DEBUG(logger, KVLOG(blockId, *outBlockActualSize));
+  std::memcpy(outBlock, ser.data(), *outBlockActualSize);
   return true;
 }
 
-std::future<bool> Replica::getBlockAsync(uint64_t blockId, char *outBlock, uint32_t *outBlockSize) {
+std::future<bool> Replica::getBlockAsync(uint64_t blockId,
+                                         char *outBlock,
+                                         uint32_t outBlockMaxSize,
+                                         uint32_t *outBlockActualSize) {
   static uint64_t callCounter = 0;
   static constexpr size_t snapshotThresh = 1000;
 
   auto future = blocksIOWorkersPool_.async(
-      [this](uint64_t blockId, char *outBlock, uint32_t *outBlockSize) {
+      [this](uint64_t blockId, char *outBlock, uint32_t outBlockMaxSize, uint32_t *outBlockActualSize) {
         bool result = false;
         auto start = std::chrono::steady_clock::now();
-        *outBlockSize = 0;
+        *outBlockActualSize = 0;
         LOG_TRACE(logger, "Job Started: " << KVLOG(blockId));
 
         // getBlock will throw exception if block doesn't exist
         try {
-          result = getBlock(blockId, outBlock, outBlockSize);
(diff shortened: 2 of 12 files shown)


bftengine/tests/bcstatetransfer/test_replica.hpp
```diff
@@ -52,6 +52,7 @@ class TestReplica : public IReplicaForStateTransfer {
 
   void changeStateTransferTimerPeriod(uint32_t timerPeriodMilli) override{};
 
+  concordUtil::Timers::Handle addOneShotTimer(uint32_t timeoutMilli) override { return concordUtil::Timers::Handle(); }
```
cloudnoize, line 55: whats is the purpose of the ```OneShotTimer``` ?

kvbc/src/Replica.cpp
```diff
@@ -539,23 +587,34 @@ std::future<bool> Replica::getBlockAsync(uint64_t blockId, char *outBlock, uint3
...
       [this](uint64_t blockId, char *outBlock, uint32_t *outBlockSize) {
+        bool result = false;
         auto start = std::chrono::steady_clock::now();
         *outBlockSize = 0;
+        LOG_TRACE(logger, "Job Started: " << KVLOG(blockId));
 
         // getBlock will throw exception if block doesn't exist
         try {
-          getBlock(blockId, outBlock, outBlockSize);
+          result = getBlock(blockId, outBlock, outBlockSize);
```
cloudnoize, line 597: isn't it worth to modify the API get block as we discussed to at least have an assertion on the sizes of the in vs out block?

bftengine/tests/bcstatetransfer/test_replica.hpp
glevkovich: The main reason is performance optimization. I'll explain via an example:  when putting block concurrently, let's say I have 4 jobs for putBlock running concurrently. None of them is ready yet. Let's assume that at moment the source is disconnected and no more ItemDataMsg arrives. Nothing will trigger the completion of these jobs, I still want to finalize these jobs ASAP. Without this timer, I will finalize them only on the next periodic timer, which is a long time from the current time.  With some logic (I am using a flag, and use this timer in rare events) I trigger a timer for 5 ms from the current time, to come back and check the future again. That way, all jobs will be finalized when switching to a different source or when (maybe) the current source is back (retransmission successful).

Again, this timer is not mandatory, in some rare cases, it will just improve performance.
Why not just wait on the jobs? since the main thread can do other things meanshile. I would like to add more optimisations such as digest computation in parallel etc. The main thread should always keep moving and only wait when it must wait, but keep moving to supervise other jobs and fetch messages from the incoming queue.

kvbc/src/Replica.cpp
glevkovich: There is a dedicated task to change this interface, and get a better idea of the maximal block size + to have the storage layer "knowing" the block size in advance. spoke about it with Gadi as well.
As for the assert, you are right, this is a good time to insert the assert (I wrote it in the other task, but will do it here).
Thanks!

kvbc/src/Replica.cpp
```diff
@@ -462,11 +465,56 @@ bool Replica::putBlock(const uint64_t blockId, const char *blockData, const uint
...
+        LOG_TRACE(logger, "Job Started: " << KVLOG(blockId, blockSize));
+        try {
+          result = putBlock(blockId, block, blockSize, trylinkSTChainFrom);
+        } catch (const std::runtime_error &ex) {
+          LOG_ERROR(logger, "runtime_error exception:" << ex.what());
+          return false;
+        } catch (const std::exception &ex) {
+          LOG_ERROR(logger, "exception:" << ex.what());
+          return false;
+        } catch (...) {
```
toly-kournik, line 494: We don't catch unknown exceptions and we are not catching exceptions in case putBlock fails, it's a reason for termination. Look at `putBlock()`

kvbc/src/Replica.cpp
```diff
@@ -539,23 +587,34 @@ std::future<bool> Replica::getBlockAsync(uint64_t blockId, char *outBlock, uint3
...
 
         // getBlock will throw exception if block doesn't exist
         try {
-          getBlock(blockId, outBlock, outBlockSize);
+          result = getBlock(blockId, outBlock, outBlockSize);
         } catch (NotFoundException &ex) {
           *outBlockSize = 0;
           LOG_ERROR(logger, "Block not found:" << KVLOG(blockId));
           return false;
+        } catch (const std::runtime_error &ex) {
```
toly-kournik, line 602: Same as above

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -663,7 +663,7 @@ ImmutableOutput KeyValueBlockchain::handleCategoryUpdates(BlockId block_id,
 }
 
 /////////////////////// state transfer blockchain ///////////////////////
-void KeyValueBlockchain::addRawBlock(const RawBlock& block, const BlockId& block_id) {
+void KeyValueBlockchain::addRawBlock(const RawBlock& block, const BlockId& block_id, bool trylinkSTChainFrom) {
```
toly-kournik, line 666: why `try` and not just `link`? and how is it used?

kvbc/src/Replica.cpp
glevkovich: @cloudnoize , In the next push I'm going to add another commit adding another argument to getBlock and getBlockAsync to address the above comment.

kvbc/src/Replica.cpp
glevkovich: OK, this is redundant. I'm actually catching an exception at the caller as well. so this is redundant for now..
Will remove..

kvbc/src/Replica.cpp
glevkovich: I'm going to catch only the case of NotFoundException and return false. This is laterally the meaning of the bool we return in the future. Else, we have no use for the bool returned in the future, we can remove it.
I will remove all other exceptions catch.

kvbc/src/categorization/kv_blockchain.cpp
glevkovich: Link is the default case, and not optimized.
Let's say we put 50,000 blocks.
code always try to link. This is costly: it calls linkSTChainFrom, which calls getRawBlock and more operations. I can skip all that easily by calling 49,999 times with false (do not try to link), and only on the last time with true (link). when putting the last block. Its a very easy change and it saves some unneeded operations.

cloudnoize (approved):
Great work, thanks

toly-kournik:
I think passing `trylinkSTChainFrom` to a putBlock reveals the storage implementation details and shouldn't be a part of an interface. It could be implemented (and it actually was in a previous version of kvbc storage) in a different way.

glevkovich:
`I think passing trylinkSTChainFrom to a putBlock reveals the storage implementation details and shouldn't be a part of an interface. It could be implemented (and it actually was in a previous version of kvbc storage) in a different way.`

For sure, but it also optimizes things? So there is a tradeoff. Also, I set a default value to behave like always, so the caller doesn't have to use it at all.
Suggestion:
I can just change the name to "bool lastBlock" and it will not reveal any internal implementation, but only "give a hint" - when true - we are putting the last block. Storage can do whatever it wants with the flag, it's like a recommendation.

toly-kournik:
ok

toly-kournik (approved):
LGTM!

PR #1712, opened 2021-07-21
TS: Provide timestamp to Reconfiguration handlers and save it in blocks

f-squirrel:
Reason:
* The time will be used by the time-based pruning and other
reconfiguration handlers in the future.
* In certain cases, replicas replay reconfiguration commands saved in the
blockchain. For the time-based pruning, the system has to replay the
command together with the time when the command was submitted.  In order
to achieve it, reconfiguration requests will be saved with the timestamp
(as CMF message).

Changes:
* Updated Reconfiguration-related interfaces
* Added a new CMF-message
* Added saving and retrieving the timestamp from the reconf block
* Refactored persistReconfigurationBlock

### DIFF
--- a/kvbc/src/reconfiguration_kvbc_handler.cpp
+++ b/kvbc/src/reconfiguration_kvbc_handler.cpp
@@ -14,20 +14,21 @@
 #include "ControlStateManager.hpp"
 #include "bftengine/EpochManager.hpp"
 #include "bftengine/ReconfigurationCmd.hpp"
+#include "concord.cmf.hpp"
 #include "endianness.hpp"
 #include "kvbc_app_filter/kvbc_key_types.h"
 
 namespace concord::kvbc::reconfiguration {
 
-kvbc::BlockId ReconfigurationBlockTools::persistReconfigurationBlock(const std::vector<uint8_t>& data,
-                                                                     const uint64_t bft_seq_num,
-                                                                     string key,
-                                                                     bool include_wedge) {
+kvbc::BlockId ReconfigurationBlockTools::persistReconfigurationBlock(
+    const std::vector<uint8_t>& data,
+    const uint64_t bft_seq_num,
+    const std::optional<bftEngine::Timestamp>& timestamp,
+    string key,
+    bool include_wedge) {
   concord::kvbc::categorization::VersionedUpdates ver_updates;
   ver_updates.addUpdate(std::move(key), std::string(data.begin(), data.end()));
 
-  // All blocks are expected to have the BFT sequence number as a key.
-  ver_updates.addUpdate(std::string{kvbc::keyTypes::bft_seq_num_key}, block_metadata_.serialize(bft_seq_num));
   uint64_t epoch = 0;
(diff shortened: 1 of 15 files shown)


kvbc/src/st_reconfiguration_sm.cpp
```diff
@@ -78,15 +82,25 @@ bool StReconfigurationHandler::handlerStoredCommand(const std::string &key, uint
...
     deserializeCmfMessage(cmd, strval);
-    return handle(cmd, seqNum, current_cp_num, blockid);
+    std::optional<bftEngine::Timestamp> timestamp = std::nullopt;
+    auto value = ro_storage_.get(kvbc::kConcordInternalCategoryId, std::string{kvbc::keyTypes::timestamp_key}, blockid);
+    if (value) {
+      const auto &data = std::get<categorization::VersionedValue>(*value).data;
+      concord::messages::Timestamp cmf_ts;
+      deserializeCmfMessage(cmf_ts, data);
+      timestamp.emplace(
+          bftEngine::Timestamp{bftEngine::ConsensusTime{cmf_ts.time_since_epoch}, cmf_ts.request_position});
```
dartdart26, line 92: Nit: Can remove the `bftEngine::Timestamp` part as emplace() takes the arguments to call the constructor itself.

kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -34,20 +36,34 @@ kvbc::BlockId ReconfigurationBlockTools::persistReconfigurationBlock(const std::
...
+  if (timestamp) {
+    concord::messages::Timestamp cmf_ts;
+    cmf_ts.time_since_epoch = timestamp->time_since_epoch.count();
+    cmf_ts.request_position = timestamp->request_position;
+    std::vector<uint8_t> serialized_ts;
+    concord::messages::serialize(serialized_ts, cmf_ts);
+    ver_updates.addUpdate(std::string{keyTypes::timestamp_key},
+                          std::string(serialized_ts.begin(), serialized_ts.end()));
   }
+  return persistReconfigurationBlock(ver_updates, bft_seq_num, timestamp);
```
dartdart26, line 49: I think we are going to serialize and add the timestamp twice - in this function and when calling `persistReconfigurationBlock(ver_updates, bft_seq_num, timestamp);`. Is that intentional?

f-squirrel: removed, now it is serialised only once

kvbc/src/st_reconfiguration_sm.cpp
f-squirrel: It fails, probably because `Timestamp` I have not defined a ctor. If it is not critical, I'd leave it.

kvbc/src/reconfiguration_kvbc_handler.cpp
f-squirrel: removed, now it is serialised only once

dartdart26 (approved):
Looks good, thanks!

f-squirrel:
**Please do not merge**

PR #1713, opened 2021-07-21
Client pool open source

robem:
This change introduces the Client Pool from closed source. The first two commits are taken from #1706 and in addition, we added clang-tidy fixes on top.


shgandhi (approved):
Thanks for the change @irisk29 and @robem! LGTM.

PR #1715, opened 2021-07-22
State transfer performance improvments put block phase2

glevkovich:
This PR includes one main improvement - putting blocks concurrently. This improves the channel throughput (MB/s) by approx 25%. It was tested also to work correctly with block chunking.
Changes rely on the interface changes done in phase1.
In addition, I had to add an IO context pool and refactor getBlock as well (using the same contexts) to simplify the code.
PR also includes a small bug fix and refactoring.
I suggest reviewing by commits to get a better understanding.

Commits:
1. BCStateTran: intro BlockIODataPool, refactor getBlock at source, fix bug.
2. BCStateTran: perform putBlock to temporary ST concurrently.
3. BCStateTran: minor refactoring (reword comments, remove unneeded code).

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -411,13 +411,13 @@ void BCStateTran::stopRunning() {
   summariesCerts.clear();
   numOfSummariesFromOtherReplicas.clear();
   sourceSelector_.reset();
-
   nextRequiredBlock_ = 0;
   metrics_.next_required_block_.Get().Set(0);
   digestOfNextRequiredBlock.makeZero();
 
-  for (auto i : pendingItemDataMsgs) replicaForStateTransfer_->freeStateTransferMsg(reinterpret_cast<char *>(i));
-
+  for (auto i : pendingItemDataMsgs) {
+    replicaForStateTransfer_->freeStateTransferMsg(reinterpret_cast<char *>(i));
+  }
   pendingItemDataMsgs.clear();
   for (auto &ctx : ioContexts_) ioPool_.free(ctx);
   ioContexts_.clear();
@@ -649,9 +649,6 @@ void BCStateTran::startCollectingStats() {
   gettingCheckpointSummariesDT_.reset();
   gettingMissingResPagesDT_.reset();
   cycleDT_.reset();
-  betweenPutBlocksStTempDT_.reset();
-  putBlocksStTempDT_.reset();
-
   sources_.clear();
 
   metrics_.overall_blocks_collected_.Get().Set(0ull);
@@ -706,7 +703,6 @@ void BCStateTran::onTimerImp() {
     last_metrics_dump_time_ = currTimeForDumping;
     LOG_INFO(getLogger(), "--BCStateTransfer metrics dump--" + metrics_component_.ToJson());
   }
-  auto currTime = getMonotonicTimeMilli();
 
   // take a snapshot and log after time passed is approx x2 of fetchRetransmissionTimeoutMs
   if (sourceFlag_ &&
@@ -719,7 +715,10 @@ void BCStateTran::onTimerImp() {
     sourceFlag_ = false;
     sourceSnapshotCounter_ = 0;
   }
+
+  // Retransmit AskForCheckpointSummariesMsg if needed
   if (fs == FetchingState::GettingCheckpointSummaries) {
+    auto currTime = getMonotonicTimeMilli();
     if ((currTime - lastTimeSentAskForCheckpointSummariesMsg) > config_.checkpointSummariesRetransmissionTimeoutMs) {
       LOG_DEBUG(getLogger(),
                 KVLOG(retransmissionNumberOfAskForCheckpointSummariesMsg, kResetCount_AskForCheckpointSummaries));
@@ -728,6 +727,7 @@ void BCStateTran::onTimerImp() {
 
       sendAskForCheckpointSummariesMsg();
     }
+    // process data if fetching
   } else if (fs == FetchingState::GettingMissingBlocks || fs == FetchingState::GettingMissingResPages) {
     processData();
   }
@@ -1495,9 +1495,6 @@ bool BCStateTran::onMessage(const FetchBlocksMsg *m, uint32_t msgLen, uint16_t r
       } catch (const std::exception &ex) {
         LOG_FATAL(getLogger(), "exception:" << ex.what());
         ConcordAssert(false);
-      } catch (...) {
-        LOG_FATAL(getLogger(), "Unknown exception!");
-        ConcordAssert(false);
       }
       ConcordAssertGT(ctx->actualBlockSize, 0);
       ConcordAssertEQ(ctx->blockId, nextBlockId);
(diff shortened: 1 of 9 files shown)


bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -357,21 +357,80 @@ class BCStateTran : public IStateTransfer {
...
+      return (fetchingState_ == FetchingState::NotFetching) ? ST_SRC_LOG : ST_DST_LOG;
+    }
+    const size_t maxNumElements_;
+    std::deque<BlockIOContextPtr> freeQ_;
+    std::deque<BlockIOContextPtr> allocatedQ_;
+    FetchingState& fetchingState_;
+  };
+
+  BlockIODataPool ioPool_;
+  std::deque<BlockIOContextPtr> ioContexes_;
```
toly-kournik, line 434: ioContexts_

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -357,21 +357,80 @@ class BCStateTran : public IStateTransfer {
...
+
+    size_t numFreeElements() const { return freeQ_.size(); }
+    size_t numAllocatedElements() const { return allocatedQ_.size(); }
+    bool empty() { return freeQ_.empty(); };
+    bool full() { return allocatedQ_.empty(); };
+    size_t maxElements() { return maxNumElements_; };
+
+    BlockIOContextPtr alloc() {
+      if (freeQ_.empty()) {
+        throw std::runtime_error("No more free elements!");
```
toly-kournik, line 390: what happens in this case? do we want to fail the entire replica?

glevkovich: This should never happen, it means I have a bug in my code. The caller should be responsible to make sure there are always enough elements. It is doing so by calling processCommitResultsAsync with the right flags.
if there are no more free elements in the pool, it should call processCommitResultsAsync with breakIfNotReady=false.
In practice, it will never happen anyway. We have no more than 5-6 concurrent at peak, while the batch size (used also in the source) is much larger.

bftengine/src/bcstatetransfer/BCStateTran.hpp
glevkovich: Thanks, I'll fix that.

toly-kournik:
I think the memory pool should be implemented as a template general utility.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2300,6 +2308,55 @@ std::string BCStateTran::logsForCollectingStatus(const uint64_t firstRequiredBlo
...
+      }
+      break;
+    }
+    ConcordAssertEQ(ctx->blockId, nextCommittedBlockId_);
+    try {
+      ConcordAssertEQ(ctx->future.get(), true);
+    } catch (const std::exception &e) {
+      LOG_FATAL(getLogger(), e.what());
+      ConcordAssert(false);
+    } catch (...) {
```
toly-kournik, line 2349: Please don't catch unknown exceptions.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2212,7 +2218,9 @@ void BCStateTran::EnterGettingCheckpointSummariesState() {
   sourceSelector_.reset();
   metrics_.current_source_replica_.Get().Set(sourceSelector_.currentReplica());
 
+  processCommitResultsAsync(false, false);
```
toly-kournik, line 2225: why is it needed here?

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2415,55 +2481,77 @@ void BCStateTran::processData() {
...
+          "Before putBlock id " << nextRequiredBlock_ << ":" << std::boolalpha << KVLOG(lastBlock, actualBlockSize));
       if (!lastBlock) {
         putBlocksStTempDT_.pause();
         betweenPutBlocksStTempDT_.start();
-        as_->getPrevDigestFromBlock(nextRequiredBlock_,
-                                    reinterpret_cast<StateTransferDigest *>(&digestOfNextRequiredBlock));
-        nextRequiredBlock_--;
-        g.txn()->setLastRequiredBlock(nextRequiredBlock_);
+        //////////////////////////////////////////////////////////////////////////
+        // Not the last block - block are committed into ST temporary blockchain
```
toly-kournik, line 2505: ST doesn't know anything about temporary blockchain - it's a storage implementation details

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2415,55 +2481,77 @@ void BCStateTran::processData() {
...
           dst_time_between_sendFetchBlocksMsg_rec_.end();
           LOG_DEBUG(getLogger(), "Sending FetchBlocksMsg: lastInBatch is true");
           sendFetchBlocksMsg(firstRequiredBlock, nextRequiredBlock_, 0);
           break;
         }
       } else {
-        // this is the last block we need
-        // report collecting status (without vblock) into log
+        //////////////////////////////////////////////////////////////////////////
+        // This is the last block we need. After committed to ST temporary blockchain, all blocks are deleted from ST
```
toly-kournik, line 2536: Same

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -425,12 +427,21 @@ class BCStateTran : public IStateTransfer {
...
+  bool oneShotTimerFlag_;  // used to control the trigger of oneShotTimer self requests
 
   // returns number of jobs pushed to queue
   uint16_t getBlocksConcurrentAsync(uint64_t nextBlockId, uint64_t firstRequiredBlock, uint16_t numBlocks);
 
+  // lastBlock: is true if we put the oldest block (firstRequiredBlock)
+  // breakIfNotReady: is true if caller would like to exit if the next future is not ready (job not ended yet).
+  //    In that case, a ONESHOT timer is invoked to check the future again soon.
+  // return: true if done procesing all futures, and false if the front one was not std::future_status::ready
+  bool processCommitResultsAsync(bool lastBlock, bool breakIfNotReady);
```
toly-kournik, line 444: breakIfNotReady is not a clear parameter and complicates both implementation and understanding

bftengine/src/bcstatetransfer/BCStateTran.cpp
glevkovich: OK, I'll remove some part of the comment in 2 places.

bftengine/src/bcstatetransfer/BCStateTran.hpp
glevkovich: I will change the name..
The whole idea of putting blocks concurrently relay on this flag.
If I don't break immediately, I'll wait for each block and blocks won't be put concurrently.
If you mean that the name should be renamed - I can change the name, but the flag is a must for the implementation of concurrent putBlock.
Many times, the future for the job we just invoked will not be ready, the main thread will break out and go to the handoff queue to fetch the next ItemDataMsag.

glevkovich: new name:
waitForAllJobsToComplete
This has the opposite meaning: when true - cannot break out the loop, must wait for each future in the queue to complete.

glevkovich:
@toly-kournik 
> In the second commit ""BCStateTran: perform putBlock to temporary ST concurrently there are no changes to related to a putBlock(). There's only a `BCStateTran::processCommitResultsAsync()` which name is misleading

But processCommitResultsAsync is not the only thing we use, the change is that we are now using the interface call putBlockAsync:
` ctx->future = as_->putBlockAsync(nextRequiredBlock_, ctx->blockData.get(), ctx->actualBlockSize, false);`
So what you've said is right for the finalize part, but the actual putBlock has changed as well - we are now putting blocks concurrently AND calling processCommitResultsAsync to finalize jobs (if possible). So I don't really understand what is wrong with the comment.
In short: put block is now concurrent by calling putBlockAsync and processCommitResultsAsync

toly-kournik (approved):
LGTM!
thanks Gil

PR #1716, opened 2021-07-22
Client Key Exchange phase 2

toly-kournik:
- store client public key in ClientsManager reserved pages;
- SigManager will first load bootstrap client public keys as now, and then will be updated by the keys loaded from reserved pages;
- few small refactorings: smart_ptrs instead of plain ptr, logs, etc.

### DIFF
--- a/bftengine/include/bftengine/KeyExchangeManager.hpp
+++ b/bftengine/include/bftengine/KeyExchangeManager.hpp
@@ -55,7 +55,10 @@ class KeyExchangeManager {
   // Publish the public keys of the clients
   void sendInitialClientsKeys(const std::string&);
   void onPublishClientsKeys(const std::string& keys, std::optional<std::string> bootstrap_keys);
-  void onClientPublicKeyExchange(const std::string& key, std::uint16_t clientId);
+  // called on a new client key
+  void onClientPublicKeyExchange(const std::string& key, KeyFormat, NodeIdType clientId);
+  // called when client keys are loaded
+  void loadClientPublicKey(const std::string& key, KeyFormat, NodeIdType clientId);
   ///////// end - Clients public keys interface///////////////
 
   std::string getStatus() const;
@@ -151,6 +154,7 @@ class KeyExchangeManager {
     IMultiSigKeyGenerator* kg{nullptr};
     IKeyExchanger* ke{nullptr};
     std::shared_ptr<concord::secretsmanager::ISecretsManagerImpl> secretsMgr;
+    IClientPublicKeyStore* cpks;
     concordUtil::Timers* timers{nullptr};
   };
 
@@ -192,6 +196,7 @@ class KeyExchangeManager {
   std::shared_ptr<IInternalBFTClient> client_;
   std::vector<IKeyExchanger*> registryToExchange_;
   IMultiSigKeyGenerator* multiSigKeyHdlr_{nullptr};
+  IClientPublicKeyStore* clientPublicKeyStore_{nullptr};
 
   struct Metrics {
     std::chrono::seconds lastMetricsDumpTime;
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -15,6 +15,8 @@
 #include "assertUtils.hpp"
 #include "Logger.hpp"
 #include "ReplicaConfig.hpp"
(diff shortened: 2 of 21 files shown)


bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -50,32 +49,37 @@ ClientsManager::ClientsManager(const std::set<NodeIdType>& proxyClients,
...
+// * remove pending request if loaded reply is newer
 void ClientsManager::loadInfoFromReservedPages() {
   for (auto const& clientId : clientIds_) {
-    const uint32_t firstPageId = getFirstPageId(clientId);
+    if (loadReservedPage(getKeyPageId(clientId), sizeOfReservedPage(), scratchPage_.data())) {
+      auto& info = clientsInfo_[clientId];
+      std::istringstream iss(scratchPage_);
+      concord::serialize::Serializable::deserialize(iss, info.pubKey);
+      ConcordAssertGT(info.pubKey.first.length(), 0);
+      KeyExchangeManager::instance().loadClientPublicKey(info.pubKey.first, info.pubKey.second, clientId);
```
cloudnoize, line 77: What happens on the first start? 
The sigmanager initializes ```clientsPublicKeys_```   and the verifiers, from files and also this method calls ```SigManager::instance()->setClientPublicKey```

toly-kournik: At the first start this code won't execute since `loadReservedPage()` will return false

cloudnoize: and at consecutive starts, there is a dependency in the order i.e. first sigmanager then the ```loadClientPublicKey```, is it guaranteed to occur correctly?

toly-kournik: `SigManager::init()` is called from `ReplicaImp::ReplicaImp()` and `ClientsManager::loadReservedPages()` first time is called from `ReplicaImp::start()`

bftengine/include/bftengine/KeyExchangeManager.hpp
```diff
@@ -192,6 +196,7 @@ class KeyExchangeManager {
   std::shared_ptr<IInternalBFTClient> client_;
   std::vector<IKeyExchanger*> registryToExchange_;
   IMultiSigKeyGenerator* multiSigKeyHdlr_{nullptr};
+  IClientPublicKeyStore* clientPublicKeyStore_{nullptr};
```
cloudnoize, line 199: Doesn't it contradict your shared_ptr philosophy :)

toly-kournik: It does, but I followed your pattern here :-)

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -230,11 +230,16 @@ void SigManager::sign(const char* data, size_t dataLength, char* outSig, uint16_
 uint16_t SigManager::getMySigLength() const { return (uint16_t)mySigner_->signatureLength(); }
 
 void SigManager::setClientPublicKey(const std::string& key, PrincipalId id, KeyFormat format) {
+  LOG_INFO(KEY_EX_LOG, "client: " << id << " key: " << key << " format: " << (uint16_t)format);
   if (replicasInfo_.isIdOfExternalClient(id)) {
-    std::unique_lock lock(mutex_);
-    verifiers_.insert_or_assign(id, std::make_shared<RSAVerifier>(key.c_str(), format));
+    try {
+      std::unique_lock lock(mutex_);
```
cloudnoize, line 236: in which thread ```setClientPublicKey``` is being called and with which threads does it have a race i.e. why do we need to lock?
and if we have a race doesn't it mean we need to lock in other places that use the ``` verifiers_``` e.g 
 ```SigManager::verifySig```

bftengine/include/bftengine/KeyExchangeManager.hpp
cloudnoize: but you convinced me to adopt yours :)

bftengine/src/bftengine/SigManager.cpp
toly-kournik: This was @glevkovich input in the previous PR.
In case of a single thread when no contention occurs it's harmless.

cloudnoize (approved):
Looks great, thanks @toly-kournik

PR #1717, opened 2021-07-22
State transfer throughput window size configuration option

mdarade:
Configuration option to control logging interval of
state transfer's throughput details.

### DIFF
--- a/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
+++ b/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
@@ -154,6 +154,7 @@ struct Config {
   uint32_t maxPendingDataFromSourceReplica = 0;  // Maximal internal buffer size for all ST data, bytes
   uint32_t maxNumOfReservedPages = 0;
   uint32_t sizeOfReservedPage = 0;  // bytes
+  uint32_t gettingMissingBlocksSummaryWindowSize = 600;
 
   // timeouts
   uint32_t refreshTimerMs = 0;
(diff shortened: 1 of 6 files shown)


bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
```diff
@@ -167,6 +167,7 @@ struct Config {
   bool runInSeparateThread = false;
   bool enableReservedPages = true;
   bool enableSourceBlocksPreFetch = true;
+  uint32_t getMissingBlocksSummaryWindowSize = 600;
```
glevkovich, line 170: For the above:
The name I chose a long time ago is not accurate.
Can we change it to gettingMissingBlocksSummaryWindowSize?
nit: consider moving this under // sizes

For this whole file:
To be able to see the parameter printed, add it to operator<<() function below, you missed it.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -268,6 +268,8 @@ BCStateTran::BCStateTran(const Config &config, IAppState *const stateApi, DataSt
   // callbacks.
   addOnTransferringCompleteCallback(
       [this](uint64_t checkpoint_num) { replicaForStateTransfer_->onTransferringComplete(checkpoint_num); });
+
+  LOG_INFO(getLogger(), "ST throughput measurement " << KVLOG(config_.getMissingBlocksSummaryWindowSize));
```
glevkovich, line 272: This is not needed, we already print the whole config by implementing config operator<<().
I think it should be removed.

kvbc/src/Replica.cpp
```diff
@@ -398,6 +399,11 @@ Replica::Replica(ICommunication *comm,
   }
 #endif
 
+  if (stConfig.getMissingBlocksSummaryWindowSize > 0 and stConfig.getMissingBlocksSummaryWindowSize < 99) {
```
glevkovich, line 402: < 100
or <= 99

glevkovich (changes requested):
@mdarade, few small comments in the body.
Also, you forgot to modify bcstatetransfer_tests.cpp
See line (approx) 39:
`Config TestConfig() {`

Need to add the new parameter.

f-squirrel (approved):
LGTM with comments from @glevkovich .

glevkovich (approved):
LGTM. thanks!

PR #1719, opened 2021-07-23
Sync MT access of TicksGen and Pruning res pages

dartdart26:
Make sure that users can safely call public methods of the
TicksGenerator and pruning::ReservedPagesClient classes. Rationale is
that these objects are:
 * constructed in the main thread
 * accessed/modified in the messaging and state transfer threads

Mutexes are used to make sure changes by one thread are visible to other
threads, even though these changes are never done in parallel.


tdimitrov (approved):

HristoStaykov (approved):
LGTM, thanks @dartdart26 !

PR #1720, opened 2021-07-23
code changes done for making self.tracker optional in SimpleKVBCProtocol methods and optimized the code

bandatarunkumar:
code changes done for making self.tracker optional in SimpleKVBCProtocol methods and optimized the code

Tested the changes by running all apollo tests

### DIFF
--- a/tests/apollo/test_skvbc_network_partitioning.py
+++ b/tests/apollo/test_skvbc_network_partitioning.py
@@ -56,7 +56,6 @@ class SkvbcNetworkPartitioningTest(unittest.TestCase):
          """
         num_ops = 500
 
-        self.skvbc = kvbc.SimpleKVBCProtocol(bft_network,tracker)
         self.bft_network = bft_network
         bft_network.start_all_replicas()
         with net.PacketDroppingAdversary(
@@ -64,7 +63,7 @@ class SkvbcNetworkPartitioningTest(unittest.TestCase):
 
             adversary.interfere()
 
-            await self.skvbc.run_concurrent_ops(num_ops)
+            await tracker.skvbc.run_concurrent_ops(num_ops)
 
     from os import environ
     @unittest.skipIf(environ.get('BUILD_COMM_TCP_TLS', "").lower() == "true", "Unstable on CI (TCP/TLS only)")
@@ -84,7 +83,6 @@ class SkvbcNetworkPartitioningTest(unittest.TestCase):
         5) Perform a "read-your-writes" check in the new view
         """
         bft_network.start_all_replicas()
-        skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
         with net.PrimaryIsolatingAdversary(bft_network) as adversary:
             initial_primary = 0
             await bft_network.wait_for_view(
@@ -97,7 +95,7 @@ class SkvbcNetworkPartitioningTest(unittest.TestCase):
             await adversary.interfere()
             expected_next_primary = 1
 
-            await self._send_random_writes(skvbc)
+            await self._send_random_writes(tracker.skvbc)
 
             await bft_network.wait_for_view(
                 replica_id=random.choice(bft_network.all_replicas(without={0})),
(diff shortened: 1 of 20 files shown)


tests/apollo/test_skvbc_network_partitioning.py
```diff
@@ -97,17 +95,17 @@ async def test_single_vc_primary_isolated(self, bft_network, tracker):
             await adversary.interfere()
             expected_next_primary = 1
 
-            await self._send_random_writes(skvbc)
+            await self._send_random_writes(tracker.skvbc)
```
teoparvanov, line 98: I don't understand this change. Why do you need to use the `skvbc` object from within the `tracker`?

teoparvanov (changes requested):

tests/apollo/test_skvbc_network_partitioning.py
teoparvanov: If you want to track, you just create an `skvbc` object with a `tracker` inside. If not, you create an `skvbc` object without a `tracker`. It should be that simple.

CC: @cjn @KTeofanidis

teoparvanov: You do understand how absurd it sounds to have a `tracker` with a reference to `skvbc` which in turn has a reference to the `tracker`, right?

bandatarunkumar: Yes True Teo.

teoparvanov: I'd suggest to create the `skvbc` object inside each test, passing the `tracker` as a parameter:
```python
skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
```
The decorator's role is to create the `tracker`, and then you pass it on to the constructor of `skvbc`.

bandatarunkumar: First case, If tracker is passed, one extra skvbc instance is created in tracker, it is not used much.

Second case, In case of no tracker, we create create skvbc. This is the current behavior.

With the current changes, to solve the duplicate skvbc, i cross referenced tracker and skvbc. This looks absurd. The second case is kept as it is.

If creation of one extra skvbc instance is fine in tracker, then I can revert this logic to old.

teoparvanov: Yes, I think it's fine to have the extra skvbc instance inside the tracker. Let's revert this particular change. The rest looks fine at a high level, I'd suggest @cjn to take a more detailed look next week. I'll try to find some time for it too.

Thanks Tarun!

bandatarunkumar: Sure, Thanks Teo. I will revert this particular change.

ananpal (approved):
LGTM !

tests/apollo/util/skvbc.py
```diff
@@ -493,90 +472,113 @@ async def run_concurrent_ops(self, num_ops, write_weight=.70):
...
+            max_read_set_size = max_set_size
+            if self.tracker is not None:
+                max_read_set_size = 0 if self.tracker.no_conflicts else max_set_size
+            readset = self.readset(0, max_read_set_size)
+            writeset = self.writeset(max_set_size)
             msg_batch.append(self.write_req(readset, writeset, read_version, long_exec))
             seq_num = client.req_seq_num.next()
             batch_seq_nums.append(seq_num)
-            self.tracker.send_write(client_id, seq_num, readset, dict(writeset), read_version)
+            if self.tracker is not None:
```
cjn, line 558: Multiple checks for self.tracker can be avoided as the whole code block process holds valid only if tracker is enabled.

bandatarunkumar: Thanks Celine, will do that

cjn (approved):
LGTM. Thanks!

teoparvanov (approved):
LGTM, thanks Tarun!

PR #1722, opened 2021-07-26
Configure Client Pool From ConcordClient

irisk29:
In this PR, the client pool will be configured from the concordclient struct.
I have added all the necessary parameters for the client pool to the concordclient structs.
A private field of client pool was added to concordclient, it will be used to send requests from concordclient to the pool.


client/concordclient/src/concord_client.cpp
```diff
@@ -31,6 +35,75 @@ ConcordClient::ConcordClient(const ConcordClientConfig& config)
   auto trc_config = std::make_unique<ThinReplicaClientConfig>(
       config_.subscribe_config.id, trc_queue_, config_.topology.f_val, std::move(trs_connections));
   trc_ = std::make_unique<ThinReplicaClient>(std::move(trc_config), metrics_);
+  ConcordClientPoolConfig client_pool_config = std::move(createClientPoolStruct(config));
```
robem, line 38: `auto client_pool_config = createClientPoolConfig(config)` (see my comment below)

client/concordclient/src/concord_client.cpp
```diff
@@ -31,6 +35,75 @@ ConcordClient::ConcordClient(const ConcordClientConfig& config)
...
+  client_pool_ = std::make_unique<concord::concord_client_pool::ConcordClientPool>(client_pool_config, metrics_);
+  if (client_pool_->HealthStatus() == concord::concord_client_pool::PoolStatus::NotServing) {
+    LOG_ERROR(KEY_EX_LOG, "Client pool health status is Not Serving");
+    throw concord::concord_client_pool::InternalError();
+  }
+}
+
+ConcordClientPoolConfig& ConcordClient::createClientPoolStruct(const ConcordClientConfig& config) {
+  ConcordClientPoolConfig* client_pool_config = new ConcordClientPoolConfig;
```
robem, line 47: No need to deal with the heap here. We can simply create the object on the stack and return it by value.
```
ConcordClientPoolConfig ConcordClient::createClientPoolConfig(const ConcordClientConfig& config) {
  ConcordClientPoolConfig client_pool_config;
```

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -91,6 +105,7 @@ struct ConcordClientConfig {
   TransportConfig transport;
   // BFT client descriptors
   std::vector<BftClientInfo> bft_clients;
+  std::uint16_t num_of_used_bft_clients;
```
robem, line 108: Isn't this `bft_clients.size()`? Why do we need an extra parameter?

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -48,17 +49,30 @@ struct BftTopology {
...
   bft::client::ClientId id;
+  uint16_t port;
+  std::string host;
 };
 
 struct TransportConfig {
   enum CommunicationType { Invalid, TlsTcp, PlainUdp };
   CommunicationType comm_type;
-
+  bool enable_mock_comm;
```
robem, line 75: Please add a comment that this is for testing purposes only. Ideally, we construct our classes so that production code doesn't have to deal with test code.

robem (changes requested):

client/concordclient/src/concord_client.cpp
irisk29: changed it, thanks.

client/concordclient/src/concord_client.cpp
irisk29: fixed.

client/concordclient/include/client/concordclient/concord_client.hpp
irisk29: done.

client/concordclient/include/client/concordclient/concord_client.hpp
irisk29: In a configuration file, we define several external clients (i.e. bft_clients) but we can actually use only some of them in the pool. This is what num_of_used_bft_clients represents and because of that, I added it as a parameter.

client/concordclient/src/concord_client.cpp
```diff
@@ -31,6 +35,75 @@ ConcordClient::ConcordClient(const ConcordClientConfig& config)
...
+  }
+
+  ParticipantNode client_pool_pn;
+  id = 0;
+  for (const auto& bft_client : config_.bft_clients) {
+    ExternalClient client_pool_ec;
+    auto external_client_id = bft_client.id.val;
+    client_pool_ec.principal_id = external_client_id;
+    client_pool_ec.client_port = bft_client.port;
+    client_pool_pn.externalClients[id] = client_pool_ec;
```
robem, line 66: There is not documentation in the header for this field but do you know if the key `id` is relevant? Is it the bft client id? If it isn't then I'm wondering why `externalClients` isn't simply a `std::vector` 🤔

irisk29: As far as I understood, the id in the `externalClients` map is for mapping between the serialized number to principal id because we don't know the principal id, hence we are creating an id of our own.

client/concordclient/include/client/concordclient/concord_client.hpp
robem: Ahh, that makes sense. Thanks! With this context I can actually read it from the variable name :)

client/concordclient/src/concord_client.cpp
robem: So, if we are creating our own then why don't we use `external_client_id`/`bft_client.id.val` instead?

robem (approved):

client/concordclient/src/concord_client.cpp
irisk29: Because we got an order inside the replicas, the IDs of the clients must be greater than replica ids + RO replicas + internal clients.

PR #1723, opened 2021-07-26
Legacy events glue

robem:
This PR contains glue code to support legacy event groups in the clientservice as well as concord client. An end-to-end test hasn't been done yet because we need TRC support which will come in a later change. Please see commit messages for details.


client/concordclient/src/concord_client.cpp
```diff
@@ -85,18 +92,31 @@ void ConcordClient::subscribe(const SubscribeRequest& request,
...
-        eg.events.push_back({e.second.begin(), e.second.end()});
-      }
-      std::chrono::duration time_now = std::chrono::system_clock::now().time_since_epoch();
-      eg.record_time = std::chrono::duration_cast<std::chrono::microseconds>(time_now);
-      eg.trace_context = {};
+      // TODO: Distinguish between events and event groups. Depends on TRC API.
+      bool is_event_group = false;
+      if (is_event_group) {
+        EventGroup eg;
+        eg.id = update->block_id;
```
shgandhi, line 99: Maybe I am missing something, how do we have a 1:1 correlation between event_group_id and block_id?

shgandhi (approved):
I just have one comment.

client/concordclient/src/concord_client.cpp
robem: There no correlation. This code is needs to be reworked once the TRC change is in. Good catch but it is example code from my point of view. I didn't bother making sense of it because it has to change anyways. Sorry about that.

PR #1727, opened 2021-07-27
TRC: Event groups

robem:
This change adds event group support to the TRC. The major focus is on backwards-compatibility, meaning, make sure that nothing breaks. Once the corresponding TRS changes are in, we will do end-to-end testing.


shgandhi (approved):
Thanks for the change @robem! LGTM! 🚀

PR #1730, opened 2021-07-28
Added wedge status and restart flag in the response

nkumar04:
Enhanced addremove status response by adding wedge status and restart flag


reconfiguration/include/reconfiguration/reconfiguration_handler.hpp
```diff
@@ -43,6 +43,10 @@ class ReconfigurationHandler : public BftReconfigurationHandler {
   bool handle(const concord::messages::AddRemoveWithWedgeCommand&,
               uint64_t,
               concord::messages::ReconfigurationResponse&) override;
+  bool handle(const concord::messages::AddRemoveWithWedgeStatus&,
```
yontyon, line 46: Why not have the whole handler in kvbc?

reconfiguration/cmf/concord.cmf
```diff
@@ -109,10 +109,13 @@ Msg AddRemoveWithWedgeCommand 29 {
 
 Msg AddRemoveWithWedgeStatus 30 {
     uint64 sender_id
+    bool bft_flag
```
yontyon, line 112: I have doubts about this flag...
We have a bft flag in the request itself (addRemoveWithWedge). 
Having a flag also here means that the operator needs to remember what was the original command, otherwise, what is the meaning of bft in the status where the original request was without bft flag (and vice versa).
What do you thing about returning the bft flag according to the actually stored cmd?

yontyon (changes requested):

yontyon (approved):

PR #1732, opened 2021-07-28
Client pool - add callback to requests

Liorgal28:
As part of moving to the client service, one of the requirements was the ability to get a response to requests from the client pool.
For this purpose, we are sending the request along with a callback; through this callback, we will return the reply once the request has been completed.

Changes made to the flow do not affect the existing flow and have been verified using Docker compose.

### DIFF
--- a/bftengine/include/bftengine/SimpleClient.hpp
+++ b/bftengine/include/bftengine/SimpleClient.hpp
@@ -21,6 +21,7 @@
 #include "Metrics.hpp"
 #include "communication/ICommunication.hpp"
 #include "PerformanceManager.hpp"
+#include "../../../bftclient/include/bftclient/base_types.h"
 
 namespace bftEngine {
 struct SimpleClientParams {
@@ -46,6 +47,7 @@ enum ClientMsgFlag : uint8_t {
 };
 
 enum OperationResult : int8_t { SUCCESS, NOT_READY, TIMEOUT, BUFFER_TOO_SMALL, INVALID_REQUEST };
+typedef std::function<void(bft::client::Reply&&)> RequestCallBack;
 
 struct ClientRequest {
   uint8_t flags = 0;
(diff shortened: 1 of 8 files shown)


bftengine/include/bftengine/SimpleClient.hpp
```diff
@@ -21,6 +21,7 @@
 #include "Metrics.hpp"
 #include "communication/ICommunication.hpp"
 #include "PerformanceManager.hpp"
+#include "../../../bftclient/include/bftclient/base_types.h"
```
robem, line 24: Ideally, we can `#include "bftclient/base_types.h`. If this doesn't work because the bftclient CMake target doesn't set its include dir properly then don't worry. We fix it later.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -35,7 +35,7 @@ struct SendError {
   };
   ErrorType type;
 };
-typedef std::variant<SendError, bft::client::Reply> SendResult;
+typedef std::variant<concord_client_pool::SubmitResult, bft::client::Reply> SendResult;
```
robem, line 38: At some point, concord_client.hpp should be self-contained so that we don't make any references to the client pool  or the TRC. With this change, client pool knowledge is required inside the request service. We implicitly included TRC and client_pool in the clientservice already. Nothing to change here, just FYI :)

robem (approved):
Lgtm. Thanks @Liorgal28 💪

bftengine/include/bftengine/SimpleClient.hpp
Liorgal28: NP

client/concordclient/include/client/concordclient/concord_client.hpp
Liorgal28: Agree I just didn't want to expose unrelated things in the pool right now.

PR #1733, opened 2021-07-28
Add ability to init Replica internals before start

dartdart26:
Introduce the Replica::initInternals() method. Used as a *workaround*,
because doing init in the constructor wouldn't work in all cases due to
some external components being set via set* methods. Additionally, some
of these external components depend on the replica itself being
constructed.
If not explicitly called by the user, Replica::start() will call
Replica::initInternals() automatically.


toly-kournik (approved):
LGTM!

PR #1734, opened 2021-07-28
Adding client keys as post reconfiguration block

cloudnoize:
Will add tests soon

### DIFF
--- a/kvbc/include/reconfiguration_kvbc_handler.hpp
+++ b/kvbc/include/reconfiguration_kvbc_handler.hpp
@@ -125,4 +125,18 @@ class InternalKvReconfigurationHandler : public concord::reconfiguration::IRecon
               uint64_t bft_seq_num,
               concord::messages::ReconfigurationResponse&) override;
 };
+
+class InternalPostKvReconfigurationHandler : public concord::reconfiguration::IReconfigurationHandler,
+                                             public ReconfigurationBlockTools {
+ public:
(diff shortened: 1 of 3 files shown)


kvbc/include/reconfiguration_kvbc_handler.hpp
```diff
@@ -125,4 +125,18 @@ class InternalKvReconfigurationHandler : public concord::reconfiguration::IRecon
               uint64_t bft_seq_num,
               concord::messages::ReconfigurationResponse&) override;
 };
+
+class InternalPostKvReconfigurationHandler : public concord::reconfiguration::IReconfigurationHandler,
+                                             public ReconfigurationBlockTools {
+ public:
+  InternalPostKvReconfigurationHandler(kvbc::IBlockAdder& block_adder, kvbc::IReader& ro_storage)
+      : ReconfigurationBlockTools{block_adder, ro_storage} {}
+  bool verifySignature(uint32_t sender_id, const std::string& data, const std::string& signature) const override {
```
toly-kournik, line 134: no need to have a trivial implementation here

toly-kournik (approved):
LGTM!

kvbc/include/reconfiguration_kvbc_handler.hpp
cloudnoize: apparently verifySignature is pure virutal
```// The verification method is pure virtual as all subclasses has to define how they verify the reconfiguration
  // requests.
  virtual bool verifySignature(uint32_t sender_id, const std::string& data, const std::string& signature) const = 0;```

PR #1735, opened 2021-07-28
Further refactoring of the Views manager and new unit tests

StefanPalashev:
This pull request includes the following changes: 

- The interface of the views manager has been made more robust by stating the necessary prerequisites in some functions.
- The logic for summarizing view change progress has been moved to the ViewsManager.
- Two new unit tests that verify the behavior of the ViewsManager have been added.

### DIFF
--- a/bftengine/src/bftengine/ViewsManager.cpp
+++ b/bftengine/src/bftengine/ViewsManager.cpp
@@ -1142,7 +1142,7 @@ void ViewsManager::processComplaintsFromViewChangeMessage(ViewChangeMsg* msg,
   // Complaints for higher view should aim to reach the view of the message.
   targetView = msg->newView();
 
-  while (msg->newView() > getCurrentView() && !(hasQuorumToLeaveView() || hasQuorumToJumpToHigherView()) &&
+  while (targetView > getCurrentView() && !(hasQuorumToLeaveView() || hasQuorumToJumpToHigherView()) &&
          iter.getAndGoToNext(complaint, size) && numberOfProcessedComplaints <= F + 1) {
     numberOfProcessedComplaints++;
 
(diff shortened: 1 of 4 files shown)


bftengine/src/bftengine/ViewsManager.cpp
```diff
@@ -1105,6 +1139,9 @@ void ViewsManager::processComplaintsFromViewChangeMessage(ViewChangeMsg* msg,
   MsgSize size = 0;
   int numberOfProcessedComplaints = 0;
 
+  // Complaints for higher view should aim to reach the view of the message.
+  targetView = msg->newView();
```
HristoStaykov, line 1143: I think it might be more readable if we replace **msg->newView()** with **targetView** in the references inside the function **processComplaintsFromViewChangeMessage**. This way we show that this is the View we are checking if we have sufficient evidence we need to reach.

bftengine/src/bftengine/ViewsManager.hpp
```diff
@@ -145,9 +145,12 @@ class ViewsManager {
 
   void storeComplaintForHigherView(std::unique_ptr<ReplicaAsksToLeaveViewMsg> &&complaintMessage);
   bool hasQuorumToJumpToHigherView() const { return complainedReplicasForHigherView.hasQuorumToLeaveView(); }
```
HristoStaykov, line 147: One remark that is related to the previous PR with refactors in ViewsManager that I missed then, but I think it is not too late to make it better :) Let's make those 2 functions protected, we only use them when processing View Change Messages when we iterate the complaints they contain.

HristoStaykov (approved):
LGTM, great work @StefanPalashev ! I only have 2 minor proposals for improvement.

PR #1736, opened 2021-07-28
Support callbacks on consuming external messages

dartdart26:
Add a functionality to **IncomingMsgsStorage** that allows users to pass
in a callback that will be called when the given message is consumed by
the replica. Such a functionality can be useful to be able to determine
if a particular message is still pending or not. An example is Concord
Cron's TicksGenerator - it will be able to know whether a tick is
pending in the external message queue.

Make pushExternal* methods return a boolean that is true on success and
false otherwise (e.g. queue capacity has been reached).

To implement, in addition to the unique_ptr to the message, store a
std::function callback, making the queue contain a std:pair of these
two. The resulting std::pair will remain movable only. Expectation is
that storing and moving an empty std::function will not degrade
performance noticeably.

Make sure we don't register the performance component if already
registered with the set of shared recorders. That allows us to have
multiple instances of **IncomingMsgsStorageImp**.

Add an unit test that mostly covers new functionality. More tests for
existing functionality will be added in separate commits.


bftengine/src/bftengine/IncomingMsgsStorageImp.hpp
```diff
@@ -68,16 +71,18 @@ class IncomingMsgsStorageImp : public IncomingMsgsStorage {
   std::shared_ptr<MsgHandlersRegistrator> msgHandlers_;
   std::chrono::milliseconds msgWaitTimeout_;
 
+  using MessageWithCallback = std::pair<std::unique_ptr<MessageBase>, Callback>;
+
   // New messages are pushed to ptrProtectedQueue.... ; protected by lock
-  std::queue<std::unique_ptr<MessageBase>>* ptrProtectedQueueForExternalMessages_;
+  std::queue<MessageWithCallback>* ptrProtectedQueueForExternalMessages_;
```
toly-kournik, line 77: I know that you used the existing style, but can we make it a smart_ptr?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1140,7 +1140,7 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
   pm_->Delay<concord::performance::SlowdownPhase::ConsensusFullCommitMsgProcess>(
       (char *)msg,
       msg->sizeNeededForObjAndMsgInLocalBuffer(),
-      std::bind(&IncomingMsgsStorage::pushExternalMsgRaw, &getIncomingMsgsStorage(), _1, _2));
+      [&s = getIncomingMsgsStorage()](char *msg, size_t &size) { s.pushExternalMsgRaw(msg, size); });
```
toly-kournik, line 1143: nice

bftengine/src/bftengine/IncomingMsgsStorageImp.hpp
toly-kournik: Oh, it's for swapping ...

toly-kournik: so `swap()` can be used

toly-kournik:
So the callback is executed in context of the incoming message processing thread?
Does this mean that the caller can't wait for a callback execution in a manner of promise/future if running in the same thread (as most of our code)?
What will be a typical callback, a change of flag?

bftengine/src/bftengine/IncomingMsgsStorageImp.hpp
dartdart26: I was thinking of making them unique_ptr, but I decided to not convolute this PR. As an off-topic, I also wanted to make the promise a non-member, in order to allow multiple start/stops of the same object. Can I make these improvements in a separate PR?

dartdart26:
> So the callback is executed in context of the incoming message processing thread?
> Does this mean that the caller can't wait for a callback execution in a manner of promise/future if running in the same thread (as most of our code)?
> What will be a typical callback, a change of flag?

Yes, the callback will be called in the message processing thread. I guess if a caller from another thread wants to, they can pass in a function that has a promise bound to it and signal it in the callback. If we are changing state from the messaging thread only, I assume we don't need promise/future. Did I get you?

A callback that I have in mind is that the messaging thread processes an internal message from Cron's TickGenerator and pushes it to the external queue with a callback. This callback will have some contextual information - in particular, the Cron component ID in question. Then, when this message gets popped out, we will notify the TicksGenerator that it can generate a new tick for this component ID. This will happen in the same message processing thread. I want to implement in such a way instead of just processing the tick from the internal message, because it preserves fairness (as internal messages have precedence over external ones) and queues the tick correctly relative to all other requests in the system.

Above is shown on this picture: https://github.com/vmware/concord-bft/tree/master/ccron

toly-kournik:
> > So the callback is executed in context of the incoming message processing thread?
> > Does this mean that the caller can't wait for a callback execution in a manner of promise/future if running in the same thread (as most of our code)?
> > What will be a typical callback, a change of flag?
> 
> Yes, the callback will be called in the message processing thread. I guess if a caller from another thread wants to, they can pass in a function that has a promise bound to it and signal it in the callback. If we are changing state from the messaging thread only, I assume we don't need promise/future. Did I get you?
> 
> A callback that I have in mind is that the messaging thread processes an internal message from Cron's TickGenerator and pushes it to the external queue with a callback. This callback will have some contextual information - in particular, the Cron component ID in question. Then, when this message gets popped out, we will notify the TicksGenerator that it can generate a new tick for this component ID. This will happen in the same message processing thread. I want to implement in such a way instead of just processing the tick from the internal message, because it preserves fairness (as internal messages have precedence over external ones) and queues the tick correctly relative to all other requests in the system.
> 
> Above is shown on this picture: https://github.com/vmware/concord-bft/tree/master/ccron

toly-kournik:
I think it somehow should be documented otherwise someone can try to wait on future forever

dartdart26:
> I think it somehow should be documented otherwise someone can try to wait on future forever

I can say that the callback will be called in the consumer thread(s) and users should be aware of that.

toly-kournik (approved):
LGTM!

PR #1739, opened 2021-07-28
Fix Clientservice startup

robem:
Now that we are starting to put all the pieces together, there are some minor issue that need to be fixed. Please see the commit messages for details.


shgandhi (approved):
Thanks @robem. LGTM!

Liorgal28 (approved):

PR #1742, opened 2021-07-29
State transfer performance improvments fix retransmissions

glevkovich:
Here we fix retransmissions issues in state transfer, and a minor bug.
This PR is adding some retransmission logic into the source selector (It’s a must for stability considerations).
What we had till now was pretty much useless, the source was never replaced after multiple retransmission failures.
Here are the commits:

1. BCStateTran: fix a bug on source snapshots.
2. BCStateTran: record message arrival time (into handoff).
3. SourceSelector: fix retransmissions, modify change source protocol.
4. source_selector_test: refactor and add test for retransmissions.


toly-kournik (approved):
LGTM!

PR #1743, opened 2021-07-29
Clientservice: Rename proto

robem:
There is never a good time to rename API but if at all then it is now. The
reason is to align with our abstractions. Clientservice and concord client are
two different components. The clientservice is a gRPC wrapper around the
concord client library. Therefore, the exposed gRPC API is clientservice API
and not concord client API. For maintainability and logevity of the code, this
change will make the code easier to grasp.


robem:
Closed in favor of https://github.com/vmware/concord-bft/pull/1744

PR #1744, opened 2021-07-29
Clientservice: Proto reorg

robem:
The file name is inconsistent with the application name but it's also wrong. Clientservice and concord client are two different components. Instead of renaming the file, and following Google's recommendations, we can split the file in two. Each file contains one service and therefore the file name is defined. No need to deal with the name clientservice at all.
I addition, the API is self-contained and can stand on its own. Clientservice just happens to implement that API.


shgandhi (approved):
Thanks for the refactoring @robem! 💯  
LGTM.

PR #1745, opened 2021-07-30
TRC: EventGroup hashing

robem:
This PR copies the hashing implementation from the TRS and adds two tests. Please see commit message for detail.

### DIFF
--- a/client/thin-replica-client/src/trc_hash.cpp
+++ b/client/thin-replica-client/src/trc_hash.cpp
@@ -14,6 +14,7 @@
 #include "client/thin-replica-client/trc_hash.hpp"
 
 #include <map>
+#include <set>
 
 #include "boost/detail/endian.hpp"
 #include "assertUtils.hpp"
@@ -22,6 +23,7 @@ using com::vmware::concord::thin_replica::Data;
 using std::invalid_argument;
 using std::list;
 using std::map;
+using std::set;
 using std::string;
 
 namespace client::thin_replica_client {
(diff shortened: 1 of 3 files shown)


client/thin-replica-client/src/trc_hash.cpp
```diff
@@ -68,40 +70,80 @@ static string hashUpdateFromEntryHashes(uint64_t block_id, const map<string, str
...
+
+    return hashUpdateFromEntryHashes(legacy_event.block_id, entry_hashes);
   }
 
-  return hashUpdateFromEntryHashes(legacy_event.block_id, entry_hashes);
+  ConcordAssert(std::holds_alternative<EventGroup>(ev));
+  auto& event_group = std::get<EventGroup>(ev);
+  set<string> entry_hashes;
+  for (const auto& event : event_group.events) {
+    string event_hash = ComputeSHA256Hash(event);
```
shgandhi, line 117: Event includes tags as well, we don't want them to be included in the hash afaik, and only hash `event.data`.

client/thin-replica-client/src/trc_hash.cpp
```diff
@@ -68,40 +70,80 @@ static string hashUpdateFromEntryHashes(uint64_t block_id, const map<string, str
...
+      entry_hashes[key_hash] = ComputeSHA256Hash(kvp.value());
     }
-    entry_hashes[key_hash] = ComputeSHA256Hash(kvp.value());
+
+    return hashUpdateFromEntryHashes(update.events().block_id(), entry_hashes);
+  }
+  ConcordAssert(update.has_event_group());
+  set<string> entry_hashes;
+  for (const auto& event : update.event_group().events()) {
+    string event_hash = ComputeSHA256Hash(event);
```
shgandhi, line 141: Same comment as above, here.

shgandhi (approved):
Thanks for the change @robem, LGTM! I have one comment though, if you could take a look.

client/thin-replica-client/src/trc_hash.cpp
robem: The `event` in this for-loop comes from EventGroup.events which is a `vector<string>` type. If we include tags on the server side then we need to remove them there before sending.

client/thin-replica-client/src/trc_hash.cpp
robem: The event in this for-loop comes from EventGroup.events (thin_replica.proto) which is a repeated bytes field so essentially a `vector<string>` type. Tags shouldn't be included by the server.

PR #1749, opened 2021-07-30
State Transfer Test Suite has been added

ananpal:
and a new state transfer test case with multiple clients
sending the concurrent request has been added

### DIFF
--- a/tests/apollo/test_skvbc_state_transfer.py
+++ b/tests/apollo/test_skvbc_state_transfer.py
@@ -1,0 +1,99 @@
+# Concord
+#
+# Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+#
+# This product is licensed to you under the Apache 2.0 license (the "License").
+# You may not use this product except in compliance with the Apache 2.0 License.
+#
+# This product may include a number of subcomponents with separate copyright
+# notices and license terms. Your use of these subcomponents is subject to the
+# terms and conditions of the subcomponent's license, as noted in the LICENSE
+# file.
+import os.path
+import random
+import unittest
+from os import environ
+
+import trio
+
+from util import skvbc as kvbc
+from util.bft import with_trio, with_bft_network, KEY_FILE_PREFIX
+from util import eliot_logging as log
+
+def start_replica_cmd(builddir, replica_id):
+    """
+    Return a command that starts an skvbc replica when passed to
+    subprocess.Popen.
+    Note each arguments is an element in a list.
+    """
+    statusTimerMilli = "500"
+    viewChangeTimeoutMilli = "10000"
+    path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
+    return [path,
+            "-k", KEY_FILE_PREFIX,
+            "-i", str(replica_id),
+            "-s", statusTimerMilli,
+            "-v", viewChangeTimeoutMilli,
+            "-e", str(True)
+            ]
+
+
+class SkvbcStateTransferTest(unittest.TestCase):
+
+    __test__ = False  # so that PyTest ignores this test scenario
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, rotate_keys=True)
+    async def test_state_transfer(self, bft_network,exchange_keys=True):
+        """
+        Test that state transfer starts and completes.
+        Stop one node, add a bunch of data to the rest of the cluster, restart
+        the node and verify state transfer works as expected. We should be able
+        to stop a different set of f nodes after state transfer completes and
+        still operate correctly.
+        """
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        stale_node = random.choice(
+            bft_network.all_replicas(without={0}))
+
+        await skvbc.prime_for_state_transfer(
+            stale_nodes={stale_node},
+            checkpoints_num=3, # key-exchange channges the last executed seqnum
+            persistency_enabled=False
+        )
+        bft_network.start_replica(stale_node)
+        await bft_network.wait_for_state_transfer_to_start()
+        await bft_network.wait_for_state_transfer_to_stop(0, stale_node)
+        await skvbc.assert_successful_put_get()
+        await bft_network.force_quorum_including_replica(stale_node)
+        await skvbc.assert_successful_put_get()
+
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, rotate_keys=True)
+    async def test_state_transfer_with_multiple_clients(self, bft_network,exchange_keys=True):
+        """
+        Test that state transfer starts and completes.
+        Stop one node, add a bunch of data to the rest of the cluster using  
+        multiple clients restart the node and verify state transfer works as 
+        expected. We should be able to stop a different set of f nodes after 
+        state transfer completes and still operate correctly.
+        """
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        stale_node = random.choice(
+            bft_network.all_replicas(without={0}))
+    
+        await skvbc.write_with_multiple_clients_for_state_transfer(
+            stale_nodes={stale_node},
+            write_run_duration=30 , # duration for which clients are sending txns
+            persistency_enabled=False
+        )
+        bft_network.start_replica(stale_node)
+        await bft_network.wait_for_state_transfer_to_start()
+        await bft_network.wait_for_state_transfer_to_stop(0, stale_node)
+        await skvbc.assert_successful_put_get()
+        await bft_network.force_quorum_including_replica(stale_node)
+        await skvbc.assert_successful_put_get()
+
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -501,6 +501,9 @@ class BftTestNetwork:
     def random_clients(self, max_clients):
         return set(random.choices(list(self.clients.values()), k=max_clients))
 
+    def get_all_clients(self):
+        return self.clients.values()
+
     def setup_txn_signing(self):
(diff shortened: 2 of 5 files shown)


tests/apollo/util/skvbc.py
```diff
@@ -227,6 +229,31 @@ async def prime_for_state_transfer(
                 verify_checkpoint_persistency=persistency_enabled)
 
             return client, known_key, known_val
+    
+    async def write_with_multiple_clients_for_state_transfer(
```
cjn, line 233: Was write_with_multiple_clients_for_state_transfer intended to send requests from multiple clients or single client?

tests/apollo/util/skvbc.py
```diff
@@ -262,6 +289,41 @@ async def fill_and_wait_for_checkpoint(
...
+            async def write_random_kv(client):
+                key = self.random_key()
+                val = self.random_value()
+                reply = await self.send_write_kv_set(client,[(key, val)])
+                assert reply.success
+            t_end = time.time() + write_run_duration
+
+            with trio.move_on_after(write_run_duration+CLIENT_REQ_GRACEFUL_END_TIMER):
+                while time.time() < t_end:
+                    async with trio.open_nursery() as nursery:
```
cjn, line 317: Can we call send_concurrent_ops() though that is driven by number of operations which can be provided?

tests/apollo/util/skvbc.py
ananpal: Yes celine its supposed to send multiple client requests.

tests/apollo/util/skvbc.py
ananpal: no celine here we are not fixing the number of operations but running the client req for a given period of time

tests/apollo/util/skvbc.py
```diff
@@ -22,6 +23,7 @@
 
 WriteReply = namedtuple('WriteReply', ['success', 'last_block_id'])
 
+CLIENT_REQ_GRACEFUL_END_TIMER=3
```
glevkovich, line 26: client requests timeout by default, if not changed can be taken from bft.py REQ_TIMEOUT_MILLI which is 5 seconds. you can assign this variable from there or just remove it and import REQ_TIMEOUT_MILLI from bft. I would add 1 more second to be on the safe side (so actual value will be 6):

```
from bft import REQ_TIMEOUT_MILLI
....
....
CLIENT_REQ_GRACEFUL_END_TIMER = REQ_TIMEOUT_MILLI + 1

```

tests/apollo/util/skvbc.py
```diff
@@ -262,6 +289,41 @@ async def fill_and_wait_for_checkpoint(
                 expected_checkpoint_num=lambda ecn: ecn == checkpoint_before + num_of_checkpoints_to_add,
                 verify_checkpoint_persistency=verify_checkpoint_persistency,
                 assert_state_transfer_not_started=assert_state_transfer_not_started)
+    
+    async def run_concurrent_txns_and_wait_for_checkpoint(
+            self, initial_nodes,
```
glevkovich, line 294: nit: The name initial_nodes is not a good one in my opinion, as this is a service function, and know nothing about the fact you are doing state transfer. Maybe it needs to wait for all replicas for example? Why not just call it replicas?

tests/apollo/util/skvbc.py
```diff
@@ -262,6 +289,41 @@ async def fill_and_wait_for_checkpoint(
...
+            write_run_duration=10,
```
glevkovich, line 295: nit: I would remove the default value for this parameter, each caller would choose its own duration.

tests/apollo/test_skvbc_state_transfer.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, rotate_keys=True)
+    async def test_state_transfer_with_multiple_clients(self, bft_network,exchange_keys=True):
+        """
+        Test that state transfer starts and completes.
+        Stop one node, add a bunch of data to the rest of the cluster, restart
+        the node and verify state transfer works as expected. We should be able
+        to stop a different set of f nodes after state transfer completes and
```
mdarade, line 81: Can this comment be any different than other test case?

tests/apollo/util/skvbc.py
glevkovich: @ananpal ,  I think @cjn might be right. we might have missed something here... it seems like send_concurrent_ops might be good for what we need? Just need to add an option to insert num_ops as null value and insert a duration instead?

glevkovich (changes requested):

tests/apollo/util/skvbc.py
ananpal: @glevkovich I feel that changing this function for our need doesn't look very straightforward and further it will complicate the existing code so it's better to use a new function.

tests/apollo/util/skvbc.py
glevkovich: How does it complicate the code?
send_concurrent_ops is used in 3-4 places, so any refactoring is not a big deal.

- replace 2nd parameter num_ops with a callable function as a parameter:
`send_concurrent_ops(self, exit_predicate, ...`

- replace
`while sent < num_ops:`
with
`while not exit_predicate():`

That's all, all you need is to define the functions (lambda) in the caller, 1 liner.
I don't see any complications.
Code reuse is important...

tests/apollo/util/skvbc.py
glevkovich: @ananpal 
How does it complicate the code?
send_concurrent_ops is used in 3-4 places, so any refactoring is not a big deal.

- replace 2nd parameter num_ops with a callable function as a parameter:
`send_concurrent_ops(self, exit_predicate, ...`

- replace
`while sent < num_ops:`
with
`while not exit_predicate():`

That's all, all you need is to define the functions (lambda) in the caller, 1 liner.
I don't see any complications.
Code reuse is important...

tests/apollo/util/bft.py
```diff
@@ -1044,6 +1048,20 @@ async def wait_for_replicas_to_collect_stable_checkpoint(self, replicas, checkpo
                     else:
                         last_stable_seqs.clear()
                         await trio.sleep(seconds=0.1)
+    async def wait_and_verify_replicas_to_reach_stable_seq_num(self, replicas, wait_duration=3, timeout=30):
```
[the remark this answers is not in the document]
ananpal: yes it was added for future use , removed it for now

tests/apollo/util/skvbc.py
ananpal: done

tests/apollo/util/bft.py
```diff
@@ -1044,6 +1048,20 @@ async def wait_for_replicas_to_collect_stable_checkpoint(self, replicas, checkpo
...
+        with log.start_action(action_type="wait_and_verify_replicas_to_reach_stable_seq_num", replicas=replicas) as action:
+            with trio.fail_after(seconds=timeout):
+                last_stable_seqs = []
```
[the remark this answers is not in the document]
ananpal: removed

tests/apollo/util/skvbc.py
```diff
@@ -482,16 +519,25 @@ async def run_concurrent_conflict_ops(self, num_ops, write_weight=.70):
         max_size = len(self.keys) // 2
         return await self.send_concurrent_ops(num_ops, max_concurrency, max_size, write_weight, create_conflicts=True)
 
-    async def send_concurrent_ops(self, num_ops, max_concurrency, max_size, write_weight, create_conflicts=False):
+    async def send_concurrent_ops(self, num_ops, max_concurrency, max_size, write_weight, create_conflicts=False,write_run_duration=None):
```
glevkovich, line 522: I think it should be more general,  the way the function is called is unclear like that. Two parameters are used for the limit and the caller must-knows of the exact internal implementation. I would also add a comment to explain. Here is a pseudo-code example (haven't checked it) in a clearer way:

```
from enum import Enum # should be added at the top of the file
    class ExitPolicy(Enum):
        COUNT = 0
        TIME = 1

    async def send_concurrent_ops(self, max_concurrency, max_size, write_weight, exit_factor, create_conflicts=False, exit_policy=ExitPolicy.COUNT):
        ''' exit_factor should be number of seconds to run if exit_policy is TIME and number of times to run if exit_policy is COUNT '''
        max_read_set_size = max_size
        if self.tracker is not None:
            max_read_set_size = 0 if self.tracker.no_conflicts else max_size
        sent = 0
        write_count = 0
        read_count = 0
        if exit_policy is ExitPolicy.COUNT:
            total_run_duration = time.time()
            total_run_duration += exit_factor
        clients = self.bft_network.random_clients(max_concurrency)

        def exit_predicate():
            if exit_policy is ExitPolicy.COUNT:
                return sent < exit_factor
            if exit_policy is ExitPolicy.TIME:
                return time.time() < total_run_duration
            assert false, "Unknown exit policy"
...
...
...
```

tests/apollo/test_skvbc_state_transfer.py
```diff
@@ -0,0 +1,99 @@
...
+    viewChangeTimeoutMilli = "10000"
+    path = os.path.join(builddir, "tests", "simpleKVBC", "TesterReplica", "skvbc_replica")
+    return [path,
+            "-k", KEY_FILE_PREFIX,
+            "-i", str(replica_id),
+            "-s", statusTimerMilli,
+            "-v", viewChangeTimeoutMilli,
+            "-e", str(True)
+            ]
+
```
glevkovich, line 39: can you please remove the line?

glevkovich (changes requested):

glevkovich (approved):
LGTM, thanks!

PR #1751, opened 2021-07-30
Cron: check for pending ticks in the external queue

dartdart26:
Fix a bug in the Concord Cron's TickGenerator that would allow
generating a tick for a component ID when there is already a pending one
for the same component ID. The path of a tick is the following:
   `internal msg queue -> external msg queue -> consensus`
Make sure that we treat ticks in the external message queue as pending
too instead of only treating ticks that are in consensus as pending.

In order to implement, introduce a set of ticks (per component ID) that
are pending in the external message queue. Attach a callback to ticks in
the external message queue that is called when the tick is processed
(either dropped or consumed for consensus). This callback deletes the
tick from the set of pending ones in the external queue.

Use unordered_set/map instead of ordered ones.

Add support for a callback when a request is popped from the external
queue in the messaging thread in IInternalBFTClient.

Force that TicksGenerator is only created as a shared_ptr as we want to
pass it around to callbacks. It is already used as a shared_ptr anyway.

Add a test case for the new behaviour.

### DIFF
--- a/bftengine/src/bftengine/InternalBFTClient.cpp
+++ b/bftengine/src/bftengine/InternalBFTClient.cpp
@@ -15,15 +15,25 @@
 #include "chrono"
 #include "Logger.hpp"
 
+#include <utility>
+
 uint64_t InternalBFTClient::sendRequest(uint64_t flags,
                                         uint32_t requestLength,
                                         const char* request,
                                         const std::string& cid) {
+  return sendRequest(flags, requestLength, request, cid, std::function<void()>{});
+}
+
+uint64_t InternalBFTClient::sendRequest(uint64_t flags,
+                                        uint32_t requestLength,
+                                        const char* request,
+                                        const std::string& cid,
+                                        std::function<void()> onPoppedFromQueue) {
   auto now = getMonotonicTime().time_since_epoch();
   auto now_ms = std::chrono::duration_cast<std::chrono::microseconds>(now);
   auto sn = now_ms.count();
   auto crm = new ClientRequestMsg(getClientId(), flags, sn, requestLength, request, 60000, cid);
-  msgComm_->getIncomingMsgsStorage()->pushExternalMsg(std::unique_ptr<MessageBase>(crm));
+  msgComm_->getIncomingMsgsStorage()->pushExternalMsg(std::unique_ptr<MessageBase>(crm), std::move(onPoppedFromQueue));
   LOG_DEBUG(GL, "Sent internal consensus: seq num [" << sn << "] client id [" << getClientId() << "]");
   return sn;
 }
--- a/bftengine/src/bftengine/InternalBFTClient.cpp
+++ b/bftengine/src/bftengine/InternalBFTClient.cpp
@@ -21,14 +21,14 @@ uint64_t InternalBFTClient::sendRequest(uint64_t flags,
                                         uint32_t requestLength,
                                         const char* request,
                                         const std::string& cid) {
-  return sendRequest(flags, requestLength, request, cid, std::function<void()>{});
+  return sendRequest(flags, requestLength, request, cid, IncomingMsgsStorage::Callback{});
 }
 
 uint64_t InternalBFTClient::sendRequest(uint64_t flags,
                                         uint32_t requestLength,
                                         const char* request,
                                         const std::string& cid,
-                                        std::function<void()> onPoppedFromQueue) {
+                                        IncomingMsgsStorage::Callback onPoppedFromQueue) {
   auto now = getMonotonicTime().time_since_epoch();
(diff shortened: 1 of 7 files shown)


bftengine/src/bftengine/InternalBFTClient.cpp
```diff
@@ -15,15 +15,25 @@
 #include "chrono"
 #include "Logger.hpp"
 
+#include <utility>
+
 uint64_t InternalBFTClient::sendRequest(uint64_t flags,
                                         uint32_t requestLength,
                                         const char* request,
                                         const std::string& cid) {
+  return sendRequest(flags, requestLength, request, cid, std::function<void()>{});
```
cloudnoize, line 24: in [https://github.com/vmware/concord-bft/pull/1736](url) you defined ```using Callback = std::function<void()>;``` maybe it's worth using it

cloudnoize:
> Fix a bug in the Concord Cron's TickGenerator that would allow
>generating a tick for a component ID when there is already a pending one
>for the same component ID. The path of a tick is the following:
>internal msg queue -> external msg queue -> consensus
>Make sure that we treat ticks in the external message queue as pending
>too instead of only treating ticks that are in consensus as pending.

for my understanding, can the TickGenerator generate a tick for a component ID that has generated a tick and not its msg is in the consensus phase? and if so why is it different than the bug?

ccron/src/ticks_generator.cpp
```diff
@@ -93,9 +108,14 @@ void TicksGenerator::run() {
...
-  const auto seq_num = bft_client_->sendRequest(
-      bftEngine::TICK_FLAG, payload.size(), reinterpret_cast<const char*>(payload.data()), kTickCid);
-  component_pending_req_seq_nums_[component_id] = seq_num;
+  const auto seq_num =
+      bft_client_->sendRequest(bftEngine::TICK_FLAG,
+                               payload.size(),
+                               reinterpret_cast<const char*>(payload.data()),
+                               kTickCid,
+                               [p = shared_from_this(), component_id]() { p->onTickPoppedFromExtQueue(component_id); });
+  pending_tick_requests_[component_id] = seq_num;
```
cloudnoize, line 117: I probably miss something, but I would have expected also a call to unset ```pending_tick_requests_[component_id] = seq_num``` on handling, no?

ccron/include/ccron/ticks_generator.hpp
```diff
@@ -65,10 +70,14 @@ class TicksGenerator {
   const IPendingRequest *pending_req_{nullptr};
```
cloudnoize, line 70: not related to this PR, but must say that ```IPendingRequest``` is confusing name, I didn't expect the client manager to implement it

bftengine/src/bftengine/InternalBFTClient.cpp
dartdart26: Sure, I will do that.

ccron/src/ticks_generator.cpp
dartdart26: The idea to keep it is that we map from component ID -> BFT seq number. In order to check if the request is still pending in the clients manager, we need the BFT sequence number if the request is not in the external queue:

```
    // If the tick for the given component ID is not in the external queue, check if it is pending. If yes, don't send.
    if (!pending_req_->isPending(bft_client_->getClientId(), req_it->second)) {
      sendClientRequestMsgTick(internal_tick.component_id);
    }
```

Does that make sense?

ccron/include/ccron/ticks_generator.hpp
dartdart26: It has been introduced for testing only, to avoid constructing a real clients manager in unit tests. It is not an interface users must rely on.

I am open to changing the name. Any ideas for a different one?

ccron/src/ticks_generator.cpp
dartdart26: I have added a diagram in the `onInternalTick()` method that explains it better. Thanks to @HristoStaykov 's comment that suggested adding it.

cloudnoize: I see, and after consensus is finished, does it stay there? can it cause an issue?

ccron/include/ccron/ticks_generator.hpp
cloudnoize: not really :) any way it's nit

cloudnoize (approved):
LGTM!

HristoStaykov (approved):
LGTM, thanks @dartdart26 !

ccron/src/ticks_generator.cpp
dartdart26: It shouldn't, because we assume the internal BFT client doesn't generate duplicate BFT sequence numbers and, therefore, an already committed request would never be pending again.

PR #1752, opened 2021-07-31
ConcordClient: Change subscribe() API

robem:
In order to simplify our current end-to-end tests, we change the current
subscribe() API. The callback API is removed and replaced with a function call
that returns the TRC queue. The latter can be looped over to get updates from
the replicas. The idea of concord client is to be the abstraction over the
client pool and trc. This change is in violation of those goals and made
intentionally at this point.


shgandhi (approved):
LGTM.

Liorgal28 (approved):
LGTM!

PR #1753, opened 2021-08-01
BCStateTran: add and maintain logger_ member

glevkovich:
logging::Logger& logger_ is added to BCStateTran:
* By default, it is set to ST_SRC_LOG
* Every time state is changed, logger_ might be re-assigned with
ST_SRC_LOG or ST_DST_LOG.

This change saves us the multiple function calls to getLogger() and then
psd_->getIsFetchingState() every time we just want to log something.


f-squirrel (approved):
LGTM!

PR #1756, opened 2021-08-02
Performance improvement for PRA

arc-vmware:
1) Improved the performance of Pre Process Result
verification flow.
2) The verification was taking 20ms. After
this fix, it will reduce from 20ms to 2 ms
3) This is tested using perf tool

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -48,6 +48,7 @@
 #include "bftengine/KeyExchangeManager.hpp"
 #include "secrets_manager_plain.h"
 #include "bftengine/EpochManager.hpp"
+#include "RequestThreadPool.hpp"
 #include <memory>
 #include <optional>
 #include <string>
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -685,27 +686,38 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+          getReplicaConfig().replicaId,
+          getReplicaConfig().fVal));
+      error_id++;
+    }
+  }
+  for (const auto &t : tasks) {
+    t.wait();
+  }
+  errors.resize(error_id);
+  for (auto err : errors) {
```
teoparvanov, line 713: Amitava, from reading this code, I don't understand how we only check for `f+1` valid signatures, instead of `n`?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -685,27 +686,38 @@ bool ReplicaImp::relevantMsgForActiveView(const T *msg) {
...
+    if (err) {
+      // trigger view change
+      LOG_INFO(VC_LOG,
```
yuliasherman, line 716: Make it WARN

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Thank you @teoparvanov for this comment. As per my understanding, every replica receives f signatures. I will discuss with @tdimitrov and then ans this comment.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Thank you, I will make this as LOG_WARN.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -48,6 +48,7 @@
 #include "bftengine/KeyExchangeManager.hpp"
 #include "secrets_manager_plain.h"
 #include "bftengine/EpochManager.hpp"
+#include "RequestThreadPool.hpp"
```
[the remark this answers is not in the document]
arc-vmware: Thank you @yuliasherman , Yes this should be static only as per our usecase separate thread pool for each thread is not needed. I will change this to static only.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: @teoparvanov , I had a discussion with @tdimitrov . We get only f signatures, because the primary also executes the requests. So the current logic is fine.

tdimitrov (approved):

bftengine/src/bftengine/ReplicaImp.cpp
teoparvanov: Hi Amitava, thanks for clarifying!

teoparvanov (approved):
LGTM!

yuliasherman:
You are probably want to return the feature to be enabled by default in this PR, too?

arc-vmware:
@yuliasherman I will set the feature true in MR. Thanks for the suggestion.

PR #1757, opened 2021-08-02
Respect USE_LOG4CPP cmake option when either set or not

toly-kournik:
Since we provide an option not to use log4cplus


yuliasherman (approved):
LGTM

PR #1758, opened 2021-08-02
Replica.cpp: fix histogram recording

glevkovich:
Since 2 of the recordings are donw in worker thread context, recording
must be done by calling recordAdomic.


HristoStaykov (approved):
LGTM, thanks @glevkovich !

PR #1760, opened 2021-08-02
EventService: Fix unsubscribe

robem:
With the recent API change of `subscribe` we didn't adjust unsubscribe and
thereby ended up with danling subscriptions. This change use atomic compare and
swap to grant access to the calling thread.


shgandhi (approved):
Thanks for the change @robem!

PR #1762, opened 2021-08-03
ThinReplica: Small fixes

robem:
Correct the calculation of the starting update that the TRC requests and fix a typo in TRS. See commit messages for details.


shgandhi (approved):
Sorry for the goof-up! 😞 
Thanks for the fix!

robem:
Note: I removed the change for block ids from this PR because we have tests in place that make sure that this behavior stays. I'm not investigating this right now. Important is that we align the behavior for event groups as outlined in `event.proto`.

PR #1764, opened 2021-08-03
SigManager compiling 

toly-kournik:
- Avoid compiling SigManager.cpp twice: for corebft and for preprocessor.
- As SigManager.hpp is included from several places, move include for generated
keys_and_signatures.cmf.hpp to SigManager.cpp .


cloudnoize (approved):
great, thanks @toly-kournik

PR #1766, opened 2021-08-03
Enforce BUILD_TESTING cmake configuration option.

toly-kournik:
Some tests were built even when BUILD_TESTING was set to FALSE.
Remove BUILD_KVBC_BENCH option. Build kvbc bench when BUILD_TESTING is
specified.


cloudnoize (approved):
LGTM

PR #1768, opened 2021-08-03
Enable graphwiz library dependency generation

toly-kournik:


cloudnoize (approved):
LGTM

cmake-build-options.sh
```diff
@@ -0,0 +1,27 @@
...
+# The list of options is obtained using the following command and are needed to be updated when options change:
+# cmake -LA ..| grep -vE "^--|^CMAKE|_DIR|_LIBRARY|NOTFOUND|^LIB|FILEPATH"
+#
+
+[[ -d build ]] || mkdir build
+cd build
+
+cmake \
+--graphviz=concord-bft.dot \
+-DBUILD_COMM_TCP_PLAIN=OFF \
```
f-squirrel, line 13: I would suggest moving the functionality to the Makelike as an additional parameter.
For example,
```
CONCORD_BFT_GRAPHIV_COMMAND?="--graphviz=concord-bft.dot"
```
and later to use it in the cmake_generation target over [here](https://github.com/vmware/concord-bft/blob/master/Makefile#L132):

```
.PHONY: gen_cmake
gen_cmake: ## Generate cmake files, used internally
	docker run ${CONCORD_BFT_USER_GROUP} ${BASIC_RUN_PARAMS} \
		${CONCORD_BFT_CONTAINER_SHELL} -c \
		"mkdir -p ${CONCORD_BFT_TARGET_SOURCE_PATH}/${CONCORD_BFT_BUILD_DIR} && \
		cd ${CONCORD_BFT_BUILD_DIR} && \
		CC=${CONCORD_BFT_CONTAINER_CC} CXX=${CONCORD_BFT_CONTAINER_CXX} \
		cmake ${CONCORD_BFT_CMAKE_FLAGS} ${CONCORD_BFT_GRAPHIV_COMMAND} .."
	@echo
	@echo "CMake finished."
```

In this case, the feature will be more accessible and will be able to reflect the most recent and usable flags by default.

toly-kournik: next step

PR #1769, opened 2021-08-04
Fix confusing log message

yuliasherman:
"The batch has been canceled as expired" pre-processor LOG not always shows up in the right cases.


toly-kournik (approved):
LGTM

PR #1770, opened 2021-08-04
Add support with all client key exchange command

yontyon:
In this PR we add support in "all" client key exchange command. 
If the target_clients in the request is empty, the reconfiguration handler writes the update for all clients.
Recall, that only the CRE client knows how to run this command. Thus, it is ok to write the update to all (the other option is to assume that the CRE client is a specific client from the group which we prefer not to do)

### DIFF
--- a/kvbc/include/reconfiguration_kvbc_handler.hpp
+++ b/kvbc/include/reconfiguration_kvbc_handler.hpp
@@ -69,8 +69,10 @@ class KvbcClientReconfigurationHandler : public concord::reconfiguration::Client
 class ReconfigurationHandler : public concord::reconfiguration::BftReconfigurationHandler,
                                public ReconfigurationBlockTools {
  public:
-  ReconfigurationHandler(kvbc::IBlockAdder& block_adder, kvbc::IReader& ro_storage)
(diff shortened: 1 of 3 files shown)


kvbc/include/reconfiguration_kvbc_handler.hpp
```diff
@@ -69,8 +69,10 @@ class KvbcClientReconfigurationHandler : public concord::reconfiguration::Client
 class ReconfigurationHandler : public concord::reconfiguration::BftReconfigurationHandler,
                                public ReconfigurationBlockTools {
  public:
-  ReconfigurationHandler(kvbc::IBlockAdder& block_adder, kvbc::IReader& ro_storage)
-      : ReconfigurationBlockTools{block_adder, ro_storage} {}
+  ReconfigurationHandler(kvbc::IBlockAdder& block_adder,
+                         kvbc::IReader& ro_storage,
+                         std::set<std::set<uint16_t>>& txKeysClientGroups)
```
toly-kournik, line 74: what's tx?

yontyon: shortcut for transactions

PR #1773, opened 2021-08-04
[Client pool] Add with_cre flag

Liorgal28:
The flag for enabling cre is presented in this PR.
Assuming we have with_cre enabled, we will remove one client from the pool.


yontyon (approved):
LGTM!

PR #1774, opened 2021-08-04
Kvbc: Protobuf is not gRPC

robem:
The concord-kvbc-proto target contains protobuf messages only and hence pulls
in a protobuf library as a dependency. It doesn't need gRPC because there are
no RPCs. In addition, the guarding `USE_GRPC` variable doesn't apply in this
case and therefore it can be removed as well.


toly-kournik (approved):
LGTM!
I would also add an option to disable protobuf in kvbc or even better to move it from kvbc.

PR #1775, opened 2021-08-04
TRS: Don't block forever on updates

robem:
In general, the TRS is waiting for new updates unconditionally. That can consume a lot of resources if we don't get new updates and if the TRC gets impatient and retries. This PR introduces conditional blocking for a fixed amount of time in order to check if we can return early. See commit messages for details.


shgandhi (approved):
These are great improvements, thanks for the change @robem, LGTM!

PR #1776, opened 2021-08-05
Primary sends its own signature with PreProcessResutMsg

tdimitrov:
When result authentication (aka non repudiation) is enabled for
PreExecution the primary should send its own result hash signature
alongside the signatures of the replicas which have preprocessed the
client request.

Also when PreProcessResultMsg is verified by non primaries, f+1
signatures should be expected instead of just f.

Primary signs the result hash just before sending the PreProcessResult
message because it is not needed earlier. Furthermore  the computation
can be skipped if the PreProcessing is not successful.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -1127,6 +1127,7 @@ void PreProcessor::finalizePreProcessing(NodeIdType clientId, uint16_t reqOffset
       // controlled by the replica while all PreProcessReply messages get released here.
       if (ReplicaConfig::instance().preExecutionResultAuthEnabled) {
         auto sigsList = reqProcessingStatePtr->getPreProcessResultSignatures();
+        sigsList.resize(numOfRequiredReplies());
         auto sigsBuf = PreProcessResultSignature::serializeResultSignatureList(sigsList);
         preProcessMsg = make_unique<PreProcessResultMsg>(clientId,
                                                          reqSeqNum,
@@ -1167,7 +1168,7 @@ void PreProcessor::finalizePreProcessing(NodeIdType clientId, uint16_t reqOffset
   if (batchedPreProcessEnabled_) batchEntry->finalizeBatchIfCompleted();
 }
 
-uint16_t PreProcessor::numOfRequiredReplies() { return myReplica_.getReplicaConfig().fVal; }
+uint16_t PreProcessor::numOfRequiredReplies() { return myReplica_.getReplicaConfig().fVal + 1; }
 
 // This function should be always called under a reqEntry->mutex lock
 bool PreProcessor::registerRequest(const string &batchCid,
@@ -1208,7 +1209,8 @@ bool PreProcessor::registerRequest(const string &batchCid,
       ongoingReqBatches_[clientId]->registerBatch(batchCid, batchSize);
   }
   if (!reqEntry->reqProcessingStatePtr) {
-    reqEntry->reqProcessingStatePtr = make_unique<RequestProcessingState>(numOfReplicas_,
+    reqEntry->reqProcessingStatePtr = make_unique<RequestProcessingState>(myReplicaId_,
+                                                                          numOfReplicas_,
                                                                           batchCid,
                                                                           clientId,
                                                                           reqOffsetInBatch,
--- a/bftengine/src/preprocessor/RequestProcessingState.hpp
+++ b/bftengine/src/preprocessor/RequestProcessingState.hpp
@@ -29,7 +29,8 @@ using ReplicaIdsList = std::vector<ReplicaId>;
 
 class RequestProcessingState {
  public:
-  RequestProcessingState(uint16_t numOfReplicas,
+  RequestProcessingState(ReplicaId myReplicaId,
+                         uint16_t numOfReplicas,
                          const std::string& batchCid,
                          uint16_t clientId,
                          uint16_t reqOffsetInBatch,
@@ -97,6 +98,7 @@ class RequestProcessingState {
 
   // The use of the class data members is thread-safe. The PreProcessor class uses a per-instance mutex lock for
   // the RequestProcessingState objects.
+  const ReplicaId myReplicaId_;
   const uint16_t numOfReplicas_;
   const std::string batchCid_;
   const uint16_t clientId_;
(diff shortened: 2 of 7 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1110,6 +1110,17 @@ void PreProcessor::cancelPreProcessing(NodeIdType clientId, uint16_t reqOffsetIn
   }
 }
 
+PreProcessResultSignature signPrimaryPreExecutionResultHash(const ReplicaId &myId,
```
yuliasherman, line 1113: A native place for this function is RequestProcessingState, where we deal with hashes and signatures.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1127,6 +1138,9 @@ void PreProcessor::finalizePreProcessing(NodeIdType clientId, uint16_t reqOffset
       // controlled by the replica while all PreProcessReply messages get released here.
       if (ReplicaConfig::instance().preExecutionResultAuthEnabled) {
         auto sigsList = reqProcessingStatePtr->getPreProcessResultSignatures();
+        // append primary's signature
+        sigsList.push_back(
```
yuliasherman, line 1142: In case the primary signature is a part of RequestProcessingState::preProcessingResultHashes_, this is not needed.

bftengine/src/preprocessor/RequestProcessingState.hpp
```diff
@@ -53,6 +53,7 @@ class RequestProcessingState {
   PreProcessingResult definePreProcessingConsensusResult();
   const char* getPrimaryPreProcessedResult() const { return primaryPreProcessResult_; }
   uint32_t getPrimaryPreProcessedResultLen() const { return primaryPreProcessResultLen_; }
+  const concord::util::SHA3_256::Digest& getPrimaryPreProcessedResultHash() { return primaryPreProcessResultHash_; }
```
yuliasherman, line 56: And also this.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1110,6 +1110,17 @@ void PreProcessor::cancelPreProcessing(NodeIdType clientId, uint16_t reqOffsetIn
   }
 }
 
+PreProcessResultSignature signPrimaryPreExecutionResultHash(const ReplicaId &myId,
+                                                            const concord::util::SHA3_256::Digest &hash) {
+  auto sm = SigManager::instance();
```
arc-vmware, line 1115: Should we add a check that this function will get executed if myId is of primary in the given view. If not then just return, so that any accidental call by other replica can be avoided.

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
```diff
@@ -57,8 +57,8 @@ std::optional<std::string> PreProcessResultMsg::validatePreProcessResultSignatur
   const auto [buf, buf_len] = getResultSignaturesBuf();
   auto sigs = preprocessor::PreProcessResultSignature::deserializeResultSignatureList(buf, buf_len);
 
-  // f because the primary also executes the request (check PreProcessor::numOfRequiredReplies())
-  const auto expectedSignatureCount = fVal;
+  // f+1 signatures are required
+  const auto expectedSignatureCount = fVal + 1;
   if (sigs.size() < static_cast<uint16_t>(expectedSignatureCount)) {
```
arc-vmware, line 62: we should change sigs.size() < static_cast<uint16_t>(expectedSignatureCount) to sigs.size() != static_cast<uint16_t>(expectedSignatureCount) if we are sure that f+1 signatures will come.
____________________________________________________________________________________
But in case the number of signatures are greater than f+1, then we should add more logic for the following scenario [Read below if there is any chance that sigs.size() > f + 1, else ignore]:
lets say in one PreProcessResultMsg there are 7 signatures from s1, s2, to s7. Here the number of signatures is greater than 3. Here f is 2, so we need 3 correct signatures out of 7. But in case we get 7, then in the verification loop we have to then check is out of 7 atleast 3 have arrived or not. The moment we get 3 correct signatures we should break the loop and return success.
case1: s1, s2 are bad. s3, s4,s5,s6 and s7 are good. The loop should run for s1 to s5. Ignore that s1 and s2 are failed and return success.
case 2: s6 and s7 are bad and rest are good. The loop should run for s1 to s3 and return success.
case 3: s2 and s6 are bad and rest are good. The loop should run for s1 to s4. Returns success.
A failure case will be, when loop don't even get 3 good signatures.

arc-vmware (changes requested):

bftengine/src/preprocessor/PreProcessor.cpp
tdimitrov: Good point! I felt this function is out of place but I wanted to defer the signature calculation. However considering that 99% of the runs will be happy cases - my optimisation makes no sense.

Please check my second commit. It implements your suggestion.

bftengine/src/preprocessor/RequestProcessingState.hpp
tdimitrov: Removed.

yuliasherman (approved):
LGTM!

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
tdimitrov: This is an interesting point, @arc-vmware. I think you are right - we should check that the message contains exactly f+1 signatures.

In practice the primary considers the PreProcessing completed once it receives f replies with result hash matching its own. So we should always receive f+1 signatures (f from replicas and 1 from primary).

I think the primary should not send invalid signatures at all - this will be an indication for a serious error.

Also a malicious primary can launch a DoS attack against the replicas by sending a PreProcessReply with a huge number of signatures. We shouldn't verify more signatures than we need.

@yuliasherman what do you think? Should check if sigs.size() == expectedSigSize?

arc-vmware (approved):
LGTM!

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
yuliasherman: Agree

bftengine/src/preprocessor/PreProcessor.cpp
tdimitrov: After addressing Yulia's comments this code is now gone.

PR #1777, opened 2021-08-05
Store full blockData in the RO storage instead of blockNum

nkumar04:
This PR includes changes to store full block data that has blockId, epochNum wedgePoint, reconfigurationCommand instead of only blockNum. Also added basic apollo test for the same


tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -1169,6 +1169,108 @@ async def test_addRemoveStatusError(self, bft_network):
             status = cmf_msgs.ReconfigurationResponse.deserialize(r)[0]
             assert status.response.error_msg == 'key_not_found'
             assert status.success is False
+    @with_trio
+    @with_bft_network(start_replica_cmd=start_replica_cmd_with_object_store_and_ke, num_ro_replicas=1, rotate_keys=True,
```
yontyon, line 1173: I would like us to have (maybe in a different PR) a real scale down/up with a read-only replica.

yontyon (approved):
LGTM!
Thanks

PR #1780, opened 2021-08-05
 Replica side support for clients scaling

yontyon:
In this PR we add the replica side support for client scaling requests.
The procedure is using the CRE in the process:
1. The operator sends a clientsAddRemove request that contains the configuration descriptor
2. The replicas write this update to the blockchain
3. The CRE pulls the update and executes the request
4. The CRE writes an update to the ledger
5. The operator can now get the status that says that this CRE was able to execute the request

### DIFF
--- a/tests/simpleKVBC/TesterCRE/main.cpp
+++ b/tests/simpleKVBC/TesterCRE/main.cpp
@@ -203,7 +203,8 @@ class ClientsAddRemoveHandler : public IStateHandler {
     return true;
   }
   logging::Logger getLogger() {
-    static logging::Logger logger_(logging::getLogger("concord.client.reconfiguration.testerCre.PublicKeyExchange"));
+    static logging::Logger logger_(
+        logging::getLogger("concord.client.reconfiguration.testerCre.ClientsAddRemoveHandler"));
     return logger_;
   }
 };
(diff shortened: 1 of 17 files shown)


tests/simpleKVBC/TesterCRE/main.cpp
```diff
@@ -177,6 +177,38 @@ class PublicKeyExchangeHandler : public IStateHandler {
...
+    concord::messages::ClientsAddRemoveCommand command =
+        std::get<concord::messages::ClientsAddRemoveCommand>(crep.response);
+
+    concord::messages::ReconfigurationRequest rreq;
+    concord::messages::ClientsAddRemoveUpdateCommand creq;
+    creq.config_descriptor = command.config_descriptor;
+    rreq.command = creq;
+    std::vector<uint8_t> req_buf;
+    concord::messages::serialize(req_buf, rreq);
+    out = {req_buf, [this, &command]() {
```
nkumar04, line 200: command is a variable on stack. i am not sure if we should capture this by reference in the lambda

nkumar04 (approved):
LGTM!
Thanks!

tests/simpleKVBC/TesterCRE/main.cpp
yontyon: fixed

PR #1788, opened 2021-08-09
Add check for unique signatures in PreProcessResultMsg validate function

tdimitrov:
At the moment validatePreProcessResultSignatures() doesn't check if the
signatures attached to the message are from unique replicas. This means
that a malicious primary can send a single signature f+1 times and it
will fool a non-primary replica that the result is valid.

This patch adds check that all signatures are from unique replicas.
SigManager has got public keys for each replica so it is guaranteed that
the message is uniquely signed by f+1 replicas.

### DIFF
--- a/bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
+++ b/bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
@@ -70,7 +70,15 @@ std::optional<std::string> PreProcessResultMsg::validatePreProcessResultSignatur
   }
 
   auto hash = concord::util::SHA3_256().digest(requestBuf(), requestLength());
+  std::unordered_set<bftEngine::impl::NodeIdType> seen_signatures;
   for (const auto& s : sigs) {
+    // insert returns std::pair<iterator, bool>. The bool indicates if the element was created or it was already in the
+    // set and no insertion was performed. The latter case indicates that we already have got a signature from this
+    // replica.
+    if (!seen_signatures.insert(s.sender_replica).second) {
+      return "PreProcessResult signatures validation failure - got more than one signatures with the same sender id";
+    }
+
     bool verificationResult = false;
     if (myReplicaId == s.sender_replica) {
       std::vector<char> mySignature(sigManager_->getMySigLength(), '\0');
--- a/bftengine/src/preprocessor/tests/messages/PreProcessResultMsg_test.cpp
+++ b/bftengine/src/preprocessor/tests/messages/PreProcessResultMsg_test.cpp
@@ -49,6 +49,17 @@ bftEngine::impl::SigManager* createSigManagerWithSigning(
                           replicasInfo);
 }
 
+// Some default message params used through the test
+struct MsgParams {
+  NodeIdType senderId = 5u;
+  uint64_t reqSeqNum = 100u;
+  const char result[12] = {"result body"};
+  const uint64_t requestTimeoutMilli = 0;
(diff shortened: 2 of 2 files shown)


yuliasherman (approved):
LGTM

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
```diff
@@ -70,7 +70,15 @@ std::optional<std::string> PreProcessResultMsg::validatePreProcessResultSignatur
   }
 
   auto hash = concord::util::SHA3_256().digest(requestBuf(), requestLength());
+  std::unordered_set<bftEngine::impl::NodeIdType> seen_signatures;
   for (const auto& s : sigs) {
+    const auto [it, created] = seen_signatures.insert(s.sender_replica);
```
arc-vmware, line 75: Very minor comment.
Instead of keeping an unused variable can we do
if(seen_signature.find(s.sender_replica) != seen_signature.end()) {
   // its present logic
} else {
   seen_signatures.insert(s.sender_replica);
}
This way we will avoid the unused variable.

Or we can also do the following to avoid the find:
if(!seen_signatures.insert(s.sender_replica).second) {
     // do the error logic. 
}

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
```diff
@@ -70,7 +70,15 @@ std::optional<std::string> PreProcessResultMsg::validatePreProcessResultSignatur
...
+    (void)it;  // unused variable
+
+    if (!created) {
```
arc-vmware, line 78: I have question:
Consider that 2 signatures are same. Then this is an error case and it will return after that.
Now will the verifySig() can return true for this signed hash for another replica. 
I think same signature with different ID will not be an issue.
But if the tuple s.signature and s.sender_replica [i.e both the signature and replica id] are same then this will be an issue.
So say a byzantine primary sends 7 same signature for his own hash with same sender_replica ids (ie. his own replica id). Then only this will be an issue.
In rare (rarest) case due to collision, signature of two different replica can also be same. So in such case we will get a false issue due to non checking of replicaid alongwith the signatures.

arc-vmware (changes requested):
Thanks @tdimitrov for the change. This is a very nice usecase to fix.

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
tdimitrov: I wanted to avoid searching and then inserting in the set, but your second idea is very good. I'll do it that way.

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
tdimitrov: Sorry @arc-vmware but I don't follow.
We keep the sender id in the set. The purpose of this check is to ensure that all sender IDs are unique. After that we verify each signature with the public key of the corresponding replica.
Even if the signatures are the same (which should not happen) everything will be fine as long as the signature verification works.
As far as I understand - this logic should be fine. What I am missing?

arc-vmware: Yes @tdimitrov , my bad, I assumed that signatures are checked. But as you are checking replica id, its fine. Thanks for correcting me.

arc-vmware (approved):

PR #1789, opened 2021-08-09
Use log4cplus.properties In Client Service

irisk29:
Added watch thread for the client service logger.
The path for the log4cplus.properties is defined by an environment variable.

### DIFF
--- a/client/clientservice/src/main.cpp
+++ b/client/clientservice/src/main.cpp
@@ -32,8 +32,9 @@ namespace po = boost::program_options;
 
 const static int kLogConfigRefreshIntervalInMs = 60 * 1000;
 
-const static char* GetLog4CplusConfigLocation() {
+const static char* getLog4CplusConfigLocation() {
   auto log_location = std::getenv("LOG4CPLUS_CONFIGURATION");
+  if (!log_location) std::cerr << "ClientService log4cplus configuration file was not set" << std::endl;
   return log_location ? log_location : "LOG4CPLUS_CONFIGURATION_NOT_SET";
 }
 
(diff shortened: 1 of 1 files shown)


client/clientservice/src/main.cpp
```diff
@@ -30,6 +30,13 @@ using concord::client::concordclient::ConcordClientConfig;
 
 namespace po = boost::program_options;
 
+const static int kLogConfigRefreshIntervalInMs = 60 * 1000;
+
+const static char* GetLog4CplusConfigLocation() {
```
robem, line 35: Let's make this more user-friendly. If the environment variable isn't set then let's print a warning to `std::cerr`. Also, function names start with lower-case.

client/clientservice/src/main.cpp
```diff
@@ -51,9 +58,8 @@ po::variables_map parseCmdLine(int argc, char** argv) {
 }
 
 int main(int argc, char** argv) {
-  // TODO: Use config file and watch thread for logger
   auto logger = logging::getLogger("concord.client.clientservice.main");
-
+  LOG_CONFIGURE_AND_WATCH(GetLog4CplusConfigLocation(), kLogConfigRefreshIntervalInMs);
```
robem, line 62: We should configure the logger *before* we get it.

robem (changes requested):

client/clientservice/src/main.cpp
irisk29: Done.

client/clientservice/src/main.cpp
irisk29: Fixed.

client/clientservice/src/main.cpp
```diff
@@ -72,6 +78,7 @@ int main(int argc, char** argv) {
   auto metrics = std::make_shared<concordMetrics::Aggregator>();
   concord_client->setMetricsAggregator(metrics);
   ClientService service(std::move(concord_client));
+  LOG_CONFIGURE_AND_WATCH(getLog4CplusConfigLocation(), kLogConfigRefreshIntervalInMs);
```
robem, line 81: We are retrieving the logger in line 62 and we configure log4cplus in line 81. Shouldn't we configure it before we get the logger or does it not matter? In any case, those two lines should be close together.

irisk29: I checked - it works either way. I moved it to be before line 62.

robem (approved):

PR #1740, opened 2021-07-29
Migrate SKVBC to use CMF messages

upshaw-alex:
This pull request modifies SimpleKVBC (implemented in `tests/simpleKVBC`) to use the CMF messages recently introduced in `tests/simpleKVBC/cmf/skvbc_messages.cmf` (see https://github.com/vmware/concord-bft/pull/1614) for client requests and replies to them, replacing SimpleKVBC's existing ad hoc message implementation which effectively used packed C structs.

This pull request also modifies Apollo's SimpleKVBC client implementation in `tests/apollo/util/skvbc.py` to use these CMF messages and remain compatible with the SimpleKVBC replicas.

I have validated all tests with names of the form `skvbc_*` appear to pass on my machine with these changes.

PR #1790, opened 2021-08-09
Revert 1740

yontyon:
https://github.com/vmware/concord-bft/pull/1740 causes issues with bumping concordbft.
Hence we revert it until it will be fixed
cc: @upshaw-alex


yuliasherman (commented):
LGTM

PR #1791, opened 2021-08-09
Fixing intermittent byzantine_primary_preexecution

arc-vmware:
This PR will help to identify the real
reasons for Timeouts in testcases.
We are seeing many testcases are failing
in a random manner. This is happenig
generally during high load.

### DIFF
--- a/tests/apollo/test_skvbc_byzantine_primary_preexecution.py
+++ b/tests/apollo/test_skvbc_byzantine_primary_preexecution.py
@@ -22,8 +22,8 @@ from util.skvbc_history_tracker import verify_linearizability
 import util.eliot_logging as log
 
 SKVBC_INIT_GRACE_TIME = 2
-BATCH_SIZE = 4
-NUM_OF_SEQ_WRITES = 25
+BATCH_SIZE = 5
+NUM_OF_SEQ_WRITES = 10
 
 def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="10000"):
     """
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc_byzantine_primary_preexecution.py
```diff
@@ -25,7 +25,7 @@
 BATCH_SIZE = 4
 NUM_OF_SEQ_WRITES = 25
 
-def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="10000"):
+def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="6000000"):
```
teoparvanov, line 28: Seriously,  1,5h ? 😄

arc-vmware: Thanks @teoparvanov for reviewing the changes.
Yes, this I kept to test the max limit. Now the test case at max took 40 - 50 secs, at peak load today and during that time this high expiry prevented any of the view change. Tomorrow I will make it to 10 mins and check. If it works fine in the peak load, then I will keep that 10 mins. Although keeping it for 1 hr 40 mins is not a problem.

teoparvanov: I don't understand what you're trying to achieve with this... The test expects a view change to happen, but you set a large timeout,  effectively disabling the view change protocol? Did you check the replica logs to verify that a view change is triggered correctly?

arc-vmware: @teoparvanov as discussed, increasing view change timeout in milli sec is not needed. We are reducing the total number of requests sent. Thanks for your review.

teoparvanov (approved):
LGTM, thanks Amitava!

HristoStaykov (approved):
LGTM, thanks @arc-vmware.

PR #1793, opened 2021-08-09
Add building crypto++ as a thirdparty inside a build tree .

toly-kournik:
Detach threshign from Crypto.hpp .
Fixed thirdparty relic and gmp builds .


yontyon (commented):
Wouldn't it mean that we will build the dependencies on every build?
If so this may extend the build time a lot.

yontyon (approved):

PR #1795, opened 2021-08-09
Docker update to version 0.28

robem:
So far we have installed `clang`, `clang-format-9`, and `clang-tidy-10`. Let's
sync all of them to the same version. We can do further compiler upgrades in a
follow-up. This PR focuses on minimal changes to sync with closed-source.
Note, `clang-tidy-9` has a bug with respect to `if constexpr` statements.
Therefore, we leave it for now.


f-squirrel (approved):
LGTM!
If we already do this, maybe we can upgrade to 10.1 or higher?

robem:
> If we already do this, maybe we can upgrade to 10.1 or higher?
The objective of this change is to sync closed- and open-source. I chose clang-9 because it is the path of least resistance to get the sync done. Let's do a follow-up change to upgrade to newer versions.

robem:
@dartdart26 you recent `d677e3821fd52d7deaa5e75a799cba58671cdfa8` changed the invocation for clang-format. This PR reverts your change given that we properly symlink all executables. Does this work for you?

dartdart26:
> @dartdart26 you recently `d677e3821fd52d7deaa5e75a799cba58671cdfa8` changed the invocation for clang-format. This PR reverts your change given that we properly symlink all executables. Does this work for you?

Yes, sure. I think the way it is done here is much better. Thanks!

PR #1797, opened 2021-08-10
Adding epoch number to bft messages 

nkumar04:
In this PR, we are adding epoch number to replica's internal messages. Epoch is the lifecycle of a system. Change in epoch number indicates that we started a new system on top of the old
one. For example, on scale-up/down to the cluster, we essentally resatrt the system with new epoch number. So, replicas need to ignore consensus messages from a replica in previous epoch(with some exceptions). Summary of the changes in this PR:
- Added epoch number of the protocol messages
- Added validity check to ignore messages from different epoch number
- Added relevant checks in checkpoint message handler so that replicas in previous epoch can get states via ST


bftengine/include/bftengine/ReconfigurationCmd.hpp
```diff
@@ -106,6 +106,21 @@ class ReconfigurationCmd : public bftEngine::ResPagesClient<ReconfigurationCmd,
     metrics_.UpdateAggregator();
     return true;
   }
+  std::optional<uint64_t> getReconfigurationCommandEpochNumber() {
```
yontyon, line 109: This method will be only used by the read-only replica?

bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -101,19 +102,27 @@ void ReadOnlyReplica::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
...
 
   // previous CheckpointMsg from the same sender
   auto pos = tableOfStableCheckpoints.find(msgGenReplicaId);
-  if (pos != tableOfStableCheckpoints.end() && pos->second->seqNumber() >= msgSeqNum) return;
+  if (pos != tableOfStableCheckpoints.end() &&
+      (pos->second->seqNumber() >= msgSeqNum || msgEpochNum < replicasLastKnownEpochVal))
+    return;
   if (pos != tableOfStableCheckpoints.end()) delete pos->second;
   CheckpointMsg *x = new CheckpointMsg(msgGenReplicaId, msgSeqNum, msgDigest, msgIsStable);
+  x->setEpochNumber(msgEpochNum);
```
yontyon, line 125: This is to handle the case where msgEpochNumber > replicaLastKnownEpochVal?

bftengine/src/bftengine/messages/PartialProofsSet.cpp
```diff
@@ -24,6 +24,7 @@
 #include "SimpleThreadPool.hpp"
 #include "ReplicaConfig.hpp"
 #include "CryptoManager.hpp"
+#include "EpochManager.hpp"
```
yontyon, line 27: Are we using the EpochManager here?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1841,13 +1842,15 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
 
   bool askForStateTransfer = false;
 
-  if (msgIsStable && msgSeqNum > lastExecutedSeqNum) {
+  if (msgIsStable && (msgSeqNum > lastExecutedSeqNum || msgEpochNum >= getSelfEpochNumber())) {
     auto pos = tableOfStableCheckpoints.find(msgGenReplicaId);
-    if (pos == tableOfStableCheckpoints.end() || pos->second->seqNumber() <= msgSeqNum) {
+    if (pos == tableOfStableCheckpoints.end() || pos->second->seqNumber() <= msgSeqNum ||
+        msgEpochNum >= getSelfEpochNumber()) {
```
yontyon, line 1848: Just to understand
1. The checkpoint message is the only one allowed to be received if the message epoch >= replica epoch, right?
2. We still start state transfer only based on sequence number?

yontyon (approved):
Great PR!
It is approved, but I have some small questions and clarification I need.
Please address them.
Thanks!

yontyon:
Also, a sanity check Apollo test could do the following:
1. run like 100 requests
2. stop one replica
3. start a new epoch - without removing metadata (using unwedge)
4.  at this point all live are in sequence number 300 and epoch 1 
4. start the crashed replica (the new replica thinks it is in sequence number 100 and epoch 0)
5. make sure the crashed replica is not getting any new requests for missing info etc. (and basically stays outside the cluster)

If you have time, this would be great to add this kind of test in a different PR

Thanks

bftengine/src/bftengine/ReadOnlyReplica.cpp
nkumar04: yes

bftengine/src/bftengine/ReplicaImp.cpp
nkumar04: yes (1) is True, i the reason i did not add this is valid method because it is called for RO replica as well and there is no self epoch number for RO replica

bftengine/include/bftengine/ReconfigurationCmd.hpp
nkumar04: yes

PR #1798, opened 2021-08-10
Start new epoch on metadata removal

yontyon:
According to the rule of "start a new epoch on metadata removal," we added logic to start a new epoch when metadata is removed using kv_blockchain_editor.
Notice that if the replicas are not synced this will break the blockchain consistency

### DIFF
--- a/kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
+++ b/kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
@@ -13,6 +13,7 @@
 
 #pragma once
 
+#include "kvbc_key_types.hpp"
 #include "db_editor_common.hpp"
 #include "categorization/kv_blockchain.h"
 #include "execution_data.cmf.hpp"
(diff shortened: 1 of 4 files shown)


nkumar04 (approved):
LGTM!

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -733,7 +734,39 @@ struct RemoveMetadata {
...
+        const auto &data = std::get<categorization::VersionedValue>(*value).data;
+        ConcordAssertEQ(data.size(), sizeof(uint64_t));
+        last_executed_sn = concordUtils::fromBigEndianBuffer<uint64_t>(data.data());
+      }
+    }
+    epoch += 1;
+    std::string epoch_str = concordUtils::toBigEndianStringBuffer(epoch);
+    concord::kvbc::categorization::VersionedUpdates ver_updates;
+    ver_updates.addUpdate(std::string{kvbc::keyTypes::reconfiguration_epoch_key}, std::move(epoch_str));
+    std::string sn_str = concordUtils::toBigEndianStringBuffer(last_executed_sn);
```
dartdart26, line 761: Do we need to add the last executed sequence number with the same value again in the new block?

dartdart26 (approved):
Looks good, thank you!

PR #1801, opened 2021-08-10
Refactor operator keys

yontyon:
In this PR we added functionality for both RSA and ECDSA keys to crypto_utils.
This will enable us to use only one verifier in the operator as well as eventually refactor out the Crypto class from concordbft/src

### DIFF
--- a/util/src/crypto_utils.cpp
+++ b/util/src/crypto_utils.cpp
@@ -26,7 +26,7 @@ class ECDSAVerifier::Impl {
   std::unique_ptr<ECDSA<ECP, SHA256>::Verifier> verifier_;
 
  public:
-  Impl(ECDSA<ECP, SHA256>::PublicKey publicKey) {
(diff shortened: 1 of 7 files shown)


util/src/crypto_utils.cpp
```diff
@@ -14,60 +14,232 @@
...
+      StringSource pub_str(key_pair.second, true, new HexDecoder());
+      ECDSA<ECP, SHA256>::PublicKey pub;
+      pub.Load(pub_str);
+      StringSink pub_string_sink(out.second);
+      PEM_Save(pub_string_sink, pub);
+    }
     return out;
   }
 
   std::pair<std::string, std::string> generateRsaKeyPairs(uint32_t sig_length, KeyFormat fmt) {
```
toly-kournik, line 188: generateRsaKeyPair

toly-kournik (approved):
LVGTM!

PR #1803, opened 2021-08-11
Do not prune on startup if pruning disabled

dartdart26:
Do not call `pruneOnStartup()` on the ST reconfiguration SM if pruning
is disabled in configuration.


tdimitrov (approved):

yontyon (approved):
LGTM!

PR #1807, opened 2021-08-11
State transfer bug fixes and minor flow control improvments

glevkovich:
Here we add multiple bug fixes and improvements observed during CI runs and manual testing:

1. BCStateTran: change the way destination handles lastInBatch flag
2. BCStateTran: add lastInBatch flag when source reply FetchResPagesMsg
2. BCStateTran: fix false assertion bug during consistency checks
3. BCStateTran: fix a bug in the source while fetching missing blocks


toly-kournik (approved):
LGTM!

PR #1808, opened 2021-08-11
Fixes for building without log4cplus.

HristoStaykov:
Remove direct includes of log4cplus headers.
Wrap logger used in macros in braces.


toly-kournik (approved):
LGTM

PR #1812, opened 2021-08-12
Cre refactoring

yontyon:
In this PR we refactor and improve the CRE code.
The main change is, instead of fetching only the next state update, we fetch all the updates at once. 
This solves a bug with recovering after CRE crash if there were multiple state changes in the meantime

### DIFF
--- a/client/reconfiguration/src/poll_based_state_client.cpp
+++ b/client/reconfiguration/src/poll_based_state_client.cpp
@@ -34,20 +34,20 @@ concord::messages::ReconfigurationResponse PollBasedStateClient::sendReconfigura
       bft::client::WriteConfig write_config{request_config, bft::client::LinearizableQuorum{}};
       rep = bftclient_->send(write_config, std::move(msg));
     }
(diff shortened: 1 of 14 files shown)


client/reconfiguration/src/poll_based_state_client.cpp
```diff
@@ -63,20 +63,35 @@ PollBasedStateClient::PollBasedStateClient(bft::client::Client* client,
       last_known_block_{last_known_block},
       sn_gen_(bft::client::ClientId{id}) {}
 
-State PollBasedStateClient::getStateUpdate(uint64_t lastKnownBlockId) const {
-  concord::messages::ClientReconfigurationStateRequest creq{id_, lastKnownBlockId};
+std::vector<State> PollBasedStateClient::getStateUpdate(bool& succ) const {
```
nkumar04, line 66: nit: please consider bool as return type and "std::vector<State>&" as argument

nkumar04 (approved):
LGTM!
Thanks!

PR #1813, opened 2021-08-13
HdrHistogram thirdparty lib

toly-kournik:
A part of an effort of making concord-bft library being built in a standard way on multiple platforms.
This will make it more development friendly and attractive to users outside vmware.
Add an option to compile a HdfHistogram thirdparty library as a part of a build using ExternalProject_Add().

Docker build stays intact!


logging/include/Logging.hpp
```diff
@@ -67,7 +67,7 @@ class LoggerImpl {
 
     // clang-format off
     std::cout << std::put_time(std::localtime(&cur_time.tv_sec), "%FT%T.") << (int)cur_time.tv_usec / 1000
-       << "|" << LoggerImpl::LEVELS_STRINGS[l].c_str()
+       << "|" << LoggerImpl::LEVELS_STRINGS[l]
```
f-squirrel, line 70: Is it related to the HDR historgram?

f-squirrel (changes requested):
Did we reach a consensus on this approach?

yontyon (approved):
LGTM!

robem:
@toly-kournik Would you mind providing a commit message which explains what this PR is changing and why?

logging/include/Logging.hpp
toly-kournik: This is related to a compilation on MacOS with a most recent clang compiler (12.05)

PR #1814, opened 2021-08-15
TS: Introduce metrics

f-squirrel:
TS: Introduce metrics 

This commit introduces the most valuable metrics for Time Service:

* time_service.ill_formed_preprepare - counter of PrePrepares without
timestamp
* time_service.soft_limit_reached_counter - counter of PrePrepares which
timestamps reached soft limit
* time_service.hard_limit_reached_counter - counter of PrePrepares which
timestamps reached hard limit
* time_service.new_time_is_less_or_equal_to_previous - counter of
synthetically increased timestamps

+ Unit tests


ukg19 (approved):
Great work @f-squirrel and thanks for offline clarification . LGTM

PR #1816, opened 2021-08-16
Use appropriate units for histograms

yuliasherman:
Increase histogram reporting units from nano to micro where applicable


toly-kournik (approved):
LGTM!

PR #1817, opened 2021-08-16
Fix non-const lvalue arguments in ThreadPool

dartdart26:
Make sure that non-const lvalue arguments are copied as we don't
transport references from async() to the callable. In general, users can
either copy or move arguments. In order to achieve that, we add a
missing std::decay_t when declaring the packaged task parameters type.

Add a comment about intentional lack of support for references. We note,
though, that std::c/ref() can explicitly be used if references are
required.

Add unit tests showing that pointers and non-const lvalues are copied
and that std::ref() works as expected.

Thanks to @arc-vmware for noticing this issue!

### DIFF
--- a/util/include/thread_pool.hpp
+++ b/util/include/thread_pool.hpp
@@ -58,12 +58,14 @@ class ThreadPool {
 
  public:
   // Executes the passed function (or any callable) in a pool thread. Returns a future to the result.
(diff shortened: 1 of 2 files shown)


arc-vmware (approved):
Thanks @dartdart26
LGTM.

PR #1818, opened 2021-08-16
State transfer fix assert failure and modify flow control on storage error

glevkovich:
Due to issues seen in RO replica, where the storage layer of RO replica was not functioning, I've found an assertion bug, and modified flow control to avoid sending FetchBlocksMsg to source when the context pool (ioPool) is empty.

Commits:
1) BCStateTran: fix wrong assertion
2) BCStateTran: avoid sending FetchBlocksMsg when ioPool is empty


toly-kournik (approved):
LGTM

PR #1820, opened 2021-08-16
BCStateTran: when stopped, wait for all jobs to gracefully end

glevkovich:
There might be some threads still working, and for a destination ST
replica we would also like to finalize all jobs to advance and persist
the last block. Call finalizePutblockAsync before stopping.


toly-kournik (approved):
LGTM

PR #1823, opened 2021-08-17
Reset wedge state on state transfer

yontyon:
When a crashed replica has done with state transfer and decides to unwedge, it needs to reset the wedge state.


nkumar04 (approved):
LGTM!
Thanks!

PR #1824, opened 2021-08-17
WIP:Add flag to disable client callback

Liorgal28:
This MR adds a flag to enable/disable the client request callback.


robem:
@Liorgal28 instead of using a flag, why don't we use a default callback `[]() { return; }` or so? The less configuration values the better :)
 
Also, you're changing `SECOND_LEG_CID_LEN`. Do you mind explaining why?

Liorgal28:
This is a PR to test if this change caused the ledger API crashes.
The second leg length has been changed in the new spider app.

PR #1827, opened 2021-08-17
Don't handle PruneRequest on ST if pruning disabled

dartdart26:
If pruning is disabled in configuration, do not handle stored
PruneRequest commands when State Transfer completes.

Also, make sure that pruneOnStartup() doesn't do anything if pruning is
disabled in configuration.


yontyon (approved):
LGTM! 
Thanks!

nkgeorgiev (approved):
LGTM! Thanks

PR #1828, opened 2021-08-17
Remove the usage of bftengine/src/Crypto from bftclient

yontyon:
As part of the crypto functionality refactoring, we remove the bftengine/src/Crypto usage of bftclient

### DIFF
--- a/bftclient/test/bft_client_api_tests.cpp
+++ b/bftclient/test/bft_client_api_tests.cpp
@@ -97,10 +97,13 @@ class ClientApiTestParametrizedFixture : public ClientApiTestFixture,
       else
         signature_data--;  // corrupt the signature
     }
-    bool sig_verification_result = transaction_verifier_->verify(reinterpret_cast<const char*>(request_data),
-                                                                 request_header->requestLength,
-                                                                 reinterpret_cast<const char*>(signature_data),
-                                                                 sig_len);
+    std::string data;
+    data.resize(request_header->requestLength);
+    std::string sig;
+    sig.resize(request_header->reqSignatureLength);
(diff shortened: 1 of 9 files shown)


bftclient/test/bft_client_api_tests.cpp
```diff
@@ -97,10 +97,13 @@ class ClientApiTestParametrizedFixture : public ClientApiTestFixture,
...
     }
-    bool sig_verification_result = transaction_verifier_->verify(reinterpret_cast<const char*>(request_data),
-                                                                 request_header->requestLength,
-                                                                 reinterpret_cast<const char*>(signature_data),
-                                                                 sig_len);
+    std::string data;
+    data.resize(request_header->requestLength);
+    std::string sig;
+    sig.resize(request_header->reqSignatureLength);
+    std::memcpy(data.data(), reinterpret_cast<const char*>(request_data), request_header->requestLength);
```
toly-kournik, line 104: Although it's a test, I think it's a bad practice to initialize a string by std::memcpy.
A better way is to use a constructor:
`std::string s{reinterpret_cast<const char*>(request_data), request_header->requestLength};`

util/include/crypto_utils.hpp
```diff
@@ -20,19 +20,22 @@ enum class KeyFormat : std::uint16_t { HexaDecimalStrippedFormat, PemFormat };
 class IVerifier {
  public:
   virtual bool verify(const std::string& data, const std::string& sig) = 0;
+  virtual uint32_t signatureLength() = 0;
```
toly-kournik, line 23: const

toly-kournik (approved):
LGTM!

PR #1831, opened 2021-08-18
TRS: Transition from legacy events to event groups

robem:
This PR introduces transitioning into event groups when a legacy event subscription request is present. In addition, we update the kvb app filter and thin replica server tests.

### DIFF
--- a/thin-replica-server/test/thin_replica_server_test.cpp
+++ b/thin-replica-server/test/thin_replica_server_test.cpp
@@ -92,20 +92,27 @@ EventGroupMap generateEventGroupMap(EventGroupId start, EventGroupId end) {
 class FakeStorage : public concord::kvbc::IReader {
  public:
   FakeStorage(BlockMap&& db) : db_(std::move(db)), block_id_(db_.size()) {}
-  FakeStorage(EventGroupMap&& db) : eg_db_(std::move(db)), eg_id_(eg_db_.size()) {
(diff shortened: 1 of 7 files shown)


thin-replica-server/test/thin_replica_server_test.cpp
```diff
@@ -954,6 +970,84 @@ TEST(thin_replica_server_test, getClientIdFromClientCert) {
...
+
+  // ThinReplica legacy events request
+  SubscriptionRequest request;
+  request.mutable_events()->set_block_id(3);
+
+  auto status =
+      replica_.SubscribeToUpdates<TestServerContext, TestServerWriter<Hash>, Hash>(&context, &request, &stream);
+  EXPECT_EQ(status.error_code(), grpc::StatusCode::OK);
+  EXPECT_EQ(state_machine.numEventGroupsReceived(), total_event_groups);
+}
```
shgandhi, line 1050: Nit: It might be useful to check if `numLegacyBlocksReceived()` returns 0, but that can be added separately.

shgandhi (approved):
Thanks a lot for this change @robem, and for the tests! LGTM!

thin-replica-server/test/thin_replica_server_test.cpp
robem: Good point! This change doesn't have immediate urgency so I will fix it now.

PR #1832, opened 2021-08-18
ConcordClient: Sync all logger prefixes

robem:
For logging configuration and for reading logs it makes the most sense to align
all logger names to start with the same `concord.client` prefix which is based
on the project name and directory path.


shgandhi (approved):
Thanks for the change @robem! LGTM.

PR #1833, opened 2021-08-19
Completely remove Crypto.cpp and Crypto.hpp

yontyon:


kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -19,13 +19,18 @@
...
-#include "Crypto.hpp"
 #include "kvbc_key_types.h"
 #include "categorization/db_categories.h"
 #include "storage/merkle_tree_key_manipulator.h"
 #include "bcstatetransfer/DBDataStore.hpp"
 #include "bftengine/PersistentStorageImp.hpp"
 #include "bftengine/DbMetadataStorage.hpp"
+#include "crypto_utils.hpp"
+
+#pragma GCC diagnostic push
```
toly-kournik, line 30: shouldn't we do it in crypto_utils.hpp?

bftengine/src/bftengine/Digest.cpp
```diff
@@ -13,11 +13,87 @@
...
+#elif defined SHA256_DIGEST
+#define DigestType SHA256
+#elif defined SHA512_DIGEST
+#define DigestType SHA512
+#endif
 
 namespace bftEngine {
 namespace impl {
 
+size_t DigestUtil::digestLength() { return DigestType::DIGESTSIZE; }
```
toly-kournik, line 44: Should it be a part of crypto_utils?

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
yontyon: no, crypto_uils.hpp is agnostic to the implementation. The implementation is in crypto_utils.cpp

bftengine/src/bftengine/Digest.cpp
yontyon: It is used only by bftengine, so I would prefer not to touch it now

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
toly-kournik: so why is there include for crypto/dll.h?

toly-kournik (approved):
LGTM! Great refactoring

PR #1835, opened 2021-08-19
TS: Provide timestamp to Reconfiguration handlers and save it in bloc…

ukg19:
Reason:
* The time will be used by the time-based pruning and other
reconfiguration handlers in the future.
* In certain cases, replicas replay reconfiguration commands saved in the
blockchain. For the time-based pruning, the system has to replay the
command together with the time when the command was submitted.  In order
to achieve it, reconfiguration requests will be saved with the timestamp
(as CMF message).

Changes:
* Updated Reconfiguration-related interfaces
* Added a new CMF-message
* Added saving and retrieving the timestamp from the reconf block
* Refactored persistReconfigurationBlock

Co-authored-by: f-squirrel <dmitry.b.danilov@gmail.com>


ukg19:
Stale now. Opened a new one.

PR #1838, opened 2021-08-20
Upgrade command enhancements with (n-f) support

nkumar04:
In this PR, we have used latest improvements done in reconfiguration engine to enhance the upgrade functionality. This also adds (n-f) support for upgrade. With this change, replicas wedge on next stable/super-stable checkpoint, runs consensus before initiating the upgrade.


bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -36,13 +36,17 @@ class ControlStateManager {
   bool getPruningProcessStatus() const { return onPruningProcess_; }
   bool getRestartBftFlag() const { return restartBftEnabled_; }
   void setRestartBftFlag(bool bft) { restartBftEnabled_ = bft; }
-
   void setRemoveMetadataFunc(std::function<void(bool)> fn) { removeMetadataCbRegistry_.add(fn); }
   void setRestartReadyFunc(std::function<void()> fn) { sendRestartReady_ = fn; }
   void sendRestartReadyToAllReplica() { sendRestartReady_(); }
+  void setInstallReadyFunc(std::function<void(const std::string&)> fn) { sendInstallReady_ = fn; }
```
yontyon, line 42: Can we somehow leverage the existing restart proof mechanism (and maybe change its name) instead of introducing another mechanism?

bftengine/src/bftengine/messages/ReplicasRestartReadyProofMsg.cpp
```diff
@@ -13,20 +13,21 @@
...
 #include "SysConsts.hpp"
 #include "EpochManager.hpp"
-#include "SigManager.hpp"
 
 namespace bftEngine {
 namespace impl {
 
 ReplicasRestartReadyProofMsg::ReplicasRestartReadyProofMsg(ReplicaId senderId,
                                                            SeqNum seqNum,
+                                                           RestartReason reason,
```
yontyon, line 22: Same as above, if we already introduce a "reason" to the restart proof, then why not have it in the restart message itself and avoid adding another message type?

kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -266,8 +266,12 @@ bool ReconfigurationHandler::handle(const concord::messages::InstallCommand& com
...
   std::vector<uint8_t> serialized_command;
   concord::messages::serialize(serialized_command, command);
-  auto blockId = persistReconfigurationBlock(
-      serialized_command, bft_seq_num, std::string{kvbc::keyTypes::reconfiguration_install_key}, false);
+  concord::kvbc::categorization::VersionedUpdates ver_updates;
+  ver_updates.addUpdate(std::string{kvbc::keyTypes::reconfiguration_install_key},
+                        std::string(serialized_command.begin(), serialized_command.end()));
+  auto epoch = bftEngine::EpochManager::instance().getSelfEpochNumber();
+  ver_updates.addUpdate(std::string{keyTypes::reconfiguration_epoch_key}, concordUtils::toBigEndianStringBuffer(epoch));
+  auto blockId = persistReconfigurationBlock(ver_updates, bft_seq_num, false);
```
yontyon, line 274: I think the wedge=true here (we wedge before install)

reconfiguration/src/reconfiguration_handler.cpp
```diff
@@ -123,6 +123,30 @@ bool ReconfigurationHandler::handle(const concord::messages::RestartCommand& com
...
+                                    uint32_t,
+                                    concord::messages::ReconfigurationResponse& rres) {
+  concord::messages::ReconfigurationErrorMsg error_msg;
+  if (cmd.version.empty()) {
+    LOG_ERROR(getLogger(), "InstallCommand received with empty version string at seq_num " << sequence_num);
+    return false;
+  }
+  LOG_INFO(getLogger(), "InstallCommand instructs replica to stop at seq_num " << KVLOG(sequence_num, cmd.version));
+  bftEngine::ControlStateManager::instance().setStopAtNextCheckpoint(sequence_num);
+  // TODO(NK): set remove_metadata and unwedge flag to True once we support start new epoch with (n-f) nodes
```
yontyon, line 137: If we don't remove the metadata on the new software install, it means that our new SW version has to be backward compatible (at least with the metadata scheme).
I wonder if it wouldn't be better to remove the metadata and state that we don't support n-f recovery (as with add/remove). WDYT?

As for the new epoch, in my opinion we need to start a new epoch even if don't remove the metadata

bftengine/include/bftengine/ControlStateManager.hpp
nkumar04: fixed

bftengine/src/bftengine/messages/ReplicasRestartReadyProofMsg.cpp
nkumar04: fixed

kvbc/src/reconfiguration_kvbc_handler.cpp
nkumar04: fixed

reconfiguration/src/reconfiguration_handler.cpp
nkumar04: Eventually, we need to remove metadata, when we remove the condition to start the cluster only when all n nodes are working. 
Right now, i am assuming that metadata will be compatible with new version of software. If we state that we support (n-f) install only (like add/remove) then we are mandating customer to spin a new VM and add that to the cluster (which also means, reconfiguration step post install). I feel, in first phase we can allow cluster to operate even after (n-f) install (with assumption that metadata is compatible with new software). Do you see any issue with this implementation?

yontyon: no, its ok.
I would only add a new epoch here (you can see how it is done in the unwedge handler without restarting concordbft, simply add a new block with the new epoch and continue running)

kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -266,8 +266,12 @@ bool ReconfigurationHandler::handle(const concord::messages::InstallCommand& com
                                     concord::messages::ReconfigurationResponse&) {
   std::vector<uint8_t> serialized_command;
   concord::messages::serialize(serialized_command, command);
-  auto blockId = persistReconfigurationBlock(
-      serialized_command, bft_seq_num, std::string{kvbc::keyTypes::reconfiguration_install_key}, false);
+  concord::kvbc::categorization::VersionedUpdates ver_updates;
```
yontyon, line 269: You can use this https://github.com/vmware/concord-bft/blob/c283616fdc0566afa64c661d7730538c55cddfe7/kvbc/src/reconfiguration_kvbc_handler.cpp#L22. It already adds the epoch.
Also, if you want to start a new epoch, you can add another post reconfiguration handler that adds the new epoch using this https://github.com/vmware/concord-bft/blob/c283616fdc0566afa64c661d7730538c55cddfe7/kvbc/src/reconfiguration_kvbc_handler.cpp#L78

yontyon (approved):
LGTM!
I had only one comment (maybe I misunderstood you)
Thanks!

PR #1839, opened 2021-08-20
TRC: Consume record_time

shgandhi:
Each event group must contain a time event, this event may be added in any order to list of events, and we no longer can use a time_key to distinguish a time event from other events because events no longer have keys associated with them.
Hence, the TRC iterates over all the events, parses them to identify the event corresponding to the TimeLogUpdateEntry message type. The record_time field is copied from TimeLogUpdateEntry object, to the record_time field defined in the EventGroup struct in update.hpp.
To allow the above, thin_replica.proto is extended to include TimeLogUpdateEntry message, and the record_time type is changed to google::Protobuf::Timestamp type in EventGroup definition in update.hpp, for simplicity.
The unit tests required for this change require updating trc_mock files, hence will be added as a separate change.


thin-replica-server/proto/thin_replica.proto
```diff
@@ -129,3 +129,10 @@ message EventGroup {
   google.protobuf.Timestamp record_time = 3;
```
robem, line 129: This field is supposed to be set by the TRS and hence the TRC shouldn't need to try finding the time from the list of events.

PR #1843, opened 2021-08-23
Send a noop command once it is ready (instead of waiting to timeout)

yontyon:
Prior to this change, once we called to tryToSendPrePreprapre(false) nothing happened because the noop command was not yet in the primary queue.
Now, we first push it to the primary queue and only then invoke the tryToSendPrePrepre(false)


nkumar04 (approved):
LGTM!

PR #1844, opened 2021-08-23
Docker update to version 0.29

ananpal:
tc - show / manipulate traffic control settings
it provides control over traffic rate which is used in apollo to control the traffic flow for specific test cases.


glevkovich (approved):
LGTM!
Remember to push the image to Artifactory before merging.

PR #1845, opened 2021-08-24
overload put method with move semantics

cloudnoize:


util/include/lru_cache.hpp
```diff
@@ -37,8 +37,8 @@ class LruCache {
         map_.erase(keys_.back());
         keys_.pop_back();
       }
-      keys_.push_front(key);
-      map_.insert({key, {value, keys_.begin()}});
+      keys_.push_front(std::move(key));
+      map_.insert({key, {std::move(value), keys_.begin()}});
```
dartdart26, line 41: The `key` variable is used after move on line 41. I think there's no way around copying the key at least once.

cloudnoize: thanks for the catch!

util/include/lru_cache.hpp
```diff
@@ -27,21 +27,12 @@ class LruCache {
...
-        keys_.pop_back();
-      }
-      keys_.push_front(key);
-      map_.insert({key, {value, keys_.begin()}});
-    }
+    putImp(std::forward<const Key&>(key), std::forward<const Value&>(value));
   }
 
+  // using this rvalue put will invalidate any usage of the key-value that were passed to the method.
+  void put(Key&& key, Value&& value) { putImp(std::forward<Key&&>(key), std::forward<Value&&>(value)); }
```
dartdart26, line 34: I think we don't need to do `std::forward` here. Instead, we need to do it inside `putImp()`, instead of the `std::moves` there.

PR #1847, opened 2021-08-24
Don't save the client keys when loading them from the reserved pages

yontyon:
Currently, we save the client exchanged keys when reading them from the reserved pages.
Other than being redundant, it also crashes the replica when it finishes state transfer.
This PR fixes the bug as well as introducing an Apollo test to validate the fix.
The test was running without the fix and failed and with the fix, it passed.


nkumar04 (approved):
LGTM!
Thanks!

PR #1848, opened 2021-08-24
testcase to verify state transfer with internal cycle added

ananpal:
it tests the scenario where during the state transfer cycle the network advances to a new state and state transfer undergoes an internal cycle of state transfer to fetch the new state


tests/apollo/util/bft.py
```diff
@@ -52,6 +52,28 @@
     'num_ro_replicas'
 ])
 
+class NetworkTrafficControl():
```
teoparvanov, line 55: have you considered placing this class in https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft_network_partitioning.py ?

PR #1853, opened 2021-08-26
Optional compaction after pruning

yontyon:
We add an option to invoke rocksb compaction explicitly after pruning

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -131,7 +131,7 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
   CONFIG_PARAM(numBlocksToKeep_, uint64_t, 0, "how much blocks to keep while pruning");
   CONFIG_PARAM(manualCompaction_,
                bool,
-               true,
+               false,
                "Enable manual rocksb compaction after pruning (to reclaim the disc space immediately)");
 
   CONFIG_PARAM(debugPersistentStorageEnabled, bool, false, "whether persistent storage debugging is enabled");


kvbc/src/pruning_handler.cpp
```diff
@@ -266,6 +267,14 @@ void PruningHandler::pruneThroughBlockId(kvbc::BlockId block_id) const {
...
         std::terminate();
       }
+      // Now that we done with pruning, lets run a manual compaction
+      if (manual_compaction) {
+        LOG_INFO(logger_, "start compaction process");
+        auto status = blocks_deleter_.reclaimDiskSpace();
+        if (!status.isOK()) {
+          LOG_ERROR(logger_, "unable to run compaction " << status.toString());
+        }
+      }
```
dartdart26, line 277: Will this work with a different pruning handler (i.e. the one used in Concord)?

storage/include/rocksdb/client.h
```diff
@@ -97,6 +97,16 @@ class Client : public concord::storage::IDBClient {
   concordUtils::Status multiPut(const SetOfKeyValuePairs& _keyValueMap) override;
   concordUtils::Status multiDel(const KeysVector& _keysVec) override;
   concordUtils::Status rangeDel(const Sliver& _beginKey, const Sliver& _endKey) override;
+
+  Status reclaimDiskSpace() override {
+    ::rocksdb::CompactRangeOptions opt;
+    opt.allow_write_stall = true;
+    opt.change_level = true;
+    auto s = dbInstance_->CompactRange(opt, nullptr, nullptr);
```
dartdart26, line 105: I think this call will only compact the default column family: https://github.com/facebook/rocksdb/blob/master/include/rocksdb/db.h#L1143

I think we should call it per column family: https://github.com/facebook/rocksdb/blob/master/include/rocksdb/db.h#L1138

dartdart26:
Do we want to merge this in 0.11.x at that stage? @teoparvanov ?

storage/include/rocksdb/client.h
yontyon: fixed

kvbc/src/pruning_handler.cpp
yontyon: yes, but we added this change on the bump also

dartdart26 (approved):
Thanks Yoni!

teoparvanov:
> 
> 
> Do we want to merge this in 0.11.x at that stage? @teoparvanov?

Hi @dartdart26 and @yontyon, yes, once we make sure it improves the "responsiveness" of pruning and it has been validated by system testing (using a custom build with the current changes).

PR #1854, opened 2021-08-26
BCStateTran, SourceSelector: Maintain an actual list of sources

glevkovich:
Till now, every time a new source is selected, a list of sources (which
isused for cycle end report) is updated immediately.
But if a source does not answer, or if it sends a corrupted data etc.,
it is not really a source.
An actual or "real" source, is one in which at least one block has been
received from (during Gettingmissingblocks or GettingMissingResPages).

Here we move the maintenance of this list into source selector, and
report to it with a new API function whenever a new block is received.

### DIFF
--- a/bftengine/src/bcstatetransfer/SourceSelector.cpp
+++ b/bftengine/src/bcstatetransfer/SourceSelector.cpp
@@ -31,6 +31,16 @@ void SourceSelector::setFetchingTimeStamp(uint64_t currTimeMilli, bool retransmi
 void SourceSelector::removeCurrentReplica() {
   preferredReplicas_.erase(currentReplica_);
   currentReplica_ = NO_REPLICA;
+  receivedValidBlockFromSrc_ = false;
+}
+
+void SourceSelector::onReceivedValidBlockFromSource() {
+  ConcordAssertNE(currentReplica_, NO_REPLICA);
+  if (!receivedValidBlockFromSrc_) {
+    receivedValidBlockFromSrc_ = true;
+    LOG_INFO(logger_, "Insert source " << currentReplica_ << " into actualSources_");
+    actualSources_.push_back(currentReplica_);
+  }
 }
 
 void SourceSelector::setAllReplicasAsPreferred() { preferredReplicas_ = allOtherReplicas_; }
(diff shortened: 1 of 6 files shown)


bftengine/src/bcstatetransfer/SourceSelector.cpp
```diff
@@ -31,6 +31,16 @@ void SourceSelector::setFetchingTimeStamp(uint64_t currTimeMilli, bool retransmi
...
   currentReplica_ = NO_REPLICA;
+  receivedValidBlockFromSrc_ = false;
+}
+
+void SourceSelector::onReceivedValidBlockFromSource() {
+  ConcordAssertNE(currentReplica_, NO_REPLICA);
+  if (!receivedValidBlockFromSrc_) {
+    receivedValidBlockFromSrc_ = true;
+    LOG_INFO(logger_, "Insert source " << currentReplica_ << " into actualSources_");
+    actualSources_.push_back(currentReplica_);
```
cloudnoize, line 42: can we give up on the flag ```receivedValidBlockFromSrc_``` if we'll use std::set instead of std::vector for ```actualSources_``` and the presence of currentReplica_ in the set will indicate the same logic as ```receivedValidBlockFromSrc_ == true```.
If we can give up on managing ```receivedValidBlockFromSrc_``` it can save future bugs, if it's not the case, can UT be added ?

glevkovich: @cloudnoize , I don't think we can use an STL set here.
The order of the elements is important and a set is associative, and also there can be re-occurring elements.
For example, I can have a vector containing [0,6,1,1,0].

As for the UT suggestion, I think it is a good idea, in practice I wanted to move into test-driven development - so just last week I've inserted the new infra into bcstatetransfer_tests.cpp , Before that it was close to impossible to do your suggestion without a work of at least a week. what was done till now is still very initial. I will see if it is possible to have a unit test with what we have now, else we will have to do it in another task (there is a task for adding unit tests, but it will require weeks of work).
I'll comment again after checking if a unit test can be added..

cloudnoize (approved):
LGTM thanks @glevkovich

bftengine/src/bcstatetransfer/SourceSelector.cpp
glevkovich: @cloudnoize , I have added a unit-test for the change.
On the way, I've found a minor bug that was fixed as well as part of this PR.

PR #1858, opened 2021-08-29
Write the new client public key to disk

yontyon:
The client's key exchange procedure in the replica side saves the new client public key in the reserved pages only.
However, in case we remove metadata (using kv_blockchain_editor) we also remove the reserved pages, and hence we left a mismatch between the client's latest private key and the replicas' public key for this client.
In this PR we introduce an ability to save the new public key in the file system in addition to the reserved pages.

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -123,7 +123,12 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
   // RSA public keys of all clients. Each public key holds set of distinct client (principal) ids which are expected to
   // sign with the matching private key
   std::set<std::pair<const std::string, std::set<uint16_t>>> publicKeysOfClients;
-
+  std::unordered_map<uint16_t, std::set<uint16_t>> clientGroups;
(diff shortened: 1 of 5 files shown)


bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -123,7 +123,12 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
   // RSA public keys of all clients. Each public key holds set of distinct client (principal) ids which are expected to
   // sign with the matching private key
   std::set<std::pair<const std::string, std::set<uint16_t>>> publicKeysOfClients;
-
+  std::unordered_map<uint16_t, std::set<uint16_t>> clientGroups;
+  CONFIG_PARAM(clientsKeysPrefix, std::string, "", "the path to the client keys directory");
+  CONFIG_PARAM(saveClinetKeyFile,
```
nkumar04, line 128: why not True?

nkumar04 (approved):
LGTM!!

bftengine/include/bftengine/ReplicaConfig.hpp
yontyon: Basically, with Apollo and Hermes tests all replicas are looking on the same files.
Saving the same file from multiple sources may cause issues (at least the Apllo tests have failed)

PR #1859, opened 2021-08-29
BCStateTran: fix assertion failure while the source was not finalize

glevkovich:
When there are 2 replicas in ST at the same time or a replica that was
a source and then enters state transfer as a destination, we fail it
fails an assertion on BCStateTran.cpp:L1333. This happens due to the
fact that it might be still regarded as a source, and its iocontexts_
was not cleared.

Here we do the following changes to finalize a source correctly:
1) New member function BCStateTran::clearIoContexts() to clear
iocontexts_.
2) New member function BCStateTran::finalizeSource() to clear source
flag, counter and log its histograms. Finally, it also calls
BCStateTran::clearIoContexts().
    *) Call if needed from  BCStateTran::startCollectingState().
    *) Call as before from  BCStateTran::onTimerIm().
3) Transform DEBUG level log entries in source to INFO level to get
better visibility when sending a batch has ended.


yontyon (approved):
LGTM!
Thanks

PR #1860, opened 2021-08-30
 Add <principal_id, token> map for addRemove command

nkumar04:
Replicas and clients need to send updateConfig request with their specific token id to config service. With this change, reconfiguration handler expects addRemove command with a list of tokens mapped to the bft principal id from the operator. replica or client uses this token while sending update config request to agent


reconfiguration/cmf/concord.cmf
```diff
@@ -192,12 +193,19 @@ Msg ClientKeyExchangeStatusResponse 48 {
...
+    bool restart
+}
+
 Msg ClientStateReply 39 {
     uint64 block_id
     oneof {
         ClientExchangePublicKey
         ClientKeyExchangeCommand
         ClientsAddRemoveCommand
+        ClientsAddRemoveExecuteCommand
```
yontyon, line 208: By adding this new message, we cannot return the client the "ClientsAddRemoveCommand" as it contains the tokens of all the clients.
I would suggest to either remove the ClientsAddRemoveCommand from the possible responses to the CRE or to assume that we send the clients' addRemove commands one by one

kvbc/include/kvbc_key_types.hpp
```diff
@@ -28,10 +28,11 @@ static const char timestamp_key = 0x2e;
 static const char reconfiguration_restart_key = 0x30;
 enum CLIENT_COMMAND_TYPES : uint8_t {
   start_ = 0x0,
-  PUBLIC_KEY_EXCHANGE = 0x1,            // identifier of public key exchange request by client
-  CLIENT_KEY_EXCHANGE_COMMAND = 0x2,    // identifier of client key exchange request by operator
-  CLIENT_SCALING_COMMAND = 0x3,         // identifier of client scaling request by operator
-  CLIENT_SCALING_COMMAND_STATUS = 0X4,  // identifier of client update request after successful scaling
+  PUBLIC_KEY_EXCHANGE = 0x1,             // identifier of public key exchange request by client
+  CLIENT_KEY_EXCHANGE_COMMAND = 0x2,     // identifier of client key exchange request by operator
+  CLIENT_SCALING_COMMAND = 0x3,          // identifier of client scaling request by operator
```
yontyon, line 33: See below

yontyon (commented):
Great PR. 
I have one comment (please see above)
Thanks!

reconfiguration/cmf/concord.cmf
nkumar04: Fixed

yontyon (approved):
LGTM!
Thanks

PR #1863, opened 2021-08-30
Don't add checkpoint messages of previous epoch

yontyon:
on receiving a checkpoint message, we cannot add it if it belongs to a previous epoch.
Otherwise, the replica won't ask for the updated checkpoint message and eventually, it get blocked

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1901,7 +1901,8 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
 
   bool askForStateTransfer = false;
 
-  if (msgIsStable && (msgSeqNum > lastExecutedSeqNum || msgEpochNum >= getSelfEpochNumber())) {
+  if (msgIsStable &&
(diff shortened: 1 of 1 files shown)


nkumar04 (approved):
Good catch!
LGTM!

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1903,7 +1904,7 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
   if (msgIsStable && (msgSeqNum > lastExecutedSeqNum || msgEpochNum >= getSelfEpochNumber())) {
     auto pos = tableOfStableCheckpoints.find(msgGenReplicaId);
     if (pos == tableOfStableCheckpoints.end() || pos->second->seqNumber() <= msgSeqNum ||
```
nkumar04, line 1906: f ( (pos == tableOfStableCheckpoints.end() &&  msgEpochNum >= getSelfEpochNumber()) || 
       pos->second->seqNumber() <= msgSeqNum ||
        msgEpochNum >= pos->second->epochNumber()) {
}
//we don't want to add checkpoint message from previous epoch to tableOfStableCheckpoints

teoparvanov:
Hi @yontyon and @nkumar04 , can this fix be merged and bumped into Concord?

teoparvanov (approved):
LGTM!

PR #1865, opened 2021-08-31
Updates on client pool

Liorgal28:
*Length of second leg commands has been changed and now we need to differentiate according to command type.
*Update default timeouts for clients
*Update client gauge


yuliasherman (approved):
LGTM!

PR #1867, opened 2021-08-31
TRC: Fix last verified block id value

shgandhi:
This commit sets the `latest_verified_block_id_` to
`block_id - 1`, when requested block_id is greater than 0.
`latest_verified_block_id_` cannot be equal to the requested block
id.


robem (approved):
Thanks for the fix @shgandhi 🚀 
According to [event.proto](https://github.com/vmware/concord-bft/blob/master/client/proto/event/v1/event.proto) we need to be inclusive.

upshaw-alex (approved):

PR #1870, opened 2021-09-01
Remove all gen-sec files on metadata removal

yontyon:
On replicas scaling, we may end up with a replica that gets the id of another replica. In that case, the old gen-sec file won't be removed.
In this PR we remove all gen-sec files on metadata removal


nkumar04 (approved):
LGTM!!

PR #1873, opened 2021-09-02
Fix latest block ID inconsistency with last reachable block ID

mdarade:
As writing blocks now happens concurrently it would result in race while
updating latest block ID with the block IDs of the first few
blocks from every checkpoint in RO replica.

### DIFF
--- a/kvbc/src/direct_kv_db_adapter.cpp
+++ b/kvbc/src/direct_kv_db_adapter.cpp
@@ -32,6 +32,7 @@
 #include <utility>
 #include <sstream>
 #include <iomanip>
+#include <mutex>
 
 using logging::Logger;
 using concordUtils::Status;
@@ -400,7 +401,7 @@ BlockId DBAdapter::addBlock(const SetOfKeyValuePairs &kv) {
   return blockId;
 }
 
-void DBAdapter::addRawBlock(const RawBlock &block, const BlockId &blockId) {
+void DBAdapter::addRawBlock(const RawBlock &block, const BlockId &blockId, bool lastBlock) {
   SetOfKeyValuePairs keys;
   if (saveKvPairsSeparately_ && block.length() > 0) {
     keys = getBlockData(block);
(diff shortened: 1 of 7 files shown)


kvbc/src/direct_kv_db_adapter.cpp
```diff
@@ -412,10 +413,27 @@ void DBAdapter::addRawBlock(const RawBlock &block, const BlockId &blockId) {
   // when ST runs, blocks arrive in batches in reverse order. we need to keep
   // track on the "Gap" and to close it. Only when it is closed, the last
   // reachable block becomes the same as the last block
-  BlockId lastReachableBlock = getLastReachableBlockId();
-  if (blockId > getLatestBlockId()) setLatestBlock(blockId);
 
-  if (blockId == lastReachableBlock + 1) setLastReachableBlockNum(getLatestBlockId());
+  // old cycle
```
glevkovich, line 417: nit: did you mean to leave lines 417-427 as a comment?
In my opinion, it should be removed...
If you decide to keep it,  better I'm multiline in one section:
```
  /*
....
  */
```

kvbc/src/direct_kv_db_adapter.cpp
```diff
@@ -412,10 +413,27 @@ void DBAdapter::addRawBlock(const RawBlock &block, const BlockId &blockId) {
...
+  // t1:1098, t2:1099
+
+  // getLatestBlockId() from earlier cycle = 849
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (blockId > getLatestBlockId()) setLatestBlock(blockId);
+  }
+  BlockId lastReachableBlock = getLastReachableBlockId();
+  // 850 == 849 + 1 will happen only once so no lock required
+  if ((lastBlock) and (blockId == lastReachableBlock + 1)) setLastReachableBlockNum(getLatestBlockId());
```
glevkovich, line 436: nit:
```
if ((lastBlock) and (blockId == getLastReachableBlockId() + 1)) setLastReachableBlockNum(getLatestBlockId());
```

glevkovich (approved):
@mdarade , thanks for the fix!
2 small minor comments.

One last thing: we should be aware of the fact this we might need to. consider to do the same changes in the future on a Markle adapter (This is not theoretical, it might happen soon):
In that case calling to:
` m_bcDbAdapter->addRawBlock(block, blockId);`
Will call the Markle adapter, where we are not doing anything with the lastBlock.
Maybe we should add a low priority task for that in improvements for epic version  1.5, since for now such a change is untestable in real deployments.

PR #1875, opened 2021-09-03
Minor fixes and added test RO replica reconfiguration handler

nkumar04:
- Added handler to test st_based_client reconfiguration handler for RO replica
- Start cre engine only on read only replica
- Add reconfiguration update only on wedge point

### DIFF
--- a/kvbc/src/Replica.cpp
+++ b/kvbc/src/Replica.cpp
@@ -64,8 +64,12 @@ Status Replica::initInternals() {
     m_replicaPtr = bftEngine::IReplica::createNewRoReplica(replicaConfig_, requestHandler, m_stateTransfer, m_ptrComm);
     m_stateTransfer->addOnTransferringCompleteCallback([this](std::uint64_t) {
       std::vector<concord::client::reconfiguration::State> stateFromReservedPages;
-      if (bftEngine::ReconfigurationCmd::instance().getStateFromResPages(stateFromReservedPages)) {
-        creClient_->pushUpdate(stateFromReservedPages);
+      uint64_t wedgePt{0};
+      if (bftEngine::ReconfigurationCmd::instance().getStateFromResPages(stateFromReservedPages, wedgePt)) {
+        LOG_INFO(GL,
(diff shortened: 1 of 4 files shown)


kvbc/src/Replica.cpp
```diff
@@ -64,8 +64,21 @@ Status Replica::initInternals() {
...
+      if (bftEngine::ReconfigurationCmd::instance().getStateFromResPages(
+              stateFromReservedPages, wedgePt, cmdEpochNum)) {
+        // get replicas latest global epoch from res pages
+        auto replicasGlobalEpochNum = bftEngine::EpochManager::instance().getGlobalEpochNumber();
+        LOG_INFO(GL,
+                 "reconfiguration command in res pages" << KVLOG(
+                     wedgePt, cmdEpochNum, replicasGlobalEpochNum, m_replicaPtr->getLastExecutedSequenceNum()));
+        // OnTransferringComplete callback is called for every 150 seqNum/checkpt, but we want to push reconfiguration
+        // command only when wedge pt is reached
+        if (static_cast<uint64_t>(m_replicaPtr->getLastExecutedSequenceNum()) >= wedgePt ||
```
yontyon, line 78: I think we should accurate this condition, for example:
replicaGlobalEpochNum < cmdEpochNum but lastExecutedSeqNum >= wedgePt (due to metadata removal), the condition will hold.
Am I right?

yontyon (approved):
LGTM!
Only one comment, please address it before merging

kvbc/src/Replica.cpp
nkumar04: explaination on the condition:
replicaGlobalEpochNum < cmdEpochNum --> this can't as i understand
case-1 - (lastExecutedSeqNum >= wedgePt) --> if replicas have not moved to new epoch and ST is complete. Then this condition holds
case-2 (replicasGlobalEpochNum > cmdEpochNum) --> replicas have move to new epoch and command belongs to old epoch. i don't care if about the sequence number. i simply push the update. If the command is already executed then it will get ignore by lastKnownReconfiguration block check

PR #1876, opened 2021-09-06
Changes to increase the retry timeout for test_primary_write

situ-s:
The python bft_client has a retry timeout of 250 miliseconds.
If the client request has not finished in 250 miliseconds, the client retries the request by sending to all replicas.
In the case of this test, we sometimes hit this timeout, making the client send more requests than expected in the test.

After Increasing retry_timeout_milli to 1000ms, we are not hitting the issue.


teoparvanov (approved):
LGTM,  thanks Situ!

PR #1878, opened 2021-09-08
Added skip_for_tls decorator to clean up code in apollo

situ-s:
Added skip_for_tls decorator in place of unittest.skipIf(environ.get('BUILD_COMM_TCP_TLS', "").lower() == "true", "Unstable on CI (TCP/TLS only)") in required apollo test cases.


teoparvanov (approved):
Looks great, thanks Situ!

PR #1879, opened 2021-09-09
Add the SynchronizedValue class template utility

dartdart26:
The `SynchronizedValue` utility allows for synchronized access to a
value from multiple threads. It supports:
 * in-place construction of values with arbitrary arguments, thus
   allowing users to use it with non-copyable and/or non-movable types
 * ability to replace the value without using type T's copy/move
   operations (that might be deleted)
 * const and non-const accessors that utilize read-write mutexes
 * RAII stype accessors that guarantee release of the mutex in all cases

Add an unit test to cover typical use cases.

### DIFF
--- a/util/include/synchronized_value.hpp
+++ b/util/include/synchronized_value.hpp
@@ -1,0 +1,92 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the
+// LICENSE file.
+
+#pragma once
+
+#include <memory>
+#include <mutex>
+#include <shared_mutex>
+#include <utility>
+
+namespace concord::util {
+
+// Syncrhonizes multi-threaded access to a value of type T.
+template <typename T>
+class SynchronizedValue {
+ private:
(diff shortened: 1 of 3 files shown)


HristoStaykov (approved):
LGTM, thanks @dartdart26 !

arc-vmware (approved):
This is very nice, Thanks Petar to implement the same.
I just have one request, it would be great to add few lines where its better to used SynchronizedValue<T> w.r.t std::atomic<T> to make sure that the user is using this correctly.

util/include/synchronized_value.hpp
```diff
@@ -0,0 +1,92 @@
...
+  // Precondition: the calling thread doesn't have any other accessors at the time of the call. If it does, the
+  // behaviour is undefined.
+  //
+  // A single thread can create one accessor only at a time. Other threads will wait until it is destroyed before they
+  // can do any operations on the synchronized value.
+  Accessor access() { return Accessor{*val_, mtx_}; }
+  ConstAccessor constAccess() const { return ConstAccessor{*val_, mtx_}; }
+
+ private:
+  std::unique_ptr<T> val_;
```
arc-vmware, line 88: Note: 
Can we replace the combination of unique_ptr<T> and the mutex with a std::atomic<std::shared_ptr<T>> ? as I can see there is some equivalence. I am not asking to make any change here, this question is to understand the approach better.

dartdart26: I think we cannot, because the intention of the mutex is to protect any thread that has an accessor and using it against another thread replacing the value.

dartdart26:
There is a race on the unique_ptr when we create the accessors. We should first create the lock, dereference the unique_ptr and, finally, move the lock to the accessor. Will fix tomorrow.

PR #1880, opened 2021-09-11
Remove redundant passing of pre_exec_all in Apollo

upshaw-alex:
This PR removes explicit values for the `pre_exec_all` parameter in instantiations of `SimpleKVBCProtocol` in Apollo where the `SimpleKVBCProtocol` is also given an `SkvbcTracker`. This PR also modifies `SimpleKVBCProtocol.__init__` to accommodate this change by defaulting the `SimpleKVBCProtocol` instance's value for `pre_exec_all` to the tracker's value of `pre_exec_all`.

This PR is intended to remove the unnecessary redundancy in instantiations of `SimpleKVBCProtocol` that give values for both tracker and `pre_exec_all`, as `SkvbcTracker`s already have a value for `pre_exec_all`, and it would be erroneous if a `SimpleKVBCProtocol` object was initialized with a tracker and an explicit value for `pre_exec_all` not matching the value the tracker already has.

### DIFF
--- a/tests/apollo/util/skvbc.py
+++ b/tests/apollo/util/skvbc.py
@@ -49,10 +49,35 @@ class SimpleKVBCProtocol:
     requests.
     """
 
-    def __init__(self, bft_network, tracker = None, pre_process=False):
+    def __init__(self, bft_network, tracker = None, pre_exec_all = None):
+        """
+        Initializer for SimpleKVBCProtocol, given a BftTestNetwork to connect to
+        as bft_network, and optionally an SkvbcTracker to use as tracker and a
+        Boolean value specifying whether to enable pre execution on all requests
+        as pre_exec_all.
+        
+        Note pre_exec_all defaults to False if no tracker is given and to the
+        tracker's value for pre_exec_all if a tracker is given, so passing an
+        explicit value for pre_exec_all is unnecessary if a tracker is in use.
+        """
+
         self.bft_network = bft_network
         self.tracker = tracker
-        self.pre_exec_all = pre_process
+
+        if (self.tracker is not None):
+            if (pre_exec_all is not None):
+                assert (self.tracker.pre_exec_all == pre_exec_all), \
+                    "SimpleKVBCProtocol constructor was given a value for " \
+                    "pre_exec_all conflicting with the value configured in " \
+                    "the SkvbcTracker it was given."
+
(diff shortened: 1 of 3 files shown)


tests/apollo/util/skvbc.py
```diff
@@ -49,10 +49,35 @@ class SimpleKVBCProtocol:
     requests.
     """
 
-    def __init__(self, bft_network, tracker = None, pre_process=False):
+    def __init__(self, bft_network, tracker = None, pre_exec_all = None):
```
teoparvanov, line 52: Hi Alex, to simplify things I suggest you removed the optional `pre_exec_all` parameter altogether. 
The value should always come from the `tracker` object.

teoparvanov (approved):
LGTM, thanks Alex! I just have one last comment before you merge this PR.

tests/apollo/util/skvbc.py
upshaw-alex: I did consider removing this parameter entirely when developing this change, though I was not sure whether or not I should do it. `tracker` is also an optional parameter, and `SimpleKVBCProtocol` currently [does use the value](https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/skvbc.py#L416) of its copy of `pre_exec_all`, regardless of whether or not`tracker` is `None`. We could still get rid of the `pre_exec_all` parameter here and just set the `SimpleKVBCProtocol`'s `pre_exec_all` attribute to `False` if no `tracker` is given, and, as far as I can tell, that would not break or otherwise affect any of the existing test cases, as I am not aware of any instances where a `SimpleKVBCProtocol` is instantiated with no `tracker` and `True` for `pre_exec_all`. However, I had some slight reservation about doing this, as it looks like `SimpleKVBCProtocol` currently does theoretically support testing with pre-execution and no history tracker, and I don't know whether we might add any test cases that actually do this in the future.

Do you have any sense of whether we intend for `SimpleKVBCProtocol` to be supporting initialization with `tracker=None` and `pre_exec_all=True` and/or what we anticipate the possibility is of future test cases that would want to test with pre-execution and no history tracker?

teoparvanov: Hi Alex, thanks for pointing this out. You are right,  we do want to preserve the option to test with pre-execution, but without a tracker.

So  I am withdrawing my concern and approve your changes for merging as they are.

PR #1881, opened 2021-09-12
Block new connections on addRemove command

yontyon:
On addRemove command, we may have a scenario in which a new TLS connection is created, but the replica already has the new configuration. In this case the replica search for the new TLS private key in the wrong place. 
For this, we introduce a new mechanism in which the replica can decide not to start new connections.

### DIFF
--- a/communication/src/TlsConnectionManager.cpp
+++ b/communication/src/TlsConnectionManager.cpp
@@ -18,6 +18,7 @@
 #include "assertUtils.hpp"
 #include "TlsConnectionManager.h"
 #include "AsyncTlsConnection.h"
+#include "communication/CommStateControl.hpp"
 
 namespace bft::communication::tls {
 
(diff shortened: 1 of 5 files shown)


communication/src/TlsConnectionManager.cpp
```diff
@@ -257,7 +258,10 @@ void ConnectionManager::startClientSSLHandshake(asio::ip::tcp::socket&& socket,
 
 void ConnectionManager::accept() {
   acceptor_.async_accept(asio::bind_executor(strand_, [this](asio::error_code ec, asio::ip::tcp::socket sock) {
-    if (ec) {
+    if (bft::communication::CommStateControl::instance().getBlockNewConnectionsFlag()) {
```
nkumar04, line 261: so, if an operator starts after config update then it will not be able to initiate a new connection. Even without this change this can result in crash. 
Do we need to mention in runbook that new session from operator won't be accepted after configuration update?

nkumar04 (approved):
LGTM!!
Thanks!

communication/src/TlsConnectionManager.cpp
yontyon: The client scaling command does not restart the operator exactly for this reason.
I think it should be fine
The only problem is if the operator is restarting after the configuration change

PR #1882, opened 2021-09-13
Block new connections on addRemove command 

yontyon:
On addRemove command, we may have a scenario in which a new TLS connection is created, but the replica already has the new configuration. In this case the replica search for the new TLS private key in the wrong place.
For this, we introduce a new mechanism in which the replica can decide not to start new connections.

(See #1881)


nkumar04 (approved):
LGTM!!

PR #1883, opened 2021-09-13
Code Coverage report with release build

itsmeakashgoyal:
Removing existing functionality where code coverage report was generated with debug build.
Now it will be generated with release build.


teoparvanov (approved):
Thanks Akash, I'm ok with this change, as long as the coverage report is correctly generated.

PR #1885, opened 2021-09-13
Add ability to crate self signed certificates

yontyon:
To enable client TLS key rotation, we need to have the ability to create new certificates programmatically. 
This PR is the first effort for clients' TLS key exchange

### DIFF
--- a/util/src/crypto_utils.cpp
+++ b/util/src/crypto_utils.cpp
@@ -17,17 +17,25 @@
 #include <cryptopp/dll.h>
 #include <cryptopp/pem.h>
 #include <cryptopp/rsa.h>
-#pragma GCC diagnostic pop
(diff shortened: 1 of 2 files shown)


util/src/crypto_utils.cpp
```diff
@@ -17,17 +17,25 @@
...
+#pragma GCC diagnostic pop
+
+#include <openssl/bio.h>
+#include <openssl/ec.h>
+#include <openssl/pem.h>
+#include <openssl/x509.h>
+#include <openssl/evp.h>
+
 #include "Logger.hpp"
 using namespace CryptoPP;
```
toly-kournik, line 31: maybe remove the `using namespace CryptoPP` if you explicitly use it in code?

toly-kournik (approved):
LGTM!

PR #1886, opened 2021-09-13
Removing tsan & asan failing badges from the repo.

pkthapa:
These badges will be added back after the tsan & asan reports become stable.


teoparvanov (approved):
LGTM, thanks Pankaj!

PR #1889, opened 2021-09-14
[Client pool] Cover all GRPC statuses

Liorgal28:
In this PR, I have included all the different gRPC statuses that the client should return.


client/client_pool/include/client/client_pool/concord_client_pool.hpp
```diff
@@ -45,6 +45,9 @@ enum SubmitResult {
   Acknowledged,  // The request has been queued for submission
   Overloaded,    // There is no available client to the moment to process the
   // request
+  Invalid_Argument,
+  Timed_Out,
+  Client_Unavailable,
```
robem, line 50: `InvalidArgument` without the underscore - same for the others.
Also, can you please add a comment to `ClientUnavailable` and provide more details.

client/client_pool/src/concord_client_pool.cpp
```diff
@@ -43,6 +43,14 @@ SubmitResult ConcordClientPool::SendRequest(std::vector<uint8_t> &&request,
                                             std::string correlation_id,
                                             std::string span_context,
                                             const bftEngine::RequestCallBack &callback) {
+  if (callback && timeout_ms.count() == 0) {
+    callback(bftEngine::SendResult{SubmitResult::Invalid_Argument});
+    return SubmitResult::Overloaded;
```
robem, line 48: We send "invalid argument" to the callback but return with "overloaded"? Which one is it?
Same below.

client/clientservice/src/request_service.cpp
```diff
@@ -45,8 +45,23 @@ Status RequestServiceImpl::Send(ServerContext* context, const Request* proto_req
   auto callback = [&](cc::SendResult&& send_result) {
     if (not std::holds_alternative<bft::client::Reply>(send_result)) {
       LOG_INFO(logger_, "Send returned error");
-      if (std::get<int>(send_result) == concord_client_pool::Overloaded)
-        status.set_value(grpc::Status(grpc::StatusCode::RESOURCE_EXHAUSTED, "Overloaded"));
+      switch (std::get<int>(send_result)) {
+        case (concord_client_pool::Overloaded):
+          status.set_value(grpc::Status(grpc::StatusCode::RESOURCE_EXHAUSTED, "ALL CLIENTS OCCUPIED"));
```
robem, line 50: There is no need to scream 🙀  at the user :) Would you mind keeping the string lower-case?

client/clientservice/src/request_service.cpp
```diff
@@ -45,8 +45,23 @@ Status RequestServiceImpl::Send(ServerContext* context, const Request* proto_req
...
+          break;
+        case (concord_client_pool::Invalid_Argument):
+          status.set_value(grpc::Status(grpc::StatusCode::INVALID_ARGUMENT, "INVALID ARGUMENT SENT"));
```
robem, line 53: "invalid argument" should be good enough

client/clientservice/src/request_service.cpp
```diff
@@ -45,8 +45,23 @@ Status RequestServiceImpl::Send(ServerContext* context, const Request* proto_req
...
+          break;
+        case (concord_client_pool::Timed_Out):
+          status.set_value(grpc::Status(grpc::StatusCode::DEADLINE_EXCEEDED, "TIME OUT REACHED FOR REQUEST"));
```
robem, line 56: "timeout"

client/clientservice/src/request_service.cpp
```diff
@@ -45,8 +45,23 @@ Status RequestServiceImpl::Send(ServerContext* context, const Request* proto_req
...
+          break;
+        case (concord_client_pool::Client_Unavailable):
+          status.set_value(grpc::Status(grpc::StatusCode::UNAVAILABLE, "NO CLIENT CONNECTED TO REPLICAS"));
+          break;
+        default:
+          status.set_value(grpc::Status(grpc::StatusCode::INTERNAL, "INTERNAL ERROR"));
```
robem, line 62: If internal errors ever occur then it would be good to have more information. Can we get the details from the logs? Or at least, can we include the error code in the message?

robem (changes requested):

client/client_pool/src/concord_client_pool.cpp
Liorgal28: As DA currently does not know how to deal with other statuses, we can only return Acknowledge/Overloaded/internal error.

client/clientservice/src/request_service.cpp
Liorgal28: The logs we currently have should cover the reason internal errors happen.

client/client_pool/include/client/client_pool/concord_client_pool.hpp
Liorgal28: Done

client/clientservice/src/request_service.cpp
Liorgal28: Done

client/clientservice/src/request_service.cpp
Liorgal28: Done

client/clientservice/src/request_service.cpp
Liorgal28: Done

client/client_pool/src/concord_client_pool.cpp
```diff
@@ -148,7 +156,12 @@ SubmitResult ConcordClientPool::SendRequest(std::vector<uint8_t> &&request,
     ClientPoolMetrics_.rejected_counter++;
     is_overloaded_ = true;
     LOG_WARN(logger_, "Cannot allocate client for" << KVLOG(correlation_id));
-    if (callback) callback(bftEngine::SendResult{SubmitResult::Overloaded});
+    if (callback) {
+      if (serving_candidates == 0 && !clients_.empty())
+        callback(bftEngine::SendResult{SubmitResult::ClientUnavailable});
+      else
+        callback(bftEngine::SendResult{SubmitResult::Overloaded});
```
robem, line 163: From the Google style guide: "[...] if an if statement has no else or else if clauses, then the curly braces for the controlled statement [...] may be omitted [...]"

Please add curly braces here and in all other places in this PR.

robem (changes requested):

client/client_pool/src/concord_client_pool.cpp
Liorgal28: Done

PR #1890, opened 2021-09-14
Save multiple replies on reserve pages

Liorgal28:
This PR adds the ability to save multiple responses per client on reserve pages.
Because client messages are handled in batches, we need to store all responses for each batch.
I have changed the map that holds the replies to be a map of sequence numbers.
We can calculate which response is the oldest based on the sequence number since it is always increasing.
Whenever we receive a reply for a client, we save it, up to N(size of the client batch).

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -98,7 +98,11 @@ void ClientsManager::loadInfoFromReservedPages() {
       ConcordAssert(replyHeader->replyLength >= 0);
       ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
 
-    if (!loadReservedPage(getReplyFirstPageId(clientId), sizeOfReservedPage(), scratchPage_.data())) continue;
+      auto& info = clientsInfo_[clientId];
+      if (info.repliesInfo.size() >= maxNumOfReqsPerClient_) deleteOldestReply(clientId);
+      const auto& res = info.repliesInfo.insert_or_assign(replyHeader->reqSeqNum, replyPtr);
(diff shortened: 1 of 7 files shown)


bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -79,28 +83,38 @@ void ClientsManager::loadInfoFromReservedPages() {
...
+                            scratchPage_.data()))
+        break;
+      ClientReplyMsgHeader* replyHeader = (ClientReplyMsgHeader*)scratchPage_.data();
+      auto replyPtr = std::make_shared<ClientReplyMsg>(myId_, replyHeader->replyLength);
+      ConcordAssert(replyHeader->msgType == 0 || replyHeader->msgType == MsgCode::ClientReply);
+      ConcordAssert(replyHeader->currentPrimaryId == 0);
+      ConcordAssert(replyHeader->replyLength >= 0);
+      ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
+
+      // YS TBD: Multiple replies for client batching should be sorted by incoming time
```
yuliasherman, line 99: Please remove this comment as irrelevant anymore.

tests/config/test_comm_config.hpp
```diff
@@ -69,7 +69,7 @@ class TestCommConfig : public ITestCommConfig {
   // Network port of the first replica. Other replicas use ports
   // basePort + (2 * index).
   static const uint16_t base_port_ = 3710;
-  static const uint32_t buf_length_ = 128 * 1024;  // 128 kB
+  static const uint32_t buf_length_ = 1280 * 1024;  // 128 kB
```
yuliasherman, line 72: Fix the comment

PR #1893, opened 2021-09-16
Added support for spdlog.

HristoStaykov:
This is an initial commit. The logging works with spdlog, but
Mapped Diagnostic Context (MDC) is not supported yet and is
temporarily disabled for the builds with spdlog.

This integration is for initial performance comparison with log4cplus.


logging/include/LoggingSpd.hpp
```diff
@@ -0,0 +1,93 @@
...
+namespace logging {
+
+// typedef log4cplus::Logger Logger;
+using Logger = std::shared_ptr<spdlog::logger>;
+
+std::string get(const std::string& key);
+
+}  // namespace logging
+
+#define LOG_TRACE(l, s)                        \
```
arc-vmware, line 30: There is a LOG4CPLUS_TRACE/DEBUG etc family of macros. I think those also need to be taken care of.

arc-vmware (commented):
This is nice work @HristoStaykov . I have few questions which I asked in the comment.

CMakeLists.txt
```diff
@@ -19,8 +19,20 @@ if(NOT CMAKE_BUILD_TYPE)
...
 
-if (NOT DEFINED USE_LOG4CPP)
-  option(USE_LOG4CPP "Enable LOG4CPP" ON)
+option(USE_SPDLOG "Enable SPDLOG" OFF)
+option(USE_LOG4CPP "Enable LOG4CPP" OFF)
+option(USE_INTERNAL_LOGGER "Enable concord-bft's logger" OFF)
+
+if (USE_SPDLOG)
+    message("-- spdlog enabled")
+    set(USE_INTERNAL_LOGGER OFF)
```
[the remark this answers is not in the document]
HristoStaykov: it is a cmake option, if it is provided by the user it will be ON, in this case I set the other 2 logger options to off. The log4cplus remains the default one. We propagate this option as a define to the build in [logging/CMakeLists.txt:45](https://github.com/vmware/concord-bft/blob/a5f98ea568fe8560eab9fcca64a0ff0b3d859ec0/logging/CMakeLists.txt#L45)

logging/include/LoggingSpd.hpp
arc-vmware: So, if USE_LOG4CPP is not defined, then also the LOG4CPLUS_* macros are wrapped into LOG_*, since I see there definition only inside Logging4cplus.hpp

arc-vmware: This is used only in daml, so we are good.

arc-vmware (approved):
LGTM.

PR #1894, opened 2021-09-16
TRC: Fix legacy transition

robem:
If we start subscribing via a legacy request but get an event group from the
server then we will never advance our request parameters because we are still
requesting for legacy events. However, as soon as the server returns event
groups, we can switch to requesting further event groups because there cannot
be anymore legacy events left to consume.


shgandhi (approved):
LGTM!

PR #1895, opened 2021-09-16
TRC: Fix legacy transition 2

robem:
The prior change addressed the actual switch-over. This change resets the
latest verified event group id when a new legacy subscription occurs. Before
this change, we would carry over the latest verified event group id to new
subscriptions which implies that they got all those event groups already.


shgandhi (approved):
Thanks for the change @robem, LGTM!

PR #1898, opened 2021-09-19
Tls key exchange status messages implementation 

yontyon:
In this PR we present the following changes:
1. We add a timestamp to each reconfiguration block
2. We add a status message for client TLS key exchange
3. We used these status messages to make TLS key exchange in Apollo tests more reliable and robust


nkumar04 (approved):
Great PR!
LGTM!

PR #1901, opened 2021-09-21
Generate tsan/asan report

pkthapa:
This PR ensures asan & tsan reports are generated even if the CI run fails.
The apollo logs are not required in asan & tsan CI workflows, hence disabled. The reason
for disabling apollo logs is to prevent significant increase in apollo logs' size because the
apollo test cases run for 6 hours continuously (until Github CI 6-hour timer expires).
Hence, disabling apollo logs since we are only concerned about asan & tsan reports.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -34,7 +34,7 @@ jobs:
           run: |
               script -q -e -c "make pull"
               sudo df -h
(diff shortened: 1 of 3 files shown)


teoparvanov (approved):
LGTM, thanks Pankaj!

PR #1902, opened 2021-09-21
Docker update to version 0.31

robem:
In the clientservice, we want to expose a Prometheus port to query metrics.
Note, concord-bft has its own metrics implementation which we use but at the
outer-most API we want to expose it via a standard metrics solution.
Example usage in #1906


install_deps.sh
```diff
@@ -262,5 +264,30 @@ git clone -b v1.37.1 --depth 1 --recurse-submodules https://github.com/grpc/grpc
     cd ${HOME} && \
     rm -r ${HOME}/grpc
 
-# After installing all libraries, let's make sure that they will be found at runtime
-ldconfig
+# libcurl needed for Prometheus-cpp
+cd ${HOME}
+wget ${WGET_FLAGS} --no-check-certificate https://curl.se/download/curl-7.66.0.tar.xz && \
```
teoparvanov, line 269: Isn't `--no-check-certificate` unsafe? We don't seem to need it elsewhere in the `install_deps.sh` script when using wget...

teoparvanov (approved):
LGTM, thanks Rob! Just one minor comment on how we download one of the new dependencies...

install_deps.sh
robem: Ha. The certificate can be validated via `curl` :D I will change this line.

PR #1904, opened 2021-09-22
Add block id to the pre-execution batch

cloudnoize:
As a batch can be retried, we want to preserve the initial block-id that was set in the initial requests on any following requests that are sent as part of the batch.
In order to do so, the block id is saved in the batch and is mapped to a cid.
The mapping of cid->block_id is not being reset but only set on setParams in order to preserve the block_id on retry,

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -28,6 +28,7 @@ using namespace std::placeholders;
 using namespace concordUtil;
 
 uint8_t RequestState::reqProcessingHistoryHeight = 10;
+std::unordered_map<std::string, uint64_t> PreProcessor::to_block_id;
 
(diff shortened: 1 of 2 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1344,6 +1355,8 @@ bool PreProcessor::registerRequestOnPrimaryReplica(const string &batchCid,
   const auto clientId = clientReqMsg->clientProxyId();
   const auto senderId = clientReqMsg->senderId();
   const auto requestTimeoutMilli = clientReqMsg->requestTimeoutMilli();
+  const auto blockid = ongoingReqBatches_[clientId]->getBlockId();
```
yuliasherman, line 1358: ->blockId  :-)

yuliasherman (approved):
LGTM

PR #1905, opened 2021-09-22
Fix pre-execution requests batch cleanup in case of cancellation

yuliasherman:
In case of no consensus reached, pre-execution requests batch needs to be cleaned.


cloudnoize (approved):
Looks great, thanks @yuliasherman

PR #1906, opened 2021-09-22
Clientservice: Add Prometheus metrics server

robem:
This change leverages Prometheus, introduced in #1902, and adds the metrics server to clientservice. Please see commit messages for details.


shgandhi (approved):

yontyon (changes requested):
I don’t think we should add Prometheus to concordbft.
1. We prefer to have as fewer deps in concordbft as possible
1. We have an internal metrics system (developed by Andrew) for concordbft, I don’t think it.   makes sense to use two different systems
1. If we use Prometheus explicitly, we would lose the agility to choose different metrics systems.
The way we worked so far in concordbft, is to use Andrew’s metrics system and translate them to whatever metrics system that was chosen in the upper level (e.g concord). I don’t think the client service should work differently

PR #1908, opened 2021-09-22
Delete transaction_signing_keys folder if it already exists

shgandhi:
This commit deletes the transaction_signing_keys folder if it
already exists instead of erroring out and exiting.


jkk-vmw:
Thanks for doing this!

jkk-vmw (approved):
LGTM. I agree that only deleting when detected makes more sense. Thanks!

robem (commented):
Looks Ok to me. The only danger here is that we will overwrite existing keys without confirmation.

robem (approved):

PR #1909, opened 2021-09-23
Add support for client restart

nkumar04:
- Added new command to restart client in reconfiguration handler
- Added appolo test case for the same


yontyon (approved):
LGTM!
Thanks!

PR #1910, opened 2021-09-23
State transfer fix assert and add logs

glevkovich:
1) BCStateTran: replace an assert with graceful ignore on arriving message
2) BCStateTran: improve logs on BCStateTran::checkConsistency()

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -2777,11 +2777,10 @@ void BCStateTran::checkConsistency(bool checkAllBlocks, bool duringInit) {
   const uint64_t lastReachableBlockNum = as_->getLastReachableBlockNum();
   const uint64_t lastBlockNum = as_->getLastBlockNum();
   const uint64_t genesisBlockNum = as_->getGenesisBlockNum();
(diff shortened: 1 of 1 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2829,6 +2830,7 @@ void BCStateTran::checkFirstAndLastCheckpoint(uint64_t firstStoredCheckpoint, ui
 
 void BCStateTran::checkReachableBlocks(uint64_t genesisBlockNum, uint64_t lastReachableBlockNum) {
   if (lastReachableBlockNum > 0) {
+    LOG_INFO(logger_, "");
```
mdarade, line 2833: Instead of putting empty logs, can we log input parameters?

mdarade:
Thanks Gil.
Just minor comment.
Looks good.
Approved.

mdarade (approved):

bftengine/src/bcstatetransfer/BCStateTran.cpp
glevkovich: Yes, I can do that, although almost all parameters are already printed outside in checkConsistency(), it would be more convenient.
Thanks.

PR #1913, opened 2021-09-23
Add metric for duration of delete blocks batch during pruning

nkgeorgiev:
This metric will be useful when debugging pruning performance issues


kvbc/src/Replica.cpp
```diff
@@ -352,9 +352,13 @@ BlockId Replica::deleteBlocksUntil(BlockId until) {
 
   const auto lastReachableBlock = m_kvBlockchain->getLastReachableBlockId();
   const auto lastDeletedBlock = std::min(lastReachableBlock, until - 1);
+  auto start = std::chrono::steady_clock::now();
```
teoparvanov, line 355: const?

teoparvanov (approved):
LGTM! @nkgeorgiev please rebase to take a recent regression fix from master.

dartdart26 (approved):

PR #1916, opened 2021-09-24
Revert "Revert "Performance Improvement : Preprepare message validati…

arc-vmware:
…on""

This reverts commit ed41a87154ec850520ac242964590d46f63537af.
Creating the pull request again and fix the reconfiguration test.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -43,6 +43,8 @@
 #include "messages/ReplicaRestartReadyMsg.hpp"
 #include "messages/ReplicasRestartReadyProofMsg.hpp"
 #include "messages/PreProcessResultMsg.hpp"
+#include "messages/ViewChangeIndicatorInternalMsg.hpp"
+#include "messages/PrePrepareCarrierInternalMsg.hpp"
 #include "CryptoManager.hpp"
 #include "ControlHandler.hpp"
 #include "bftengine/KeyExchangeManager.hpp"
@@ -134,14 +136,24 @@ void ReplicaImp::messageHandler(MessageBase *msg) {
       return;
     }
   }
-  if (validateMessage(trueTypeObj) && !isCollectingState())
-    onMessage<T>(trueTypeObj);
-  else
-    delete trueTypeObj;
-}
 
-template <class T>
-void onMessage(T *);
+  if constexpr (std::is_same_v<T, PrePrepareMsg>) {
+    if (getReplicaConfig().prePrepareFinalizeAsyncEnabled) {
+      if (!isCollectingState()) {
+        validatePrePrepareMsg(trueTypeObj);
+        return;
+      } else {
+        delete trueTypeObj;
+        return;
+      }
+    }
+  }
+  if (!isCollectingState() && validateMessage(trueTypeObj)) {
+    onMessage<T>(trueTypeObj);
+  } else {
+    delete trueTypeObj;
+  }
+}
 
 bool ReplicaImp::validateMessage(MessageBase *msg) {
   if (config_.debugStatisticsEnabled) {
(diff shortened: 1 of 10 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -134,15 +136,19 @@ void ReplicaImp::messageHandler(MessageBase *msg) {
       return;
     }
   }
-  if (validateMessage(trueTypeObj) && !isCollectingState())
+  if constexpr (std::is_same_v<T, PrePrepareMsg>) {
```
yontyon, line 139: You cannot handle any message while state transfer. Please add here this condition
```
if(!isCollectingState())
```

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1274,6 +1312,31 @@ void ReplicaImp::onInternalMsg(InternalMessage &&msg) {
   if (auto *ppm = std::get_if<PrePrepareMsg *>(&msg)) {
     return startConsensusProcess(*ppm, true);
   }
+
+  // Handle a the PrePrepare Carrier
+  if (auto *ppcim = std::get_if<PrePrepareCarrierInternalMsg>(&msg)) {
+    if (bftEngine::ControlStateManager::instance().getPruningProcessStatus()) {
```
yontyon, line 1318: Here also, you cannot handle any message while state transfer. You should reject and delete the message if the replica is collecting state
e.g.
```
if(!isCollectingState())
```

yontyon (changes requested):

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Thanks @yontyon for your comment. Let me add this check.

yuliasherman (approved):
LGTM

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Thanks @yontyon , yes I am now keeping the isCollectingState() check all the places before the onMessage() function call.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -467,7 +513,9 @@ PrePrepareMsg *ReplicaImp::createPrePrepareMessage() {
     return nullptr;
   }
 
-  controller->onSendingPrePrepare((primaryLastUsedSeqNum + 1), firstPath);
+  if (!getReplicaConfig().prePrepareFinalizeAsyncEnabled) {
```
arc-vmware, line 516: @yontyon Please review if this change is good enough.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -630,6 +678,7 @@ void ReplicaImp::startConsensusProcess(PrePrepareMsg *pp, bool isCreatedEarlier)
   metric_bft_batch_size_.Get().Set(pp->numberOfRequests());
   primaryLastUsedSeqNum++;
   if (isCreatedEarlier) {
+    controller->onSendingPrePrepare(primaryLastUsedSeqNum, firstPath);
```
arc-vmware, line 681: @yontyon : Please review if this is right todo.

tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -67,7 +67,6 @@ std::unique_ptr<TestSetup> TestSetup::ParseArgs(int argc, char** argv) {
     replicaConfig.clientBatchingEnabled = true;
     replicaConfig.pruningEnabled_ = true;
     replicaConfig.numBlocksToKeep_ = 10;
-    replicaConfig.timeServiceEnabled = true;
```
yontyon, line 70: I don't think we should disable time service in concordbft.
Our issues with the new time service were with the product. I think that in order to keep it updated and not break it in concordbft, we should leave it enabled

arc-vmware: @yontyon  Time service is enabled in concord-bft. Its enabled using the -f parameter in the testcase. After the the timeservice v2 is fully implemented and tested, it will be enabled by default. I am code reviewing that part, so I will make sure that its switched on for all. Now its switched on for skvbc_reconfiguration as it requires time service and its tested.

yontyon (approved):
LGTM
Thanks

PR #1920, opened 2021-09-27
Fix for gcc build failure

itsmeakashgoyal:
github action workflow for was failing for gcc build.
Fix has been implemented and the same can be verified from below actions,
https://github.com/itsmeakashgoyal/concord-bft/actions/runs/1277533663
https://github.com/itsmeakashgoyal/concord-bft/actions/runs/1277533659


teoparvanov:
Hi Akash, thanks so much for fixing this!

teoparvanov (approved):
LGTM!

PR #1923, opened 2021-09-30
Replicas' TLS key exchange - basic implementation 

yontyon:
In this PR we introduce a basic implementation for replicas' TLS key exchange.
A replica TLS key exchange is composed of two phases:
1. The operator sends a request with up to f replicas that we want to exchange (see below why we don't exchange for more than f replicas at once)
2. Once the request is executed, each replica in the target_replicas generates a new key-pair and certificate and sends another BFT request that contains the new certificate.
3. On this second request execution, the replica saves the new certificate and closes the old TLS channel to the exchanged replica.

Note: 
We have a basic problem with TLS replicas' TLS key exchange: While the replicas in the lead quorum replace the key on the same stage, the clients need to fetch the data from the replicas in order to exchange the replicas' key.
However, once the replicas changed their TLS keys, the clients cannot connect to the replica and fetch the data.
To overcome the problem we don't allow to rotate more than f replicas at once. This way, both, CRE and ST are still able to work with the remaining quorum. It is the user's responsibility to verify that the clients were able to get the replicas key exchange command before exchanging more keys.
Note, that in case we have lagged/ST replicas, it is the client's responsibility to include these replicas in the exchanged f replicas (otherwise, we won't have a functional quorum).

To complete the feature we need to implement also the following:
1. ST handler for this command
2. Client's reconfiguration handler for this command

### DIFF
--- a/bftengine/src/bftengine/KeyExchangeManager.cpp
+++ b/bftengine/src/bftengine/KeyExchangeManager.cpp
@@ -150,7 +150,7 @@ void KeyExchangeManager::exchangeTlsKeys(const SeqNum& bft_sn) {
   std::string cid = "replicaTlsKeyExchange_" + std::to_string(bft_sn) + "_" + std::to_string(repId);
   client_->sendRequest(RECONFIG_FLAG, strMsg.size(), strMsg.c_str(), cid);
   metrics_->tls_key_exchange_requests_++;
-  LOG_INFO(KEY_EX_LOG, "Replica have generated a new tls keys");
+  LOG_INFO(KEY_EX_LOG, "Replica has generated a new tls keys");
 }
 void KeyExchangeManager::sendKeyExchange(const SeqNum& sn) {
   // first check whether we've already generated keys lately
--- a/bftengine/src/bftengine/KeyExchangeManager.cpp
+++ b/bftengine/src/bftengine/KeyExchangeManager.cpp
@@ -15,6 +15,8 @@
 #include "kvstream.h"
 #include "json_output.hpp"
 #include "SigManager.hpp"
+#include "secrets_manager_plain.h"
+#include "concord.cmf.hpp"
 #include <thread>
 
 namespace bftEngine::impl {
@@ -28,7 +30,8 @@ KeyExchangeManager::KeyExchangeManager(InitData* id)
       client_(id->cl),
       multiSigKeyHdlr_(id->kg),
       clientPublicKeyStore_{id->cpks},
-      timers_(*(id->timers)) {
+      timers_(*(id->timers)),
+      secretsMgr_(id->secretsMgr) {
   registerForNotification(id->ke);
   notifyRegistry();
   if (!ReplicaConfig::instance().getkeyExchangeOnStart()) initial_exchange_ = true;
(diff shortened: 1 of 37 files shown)


communication/include/communication/CommStateControl.hpp
```diff
@@ -11,6 +11,8 @@
 // LICENSE file.
 
 #pragma once
+#include <functional>
+#include "callback_registry.hpp"
 namespace bft::communication {
 class CommStateControl {
```
toly-kournik, line 17: it's in `bft::communication` namespace (why not `concord::bft::communication`?), so `Comm` is redundant

communication/include/communication/CommStateControl.hpp
```diff
@@ -21,7 +23,14 @@ class CommStateControl {
   void setBlockNewConnectionsFlag(bool flag) { blockNewConnectionsFlag_ = flag; }
   bool getBlockNewConnectionsFlag() { return blockNewConnectionsFlag_; }
```
toly-kournik, line 24: const method

kvbc/src/Replica.cpp
```diff
@@ -431,6 +432,10 @@ Replica::Replica(ICommunication *comm,
       secretsManager_{secretsManager},
       blocksIOWorkersPool_((replicaConfig.numWorkerThreadsForBlockIO > 0) ? replicaConfig.numWorkerThreadsForBlockIO
                                                                           : std::thread::hardware_concurrency()) {
+  bft::communication::CommStateControl::instance().setCommRestartCallBack([this]() {
+    this->m_ptrComm->Stop();
```
toly-kournik: style:
`m_ptrComm` -> comm_
`Start()` and `Stop()` -> lowercase

reconfiguration/include/reconfiguration/reconfiguration_handler.hpp
```diff
@@ -19,6 +19,7 @@
 #include "OpenTracing.hpp"
 #include "SigManager.hpp"
 #include "crypto_utils.hpp"
+#include "bftengine/InternalBFTClient.hpp"
```
toly-kournik, line 22: why did you add this include?

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -113,6 +116,42 @@ void KeyExchangeManager::loadPublicKeys() {
...
+  std::fstream nec_f(pk_path);
+  if (nec_f.good()) {
+    psm_.encryptFile(pk_path, keys.first);
+  }
+  std::fstream ec_f(pk_path + ".enc");
+  if (ec_f.good()) {
+    secretsMgr_->encryptFile(pk_path + ".enc", keys.first);
+  }
+
+  concord::messages::ReconfigurationRequest req;
```
toly-kournik, line 137: Why to encapsulate `ReplicaTlsExchangeKey` in a `ReconfigurationRequest`?
`ReconfigurationRequest` is usually sent by an operator.

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -113,6 +116,42 @@ void KeyExchangeManager::loadPublicKeys() {
...
 
+void KeyExchangeManager::exchangeTlsKeys(const SeqNum& bft_sn) {
+  auto repId = bftEngine::ReplicaConfig::instance().replicaId;
+  auto keys = concord::util::crypto::Crypto::instance().generateECDSAKeyPair(
+      concord::util::crypto::KeyFormat::PemFormat, concord::util::crypto::CurveType::secp384r1);
+  auto cert = concord::util::crypto::Crypto::instance().generateSelfSignedCertificate(
+      keys, "node" + std::to_string(repId) + "ser", repId);
+  std::string pk_path =
+      bftEngine::ReplicaConfig::instance().certificatesRootPath + "/" + std::to_string(repId) + "/server/pk.pem";
+  concord::secretsmanager::SecretsManagerPlain psm_;
```
toly-kournik, line 127: We don't need to decide which secrets manager to use, we work with a provided one

tests/simpleKVBC/TesterReplica/setup.hpp
```diff
@@ -36,7 +37,7 @@ class TestSetup {
   static std::unique_ptr<TestSetup> ParseArgs(int argc, char** argv);
 
   std::unique_ptr<IStorageFactory> GetStorageFactory();
-
+  std::shared_ptr<concord::secretsmanager::ISecretsManagerImpl> GetSecretManager() { return sm_; }
```
toly-kournik, line 40: const

toly-kournik:
How replicas which were down during the tls key exchange would know that the key has been exchanged (hint: reserved pages?)

communication/include/communication/CommStateControl.hpp
yontyon: fixed

communication/include/communication/CommStateControl.hpp
yontyon: fixed

kvbc/src/Replica.cpp
yontyon: fixed

reconfiguration/include/reconfiguration/reconfiguration_handler.hpp
yontyon: fixed

bftengine/src/bftengine/KeyExchangeManager.cpp
yontyon: fixed

tests/simpleKVBC/TesterReplica/setup.hpp
yontyon: fixed

bftengine/src/bftengine/KeyExchangeManager.cpp
yontyon: As a response to the operator `reconfiguraitonRequest`, the exchanged replicas send another `reconfigurationRequest` that contains the new certificate.
This is to avoid having a different execution handler for this method (like we have with regular key exchange) and to work with a single framework

yontyon:
@toly-kournik , as for state transfer, after merging this PR we will post another one. 
The method is to scan, once we finished state transfer, for the latest block that contains the tls reconfiguration request and execute it if needed (if it wasn't already executed by the replica).
We use this method in all our reconfiguration requests (like, wedge, addRemove, and more)

reconfiguration/src/reconfiguration_handler.cpp
```diff
@@ -52,18 +53,26 @@ bool ReconfigurationHandler::handle(const KeyExchangeCommand& command,
                                     uint64_t sequence_number,
                                     uint32_t,
                                     const std::optional<bftEngine::Timestamp>&,
-                                    concord::messages::ReconfigurationResponse&) {
+                                    concord::messages::ReconfigurationResponse& rres) {
+  if (command.tls && command.target_replicas.size() > bftEngine::ReplicaConfig::instance().fVal) {
```
nkumar04, line 57: what happens when we have only (n-f) functional nodes? should we ignore replica tls key exchange?

toly-kournik (approved):
LVGTM!

reconfiguration/src/reconfiguration_handler.cpp
yontyon: There are two aspects here:
(1) whether to run the command at al - in this case, this is the user responsibility to update the clients correctly (or wait for the f faulty node to return)
(2) How the late replicas are catching up - this will be done by the state transfer reconfiguration mechanism
as for (1), the point is that replicas cannot know (at least without very complicated BFT failure detectors) that other replicas are not available
They can use some heuristics, but I prefer to leave it for the user's decision (at least for the fears phase). In the future, we can implement a timed exchanged mechanism (in which the replica runs the actual exchange after a predefined amount of time). But notice that even this mechanism will not solve the problem of n-f functional nodes if the client wasn’t able to get the update in the predefined timeout

PR #1924, opened 2021-09-30
Identify the case where we restart during pruning

yontyon:
When pruning run in the background, we may delete old overridden values (like wedge key) from the database.
For this, we need to remove the assert that assumes we always have the wedge value in the blockchain.


nkumar04 (approved):
LGTM!!

PR #1927, opened 2021-09-30
Add metric for duration of delete blocks batch during pruning (#1913)

nkgeorgiev:
* Add metric for duration ot delete block batch during pruning

* make 'now' variable const


dartdart26 (approved):
Thanks!

PR #1928, opened 2021-09-30
Remove duplicate definition of kConcordInternalCategoryId

shgandhi:
This commit keeps the internal category definition in db_categories.h (with the rest of the category definitions) and removes the one in db_interfaces.h. The change is reflected in all the files that use internal category.


dartdart26 (approved):
Thanks!

PR #1929, opened 2021-10-01
Fixing the data type of maxSpaceForReqs and review of primaryCombinedReqSize

arc-vmware:
Following 2 issues are handled in this PR
1) The datatype of maxSpaceForReqs is uint_16. This variable
stores the size of max space needed for the PrePrepareMsg.
So this variable can limit the size of PPM to 64KB.
2) The primaryCombinedReqSize variable should keep track of
actual total size of ClientRequestMsg. So every time we push CRM
into PPM, we have to increment it. And every time we pop CRM out of PPM
we have to reduce it with the removed CRM size. In this PR we will
review and change if we see some issues.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -269,8 +269,8 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
                                                         << "], senderId=" << senderId);
         if (time_to_collect_batch_ == MinTime) time_to_collect_batch_ = getMonotonicTime();
         requestsQueueOfPrimary.push(m);
-        primary_queue_size_.Get().Set(requestsQueueOfPrimary.size());
         primaryCombinedReqSize += m->size();
+        primary_queue_size_.Get().Set(requestsQueueOfPrimary.size());
         tryToSendPrePrepareMsg(true);
         return;
       } else {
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -269,8 +269,8 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
                                                         << "], senderId=" << senderId);
         if (time_to_collect_batch_ == MinTime) time_to_collect_batch_ = getMonotonicTime();
         requestsQueueOfPrimary.push(m);
-        primary_queue_size_.Get().Set(requestsQueueOfPrimary.size());
         primaryCombinedReqSize += m->size();
+        primary_queue_size_.Get().Set(requestsQueueOfPrimary.size());
```
yuliasherman, line 273: Why did you change the order of operations?

yuliasherman (approved):
LGTM

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: I am trying to follow the following conventions:
1) Whenever we push into requestsQueueOfPrimary, we update the (byte)size in  primaryCombinedReqSize.
2) Whenever we pop from the requestsQueueOfPrimary we do the following:
    a) Access the front() of the queue in some variable say 'f' (its a shared ptr to CRM).
    b) Update the (byte)size in  primaryCombinedReqSize.
    c) Pop the queue.
    d) delete the shared ptr.

PR #1930, opened 2021-10-01
TRC: Fix log spew delay

robem:
So far we only delayed the second message but kept spewing logs afterwards.
This change fixes this by resetting the last log time.


shgandhi (approved):
Thanks for fixing this @robem! LGTM.

PR #1935, opened 2021-10-05
Fixing the difference in blockchain due to indeterminism

arc-vmware:
This fix removes the race between execute of preprepare message
and Pruning. During the execution of preprepare message, pruning
check is added to avoid the execution. But the check is kept in
such a manner that commit of pre prepare message is happening.


yontyon (approved):
LGTM!
Thanks!

PR #1941, opened 2021-10-05
Fix the case of "Protobuf" in a CMake file.

SamirTalwar-DA:
On my machine, using the incorrect case means that CMake can't find the
appropriate package, because the file it's looking for is called
"FindProtobuf.cmake", not "FindPROTOBUF.cmake".


robem (approved):
The docker image has the package installed as `FindProtobuf.cmake`. Hence, this change makes sense. Thanks for the contribution! 🚀

SamirTalwar-DA:
Can we merge this?

PR #1942, opened 2021-10-06
State transfer functionality for TLS key exchange

yontyon:
In this PR we introduce the replicas logic for handling TLS key exchange after state transfer (including Apollo tests).
Note that client reconfiguration for replicas TLS keys and Read only replica reconfiguration are not part of this PR

### DIFF
--- a/tests/apollo/test_skvbc_reconfiguration.py
+++ b/tests/apollo/test_skvbc_reconfiguration.py
@@ -251,6 +251,58 @@ class SkvbcReconfigurationTest(unittest.TestCase):
 
             for i in range(100):
                 await skvbc.send_write_kv_set()
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: n == 7, with_cre=True)
+    async def test_client_tls_key_exchange_command_with_st(self, bft_network):
+        """
+            Operator sends client key exchange command for cre client
+        """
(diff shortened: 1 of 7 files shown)


tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -410,9 +462,76 @@ async def test_replicas_tls_key_exchange_command_without_primary(self, bft_netwo
...
+    async def test_replicas_tls_key_exchange_command_with_st(self, bft_network):
+        """
+            Operator sends client key exchange command for f - 1 replicas
+        """
+        with log.start_action(action_type="test_replica_key_exchange_command"):
+            bft_network.start_all_replicas()
+            skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+            initial_prim = 0
+            crashed_replica = list(bft_network.random_set_of_replicas(1, {initial_prim}))
+            exchanged_replicas = list(bft_network.random_set_of_replicas(bft_network.config.f - 2, {initial_prim, crashed_replica[0]}))
```
nkumar04, line 476: here exchanged_replicas list is empty? (because of f-2)

yontyon: Yes, but we later add to it the current primary (tp make sure that the primary key exchange was done by the state transfer handler)

PR #1946, opened 2021-10-06
Print a separate log message when no updates generated by replicas

shgandhi:
The client will not have any agreeing servers, if no updates are getting
generated by the replicas. This commit makes the distinction between
the two scenarios i.e., no agreement b/w replicas and no updates getting
generated by replicas explicit via log messages.
If readBlock doesn't result in any update, and findBlockHashAgreement
method, which opens hash streams to all the replicas to check for
maximal agreement, times out waiting for hash updates at all the replicas,
we can say that no updates are being generated by any of the replicas.
This commit updates the findBlockHashAgreement method s.t., we can
check if all the hash streams opened timed out waiting for a hash.

### DIFF
--- a/client/thin-replica-client/include/client/thin-replica-client/thin_replica_client.hpp
+++ b/client/thin-replica-client/include/client/thin-replica-client/thin_replica_client.hpp
@@ -243,7 +243,13 @@ class ThinReplicaClient final {
                                      size_t& most_agreeing,
                                      HashRecord& most_agreed_block,
                                      std::unique_ptr<LogCid>& cid);
-  void findBlockHashAgreement(std::vector<bool>& servers_tried,
+
+  // Opens hash streams to all the replicas and tries to read hash updates
+  // from opened streams to check for maximal agreement.
+  // If none of the replicas return a hash update i.e., the connections to
+  // all the replicas either time out or fail while waiting for an update,
+  // findBlockHashAgreement returns false, otherwise returns true.
+  bool findBlockHashAgreement(std::vector<bool>& servers_tried,
                               HashRecordMap& agreeing_subset_members,
                               size_t& most_agreeing,
                               HashRecord& most_agreed_block,
@@ -268,7 +274,9 @@ class ThinReplicaClient final {
                            HashRecordMap& server_indexes_by_reported_update,
                            size_t& maximal_agreeing_subset_size,
                            HashRecord& maximally_agreed_on_update);
-  void readUpdateHashFromStream(size_t server_index,
+
+  // Returns true if a hash update is received from a hash stream, returns false otherwise
+  bool readUpdateHashFromStream(size_t server_index,
                                 HashRecordMap& server_indexes_by_reported_update,
                                 size_t& maximal_agreeing_subset_size,
                                 HashRecord& maximally_agreed_on_update);
--- a/client/thin-replica-client/src/thin_replica_client.cpp
+++ b/client/thin-replica-client/src/thin_replica_client.cpp
@@ -96,7 +96,7 @@ void ThinReplicaClient::recordCollectedHash(size_t update_source,
   }
 }
 
-void ThinReplicaClient::readUpdateHashFromStream(size_t server_index,
+bool ThinReplicaClient::readUpdateHashFromStream(size_t server_index,
                                                  HashRecordMap& server_indexes_by_reported_update,
                                                  size_t& maximal_agreeing_subset_size,
                                                  HashRecord& maximally_agreed_on_update) {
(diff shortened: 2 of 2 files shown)


robem (changes requested):

client/thin-replica-client/include/client/thin-replica-client/thin_replica_client.hpp
```diff
@@ -243,7 +243,7 @@ class ThinReplicaClient final {
                                      size_t& most_agreeing,
                                      HashRecord& most_agreed_block,
                                      std::unique_ptr<LogCid>& cid);
-  void findBlockHashAgreement(std::vector<bool>& servers_tried,
+  bool findBlockHashAgreement(std::vector<bool>& servers_tried,
```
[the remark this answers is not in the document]
shgandhi: Done.

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -304,13 +308,20 @@ void ThinReplicaClient::findBlockHashAgreement(std::vector<bool>& servers_tried,
       }
     }
 
-    readUpdateHashFromStream(server_index, agreeing_subset_members, most_agreeing, most_agreed_block);
+    readUpdateHashFromStream(
+        server_index, agreeing_subset_members, most_agreeing, most_agreed_block, unsuccessful_hash_stream_subset_size);
```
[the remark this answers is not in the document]
shgandhi: Done.

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -483,24 +495,34 @@ void ThinReplicaClient::receiveUpdates() {
...
     if (stop_subscription_thread_) {
       break;
     }
 
     // At this point we need to have agreeing servers.
     if (most_agreeing < (config_->max_faulty + 1)) {
       // print the warning every minute to avoid flooding with logs
-      std::string msg = "Couldn't find agreement amongst all servers. Try again.";
+      std::string no_agreement_msg = "Couldn't find agreement amongst all servers. Try again.";
+      std::string no_update_msg = "No new update received from replica" + std::to_string(data_conn_index_);
```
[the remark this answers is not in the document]
shgandhi: Thanks for catching, corrected.

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -96,23 +96,26 @@ void ThinReplicaClient::recordCollectedHash(size_t update_source,
   }
 }
 
-void ThinReplicaClient::readUpdateHashFromStream(size_t server_index,
+bool ThinReplicaClient::readUpdateHashFromStream(size_t server_index,
                                                  HashRecordMap& server_indexes_by_reported_update,
                                                  size_t& maximal_agreeing_subset_size,
                                                  HashRecord& maximally_agreed_on_update) {
   Hash hash;
+  bool has_hash = true;
```
robem, line 104: We only ever change this variable if we are about to return. So I suggest to simply `return true;` or `return false;` instead of this variable. Given that the return value is documented the code becomes clear.

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -304,13 +309,20 @@ void ThinReplicaClient::findBlockHashAgreement(std::vector<bool>& servers_tried,
...
     servers_tried[server_index] = true;
 
     if (most_agreeing >= (config_->max_faulty + 1)) {
-      return;
+      return true;
     }
   }
+  auto all_servers_tried =
+      std::all_of(servers_tried.begin(), servers_tried.end(), [](bool elem) { return elem == true; });
+  if (all_servers_tried && unsuccessful_hash_stream_subset_size > servers_tried.size()) {
```
robem, line 322: Shouldn't it be `>= servers_tried.size()`? E.g., we tried 7 servers and 7 servers returned no hash update. Also, can the amount of failed hashes even be greater than the amount of servers if we only query each server once?

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -483,24 +496,34 @@ void ThinReplicaClient::receiveUpdates() {
...
     if (stop_subscription_thread_) {
       break;
     }
 
     // At this point we need to have agreeing servers.
     if (most_agreeing < (config_->max_faulty + 1)) {
       // print the warning every minute to avoid flooding with logs
-      std::string msg = "Couldn't find agreement amongst all servers. Try again.";
+      std::string no_agreement_msg = "Couldn't find agreement amongst all servers. Try again.";
+      std::string no_update_msg = "No new update from replica";
```
robem, line 509: Either plural "replicas" or "any replica".

client/thin-replica-client/src/thin_replica_client.cpp
shgandhi: Updated.

client/thin-replica-client/src/thin_replica_client.cpp
shgandhi: Updated.

client/thin-replica-client/src/thin_replica_client.cpp
shgandhi: Yes, thanks, updated.

shgandhi:
Manually tested and updated the PR.

PR #1952, opened 2021-10-08
Coverity fix - use of pointer after delete

nkumar04:
This change is fix a coverity bug in function ReplicaRestartReadyMsg handler, where pointer is used after free


arc-vmware (approved):
LGTM!!

PR #1953, opened 2021-10-08
Coverity bug fix

nkumar04:
PreProcessBatchRequestMsg and ClientBatchRequestMsg contains a batch of client messages and each message is parsed from the message body for requestLen, signatureLen etc. Theses len parameters are used a offset to point to the message buffer. Coverity tool complains that sanity/range check is not done on these len parameters and they are used in allocation of buffer for these message in PreProcessBatchRequestMsg::getPreProcessRequestMsgs and ClientBatchRequestMsg::getClientPreProcessRequestMsgs.
This PR, adds a function to validate these len parameters

### DIFF
--- a/bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
+++ b/bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
@@ -60,6 +60,41 @@ void ClientBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
     LOG_ERROR(logger(), "Message sender is invalid" << KVLOG(senderId()));
     throw std::runtime_error(__PRETTY_FUNCTION__);
   }
+  if (!checkElements()) {
+    LOG_ERROR(logger(), "One or more ClientMsg in the list is invalid");
+    throw std::runtime_error(__PRETTY_FUNCTION__);
+  }
+}
+
+bool ClientBatchRequestMsg::checkElements() const {
+  const auto totalMsgSize = size();
+  const auto& numOfMessagesInBatch = msgBody()->numOfMessagesInBatch;
+  if (!numOfMessagesInBatch || (numOfMessagesInBatch > MAX_BATCH_SIZE)) {
+    LOG_ERROR(logger(), KVLOG(numOfMessagesInBatch));
+    return false;
+  }
+  char* dataPosition = body() + sizeof(ClientBatchRequestMsgHeader) + msgBody()->cidSize;
+  const auto& sigManager = SigManager::instance();
+  const auto& isClientTransactionSigningEnabled = sigManager->isClientTransactionSigningEnabled();
+  for (auto i = 0u; i < numOfMessagesInBatch; i++) {
+    const auto& singleMsgHeader = *(ClientRequestMsgHeader*)dataPosition;
+    PrincipalId clientId = singleMsgHeader.idOfClientProxy;
+    auto expectedSigLen = (isClientTransactionSigningEnabled ? sigManager->getSigLength(clientId) : 0);
+    if ((expectedSigLen != singleMsgHeader.reqSignatureLength) || (totalMsgSize < singleMsgHeader.requestLength) ||
+        (totalMsgSize < singleMsgHeader.cidLength)) {
+      LOG_ERROR(logger(),
+                KVLOG(clientId,
+                      totalMsgSize,
+                      expectedSigLen,
+                      singleMsgHeader.reqSignatureLength,
+                      singleMsgHeader.requestLength,
+                      singleMsgHeader.cidLength));
+      return false;
+    }
+    dataPosition += sizeof(ClientRequestMsgHeader) + singleMsgHeader.spanContextSize + singleMsgHeader.requestLength +
+                    singleMsgHeader.cidLength + singleMsgHeader.reqSignatureLength;
+  }
+  return true;
 }
 
 ClientMsgsList& ClientBatchRequestMsg::getClientPreProcessRequestMsgs() {
--- a/bftengine/src/preprocessor/messages/ClientBatchRequestMsg.hpp
+++ b/bftengine/src/preprocessor/messages/ClientBatchRequestMsg.hpp
@@ -51,6 +51,7 @@ class ClientBatchRequestMsg : public MessageBase {
     static logging::Logger logger_ = logging::getLogger("concord.preprocessor");
     return logger_;
   }
+  bool checkElements() const;
   std::string cid_;
   ClientMsgsList clientMsgsList_;
 };
--- a/bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.cpp
+++ b/bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.cpp
@@ -11,7 +11,7 @@
 
 #include "PreProcessBatchRequestMsg.hpp"
 #include "assertUtils.hpp"
-
+#include "SigManager.hpp"
 namespace preprocessor {
 
 using namespace std;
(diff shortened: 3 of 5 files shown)


bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.hpp
```diff
@@ -65,6 +65,7 @@ class PreProcessBatchRequestMsg : public MessageBase {
   Header* msgBody() const { return ((Header*)msgBody_); }
 
  private:
+  inline bool checkElements() const;
```
glevkovich, line 68: I think you should remove the inline keyword, as the function is not defined in the header.

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
```diff
@@ -60,6 +60,36 @@ void ClientBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
     LOG_ERROR(logger(), "Message sender is invalid" << KVLOG(senderId()));
     throw std::runtime_error(__PRETTY_FUNCTION__);
   }
+  if (!checkElements()) {
+    LOG_ERROR(logger(), "One or more ClientMsg in the list is invalid");
+    throw std::runtime_error(__PRETTY_FUNCTION__);
+  }
+}
```
glevkovich, line 67: can you please add an empty line between functions?

bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.cpp
```diff
@@ -52,6 +52,36 @@ void PreProcessBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
     LOG_ERROR(logger(), "Message sender is invalid" << KVLOG(senderId()));
     throw std::runtime_error(__PRETTY_FUNCTION__);
   }
+  if (!checkElements()) {
+    LOG_ERROR(logger(), "One or more PreProcessReqMsg in the list is invalid");
+    throw std::runtime_error(__PRETTY_FUNCTION__);
+  }
+}
```
glevkovich, line 59: can you please add a line space here?

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
```diff
@@ -60,6 +60,36 @@ void ClientBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
...
+bool ClientBatchRequestMsg::checkElements() const {
+  const auto totalMsgSize = size();
+  const auto& numOfMessagesInBatch = msgBody()->numOfMessagesInBatch;
```
glevkovich, line 70: I think we should consider a macro MAX_NUM_ELEMENT_IN_BATCH. You can set it for now to 200 for example, but maybe reserch a little more (speak with some other people) on the exact max value. If you are checking for validity, every data can be corrupted. Therfor, also the number of messages. If this number is 0 or too big, we should also drop.

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
```diff
@@ -60,6 +60,36 @@ void ClientBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
...
+  char* dataPosition = body() + sizeof(ClientBatchRequestMsgHeader) + msgBody()->cidSize;
+  auto sigManager = SigManager::instance();
+  bool isClientTransactionSigningEnabled = sigManager->isClientTransactionSigningEnabled();
+  for (auto i = 0u; i < numOfMessagesInBatch; i++) {
+    const auto& singleMsgHeader = *(ClientRequestMsgHeader*)dataPosition;
```
glevkovich, line 75: Why copying? Can't you just use a const pointer?

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
```diff
@@ -60,6 +60,36 @@ void ClientBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
...
+    PrincipalId clientId = singleMsgHeader.idOfClientProxy;
+    auto expectedSigLen = (isClientTransactionSigningEnabled ? sigManager->getSigLength(clientId) : 0);
```
glevkovich, line 77: just a comment: are you aware that tx signing can be enabled, and clientid is not recognized, and the function getSigLength will return 0? The resut will be the same (go out with error) but for different reason - bad client id. Luckily, an error is printed inside the function so you don't need to do anything special here.

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
```diff
@@ -60,6 +60,36 @@ void ClientBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
...
+    if (expectedSigLen != singleMsgHeader.reqSignatureLength || totalMsgSize < singleMsgHeader.requestLength ||
```
glevkovich, line 78: Is it possible to put in all complex conditions brackets around? I think it makes the code more readble...and reduce amounts of errors..

bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.cpp
```diff
@@ -52,6 +52,36 @@ void PreProcessBatchRequestMsg::validate(const ReplicasInfo& repInfo) const {
...
+bool PreProcessBatchRequestMsg::checkElements() const {
```
glevkovich, line 60: Is there any common code you detect? If yes, maybe we should consider putting another common class between
PreProcessBatchRequestMsg/ClientBatchRequestMsg and MessageBase? 
These 2 functions are almost identical?

Same comments like in ClientBatchRequestMsg::checkElements()

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.hpp
```diff
@@ -51,6 +51,7 @@ class ClientBatchRequestMsg : public MessageBase {
     static logging::Logger logger_ = logging::getLogger("concord.preprocessor");
     return logger_;
   }
+  inline bool checkElements() const;
```
glevkovich, line 54: can remove inline keyword.

glevkovich (changes requested):
@nkumar04 , Thanks for these changes, I think it's really important to validate these 2 messages. - to avoid byzanite peer replica which sends corrupted data + in term of stability.
Wrote some  few comments.

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
nkumar04: Thanks for the info

bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.cpp
nkumar04: fixed

bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.cpp
glevkovich: Not sure I understood. do you want to discuss that on zoom?

bftengine/src/preprocessor/messages/ClientBatchRequestMsg.cpp
nkumar04: Added check for MAX_BATCH_SIZE

bftengine/src/preprocessor/messages/PreProcessBatchRequestMsg.cpp
glevkovich: yes, lets keep it that way... it will just complicate things.

PR #1955, opened 2021-10-08
Fix order of block deletes relative to adds on ST

dartdart26:
Given healthy replicas execute pruning (block deletes) and block
additions in a specified order, e.g.:
  "block(1, 1), block(2, 1), block(3, 1), prune 1, prune 2, block(4, 3)"
where block(X, Y) means block X being added when current genesis is Y
and prune N means prune block N, we want that a fallen-behind or crashed
replica executes them in the same order on top of its blockchain. This
is required so that all replicas reach to the same hash commitments at
the end. Currently, the replica that receives blocks from ST will do the
following:
  1. Add all blocks to its chain.
  2. Prune up to the same point as other replicas.
However, this would lead to the following different order:
  "block(1, 1), block(2, 1), block(3, 1), block(4, 1), prune 1, prune 2"

In order to provide the same order, we persist the genesis block ID at
the time any subsequent block is added in that block. That allows us to
then iterate over blocks received by ST and before adding them to the local
blockchain, we first prune up to the genesis block ID in the ST block
currently being processed. For example, if block 1 is already existing
at the crashed replica, its actions would be:
1. block(2, 1) -> add block 2
2. block(3, 1) -> add block 3
3. block(4, 3):
  * prune block 1 and 2 (up to 3)
  * add block 4 as block (4, 3)

Above solution adds a single 8-byte "concord_internal" category key-value in every block.

Remove `last_scheduled_block_for_pruning_` in PruningHandler and just
use the genesis block ID for pruning status.

Change unit tests so that they accommodate the new key.

Introduce a new unit test that tests the new behaviour. Update the
`test_pruning_command_with_failures` reconfiguration test so that it
always does ST and add an extra step - a subsequent ST after pruning.

### DIFF
--- a/kvbc/include/kvbc_key_types.hpp
+++ b/kvbc/include/kvbc_key_types.hpp
@@ -13,6 +13,8 @@
 #pragma once
 
 #include <cstdint>
+#include <string>
+
 namespace concord::kvbc::keyTypes {
 static const char bft_seq_num_key = 0x21;
 static const char reconfiguration_pruning_key = 0x24;
@@ -27,6 +29,7 @@ static const char reconfiguration_tls_exchange_key = 0x2e;
 
 static const char reconfiguration_restart_key = 0x30;
 static const char reconfiguration_ts_key = 0x31;
+static const std::string genesis_block_key(1, 0x32);
 
 enum CLIENT_COMMAND_TYPES : uint8_t {
   start_ = 0x0,
--- a/kvbc/src/categorization/kv_blockchain.cpp
+++ b/kvbc/src/categorization/kv_blockchain.cpp
@@ -17,6 +17,9 @@
 #include "json_output.hpp"
 #include "diagnostics.h"
 #include "performance_handler.h"
+#include "kvbc_key_types.hpp"
+#include "categorization/db_categories.h"
+#include "endianness.hpp"
 
 #include <stdexcept>
 
@@ -146,6 +149,15 @@ void KeyValueBlockchain::loadCategories() {
   }
 }
 
+void KeyValueBlockchain::addGenesisBlockKey(Updates& updates) const {
+  const auto stale_on_update = true;
+  updates.addCategoryIfNotExisting<VersionedInput>(kConcordInternalCategoryId);
+  updates.appendKeyValue<VersionedUpdates>(
+      kConcordInternalCategoryId,
+      std::string{keyTypes::genesis_block_key},
+      VersionedUpdates::ValueType{concordUtils::toBigEndianStringBuffer(getGenesisBlockId()), stale_on_update});
+}
+
 // 1) Defines a new block
 // 2) calls per cateogry with its updates
 // 3) inserts the updates KV to the DB updates set per column family
@@ -154,6 +166,7 @@ BlockId KeyValueBlockchain::addBlock(Updates&& updates) {
   diagnostics::TimeRecorder scoped_timer(*histograms_.addBlock);
   // Use new client batch and column families
   auto write_batch = native_client_->getBatch();
+  addGenesisBlockKey(updates);
   auto block_id = addBlock(std::move(updates.category_updates_), write_batch);
   native_client_->write(std::move(write_batch));
   block_chain_.setAddedBlockId(block_id);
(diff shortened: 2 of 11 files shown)


kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -146,6 +149,15 @@ void KeyValueBlockchain::loadCategories() {
   }
 }
 
+void KeyValueBlockchain::addGenesisBlockKey(Updates& updates) const {
+  const auto stale_on_update = true;
+  updates.addCategoryIfNotExisting<VersionedInput>(kConcordInternalCategoryId);
```
arc-vmware, line 154: Just a que, should we add kConcordReconfigurationCategoryId and add the genesis block only for 
 for those blocks with kConcordReconfigurationCategoryId ?

dartdart26: Actually, I am not sure what the intention of `kConcordReconfigurationCategoryId` is. Maybe it can be used instead. Maybe @yontyon can give us a hint.

dartdart26: There are two `kvbc_key_types` files, it might be a bit confusing.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -729,13 +742,32 @@ void KeyValueBlockchain::linkSTChainFrom(BlockId block_id) {
...
+  if (cat_it == block.data.updates.kv.cend()) {
+    return;
+  }
+  const auto& internal_kvs = std::get<VersionedInput>(cat_it->second).kv;
+  auto key_it = internal_kvs.find(keyTypes::genesis_block_key);
+  if (key_it != internal_kvs.cend()) {
+    const auto block_genesis_id = concordUtils::fromBigEndianBuffer<BlockId>(key_it->second.data.data());
+    while (getGenesisBlockId() >= INITIAL_GENESIS_BLOCK_ID && getGenesisBlockId() < getLastReachableBlockId() &&
+           block_genesis_id > getGenesisBlockId()) {
+      deleteGenesisBlock();
```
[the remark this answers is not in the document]
dartdart26: Thank you @arc-vmware for finding this issue and your help!

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -146,6 +149,15 @@ void KeyValueBlockchain::loadCategories() {
   }
 }
 
+void KeyValueBlockchain::addGenesisBlockKey(Updates& updates) const {
+  const auto stale_on_update = true;
+  updates.addCategoryIfNotExisting<VersionedInput>(kConcordInternalCategoryId);
+  updates.appendKeyValue<VersionedUpdates>(
+      kConcordInternalCategoryId,
+      std::string{keyTypes::genesis_block_key},
+      VersionedUpdates::ValueType{concordUtils::toBigEndianStringBuffer(getGenesisBlockId()), stale_on_update});
```
cloudnoize, line 158: nice :)

kvbc/include/kvbc_key_types.hpp
```diff
@@ -27,6 +29,7 @@ static const char reconfiguration_tls_exchange_key = 0x2e;
 
 static const char reconfiguration_restart_key = 0x30;
 static const char reconfiguration_ts_key = 0x31;
+static const std::string genesis_block_key(1, 0x32);
```
yontyon, line 32: All reconfiguration keys are now being written under the reconfiguration category, so you can reuse keys under a different category if you want.

kvbc/src/categorization/kv_blockchain.cpp
dartdart26: @yontyon We call `updates.addCategoryIfNotExisting()`to add it to the updates only, not globally. Is that what you meant?

In terms of the ID, we use 0x32. Is that OK for `kConcordInternalCategoryId`?

kvbc/src/pruning_handler.cpp
```diff
@@ -248,8 +244,8 @@ bool PruningHandler::handle(const concord::messages::PruneStatusRequest&,
   if (!pruning_enabled_) return true;
   concord::messages::PruneStatus prune_status;
   std::lock_guard lock(pruning_status_lock_);
-  prune_status.last_pruned_block =
-      last_scheduled_block_for_pruning_.has_value() ? last_scheduled_block_for_pruning_.value() : 0;
+  const auto genesis_id = ro_storage_.getGenesisBlockId();
+  prune_status.last_pruned_block = ((genesis_id != INITIAL_GENESIS_BLOCK_ID && genesis_id != 0) ? genesis_id - 1 : 0);
```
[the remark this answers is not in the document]
dartdart26: Yes, you are right. Will change it, thanks.

PR #1961, opened 2021-10-12
Changes in Concord Client request service

tdimitrov:
This patch applies a minor change to the Client Request gRPC service
imposed by the new execution engine interface.
The type of request field (in Request message) and response field (in
Response message) are changed from bytes to protobuf.Any.

This change allows a type to be attached to the response field so that
the user of the service can interpret the data easier.

### DIFF
--- a/client/proto/request/v1/request.proto
+++ b/client/proto/request/v1/request.proto
@@ -6,6 +6,7 @@ syntax = "proto3";
 package vmware.concord.client.request.v1;
 
 import "google/protobuf/duration.proto";
+import "google/protobuf/any.proto";
 
 // Specifies Java package name, using the standard prefix "com."
 option java_package = "com.vmware.concord.client.request.v1";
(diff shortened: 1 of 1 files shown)


client/proto/request/v1/request.proto
```diff
@@ -28,12 +29,13 @@ service RequestService {
   // RESOURCE_EXHAUSTED: if Concord Client is overloaded. The caller should retry with a backoff.
   // UNAVAILABLE: if Concord Client is currently unable to process any requests. The caller should retry with a backoff.
   // INTERNAL: if Concord Client cannot progress independent of the request.
+  // ABORTED: this error is returned if the PreExecution read/time conditions are not met. TODO: which preexecution condition was violated?
```
meiersi-da, line 32: @tdimitrov : please expand on these three kinds of error.

SamirTalwar-DA: I think the issue is simply that we are likely to have many more requests open at once. I'm afraid I don't have a good enough understanding of all of the various components to know if it affects anything else.

This may or may not be a problem, but I suggest we pay close attention to any performance impact as a result. Do you think we can test this in isolation to see if it makes any difference to performance under load?

tdimitrov:
Hey @cjn do you need this PR?

PR #1963, opened 2021-10-13
TRS: Do not remove live_updates when transitioning to event groups

shgandhi:
When there are no more legacy events, and TRS is preparing to transition
to event groups, TRS removes the live_updates buffer from the subscriber
list, while still using live_updates buffer to sync event groups with
storage.
TRS incorrectly tries to remove the buffer again from the subscriber_list
whenever the server context is cancelled. If is_event_group_transition
is set to true, live_updates must not be removed from the subscriber list
to avoid double removal from subscriber_list.

### DIFF
--- a/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
+++ b/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
@@ -347,10 +347,12 @@ class ThinReplicaImpl {
             update_aggregator_counter = 0;
           }
         }
-        config_->subscriber_list.removeBuffer(live_updates);
-        live_updates->removeAllUpdates();
-        metrics_.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
-        metrics_.updateAggregator();
+        if (!is_event_group_transition) {
+          config_->subscriber_list.removeBuffer(live_updates);
+          live_updates->removeAllUpdates();
+          metrics_.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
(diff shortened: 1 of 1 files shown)


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -347,10 +347,12 @@ class ThinReplicaImpl {
             update_aggregator_counter = 0;
           }
         }
-        config_->subscriber_list.removeBuffer(live_updates);
-        live_updates->removeAllUpdates();
-        metrics_.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
-        metrics_.updateAggregator();
+        if (!is_event_group_transition) {
```
robem, line 350: We can even put the whole `try-catch` block under this condition. Good catch! 👍

shgandhi: It is [this](https://github.com/vmware/concord-bft/blob/c48238d8501f1309dffd9db007762ff0b429788f/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp#L428) call to removeBuffer that results in the crash. 
I was thinking we add a `try-catch` block inside the `removeBuffer` method, and remove the assert, WDYT?

shgandhi: I was thinking of throwing a runtime exception. But I can't think of a situation where this error will be seen other than a coding error, so I agree, assert is the right way to go.

PR #1968, opened 2021-10-18
Add block id and its signature to the pre-process result

cloudnoize:
On pre-execution, each replica adds to the PreProcessReplyMsg:

- block_id - the replica block id at the pre-execution time.

- block_id signature - the replica signs the block id for other replicas to validate and prevent a malicious replica from fooling us into choosing a block id that is not correct.

On collecting enough pre-execution results, the primary sends to consensus a PreProcessReslt msg that contains a list of result signatures.

During consensus, the signatures are being validated.

Before the execution, the minimum block id is chosen from the list and passed to the post-execution.

-- The previous method where the primary sends its block id view the pre-execution result was removed.


tdimitrov (approved):
Good job @cloudnoize

PR #1969, opened 2021-10-18
stabilizing skvbc_batch_preexecution apollo test

nkumar04:
In some intermittent failure cases, previous batch execution is not finished on all replicas when primary is stopped. As a result, new requests are ignored and if we have only f or less replicas that complains then view change is not triggered.
So, in such a case we need to send new requests for view change. 
Adding this check to make sure that, previous batch execution is finished before we stop primary to test view change


HristoStaykov (approved):
LGTM!

PR #1972, opened 2021-10-19
TS: Time-Service Fixes

ukg19:
This PR handles Time-service related fixes.

See each commit for more info

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -517,6 +517,7 @@ PrePrepareMsg *ReplicaImp::createPrePrepareMessage() {
   }
 
   if (config_.timeServiceEnabled) {
+    // If time-service is enabled, the first client request will be time request
     auto timeServiceMsg = time_service_manager_->createClientRequestMsg();
     auto pp = new PrePrepareMsg(config_.getreplicaId(),
                                 getCurrentView(),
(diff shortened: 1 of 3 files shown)


ukg19:
> 1. Add a comment in replicaimp where the first message is added for timeservice. Also add some comment whereever timeservice related changed are done.
Done
> 2. Why are we using full reserved page for the timeservice since its only a single timestamp? Review this, if it can be added to the existing timeservice.
We are using a single reserved page, so it is fine.
> 3. Add the following comment in Pre-execution : “No one will use the time for computation in pre-execution”.
Added
> 4. Add real scenarios like delay in sending PrePrepare Messages, and then allow replica to recover.
Done as part of this PR https://github.com/vmware/concord-bft/pull/1983
> 5. Drift of time for more than f+1 replicas.
Done as part of this PR https://github.com/vmware/concord-bft/pull/1983
> 6. Add metric to find the number of view changes happening due to time service, if possible.
Not possible
> 7. Change the method name compareAndSwap to compareAndUpdate (actually no swapping is happening).
Done
> 8. We cshould not change the FullCommitProofMsg for Timeservice. Please review and revert those changes if make sense.
Done as part of this PR https://github.com/vmware/concord-bft/pull/1987
> 9. ConsensusTime is currently std::chrono::milliseconds which doesn’t utilise full 64 bits ( uses 56 bits) , it may be a problem on some platforms. Review and convert it to use unit64_t
Currently we are getting time using ClockT::now().time_since_epoch() which gives a duration number. This number needs to be converted using functions like : 

std::chrono::milliseconds | duration</signed integer type of at least 45 bits/, std::milli>
-- | --
where again we will face the same problem because std::chrono::milliseconds uses 45 bits at least. So it is not possible using current approach
> 10. Update TS v2 documentation wherever there is a difference between design and code
Done
> 11. Check that the soft and hard limits are in deployment tag and customer should be able to change those.
Done in concord
> 12. Why do we need serialize_impl and deserialize_impl functions, lets check if we can remove those.
They are needed for duration types
> 13. Check for different configuration of f, c & n in Apollo testcases and unit testcases.
Done as part of this PR https://github.com/vmware/concord-bft/pull/1983
> 14. Don't send PrePreparePartialMsg and PrePrepareCommitProof message if time is not ok.
Handling in progress

arc-vmware (approved):
LGTM

PR #1974, opened 2021-10-20
KvbAppFilter Event Group Bug fixes

shgandhi:
This MR adds a few bug fixes for issues observed in end-to-end testing of event group path. Please see individual commits for more detail.


kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -388,7 +388,11 @@ std::optional<uint64_t> KvbAppFilter::getNextEventGroupId(std::shared_ptr<EventG
   // Vice versa is true if event_group_id_batch is populated with only private event group ids
   // Note: each private_event_group_ids and public_event_group_ids hold a sorted list of global event group ids at all
   // times
-  if (!public_event_group_ids.empty() &&
+  bool is_lte_batch_size = (!public_event_group_ids.empty() && !private_event_group_ids.empty() &&
+                            public_event_group_ids.size() + private_event_group_ids.size() <= kBatchSize)
+                               ? true
+                               : false;
+  if (!public_event_group_ids.empty() && !is_lte_batch_size &&
```
robem, line 395: Instead of `is_lte_batch_size`, could we use `public_event_group_ids.size() >= kBatchSize`? Similar for private events?
IMHO, this makes it more readable because you only consider public events if all public events available can fill the whole batch + other conditions.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -440,7 +440,7 @@ std::optional<uint64_t> KvbAppFilter::getNextEventGroupId(std::shared_ptr<EventG
       if (!has_pub_offset_updated) {
         if (auto pub_it = std::find(public_event_group_ids.begin(), public_event_group_ids.end(), *r_it);
             pub_it != public_event_group_ids.end()) {
-          eg_state->public_offset = ++pub_it - public_event_group_ids.begin();
+          eg_state->public_offset += ++pub_it - public_event_group_ids.begin();
```
robem, line 443: Good catch! We probably need a test for the `else` part of this function :)

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
shgandhi: Thanks, yes it does. Updated.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -384,13 +383,16 @@ std::optional<uint64_t> KvbAppFilter::getNextEventGroupId(std::shared_ptr<EventG
   }
   // populate event_group_id_batch with only public event group ids iff one of the following occurs -
   // 1. No private event groups exist in storage
-  // 2. The first global event group id in private_event_group_ids is greater than the last global event group id
-  // public_event_group_ids
+  // 2. The first global event group id in private_event_group_ids is greater than the last global event group id in
+  // public_event_group_ids and public_event_group_ids has atleast kBatchSize elements
```
robem, line 387: "at least"

shgandhi: Corrected.

PR #1978, opened 2021-10-21
EventGroup: Update tracing

robem:
This change propagates the trace context from concord to the caller of concord
client. Thereby, we leverage the existing legacy interface (bytes) for internal
trace context transfer (via thin_replica.proto) but expose it in W3C HTTP
header format within a "text map".
Note, the functionality has not been end-to-end tested given that we still lack
a tracer implementation (such as Jaeger). This will be introduced soon and
should uncover any bugs.


client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -202,8 +202,8 @@ std::pair<bool, ThinReplicaClient::SpanPtr> ThinReplicaClient::readBlock(Data& u
   SpanPtr span;
   if (update_in.has_event_group()) {
     id = update_in.event_group().id();
-    // TODO: Event Group traces
-    span.reset();
+    span = TraceContexts::CreateChildSpanFromBinary(
+        update_in.event_group().trace_context(), "trc_read_event_group", {}, logger_);
```
shgandhi, line 206: Maybe I am missing something, why are we not passing `std::to_string(id)` as cid in this call to `CreateChildSpanFromBinary`?

PR #1985, opened 2021-10-25
 changes for periodic source replacement in state transfer

ananpal:
changes for periodic source replacement in state transfer

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -2463,24 +2463,24 @@ void BCStateTran::processData(bool lastInBatch) {
   bool badDataFromCurrentSourceReplica = false;
 
   while (true) {
-    bool newSourceReplica = sourceSelector_.shouldReplaceSource(currTime, badDataFromCurrentSourceReplica);
+    const auto sourceReplacmentScheme =
+        sourceSelector_.shouldReplaceSource(currTime, badDataFromCurrentSourceReplica, lastInBatch);
 
-    if (newSourceReplica) {
+    if (sourceReplacmentScheme != SourceReplaceType::DONOT) {
       //////////////////////////////////////////////////////////////////////////
       // Select a source replica
       //////////////////////////////////////////////////////////////////////////
-      sourceSelector_.removeCurrentReplica();
       if (fs == FetchingState::GettingMissingResPages && sourceSelector_.noPreferredReplicas()) {
         EnterGettingCheckpointSummariesState();
         return;
       }
-      sourceSelector_.updateSource(currTime);
+      sourceSelector_.updateSource(currTime, sourceReplacmentScheme);
       auto currentSource = sourceSelector_.currentReplica();
       LOG_DEBUG(logger_, "Selected new source replica: " << currentSource);
       metrics_.current_source_replica_.Get().Set(currentSource);
       metrics_.preferred_replicas_.Get().Set(sourceSelector_.preferredReplicasToString());
       badDataFromCurrentSourceReplica = false;
-      clearAllPendingItemsData();
+      if (sourceReplacmentScheme == SourceReplaceType::IMMEDIATE) clearAllPendingItemsData();
     }
 
     // We have a valid source replica at this point
@@ -2514,6 +2514,7 @@ void BCStateTran::processData(bool lastInBatch) {
       }
     }
 
+    bool newSourceReplica = (sourceReplacmentScheme != SourceReplaceType::DONOT);
     ConcordAssertNE(nextRequiredBlock_, 0);
     ConcordAssertOR(
         (fetchingState == FetchingState::GettingMissingBlocks) && (nextCommittedBlockId_ >= nextRequiredBlock_),
@@ -2569,7 +2570,6 @@ void BCStateTran::processData(bool lastInBatch) {
       //////////////////////////////////////////////////////////////////////////
       // if we have a new block
       //////////////////////////////////////////////////////////////////////////
-      sourceSelector_.setSourceSelectionTime(currTime);
       sourceSelector_.onReceivedValidBlockFromSource();
 
       ConcordAssertAND(lastChunkInRequiredBlock >= 1, actualBlockSize > 0);
@@ -2659,7 +2659,6 @@ void BCStateTran::processData(bool lastInBatch) {
       // if we have a new vblock
       //////////////////////////////////////////////////////////////////////////
       DataStoreTransaction::Guard g(psd_->beginTransaction());
-      sourceSelector_.setSourceSelectionTime(currTime);
       sourceSelector_.onReceivedValidBlockFromSource();
 
       if (config_.enableReservedPages) {
--- a/bftengine/src/bcstatetransfer/SourceSelector.cpp
+++ b/bftengine/src/bcstatetransfer/SourceSelector.cpp
@@ -19,7 +19,10 @@ namespace impl {
 
 bool SourceSelector::hasSource() const { return currentReplica_ != NO_REPLICA && sourceSelectionTimeMilli_ > 0; }
 
-void SourceSelector::setSourceSelectionTime(uint64_t currTimeMilli) { sourceSelectionTimeMilli_ = currTimeMilli; }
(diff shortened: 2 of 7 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2460,7 +2460,7 @@ void BCStateTran::processData(bool lastInBatch) {
   bool badDataFromCurrentSourceReplica = false;
 
   while (true) {
-    bool newSourceReplica = sourceSelector_.shouldReplaceSource(currTime, badDataFromCurrentSourceReplica);
+    bool newSourceReplica = sourceSelector_.shouldReplaceSource(currTime, badDataFromCurrentSourceReplica, lastInBatch);
```
mdarade, line 2463: You may end up choosing new source just before processing last chunk from batch.

bftengine/src/bcstatetransfer/SourceSelector.cpp
```diff
@@ -143,7 +149,7 @@ bool SourceSelector::shouldReplaceSource(uint64_t currTimeMilli, bool badDataFro
     return true;
   }
 
-  if (sourceReplacementTimeoutMilli_ > 0) {
+  if (lastInBatch && (sourceReplacementTimeoutMilli_ > 0)) {
```
mdarade, line 152: Should you remove current source from preferred replicas in case of timeout? Maybe updateSource() needs to handle this case differently?

bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
```diff
@@ -689,6 +701,38 @@ TEST_F(BcStTest, dstValidateRealSourceListReported) {
...
+// Validate a recurring source selection, during ongoing state transfer;
+TEST_F(BcStTest, validatePeriodicSourceReplacement) {
+  ASSERT_NO_FATAL_FAILURE(SendCheckpointSummaries());
+  mockedSrc_->ReplyAskForCheckpointSummariesMsg();
+  int batch_count = 0;
+  while (true) {
+    ASSERT_NO_FATAL_FAILURE(
+        AssertFetchBlocksMsgSent(testParams_.expectedFirstRequiredBlockNum, testParams_.expectedLastRequiredBlockNum));
+    // once the source is selected, adding sleep for more than source replacement time duration
+    if (batch_count < 2) {
```
glevkovich, line 713: This doesn't make sense to me. Why do we need to add 5 seconds delay between request and reply? This does never happen in real world and shouldn't happen in happy flow test scenario.
you have the mock delays as part of the reply (2 millisec).
Adding sleeps here will trigger unwanted retransmissions..since no replay is received in 5000ms and I believe the retransmission timeout is smaller. This creates an unwanted scenario in test, where you involve replacement  due to retransmissions.
You first spoke on makeing sourceReplicaReplacementTimeoutMs=500ms when we had 1 on 1 chat, why is it now 5000ms? sounds too much to me.
We have approx 6-7 batches here, 1650 blocks needed to be fetched. with 2ms delay it will take 3 to 3.5 second to transfer all. 
If you reduce the replacement to 500 it should replace multiple times.
Also, the test should be general enough to work if I change number of blocks fetched.

bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
```diff
@@ -689,6 +701,38 @@ TEST_F(BcStTest, dstValidateRealSourceListReported) {
...
+    mockedSrc_->ReplyFetchBlocksMsg();
+    if (testParams_.expectedLastRequiredBlockNum <= config_.maxNumberOfChunksInBatch) break;
+    testParams_.expectedLastRequiredBlockNum -= config_.maxNumberOfChunksInBatch;
+    // There might be pending jobs for putBlock, we need to wait some time and then finalize them by calling
+    this_thread::sleep_for(chrono::milliseconds(20));
+    onTimerImp();
+    batch_count++;
+  }
+  const auto& sources_ = GetSourceSelector().getActualSources();
+  ASSERT_EQ(sources_.size(), 3);
```
glevkovich, line 725: I think we need to expect some computed number, in case we increase number of blocks fetched or change periodic replacment config param. Test should be adjustable to configuration.
In that case I would expect the number of sources to be:
time to send batches / periodic replacment time + 1

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2511,6 +2512,7 @@ void BCStateTran::processData(bool lastInBatch) {
       }
     }
 
+    bool newSourceReplica = replaceSourceType != SourceReplaceType::DONT_REPLACE ? true : false;
```
glevkovich, line 2515: nit:
`bool newSourceReplica = (replaceSourceType != SourceReplaceType::DONT_REPLACE);`

or you can completely remove this verialbe and use replaceSourceType for the rest of the code.

bftengine/src/bcstatetransfer/SourceSelector.cpp
```diff
@@ -32,6 +35,7 @@ void SourceSelector::removeCurrentReplica() {
   preferredReplicas_.erase(currentReplica_);
   currentReplica_ = NO_REPLICA;
   receivedValidBlockFromSrc_ = false;
+  setSourceSelectionTime(0);
```
glevkovich, line 38: Can you please explain why are you nulling the source selection time?

bftengine/src/bcstatetransfer/SourceSelector.cpp
```diff
@@ -89,8 +93,8 @@ uint64_t SourceSelector::timeSinceSourceSelectedMilli(uint64_t currTimeMilli) co
 }
 
 // Replace the source.
-void SourceSelector::updateSource(uint64_t currTimeMilli) {
-  if (currentReplica_ != NO_REPLICA) {
+void SourceSelector::updateSource(uint64_t currTimeMilli, SourceReplaceType replaceSourceType) {
+  if ((currentReplica_ != NO_REPLICA) && replaceSourceType == SourceReplaceType::HARD_REPLACE) {
```
glevkovich, line 97: Can you please add round brackets to the 2nd expression?

bftengine/src/bcstatetransfer/SourceSelector.cpp
glevkovich: @mdarade , I think he doesn't do this, at least not at the current code I'm reviewing. see line 97. Am I missing something?

glevkovich (changes requested):
Hi @ananpal, I think you are closer to completing the task. I have some more comments. If you would like - lets work together to fix the issues in the code.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2460,24 +2460,25 @@ void BCStateTran::processData(bool lastInBatch) {
...
       // Select a source replica
       //////////////////////////////////////////////////////////////////////////
-      sourceSelector_.removeCurrentReplica();
+      // sourceSelector_.removeCurrentReplica();
       if (fs == FetchingState::GettingMissingResPages && sourceSelector_.noPreferredReplicas()) {
         EnterGettingCheckpointSummariesState();
         return;
       }
-      sourceSelector_.updateSource(currTime);
+      sourceSelector_.updateSource(currTime, replaceSourceType);
```
mdarade, line 2475: In case of GRACEFUL_REPLACE, when & where source would be replaced?

bftengine/src/bcstatetransfer/SourceSelector.hpp
```diff
@@ -26,6 +26,8 @@ namespace impl {
 
 static const uint16_t NO_REPLICA = UINT16_MAX;
 
+enum SourceReplaceType { GRACEFUL_REPLACE, HARD_REPLACE, DONT_REPLACE };
```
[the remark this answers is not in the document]
ananpal: done

glevkovich (approved):
@ananpal , thank you for the hard work!
Please do not merge this PR, as we are waiting for v1.5 branch out. We will probably merge this one into the v1.6 dev branch anyhow.

PR #1988, opened 2021-10-26
WIP: Introducing CRE handler for scaling during state transfer

yontyon:
In this PR we address a fundamental issue with reconfiguration commands that change the network topology, remove metadata and/or restart.

Our initial approach was to rely on state transfer to get an update on such changes, while the leading quorum, commits the relevant update to the chain.
The main problem with this approach is that if the network's topology has changed, we cannot assume anything about the state-transfer's ability to complete the state transfer cycle (and hence get the update).
For example, if we scaled down from 7 to 4, the replica is state transfer is unable to complete the state transfer cycle (it needs at least 5 replicas for state transfer).

To solve this issue, we designed the state transfer CRE for executing reconfiguration commands asynchronously to state transfer compilation (see https://github.com/vmware/concord-bft/commit/89484a67ba56b68db8257ed9b555661e7bba7c16).
However, with scaling the procedure is a bit more complicated due to the following reasons:
1. We cannot execute the scaling command twice - meaning, that we need some way to mark which scaling command has already been executed.
2. In order to be able to start after the scaling, we need to remove the metadata, however, this may bring the replica to an inconsistent state once it completes the state transfer (mainly due to changes in reserved pages).

To overcome this issues we do the following:
1. We save in a dedicated file the new configuration ids. The reason why we don't use the same approach as a regular client (and save an update on the chain) is that the leading quorum will be probably wedged at this time and the state-transferred replica will not be able to update its status on the chain.
2. On state transfer compilation, the replica detect if there is some new configuration block (at this point, if there were some scaling command it was already executed by the replica's CRE). If there is such an update, the replica removes its metadata and restart itself in order to be consistent with the other replicas.

This PR also contains an Apollo test that demonstrates how this should work.
Notice that read-only replica scaling is not part of this PR.

### DIFF
--- a/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
+++ b/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
@@ -189,7 +189,7 @@ std::shared_ptr<ClientReconfigurationEngine> CreFactory::create(std::shared_ptr<
   Config cre_config;
   cre_config.id_ = repConfig.replicaId;
   cre_config.interval_timeout_ms_ = 1000;
-  PollBasedStateClient* pbc = new PollBasedStateClient(bftClient, cre_config.interval_timeout_ms_, 0, cre_config.id_);
+  IStateClient* pbc = new PollBasedStateClient(bftClient, cre_config.interval_timeout_ms_, 0, cre_config.id_);
   auto cre =
       std::make_shared<ClientReconfigurationEngine>(cre_config, pbc, std::make_shared<concordMetrics::Aggregator>());
   cre->registerHandler(std::make_shared<ReplicaTLSKeyExchangeHandler>());
--- a/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
+++ b/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
@@ -20,6 +20,7 @@
 #include "communication/ICommunication.hpp"
 #include "bftclient/bft_client.h"
 #include "ControlStateManager.hpp"
+#include "reconfiguration/ireconfiguration.hpp"
 
 namespace bftEngine::bcst::asyncCRE {
 using namespace concord::client::reconfiguration;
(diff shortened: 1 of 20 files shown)


bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
```diff
@@ -113,6 +115,62 @@ class InternalSigner : public concord::util::crypto::ISigner {
...
+    return false;
+  }
+
+  bool execute(const State& state, WriteState&) override {
+    LOG_INFO(getLogger(), "execute AddRemoveWithWedgeCommand");
+    concord::messages::ClientStateReply crep;
+    concord::messages::deserialize(state.data, crep);
+    concord::messages::ClientsAddRemoveExecuteCommand command =
+        std::get<concord::messages::ClientsAddRemoveExecuteCommand>(crep.response);
+    LOG_INFO(getLogger(), "b(1)" << KVLOG(command.token, command.config_descriptor, command.restart));
```
nkumar04, line 149: nit: b(1)?

bftengine/src/bftengine/ControlStateManager.cpp
```diff
@@ -56,13 +56,18 @@ void ControlStateManager::addOnRestartProofCallBack(std::function<void()> cb,
...
   // configuration update happens on stable checkpoint.
 
   if ((restartBftEnabled_ && IControlHandler::instance()->isOnStableCheckpoint()) ||
       IControlHandler::instance()->isOnNOutOfNCheckpoint()) {
     auto seq_num_to_stop_at = getCheckpointToStopAt();
     if (seq_num_to_stop_at.has_value() && seq_num) {
+      // A nasty hack to allow state transfer replicas to get the update in high probability - we are going to sleep for
+      // 10 seconds, assuming that it is enough for state transfer replicas to get the state change and perform the
+      // update
+      if (reason == static_cast<uint8_t>(ReplicaRestartReadyMsg::Reason::Scale))
```
nkumar04, line 69: not sure i understood. why we need to sleep in case of superStableCheckpoint?

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
yontyon: fixed

bftengine/src/bftengine/ControlStateManager.cpp
yontyon: We want to let the state transferred replica a chance to poll the latest update from the blockchain. 
Because the replicas may restart themselves (and after the restart, the state transferred replica may not be able to connect to other replicas) we wait for a while (10 seconds, where the CRE timeout is 1 second) before shutting down the replicas

nkumar04: right, but if IControlHandler::instance()->isOnNOutOfNCheckpoint() == True that means, all replicas are at same checkpoint and there is no lagging replica. right?
But yes, we need it for RO replica.

yontyon: exactly

kvbc/src/st_reconfiguration_sm.cpp
```diff
@@ -194,12 +194,12 @@ bool StReconfigurationHandler::handleWedgeCommands(const T &cmd,
   if (my_last_known_epoch == command_epoch && my_last_known_epoch == last_known_global_epoch && cp_sn < wedge_point)
     return true;  // We still need to complete another state transfer
 
-  // If we reached to this point, we are defiantly going to run the addRemove command,
-  // so lets invoke all original reconfiguration handlers from the product layer (without concord-bft's ones)
-  concord::messages::ReconfigurationResponse response;
-  for (auto &h : orig_reconf_handlers_) {
-    h->handle(cmd, bft_seq_num, UINT32_MAX, std::nullopt, response);
-  }
+  //  // If we reached to this point, we are defiantly going to run the addRemove command,
```
nkumar04, line 197: you can remove commented code

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
```diff
@@ -132,10 +190,11 @@ std::shared_ptr<ClientReconfigurationEngine> CreFactory::create(std::shared_ptr<
   Config cre_config;
   cre_config.id_ = repConfig.replicaId;
   cre_config.interval_timeout_ms_ = 1000;
-  IStateClient* pbc = new PollBasedStateClient(bftClient, cre_config.interval_timeout_ms_, 0, cre_config.id_);
+  PollBasedStateClient* pbc = new PollBasedStateClient(bftClient, cre_config.interval_timeout_ms_, 0, cre_config.id_);
```
nkumar04, line 192: isn't PollBasedStateClient derives from  IStateClient? do we need this change?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -1534,45 +1535,81 @@ async def test_add_nodes_with_failures(self, bft_network):
...
+            await bft_network.check_initital_key_exchange(stop_replicas=False, full_key_exchange=False, replicas_to_start=live_replicas)
+            bft_network.start_replicas(crashed_replica)
+            for i in range(601):
+                await skvbc.send_write_kv_set()
 
+            for r in crashed_replica:
+                await bft_network.wait_for_state_transfer_to_stop(initial_prim,
+                                                                  r,
+                                                                  stop_on_stable_seq_num=False)
+            live_replicas = live_replicas = bft_network.all_replicas(without=replicas_for_st)
```
nkumar04, line 1600: live_replicas = live_replicas = //plz remove duplicate

nkumar04 (approved):
LGTM!!
Thanks!

PR #1999, opened 2021-10-31
WIP: Use simple test

Liorgal28:


teoparvanov:
Hi @Liorgal28 , doesn't the "simple test" use the old BFT client ("simple client") instead of the new one ("bftclient")? If this is so, what is the point of using the simple test?

Liorgal28:
Currently, it's just for fast testing of the post-execution separation, after that we need to think if/how to improve this test.

PR #2000, opened 2021-10-31
WIP: revise the ERROR log level

yontyon:
In this PR we revise the usage of ERROR log level in order to have clean and accurate logs.

### DIFF
--- a/bftengine/src/bftengine/RequestHandler.cpp
+++ b/bftengine/src/bftengine/RequestHandler.cpp
@@ -98,7 +98,7 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
       req.outExecutionStatus = 0;
 
       if (req.flags & READ_ONLY_FLAG) {
-        LOG_ERROR(GL, "Received a read-only Tick, ignoring");
+        LOG_WARN(GL, "Received a read-only Tick, ignoring");
         req.outExecutionStatus = 1;
       } else if (cron_table_registry_) {
         using namespace concord::cron;
@@ -108,7 +108,7 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
         const auto tick = Tick{payload.component_id, req.executionSequenceNum};
         (*cron_table_registry_)[payload.component_id].evaluate(tick);
       } else {
-        LOG_ERROR(GL, "Received a Tick, but the cron table registry is not initialized");
+        LOG_WARN(GL, "Received a Tick, but the cron table registry is not initialized");
         req.outExecutionStatus = 2;
       }
     }
--- a/bftengine/src/bftengine/SigManager.cpp
+++ b/bftengine/src/bftengine/SigManager.cpp
@@ -183,7 +183,7 @@ uint16_t SigManager::getSigLength(PrincipalId pid) const {
     if (auto pos = verifiers_.find(pid); pos != verifiers_.end()) {
       return pos->second->signatureLength();
     } else {
-      LOG_ERROR(GL, "Unrecognized pid " << pid);
+      LOG_WARN(GL, "Unrecognized pid " << pid);
       return 0;
     }
   }
@@ -199,7 +199,7 @@ bool SigManager::verifySig(
     if (auto pos = verifiers_.find(pid); pos != verifiers_.end()) {
       result = pos->second->verify(str_data, str_sig);
     } else {
-      LOG_ERROR(GL, "Unrecognized pid " << pid);
+      LOG_WARN(GL, "Unrecognized pid " << pid);
       metrics_.sigVerificationFailedOnUnrecognizedParticipantId_++;
       metrics_component_.UpdateAggregator();
       return false;
--- a/bftengine/src/bftengine/SimpleClientImp.cpp
+++ b/bftengine/src/bftengine/SimpleClientImp.cpp
@@ -390,12 +390,12 @@ OperationResult SimpleClientImp::sendRequest(uint8_t flags,
 OperationResult SimpleClientImp::isBatchRequestValid(const ClientRequest& req) {
   OperationResult res = SUCCESS;
   if (req.flags & READ_ONLY_REQ) {
-    LOG_ERROR(logger_, "Read-only requests cannot be sent in a batch" << KVLOG(req.reqSeqNum, clientId_, req.cid));
+    LOG_WARN(logger_, "Read-only requests cannot be sent in a batch" << KVLOG(req.reqSeqNum, clientId_, req.cid));
     res = INVALID_REQUEST;
   } else if (!(req.flags & PRE_PROCESS_REQ)) {
-    LOG_ERROR(logger_,
-              "Requests batching is supported only for requests intended for pre-processing"
-                  << KVLOG(req.reqSeqNum, clientId_, req.cid));
+    LOG_WARN(logger_,
+             "Requests batching is supported only for requests intended for pre-processing"
+                 << KVLOG(req.reqSeqNum, clientId_, req.cid));
     res = INVALID_REQUEST;
   }
   if (res != SUCCESS) reset();
--- a/client/reconfiguration/src/poll_based_state_client.cpp
+++ b/client/reconfiguration/src/poll_based_state_client.cpp
@@ -36,7 +36,7 @@ concord::messages::ReconfigurationResponse PollBasedStateClient::sendReconfigura
     }
     concord::messages::deserialize(rep.matched_data, rres);
   } catch (std::exception& e) {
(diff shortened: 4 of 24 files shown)


teoparvanov (approved):
LGTM, thanks for the fixes Yoni!

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -108,7 +108,7 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
         const auto tick = Tick{payload.component_id, req.executionSequenceNum};
         (*cron_table_registry_)[payload.component_id].evaluate(tick);
       } else {
-        LOG_ERROR(GL, "Received a Tick, but the cron table registry is not initialized");
```
evdzhurov, line 111: I would say this might be an actual error depending if the cron table is really optional or not. As far as I can see an empty cron table is constructed by default in a Replica so it would be very difficult for this object to be missing in normal circumstances.

evdzhurov, line 111: I would say this might be an actual error depending if the cron table is really optional or not. As far as I can see an empty cron table is constructed by default in a Replica so it would be very difficult for this object to be missing in normal circumstances.

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -183,7 +183,7 @@ uint16_t SigManager::getSigLength(PrincipalId pid) const {
     if (auto pos = verifiers_.find(pid); pos != verifiers_.end()) {
       return pos->second->signatureLength();
     } else {
-      LOG_ERROR(GL, "Unrecognized pid " << pid);
+      LOG_WARN(GL, "Unrecognized pid " << pid);
```
evdzhurov, line 186: I think this should remain an error. The SigManager doesn't know where the pid is comming from so it could be the case that there's a programmer error isntead of a corrupted data from an external message.

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -199,7 +199,7 @@ bool SigManager::verifySig(
     if (auto pos = verifiers_.find(pid); pos != verifiers_.end()) {
       result = pos->second->verify(str_data, str_sig);
     } else {
-      LOG_ERROR(GL, "Unrecognized pid " << pid);
+      LOG_WARN(GL, "Unrecognized pid " << pid);
```
evdzhurov, line 202: Same as above.
IMO code that uses SigManager should first check for an invalid pid (coming from a corrupted message) and then proceed to use its methods. Therefore at this point invalid pids should be treated as errors.

bftengine/src/bftengine/SimpleClientImp.cpp
```diff
@@ -390,12 +390,12 @@ OperationResult SimpleClientImp::sendRequest(uint8_t flags,
 OperationResult SimpleClientImp::isBatchRequestValid(const ClientRequest& req) {
```
evdzhurov, line 390: I would say logs in isBatchRequestValid could stay as errors because there's an opportunity for earlier pruning of messages that shouldn't have be batched in the first place. These errors could help catch misuse of the batching feature.

client/reconfiguration/src/poll_based_state_client.cpp
```diff
@@ -36,7 +36,7 @@ concord::messages::ReconfigurationResponse PollBasedStateClient::sendReconfigura
     }
     concord::messages::deserialize(rep.matched_data, rres);
   } catch (std::exception& e) {
-    LOG_ERROR(getLogger(), "error while initiating bft request " << e.what());
```
evdzhurov, line 39: I would say there's a deeper issue here. If an exception occurs it could be from the sent operations or deserialization. If a sent operation throws this could be treated as an error due to internal state inconsistencies. However if deserialization fails because the length of the reply is inconsistent (externally controlled) then this would cause a termination via a failed ConcordAssert inside deserialize. In other words if a byzantine actor sends a message with inconsistent size this will cause a crash failure.

reconfiguration/src/reconfiguration_handler.cpp
```diff
@@ -197,7 +197,7 @@ BftReconfigurationHandler::BftReconfigurationHandler() {
   std::ifstream key_content;
   key_content.open(operatorPubKeyPath);
   if (!key_content) {
-    LOG_ERROR(getLogger(), "unable to read the operator public key file");
```
evdzhurov, line 200: This could be an error depending how the operator public key file is installed on a machine. If the file is pre-installed or there's no mechanism to obtain it dynamically (and recover) an error could be more useful to draw in attention.

util/src/MetricsServer.cpp
```diff
@@ -137,7 +137,7 @@ void Server::sendError(sockaddr_in* cliaddr, socklen_t addrlen) {
   memcpy(buf_ + sizeof(Header), msg, msglen);
   auto len = sendto(sock_, buf_, msglen + sizeof(Header), 0, (const struct sockaddr*)cliaddr, addrlen);
   if (len < 0) {
-    LOG_ERROR(logger_, "Failed to send error msg: " << concordUtils::errnoString(errno));
```
evdzhurov, line 140: I would consider leaving this as an error due to the following explanation of the sendto system call which should be also applicable to sendto.
"No indication of failure to deliver is implicit in a send(). Locally detected errors are indicated by a return value of -1."
In other words an error would most likely be a local failure and won't reveal a byzantine peer that didn't deliver the message. Especially since this send is over UDP made by the metrics server.

evdzhurov:
I generally agree with the idea that validating messages from the outside should produce warnings instead of errors. In a BFT system, messages cannot be guaranteed to be correct. I've added comments where the change seems too permissive or falls into a different category.

evdzhurov (approved):

bftengine/src/bftengine/RequestHandler.cpp
yontyon: A malicious primary may send a tick even if the corn table is not initialized. 
In this case, this is not a local ERROR but an actual malicious activity.
The "rule" I'm trying to have here is that if the error might not be as a result of a local error, it should be only WARN

bftengine/src/bftengine/SigManager.cpp
yontyon: same as above, a malicious client may set an unknown id in its message

bftengine/src/bftengine/SigManager.cpp
yontyon: same

bftengine/src/bftengine/SimpleClientImp.cpp
yontyon: same as above, a malicious client may send an invalid batch request

client/reconfiguration/src/poll_based_state_client.cpp
yontyon: We deserialize here the matched_date. this data must be received from at least n-f replicas. 
Hence, there is no way we would have a corrupted reply (unless the quorum itself is corrupted, in this case, I don't think it is very important what we would do)

reconfiguration/src/reconfiguration_handler.cpp
yontyon: agreed. We need to add additional checks here

util/src/MetricsServer.cpp
yontyon: Agreed. 
The only way this is not an error is during the server shutdown.
However, this may happen even in the regular case. leaving it in error level won't pass the ERROR checks (unless you are going to change it to detect only FATAL logs)
WDYT?

evdzhurov: We've already made the change to log FATAL only, I guess errors will need to be investigated from time to time.

bftengine/src/bftengine/RequestHandler.cpp
evdzhurov: Ok, I agree if that's the case.

bftengine/src/bftengine/SigManager.cpp
evdzhurov: It's a given that the id could be invalid in the message but will it be always the case that SigManager is used with ids from a message? The purpose of the error is to remind that a pid could be sanitized beforehand. I've seen it done in a few places. It's up to you.

client/reconfiguration/src/poll_based_state_client.cpp
evdzhurov: Ok, if that's the case I agree

teoparvanov:
Closing PR - @evdzhurov will continue this change.

PR #2001, opened 2021-11-01
Add a test that syncs a replica with higher view

StefanPalashev:
The test validates that once restarted, a crashed replica catches up to
the view that the other replicas are in after multiple view changes were
done while the replica was offline.

### DIFF
--- a/tests/apollo/test_skvbc_view_change.py
+++ b/tests/apollo/test_skvbc_view_change.py
@@ -364,7 +364,7 @@ class SkvbcViewChangeTest(unittest.TestCase):
             err_msg="Make sure the initial primary replica works in the new view."
         )
 
-        # waiting for the active window to be rebuilt after the view change
+        # waiting for the active window in the initial primary to be rebuilt after the view change
         await trio.sleep(seconds=5)
 
         # crash the new primary to trigger a view change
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc_view_change.py
```diff
@@ -294,6 +294,103 @@ async def test_restart_replica_after_view_change(self, bft_network, tracker):
...
+        bft_network.start_replica(initial_primary)
+        await skvbc.run_concurrent_ops(num_ops=10)
+
+        await bft_network.wait_for_view(
+            replica_id=initial_primary,
+            expected=lambda v: v == expected_next_primary,
+            err_msg="Make sure the initial primary replica works in the new view."
+        )
+
+        # waiting for the active window to be rebuilt after the view change
```
HristoStaykov, line 367: waiting for the active window to be rebuilt after the view change -> waiting for the active window to be rebuilt after the view change in initial_primary.

HristoStaykov (approved):
Great work, @StefanPalashev ! I am happy to see we have this situation covered! Just a minor recommendation for comment clarity.

tests/apollo/test_skvbc_view_change.py
StefanPalashev: I will change the comment to: "waiting for the active window in the initial primary to be rebuilt after the view change" so that it emphasizes that the rebuilding of the active window in the primary is being waited for.

PR #2003, opened 2021-11-02
Check fatal error logs only

evdzhurov:
Modify the workflow so that only fatal error logs
are exported to ReplicaErrorLog.txt and would
fail the build.


evdzhurov:
This change has been requested by @teoparvanov. This PR relaxes log checking to include only fatal errors but there's still an ongoing investigation for all errors manifested through testing.

teoparvanov (approved):
LGTM, thanks @evdzhurov !

PR #2004, opened 2021-11-02
Added thread safe version of bft_client send request api

nkumar04:
Current api to send request to replicas is not thread safe. This PR adds the thread safe version of the api, which is required if any application calls send from mutiple threads.


Liorgal28 (approved):
LGTM

PR #2007, opened 2021-11-03
Wait for all replicas to be up before interfering them

yontyon:
The `ReplicaSubsetIsolatingAdversary` expects to have all replicas up and running before starting.
In this PR we stabilize `test_add_nodes_with_failures` by verifying that all replicas are up before interfering them.


nkumar04 (approved):
LGTM!

PR #2011, opened 2021-11-04
Allow TLS support for TRS-TRC via Clientservice

shgandhi:
This PR enables TLS support for TRS-TRC connection. Please see individual commits for more detail.
Note, this still needs manual testing, which will be done once the close-sourced change corresponding to this change is made.

### DIFF
--- a/client/clientservice/include/client/clientservice/configuration.hpp
+++ b/client/clientservice/include/client/clientservice/configuration.hpp
@@ -26,6 +26,7 @@ void parseConfigFile(concord::client::concordclient::ConcordClientConfig&, const
 void configureSubscription(concord::client::concordclient::ConcordClientConfig&,
                            const std::string& tr_id,
                            bool insecure,
-                           const std::string& tls_path);
+                           const std::string& tls_path,
+                           const std::string& secrets_url);
 
 }  // namespace concord::client::clientservice
--- a/client/clientservice/include/client/clientservice/configuration.hpp
+++ b/client/clientservice/include/client/clientservice/configuration.hpp
@@ -29,6 +29,10 @@ void configureSubscription(concord::client::concordclient::ConcordClientConfig&,
                            const std::string& tls_path,
                            const std::string& secrets_url);
 
+void configureTransport(concord::client::concordclient::ConcordClientConfig& config,
+                        bool is_insecure,
+                        const std::string& tls_path);
+
 const std::string decryptPrivateKey(const std::optional<std::string>& secrets_url, const std::string& path);
 
 // This method reads certificates from file
--- a/client/clientservice/include/client/clientservice/configuration.hpp
+++ b/client/clientservice/include/client/clientservice/configuration.hpp
@@ -31,7 +31,7 @@ void configureSubscription(concord::client::concordclient::ConcordClientConfig&,
 
 const std::string decryptPrivateKey(const std::optional<std::string>& secrets_url, const std::string& path);
 
-// This method reads certificates from file if TLS is enabled
+// This method reads certificates from file
 void readCert(const std::string& input_filename, std::string& out_data);
 
 // This method gets the client_id from the OU field in the client certificate
--- a/client/clientservice/include/client/clientservice/configuration.hpp
+++ b/client/clientservice/include/client/clientservice/configuration.hpp
@@ -29,7 +29,7 @@ void configureSubscription(concord::client::concordclient::ConcordClientConfig&,
                            const std::string& tls_path,
                            const std::string& secrets_url);
 
-const std::string decryptPK(const std::optional<std::string>& secrets_url, const std::string& path);
+const std::string decryptPrivateKey(const std::optional<std::string>& secrets_url, const std::string& path);
 
 // This method reads certificates from file if TLS is enabled
 void readCert(const std::string& input_filename, std::string& out_data);
--- a/client/clientservice/include/client/clientservice/configuration.hpp
+++ b/client/clientservice/include/client/clientservice/configuration.hpp
@@ -29,4 +29,16 @@ void configureSubscription(concord::client::concordclient::ConcordClientConfig&,
                            const std::string& tls_path,
                            const std::string& secrets_url);
 
+const std::string decryptPK(const std::optional<std::string>& secrets_url, const std::string& path);
+
+// This method reads certificates from file if TLS is enabled
+void readCert(const std::string& input_filename, std::string& out_data);
+
+// This method gets the client_id from the OU field in the client certificate
+std::string getClientIdFromClientCert(const std::string& client_cert_path);
+
+// This method is used by getClientIdFromClientCert to get the client_id from
+// the subject in the client certificate
+std::string parseClientIdFromSubject(const std::string& subject_str);
+
 }  // namespace concord::client::clientservice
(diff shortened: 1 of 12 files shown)


yontyon:
Hi @shgandhi 
A question, why we are not using the same certificates we use for bftclients<->bft_replcias also for TLS<->TRC?

shgandhi:
> Hi @shgandhi A question, why we are not using the same certificates we use for bftclients<->bft_replcias also for TLS<->TRC?
@yontyon primarily because each client in the client pool uses its own communication object to communicate with the committer nodes, whereas with TRC we have a more straight-forward case of one TRC object per participant at the moment communicating with the committer nodes. So for one, we don't need as many certificates as the client pool.
Another point of differentiation is that for TRC/TRS communication we need to encode client IDs in the certificates for client authentication and authorization.

client/clientservice/include/client/clientservice/configuration.hpp
```diff
@@ -26,6 +26,19 @@ void parseConfigFile(concord::client::concordclient::ConcordClientConfig&, const
 void configureSubscription(concord::client::concordclient::ConcordClientConfig&,
                            const std::string& tr_id,
                            bool insecure,
-                           const std::string& tls_path);
+                           const std::string& tls_path,
+                           const std::string& secrets_url);
+
+const std::string decryptPK(const std::optional<std::string>& secrets_url, const std::string& path);
```
robem, line 32: `decryptPk` (following the suggestions in the coding standard) or `decryptPrivateKey` or `decryptSecret`

client/clientservice/include/client/clientservice/configuration.hpp
```diff
@@ -26,6 +26,19 @@ void parseConfigFile(concord::client::concordclient::ConcordClientConfig&, const
...
+
+// This method reads certificates from file if TLS is enabled
```
robem, line 34: This function should read from file independent of TLS. The TLS condition should be the condition for calling the function.

client/clientservice/src/configuration.cpp
```diff
@@ -142,14 +145,117 @@ void parseConfigFile(ConcordClientConfig& config, const YAML::Node& yaml) {
...
+}
+
+void readCert(const std::string& input_filename, std::string& out_data) {
+  std::ifstream input_file(input_filename.c_str(), std::ios::in);
+
+  if (!input_file.is_open()) {
+    LOG_FATAL(logger, "Failed to construct concord client.");
+    throw std::runtime_error(__PRETTY_FUNCTION__ + std::string(": Could not open the input file (") + input_filename +
+                             std::string(") to establish TLS connection with the thin replica server."));
+  } else {
```
robem, line 203: We don't need `else` if we `throw` above.

client/clientservice/src/configuration.cpp
```diff
@@ -142,14 +145,117 @@ void parseConfigFile(ConcordClientConfig& config, const YAML::Node& yaml) {
...
+      out_data = read_buffer.str();
+      LOG_INFO(logger, "Successfully loaded the contents of " + input_filename);
+    } catch (std::exception& e) {
+      LOG_FATAL(logger, "Failed to construct concord client.");
+      throw std::runtime_error(__PRETTY_FUNCTION__ +
+                               std::string(": An exception occurred while trying to read the input file (") +
+                               input_filename + std::string("): ") + std::string(e.what()));
+    }
+  }
+  return;
```
robem, line 217: No need for return

client/clientservice/src/main.cpp
```diff
@@ -53,6 +53,7 @@ po::variables_map parseCmdLine(int argc, char** argv) {
     ("tr-id", po::value<std::string>()->required(), "ID used to subscribe to replicas for data/hashes")
     ("tr-insecure", po::value<bool>()->default_value(false), "Testing only: Allow insecure connection with TRS on replicas")
     ("tr-tls-path", po::value<std::string>()->default_value(""), "Path to thin replica TLS certificates")
+    ("secrets-url", po::value<std::string>()->default_value(""), "URL to decrypt private keys")
```
robem, line 56: Is it used for private keys only? Or is it more generic "configuration"?

secretsmanager/secretretriever/CMakeLists.txt
```diff
@@ -33,5 +28,9 @@ target_compile_definitions(secret_retriever_shared PUBLIC
         $<$<BOOL:${HTTPLIB_IS_USING_OPENSSL}>:CPPHTTPLIB_OPENSSL_SUPPORT>
         )
 
+target_link_libraries(secret_retriever PRIVATE nlohmann_json::nlohmann_json ${OPENSSL_LIBRARIES})
+
+target_link_libraries(secret_retriever_shared PRIVATE nlohmann_json::nlohmann_json ${OPENSSL_LIBRARIES}       )
```
robem, line 33: The closing bracket seems off.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -89,10 +89,18 @@ struct SubscribeConfig {
   std::string id;
   // If set to false then all certificates related to subscription will be ignored
   bool use_tls;
+  // TRC/S certificate path
+  std::string trsc_tls_cert_path;
```
robem, line 93: Why do we need this here? You only need the path when you configure the client and read the certs.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -89,10 +89,18 @@ struct SubscribeConfig {
...
+  // URL used to decrypt private keys
+  std::string secrets_url;
```
robem, line 95: same

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -89,10 +89,18 @@ struct SubscribeConfig {
...
   // Buffer with the client's PEM encoded certififacte chain
   std::string pem_cert_chain;
   // Buffer with the client's PEM encoded private key
   std::string pem_private_key;
+  // TRC ID obtained from OU field in pem_cert_chain
+  std::string id_from_cert;
```
robem, line 101: Not sure if this provides any value either. If the user can fill out `id` and `id_from_cert` then we cannot really compare because we don't know if `id_from_cert` actually came from the certificate.

client/thin-replica-client/include/client/thin-replica-client/trs_connection.hpp
```diff
@@ -31,21 +31,25 @@ struct TrsConnectionConfig {
...
-  const std::string client_cert_path;
-  const std::string server_cert_path;
-
-  TrsConnectionConfig(bool use_tls_, const std::string& thin_replica_tls_cert_path_, const std::string& client_key_)
+  const std::string& client_key;
+  const std::string& client_cert;
+  const std::string& server_cert;
+  // If TLS is enabled for TRC-TRS connection, the client cert must have the
+  // client ID in the OU field which should match the client ID of the client
+  const std::string& cert_client_id;
```
robem, line 41: Similar to the comment above. If the user provides this field then we don't know where it actually comes from. Instead call it `id` or `client_id`.

robem (changes requested):

client/clientservice/src/main.cpp
shgandhi: It is only used for decrypting private keys.

client/concordclient/include/client/concordclient/concord_client.hpp
shgandhi: Thanks for catching, updated.

client/concordclient/include/client/concordclient/concord_client.hpp
shgandhi: Yes, thanks. This shouldn't be configurable. Updated.

robem (approved):

PR #2012, opened 2021-11-04
Exchange the client tls certs of a server

yontyon:
In our current setup, each server has two certificates being used. One when it serves as serve and one when it serves as a client (the specific logic is: A server is a client for the servers with id smaller than its own and server to the rest)
In this PR, we adjust the TLS key rotation logic to this and fix some bugs and stabilize it.


Liorgal28 (approved):

tests/apollo/util/skvbc_history_tracker.py
```diff
@@ -572,6 +574,14 @@ def _match_filled_blocks(self):
...
                 else:
                     unmatched.append(req)
+            if success is True:
+                continue
+            for _ in range(0, len(failures_write_requests)):
+                req = failures_write_requests.pop()
+                if req.writeset == block_kvpairs:
+                    matched_blocks.append((req, block_id))
+                    success = True
+                    break
```
teoparvanov, line 584: maybe move this under the `if success is True:` block?

PR #2014, opened 2021-11-04
Support byzantine behaviour for the conflict detection optimization

cloudnoize:
Before this change, a replica blindly used a block id given from the primary to determine whether it could rely on its cache as a single source of truth instead of going to storage.
It means that the primary could have fooled us by sending a larger than the actual block id.

This PR introduces a mechanism to evaluate whether the primary block id is reliable.

A replica maintains a delta variable that is used to determine if the primary block id is too large.
The calculation is as follows: if the replica block id + the delta is smaller than the given primary block id, then the primary block id is rejected, and `0` is used.

The Block ID is being appended to the pre-execution result and therefore being signed.

**If a quorum of replicas decides to reject the value, the pre-execution is at a state where there is a quorum that does not include the primary, to recover, the primary modifies its appended block id to `0`.**

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -4177,11 +4177,12 @@ void ReplicaImp::executeReadOnlyRequest(concordUtils::SpanWrapper &parent_span, 
 
 void ReplicaImp::setConflictDetectionBlockId(const ClientRequestMsg &clientReqMsg,
                                              IRequestsHandler::ExecutionRequest &execReq) {
-  auto pos = clientReqMsg.requestLength() - sizeof(uint64_t);
-  auto opt_block_id = *(reinterpret_cast<uint64_t *>(clientReqMsg.requestBuf() + pos));
-  LOG_DEBUG(GL, "cd optimization block id is " << opt_block_id);
+  ConcordAssertGT(clientReqMsg.requestLength(), sizeof(uint64_t));
+  auto requestSize = clientReqMsg.requestLength() - sizeof(uint64_t);
+  auto opt_block_id = *(reinterpret_cast<uint64_t *>(clientReqMsg.requestBuf() + requestSize));
+  LOG_DEBUG(GL, "Conflict detection optimization block id is " << opt_block_id);
   execReq.blockId = opt_block_id;
-  execReq.requestSize = pos;
+  execReq.requestSize = requestSize;
 }
 
 void ReplicaImp::executeRequestsInPrePrepareMsg(concordUtils::SpanWrapper &parent_span,
@@ -4420,8 +4421,8 @@ void ReplicaImp::executeRequestsAndSendResponses(PrePrepareMsg *ppMsg,
         static_cast<uint32_t>(config_.getmaxReplyMessageSize() - sizeof(ClientReplyMsgHeader)),
         (char *)std::malloc(config_.getmaxReplyMessageSize() - sizeof(ClientReplyMsgHeader)),
         req.requestSeqNum()});
-    // Decode the block id of the conflict detection optimization,
-    // and pass it to the post-execution, change the size of the
+    // Decode the pre-execution block-id for the conflict detection optimization,
+    // and pass it to the post-execution.
     if (((MessageBase::Header *)requestBody)->msgType == MsgCode::PreProcessResult) {
       setConflictDetectionBlockId(req, accumulatedRequests.back());
     }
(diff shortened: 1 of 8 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4175,6 +4175,16 @@ void ReplicaImp::executeReadOnlyRequest(concordUtils::SpanWrapper &parent_span,
   }
 }
 
+void ReplicaImp::setConflictDetectionBlockId(const ClientRequestMsg &clientReqMsg,
+                                             IRequestsHandler::ExecutionRequest &execReq) {
+  ConcordAssertGT(clientReqMsg.requestLength(), sizeof(uint64_t));
+  auto pos = clientReqMsg.requestLength() - sizeof(uint64_t);
+  auto opt_block_id = *(reinterpret_cast<uint64_t *>(clientReqMsg.requestBuf() + pos));
+  LOG_DEBUG(GL, "cd optimization block id is " << opt_block_id);
```
yuliasherman, line 4183: What "cd" states for?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4175,6 +4175,16 @@ void ReplicaImp::executeReadOnlyRequest(concordUtils::SpanWrapper &parent_span,
   }
 }
 
+void ReplicaImp::setConflictDetectionBlockId(const ClientRequestMsg &clientReqMsg,
+                                             IRequestsHandler::ExecutionRequest &execReq) {
+  ConcordAssertGT(clientReqMsg.requestLength(), sizeof(uint64_t));
+  auto pos = clientReqMsg.requestLength() - sizeof(uint64_t);
```
yuliasherman, line 4181: If 'pos' is actually 'requestSize', call it 'requestSize'.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1519,12 +1529,16 @@ uint32_t PreProcessor::launchReqPreProcessing(const PreProcessRequestMsgSharedPt
...
-  accumulatedRequests.back().blockId = preProcessReqMsg->primaryBlockId();
+      preProcessResultBuffer});
   requestsHandler_.execute(accumulatedRequests, std::nullopt, cid, span);
   const IRequestsHandler::ExecutionRequest &request = accumulatedRequests.back();
   const auto status = request.outExecutionStatus;
-  const auto resultLen = request.outActualReplySize;
+  // Append the conflict detection block id and add sizeof(uint64_t) the resulting length.
+  memcpy(preProcessResultBuffer + request.outActualReplySize, reinterpret_cast<char *>(&blockId), sizeof(uint64_t));
+  const auto resultLen = request.outActualReplySize + sizeof(uint64_t);
+  LOG_INFO(logger(),
```
yuliasherman, line 1539: Do we want to have this hash calculation happen always? We need it only for a log.

bftengine/src/preprocessor/RequestProcessingState.cpp
```diff
@@ -167,6 +167,34 @@ bool RequestProcessingState::isReqTimedOut() const {
...
+    const concord::util::SHA3_256::Digest &other, uint64_t blockId) {
+  // since this scenario is rare, a new string is allocated for safety.
+  std::string modifiedResult(primaryPreProcessResult_, primaryPreProcessResultLen_);
+  ConcordAssertGT(modifiedResult.size(), sizeof(uint64_t));
+  memcpy(modifiedResult.data() + modifiedResult.size() - sizeof(uint64_t),
+         reinterpret_cast<char *>(&blockId),
+         sizeof(uint64_t));
+  auto modifiedHash = convertToArray(SHA3_256().digest(modifiedResult.c_str(), modifiedResult.size()).data());
+  if (other == modifiedHash) {
+    LOG_INFO(logger(), "Primary hash is different from quorum due to mismatch in block id");
```
yuliasherman, line 180: Your logs don't have any reference to the command. How are you going to look for them in the logs?

yuliasherman (approved):
LGTM, with some comments.

bftengine/src/bftengine/ReplicaImp.cpp
cloudnoize: Conflict detection :) I'll change

bftengine/src/preprocessor/PreProcessor.cpp
cloudnoize: it should be on debug level, thanks

tdimitrov (approved):

PR #2016, opened 2021-11-05
Fix metrics server sendError failing with "Invalid argument"

evdzhurov:
After calling shutdown on the socket the code expects that
recvfrom will return a -1 result but 0 could be returned instead.
Following that, there's an attempt to send an error back to the peer
but since the address of the peer is empty sendError fails
with "Invalid argument".

This is a quick fix and a better approach would be to use
an alternative I/O model to wake up the RecvLoop thread, but this is a
non-trivial refactor best reserved for a future task.


HristoStaykov (approved):
LGTM, thanks @evdzhurov !

PR #2017, opened 2021-11-07
[Client service] Update error on batches

Liorgal28:
If we are not updating clientrequesterror_ when we are sending responses we will not call the callback with the right value.
Added updating of clientrequesterror_ inside sendPendingRequests in order to use callback also in batch requests.


Liorgal28:
image (7)

Here we can see that the request send and we are waiting for a callback (LIOR before logline), and we can see LIOR after log when we are getting a callback with a timeout status.

irisk29 (approved):
Thanks for the change @Liorgal28. LGTM!

PR #2020, opened 2021-11-08
Get External Client Error For Testing

irisk29:
To check the external client's error that happened during the handle of a request, I've added a function in the client pool to retrieve the error.

Moreover, there is a not-used code in this class and therefore I erased it.


Liorgal28 (approved):
LGTM

PR #2021, opened 2021-11-09
CID correction for preprepare message

arc-vmware:
The batched CID is set same as seq number.
But when we build Preprepare message in a
separate thread, we assign a temporary seq num
to the PPM (PrePrepareMsg). This seq num is
used to assign the cid to PPM. Later during
consensus, the primary replica assigns the
actual seq num and during that time cid
should be updated.
Another point to note here, if the temporary
cid allocated (based on seq num) is 999
and the actual seq num is 1003, then it will
result into a problem as the space reserved for
cid is 3. To avoid this problem we will put
the CID as 0999 and then fill 1003.
So this will result into addition of an extra
0 for other cases, which should be fine.


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -619,7 +619,10 @@ std::pair<PrePrepareMsg *, bool> ReplicaImp::buildPrePrepareMessage() {
   TimeRecorder scoped_timer(*histograms_.buildPrePrepareMessage);
   PrePrepareMsg *prePrepareMsg = createPrePrepareMessage();
   if (!prePrepareMsg) return std::make_pair(nullptr, false);
-  SCOPED_MDC("pp_msg_cid", prePrepareMsg->getCid());
+
+  if (!getReplicaConfig().prePrepareFinalizeAsyncEnabled) {
+    SCOPED_MDC("pp_msg_cid", prePrepareMsg->getCid());
```
yuliasherman, line 624: In case prePrepareFinalizeAsyncEnabled is false, what CID will appear in the logs now?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -122,7 +122,8 @@ PrePrepareMsg::PrePrepareMsg(ReplicaId sender,
                              CommitPath firstPath,
                              const concordUtils::SpanContext& spanContext,
                              size_t size)
-    : PrePrepareMsg::PrePrepareMsg(sender, v, s, firstPath, spanContext, std::to_string(s), size) {}
+    : PrePrepareMsg::PrePrepareMsg(
+          sender, v, s, firstPath, spanContext, std::string(1, '0') + std::to_string(s), size) {}
```
yuliasherman, line 126: This is not trivial and needs to be explained in a comment.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -330,5 +331,14 @@ std::string PrePrepareMsg::getCid() const {
   return std::string(this->body() + this->payloadShift() - b()->batchCidLength, b()->batchCidLength);
 }
 
+void PrePrepareMsg::updateCid(SeqNum s) {
+  std::string cidStr = std::to_string(s);
+  if (cidStr.size() >= b()->batchCidLength) {
+    memcpy(this->body() + this->payloadShift() - b()->batchCidLength, cidStr.data(), b()->batchCidLength);
+  } else {
+    memcpy(this->body() + this->payloadShift() - b()->batchCidLength + 1, cidStr.data(), b()->batchCidLength - 1);
```
yuliasherman, line 339: How this could happen now?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -330,5 +331,14 @@ std::string PrePrepareMsg::getCid() const {
   return std::string(this->body() + this->payloadShift() - b()->batchCidLength, b()->batchCidLength);
 }
 
+void PrePrepareMsg::updateCid(SeqNum s) {
```
yuliasherman, line 334: Having two different functions for CID changes - 'set' and 'update' confuses. I propose to replace 'set' functionality with 'update' and this way to keep only one function.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: if prePrepareFinalizeAsyncEnabled is false then preprepare will be created in single thread. So the correct CID will appear in logs. If prePrepareFinalizeAsyncEnabled is true then the CID will be added only at the time of consensus, since at the time of building the preprepare, the CID allocated might change later.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
arc-vmware: Thanks @yuliasherman , I will add detailed comment on this.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
arc-vmware: As this is a new function, after adding this function, the cid will be populated again with the new seq number. So the cid and seq number will become same now for all PPMs after this change.

yuliasherman (approved):
LGTM

PR #2022, opened 2021-11-09
Add a benchmark for hashing the whole state

dartdart26:
Simulate an iteration of the state of the system in the block merkle
category by including only the latest version of the keys and excluding
deleted keys. Additionally, form a blockchain from all state keys using
SHA2-256.

Use an iterator to iterate the latest version column family and
accumulate keys with versions. Then, after a certain configurable amount
has ben accumulated, use RocksDB MultiGet() in multiple threads to fetch
the values.

At the end, report the hash that covers all state keys, data sizes and
speed.


cloudnoize (approved):
Looks good

PR #2023, opened 2021-11-10
Status handler for ReadOnlyReplica

tdimitrov:
Regular replicas has got the so called class ReplicaStatusHandlers which
exposes various parameters. They can be used for diagnostic purposes for
example.
ReadOnly replica doesn't register any status handlers and it can't use
ReplicaStatusHandlers due to difference in the internal metrics used by
two replicas. For example ROR doesn't use last_stable_seq_num metric.
Solution for this is to provide a simple function which registers some
status handlers specific for the ROR.

This patch adds only a single one - last_executed_seq_num.


toly-kournik (approved):
LGTM

PR #2028, opened 2021-11-12
Revise error log levels (based on PR #2000)

evdzhurov:
Most of the changes in PR #2000 are preserved except two cases where the error log level is kept as is:
* SigManager should log an unrecognized pid as an error to remind
that invalid pids from messages should be sanitized earlier.
* MetricsServer sendError should log an error instead of
a warning - the fix for the error is in PR #2016
And one minor addition:
* Added more suitable wording for the warning
in getLastKnownReconfigurationCmdBlock.


teoparvanov (approved):
LGTM, thanks @evdzhurov !

PR #2030, opened 2021-11-12
Bc 13442  source selector primary awareness

mdarade:
Consensus messages would be interpreted by state transfer to
keep track of current primary replica so that it can be excluded
from list of sources.


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -148,7 +154,7 @@ void ReplicaImp::messageHandler(MessageBase *msg) {
       }
     }
   }
-  if (!isCollectingState() && validateMessage(trueTypeObj)) {
+  if (!isCollectingState()) {
```
arc-vmware, line 157: validateMessage() should be called here.

arc-vmware (changes requested):

bftengine/src/bftengine/ReplicaImp.cpp
mdarade: Again? Its already being called at line 131.

PR #2033, opened 2021-11-14
Asynchronous Incoming External Message validation

arc-vmware:
The message validation for various message involves
many time consuming tasks. Such tasks leads to slower
validation of messages. If the entire process of
validation happens asynchronously then the throughput
of our system will improve. This PR is achieving this
by the following step:
If the message requires asynchronous validation
then the validation happens in a separate thread;
     After validation, the message is translated
     into an internal message and pushed into the
     internal message queue.
     The internal message is handled by the replica
Criteria for async message validation:
If for a message the validation require any time consuming
operation, like signature verification, then only such message's
validation will happen asynchronously.
List of messages chosen in this PR are as below:
1) CheckpointMsg
2) ClientRequestMsg
3) PrePrepareMsg (This message is handled separately)
4) ReplicaRestartReadyMsg
5) ReplicasRestartReadyProofMsg
6) ViewChangeMsg
Adding existing message for Async validation is simple:
step1: Review if the validation can happen in a separate
thread
step2: Add bool canValidateAsync() const override { 
return true; 
}
In the message as public member function.
Adding a new message for Async validation require the
following extra step:
step3: Register the validated message callback in
MsgHandlersRegistrator.

### DIFF
--- a/bftengine/src/bftengine/MsgHandlersRegistrator.hpp
+++ b/bftengine/src/bftengine/MsgHandlersRegistrator.hpp
@@ -10,16 +10,24 @@
 // file.
 
 #pragma once
-
-#include "messages/MessageBase.hpp"
-#include "messages/InternalMessage.hpp"
 #include <functional>
 #include <unordered_map>
 
+#include "messages/MessageBase.hpp"
+#include "messages/InternalMessage.hpp"
+#include "messages/ValidatedMessageCarrierInternalMsg.hpp"
+
 namespace bftEngine::impl {
 
-typedef std::function<void(MessageBase*)> MsgHandlerCallback;
-typedef std::function<void(InternalMessage&&)> InternalMsgHandlerCallback;
+template <typename T>
+using CallbackTypeWithPtrArg = std::function<void(T*)>;
+
+template <typename T>
+using CallbackTypeWithRefArg = std::function<void(T&&)>;
+
+using MsgHandlerCallback = CallbackTypeWithPtrArg<MessageBase>;
+using ValidatedMsgHandlerCallback = CallbackTypeWithPtrArg<CarrierMesssage>;
+using InternalMsgHandlerCallback = CallbackTypeWithRefArg<InternalMessage>;
 
 // MsgHandlersRegistrator class contains message handling callback functions.
 // Logically it's a singleton - only one message handler could be registered for every message type,
@@ -27,11 +35,26 @@ typedef std::function<void(InternalMessage&&)> InternalMsgHandlerCallback;
 // needs to have its own callbacks being called.
 // For each new message corresponding function of type MsgHandlerCallback should be registered via this class
 // otherwise the message could not be handled later.
+// MsgHandlersRegistrator class contains three kinds of callback functions:
+//  1) Default callback handler which will be called by the external message dispatcher.
+//  2) Validated callback handler which will be called when redispatching of internal
+//     message will happen.
+//  3) Internal message callback handler which will be called by the Internal message
+//     dispatcher.
+// MsgHandlersRegistrator is a repository of all kinds of callbacks.
 
 class MsgHandlersRegistrator {
  public:
   void registerMsgHandler(uint16_t msgId, const MsgHandlerCallback& callbackFunc) {
     msgHandlers_[msgId] = callbackFunc;
+  }
+
+  // This variation will allow addition of validated message callback function.
+  void registerMsgHandler(uint16_t msgId,
+                          const MsgHandlerCallback& callbackFunc,
+                          const ValidatedMsgHandlerCallback& validatedMsgCallbackFunc) {
+    msgHandlers_[msgId] = callbackFunc;
+    validatedMsgHandlers_[msgId] = validatedMsgCallbackFunc;
   }
 
   void registerInternalMsgHandler(const InternalMsgHandlerCallback& cb) { internalMsgHandler_ = cb; }
@@ -42,10 +65,20 @@ class MsgHandlersRegistrator {
     return nullptr;
   }
 
+  // Returns the validated callback message handler for the given message id.
+  // Validated callback is possible only for External Messages which were validated,
+  // so they are the subset of message which has a valid message type (msgId).
+  ValidatedMsgHandlerCallback getValidatedMsgCallback(uint16_t msgId) {
+    auto iterator = validatedMsgHandlers_.find(msgId);
+    if (iterator != validatedMsgHandlers_.end()) return iterator->second;
+    return nullptr;
+  }
+
   void handleInternalMsg(InternalMessage&& msg) { internalMsgHandler_(std::move(msg)); }
 
  private:
   std::unordered_map<uint16_t, MsgHandlerCallback> msgHandlers_;
+  std::unordered_map<uint16_t, ValidatedMsgHandlerCallback> validatedMsgHandlers_;
   InternalMsgHandlerCallback internalMsgHandler_;
 };
 
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -9,6 +9,12 @@
 // these subcomponents is subject to the terms and conditions of the sub-component's license, as noted in the LICENSE
 // file.
 
+#include <memory>
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <bitset>
+
 #include "ReplicaImp.hpp"
 #include "Timers.hpp"
 #include "assertUtils.hpp"
@@ -45,17 +51,13 @@
 #include "messages/PreProcessResultMsg.hpp"
 #include "messages/ViewChangeIndicatorInternalMsg.hpp"
 #include "messages/PrePrepareCarrierInternalMsg.hpp"
+#include "messages/ValidatedMessageCarrierInternalMsg.hpp"
 #include "CryptoManager.hpp"
 #include "ControlHandler.hpp"
 #include "bftengine/KeyExchangeManager.hpp"
 #include "secrets_manager_plain.h"
 #include "bftengine/EpochManager.hpp"
 #include "RequestThreadPool.hpp"
-#include <memory>
-#include <optional>
-#include <string>
-#include <type_traits>
-#include <bitset>
 
 #define getName(var) #var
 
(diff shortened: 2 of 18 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -148,13 +177,65 @@ void ReplicaImp::messageHandler(MessageBase *msg) {
       }
     }
   }
-  if (!isCollectingState() && validateMessage(trueTypeObj)) {
+
+  if (!isCollectingState()) {
+    // If replica is not doing state transfer, then for messages whose validation is needed
```
yuliasherman, line 182: Please re-phrase the comment. Not clear.

bftengine/src/bftengine/messages/CheckpointMsg.hpp
```diff
@@ -48,6 +48,8 @@ class CheckpointMsg : public MessageBase {
 
   void validate(const ReplicasInfo& repInfo) const override;
 
+  bool canValidateAsync() const override { return true; }
```
yuliasherman, line 51: canValidateAsync->shouldValidateAsync

bftengine/src/bftengine/messages/InternalMessage.hpp
```diff
@@ -63,6 +64,10 @@ using InternalMessage = std::variant<FullCommitProofMsg*,
                                      // Carries PrePrepare message after validation.
                                      PrePrepareCarrierInternalMsg,
 
+                                     // Add Carriers of messages which will translate
```
yuliasherman, line 67: It's not a translation - in fact, CarrierMesssage encapsulates an external message, right?

bftengine/src/bftengine/messages/InternalMessage.hpp
```diff
@@ -63,6 +64,10 @@ using InternalMessage = std::variant<FullCommitProofMsg*,
...
+                                     // Incoming External messages to Internal message.
+                                     CarrierMesssage*,
```
yuliasherman, line 69: What '*' does here?

yuliasherman (approved):
LGTM we some comments.

bftengine/src/bftengine/messages/InternalMessage.hpp
arc-vmware: Thanks @yuliasherman , encapsulate is a better word than translate in our scenario, I will update the comment accordingly.
CarrierMessage can be thought of as a generic message which can represent a generic encapsulator message. The encapsulation (or even translation in some case) can happen for many reasons. 
In this PR we can see that the encapsulation is happening to translate an external Message into an internal message using ValidatedMessageCarrierInternalMsg.
ValidatedMessageCarrierInternalMsg<MSG> is giving us a way to do this in a generic manner so that for any new External Message and we don't have to add a new type in InternalMessage variant. All such types are represented by CarrierMessage. 
Tomorrow let's assume that we want to encapsulate for a different reason. (assume hypothetically) we want to encapsulate a request message and translate / treat the message as a reply message. We will create a RequestToReplyMsgConv<MSG> : public CarrierMessage. In that case CarrierMessage will represent either validated message or Reply Message based on the context. There can be other scenarios aswell. 
In current context as encapsulation is happening, I will update the comment.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Thanks @yuliasherman, let me rephrase.

bftengine/src/bftengine/messages/InternalMessage.hpp
arc-vmware: CarrierMessage just know the type of the message whose encapsulation it's carrying. So I am using a Pointer Type of CarrierMessage. This means, the internal queue will point to a location which will contain an encapsulated message with the type.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -69,58 +70,86 @@ using namespace concord::diagnostics;
 namespace bftEngine::impl {
 
 void ReplicaImp::registerMsgHandlers() {
-  msgHandlers_->registerMsgHandler(MsgCode::Checkpoint, bind(&ReplicaImp::messageHandler<CheckpointMsg>, this, _1));
+  msgHandlers_->registerMsgHandler(MsgCode::Checkpoint,
+                                   bind(&ReplicaImp::messageHandler<CheckpointMsg>, this, _1),
+                                   bind(&ReplicaImp::validatedMessageHandler<CheckpointMsg>, this, _1));
```
glevkovich, line 75: nit:
I assume you are registering here some messages which will never be handled by the handler?
Another option would be to have a null default value as 2nd parameter for registerMsgHandler which signals "I do not wish to register this message, it is too short to validate async..".

example:

```
msgHandlers_->registerMsgHandler(MsgCode::Checkpoint,
                                   bind(&ReplicaImp::messageHandler<CheckpointMsg>, this, _1),
                                   CheckpointMsg::shouldValidateAsync() ? 
                                   bind(&ReplicaImp::validatedMessageHandler<CheckpointMsg>, this, _1) : nullptr);
);
```

comments:

- It might look uglier, but maybe with more thinking there is a better way.
- You will need to make shouldValidateAsync static, whic hmakes sense.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -148,13 +177,70 @@ void ReplicaImp::messageHandler(MessageBase *msg) {
       }
     }
   }
-  if (!isCollectingState() && validateMessage(trueTypeObj)) {
+
```
glevkovich, line 180: I think I'm confused about the above code. This is some code you had in the past specific to pre-preapre , right?
I was assuming you are planning to embed it in the common solution?
In that case, even the code handle preprepare async. , why is it left and not moved to the validation handler?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -148,13 +177,70 @@ void ReplicaImp::messageHandler(MessageBase *msg) {
...
+  // The Carrier message is allocated by this replica and after validation its given back
+  // to itself without going through the network stack (as an internal message).
+  // So the carrier message is quickly reinterpreted and used instead of fresh memory allocation.
+  // So this reinterpret_cast is intended.
+  ValidatedMessageCarrierInternalMsg<T> *validatedTrueTyeObj =
+      reinterpret_cast<ValidatedMessageCarrierInternalMsg<T> *>(msg);
+  T *trueTypeObj = validatedTrueTyeObj->returnMessageToOwner();
+  delete msg;
+
+  if (bftEngine::ControlStateManager::instance().getPruningProcessStatus()) {
```
glevkovich, line 224: can this block ever happen?
We already passed same check in ReplicaImp::messageHandler?
Maybe you should assert here?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -168,6 +254,34 @@ bool ReplicaImp::validateMessage(MessageBase *msg) {
...
+ * the validation in a thread bag and returns peacefully. This will also translate the message
+ * into internal message.
+ *
+ * @param msg : This is the message whose validation is complex and should happen asynchronously.
+ *
+ * @return : returns nothing
+ */
+template <typename MSG>
+void ReplicaImp::asyncValidateMessage(MSG *msg) {
+  RequestThreadPool::getThreadPool().async(
```
glevkovich, line 268: nit:
```
I don't think compiler do optimizations here. Slightly better:
```
static std::once_flag get_once;
static concord::util::ThreadPool tp;
std::call_once(get_once, [&]() { tp = RequestThreadPool::getThreadPool(); });
tp.async( ...

```

File: tests/apollo/test_skvbc_reconfiguration.py [hunk 9]
```diff
@@ -1797,7 +1797,7 @@ async def test_install_command(self, bft_network):
     
 
     async def try_to_unwedge(self, bft_network, bft, restart, quorum=None):
-        with trio.fail_after(seconds=60):
+        with trio.fail_after(seconds=90):
```
glevkovich, line 1800: nit: why did you have to increase it?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -582,7 +582,7 @@ async def run_replica_tls_key_exchange_cycle(self, bft_network, replicas, affect
         for rep in affected_replicas:
             reps_data[rep] = {}
             for r in replicas:
-                cert_path = os.path.join(bft_network.certdir + "/" + str(r), str(r),"server", "server.cert")
+                cert_path = os.path.join(bft_network.certdir + "/" + str(rep), str(r),"server", "server.cert")
```
glevkovich, line 585: nit: did you find a bug here? maybe it should be in a seperate commit

bftengine/src/bftengine/messages/ValidatedMessageCarrierInternalMsg.hpp
```diff
@@ -0,0 +1,60 @@
...
+ public:
+  CarrierMesssage(MsgType msgType) : msgType_(msgType) {}
+  virtual ~CarrierMesssage() {}
+  MsgType getMsgType() const { return msgType_; }
+
+ private:
+  MsgType msgType_;
+};
+
+// One usecase of translation of message is doing validation in a separate thread and then
```
glevkovich, line 33: typo
usecase -> use case

glevkovich (approved):
@arc-vmware , I think it looks really good, left a few comments and questions.
Only important thing I do not understand is why we are still processing async. the preprepare message using validatePrePrepareMsg. Is it due to the fact that this is a message that might batch other messages? Can't we have the code moved into the general case? Or is it too complex?

bftengine/src/bftengine/MsgHandlersRegistrator.hpp
```diff
@@ -13,12 +13,14 @@
 
 #include "messages/MessageBase.hpp"
 #include "messages/InternalMessage.hpp"
+#include "messages/ValidatedMessageCarrierInternalMsg.hpp"
```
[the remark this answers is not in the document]
arc-vmware: thanks @glevkovich , i will make sure that this order is maintained for all the files I touched in this PR.

bftengine/src/bftengine/MsgHandlersRegistrator.hpp
```diff
@@ -13,12 +13,14 @@
...
 #include <functional>
 #include <unordered_map>
 
 namespace bftEngine::impl {
 
 typedef std::function<void(MessageBase*)> MsgHandlerCallback;
+typedef std::function<void(CarrierMesssage*)> ValidatedMsgHandlerCallback;
```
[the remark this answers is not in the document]
arc-vmware: @glevkovich thanks, will convert this to using.

bftengine/src/bftengine/MsgHandlersRegistrator.hpp
```diff
@@ -42,10 +59,20 @@ class MsgHandlersRegistrator {
     return nullptr;
   }
 
+  // Returns the validated callback message handler for the given message id.
+  // Validated callback is possible only for External Messages which were validated,
+  // so they are the subset of message which has a valud message type (msgId).
```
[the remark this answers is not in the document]
arc-vmware: thanks @glevkovich , this is very helpful.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Thanks @glevkovich, this is a great comment to discuss and then finalize.
Comment 1:
I thought about it, to add the handler for specific messages. But later I thought that it might increase an extra step when we would like to add a new message for asynchronous validation. If we add all message then choice of adding or not can happen through shouldValidateAsync(). So I preferred this way.
Comment 2:
Yes we can make it async, but sometimes for some complex message, it might be possible that whether to do async validation or not, will depend on some other cases. e,g, there are many kind of client req messages. Some might not be originating from clients and this may not have any signature. So we can decide case by case whether for this message any asynchronous validation is needed or not.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: There are 3 kinds of validation case and 2 kind of message handlers.
Validation case:
1) Generic sync validation.
2) Generic Async validation
3) PrePrepare specific Async validation.

When a message comes its handled by messageHandler() which decides which of the above validation will take place.
If its sync validation then the onMessage is also called.
If it's any kind of async validation, then the validated message handler will be called later. The validated message handler will have to check the pruning status and state transfer status again, before calling the onMessage.
With the above background, below is the ans to the ques:
1) This code is not for preprepare, its for all messages. This is for sync validation.
2) Currently I haven't merged the async validation, since the pre process result validation is specific to preprepare. This flow need to be completely checked, so it should be taken separately. I think its doable, and it will definitely reduce the complexity.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Pruning happens when all nodes are ready (its not based on quorum). During pruning, it sets a replica wide flag as during pruning we should never update the blockchain. So before starting the onMessage, the handler checks pruning flag and allows only client req messages and ignores all other protocol messages. As validation would have happened in a separate thread, both pruning and state transfer need to be checked.

tests/apollo/test_skvbc_reconfiguration.py
arc-vmware: This will be reverted as all the reconfiguration testcases are fixed in another PR. Once that goes in I will merge and commit.

tests/apollo/test_skvbc_reconfiguration.py
arc-vmware: This will also be reverted.

bftengine/src/bftengine/messages/ValidatedMessageCarrierInternalMsg.hpp
arc-vmware: Thanks I will change this.

arc-vmware:
@glevkovich Thanks a lot for the review.
The preprepare message is a set of many different kind of client req message. There is a specific validation that is happening within a subset of the the set of client message. This validation requires all the messages of subset to finalize the final verdict of the validation. The name of the message is pre process result validation. This validation is done in replicaimp and if validation fails then an instant view change is needed.
To change this, we have to move this to Prepreparemsg's validation and then pursue the natural course of view change. 
As this is more related to complexity reduction and less with performance, I am thinking of taking this in a separate PR. For now I will raise a JIRA as this is also an important task.

bftengine/src/bftengine/ReplicaImp.cpp
arc-vmware: Thanks @glevkovich , this is good. As the RequestThreadPool::getThreadPool() returns a reference to the static threadpool, we directly use the same. This is good for low level, and it definitely improves the understanding.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1323,13 +1437,40 @@ void ReplicaImp::onMessage<FullCommitProofMsg>(FullCommitProofMsg *msg) {
...
+ */
+void ReplicaImp::onCarrierMessage(CarrierMesssage *msg) {
+  auto validatedMsgCallBack = msgHandlers_->getValidatedMsgCallback(msg->getMsgType());
+  if (validatedMsgCallBack) {
+    validatedMsgCallBack(msg);
+  } else {
+    // Below delete can be a memory leak and must be investigated.
+    // This can happen only when non-translatable message is sent,
+    // this should never happen.
+    LOG_ERROR(GL, "Received unknown external translated Message: " << KVLOG(msg->getMsgType()));
```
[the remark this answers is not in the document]
arc-vmware: Thanks @glevkovich , let me use concord assert here.

bftengine/src/bftengine/ReplicaImp.cpp
glevkovich: @arc-vmware 
OK thanks for the explenation - answer 1 - I compeltely understood, for answer 2 - I now I understand you leave the case of merging the cases for the future.

bftengine/src/bftengine/ReplicaImp.cpp
glevkovich: @arc-vmware , yes - my suggestion was on how to improve the fetch of this reference (fetch only once and not millions of times), not on it's creation which is done only once.

PR #2040, opened 2021-11-15
final TLS rotation fixes

yontyon:
In this PR we fix some issues with TLS key exchange as well as adding the handler to tester CRE that handles replicas TLS key exchange (this handler should be implemented in every application-level CRE)
Also, we stabilized some of the reconfiguration Apollo tests.


Liorgal28 (approved):
LGTM 👑

PR #2043, opened 2021-11-16
Add FAILED_PRECONDITION as error for StreamSnapshot

dartdart26:
Rationale is that we might return it if Concord Client cannot verify the
state from the replicas (i.e. the signatures don't match). It can be
used in different scenarios too.


meiersi-da (approved):
Good call @dartdart26 ! I agree.

miklos-da (approved):

tdimitrov (approved):

PR #2044, opened 2021-11-16
Object Store Integrity

toly-kournik:
- ReadOnlyReplica to use MetadataStorage for storing
  CheckpointMsgDescriptor
- CheckpontMsg to include a separate state(block) digest along with an
  additional state digest (which in case of BCStateTransfer is a digest of reserved pages and may include additional info in future)

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -73,7 +73,7 @@ IStateTransfer *create(const Config &config,
 
   impl::DataStore *ds = nullptr;
 
-  if (dynamic_cast<concord::storage::memorydb::Client *>(dbc.get()))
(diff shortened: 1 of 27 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -73,7 +73,7 @@ IStateTransfer *create(const Config &config,
 
   impl::DataStore *ds = nullptr;
 
-  if (dynamic_cast<concord::storage::memorydb::Client *>(dbc.get()))
+  if (dynamic_cast<concord::storage::memorydb::Client *>(dbc.get()) || config.isReadOnly)
     ds = new impl::InMemoryDataStore(config.sizeOfReservedPage);
```
yontyon, line 77: Why a read-only replica has only an in-memory DB?
what happens if it crashed

yontyon (approved):
Very very very good to me
staff 1000

PR #2045, opened 2021-11-17
kvbc/Replica: Expose native db client

robem:
When adjusting event group counters at pruning we need the ability of the
native client to iterate over the kv-store efficiently.


dartdart26 (approved):

kvbc/include/Replica.h
```diff
@@ -151,6 +151,11 @@ class Replica : public IReplica,
   std::shared_ptr<cron::TicksGenerator> ticksGenerator() const { return m_replicaPtr->ticksGenerator(); }
   void registerStBasedReconfigurationHandler(std::shared_ptr<concord::client::reconfiguration::IStateHandler>);
 
+  std::optional<std::shared_ptr<concord::storage::rocksdb::NativeClient>> nativeDbClient() {
```
dartdart26, line 154: Maybe we can document that if not nullopt, the shared_ptr cannot be nullptr. Or just use nullptr instead of std::optional.

PR #2046, opened 2021-11-17
Add an advanced unit test for the ViewsManager

StefanPalashev:
The test verifies that when the ViewsManager receives a View Change message with Pre-Prepare message and PreparedCertificate for lower view and at least f + 1 View Change messages with Pre-Prepare message and no PreparedCertificate for higher view, it will ignore the message with PreparedCertificate and instead move to the higher view, represented by f + 1 View Change messages.


HristoStaykov (approved):
LGTM, thanks @StefanPalashev!

PR #2047, opened 2021-11-17
EventGroup: Fix counter in event group state

robem:
When reading event groups from storage, we have to compute the tag-specifc
event group id because it isn't stored. We know that the it always increases by
one but the starting point can be different if we prune. Therefore, this change
computes the starting tag-specific event group id based on the oldest
public/tag-specific event groups available. Given how the counter is used we
have initialize it to an ID that is one before the oldest or zero.


shgandhi (approved):

shgandhi:

PR #2048, opened 2021-11-18
Making sure to get the reference only once.

arc-vmware:
https://github.com/vmware/concord-bft/pull/2033#discussion_r750131695
Improve the fetch of the reference (fetch only once and not millions
of times) to the static thread pool, not on it's creation which is
done only once.


glevkovich (approved):
@arc-vmware , thanks for the change.
Remember that this will be valid only when RequestThreadPool is a singelton.
If not (due to some change in the future) this will not work: if we create/destroy the whole replica without loading the library. Even if this doesn't work, it's an easy fix and easy to detect, so I'm not worried about a possible future crash during development.

arc-vmware:
thanks @glevkovich

PR #2053, opened 2021-11-18
Correctly propagate exceptions from the TRC to the event service

shgandhi:
This PR adds changes to propagate exceptions correctly from the TRC to the client service. Please see individual commits for more detail.

### DIFF
--- a/client/concordclient/include/client/concordclient/event_update_queue.hpp
+++ b/client/concordclient/include/client/concordclient/event_update_queue.hpp
@@ -17,6 +17,7 @@
 #include <list>
 #include <memory>
 #include <mutex>
+#include <exception>
 
 #include "client/concordclient/event_update.hpp"
 
@@ -73,6 +74,8 @@ class UpdateQueue {
   virtual std::unique_ptr<EventVariant> tryPop() = 0;
 
   virtual uint64_t size() = 0;
+
+  virtual void setException(std::exception_ptr e) = 0;
 };
 
 // Basic UpdateQueue implementation provided by this library. This class can be expected to adhere to the UpdateQueue
@@ -85,6 +88,7 @@ class BasicUpdateQueue : public UpdateQueue {
   std::mutex mutex_;
   std::condition_variable condition_;
   bool release_consumers_;
+  std::exception_ptr exception;
 
  public:
   // Construct a BasicUpdateQueue.
@@ -107,6 +111,7 @@ class BasicUpdateQueue : public UpdateQueue {
   virtual std::unique_ptr<EventVariant> pop() override;
   virtual std::unique_ptr<EventVariant> tryPop() override;
   virtual uint64_t size() override;
+  virtual void setException(std::exception_ptr e) override;
 };
 
 }  // namespace concord::client::concordclient
--- a/client/concordclient/src/event_update_queue.cpp
+++ b/client/concordclient/src/event_update_queue.cpp
@@ -49,6 +49,9 @@ void BasicUpdateQueue::push(unique_ptr<EventVariant> update) {
 
 unique_ptr<EventVariant> BasicUpdateQueue::pop() {
   unique_lock<mutex> lock(mutex_);
+  if (exception) {
+    std::rethrow_exception(exception);
+  }
   while (!(release_consumers_ || (queue_data_.size() > 0))) {
     condition_.wait(lock);
   }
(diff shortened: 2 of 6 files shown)


client/concordclient/include/client/concordclient/event_update_queue.hpp
```diff
@@ -73,6 +73,8 @@ class UpdateQueue {
   virtual std::unique_ptr<EventVariant> tryPop() = 0;
 
   virtual uint64_t size() = 0;
+
+  virtual void setException(std::exception_ptr e) = 0;
```
robem, line 77: There is probably a transient include happening. We should add `#include <exception>`.

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -733,13 +736,13 @@ ThinReplicaClient::~ThinReplicaClient() {
     ConcordAssert(subscription_thread_->joinable());
     try {
       subscription_thread_->join();
```
robem, line 739: `join` will not throw any application-specific exceptions. Instead, it might throw `system_error`.
Meaning, we only need to catch `std::exception` or `system_error` directly.
Therefore, we can remove all the "catch-code" here and for all other `join()` calls.
[https://en.cppreference.com/w/cpp/language/throw](https://en.cppreference.com/w/cpp/language/throw)

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -993,7 +995,8 @@ void ThinReplicaClient::Subscribe(uint64_t block_id) {
   // Create and launch thread to stream updates from the servers and push them
   // into the queue.
   stop_subscription_thread_ = false;
-  subscription_thread_.reset(new thread(&ThinReplicaClient::receiveUpdates, this));
+  subscription_thread_.reset(
+      new thread(&ThinReplicaClient::receiveUpdatesWrapper, this, std::ref(config_->update_queue)));
```
robem, line 999: `config_` is a member variable of the ThinReplicaClient object. We invoke the wrapper function on an already instantiated ThinReplicaClient object which is `this`. Therefore, the function will have access to `config_` so that we don't need this parameter.

robem (changes requested):

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -573,6 +573,7 @@ void ThinReplicaClient::receiveUpdates() {
       if (servers_with_hash >= (2 * config_->max_faulty + 1) && !hash_has_agreement) {
         std::string msg{"Internal Error: Couldn't find agreement, 2f+1 replicas generated mismatching hashes"};
         LOG_ERROR(logger_, msg);
+        stop_subscription_thread_ = true;
```
robem, line 577: Instead of setting it before every `throw`, we can set this variable inside the wrapper's catch statement.

client/concordclient/src/event_update_queue.cpp
```diff
@@ -63,6 +63,9 @@ unique_ptr<EventVariant> BasicUpdateQueue::pop() {
 
 unique_ptr<EventVariant> BasicUpdateQueue::tryPop() {
   lock_guard<mutex> lock(mutex_);
+  if (exception) {
```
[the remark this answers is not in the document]
shgandhi: Thanks for catching, added.

client/concordclient/include/client/concordclient/event_update_queue.hpp
shgandhi: Updated.

client/thin-replica-client/src/thin_replica_client.cpp
shgandhi: Thanks for catching, updated.

client/thin-replica-client/src/thin_replica_client.cpp
shgandhi: Yes, updated.

client/thin-replica-client/src/thin_replica_client.cpp
shgandhi: Updated.

shgandhi:
Thanks for the review @robem, updated the PR as suggested.

robem (approved):
Looks good 💪  Thanks for taking care of this!

PR #2060, opened 2021-11-23
[Client pool] Push client to the top of the queue

Liorgal28:
As of now, when a client finishes handling some work, he returns to the end of the queue to allow the other clients to work.

We are changing it to be pushed to the top of the queue so that we can use as few new clients as possible.


yuliasherman (approved):
LGTM, thanks!

teoparvanov (approved):
LGTM, thanks Lior!

PR #2063, opened 2021-11-23
Do not rethrow same exception everytime upon pop() and tryPop() call

shgandhi:
This commit clears the BasicUpdateQueue exception_ptr when an exception is caught during update_queue pop() and tryPop(). If exception_ptr is not reset to nullptr, after the first time an exception is caught and set, we keep on throwing the same exception every time pop()/tryPop() is called.

This commit also includes a configuration fix and some refactoring, see individual commits for more detail.

### DIFF
--- a/client/concordclient/test/cc_basic_update_queue_test.cpp
+++ b/client/concordclient/test/cc_basic_update_queue_test.cpp
@@ -12,6 +12,7 @@
 // file.
 
 #include "client/concordclient/event_update_queue.hpp"
+#include "client/concordclient/concord_client_exceptions.hpp"
 
 #include <thread>
 #include <chrono>
@@ -37,6 +38,7 @@ using namespace std::chrono_literals;
 using concord::client::concordclient::BasicUpdateQueue;
 using concord::client::concordclient::EventVariant;
 using concord::client::concordclient::Update;
+using concord::client::concordclient::OutOfRangeSubscriptionRequest;
 
 const milliseconds kBriefDelayDuration = 10ms;
 const uint64_t kNumUpdatesToTest = (uint64_t)1 << 18;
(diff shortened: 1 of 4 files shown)


client/concordclient/test/cc_basic_update_queue_test.cpp
```diff
@@ -81,6 +83,42 @@ TEST(trc_basic_update_queue_test, test_release_consumers) {
...
+TEST(trc_basic_update_queue_test, test_pop_with_exception) {
+  auto queue = make_unique<BasicUpdateQueue>();
+  auto test_exception = std::make_exception_ptr(OutOfRangeSubscriptionRequest());
+  queue.reset(new BasicUpdateQueue());
+  EXPECT_NO_THROW(queue->setException(test_exception));
+  EXPECT_THROW(queue->pop();, OutOfRangeSubscriptionRequest);
+
+  queue->push(MakeUniqueUpdate(1, vector<pair<string, string>>{{"key1", "value1"}}));
+  queue->push(MakeUniqueUpdate(2, vector<pair<string, string>>{{"key2", "value2"}}));
+  thread consumer0([&]() { EXPECT_NO_THROW(queue->pop()) << "BasicUpdateQueue::pop call failed with an exception."; });
```
robem, line 95: No need for threading here. You're testing the exception behaviour of the update queue only. Same for `tryPop`.

```
queue->push()
queue->push()
queue->pop()
queue->setException()
EXPECT_THROW(queue->pop())
EXPECT_NO_THROW(queue->pop())
```

robem (approved):

client/concordclient/test/cc_basic_update_queue_test.cpp
shgandhi: Updated.

PR #2068, opened 2021-11-25
Stop pre execution on wedged replica

yontyon:
Stop wedged replica from processing pre-process messages.
This is important to solve the case when the primary is wedged and other replicas aren't.
In this case, pre-execution works correctly but the consensus itself doesn't. This will cause liveness issues as no one will complain about the primary.

### DIFF
--- a/bftengine/include/bftengine/ControlStateManager.hpp
+++ b/bftengine/include/bftengine/ControlStateManager.hpp
@@ -30,7 +30,9 @@ class ControlStateManager {
   }
   void setStopAtNextCheckpoint(int64_t currentSeqNum);
   std::optional<int64_t> getCheckpointToStopAt();
-
+  bool isWedged() { return wedged; }
+  void wedge() { wedged = true; }
+  void unwedge() { wedged = false; }
   void markRemoveMetadata(bool include_st = true) { removeMetadataCbRegistry_.invokeAll(include_st); }
   void setPruningProcess(bool onPruningProcess) { onPruningProcess_ = onPruningProcess; }
   bool getPruningProcessStatus() const { return onPruningProcess_; }
@@ -60,6 +62,7 @@ class ControlStateManager {
   ControlStateManager(const ControlStateManager&) = delete;
 
   uint64_t wedgePoint{0};
+  uint64_t wedged = false;
   std::atomic_bool restartBftEnabled_ = false;
   std::unordered_map<uint8_t, SeqNum> hasRestartProofAtSeqNum_;  // reason for restart is the key
   std::atomic_bool onPruningProcess_ = false;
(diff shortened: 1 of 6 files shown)


bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -60,6 +62,7 @@ class ControlStateManager {
   ControlStateManager(const ControlStateManager&) = delete;
 
   uint64_t wedgePoint{0};
+  uint64_t wedged = false;
```
Liorgal28, line 65: Should be bool?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4709,6 +4712,10 @@ void ReplicaImp::executeNextCommittedRequests(concordUtils::SpanWrapper &parent_
   while (lastExecutedSeqNum < lastStableSeqNum + kWorkWindowSize) {
     SeqNum nextExecutedSeqNum = lastExecutedSeqNum + 1;
     SCOPED_MDC_SEQ_NUM(std::to_string(nextExecutedSeqNum));
+    if (ControlStateManager::instance().isWedged()) {
+      LOG_INFO(CNSUS, "system is wedged, no new prePrepare requests will be executed until its unwedged");
```
Liorgal28, line 4716: Most of the times logs without any variables should be DEBUG but it's your call to say if it should be INFO.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4743,36 +4750,38 @@ void ReplicaImp::executeNextCommittedRequests(concordUtils::SpanWrapper &parent_
...
+      std::vector<uint8_t> data_vec;
+      concord::messages::serialize(data_vec, req);
+      std::string sig(SigManager::instance()->getMySigLength(), '\0');
+      uint16_t sig_length{0};
+      SigManager::instance()->sign(reinterpret_cast<char *>(data_vec.data()), data_vec.size(), sig.data(), sig_length);
+      req.signature = std::vector<uint8_t>(sig.begin(), sig.end());
+      data_vec.clear();
+      concord::messages::serialize(data_vec, req);
+      std::string strMsg(data_vec.begin(), data_vec.end());
+      auto sn = std::chrono::duration_cast<std::chrono::microseconds>(getMonotonicTime().time_since_epoch()).count();
```
Liorgal28, line 4773: sn is not a good name - Yulia just asked me to change it in the post exec PR.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1017,6 +1018,11 @@ void PreProcessor::msgProcessingLoop() {
     while (!msgLoopDone_ && msgs_.read_available()) {
       auto msg = msgs_.front();
       msgs_.pop();
+      if (bftEngine::ControlStateManager::instance().isWedged()) {
+        LOG_INFO(logger(), "replica is wedged, ignoring all pre-process requests");
```
Liorgal28, line 1022: Same about the log level.

yuliasherman, line 1022: -> The replica is wedged, the request is ignored

PR #2072, opened 2021-11-26
Initial Key-exchange for lagged replicas on completing ST

nkumar04:
With (n-f) initial key exchange support, we do initial key exchange for live replicas only. Lagged replicas get the public keys of peer replicas from reserved pages. But, we also need to generate key pair and perform key-exchange for lagged replicas as well. this PR addresses this issue.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -795,6 +795,9 @@ void BCStateTran::addOnTransferringCompleteCallback(std::function<void(uint64_t)
   }
   on_transferring_complete_cb_registry_.at(uint64_t(priority)).add(std::move(callback));
 }
+void BCStateTran::addOnFetchingStateChangeCallback(std::function<void(uint64_t)> cb) {
+  if (cb) on_fetching_state_change_cb_registry_.add(std::move(cb));
+}
 
 // this function can be executed in context of another thread.
 void BCStateTran::handleStateTransferMessageImp(char *msg,
@@ -2733,9 +2736,9 @@ void BCStateTran::processData(bool lastInBatch) {
 
       cycleEndSummary();
       sourceSelector_.reset();
-
       g.txn()->setIsFetchingState(false);
       ConcordAssertEQ(getFetchingState(), FetchingState::NotFetching);
+      on_fetching_state_change_cb_registry_.invokeAll(cp.checkpointNum);
       break;
     } else if (!badDataFromCurrentSourceReplica) {
       //////////////////////////////////////////////////////////////////////////
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -474,7 +474,7 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
     }
   }
   delete m;
-}
+}  // namespace bftEngine::impl
 
 template <>
 void ReplicaImp::onMessage<preprocessor::PreProcessResultMsg>(preprocessor::PreProcessResultMsg *m) {
@@ -3101,6 +3101,7 @@ void ReplicaImp::onTransferringCompleteImp(uint64_t newStateCheckpoint) {
   clientsManager->loadInfoFromReservedPages();
 
   KeyExchangeManager::instance().loadPublicKeys();
+
   if (config_.timeServiceEnabled) {
     time_service_manager_->load();
   }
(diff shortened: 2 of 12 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2736,6 +2736,10 @@ void BCStateTran::processData(bool lastInBatch) {
 
       g.txn()->setIsFetchingState(false);
       ConcordAssertEQ(getFetchingState(), FetchingState::NotFetching);
+      // Note: Key exchange message is sent as internal bft client request and
+      // it will be dropped by message handler if curr_state == fetching
+      // So we need to invoke this cb after setting state to NotFetching
+      replicaForStateTransfer_->checkForKeyExchange();
```
yontyon, line 2742: Instead of defining a virtual method in IState transfer, you can simply register a callback to the state transfer call back registry (on BCState transfer)
See https://github.com/vmware/concord-bft/blob/master/bftengine/src/bcstatetransfer/BCStateTran.cpp#L2730

yontyon (changes requested):
The overall logic is good.
I have to comments:
1. BCstate transfer has a callback registry for state transfer completion. You can simply register a call back in the key exchange manager (and pass it a reference to state transfer, if it is not already done).
2. We also talked about rejecting the key exchange messages by the leading quorum if the generated sequence number is out of the working window, correct me if I'm wrong, but I didn't see such a logic in the PR

Other than that, great PR (as usual) :)

bftengine/src/bcstatetransfer/BCStateTran.cpp
nkumar04: The reason i did not register a callback with BCStateTansfer::onTransferringComplete because callbacks are invoked first and ST fetching state is changed later to NOT_FETCHING. 
KeyExchangeManager::sendKeyExchange sends an internal bft client request which non-primary replica forwards to primary for consensus. 
Problem is that, this internal client request msg will get dropped by ReplicaImp::msgHandler because ST fetching state is FETCHING when onTransferComplete() invokes the callbacks. Please check this condition here:
https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#:~:text=%7D-,if%20(!isCollectingState())%20%7B,-//%20The%20message%20validation 

Please let me know if this makes sense

yontyon: correct, you are right.
But wouldn't it be better to have another registry for callbacks after the state == not fetching?
I think that it will simplify the MR as well as create a good infra for future callbacks of this kind.
WDYT?

nkumar04: yes, thats a good point. Let me update the PR as per this. Thanks

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -380,6 +380,16 @@ void ReplicaImp::onMessage<ClientRequestMsg>(ClientRequestMsg *m) {
       return;
     }
   }
+  if ((flags & KEY_EXCHANGE_FLAG) && repsInfo->isIdOfPeerReplica(senderId)) {
```
yontyon, line 383: You can have this logic in the onKeyExchange inside the keyExchangeManager.
Here https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/KeyExchangeManager.cpp#L68 the `sn` argument is for the bft sequence number of the prePrepare containing this keyExchange message.
In addition, you have the generated sequence number in the message itself.
I think it would be more clear to have this logic there

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4099,6 +4110,16 @@ ReplicaImp::ReplicaImp(bool firstTime,
       KeyExchangeManager::instance().sendInitialKey(currentPrimary());
     }
   });
+  stateTransfer->addOnFetchingStateChangeCallback([&](uint64_t checkPointNum) {
+    // With (n-f) initial key exchange support, if we have a lagged replica
+    // which syncs its state through ST, we need to make sure that it completes
+    // initial key exchange after completing ST
+    if (!isCollectingState()) {
+      if (ReplicaConfig::instance().getkeyExchangeOnStart() && !KeyExchangeManager::instance().exchanged()) {
+        KeyExchangeManager::instance().sendInitialKey(currentPrimary(), checkPointNum * checkpointWindowSize);
```
yontyon, line 4119: Nit: You can use here the `lastExecutedSequenceNum` or `lastStableSeqNum` instead of `checkPointNum * checkpointWindowSize`

yontyon (approved):
LGTM!
I have one comment about where you reject the out-of-range keyExhcange message.
Let me know what you think of it

Thanks

PR #2074, opened 2021-11-28
Send prePreapre as response to RFMD even from non-primary replica

yontyon:
Until now, when a replica requests a prePreapre message in RFMD message, only the primary was answering.
This is not good for two reasons:
1. It implies the primary is the only one responsible for distributing missing prePreapre messages - more job to do for the primary
2. In case a replica X is not able to talk with the primary (for example, due to TLS key rotation) it won't be able to execute the request (as it will never get the prePrepare message)

Recall that all of the replicas have the prePrepare messages, so why not let them help the late replica?
This PR fixes this issue


yuliasherman:
Yehonatan, could you make me a favor? Please correct the typo in prePreapre/Prepreare/prepreapre etc.

Liorgal28 (approved):

PR #2078, opened 2021-11-29
Enable TS for state-transfer tests and fix intermittent failures in tests

ukg19:
Enable TS for state-transfer tests and fix intermittent failures in tests

### DIFF
--- a/tests/apollo/util/bft.py
+++ b/tests/apollo/util/bft.py
@@ -1184,7 +1184,51 @@ class BftTestNetwork:
                                     return
 
                                 await trio.sleep(0.5)
-                                               
+
+    async def wait_for_state_transfer_to_stop_with_RFMD(self, up_to_date_node, stale_node, stop_on_stable_seq_num=False):
+        with log.start_action(action_type="wait_for_state_transfer_to_stop_with_RFMD", up_to_date_node=up_to_date_node, stale_node=stale_node, stop_on_stable_seq_num=stop_on_stable_seq_num):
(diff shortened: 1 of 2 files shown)


tests/apollo/util/bft.py
```diff
@@ -1184,7 +1184,51 @@ async def wait_for_state_transfer_to_stop(self, up_to_date_node, stale_node, sto
...
+                with log.start_action(action_type='start_polling', key=key[2], expected_seq_num=expected_seq_num) as action:
+                    last_n = -1
+                    while True:
+                        with trio.move_on_after(.5): # seconds
+                            metrics = await self.metrics.get_all(stale_node)                           
+                            try:
+                                n = self.metrics.get_local(metrics, *key)
+                                # If seq_num is not moving send the new message to advance it
+                                if (n == last_n):
+                                    skvbc = kvbc.SimpleKVBCProtocol(self)
```
cjn, line 1206: Just a query as to do we really need a separate method and instead send these additional requests for advancing time in the existing method wait_for_state_transfer_to_stop()

arc-vmware: We thought about it. But as wait_for_state_transfer_to_stop() is a common function used by many testcases, which donot require this feature of advancing time, will get affected. But we wanted to create a new function so that we can provide RFMD support for the state transfer method.

cjn (approved):

cjn:
LGTM!

PR #2079, opened 2021-11-29
Use raw key in the block merkle latest version CF

dartdart26:
Use the actual raw key instead of the key's hash in the block merkle's
BLOCK_MERKLE_LATEST_KEY_VERSION_CF column family. We need that so we can
serve a state snapshot in lexicographic order on keys. Hashes prevented
us, because:
 * we lose the key itself when we hash and then prune the block the key
   was added in (as we only persist the key itself in blocks)
 * we won't be able to stream in lexicographic order if we store the key
   hashes only

The BLOCK_MERKLE_KEYS_CF column family and other parts of block merkle
don't change in any way.

**Important note**: when this change is merged, the DB will no longer be
backwards compatible. A tool that migrates an existing DB so that it is
compatible with the change in this PR will be provided in a separate
commit.

### DIFF
--- a/kvbc/src/categorization/block_merkle_category.cpp
+++ b/kvbc/src/categorization/block_merkle_category.cpp
@@ -158,8 +158,8 @@ std::vector<Buffer> versionedKeys(const std::vector<std::string>& keys, const st
   return versioned_keys;
 }
 
-void putLatestKeyVersion(NativeWriteBatch& batch, const Hash& key_hash, TaggedVersion version) {
-  batch.put(BLOCK_MERKLE_LATEST_KEY_VERSION_CF, key_hash, serializeThreadLocal(LatestKeyVersion{version.encode()}));
+void putLatestKeyVersion(NativeWriteBatch& batch, const std::string& key, TaggedVersion version) {
+  batch.put(BLOCK_MERKLE_LATEST_KEY_VERSION_CF, key, serializeThreadLocal(LatestKeyVersion{version.encode()}));
 }
 
 void putKeys(NativeWriteBatch& batch,
@@ -167,7 +167,7 @@ void putKeys(NativeWriteBatch& batch,
              std::vector<KeyHash>&& hashed_added_keys,
              std::vector<KeyHash>&& hashed_deleted_keys,
              BlockMerkleInput& updates) {
-  auto kv_it = updates.kv.begin();
+  auto kv_it = updates.kv.cbegin();
   for (auto key_it = hashed_added_keys.begin(); key_it != hashed_added_keys.end(); key_it++) {
     // Only serialize the Header of a DBValue, to prevent the need to copy a potentially large value.
     auto header = toSlice(serializeThreadLocal(DbValueHeader{false, static_cast<uint32_t>(kv_it->second.size())}));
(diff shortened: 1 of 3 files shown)


kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -158,16 +158,16 @@ std::vector<Buffer> versionedKeys(const std::vector<std::string>& keys, const st
   return versioned_keys;
 }
 
-void putLatestKeyVersion(NativeWriteBatch& batch, const Hash& key_hash, TaggedVersion version) {
-  batch.put(BLOCK_MERKLE_LATEST_KEY_VERSION_CF, key_hash, serializeThreadLocal(LatestKeyVersion{version.encode()}));
+void putLatestKeyVersion(NativeWriteBatch& batch, const std::string& key, TaggedVersion version) {
+  batch.put(BLOCK_MERKLE_LATEST_KEY_VERSION_CF, key, serializeThreadLocal(LatestKeyVersion{version.encode()}));
```
cloudnoize, line 162: Do we need to be concerned about data continuity? 
If we upgrade, and previously wrote hash(X) and now X, do we have a way to access the old data?

dartdart26: Yes, we do, I've listed that as an important note in the description. We are going to provide a migration tool for existing blockchains in a separate commit.

cloudnoize (approved):
LGTM!

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -616,34 +611,37 @@ void BlockMerkleCategory::deleteLastReachableBlock(BlockId block_id,
...
     }
     // Remove the value for the key at `block_id`.
     batch.del(BLOCK_MERKLE_KEYS_CF, versioned_key);
   }
   removeMerkleNodes(batch, block_id, out.state_root_version);
 }
 
-std::pair<std::vector<Hash>, std::vector<std::optional<TaggedVersion>>> BlockMerkleCategory::getLatestVersions(
-    const BlockMerkleOutput& out) const {
+std::tuple<std::vector<Hash>, std::vector<std::string>, std::vector<std::optional<TaggedVersion>>>
```
arc-vmware, line 629: should we create vector of tuples:
std::vector<std::tuple<Hash, std::string, std::optional<TaggedVersion>>> 
This will help us to get the complete entity together, but in case of tuple of vector, the mapping across vector for each row need to be done by us. Is there any advantage of tuple of vectors over vector of tuples in this case.

arc-vmware (approved):
LGTM
Thanks @dartdart26 for the changes. I few minor comments or questions.

kvbc/src/categorization/block_merkle_category.cpp
dartdart26: I think it makes sense and is doable. It will require more refactoring, though. I don't have a strong preference. If you think we should change it, I will do it.

arc-vmware: @dartdart26 this change is not needed if its not going to improve anything, please ignore.

dartdart26:
Will open a separate PR that includes the changes in this one plus the migration tool.

PR #2081, opened 2021-11-29
Fix preProcInFlyRequestsNum Metric For Primary

irisk29:
In the primary, the value for this metric was defect (18E).
This metric was never incremented when the replica is the primary but this check was not conducted when we decremented it.


irisk29:
this is the original value in the metric for the primary:
Screen Shot 2021-11-29 at 20 30 06

and this is the value after the fix:
Screen Shot 2021-11-29 at 20 40 20

I tested it by running docker-compose-daml-nano.yml

yuliasherman (approved):
LGTM! Thanks, Iris.

PR #2082, opened 2021-11-29
Fixing non-deterministic case in request handling

arc-vmware:
We have 2 levels of request thread pools, If level 1 schedules all the N threads for its tasks, then a deadlock situation happens where level 1 threads wait for level 2 threads, and level 2 will not be able to schedule any threads. 
This situation will happen only for PrePrepareMsg. If PrePrepareMsg and CheckPointMsg are scheduled together then the checkpoint message will be validated as there are no 2 levels. But if the thread pool is filled only with PrePrepareMsg, then such a situation can happen. This non-determinism would lead to the failure of many Apollo test cases.
Any other combination of task allocation among these 2 levels will not lead to this deadlock.

There are two conditions required to break this deadlock:
a) The creation of threads must be a DAG [This is already happening]
b) For each level of a DAG a different thread pool is used.

The solution to this issue: create a thread pool for each level and make sure that threads in level 'i' don't share the thread pool of level j, for any 2 levels.

In our current solution, we are setting the max level as 2, since we have 2 levels.
The number of threads in level 1 is set to 40
The number of threads in level 2 is set to 24.

So the total number of threads remained same.


yontyon (approved):
LGTM!
Thanks!
Please create a corresponding PR for v0.13.x

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -198,10 +198,17 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
...
+  CONFIG_PARAM(
+      threadbagConcurrencyLevel1,
+      uint32_t,
+      48u,
+      "Number of threads given to thread pool that is created for any request processing for the parent of validation");
+
+  CONFIG_PARAM(
+      threadbagConcurrencyLevel2,
+      uint32_t,
+      32u,
```
yuliasherman, line 210: We are requested to minimize the threads' usage. Are you sure about those numbers?

arc-vmware: Thanks @yuliasherman for the comment. The total number of thread are increasing from 64 to 80. This is a big increase, so I think I should keep the total thread count same. The outer thread will be used by all the messages, so I will keep that at 40. The inner thread is only for PrePrepareMsg, so I will keep that at 24. 
The result for threadcount of threadpool >= 32 showed performance improvement as per our experiments. So Keeping outer thread to 40 will not have any performance degradation.

yuliasherman (approved):
LGTM

PR #2083, opened 2021-11-30
Stabilize reconfiguration tests v0.13.x

yontyon:
To validate the client key exchange, we compared the ".new" file and the old file after the key exchange.
However, this was not good for 2 reasons:

leaving the ".new" fine in the directory
When having two key exchange phases in a row (as in the tests) and CRE is not fast enough, we may restart the clients before even exchanging the keys.
The fix is to remove the ".new" file once we have done with the exchange and compare the actual key before & after the key exchange command


Liorgal28 (approved):

teoparvanov (approved):
Yoni, thanks for the fix!

PR #2085, opened 2021-11-30
Fixing non-deterministic case in request handling

arc-vmware:
We have 2 levels of request thread pools, If level 1 schedules all the N threads for its tasks, then a deadlock situation happens where level 1 threads wait for level 2 threads, and level 2 will not be able to schedule any threads.
This situation will happen only for PrePrepareMsg. If PrePrepareMsg and CheckPointMsg are scheduled together then the checkpoint message will be validated as there are no 2 levels. But if the thread pool is filled only with PrePrepareMsg, then such a situation can happen. This non-determinism would lead to the failure of many Apollo test cases.
Any other combination of task allocation among these 2 levels will not lead to this deadlock.

There are two conditions required to break this deadlock:
a) The creation of threads must be a DAG [This is already happening]
b) For each level of a DAG a different thread pool is used.

The solution to this issue: create a thread pool for each level and make sure that threads in level 'i' don't share the thread pool of level j, for any 2 levels.

In our current solution, we are setting the max level as 2, since we have 2 levels.
The number of threads in level 1 is set to 40
The number of threads in level 2 is set to 24.

So the total number of threads remained same.


yontyon (approved):
LGTM!
Thanks!

yuliasherman (commented):
Could you organize the description of the PR the same way I did it for you in https://github.com/vmware/concord-bft/pull/2082, pls? :-)

arc-vmware:
Thank you @yuliasherman , I updated.

yuliasherman (approved):
LGTM, 10x!

PR #2088, opened 2021-11-30
EventService: Unsubscribe on error

robem:
We return the proper error code to the application but we don't call
unsubscribe which results in us rejecting future subscription requests.
This change makes sure we unsubscribe before we return the error code.


jkk-vmw:
Looks good to me.
A whole lot better than returning early and failing to not unsub.

bayarcher (approved):
Looks good to me.

PR #2089, opened 2021-12-01
Remove span argument from send(). In future use ReadConfig or WriteConfig Hierarchy

bayarcher:
Remove span_context argument from send() function as it is not used to pass span context to call hierarchy.  send() function takes either bft::client::ReadConfig or 'bft::client::WriteConfig' as argument. In both the config hierarchy, need to populate 'request.span_context' to enable Opentracing.

### DIFF
--- a/client/clientservice/src/request_service.cpp
+++ b/client/clientservice/src/request_service.cpp
@@ -74,12 +74,12 @@ Status RequestServiceImpl::Send(ServerContext* context, const Request* proto_req
     bft::client::ReadConfig config;
     config.request = req_config;
     auto span = opentracing::Tracer::Global()->StartSpan("send_ro", {});
-    client_->send(config, std::move(msg), span, callback);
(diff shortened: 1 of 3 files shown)


client/clientservice/src/request_service.cpp
```diff
@@ -74,12 +74,12 @@ Status RequestServiceImpl::Send(ServerContext* context, const Request* proto_req
     bft::client::ReadConfig config;
     config.request = req_config;
     auto span = opentracing::Tracer::Global()->StartSpan("send_ro", {});
```
robem, line 76: Now this is an unused variable. How do we forward traces without this API?

bayarcher: Added how to use bft::client::ReadConfig or 'bft::client::WriteConfig' to the function call send() in the comments.

robem (approved):
Lgtm. Thanks for correcting this! 💪

PR #2092, opened 2021-12-01
Increase the reply size of pollBasedStateClient of CRE

yontyon:
Some of the reconfiguration status replies may be larger than 64KB (like key exchange status for many clients)
In this PR we increase this size to 1MB


nkumar04 (approved):
LGTM!!

PR #2097, opened 2021-12-02
Add validation for max client batch size

evdzhurov:
Add validation for max client batch size when receiving ClientBatchRequestMsg, otherwise a faulty client will be able to bring down a replica by triggering later assertions predicated on the correct size of the accepted batch.


yuliasherman (approved):
LGTM, tahnks for fixing that!

HristoStaykov (approved):
LGTM, thanks for the fix, @evdzhurov !

PR #2100, opened 2021-12-02
Connect Opentracing traces from  Clientservice

bayarcher:
Opentracing root  traces from ClientService are sent in the request using proper API.

### DIFF
--- a/client/clientservice/src/request_service.cpp
+++ b/client/clientservice/src/request_service.cpp
@@ -74,11 +74,17 @@ Status RequestServiceImpl::Send(ServerContext* context, const Request* proto_req
     bft::client::ReadConfig config;
     config.request = req_config;
     auto span = opentracing::Tracer::Global()->StartSpan("send_ro", {});
+    std::ostringstream carrier;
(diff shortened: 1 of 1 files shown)


client/client_pool/include/client/client_pool/client_pool_config.hpp
```diff
@@ -64,6 +64,7 @@ typedef struct ConcordClientPoolConfig {
   std::string signing_key_path = "resources/signing_keys";
   std::uint32_t external_requests_queue_size = 0;
   std::uint32_t trace_sampling_rate = 0;
+  bool enable_traces_in_concord_only_env = false;
```
robem, line 67: Can you explain why we need this flag and why isn't it used in this change?

bayarcher: It will be used in concord code for concord components only test deployments

bayarcher: Removed this changes after discussion with robem.

robem (approved):

PR #2101, opened 2021-12-02
Reverting the change that renames event group column families

shgandhi:
Reverting changes done in commit https://github.com/vmware/concord-bft/commit/1c1a09f608329701b7d344a479edb26ca2e416e9


pkthapa (approved):
LGTM! Thanks for the change @shgandhi

PR #2102, opened 2021-12-03
Ensure exactly once semantics for client batching

evdzhurov:
In this PR an experimental fix is proposed following an investigation about the occurrence of phantom blocks in the skvbc_network_partitioning Apollo test.

Full details on the fix and the investigation can be found here:
https://confluence.eng.vmware.com/display/BLOC/Ensuring+exactly+once+semantics+in+the+presence+of+client+batching

Contents of the fix (note that some unrelated and minor changes have been removed after the initial commit to simplify the PR):
* ClientsManager::deleteOldestReply has been changed to take into account only the sequence number of the reply instead of the wall clock time. Wall clock time should not be necessary for properly deleting the oldest reply. This also simplifies the logic a lot.
* A new method ClientsManager::lastInOrderReplyId has been added. It returns the last consecutive reply id which can be used to reject older replies (from previous batches in the case of client batching).
* ReplicaImp::isReplyAlreadySentToClient now works properly for requests that do not have a corresponding reply in ClientsManager but are older than the last in-order reply.
* The logic for rejecting old requests (older than the last in-order reply) has been restored in ClientsManager::canBecomePending. Note one of the comments "* request seq number is bigger than the last reply seq number". This should be now satisfied with client batching as well.
* ClientsManager::allocateReplyFromSavedOne will no longer trigger a FATAL error if a reply that has been already sent does not exist in the case of disabled batching. This invariant is no longer guaranteed by the calling code. See the code comments for further explanation.

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -120,30 +120,29 @@ bool ClientsManager::hasReply(NodeIdType clientId, ReqId reqSeqNum) const {
   }
 }
 
-void ClientsManager::deleteOldestReply(NodeIdType clientId) {
-  // YS TBD: Once multiple replies for client batching are sorted by incoming time, they could be deleted properly
-  Time earliestTime = MaxTime;
-  ReqId earliestReplyId = 0;
+ReqId ClientsManager::lastInOrderReplyId(NodeIdType clientId) const {
+  auto it = clientsInfo_.find(clientId);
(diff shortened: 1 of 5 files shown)


yuliasherman (changes requested):

yuliasherman (approved):
LGTM

yuliasherman:
This MR changes the central systems' component behavior and needs to pass Maestro tests with a load of at least 700 TPS before a merge.

PR #2103, opened 2021-12-03
TRS: Fix pruned event group transition

robem:
The code assumed that we would always have event group 1 available which is not
true with pruning enabled. Hence, we need to query the oldest available event
group id.


shgandhi:
Thanks for the fix @robem. LGTM!

shgandhi (approved):

PR #2108, opened 2021-12-06
Pass ClientRequest cid to ExecutionRequest instead of the PrePrepareMsg cid

KTeofanidis:
We're setting the cid on all ExecutionRequests to the PrePrepareMsg cid. We need the actual client request cid to be passed in instead for use in some execution implementations.


teoparvanov (approved):
LGTM!

PR #2109, opened 2021-12-06
Adjust remaining window size while building PPM

arc-vmware:
As we know the preprepare creation is now happening in parallel. The threadpool can schedule atmost 64 (this is configurable) preprepare building together. When we try building the preprepare message, we check if "main log" (this is a local rough notebook to keep track of built and in-flight preprepare message), can hold new preprepare or not. If we bypass this check then during actual allocation, we terminate. The check for available space in main log and actual assignment of a place in log for preprepare message is separated, since building of preprepare message is an elaborate process.
But the check and assignment should be in sync and this should also happen for inflight messages.
When the preprepare message was built completely and checked in the same thread, this was little easier. But when it became asynchronous, we have to add a best effort based check which can assure that a place for the fully-built preprepare will always be guaranteed.

So the fix for this issue, is to keep a track of the number of preprepare messages who are not yet built. We are calling such a pre prepare message as active transient preprepare message. This count of active transient preprepare messages will help us to get the instantaneous upper bound of the number of message that will require a place in the main log.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.hpp
+++ b/bftengine/src/bftengine/ReplicaImp.hpp
@@ -120,8 +120,17 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
   size_t NonPrimaryCombinedReqSize = 1000;
   //
   const std::thread::id MAIN_THREAD_ID;
+  // A preprepare message which is still building is called transient preprepare.
+  // activeTransientPreprepare_ represents the number of preprepare messages that are
+  // currently getting build in a separate thread.
+  // This should not be thread safe because its used only in the main thread
+  // This should start with 0, since by default we will assume that a preprepare is taken
+  // to be added to main log. And this will solely represent those preprepares which are
+  // to be added to main log, but are not added as yet.
+  uint16_t numOfTransientPreprepareMsgs_ = 0;
+
   // represents the number of running executions at this moment.
-  // This shoukd not be thread safe because its been changed in 3 places:
+  // This should not be thread safe because its been changed in 3 places:
   // 1) startPrePrepareMsgExecution - when we are getting empty PP its simbulize that we should push
   // FinishPrePrepareExecutionInternalMsg which can happen only on the main thread. 2) executeAllPrePreparedRequests -
   // just before we insert the job to execution thread which means we are in the main thread. 3)
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -14,6 +14,7 @@
 #include <string>
 #include <type_traits>
 #include <bitset>
+#include <limits>
 
 #include "ReplicaImp.hpp"
 #include "Timers.hpp"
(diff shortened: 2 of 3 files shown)


bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -120,8 +120,17 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
...
   //
   const std::thread::id MAIN_THREAD_ID;
+  // A preprepare message which is still building is called transient preprepare.
+  // activeTransientPreprepare_ represents the number of preprepare messages that are
+  // currently getting build in a separate thread.
+  // This should not be thread safe because its used only in the main thread
+  // This should start with 0, since by default we will assume that a preprepare is taken
+  // to be added to main log. And this will solely represent those preprepares which are
+  // to be added to main log, but are not added as yet.
+  uint16_t activeTransientPreprepare_ = 0;
```
yuliasherman, line 130: -> numOfTransientPreprepareMsgs_

yuliasherman (commented):
LGTM. Can we reproduce the issue and verify the fix?

bftengine/src/bftengine/ReplicaImp.hpp
arc-vmware: Thanks @yuliasherman , updated with this change.

arc-vmware:
@yuliasherman yes we will run maestro multiple times and verify the fix. Currently the MR is not going due to some promethius log issue. I have the MR ready, will raise it shortly.

yuliasherman (approved):
LGTM

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -120,8 +120,17 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
...
   //
   const std::thread::id MAIN_THREAD_ID;
+  // A preprepare message which is still building is called transient preprepare.
+  // activeTransientPreprepare_ represents the number of preprepare messages that are
+  // currently getting build in a separate thread.
+  // This should not be thread safe because its used only in the main thread
+  // This should start with 0, since by default we will assume that a preprepare is taken
+  // to be added to main log. And this will solely represent those preprepares which are
+  // to be added to main log, but are not added as yet.
+  uint16_t numOfTransientPreprepareMsgs_ = 0;
```
cloudnoize, line 130: I think this variable should be initialized also when state transfer is completed.

cloudnoize: I'm thinking about what can happen on a view change if, let's say, replica 0 is the primary. A VC happens when some PP is inflight. Can this variable be on a non 0 state after the VC? If it becomes the primary again, will this value be dirty?

arc-vmware: Thanks @cloudnoize for reviewing and raising this question. Let me discuss the interaction with view change, State Transfer and Pruning and effect on this variable here.
How is view change handled?
Ans:
Inside [finishAddingRequestsToPrePrepareMsg()](https://github.com/arc-vmware/concord-bft/blob/25960bdadfb803ba0bda7510aa9361935f7c36d0/bftengine/src/bftengine/ReplicaImp.cpp#L742) function, a thread is created to compute the digest of digests for the given preprepare message. The digest of digests is a time consuming process (it might even take upto 2 ms).
So lets assume that there are ‘x’ preprepare messages in transient state.
 
If any of the preprepare message is done with this computation, it is sent as an internal message (in any order). When this message comes to the main thread it is handled in the [onInternalMsg()](https://github.com/arc-vmware/concord-bft/blob/25960bdadfb803ba0bda7510aa9361935f7c36d0/bftengine/src/bftengine/ReplicaImp.cpp#L1536) handler and it calls the [startConsensusProcess()](https://github.com/arc-vmware/concord-bft/blob/25960bdadfb803ba0bda7510aa9361935f7c36d0/bftengine/src/bftengine/ReplicaImp.cpp#L1563) function. The first line of startConsensusProcess() function calls [isCurrentPrimary()](https://github.com/arc-vmware/concord-bft/blob/25960bdadfb803ba0bda7510aa9361935f7c36d0/bftengine/src/bftengine/ReplicaImp.cpp#L865) and then takes a position in the main log of that replica. If in the meantime a view change happens and it’s no longer a primary, then the preprepare is simply dropped (and a memory leak might happen, which I will fix in 1.6 for the memory leak fix for preprepare action item, which we had discussed earlier).
 
So the view change is actually handled in this scenario.

How ST and Pruning is handled?
The pruning and ST are also handled as below:
Inside [finishAddingRequestsToPrePrepareMsg()](https://github.com/arc-vmware/concord-bft/blob/25960bdadfb803ba0bda7510aa9361935f7c36d0/bftengine/src/bftengine/ReplicaImp.cpp#L742) function, the ST is checked as a part of the prerequisite. Assume that during the handling of preprepare digest of digests the state transfer starts. When we receive the message in the onInternalMsg() handler we [check again for both pruning and ST](https://github.com/arc-vmware/concord-bft/blob/25960bdadfb803ba0bda7510aa9361935f7c36d0/bftengine/src/bftengine/ReplicaImp.cpp#L1557). We will not start consensus and delete the message.
 
So this way both pruning and ST are handled together. I will setup a call today, whenever you have time and discuss the above.

Can this variable become dirty?
There are 3 regions of this variable:
a) negative region -- This will never happen as this is an unsigned variable. I will add a check to reduce this variable only when its positive. But this variable is never reduced more than required. Also this variable is always decremented in the message thread and none of the operations on this variable is happening in parallel. 
b) The count of transient PPM : Desired state and its fine.
c) Its incremented more than the count of transient ppms. This is not possible as its incremented only when we are creating a new thread for computation. And its decrementing only when we are receiving the result of computation. This entire process is happening is same machine, so this variable will become zero.

cloudnoize (approved):
Looks great, thanks @arc-vmware !

PR #2110, opened 2021-12-07
ClientPool: Distinguish reply processing

robem:
The client pool ignores replies that are exactly as big as the maximum
supported size. If there are any callbacks registered with such a reply then
the callback will never be invoked and the application will timeout eventually.
This change distinguishes how replies are handled. If a callback is provided
then we simply hand-over the reply. If no callback is provided then we need to
check the reply size before we copy it into the reply buffer.

### DIFF
--- a/client/client_pool/src/external_client.cpp
+++ b/client/client_pool/src/external_client.cpp
@@ -149,16 +149,24 @@ std::pair<int8_t, ConcordClient::PendingReplies> ConcordClient::SendPendingReque
   }
   try {
     auto res = new_client_->sendBatch(request_queue, batch_cid);
-    for (auto rep : res) {
-      for (auto reply : pending_replies_) {
-        if (reply.lengthOfReplyBuffer == max_reply_size_) continue;
-        auto cid = seq_num_to_cid.find(rep.first)->second;
-        if (reply.cid == cid) {
-          reply.actualReplyLength = rep.second.matched_data.size();
-          memcpy(reply.replyBuffer, rep.second.matched_data.data(), rep.second.matched_data.size());
+    for (const auto& rep : res) {
+      auto cid = seq_num_to_cid.find(rep.first)->second;
+      auto data_size = rep.second.matched_data.size();
+      for (auto& reply : pending_replies_) {
+        if (reply.cid != cid) continue;
+        if (reply.cb) {
           auto result = bftEngine::SendResult{rep.second};
-          if (reply.cb) reply.cb(std::move(result));
-          LOG_DEBUG(logger_, "Request has completed processing" << KVLOG(client_id_, batch_cid, reply.cid));
+          reply.cb(std::move(result));
+          LOG_DEBUG(logger_, "Request processing via callback completed " << KVLOG(client_id_, batch_cid, reply.cid));
+        } else {
+          // TODO: Used for testing only
+          if (data_size > reply.lengthOfReplyBuffer) {
+            LOG_WARN(logger_, "Reply too big " << KVLOG(client_id_, cid, reply.lengthOfReplyBuffer, data_size));
+            continue;
+          }
+          memcpy(reply.replyBuffer, rep.second.matched_data.data(), data_size);
+          reply.actualReplyLength = data_size;
+          LOG_DEBUG(logger_, "Request processing completed " << KVLOG(client_id_, batch_cid, reply.cid));
         }
       }
     }


shgandhi (approved):
Thanks for the fix @robem! LGTM!

client/client_pool/src/external_client.cpp
```diff
@@ -149,17 +154,21 @@ std::pair<int8_t, ConcordClient::PendingReplies> ConcordClient::SendPendingReque
   }
   try {
     auto res = new_client_->sendBatch(request_queue, batch_cid);
-    for (auto rep : res) {
-      for (auto reply : pending_replies_) {
-        if (reply.lengthOfReplyBuffer == max_reply_size_) continue;
```
Liorgal28, line 154: This line was not intended to determine if the returned reply exceeded the limit, but to determine if we even want to do anything with it.
The agreement with DA was to use some internal buffer (like a garbage can) to save the response when the parameter reply_size was set to zero.
Here, the reply size is set to the max_size based on the idea that DA should only send buffer with less than the max size.

Liorgal28: `if (reply_size) {
    pending_reply.replyBuffer = reply_buffer;
    pending_reply.lengthOfReplyBuffer = reply_size;
  } else {
    pending_reply.replyBuffer = reply_->data() + batching_buffer_reply_offset_ * max_reply_size_;
    LOG_DEBUG(logger_,
              "Given reply size is 0, setting internal buffer with offset="
                  << (batching_buffer_reply_offset_ * max_reply_size_));
    ++batching_buffer_reply_offset_;
    pending_reply.lengthOfReplyBuffer = max_reply_size_;
  }`

This is the part of the code that we do that.

yuliasherman: You are right, Lior. But a more correct check would be for the existence of a callback.

client/client_pool/src/external_client.cpp
```diff
@@ -149,17 +154,21 @@ std::pair<int8_t, ConcordClient::PendingReplies> ConcordClient::SendPendingReque
...
-          memcpy(reply.replyBuffer, rep.second.matched_data.data(), rep.second.matched_data.size());
-          auto result = bftEngine::SendResult{rep.second};
-          if (reply.cb) reply.cb(std::move(result));
-          LOG_DEBUG(logger_, "Request has completed processing" << KVLOG(client_id_, batch_cid, reply.cid));
+    for (const auto& rep : res) {
+      auto cid = seq_num_to_cid.find(rep.first)->second;
+      auto data_size = rep.second.matched_data.size();
+      for (auto& reply : pending_replies_) {
+        if (reply.cid != cid) continue;
+        if (data_size > reply.lengthOfReplyBuffer) {
```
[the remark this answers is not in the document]
robem: We agreed offline that the code needs more refactoring :) This PR addresses 1) and 3) we will do 2) in a later change.

client/client_pool/src/external_client.cpp
robem: Thanks!

client/client_pool/include/client/client_pool/concord_client_pool.hpp
```diff
@@ -48,6 +48,7 @@ enum SubmitResult {
   InvalidArgument,
   TimedOut,
   ClientUnavailable,  // There are clients in the queue but none of them are connected to enough replicas
+  Internal,
```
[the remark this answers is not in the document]
robem: `InternalError` is in use already `InternalIssue` it is.

client/client_pool/src/external_client.cpp
```diff
@@ -112,6 +117,9 @@ void ConcordClient::AddPendingRequest(std::vector<uint8_t>&& request,
 
   bftEngine::ClientReply pending_reply;
   if (reply_size) {
+    // We assume a callback is provided in order to distinguish later whether the replyBuffer should be used
+    // TODO: Do we need the else-path?
+    ConcordAssert(callback != nullptr);
```
yuliasherman, line 122: This code works now without CB provided and reply_size set to max_reply_size_.

yuliasherman (approved):
LGTM

client/client_pool/src/external_client.cpp
robem: We clarified: We have tests that don't provide a callback but valid reply buffers. So we cannot have an assert here.

Liorgal28 (approved):

shgandhi (approved):

yuliasherman (approved):
LGTM

PR #2111, opened 2021-12-07
Persist time and recover from it if needed

cloudnoize:
If we have a crash during execution, we recover by reverting any byproduct of the faulty execution and re-executing
the request.
The data needed for the recovery is saved in a struct ```DescriptorOfLastExecution``` that is being filled and persists prior to the execution of the request.
Time service V2 has a mechanism that adds an epsilon to the timestamp in case it's identical to the stored timestamp.
This mechanism's current behavior is faulty during the recovery process since it adds an epsilon to the time of the recovered request, if that request already visited its ```compareAndUpdate``` method prior to the crash.

The addition of the epsilon creates deviation of the state of the crashed replica from the other replicas as its time for the recovered request is different.

This PR solves it by recovering the time that was prior to the execution of the crashed request.
By persisting the time in the ```DescriptorOfLastExecution``` and setting the time in case we recover from a crash.

### DIFF
--- a/bftengine/src/bftengine/DebugPersistentStorage.cpp
+++ b/bftengine/src/bftengine/DebugPersistentStorage.cpp
@@ -196,7 +196,7 @@ void DebugPersistentStorage::setDescriptorOfLastExecution(const DescriptorOfLast
   ConcordAssert(d.validRequests.numOfBits() <= maxNumOfRequestsInBatch);
 
   hasDescriptorOfLastExecution_ = true;
-  descriptorOfLastExecution_ = DescriptorOfLastExecution{d.executedSeqNum, d.validRequests};
+  descriptorOfLastExecution_ = DescriptorOfLastExecution{d.executedSeqNum, d.validRequests, d.timeInTicks};
 }
 
 void DebugPersistentStorage::setDescriptorOfLastStableCheckpoint(
@@ -376,7 +376,7 @@ DescriptorOfLastExecution DebugPersistentStorage::getDescriptorOfLastExecution()
 
   DescriptorOfLastExecution &d = descriptorOfLastExecution_;
 
-  return DescriptorOfLastExecution{d.executedSeqNum, d.validRequests};
+  return DescriptorOfLastExecution{d.executedSeqNum, d.validRequests, d.timeInTicks};
 }
 
 DescriptorOfLastStableCheckpoint DebugPersistentStorage::getDescriptorOfLastStableCheckpoint() {
--- a/bftengine/src/bftengine/PersistentStorageDescriptors.cpp
+++ b/bftengine/src/bftengine/PersistentStorageDescriptors.cpp
@@ -298,7 +298,8 @@ void DescriptorOfLastNewView::deserializeElement(uint32_t id, char *buf, size_t 
 /***** DescriptorOfLastExecution *****/
 
 bool DescriptorOfLastExecution::equals(const DescriptorOfLastExecution &other) const {
-  return (other.executedSeqNum == executedSeqNum && other.validRequests.equals(validRequests));
+  return (other.executedSeqNum == executedSeqNum && other.validRequests.equals(validRequests) &&
+          other.timeInTicks == timeInTicks);
 }
 
 void DescriptorOfLastExecution::serialize(char *&buf, size_t bufLen, size_t &actualSize) const {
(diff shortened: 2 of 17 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3872,6 +3872,11 @@ ReplicaImp::ReplicaImp(const LoadedReplicaData &ld,
   metric_last_agreed_view_.Get().Set(lastAgreedView);
   metric_current_primary_.Get().Set(getCurrentView() % config_.getnumReplicas());
 
+  if (config_.timeServiceEnabled) {
```
arc-vmware, line 3875: There are 3 constructors for ReplicaImp:
1) ReplicaImp::ReplicaImp(bool firstTime,
2) ReplicaImp::ReplicaImp(const ReplicaConfig &config,
3) ReplicaImp::ReplicaImp(const LoadedReplicaData &ld,
I think this is needed for both firstTime (true or false).
Also this requires ReplicaConfig. So its not possible to add it in ctor (1). So it need to be added to both ctors (1) and (2). My suggestion is to create a separate function which will take the time service enabled parameter and aggregator and then that function should be called from both ctor (2) and (3)

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4724,7 +4725,7 @@ void ReplicaImp::markSpecialRequests(RequestsIterator &reqIter,
   reqIter.restart();
 
   if (ps_) {
-    DescriptorOfLastExecution execDesc{lastExecutedSeqNum + 1, requestSet};
+    DescriptorOfLastExecution execDesc{lastExecutedSeqNum + 1, requestSet, time_service_manager_->getTime().count()};
```
arc-vmware, line 4728: Should this happen for both timeservice enabled and disabled case?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5221,7 +5222,8 @@ void ReplicaImp::executeRequestsInPrePrepareMsg(concordUtils::SpanWrapper &paren
       reqIter.restart();
 
       if (ps_) {
-        DescriptorOfLastExecution execDesc{lastExecutedSeqNum + 1, requestSet};
+        DescriptorOfLastExecution execDesc{
```
arc-vmware, line 5225: Should this happen in both case, time service enabled and disabled?

bftengine/src/bftengine/DebugPersistentStorage.cpp
```diff
@@ -196,7 +196,7 @@ void DebugPersistentStorage::setDescriptorOfLastExecution(const DescriptorOfLast
   ConcordAssert(d.validRequests.numOfBits() <= maxNumOfRequestsInBatch);
 
   hasDescriptorOfLastExecution_ = true;
-  descriptorOfLastExecution_ = DescriptorOfLastExecution{d.executedSeqNum, d.validRequests};
+  descriptorOfLastExecution_ = DescriptorOfLastExecution{d.executedSeqNum, d.validRequests, d.timeInTicks};
```
arc-vmware, line 199: Generic comment for the persistent files, if timeservice is not enabled then everything should be either set to 0 or to some INVALID_TIME (which can be a time in far far future). So all these files should not assume that time service is available by default till 1.7 atleast.

arc-vmware (commented):
Thanks @cloudnoize for making these changes. I have very few comments as most of the code looks good to me.

bftengine/src/bftengine/ReplicaImp.cpp
cloudnoize: great catch thanks

bftengine/src/bftengine/DebugPersistentStorage.cpp
cloudnoize: thanks, I'll make sure proper behavior in both cases

arc-vmware (approved):
LGTM

bftengine/src/bftengine/PersistentStorageDescriptors.cpp
```diff
@@ -312,21 +313,31 @@ void DescriptorOfLastExecution::serialize(char *&buf, size_t bufLen, size_t &act
   uint32_t bitMapSize = 0;
   validRequests.writeToBuffer(buf, validRequests.sizeNeededInBuffer(), &bitMapSize);
   buf += bitMapSize;
-  actualSize = executedSeqNumSize + bitMapSize;
+
+  auto sizeOfTick = sizeof(ConsensusTickRep);
+  memcpy(buf, &timeInTicks, sizeOfTick);
+
```
yuliasherman, line 319: You need to move the buf pointer forward: buf += sizeOfTick;

yuliasherman (approved):
LGTM

PR #2117, opened 2021-12-08
Clientservice: Remove unused parameter

robem:
Client batching is set inside the configuration file. The command-line
parameter is never used and was only a placeholder for bft-specific arguments.


shgandhi (commented):
LGTM!

shgandhi (approved):

PR #2118, opened 2021-12-09
Persist time and recover from it if needed

cloudnoize:
Cherry pick of 92784efbad2fc0d75a82c2853f6ccfcfd9900779

If we have a crash during execution, we recover by reverting any byproduct of the faulty execution and re-executing
the request.
The data needed for the recovery is saved in a struct DescriptorOfLastExecution that is being filled and persists prior to the execution of the request.
Time service V2 has a mechanism that adds an epsilon to the timestamp in case it's identical to the stored timestamp.
This mechanism's current behavior is faulty during the recovery process since it adds an epsilon to the time of the recovered request, if that request already visited its compareAndUpdate method prior to the crash.

The addition of the epsilon creates deviation of the state of the crashed replica from the other replicas as its time for the recovered request is different.

This PR solves it by recovering the time that was prior to the execution of the crashed request.
By persisting the time in the DescriptorOfLastExecution and setting the time in case we recover from a crash.

### DIFF
--- a/bftengine/include/bftengine/TimeService.hpp
+++ b/bftengine/include/bftengine/TimeService.hpp
@@ -14,7 +14,8 @@
 #include <chrono>
 
 namespace bftEngine {
-using ConsensusTime = std::chrono::milliseconds;
+using ConsensusTickRep = std::int64_t;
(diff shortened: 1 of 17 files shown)


yuliasherman (approved):
LGTM

PR #2120, opened 2021-12-09
Remove assert condition that checks tx signing key based on secret url

nkumar04:
Secret manager config does not depend of transaction signing feature. In this PR, we remove the assert that manadates the presence of transaction_signing_key when secret_config is enabled


glevkovich (approved):
LGTM!

PR #2122, opened 2021-12-09
Fix handling of encrypted private keys

shgandhi:
This PR fixes two issues -
1. If encrypted_config_enabled is disabled, but secrets_url was non-empty, concord client still tried to decrypt non-existent encrypted private keys. 
2. secret_data in client pool config was not being configured via the secrets_url value in ConcordClientConfig.

### DIFF
--- a/client/clientservice/src/main.cpp
+++ b/client/clientservice/src/main.cpp
@@ -127,7 +127,7 @@ int main(int argc, char** argv) {
     auto yaml = YAML::LoadFile(opts["config"].as<std::string>());
     parseConfigFile(config, yaml);
     std::optional<std::string> secrets_url = std::nullopt;
-    if (opts.count("secrets-url")) {
+    if (opts.count("secrets-url") && config.topology.encrypted_config_enabled) {
       secrets_url = {opts["secrets-url"].as<std::string>()};
     }
     configureSubscription(config,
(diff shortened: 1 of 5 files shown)


client/clientservice/src/main.cpp
```diff
@@ -128,8 +129,11 @@ int main(int argc, char** argv) {
     auto yaml = YAML::LoadFile(opts["config"].as<std::string>());
     parseConfigFile(config, yaml);
     std::optional<std::string> secrets_url = std::nullopt;
-    if (opts.count("secrets-url")) {
+    if (opts.count("secrets-url") && config.topology.encrypted_config_enabled) {
       secrets_url = {opts["secrets-url"].as<std::string>()};
+      if (secrets_url) {
+        config.transport.secret_data = concord::secretsmanager::secretretriever::retrieveSecret(*secrets_url);
```
robem, line 135: To me, secrets manager is an implementation details of TRC and the client pool but you could argue either way. I'm not a fan of retrieving the secret different for each of them. Would you mind adding a todo comment? Either we change the TRC or we change the client pool so that it is unified.

robem (approved):

client/clientservice/src/main.cpp
shgandhi: Updated the TRC implementation of decryptPrivateKey to directly consume secret_data from the configuration.

PR #2124, opened 2021-12-10
Fix handling of encrypted private keys

shgandhi:
This PR fixes two issues -

1. If encrypted_config_enabled is disabled, but secrets_url was non-empty, concord client still tried to decrypt non-existent encrypted private keys.
2. secret_data in client pool config was not being configured via the secrets_url value in ConcordClientConfig.


robem (approved):

shgandhi:
Added a commit that disabled failing test - `test_positive_write_batching_enabled` in `test_skvbc_client_transaction_signing.py`.

PR #2127, opened 2021-12-12
[WIP] Don't save RSI in reserved pages

yontyon:
Apparently, we save in the reserved pages the replica's RSI data. 
As this data may be different across the replicas, it means that we may have inconsistency in the hash of the reserved page, resulting in state split.
This PR fixes this issue

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -36,13 +36,16 @@ void ClientsManager::ClientRsiManager::init() {
     std::vector<std::tuple<uint64_t, uint64_t, std::string>> clientSavedData;
     for (uint32_t offset = 0; offset < maxClientBatchSize_; offset++) {
       std::string data = ps_->getReplicaSpecificInfo(clientId * maxClientBatchSize_ + offset);
+      if (data.empty()) continue;
       uint64_t index = 0;
-      memcpy(&index, data.data(), sizeof(uint64_t));
+      memcpy(&index, data.data(), indexSize);
       uint64_t reqSeqNum = 0;
-      memcpy(&reqSeqNum, data.data() + sizeof(uint64_t), sizeof(uint64_t));
+      memcpy(&reqSeqNum, data.data() + indexSize, reqSeqNumSize);
       uint64_t rsi_size = 0;
-      memcpy(&rsi_size, data.data() + 2 * sizeof(uint64_t), sizeof(uint64_t));
-      std::string rsiData(data.data() + rsiPrefixSize, rsi_size);
+      memcpy(&rsi_size, data.data() + indexSize + reqSeqNumSize, dataLengthSize);
+      std::string rsiData;
+      rsiData.resize(rsi_size);
+      memcpy(rsiData.data(), data.data() + rsiPrefixSize, rsi_size);
       clientSavedData.emplace_back(index, reqSeqNum, rsiData);
     }
     std::sort(clientSavedData.begin(), clientSavedData.end(), [](auto& data1, auto& data2) {
@@ -51,7 +54,7 @@ void ClientsManager::ClientRsiManager::init() {
     for (const auto& rsiData : clientSavedData) {
       rsiCache_[clientId].emplace_back(std::get<1>(rsiData), std::get<2>(rsiData));
     }
-    clientsIndex_[clientId] = std::get<0>((clientSavedData.back()));
+    if (!clientSavedData.empty()) clientsIndex_[clientId] = std::get<0>((clientSavedData.back()));
   }
 }
 std::string ClientsManager::ClientRsiManager::getRsiForClient(uint32_t clientId, uint64_t reqSenNum) {
@@ -70,10 +73,10 @@ void ClientsManager::ClientRsiManager::setRsiForClient(uint32_t clientId,
   // The structure of the RSI in the persistent storage is: [(uint64_t) index | (uint64_t) requestSeqNum | (uint64_t)
   // dataSize | char* data]
   UniquePtrToChar serializedData(new char[rsiPrefixSize + rsiData.size()]);
-  memcpy(serializedData.get(), &nextClientIndex, sizeof(uint64_t));
-  memcpy(serializedData.get() + sizeof(uint64_t), &reqSeqNum, sizeof(uint64_t));
+  memcpy(serializedData.get(), &nextClientIndex, indexSize);
+  memcpy(serializedData.get() + indexSize, &reqSeqNum, reqSeqNumSize);
   uint64_t dataSize = rsiData.size();
-  memcpy(serializedData.get() + 2 * sizeof(uint64_t), &dataSize, sizeof(uint64_t));
+  memcpy(serializedData.get() + indexSize + reqSeqNumSize, &dataSize, dataLengthSize);
   memcpy(serializedData.get() + rsiPrefixSize, rsiData.data(), rsiData.size());
   ps_->beginWriteTran();
   ps_->setReplicaSpecificInfo(storageIndex, serializedData.get(), rsiData.size() + rsiPrefixSize);
@@ -265,7 +268,6 @@ std::unique_ptr<ClientReplyMsg> ClientsManager::allocateNewReplyMsgAndWriteToSto
     const uint32_t sizePage = ((i < numOfPages - 1) ? sizeOfReservedPage() : sizeLastPage);
     saveReservedPage(firstPageId + i, sizePage, ptrPage);
   }
-
   // now save the RSI in the rsiManager
   rsiManager_->setRsiForClient(clientId, requestSeqNum, std::string(reply + commonMsgSize, rsiLength));
   // we cannot set the RSI metadata before saving the reply to the reserved paged, hence save it now.
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -17,6 +17,7 @@
 #include "ReplicaConfig.hpp"
 #include "bftengine/KeyExchangeManager.hpp"
 #include "Serializable.h"
+#include "PersistentStorageImp.hpp"
 
 #include <chrono>
 using namespace std::chrono;
@@ -25,6 +26,14 @@ namespace bftEngine::impl {
 // Initialize:
 // * map of client id to indices.
 // * Calculate reserved pages per client.
+ClientsManager::ClientsManager(std::shared_ptr<PersistentStorage> ps,
+                               const std::set<NodeIdType>& proxyClients,
+                               const std::set<NodeIdType>& externalClients,
+                               const std::set<NodeIdType>& internalClients,
+                               concordMetrics::Component& metrics)
+    : ClientsManager{proxyClients, externalClients, internalClients, metrics} {
+  ps_ = ps;
+}
 ClientsManager::ClientsManager(const std::set<NodeIdType>& proxyClients,
                                const std::set<NodeIdType>& externalClients,
                                const std::set<NodeIdType>& internalClients,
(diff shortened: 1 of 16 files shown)


bftengine/src/bftengine/PersistentStorageImp.hpp
```diff
@@ -99,14 +99,18 @@ enum DescMetadataParameterIds {
   LAST_NEW_VIEW_DESC
 };
 
+const uint32_t clientsDataDescNum = LAST_NEW_VIEW_DESC + 1;
```
yuliasherman, line 102: Define enum instead for better readability.

bftengine/src/bftengine/PersistentStorageImp.hpp
```diff
@@ -99,14 +99,18 @@ enum DescMetadataParameterIds {
...
 typedef unique_ptr<MetadataStorage::ObjectDesc[]> ObjectDescUniquePtr;
 
 class PersistentStorageImp : public PersistentStorage {
  public:
   static constexpr auto kMaxUserDataSizeBytes = 256;
 
  public:
-  PersistentStorageImp(uint16_t numReplicas, uint16_t fVal, uint16_t cVal);
+  PersistentStorageImp(
+      uint16_t numReplicas, uint16_t fVal, uint16_t cVal, uint64_t numClients, uint64_t numOfClientBatch);
```
yuliasherman, line 113: numOfClientBatch -> maxClientBatchSize

yuliasherman, line 113: numClients could not include RO replicas, split the parameter to two or give a more descriptive name.

bftengine/src/bftengine/PersistentStorageImp.hpp
```diff
@@ -250,6 +256,11 @@ class PersistentStorageImp : public PersistentStorage {
   const uint16_t fVal_;
   const uint16_t cVal_;
 
+  const uint16_t numClients_;
+  const uint16_t numClientBatch_;
```
yuliasherman, line 260: numClientBatch_ -> maxClientBatchSize_

bftengine/src/bftengine/PersistentStorageImp.hpp
```diff
@@ -250,6 +256,11 @@ class PersistentStorageImp : public PersistentStorage {
...
+  std::unordered_map<uint32_t, std::unordered_map<uint64_t, std::string>> rsiCache_;
```
yuliasherman, line 261: Add a comment for uint64_t and std::string meanings

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -121,6 +128,13 @@ ObjectDescUniquePtr PersistentStorageImp::getDefaultMetadataObjectDescriptors(ui
   metadataObjectsArray.get()[LAST_STABLE_CHECKPOINT_DESC].maxSize =
       DescriptorOfLastStableCheckpoint::maxSize(numReplicas_);
 
+  for (auto i = 0; i < numReplicas_ + numClients_; i++) {
+    uint32_t baseDescNum = clientsDataDescNum + viewChangeMsgsNum + numClientBatch_ * i;
+    for (auto j = 0; j < numClientBatch_; j++) {
+      metadataObjectsArray.get()[baseDescNum + j].maxSize =
+          (sizeof(size_t) + 2 * sizeof(uint64_t) + replicaSpecificInfoMaxSize);
+    }
+  }
```
yuliasherman, line 137: Define a new function for this.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -517,8 +531,91 @@ void PersistentStorageImp::setUserDataInTransaction(const void *data, std::size_
...
+                                                  char *rsiData,
+                                                  size_t rsiSize) {
+  // First, find the next available sequence number for this client
+  beginWriteTran();
+  if (rsiLatestIndex.count(clientId) == 0) rsiLatestIndex[clientId] = 0;
+  auto client_index = rsiLatestIndex.at(clientId)++;
+  auto client_circular_index = client_index % numClientBatch_;
+  uint32_t viewChangeMsgsNum = DescriptorOfLastNewView::getViewChangeMsgsNum();
+  uint32_t baseDescNum = clientsDataDescNum + viewChangeMsgsNum + clientId * numClientBatch_;
+  UniquePtrToChar outBuf(new char[replicaSpecificInfoMaxSize + 2 * sizeof(uint64_t) + sizeof(size_t)]);
```
yuliasherman, line 545: Could you use variable names instead of sizeof(uint64_t) and sizeof(size_t)? It is not clear overwise.
You can define constants for those parameters and re-use them. There are a lot of such places below.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -517,8 +531,91 @@ void PersistentStorageImp::setUserDataInTransaction(const void *data, std::size_
...
+  char *outBufPtr = outBuf.get();
+  uint32_t outActualObjectSize = 0;
+  metadataStorage_->read(baseDescNum + client_circular_index,
+                         replicaSpecificInfoMaxSize + 2 * sizeof(uint64_t) + sizeof(size_t),
```
yuliasherman, line 549: The same as above.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -517,8 +531,91 @@ void PersistentStorageImp::setUserDataInTransaction(const void *data, std::size_
...
+                         outBufPtr,
+                         outActualObjectSize);
+  if (outActualObjectSize > 0) {
+    // If we do have a saved value on this index, erase it from the cache
+    uint64_t oldSn = 0;
+    std::memcpy(&oldSn, outBufPtr + sizeof(uint64_t), sizeof(uint64_t));
```
yuliasherman, line 555: The same as above.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -517,8 +531,91 @@ void PersistentStorageImp::setUserDataInTransaction(const void *data, std::size_
...
+    ConcordAssert(oldSn > 0);
+    rsiCache_[clientId].erase(oldSn);
+  }
+
+  // Now write the rsi data for this [client, sequence number] to the storage
+  std::string data(rsiData, rsiSize);
+  UniquePtrToChar inBuf(new char[rsiSize + 2 * sizeof(uint64_t) + sizeof(size_t)]);
```
yuliasherman, line 562: The same as above.

bftengine/src/bftengine/PersistentStorageImp.cpp
```diff
@@ -517,8 +531,91 @@ void PersistentStorageImp::setUserDataInTransaction(const void *data, std::size_
...
+  char *inBufPtr = inBuf.get();
+  std::memcpy(inBufPtr, &client_index, sizeof(uint64_t));
+  std::memcpy(inBufPtr + sizeof(uint64_t), &requestSeqNum, sizeof(uint64_t));
+  std::memcpy(inBufPtr + 2 * sizeof(uint64_t), &rsiSize, sizeof(size_t));
+  std::memcpy(inBufPtr + 2 * sizeof(uint64_t) + sizeof(size_t), rsiData, rsiSize);
```
yuliasherman, line 567: The same as above.

bftengine/src/bftengine/PersistentStorageImp.hpp
yontyon: fixed

bftengine/src/bftengine/PersistentStorageImp.hpp
yontyon: done

yontyon: I changed it to include all principles ids (including replicas).
The name has changed accordingly

bftengine/src/bftengine/PersistentStorageImp.hpp
yontyon: fixed

bftengine/src/bftengine/PersistentStorageImp.cpp
yontyon: fixed

bftengine/src/bftengine/PersistentStorageImp.cpp
yontyon: fixed

bftengine/src/bftengine/PersistentStorageImp.cpp
yontyon: fixed

yuliasherman (changes requested):
ClientsManager is not the correct place for those changes either. Please create a separate class for them.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -164,20 +178,22 @@ std::unique_ptr<ClientReplyMsg> ClientsManager::allocateNewReplyMsgAndWriteToSto
 
   c.repliesInfo.insert_or_assign(requestSeqNum, getMonotonicTime());
   LOG_DEBUG(CL_MNGR, KVLOG(clientId, requestSeqNum));
-  auto r = std::make_unique<ClientReplyMsg>(myId_, requestSeqNum, reply, replyLength);
+  auto r = std::make_unique<ClientReplyMsg>(myId_, requestSeqNum, reply, replyLength - rsiLength);
 
-  uint32_t numOfPages = r->size() / sizeOfReservedPage();
+  // At this point, the rsi data is not part of the reply
+  uint32_t commonMsgSize = r->size();
```
yuliasherman, line 184: commonMsgSize -> wholeMsgSize? -> fullMsgSize?

yuliasherman (approved):
LGTM

bftengine/src/bftengine/ClientsManager.cpp
yontyon: The common message size the part of the message that has to be identical between all replicas

yuliasherman: Add a comment explaining that, please.

yontyon:
There is one major limitation with this approach:
A late replica cannot reproduce its RSI after state transfer, this means that these RSIs are gone forever.

A different approach (which we plan to implement) is to have RSIs which can be derived from the common part of the reply.
This way, once a replica gets the replies in the reserved pages, it is able to create its corresponding RSI data.
CC: @guyg7

upshaw-alex:
@yontyon, I am currently attempting to resolve merge conflicts between this PR and my PR https://github.com/vmware/concord-bft/pull/2067, and I had a few questions regarding the changes this PR makes to `ClientsManager`:
1. This PR overloads `ClientsManager`'s constructor, with the added version of the constructor accepting a parameter `std::shared_ptr<PersistentStorage> ps`. Can you confirm it was your intention to overload this constructor rather than just changing its signature and requiring the new `ps` parameter for all `ClientsManager` constructions?
2. I notice your implementation for the new version of `ClientsManager`'s constructor calls the existing version. Assuming you did intentionally overload the constructor, can you confirm it was intentional that the existing version of the constructor has remained `public`, thereby allowing code using `ClientsManager`s to construct and use them without giving a `ps` parameter?
3. . Assuming the answer to my questions 1 and 2 are yes, is it intentional that `ClientsManager::allocateNewReplyMsgAndWriteToStorage` and `ClientsManager::allocateReplyFromSavedOne` segfault when they are used with a `ClientsManager` that was constructed without a `ps` parameter?

yontyon:
@upshaw-alex 
Hi, 
1. Eventually I want to use only the new constructor, to have a smoother change I first only overloaded it.
2. same as above, this is temporarily, only for having a smoothing change.
3. yes, these two methods use persistent storage, so if it is not initiated, they will create segfault.
In general, if you want to remove the old constructor as part of your PR it would be great!

PR #2128, opened 2021-12-13
Use raw key in the block merkle latest version CF

dartdart26:
Use the actual raw key instead of the key's hash in the block merkle's
BLOCK_MERKLE_LATEST_KEY_VERSION_CF column family. We need that so we can
serve a state snapshot in lexicographic order on keys. Hashes prevented
us, because:
 * we lose the key itself when we hash and then prune the block the key
   was added in (as we only persist the key itself in blocks)
 * we won't be able to stream in lexicographic order if we store the key
   hashes only

The BLOCK_MERKLE_KEYS_CF column family and other parts of block merkle
don't change in any way.

The migration keeps track of what steps have been executed such that if
the migration process crashes, it can start fresh or continue from where
it left off. Persisting migration state is implemented by introducing
the `Migration` EDBKeyType.

**Important note**: when this change is merged, the DB will no longer be
backwards compatible. Therefore, a tool that migrates an existing DB so
that it is compatible with the change in this PR is provided too.

In terms of the migration tool, we provide:
 * ability to recover if migration crashed in the middle
 * unit tests
 * automatically detect if migration is needed

### DIFF
--- a/kvbc/CMakeLists.txt
+++ b/kvbc/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required (VERSION 3.2)
 project(libkvbc VERSION 0.1.0.0 LANGUAGES CXX)
 
+find_package(Boost ${MIN_BOOST_VERSION} COMPONENTS filesystem REQUIRED)
 
 add_library(concord_block_update INTERFACE)
 target_include_directories(concord_block_update INTERFACE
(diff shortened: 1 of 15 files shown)


arc-vmware (approved):
LGTM
This code was reviewed earlier in https://github.com/vmware/concord-bft/pull/2079
The current adding for the migration tool looks good.

PR #2131, opened 2021-12-14
Enable time service for remaining apollo tests

ukg19:
This PR aims to enable time service for remaining apollo tests(for which TS was disabled) which are following:
test_skvbc_chaotic_startup
test_skvbc_checkpoints
test_skvbc_network_partitioning
test_skvbc_persistence
test_skvbc_transaction_signing

### DIFF
--- a/tests/apollo/test_skvbc_chaotic_startup.py
+++ b/tests/apollo/test_skvbc_chaotic_startup.py
@@ -36,8 +36,7 @@ def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="10000"):
             "-k", KEY_FILE_PREFIX,
             "-i", str(replica_id),
             "-s", statusTimerMilli,
-            "-v", view_change_timeout_milli,
-            "-f", '0' # disable time service until intermittent failures are fixed for this test
+            "-v", view_change_timeout_milli
             ]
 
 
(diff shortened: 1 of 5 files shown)


tests/apollo/test_skvbc_network_partitioning.py
```diff
@@ -38,7 +38,7 @@ def start_replica_cmd(builddir, replica_id):
             "-i", str(replica_id),
             "-s", statusTimerMilli,
             "-v", viewChangeTimeoutMilli,
-            "-f", '0' # disable time service until intermittent failures are fixed for this test
+            "-f", '1'
```
arc-vmware, line 41: Should we do this after https://github.com/vmware/concord-bft/pull/2102 ?

tests/apollo/test_skvbc_chaotic_startup.py
```diff
@@ -37,7 +37,7 @@ def start_replica_cmd(builddir, replica_id, view_change_timeout_milli="10000"):
             "-i", str(replica_id),
             "-s", statusTimerMilli,
             "-v", view_change_timeout_milli,
-            "-f", '0' # disable time service until intermittent failures are fixed for this test
+            "-f", '1'
```
arc-vmware, line 40: As time service is true by default, all we should do is to remove -f from the command. And anytime we need to configure with disabled TS, we should add -f 0.

ukg19: done

tests/apollo/test_skvbc_network_partitioning.py
ukg19: I have not observed any failure due to TS in this test locally after running it multiple times. I'm testing it in CI with enabled TS. Will give all tests multiple CI runs. If any failures are observed due to TS, I will disable TS for it.

arc-vmware (approved):
LGTM

PR #2135, opened 2021-12-15
EventGroups: Fix event group state initialization

robem:
When all tag-specific event groups got pruned and haven't received any new
updates yet then we need to account for the fact that those event groups have
existed locally. Meaning, our tag-specific counter needs to be initialized with
the last known event group id.
While debugging, the asserts verified our assumptions and keeping them in the
code makes it more readable. In addition, this change adds a test for all
scenarios when a new tag-specific event group state is initialized.

### DIFF
--- a/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
+++ b/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
@@ -305,7 +305,11 @@ uint64_t KvbAppFilter::getValueFromTagTable(const std::string &key) {
     LOG_ERROR(logger_, msg.str());
     throw std::runtime_error(msg.str());
   }
-  return concordUtils::fromBigEndianBuffer<uint64_t>(val->data.data());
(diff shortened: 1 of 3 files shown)


kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -376,6 +381,7 @@ std::optional<uint64_t> KvbAppFilter::getNextEventGroupId(std::shared_ptr<EventG
       // kExecutionEventGroupTagCategory
       uint64_t global_eg_id =
           getValueFromTagTable(kPublicEgId + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(i));
+      ConcordAssertNE(global_eg_id, 0);
```
shgandhi, line 384: Nit: Why do we not assert this in `getValueFromTagTable`?

shgandhi (approved):
Thanks for this fix @robem, and adding the many asserts we've been missing. 💪 
LGTM!

PR #2136, opened 2021-12-15
Allow returning error replies to the client

yuliasherman:
Define an end-to-end infrastructure that enables returning error replies to the clients.


Liorgal28 (approved):
Looks good :)

PR #2137, opened 2021-12-15
[1.5] Fix memory leak 

Liorgal28:
When added post-execution separation we added a defer mechanism.
We defer messages if post-execution is running and push the messages to a bounded queue.

I forgot to delete the message if we already cross the boundary of queue size.


HristoStaykov (approved):
LGTM, good catch @Liorgal28 !

PR #2139, opened 2021-12-16
EventGroups: Fix event group state initialization

robem:
Cherry pick #2135 

When all tag-specific event groups got pruned and haven't received any new
updates yet then we need to account for the fact that those event groups have
existed locally. Meaning, our tag-specific counter needs to be initialized with
the last known event group id.
While debugging, the asserts verified our assumptions and keeping them in the
code makes it more readable. In addition, this change adds a test for all
scenarios when a new tag-specific event group state is initialized.


shgandhi (approved):
I don't know how I missed this. LGTM!

robem:
I cannot seem to re-run the CI but the cherry-pick isn't needed given that event groups are not used in `v0.13.x`.

PR #2142, opened 2021-12-16
Adaptive pruning feature - cmf messages

titusen:
addition of the PruneSwitchModeRequest  message and alteration to existing ones.

### DIFF
--- a/reconfiguration/cmf/concord.cmf
+++ b/reconfiguration/cmf/concord.cmf
@@ -55,77 +55,83 @@ Msg PruneStatus 18 {
     uint64 sender
     bool in_progress
     uint64 last_pruned_block
+    string configuration_json
 }
 
-Msg GetVersionCommand 19 {
+Msg PruneSwitchModeRequest 19 {
+    uint64 sender_id
+    string configuration_json
+}
+
+Msg GetVersionCommand 20 {
   bytes place_holder
 }
 
-Msg InstallCommand 20 {
+Msg InstallCommand 21 {
     string version
     bool bft_support
 }
 
-Msg InstallStatusCommand 21 {
+Msg InstallStatusCommand 22 {
     optional string version
 }
 
-Msg InstallStatusResponse 22 {
+Msg InstallStatusResponse 23 {
     string version
     bool in_progress
     bool install_completed
 }
 
-Msg GetVersionResponse 23 {
+Msg GetVersionResponse 24 {
   string version
 }
 
-Msg ReconfigurationErrorMsg 24 {
+Msg ReconfigurationErrorMsg 25 {
     string error_msg
 }
 
-Msg KeyExchangeCommand 25 {
+Msg KeyExchangeCommand 26 {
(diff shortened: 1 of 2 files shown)


reconfiguration/cmf/concord.cmf
```diff
@@ -55,105 +55,112 @@ Msg PruneStatus 18 {
...
 
-Msg InstallStatusResponse 22 {
+Msg InstallStatusResponse 23 {
     string version
     bool in_progress
     bool install_completed
 }
 
-Msg GetVersionResponse 23 {
+Msg GetVersionResponse 24 {
```
teoparvanov, line 85: What's the impact of shifting all the message id-s? Doesn't it break backwards-compatibility?

teoparvanov: AFAIK, you should add new messages at the end, without changing existing IDs.

reconfiguration/cmf/concord.cmf
```diff
@@ -55,105 +55,112 @@ Msg PruneStatus 18 {
     uint64 sender
     bool in_progress
     uint64 last_pruned_block
+    string configuration_json
 }
 
-Msg GetVersionCommand 19 {
+Msg PruneSwitchModeRequest 19 {
+    uint64 sender_id
+    string configuration_json
```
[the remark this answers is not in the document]
titusen: e.g. 
{ "mode" : "adaptive",
"max" : {"db": 400 , "consensus":800 }}

reconfiguration/cmf/concord.cmf
```diff
@@ -247,6 +248,11 @@ Msg ClientReconfigurationStateReply 49 {
     list ClientStateReply states
 }
 
+Msg PruneSwitchModeRequest 50 {
+    uint64 sender_id
+    string configuration_json
```
cloudnoize, line 253: Who will compose this JSON? the operator?
Maybe it's better to stay in the CMF world and to define a new message for the configuration like
Msg PruneConfiguration{
        uint8 mode
        int rate
}

And then to compose it in that msg, this is how we work with other compositions.

teoparvanov: +1, I'd also prefer that - no need to mix CMF and JSON I believe

reconfiguration/cmf/concord.cmf
titusen: Done

reconfiguration/cmf/concord.cmf
titusen: IMO it couples message format with pruning mechanism. Let's say that there are 10 configurations of adaptive pruning. R we going to have 10 messages for each of them?

teoparvanov: You would have this logical coupling regardless of the format. If you use JSON, the coupling would simply be hidden as a string.

teoparvanov (approved):
LGTM, thanks Tytus!

PR #2145, opened 2021-12-16
0 ticks command handle as stop

titusen:
Due to necessity of stopping pruning, the command start pruning with a pace set 0 will do it.


cloudnoize (approved):
LGTM!

teoparvanov (approved):
Nice!

PR #2146, opened 2021-12-16
Create verifiers if key paths are set in config

nkumar04:
when transaction siging is not enabled, sigmanager does not create verifiers for client pub keys. To validate the requests from cre, we need the verifiers to validate the requests. This PR adds changes to create the verifier if keys path is set


yontyon (approved):
LGTM.
Make sure that the actual transaction signing says off if needed

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -150,6 +150,12 @@ void ClientRequestMsg::validateImp(const ReplicasInfo& repInfo) const {
       }
     }
   }
+  if ((isClientTransactionSigningEnabled == false) && (header->flags & RECONFIG_FLAG)) {
```
yontyon, line 153: `isClientTransactionSigningEnabled == false` is not correct here, because in real deployment the operator is not part of any client group, so its messages cannot be verified here

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -80,13 +81,14 @@ SigManager* SigManager::initImpl(ReplicaId myId,
...
-                        publicKeysMapping,
-                        (publicKeysOfClients != nullptr),
-                        replicasInfo);
+  return new SigManager(
+      myId,
+      numReplicas,
+      make_pair(mySigPrivateKey, replicasKeysFormat),
+      publickeys,
+      publicKeysMapping,
+      ((ReplicaConfig::instance().clientTransactionSigningEnabled) && (publicKeysOfClients != nullptr)),
```
yontyon, line 90: Is it still creating the verifiers even if transaction signing is disabled?

glevkovich:
@nkumar04 , some general comment: you have added here a new functionality to allow loading verifers even when tx signing is disabled, but I don't see any test. Adding the code changes should come with some tests validating the changes, where applicable - in my opinion.
I would assume you've made some manual test, that verifiers are loaded even when TX signing is enabled? That's good for a start, but I would recommend adding a test in Apollo/Hermes (both are possible, Apollo would require some infra change) to validate that your changes actually work, else we have a probabilty to have bugs now or in the future if your assumptions are broken.

bftengine/src/bftengine/SigManager.cpp
nkumar04: yes

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
nkumar04: We are skipping signature verification here. Reason for adding this check here is cre requests are always signed. If tx signing is disabled, we need to skip verification here and it is verified by reconfiguration handler

bftengine/src/bftengine/ReplicaLoader.cpp
```diff
@@ -53,7 +53,9 @@ ReplicaLoader::ErrorCode loadConfig(LoadedReplicaData &ld) {
                                    config.replicaPrivateKey,
                                    config.publicKeysOfReplicas,
                                    concord::util::crypto::KeyFormat::HexaDecimalStrippedFormat,
-                                   config.clientTransactionSigningEnabled ? &config.publicKeysOfClients : nullptr,
+                                   (config.clientTransactionSigningEnabled || !config.clientsKeysPrefix.empty())
```
glevkovich, line 56: nit: same logic repeats 3 times.
For better readibility, and also because there is more logic here, maybe we should have an inline function as part of ReplicaConfig?

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
glevkovich, line 153: 2 comments:
1) The validation of the client ID is very important here. I might not fully understandf the code, but do you ignore isIdOfExternalClient on purpose?
2) I think this statement  should start with an else.
I would refactor this part with the above:
`if (isIdOfExternalClient && isClientTransactionSigningEnabled) {`
Change to
`if (iisClientTransactionSigningEnabled) {`
and have here an else with the rest of the applied logic refactored accordingly.
Else you are always leave the chance to override previous set to true as part of future changes.

bftengine/src/preprocessor/tests/messages/PreProcessResultMsg_test.cpp
```diff
@@ -30,6 +30,7 @@ bftEngine::ReplicaConfig& createReplicaConfigWithExtClient(uint16_t fVal, uint16
 
   auto clientPrincipalIds = std::set<uint16_t>{uint16_t(config.numReplicas + 1)};
   config.publicKeysOfClients.insert(make_pair(config.publicKeysOfReplicas.begin()->second, clientPrincipalIds));
+  config.clientTransactionSigningEnabled = true;
```
glevkovich, line 33: Does the test for  the case
`config.clientTransactionSigningEnabled = false
`
Is completely irrelevant now?
If it is still relevant now, I would recommend having here a parametrized test - run all tests or some parts with thsi option set to true and false.

glevkovich (changes requested):

bftengine/src/bftengine/ReplicaLoader.cpp
nkumar04: fixed

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
nkumar04: fixed

bftengine/src/preprocessor/tests/messages/PreProcessResultMsg_test.cpp
nkumar04: added tests for tx_signing disabled also.

yontyon (approved):
LGTM!
Thanks

nkumar04:
> @nkumar04 , some general comment: you have added here a new functionality to allow loading verifers even when tx signing is disabled, but I don't see any test. Adding the code changes should come with some tests validating the changes, where applicable - in my opinion. I would assume you've made some manual test, that verifiers are loaded even when TX signing is enabled? I That's good for , but I would recommend (I might be missing your test, maybe in another PR?) adding a test in Apollo/Hermes (both are possible, Apollo would require some infra change) to validate that your changes actually work, else we have a probabilty to have bugs now or in the future if your assumptions are broken.

Thanks @glevkovich for the review, i have created jira to add tests on hermes and apollo. i will work on adding the test. I have addressed your comments. Please let me know if you have any other comment. Thank you

glevkovich (approved):
Thank you!

PR #2147, opened 2021-12-16
Remove RSIs of the source replica on metadata reset

yontyon:
Following https://github.com/vmware/concord-bft/pull/2127, on metadata reset we also need to remove the source RSIs (because they are not generated by the restored replica).
We present 4 new parameters one needs to pass to the `resetMetadata` for removing the RSIs - 
F - the fVal of the original blockchain
C - the cVal of the original blockchain
principles - the number of principles in the original blockchain (including, replicas, read-only replicas and bft clients)
MAX_CLIENT_BATCH_SIZE - the client's ma batch size in the original blockchain.

In order not to break the current API, if one of the latest 4 parameters is missing, we won't remove the RSI at all.

### DIFF
--- a/bftengine/src/bftengine/ReplicaSpecificInfoManager.cpp
+++ b/bftengine/src/bftengine/ReplicaSpecificInfoManager.cpp
@@ -64,7 +64,9 @@ void RsiDataManager::init() {
     for (uint32_t offset = 0; offset < max_client_batch_size_; offset++) {
       std::vector<uint8_t> data = ps_->getReplicaSpecificInfo(clientId * max_client_batch_size_ + offset);
       if (data.empty()) continue;
-      clientRsiData.emplace_back(RsiItem::deserialize(data));
+      auto rsi_item = RsiItem::deserialize(data);
+      if (rsi_item.data().empty()) continue;
+      clientRsiData.emplace_back(rsi_item);
     }
     std::sort(clientRsiData.begin(), clientRsiData.end(), [](auto& data1, auto& data2) {
       return data1.index() < data2.index();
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaSpecificInfoManager.cpp
```diff
@@ -64,7 +64,9 @@ void RsiDataManager::init() {
     for (uint32_t offset = 0; offset < max_client_batch_size_; offset++) {
       std::vector<uint8_t> data = ps_->getReplicaSpecificInfo(clientId * max_client_batch_size_ + offset);
       if (data.empty()) continue;
-      clientRsiData.emplace_back(RsiItem::deserialize(data));
+      auto rsi_item = RsiItem::deserialize(data);
+      if (rsi_item.data().empty()) continue;
```
toly-kournik, line 68: Just curious how it could happen?

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -920,6 +931,15 @@ struct ResetMetadata {
       result["stable seq num"] = std::to_string(stableSeqNum);
     }
     p->setPrimaryLastUsedSeqNum(lastExecutedSn);
+    if (removeRsis) {
+      for (uint32_t principle = 0; principle < nVal + principles; principle++) {
```
toly-kournik, line 935: nit: "principle" is not our terminology

toly-kournik (approved):
LGTM!

bftengine/src/bftengine/ReplicaSpecificInfoManager.cpp
yontyon: In case we "remove" the RSI, which means simply setting it to empty string

PR #2148, opened 2021-12-17
Set gRPC max receive message size for clientservice

shgandhi:
The default value for gRPC max receive message size is 4MiB
(see GRPC_DEFAULT_MAX_RECV_MESSAGE_LENGTH). This commit
changes it to 50MiB to avoid gRPC RESOURCE_EXHAUSTED errors.

### DIFF
--- a/client/clientservice/src/main.cpp
+++ b/client/clientservice/src/main.cpp
@@ -57,6 +57,7 @@ po::variables_map parseCmdLine(int argc, char** argv) {
     ("metrics-port", po::value<int>()->default_value(9891), "Prometheus port to query clientservice metrics")
     ("secrets-url", po::value<std::string>(), "URL to decrypt private keys")
     ("jaeger", po::value<std::string>(), "Push trace data to this Jaeger Agent")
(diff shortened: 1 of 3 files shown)


client/clientservice/src/main.cpp
```diff
@@ -57,6 +57,7 @@ po::variables_map parseCmdLine(int argc, char** argv) {
     ("metrics-port", po::value<int>()->default_value(9891), "Prometheus port to query clientservice metrics")
     ("secrets-url", po::value<std::string>(), "URL to decrypt private keys")
     ("jaeger", po::value<std::string>(), "Push trace data to this Jaeger Agent")
+    ("max-receive-msg-size", po::value<int>()->default_value(52428800), "Clientservice max receive message size in bytes")
```
robem, line 60: Ideally, the default would be a much smaller value (such as 4MB or even lower) and during deployment, we decide how big the value should be. I don't know about the gRPC internals so this may or may not lead to unnecessary memory allocations.

robem (approved):
Thanks for the fix and also for making it configurable!

client/clientservice/src/main.cpp
shgandhi: Updated.

PR #2151, opened 2021-12-17
Added validation for message size in client request.

agrkushvm:
Issue : This method may cause the process to crash because of an invalid message msgSize_.
            For example, if the message is not long enough, then header may access to an illegal memory location.

Fix : Added check to verify the size is greater than the default ClientRequestMsgHeader.

### DIFF
--- a/bftengine/src/bftengine/messages/ClientRequestMsg.cpp
+++ b/bftengine/src/bftengine/messages/ClientRequestMsg.cpp
@@ -98,13 +98,28 @@ bool ClientRequestMsg::shouldValidateAsync() const {
 
 void ClientRequestMsg::validateImp(const ReplicasInfo& repInfo) const {
   const auto* header = msgBody();
+  const auto msgSize = size();
+
+  std::stringstream msg;
+  // Check message size is greater than minimum header size
+  if (msgSize < sizeof(ClientRequestMsgHeader) + spanContextSize()) {
+    msg << "Invalid Message Size " << KVLOG(msgSize, sizeof(ClientRequestMsgHeader), spanContextSize());
+    LOG_ERROR(CNSUS, msg.str());
+    throw std::runtime_error(msg.str());
+  }
+
   PrincipalId clientId = header->idOfClientProxy;
   if ((header->flags & RECONFIG_FLAG) == 0) ConcordAssert(this->senderId() != repInfo.myId());
+
   /// to do - should it be just the header?
   auto minMsgSize = sizeof(ClientRequestMsgHeader) + header->cidLength + spanContextSize() + header->reqSignatureLength;
-  const auto msgSize = size();
+  if (msgSize < minMsgSize) {
+    msg << "Invalid msgSize: " << KVLOG(msgSize, minMsgSize);
+    LOG_WARN(CNSUS, msg.str());
+    throw std::runtime_error(msg.str());
+  }
+
   uint16_t expectedSigLen = 0;
-  std::stringstream msg;
   auto sigManager = SigManager::instance();
   bool isClientTransactionSigningEnabled = sigManager->isClientTransactionSigningEnabled();
   bool isIdOfExternalClient = repInfo.isIdOfExternalClient(clientId);
--- a/bftengine/tests/messages/ClientRequestMsg_test.cpp
+++ b/bftengine/tests/messages/ClientRequestMsg_test.cpp
@@ -282,6 +282,63 @@ TEST_F(ClientRequestMsgTestFixture, extra_buffer) {
   ASSERT_EQ(msg.size(), mainMsgSize + extraBufSize);
   ASSERT_EQ(msg.getExtraBufPtr().first, msg.body() + mainMsgSize);
 }
+
+TEST_F(ClientRequestMsgTestFixture, validate_size) {
+  // Inherit from ClientRequestMsg so that some protected methods can be exposed for the test
+  struct TestClientRequestMsg : ClientRequestMsg {
+    TestClientRequestMsg(NodeIdType sender,
+                         uint64_t flags,
+                         uint64_t reqSeqNum,
+                         uint32_t requestLength,
+                         const char* request,
+                         uint64_t reqTimeoutMilli,
+                         const std::string& cid,
+                         const concordUtils::SpanContext& spanContext,
+                         const char* requestSignature,
+                         uint32_t requestSignatureLen,
+                         const uint32_t extraBufSize)
+        : ClientRequestMsg(sender,
+                           flags,
+                           reqSeqNum,
+                           requestLength,
+                           request,
+                           reqTimeoutMilli,
+                           cid,
+                           0,
+                           spanContext,
+                           requestSignature,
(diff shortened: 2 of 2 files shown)


teoparvanov (commented):
I think it's worth adding a unit test for this scenario here:
https://github.com/vmware/concord-bft/blob/master/bftengine/tests/messages/ClientRequestMsg_test.cpp

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -102,11 +102,19 @@ void ClientRequestMsg::validateImp(const ReplicasInfo& repInfo) const {
   const auto* header = msgBody();
   PrincipalId clientId = header->idOfClientProxy;
   if ((header->flags & RECONFIG_FLAG) == 0) ConcordAssert(this->senderId() != repInfo.myId());
+
+  const auto msgSize = size();
+  std::stringstream msg;
+  // Check message size is greater than muinimum header size
+  if (size() < sizeof(ClientRequestMsgHeader) + spanContextSize()) {
```
arc-vmware, line 109: use msgSize instead of size() here.

agrkushvm: Ok.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -102,11 +102,19 @@ void ClientRequestMsg::validateImp(const ReplicasInfo& repInfo) const {
   const auto* header = msgBody();
   PrincipalId clientId = header->idOfClientProxy;
   if ((header->flags & RECONFIG_FLAG) == 0) ConcordAssert(this->senderId() != repInfo.myId());
+
+  const auto msgSize = size();
+  std::stringstream msg;
+  // Check message size is greater than muinimum header size
```
arc-vmware, line 108: sorry, i missed it earlier:
the spelling of minimum should be corrected

agrkushvm:
> I think it's worth adding a unit test for this scenario here: https://github.com/vmware/concord-bft/blob/master/bftengine/tests/messages/ClientRequestMsg_test.cpp

@teoparvanov 
OK.
Should I create a different PR for the unit test case addition or add in this PR only ?

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -102,11 +102,19 @@ void ClientRequestMsg::validateImp(const ReplicasInfo& repInfo) const {
...
+  if (msgSize < sizeof(ClientRequestMsgHeader) + spanContextSize()) {
+    msg << "Invalid Message Size ";
```
arc-vmware, line 110: Add KVLOG(msgSize, sizeof(ClientRequestMsgHeader), spanContextSize(), clientId) to the message, so that we can see what size is coming.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -102,11 +102,19 @@ void ClientRequestMsg::validateImp(const ReplicasInfo& repInfo) const {
...
+    LOG_ERROR(GL, msg.str());
```
arc-vmware, line 111: Use the logger as CNSUS instead of GL.

arc-vmware (changes requested):

arc-vmware:
Please add testcases in this PR for this new check.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
agrkushvm: Ok.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
agrkushvm: Ok

teoparvanov:
> > I think it's worth adding a unit test for this scenario here: https://github.com/vmware/concord-bft/blob/master/bftengine/tests/messages/ClientRequestMsg_test.cpp
> 
> @teoparvanov OK. Should I create a different PR for the unit test case addition or add in this PR only ?

No, I think you should add the unit test as part of this PR.

glevkovich:
```
@teoparvanov
OK.
Should I create a different PR for the unit test case addition or add in this PR only ?
```
@agrkushvm , I think that adding this in another PR misses all the point.
The idea is that if someone adds a change, it also adds the test as part of the current PR (where it makes sense).
Let us assume that your change has a bug we haven't tracked -  pushing it into master and only then testing to find out the test fails won't be a good idea.
In some cases, if the tests require a lot of work it is possible to do so, but mark the 2 PRs as dependent on each other and push them together.

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
```diff
@@ -102,11 +102,19 @@ void ClientRequestMsg::validateImp(const ReplicasInfo& repInfo) const {
   const auto* header = msgBody();
   PrincipalId clientId = header->idOfClientProxy;
   if ((header->flags & RECONFIG_FLAG) == 0) ConcordAssert(this->senderId() != repInfo.myId());
+
```
glevkovich, line 105: @agrkushvm , I think execution might potentially crash already in line 103, you have to place you check before accessing the header fields.
To safetly access any field in header you must check first that
`(msgSize >= sizeof(ClientRequestMsgHeader)`. 

In addition, line 177 is doing a similar check, but potentially too late.
After validating the 1st check and being able to access the header, I would move the check (msgSize < minMsgSize) , i would move this check to the top of the function.
So you need to have 2 checks at the beggining of the function:

1. Before accessing any field in messge body, check that (msgSize >= sizeof(ClientRequestMsgHeader) + spanContextSize())
2. Immidiately after check that  (msgSize < minMsgSize)

glevkovich (changes requested):

bftengine/src/bftengine/messages/ClientRequestMsg.cpp
agrkushvm: ok

arc-vmware (approved):
LGTM

bftengine/tests/messages/ClientRequestMsg_test.cpp
```diff
@@ -282,6 +282,63 @@ TEST_F(ClientRequestMsgTestFixture, extra_buffer) {
...
+                           request,
+                           requestTimeoutMilli,
+                           correlationId,
+                           concordUtils::SpanContext{spanContext},
+                           nullptr,
+                           0,
+                           extraBufSize);
+
+  // Setting the msg size to zero so that it should throw an error. Validatng the size check
+  msg.setSize(0);
```
glevkovich, line 338: nit:
 why not checking on a loop up to (but not include) minMsgSize?

glevkovich (approved):
Thanks @agrkushvm ! 
one smalle comment for the Google Test test case

PR #2152, opened 2021-12-19
Fixed Log Message In The Client Pool

irisk29:
Fix for BC-9727


client/client_pool/src/external_client.cpp
```diff
@@ -154,6 +154,9 @@ std::pair<int8_t, ConcordClient::PendingReplies> ConcordClient::SendPendingReque
       auto data_size = rep.second.matched_data.size();
       for (auto& reply : pending_replies_) {
         if (reply.cid != cid) continue;
+        LOG_INFO(logger_,
```
teoparvanov, line 157: Wouldn't this result in a lot of output in the logs? Just imagine if the system is under a load of 900tps...

Liorgal28: After talking with @cjn we are going to suggest different logs to track the start/finish of single requests.

PR #2157, opened 2021-12-20
Remove Static Reply Buffer In Client Pool

irisk29:
In the current BFT client pool implementation, we allocate one static buffer for all clients in the pool (per request in the batch) and use it as a trash can for reply messages. 
There is no need for that anymore.


yuliasherman (approved):
LGTM, 10x!

PR #2159, opened 2021-12-21
Sign unique request fields along with PrePrepareResult

yuliasherman:
Signing only PrePrepareResult buffer allows a malicious primary replica to make the system fool. Additional request fields should be used in the hash creation process.


Liorgal28 (approved):
Amazing work!

PR #2161, opened 2021-12-21
Merge pull request #2146 from nkumar04/nkumar4/BC-15732

nkumar04:
Create verifiers if key paths are set in config


yontyon (approved):
LGTM!
Thanks

PR #2162, opened 2021-12-22
Rotate the source of preprepare in rfmd mechanism

yontyon:
Recently we introduced a change in which every replica can answer with preprepare as a response for RFMD.
Alas, as preprepare messages might be large, we wouldn't want to overload the system with redundant preprepare messages.
For this, we introduce in this PR a mechanism in which the late replica asks every time from a different node the preprepare message.
This way we benefit from two things:
1. All replicas can answer to the preprepare requests (balancing the load on the primary) and
2. Only one replica answers the preprepare in a round (balancing the overall network usage)

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -3512,7 +3512,7 @@ void ReplicaImp::tryToSendReqMissingDataMsg(SeqNum seqNumber, bool slowPathOnly,
 
     reqData.resetFlags();
 
-    if (missingPrePrepare) reqData.setPrePrepareIsMissing();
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3533,6 +3533,7 @@ void ReplicaImp::tryToSendReqMissingDataMsg(SeqNum seqNumber, bool slowPathOnly,
     send(&reqData, destRep);
     metric_sent_req_for_missing_data_++;
   }
+  source_for_rfmd_pp_ = (source_for_rfmd_pp_ + 1) & config_.numReplicas;
```
Liorgal28, line 3536: If I'm not wrong it should be config_.numReplicas-1 (when source_for_rfmd_pp_ = 6  the result will be 7 that is not good for us)

Liorgal28 (approved):

PR #2166, opened 2021-12-22
Filter duplicate preprocess replies from the same sender

cloudnoize:
Before this PR, multiple identical replies from the same sender could have been added to the preprocess result set.
The duplication would have been discovered in the signatures' validation phase, resulting in a view change.

This PR filters the duplications at the preprocessing phase eliminating the possibility for a view change unless the primary is malicious and adds duplications on purpose.

### DIFF
--- a/bftengine/src/preprocessor/RequestProcessingState.cpp
+++ b/bftengine/src/preprocessor/RequestProcessingState.cpp
@@ -120,10 +120,18 @@ void RequestProcessingState::detectNonDeterministicPreProcessing(const uint8_t *
 }
 
 void RequestProcessingState::handlePreProcessReplyMsg(const PreProcessReplyMsgSharedPtr &preProcessReplyMsg) {
+  SCOPED_MDC_CID(cid_);
   const auto &senderId = preProcessReplyMsg->senderId();
   if (preProcessReplyMsg->status() == STATUS_GOOD) {
-    numOfReceivedReplies_++;
     const auto &newHashArray = convertToArray(preProcessReplyMsg->resultsHash());
+    // Do not add a duplicate signature from the same sender to the list.
+    // insert returns a pair, where the second is a boolean denoting whether the insertion took place.
+    // TODO - identify the scenarios that can cause duplicate replies to arrive.
+    if (!(seenSenders_[newHashArray]).insert(preProcessReplyMsg->senderId()).second) {
+      LOG_INFO(logger(), "Signature was already accepted for this sender " << KVLOG(senderId, reqSeqNum_, clientId_));
+      return;
+    }
+    numOfReceivedReplies_++;
     // Counts equal hashes and saves the signatures with the replica ID. They will be used as a proof that the primary
     // is sending correct pre-execution result to the rest of the replicas.
     preProcessingResultHashes_[newHashArray].emplace_back(preProcessReplyMsg->getResultHashSignature(),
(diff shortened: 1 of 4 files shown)


bftengine/src/preprocessor/RequestProcessingState.hpp
```diff
@@ -128,6 +130,7 @@ class RequestProcessingState {
   // Maps result hash to a list of replica signatures sent for this hash. This also implicitly gives the number of
   // replicas returning a specific hash.
   std::map<concord::util::SHA3_256::Digest, std::list<PreProcessResultSignature>> preProcessingResultHashes_;
+  std::map<concord::util::SHA3_256::Digest, std::unordered_set<bftEngine::impl::NodeIdType>> seen_senders_;
```
eyalrund, line 133: change to camelCase style

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
```diff
@@ -71,9 +71,8 @@ std::optional<std::string> PreProcessResultMsg::validatePreProcessResultSignatur
   auto hash = concord::util::SHA3_256().digest(requestBuf(), requestLength());
   std::unordered_set<bftEngine::impl::NodeIdType> seen_signatures;
   for (const auto& s : sigs) {
-    // insert returns std::pair<iterator, bool>. The bool indicates if the element was created, or it was already in the
-    // set and no insertion was performed. The latter case indicates that we already have got a signature from this
-    // replica.
+    // logic that filters duplicate signatures exist in the preprocessing phase,
```
eyalrund, line 74: logic that filters duplicate signatures was performed in the ....
therefore duplication implies that the primary is probably malicious

bftengine/src/preprocessor/RequestProcessingState.cpp
```diff
@@ -115,17 +115,24 @@ void RequestProcessingState::detectNonDeterministicPreProcessing(const uint8_t *
...
+  SCOPED_MDC_CID(cid_);
   const auto &senderId = preProcessReplyMsg->senderId();
   if (preProcessReplyMsg->status() == STATUS_GOOD) {
-    numOfReceivedReplies_++;
     const auto &newHashArray = convertToArray(preProcessReplyMsg->resultsHash());
+    // Do not add a duplicate signature from the same sender to the list.
+    // insert returns a pair, where the second is a boolean denoting whether the insertion took place.
+    // TODO - identify the scenarios that can cause duplicate replies to arrive.
+    if (!(seen_senders_[newHashArray]).insert(preProcessReplyMsg->senderId()).second) {
+      LOG_INFO(logger(), "Signature was already accepted for this sender " << KVLOG(senderId, reqSeqNum_, clientId_));
```
eyalrund, line 126: should log level be INFO or WARNING?

eyalrund (changes requested):

bftengine/src/preprocessor/RequestProcessingState.cpp
cloudnoize: I've validated with Guy, this scenario is normal

bftengine/src/preprocessor/RequestProcessingState.hpp
```diff
@@ -146,6 +149,7 @@ class RequestProcessingState {
   // Maps result hash to a list of replica signatures sent for this hash. This also implicitly gives the number of
   // replicas returning a specific hash.
   std::map<concord::util::SHA3_256::Digest, std::list<PreProcessResultSignature>> preProcessingResultHashes_;
+  std::map<concord::util::SHA3_256::Digest, std::unordered_set<bftEngine::impl::NodeIdType>> seenSenders_;
```
yuliasherman, line 152: I've checked the serialization/deserialization mechanism in PreProcessResultSignature. It is not affected by the type of used container. I think it is much cleaner to replace the list with the set here. It will also avoid unneeded checks in PreProcessResultMsg::validatePreProcessResultSignatures function.

cloudnoize: yep, the API ```std::string serializeResultSignatureList(const std::list<PreProcessResultSignature>& sigs)``` confused me, I'll use a set and convert it to a list when calling RequestProcessingState::getPreProcessResultSignatures()

yuliasherman: I would change the container type end-to-end.

yuliasherman (approved):
LGTM

PR #2169, opened 2021-12-23
Tls fix in 1.5

Liorgal28:
This PR insert:
1) On TLS code we find that in some situations the write_msg_ could be overridden by 2 different threads and cause the disappearing of messages.


yontyon (approved):
LGTM!

PR #2171, opened 2021-12-23
Don't exchange tls keys if already exist v0.13

yontyon:
Assume we have a blockchain: **b0, b1, ..., bx, ... by, ..., bn**
where **bx** is the block that contains the key exchange command for node **x** and by is the block that contains x's new certificate.
Now, assume we restore node **x** to a block prior to **bx**

before restore:

**x** creates private key **k1** and certificate **c1** and publish **c1** in **by**. Once by is committed, then **x** exchange the old key with **k1** and the old certificate with the certificate that we have in **by**.

After restore:

We retore the blockchain prior to **bx**, so when **x** commits **bx** again, it creates **k2** and **c2** (which are different than **k1, c1**). Then, when it commits by it exchanges the key to **k2** and the certificate to **c1** (as **c1** is the certificate stored in the **by**).

Hence, we have a key mismatch with the certificate.


nkumar04 (approved):
LGTM!!
Thanks!

PR #2172, opened 2021-12-23
Don't execute requests from previous epoch on CRE

yontyon:
In CRE, executing requests from previous epochs may create problems.
For example, consider the scaling procedure.
On scaling the components may get new TLS certificates from the configuration service.
Now, consider the following scenario:
1. TLS exchange for replica X
2. scale up clients
3. the new client gets the certificate of X using CRE, but X's certificate has already changed due to the reconfiguration procedure.

To conclude, we should not run commands from previous epochs in the CRE


nkumar04 (approved):
LGTM!
Thanks!

PR #2176, opened 2022-01-02
 TLS key-exchange improvements (replica-replica and client-replica)

yontyon:
In this PR we introduce the first stage of the TLS key rotation improvements.
Until now, on TLS key exchange, the exchanged component created a key-pair and a self-signed certificate (signed by the new generated private key). Then, the replica published the certificate on the blockchain and all other components had to read it from the blockchain itself.

In this PR we introduce a different method. Instead of generating a self-signed certificate with the new private key, the replica signs the new certificate with its current TLS key. This way, other replicas can validate the certificate against the replica's RSA key, instead of reading it from the blockchain.

This PR is for the replicas' side only (meaning, not for replica-client exchange), but once we complete this change for the replica-client exchange also, we will get rid of the limitation of not exchanging more than f replicas at once.

### DIFF
--- a/bftengine/include/bftengine/KeyExchangeManager.hpp
+++ b/bftengine/include/bftengine/KeyExchangeManager.hpp
@@ -28,7 +28,7 @@ typedef int64_t SeqNum;  // TODO [TK] redefinition
 
 class KeyExchangeManager {
  public:
-  void exchangeTlsKeys(uint64_t bft_sn);
+  void exchangeTlsKeys(const SeqNum& bft_sn);
   // Generates and publish key to consensus
   void sendKeyExchange(const SeqNum&);
   // Generates and publish the first replica's key,
@@ -180,7 +180,7 @@ class KeyExchangeManager {
   std::string generateCid(std::string);
   // build cryptosystem
   void notifyRegistry();
-  void exchangeTlsKeys(const std::string& type, uint64_t bft_sn);
+  void exchangeTlsKeys(const std::string& type, const SeqNum& bft_sn);
   /**
    * Samples periodically how many connections the replica has with other replicas.
    * returns when num of connections is (clusterSize - 1) i.e. full communication.
--- a/bftengine/include/bftengine/KeyExchangeManager.hpp
+++ b/bftengine/include/bftengine/KeyExchangeManager.hpp
@@ -28,7 +28,7 @@ typedef int64_t SeqNum;  // TODO [TK] redefinition
 
 class KeyExchangeManager {
  public:
-  void exchangeTlsKeys(const SeqNum&);
+  void exchangeTlsKeys(uint64_t bftSn);
   // Generates and publish key to consensus
   void sendKeyExchange(const SeqNum&);
   // Generates and publish the first replica's key,
@@ -180,6 +180,7 @@ class KeyExchangeManager {
   std::string generateCid(std::string);
   // build cryptosystem
   void notifyRegistry();
+  void exchangeTlsKeys(const std::string& type, uint64_t bftSn);
   /**
    * Samples periodically how many connections the replica has with other replicas.
    * returns when num of connections is (clusterSize - 1) i.e. full communication.
--- a/bftengine/include/bftengine/KeyExchangeManager.hpp
+++ b/bftengine/include/bftengine/KeyExchangeManager.hpp
@@ -28,7 +28,7 @@ typedef int64_t SeqNum;  // TODO [TK] redefinition
 
 class KeyExchangeManager {
  public:
-  void exchangeTlsKeys(uint64_t bftSn);
+  void exchangeTlsKeys(uint64_t bft_sn);
   // Generates and publish key to consensus
   void sendKeyExchange(const SeqNum&);
   // Generates and publish the first replica's key,
@@ -180,7 +180,7 @@ class KeyExchangeManager {
   std::string generateCid(std::string);
   // build cryptosystem
   void notifyRegistry();
-  void exchangeTlsKeys(const std::string& type, uint64_t bftSn);
+  void exchangeTlsKeys(const std::string& type, uint64_t bft_sn);
   /**
    * Samples periodically how many connections the replica has with other replicas.
    * returns when num of connections is (clusterSize - 1) i.e. full communication.
--- a/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
+++ b/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
@@ -117,6 +117,7 @@ class InternalSigner : public concord::util::crypto::ISigner {
     return out;
   }
   uint32_t signatureLength() const override { return bftEngine::impl::SigManager::instance()->getMySigLength(); }
+  std::string getPrivKey() const override { return ""; }
 };
 
 class ScalingReplicaHandler : public IStateHandler {
--- a/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
+++ b/bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
@@ -101,7 +101,6 @@ class ReplicaTLSKeyExchangeHandler : public IStateHandler {
                              std::to_string(sender_id) + "/client/client.cert";
     sm_.encryptFile(bft_replicas_cert_path, cert);
     LOG_INFO(GL, bft_replicas_cert_path + " is updated on the disk");
-    StateControl::instance().restartComm(sender_id);
     return succ;
   }
 
@@ -197,7 +196,8 @@ std::shared_ptr<ClientReconfigurationEngine> CreFactory::create(std::shared_ptr<
   IStateClient* pbc = new PollBasedStateClient(bftClient, cre_config.interval_timeout_ms_, 0, cre_config.id_);
   auto cre =
       std::make_shared<ClientReconfigurationEngine>(cre_config, pbc, std::make_shared<concordMetrics::Aggregator>());
-  cre->registerHandler(std::make_shared<ReplicaTLSKeyExchangeHandler>());
+  if (bftEngine::ReplicaConfig::instance().isReadOnly)
+    cre->registerHandler(std::make_shared<ReplicaTLSKeyExchangeHandler>());
   if (!bftEngine::ReplicaConfig::instance().isReadOnly) cre->registerHandler(std::make_shared<ScalingReplicaHandler>());
   return cre;
 }
--- a/bftengine/src/bftengine/KeyExchangeManager.cpp
+++ b/bftengine/src/bftengine/KeyExchangeManager.cpp
@@ -136,7 +136,7 @@ void KeyExchangeManager::loadPublicKeys() {
   notifyRegistry();
 }
 
-void KeyExchangeManager::exchangeTlsKeys(const std::string& type, uint64_t bft_sn) {
+void KeyExchangeManager::exchangeTlsKeys(const std::string& type, const SeqNum& bft_sn) {
   auto keys = concord::util::crypto::Crypto::instance().generateECDSAKeyPair(
       concord::util::crypto::KeyFormat::PemFormat, concord::util::crypto::CurveType::secp384r1);
   std::string root_path =
@@ -185,7 +185,7 @@ void KeyExchangeManager::exchangeTlsKeys(const std::string& type, uint64_t bft_s
   client_->sendRequest(RECONFIG_FLAG, strMsg.size(), strMsg.c_str(), cid);
 }
 
-void KeyExchangeManager::exchangeTlsKeys(uint64_t bft_sn) {
+void KeyExchangeManager::exchangeTlsKeys(const SeqNum& bft_sn) {
   exchangeTlsKeys("server", bft_sn);
   exchangeTlsKeys("client", bft_sn);
   metrics_->tls_key_exchange_requests_++;
--- a/bftengine/src/bftengine/KeyExchangeManager.cpp
+++ b/bftengine/src/bftengine/KeyExchangeManager.cpp
@@ -136,7 +136,7 @@ void KeyExchangeManager::loadPublicKeys() {
   notifyRegistry();
 }
 
-void KeyExchangeManager::exchangeTlsKeys(const std::string& type, uint64_t bftSn) {
+void KeyExchangeManager::exchangeTlsKeys(const std::string& type, uint64_t bft_sn) {
   auto keys = concord::util::crypto::Crypto::instance().generateECDSAKeyPair(
       concord::util::crypto::KeyFormat::PemFormat, concord::util::crypto::CurveType::secp384r1);
   std::string root_path =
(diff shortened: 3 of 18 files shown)


bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -135,15 +136,38 @@ void KeyExchangeManager::loadPublicKeys() {
...
+                                 .first;
+  auto cert = concord::util::crypto::CertificateUtils::generateSelfSignedCert(cert_path, keys.second, prev_key_pem);
+  // Now that we have generated new key pair and certificate, lets do the actual exchange on this replica
+  std::string pk_path = root_path + "/pk.pem";
+  std::fstream nec_f(pk_path);
+  std::fstream ec_f(pk_path + ".enc");
+
+  // 1. exchange private key
+  if (nec_f.good()) {
+    secretsMgr_->encryptFile(pk_path, keys.first);
```
robem, line 157: I assume `nec_f` stands for "not encrypted file". Therefore, why do we call `encryptFile`? Shouldn't we just write the key via `writeFile`?

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -135,15 +136,38 @@ void KeyExchangeManager::loadPublicKeys() {
...
+    nec_f.close();
+  }
+  if (ec_f.good()) {
+    secretsMgr_->encryptFile(pk_path + ".enc", keys.first);
+    ec_f.close();
+  }
 
+  concord::secretsmanager::SecretsManagerPlain psm_;
+  // 2. exchange the certificate itself
+  psm_.encryptFile(cert_path, cert);
```
robem, line 167: Same here, we don't encrypt but simply write the cert to the filesystem.

bftengine/include/bftengine/KeyExchangeManager.hpp
```diff
@@ -180,6 +180,7 @@ class KeyExchangeManager {
   std::string generateCid(std::string);
   // build cryptosystem
   void notifyRegistry();
+  void exchangeTlsKeys(const std::string& type, uint64_t bftSn);
```
robem, line 183: An `enum` or `variant` might make more sense than `string` or even `bool` if the type is binary.

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -135,15 +136,38 @@ void KeyExchangeManager::loadPublicKeys() {
   notifyRegistry();
 }
 
-void KeyExchangeManager::exchangeTlsKeys(const SeqNum& bft_sn) {
-  auto repId = bftEngine::ReplicaConfig::instance().replicaId;
+void KeyExchangeManager::exchangeTlsKeys(const std::string& type, uint64_t bftSn) {
```
robem, line 139: `bft_sn`

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -157,9 +181,15 @@ void KeyExchangeManager::exchangeTlsKeys(const SeqNum& bft_sn) {
   data_vec.clear();
   concord::messages::serialize(data_vec, req);
   std::string strMsg(data_vec.begin(), data_vec.end());
-  std::string cid = "replicaTlsKeyExchange_" + std::to_string(bft_sn) + "_" + std::to_string(repId);
+  std::string cid = "replicaTlsKeyExchange_" + std::to_string(bftSn) + "_" + std::to_string(repId);
   client_->sendRequest(RECONFIG_FLAG, strMsg.size(), strMsg.c_str(), cid);
+}
+
+void KeyExchangeManager::exchangeTlsKeys(uint64_t bftSn) {
```
robem, line 188: `bft_sn`

communication/include/communication/StateControl.hpp
```diff
@@ -29,9 +30,15 @@ class StateControl {
...
+  void setGetPeerPubKeyMethod(std::function<std::string(uint32_t)> m) { getPeerPubKey_ = std::move(m); }
+  std::string getPeerPubKey(uint32_t id) {
+    if (getPeerPubKey_) return getPeerPubKey_(id);
+    return std::string();
+  }
 
  private:
   std::mutex lock_comm_;
   concord::util::CallbackRegistry<uint32_t> comm_restart_cb_registry_;
+  std::function<std::string(uint32_t)> getPeerPubKey_;
```
robem, line 42: Please follow the coding standard, snake_case variables.

communication/src/AsyncTlsConnection.cpp
```diff
@@ -384,103 +385,60 @@ bool AsyncTlsConnection::verifyCertificateServer(asio::ssl::verify_context& ctx)
...
-  auto [valid, peer_id] = checkCertificate(cert, "client", std::string(subject), std::nullopt);
+  auto [valid, peer_id] = checkCertificate(cert, std::nullopt);
   peer_id_ = peer_id;
   return valid;
 }
 
 std::pair<bool, NodeNum> AsyncTlsConnection::checkCertificate(X509* receivedCert,
-                                                              std::string connectionType,
-                                                              const std::string& subject,
                                                               std::optional<NodeNum> expectedPeerId) {
```
robem, line 399: Given that you touch this function, would you mind changing the variables to snake_case as well?

communication/src/AsyncTlsConnection.cpp
```diff
@@ -384,103 +385,60 @@ bool AsyncTlsConnection::verifyCertificateServer(asio::ssl::verify_context& ctx)
...
-  if (expectedPeerId) {
-    if (remotePeerId != expectedPeerId) {
-      LOG_WARN(logger_, "Peers don't match, expected: " << expectedPeerId.value() << ", received: " << remPeer);
-      return std::make_pair(false, remotePeerId);
+  uint32_t peerId = UINT32_MAX;
+  std::string conn_type;
+  bool res = concord::util::crypto::CertificateUtils::verifyCertificate(
+      receivedCert, config_.certificatesRootPath, peerId, conn_type);
+  if (expectedPeerId.has_value() && peerId != expectedPeerId.value()) return std::make_pair(false, peerId);
+  if (!res) {
```
robem, line 405: Suggestion: We can return right here and therefore skip the indentation.
```
if (res)  return std::make_pair(res, peer_id);
```

communication/src/AsyncTlsConnection.cpp
```diff
@@ -384,103 +385,60 @@ bool AsyncTlsConnection::verifyCertificateServer(asio::ssl::verify_context& ctx)
...
+      int certLen = BIO_pending(outbio);
+      certStr.resize(certLen);
+      BIO_read(outbio, (void*)&(certStr.front()), certLen);
+      std::ofstream out(local_cert_path.data());
+      out << certStr;
+      out.close();
+      BIO_free(outbio);
+      LOG_INFO(logger_, "new certificate has been updated on local storage, peer: " << peerId);
+    } else {
+      std::make_pair(false, peerId);
```
robem, line 438: return?

communication/src/AsyncTlsConnection.cpp
```diff
@@ -384,103 +385,60 @@ bool AsyncTlsConnection::verifyCertificateServer(asio::ssl::verify_context& ctx)
...
+    if (!pem_pub_key.empty()) {
+      if (concord::util::crypto::Crypto::instance().getFormat(pem_pub_key) !=
+          concord::util::crypto::KeyFormat::PemFormat) {
+        pem_pub_key = concord::util::crypto::Crypto::instance()
+                          .RsaHexToPem(std::make_pair("", StateControl::instance().getPeerPubKey(peerId)))
+                          .second;
+      }
+      res = concord::util::crypto::CertificateUtils::verifyCertificate(receivedCert, pem_pub_key);
+    }
+    if (res) {
```
robem, line 419: Same suggestion, if you negate the condition then you can return right here and save one level of indentation. Overall, this will make it more readable.

kvbc/include/reconfiguration_kvbc_handler.hpp
```diff
@@ -209,6 +204,11 @@ class InternalKvReconfigurationHandler : public concord::reconfiguration::IRecon
               uint32_t,
               const std::optional<bftEngine::Timestamp>&,
               concord::messages::ReconfigurationResponse&) override;
+  bool handle(const concord::messages::ReplicaTlsExchangeKey&,
+              uint64_t,
+              uint32_t,
```
robem, line 209: The whole file is skipping the second and third parameter but the semantics are unclear (at least from just looking at this function signature). Would you mind adding variable names?

kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -355,39 +319,25 @@ bool ReconfigurationHandler::handle(const concord::messages::ClientKeyExchangeSt
...
+          LOG_INFO(getLogger(), "found transactions public key exchange status for client" << KVLOG(cid));
+          stats.clients_data.push_back(std::make_pair(cid, cmd.pub_key));
         }
       }
     }
   }
   rres.response = stats;
   return true;
-}
+}  // namespace concord::kvbc::reconfiguration
```
robem, line 340: Isn't this the end of a function? Not a namespace?

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -135,15 +136,38 @@ void KeyExchangeManager::loadPublicKeys() {
...
+  if (nec_f.good()) {
+    secretsMgr_->encryptFile(pk_path, keys.first);
+    nec_f.close();
+  }
+  if (ec_f.good()) {
+    secretsMgr_->encryptFile(pk_path + ".enc", keys.first);
+    ec_f.close();
+  }
 
+  concord::secretsmanager::SecretsManagerPlain psm_;
```
robem, line 165: Just `psm` no underscore.

robem (changes requested):

bftengine/src/bftengine/KeyExchangeManager.cpp
yontyon: the secertMgr_ might be either plain or encrypted. The reason why we need to check the suffix of the file is that currently encrypted files have the .enc suffix (this should be fixed, but not as part of this PR)

bftengine/src/bftengine/KeyExchangeManager.cpp
yontyon: the plain secret manager here is used as a file handler (it wraps the read/write operations with error handling and so on). We use it in order to have less code duplications

bftengine/include/bftengine/KeyExchangeManager.hpp
yontyon: The type (either server or client) is part of the final file path. This is why we move it here directly through a string.
This way, in the future, it will be easier to change the file path directly from the caller function

bftengine/src/bftengine/KeyExchangeManager.cpp
yontyon: fixed

bftengine/src/bftengine/KeyExchangeManager.cpp
yontyon: fixed

communication/include/communication/StateControl.hpp
yontyon: fixed

communication/src/AsyncTlsConnection.cpp
yontyon: fixed

communication/src/AsyncTlsConnection.cpp
yontyon: fixed

communication/src/AsyncTlsConnection.cpp
yontyon: fixed

communication/src/AsyncTlsConnection.cpp
yontyon: fixed

kvbc/include/reconfiguration_kvbc_handler.hpp
yontyon: This should be fixed, but not as part of this PR

kvbc/src/reconfiguration_kvbc_handler.cpp
yontyon: fixed

bftengine/src/bftengine/KeyExchangeManager.cpp
yontyon: fixed

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
```diff
@@ -197,7 +196,8 @@ std::shared_ptr<ClientReconfigurationEngine> CreFactory::create(std::shared_ptr<
   IStateClient* pbc = new PollBasedStateClient(bftClient, cre_config.interval_timeout_ms_, 0, cre_config.id_);
   auto cre =
       std::make_shared<ClientReconfigurationEngine>(cre_config, pbc, std::make_shared<concordMetrics::Aggregator>());
-  cre->registerHandler(std::make_shared<ReplicaTLSKeyExchangeHandler>());
+  if (bftEngine::ReplicaConfig::instance().isReadOnly)
+    cre->registerHandler(std::make_shared<ReplicaTLSKeyExchangeHandler>());
   if (!bftEngine::ReplicaConfig::instance().isReadOnly) cre->registerHandler(std::make_shared<ScalingReplicaHandler>());
```
toly-kournik, line 201: else?

toly-kournik (approved):
LGTM with minor comments

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
yontyon: we don't need the replicaTLSKeyExchangeHandler in regular replicas.

bftengine/include/bftengine/KeyExchangeManager.hpp
```diff
@@ -28,7 +28,7 @@ typedef int64_t SeqNum;  // TODO [TK] redefinition
 
 class KeyExchangeManager {
  public:
-  void exchangeTlsKeys(const SeqNum&);
+  void exchangeTlsKeys(uint64_t bftSn);
```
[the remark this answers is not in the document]
yontyon: fixed

util/src/crypto_utils.cpp
```diff
@@ -337,13 +265,171 @@ std::pair<std::string, std::string> Crypto::ECDSAHexToPem(const std::pair<std::s
...
+bool CertificateUtils::verifyCertificate(X509* cert_to_verify,
+                                         const std::string& cert_root_directory,
+                                         uint32_t& remote_peer_id,
+                                         std::string& conn_type) {
+  // First get the source ID
+  static constexpr size_t SIZE = 512;
+  std::string subject(SIZE, 0);
+  X509_NAME_oneline(X509_get_subject_name(cert_to_verify), subject.data(), SIZE);
+
+  int peerIdPrefixLength = 3;
```
robem, line 285: `snake_case` - please check and apply to the rest of the PR as well.

util/include/crypto_utils.hpp
```diff
@@ -15,68 +15,97 @@
...
+#include <openssl/ec.h>
+#include <openssl/pem.h>
+#include <openssl/x509.h>
+#include <openssl/evp.h>
+
 namespace concord::util::crypto {
 enum class KeyFormat : std::uint16_t { HexaDecimalStrippedFormat, PemFormat };
 enum class CurveType : std::uint16_t { secp256k1, secp384r1 };
+
+class CertificateUtils {
```
robem, line 29: Do we need this class? Couldn't the functions simply sit in the `crypto` namespace? Or a new sub-namespace instead of the class - `concord::util::crypto::certs` or similar?

bftengine/include/bftengine/KeyExchangeManager.hpp
robem: I don't think it will be easier. If you want to change it then you now have to change it in multiple places. Everywhere, where `exchangeTlsKeys` is invoked and then in the function itself which checks the type.
Instead, if you care about changing the path only then you could construct the type string inside the function instead of relying on the caller.
Meaning, I suggest to introduce an enum that specifies the type and within `exchangeTlsKeys` you construct the path based on that enum. If the path ever changes then you have to change it inside `exchangeTlsKeys` only.

robem (approved):

PR #2178, opened 2022-01-03
Adding the EDBKeyType for Migration

arc-vmware:
This will handle the case of EDBKeyType for the block merkel lastest
key version column family.


nkumar04 (approved):
LGTM!

PR #2179, opened 2022-01-03
Adding exception in place of assert.

agrkushvm:


bftengine/src/bftengine/SigManager.cpp
```diff
@@ -213,7 +213,12 @@ bool SigManager::verifySig(
     idOfReplica = replicasInfo_.isIdOfReplica(pid);
   }
   idOfReadOnlyReplica = replicasInfo_.isIdOfPeerRoReplica(pid);
-  ConcordAssert(idOfReplica || idOfExternalClient || idOfReadOnlyReplica);
```
robem, line 216: Asserts are usually in place to check programming assumptions. Why are you converting this to an exception? How is the exception handled at runtime?
A proper commit message would be helpful. What problem are you solving, why and how?

agrkushvm: @robem 
As discussed with @glevkovich, currently asserts are placed when we internally check the parameters.
But here pid is coming as a parameter in verifySig API
bool SigManager::verifySig(
    PrincipalId pid, const char* data, size_t dataLength, const char* sig, uint16_t sigLength)
 Therefore we have replaced assert with exception

glevkovich: @robem , the input is comming from the network. Any byzantine replica may send bad data on message header and crash us. For the specific pid, I don't think it can ever happen since it is validated with tls ceritificate on the source on communication level. But we don't want to replay on external module (communication) - and what if we run tCP/udp?
So asserting here is not safe. replacing to exception.

glevkovich:
@agrkushvm , I think it is not a good idea you are throwing a general exception here when you know you want to catch it on the caller.
catching this way:
`catch (const std::exception& e)`
will catch any exception thrown in this function, and continue, while this is not always the way we want to behave.
Let's assume we have some other bug in 3rd party or library code (cryptoPP, std) that throws an exception, here you are catching any exception and continues.

Suggestion: why not force the caller to get the returned value and return false (not throwing an exception) ?
see:
https://en.cppreference.com/w/cpp/language/attributes/nodiscard
use nodiscard option (started on c++17).

I would change the function to return an error and use nodiscard to force the caller to check returned value.
I will continue reviewing the code after we discuss this 1st comment.

agrkushvm:
> @agrkushvm , I think it is not a good idea you are throwing a general exception here when you know you want to catch it on the caller. catching this way: `catch (const std::exception& e)` will catch any exception thrown in this function, and continue, while this is not always the way we want to behave. Let's assume we have some other bug in 3rd party or library code (cryptoPP, std) that throws an exception, here you are catching any exception and continues.
> 
> Suggestion: why not force the caller to get the returned value and return false (not throwing an exception) ? see: https://en.cppreference.com/w/cpp/language/attributes/nodiscard use nodiscard option (started on c++17).
> 
> I would change the function to return an error and use nodiscard to force the caller to check returned value. I will continue reviewing the code after we discuss this 1st comment.

@glevkovich Comments addressed.

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -215,7 +215,12 @@ bool SigManager::verifySig(
...
+  if (!(idOfReplica || idOfExternalClient || idOfReadOnlyReplica)) {
+    std::stringstream msg;
+    msg << "Invalid ID of replicas : " << KVLOG(idOfReplica, idOfExternalClient, idOfReadOnlyReplica);
+    LOG_ERROR(GL, msg.str());
```
glevkovich, line 221: why do we need the std::stringstream?
`LOG_ERROR(GL, "Invalid ID of replicas:" << KVLOG(idOfReplica, idOfExternalClient, idOfReadOnlyReplica));`
please pay attention: KVLOG already adds a space after the ':'.

glevkovich (approved):
@agrkushvm  Thank You, approved.
Just one small comment.

PR #2180, opened 2022-01-03
Make deferred messages queues configurable

Liorgal28:
This PR inserts 2 main changes:
1) Make the deferred messages queue size configurable.
2) Decrease the num of TLS threads to 1 in order to avoid thread race.
3) Fix Asan heap-use-after-free in post-execution handle of the deferred message.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1047,7 +1047,6 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
     LOG_INFO(GL,
              "Ignoring PrePrepareMsg because system is stopped at checkpoint pending control state operation (upgrade, "
              "etc...)");
-    // TODO(LG): we should delete the message.
     return;
   }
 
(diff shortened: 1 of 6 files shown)


yontyon (approved):

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1043,6 +1045,7 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
     LOG_INFO(GL,
              "Ignoring PrePrepareMsg because system is stopped at checkpoint pending control state operation (upgrade, "
              "etc...)");
+    // TODO(LG): we should delete the message.
```
robem, line 1048: Then why don't we do it?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1078,9 +1081,14 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
 
     return;  // TODO(GG): memory deallocation is confusing .....
   }
+  if (isCurrentPrimary() && (msg->senderId() != getReplicaConfig().replicaId)) {
+    LOG_INFO(GL, "Ignoring PrePrepareMsg since im the current primary");
```
robem, line 1085: "... since this is the current primary" or "... since I'm the current primary"

robem (commented):
Where are the commit messages? What problem are you solving and how?

communication/src/TlsTCPCommunication.cpp
```diff
@@ -14,7 +14,7 @@
 #include "TlsRunner.h"
 
 // TODO: Make this configurable
-static constexpr size_t NUM_THREADS = 2;
+static constexpr size_t NUM_THREADS = 1;
```
robem, line 17: Why is this constant being changed?

bftengine/src/bftengine/ReplicaImp.cpp
Liorgal28: You're right it shouldn't be in this PR - removed.

PR #2181, opened 2022-01-03
Attach epoch number per reconfiguration action key

yontyon:
Until now, we overwrote the epoch number on every new reconfiguration request.
However, in the case of pruning, we may remove the old epoch number and hence lose a command's epoch number.
For example:
1. execute reconfiguration request of replica's scaling the command is written to block X and we have an epoch record on block X
2. execute another request of key exchange and write it to block X + 1 ==> we have a new epoch record on block X+1
3. execute pruning --> the epoch record on block X is stale and hence removed

--> We lost the epoch number of the first scaling command!
This PR is for solving this issue (in v0.13.x). It is solved by writing an epoch record attached to each command key. This way, only if we execute the same command again (and overwrite the first command block) we also overwrite this specific command epoch record.


nkumar04 (approved):
LGTM!
Thanks!

PR #2182, opened 2022-01-03
Move OperationResult to a separate include file.

yuliasherman:
As OperationResult is used by both - client and replica sides, separate it to a new include file.

### DIFF
--- a/bftengine/include/bftengine/SharedTypes.hpp
+++ b/bftengine/include/bftengine/SharedTypes.hpp
@@ -1,0 +1,30 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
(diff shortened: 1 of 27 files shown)


cloudnoize (approved):
Great! thanks

bftengine/include/bftengine/SharedTypes.hpp
```diff
@@ -0,0 +1,30 @@
...
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+namespace shared {
```
robem, line 14: This is a top-level namespace. This should be either `concord::shared` or `bftengine::shared`. Not just `shared`. Namespaces provide a way to avoid common name collision and should start with a project name.

yuliasherman: I thought about it. It needs to be common between bftengine and bftclient, so I chose 'shared'.

robem: But `shared` doesn't mean anything. Shared between whom? The namespace should at least narrow down the domain.

yuliasherman: I can leave it bftEngine, like before.

PR #2184, opened 2022-01-04
Make the number of objects dynamic in the metadata storage

yontyon:
Now that we save RSI in the persistent storage, we may end up with the default number of object descriptors not matching the actual number of objects. In case we have x bft clients and batch size of y we need at least x*y objects in the persistent storage objects map.
This number may exceed the hardcoded 10000 number.
In this PR we introduced a dynamic way to initiate and determine the number of objects depending on the numbers x and y

### DIFF
--- a/bftengine/src/bftengine/DbMetadataStorage.cpp
+++ b/bftengine/src/bftengine/DbMetadataStorage.cpp
@@ -46,12 +46,14 @@ bool DBMetadataStorage::isNewStorage() {
   return (outActualObjectSize == 0);
 }
 
-bool DBMetadataStorage::initMaxSizeOfObjects(ObjectDesc *metadataObjectsArray, uint32_t metadataObjectsArrayLength) {
+bool DBMetadataStorage::initMaxSizeOfObjects(const std::map<uint32_t, ObjectDesc> &metadataObjectsArray,
+                                             uint32_t metadataObjectsArrayLength) {
   for (uint32_t i = objectsNumParameterId_ + 1; i < metadataObjectsArrayLength; ++i) {
-    objectIdToSizeMap_[i] = metadataObjectsArray[i].maxSize;
-    LOG_TRACE(logger_,
-              "initMaxSizeOfObjects i=" << i << " object data: id=" << metadataObjectsArray[i].id
-                                        << ", maxSize=" << metadataObjectsArray[i].maxSize);
+    const auto objectData = metadataObjectsArray.at(i);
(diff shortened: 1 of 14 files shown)


bftengine/src/bftengine/DbMetadataStorage.cpp
```diff
@@ -46,12 +46,14 @@ bool DBMetadataStorage::isNewStorage() {
...
-bool DBMetadataStorage::initMaxSizeOfObjects(ObjectDesc *metadataObjectsArray, uint32_t metadataObjectsArrayLength) {
+bool DBMetadataStorage::initMaxSizeOfObjects(const std::map<uint32_t, ObjectDesc> &metadataObjectsArray,
+                                             uint32_t metadataObjectsArrayLength) {
   for (uint32_t i = objectsNumParameterId_ + 1; i < metadataObjectsArrayLength; ++i) {
-    objectIdToSizeMap_[i] = metadataObjectsArray[i].maxSize;
-    LOG_TRACE(logger_,
-              "initMaxSizeOfObjects i=" << i << " object data: id=" << metadataObjectsArray[i].id
-                                        << ", maxSize=" << metadataObjectsArray[i].maxSize);
+    const auto objectData = metadataObjectsArray.at(i);
+    objectIdToSizeMap_[i] = objectData.maxSize;
```
arc-vmware, line 53: We can change this to 
objectIdToSizeMap_[objectData.id] instead of i.

bftengine/tests/simpleStorage/FileStorage.cpp
```diff
@@ -111,7 +111,8 @@ void FileStorage::write(
 
 bool FileStorage::isNewStorage() { return (objectsMetadata_ == nullptr); }
 
-bool FileStorage::initMaxSizeOfObjects(ObjectDesc *metadataObjectsArray, uint32_t metadataObjectsArrayLength) {
+bool FileStorage::initMaxSizeOfObjects(const std::map<uint32_t, ObjectDesc> &metadataObjectsArray,
```
arc-vmware, line 114: As it's a map now, can we remove metadataObjectsArrayLength

arc-vmware (approved):
LGTM, this is done very well, brilliant.

Thanks @yontyon 
I have few nice to have suggestions .

bftengine/tests/simpleStorage/FileStorage.cpp
yontyon: agreed, but I would like to change the existing API as less as possible

PR #2185, opened 2022-01-04
ResourceManager skeleton

titusen:
ResourceMonitor initial version with necessary interfaces.


performance/include/IResourceComputingStrategy.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
robem, line 1: Copyrights missing in all files.

performance/include/IResourceComputingStrategy.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+namespace concord::performance {
+
+class IResourceComputingStrategy {
+ public:
+  virtual std::uint64_t computeAvailableResources(const ISystemResourceEntity &consensusMonitor,
+                                                  const ISystemResourceEntity &storageMonitor) = 0;
+};
+
+}  // namespace concord::performance
```
robem, line 14: Newline missing - Github shows it with this symbol.

titusen: Done

performance/include/IResourceComputingStrategy.hpp
titusen: Done

teoparvanov (commented):
Hi @titusen, can you please provide a more descriptive title? Initial version of what, for instance 😄 also, I see there is some unit test, so this isn't very accurate either. 

Thank you,
Teo

performance/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
-if(BUILD_SLOWDOWN)
-    add_subdirectory(test)
-endif()
+add_subdirectory(test)
+
+add_library(concord_resources_monitor INTERFACE)
+
+target_include_directories(concord_resources_monitor INTERFACE
+  include
+)
```
dartdart26, line 7: I think the intention for the `performance` directory was to have the performance testing framework inside. Not sure if it is the right place for the resources monitor. Maybe we can have it in a separate directory? What do you think?

performance/include/IResourceComputingStrategy.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <cstdint>
+#include "ISystemResourceEntity.hpp"
+
+namespace concord::performance {
+
+class IResourceComputingStrategy {
+ public:
+  virtual ~IResourceComputingStrategy() = default;
+  virtual std::uint64_t computeAvailableResources(const ISystemResourceEntity &consensusMonitor,
+                                                  const ISystemResourceEntity &storageMonitor) = 0;
```
dartdart26, line 24: Could we please define what the return value of this method signifies? It would be really helpful to its users.

performance/include/IResourceManager.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#include <cstdint>
+
+namespace concord::performance {
+class IResourceManager {
+ public:
+  virtual ~IResourceManager() = default;
+  virtual uint64_t getAvailableResources() const = 0;
```
dartdart26, line 21: Same as above - let's please define what the return value means.

performance/include/ISystemResourceEntity.hpp
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include <cstdint>
+#include <string>
+
+namespace concord::performance {
+class ISystemResourceEntity {
+ public:
+  virtual int64_t getAvailableResources() const = 0;
+  virtual uint64_t getMeasurements() const = 0;
```
dartdart26, line 22: Same as above - let's please define what the return values of these methods mean.

performance/include/IntervalMappingComputingStrategy.hpp
```diff
@@ -0,0 +1,40 @@
...
+      : intervalMapping(std::move(intervalMappingSorted)) {
+    intervalMapping.push_back(std::make_pair(UINT64_MAX, 0));
+  };
+  virtual ~IntervalMappingComputingStrategy() = default;
+  virtual std::uint64_t computeAvailableResources(const ISystemResourceEntity &consensusMonitor,
+                                                  const ISystemResourceEntity &storageMonitor) override {
+    return std::upper_bound(intervalMapping.begin(),
+                            intervalMapping.end(),
+                            std::make_pair((u_int64_t)consensusMonitor.getAvailableResources(), (u_int64_t)0))
+        ->second;
```
dartdart26, line 34: Nit: I know this is a skeleton only, but it would be helpful for the future implementation to explain what the algorithm is doing to determine the available resources.

performance/include/IntervalMappingComputingStrategy.hpp
```diff
@@ -0,0 +1,40 @@
...
+  };
+
+ private:
+  std::vector<std::pair<uint64_t, uint64_t>> &&intervalMapping;
```
dartdart26, line 38: If I am not missing something and judging by the use in the constructor, I think the member should be just `intervalMapping` and not a reference.

performance/include/IResourceComputingStrategy.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <cstdint>
+#include "ISystemResourceEntity.hpp"
+
+namespace concord::performance {
+
+class IResourceComputingStrategy {
+ public:
+  virtual ~IResourceComputingStrategy() = default;
+  virtual std::uint64_t computeAvailableResources(const ISystemResourceEntity &consensusMonitor,
```
cloudnoize, line 23: As a generic interface, it seems better to pass something like a vector of ISystemResourceEntity instead of explicit resources as it is more flexible for future extensions.

cloudnoize (commented):
I'm missing comments that explain the functionality of each class/interface, their interaction, and the functionality as a whole.

performance/include/IResourceComputingStrategy.hpp
titusen: Some computing strategies may distinguish the source of measurement e.g. IntervalMappingComputingStrategy which takes into account only the ConsensusMonitor Metric. I came to the conslusion that it will take no arguments. Each particular implementation will take all SystemResourceEntities it needs to compute available resources.

titusen: I've removed a ComputingStrategy interface and now the implementation of an resource manager is taking care of computations

performance/include/IntervalMappingComputingStrategy.hpp
titusen: Yep, corrected.

resources-manager/include/IntervalMappingResourceManager.hpp
```diff
@@ -0,0 +1,53 @@
...
+    return std::upper_bound(intervalMapping.begin(),
+                            intervalMapping.end(),
+                            std::make_pair((uint64_t)consensusEngine->getMeasurements(), (u_int64_t)0))
+        ->second;
+  }
+  static IntervalMappingResourceManager *createIntervalMappingResourceManager(
+      const std::shared_ptr<ISystemResourceEntity> &consensusEngine,
+      std::vector<std::pair<uint64_t, uint64_t>> &&intervalMapping) {
+    intervalMapping.push_back(std::make_pair(UINT64_MAX, 0));
+    return new IntervalMappingResourceManager(consensusEngine, std::move(intervalMapping));
```
cloudnoize, line 41: Is it a Singleton? who calls this and who is the owner of the pointer?

resources-manager/include/SumsResourceEntitiesAvailabilityManager.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include "IResourceManager.hpp"
+#include "ISystemResourceEntity.hpp"
+
+#include <vector>
+#include <memory>
+#include <algorithm>
+
+namespace concord::performance {
+
+class SumsResourceEntitiesAvailabilityManager : public IResourceManager {
```
cloudnoize, line 24: Do you have a diagram that shows the relationship between those classes and how they interact in order to calculate a result?
For example, I don't understand SumsResourceEntitiesAvailabilityManager and SubstractFromMaxResourceManager and where and how they are being used.

resources-manager/include/IntervalMappingResourceManager.hpp
titusen: It's a standard factory method. Caller is responsible for memory. AFAIK the common convention is that anytime method returns a pointer it is a caller, who is responsible for assign it to proper smart pointer. Is that assumption correct?

resources-manager/include/SumsResourceEntitiesAvailabilityManager.hpp
titusen: So far the only usage is in the unittests. The responsiblity of computations is on the implementation of the IResourceManager interface. The SumsResourceEntitiesAvailabilityManager sums up the getAvailableResources answers.
The SubstractFromMaxResourceManager  substracts from configrued maximum capacity the sum of getMeasurements.

titusen:
A graph explanation of ResourceManager implementations.
[Uploading ResourceManagerScenarios.pdf…]()

resources-manager/include/IResourceManager.hpp
```diff
@@ -0,0 +1,27 @@
+// Concord
+//
+// Copyright (c) 2018-2021 VMware, Inc. All Rights Reserved.
```
dartdart26, line 3: Nit: 2022.

resources-manager/include/IResourceManager.hpp
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include <cstdint>
+
+namespace concord::performance {
+class IResourceManager {
+ public:
+  virtual ~IResourceManager() = default;
+  /*
+    getAvailableResources indicates what capacity is left for pruning. Return value is a
```
dartdart26, line 22: Could you please describe what the return value is?

resources-manager/include/ISystemResourceEntity.hpp
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include <cstdint>
+#include <string>
+
+namespace concord::performance {
+class ISystemResourceEntity {
+ public:
+  virtual ~ISystemResourceEntity() = default;
+  // Implementation returns how much more pruned blocks can this entity support.
```
dartdart26, line 22: Is that comment relevant for the interface? Does this interface only relate to pruning?

resources-manager/include/ISystemResourceEntity.hpp
```diff
@@ -0,0 +1,29 @@
...
+  virtual int64_t getAvailableResources() const = 0;
+  // Measurements of handled operations. Can be transaction, prune blocks, etc.
```
dartdart26, line 24: I think returning an opaque integer type is hard to reason about for the caller. Can we maybe define it it as "count of operations per some period" or similar?

resources-manager/include/ISystemResourceEntity.hpp
```diff
@@ -0,0 +1,29 @@
...
+  virtual uint64_t getMeasurements() const = 0;
+  // Name of the resource for metrics purposes.
+  virtual const std::string& getResourceName() const = 0;
```
dartdart26, line 27: Nit: I think returning a reference in an interfaces forces all implementations to have a member resource name string. I would suggest we either:

- change return type to std::string
- make a public const resource name std::string available in the interface

resources-manager/include/IResourceManager.hpp
dartdart26: Oh, I got it now. The comment `getAvailableResources` is wrong in the comment, not the actual method name.

resources-manager/include/ISystemResourceEntity.hpp
titusen: I thought that negative value could be used when a counter/metric exceeds it's capacity.

resources-manager/include/IResourceManager.hpp
titusen: That's really strange, cause on my vm, it's corrected and the code committed and pushed.

resources-manager/include/ISystemResourceEntity.hpp
titusen: Will correct this to uint64.

resources-manager/include/ISystemResourceEntity.hpp
titusen: done

resources-manager/include/ISystemResourceEntity.hpp
titusen: Done

resources-manager/include/ISystemResourceEntity.hpp
dartdart26: Could you please explain that in a comment? Or maybe define another type that is even more explicit than an integer? It would make the interface easier to use.

resources-manager/include/SumsResourceEntitiesAvailabilityManager.hpp
titusen: More explanations are in attached pdf file.

titusen: [ResourceManagerScenarios.pdf](https://github.com/vmware/concord-bft/files/7884184/ResourceManagerScenarios.pdf)

cloudnoize (approved):
LGTM thanks for the effort!

PR #2187, opened 2022-01-04
Return OperationResult back under bftEngine namespace

yuliasherman:
Return OperationResult back under bftEngine namespace.


robem (approved):
👍  Much better. Thanks!

PR #2193, opened 2022-01-05
TRS: Fix clean-up code

robem:
In case we catch an exception while we wait for updates on the live-update
queue, we need to clean-up properly (unregister the subscription buffer).
Worst case scenario: This happens constantly and therefore, we eventually run
out of memory due to we keep subscription buffer queues with 100 elements
around forever.

### DIFF
--- a/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
+++ b/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
@@ -365,19 +365,20 @@ class ThinReplicaImpl {
             update_aggregator_counter = 0;
           }
         }
-        if (!is_event_group_transition) {
-          config_->subscriber_list.removeBuffer(live_updates);
-          live_updates->removeAllUpdates();
-          metrics_.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
-          metrics_.updateAggregator();
-        }
       } catch (std::exception& error) {
         LOG_INFO(logger_, "Subscription stream closed: " << error.what());
       }
-      if (context->IsCancelled()) {
-        return grpc::Status::CANCELLED;
-      }
-      if (not is_event_group_transition) {
+
+      // Clean up if we need to return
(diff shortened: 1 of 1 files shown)


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -365,19 +365,18 @@ class ThinReplicaImpl {
...
-      }
-      if (not is_event_group_transition) {
+      // Clean up if we need to return
+      if (not is_event_group_transition || context->IsCancelled()) {
+        config_->subscriber_list.removeBuffer(live_updates);
+        live_updates->removeAllUpdates();
+        metrics_.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
+        metrics_.updateAggregator();
+        if (context->IsCancelled()) {
+          return grpc::Status::CANCELLED;
```
shgandhi, line 378: Nit: I think a log message `LOG_WARN(logger_, "Subscription cancelled");` like we have in `TryPopEventGroup()` can be helpful for debugging.

shgandhi (approved):
Thanks for the fix @robem! LGTM.
I think we also need to put calls to `Pop()` and `PopEventGroup()` in try-catch block, and add a clean up code in the catch block. WDYT?

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
robem: This gets triggered more often than not if there are no updates. Having a warning message would actually spam the logs. We can do a debug message but even then it would be pure spam :/

robem:
> Thanks for the fix @robem! LGTM. I think we also need to put calls to `Pop()` and `PopEventGroup()` in try-catch block, and add a clean up code in the catch block. WDYT?

They're in try-catch blocks already just one level higher.

shgandhi (approved):

PR #2195, opened 2022-01-06
Add specific message type for latest known epoch

yontyon:
Apparently, with an empty response, the deserialization set the first class in the variant. Hence we are unable to return a state reply with an empty response. In this PR we add a specific type to the epoch response


nkumar04 (approved):
LGTM!
Thanks!

Liorgal28 (approved):

PR #2196, opened 2022-01-06
Infra for replicas main public key exchange

yontyon:
In this PR we introduce the infrastructure for publishing the replicas' main key (currently RSA).
This is done by publishing a special block on the blockchain which contains the replica's public key.

This feature is working independently on the initial key exchange feature


nkumar04 (approved):
Great PR!
LGTM!

PR #2203, opened 2022-01-09
Collect preprocess signatures in a set instead of list

cloudnoize:
In order to eliminate duplicate signatures from the same sender ID in the preprocess result.


yuliasherman (approved):
LGTM

PR #2204, opened 2022-01-09
Return reply data from the client pool to the client service when operation succeeds

yuliasherman:
The client pool needs to call 'callback_(result_data)' in case of request succeeds and 'callback_(operation_result)' when it fails.


situ-s (approved):
LGTM! Thanks.

PR #2205, opened 2022-01-10
Bug fix. Store future in class scope and some code cleanup

nkumar04:
std::async call returns a future object. if this is in local scope, then thread blocks on destructor of the future, which essentially is equivalent to sync operation. This PR addresses this issue.
Also, CreateDbCheckpointAsync api now returns the lastBlockId to the caller.

### DIFF
--- a/bftengine/include/bftengine/DbCheckpointManager.hpp
+++ b/bftengine/include/bftengine/DbCheckpointManager.hpp
@@ -32,15 +32,8 @@
 #include "Metrics.hpp"
 #include "InternalBFTClient.hpp"
 #include "storage/db_interface.h"
-#if __has_include(<filesystem>)
-#include <filesystem>
-namespace fs = std::filesystem;
-#elif __has_include(<experimental/filesystem>)
-#include <experimental/filesystem>
-namespace _fs = std::experimental::filesystem;
-#else
-#error "Missing filesystem support"
-#endif
+#include <boost/filesystem.hpp>
+namespace _fs = boost::filesystem;
 namespace bftEngine::impl {
 using std::chrono::duration_cast;
 using Status = concordUtils::Status;
@@ -49,8 +42,7 @@ using Seconds = std::chrono::seconds;
 using InternalBftClient = bftEngine::impl::InternalBFTClient;
 class DbCheckpointManager {
  public:
-  // void createDbCheckpoint(const SeqNum& seqNum);
-  void createDbCheckpointAsync(const SeqNum& seqNum, const std::optional<Timestamp>& timestamp);
+  uint64_t createDbCheckpointAsync(const SeqNum& seqNum, const std::optional<Timestamp>& timestamp);
   Seconds getLastCheckpointCreationTime() const { return lastCheckpointCreationTime_; }
   void initializeDbCheckpointManager(std::shared_ptr<concord::storage::IDBClient> dbClient,
                                      std::shared_ptr<bftEngine::impl::PersistentStorage> p,
@@ -112,9 +104,13 @@ class DbCheckpointManager {
   void checkAndRemove();
   void removeOldestDbCheckpoint();
   void updateDbCheckpointMetadata();
+  void removeDbCheckpointFuture(uint64_t);
+  void updateMetrics();
   InternalBftClient* client_{nullptr};
   std::atomic<bool> stopped_ = false;
   DbCheckpointMetadata dbCheckptMetadata_;
+  std::map<uint64_t, std::future<void> > dbCreateCheckPtFuture_;
+  std::future<void> cleanUpFuture_;
   std::shared_ptr<concord::storage::IDBClient> dbClient_;
   std::shared_ptr<bftEngine::impl::PersistentStorage> ps_;
   // this thread minitors if we over use the disk space
--- a/bftengine/src/bftengine/DbCheckpointManager.cpp
+++ b/bftengine/src/bftengine/DbCheckpointManager.cpp
@@ -86,22 +86,7 @@ Status DbCheckpointManager::createDbCheckpoint(const CheckpointId& checkPointId,
       removeOldestDbCheckpoint();
     }
   }
-
-  // update metrics
-  auto ret = std::async(std::launch::async, [this]() {
-    const auto& checkpointDir = dbClient_->getCheckpointPath();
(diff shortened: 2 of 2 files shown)


bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -199,15 +184,17 @@ void DbCheckpointManager::initializeDbCheckpointManager(std::shared_ptr<concord:
...
-  auto ret = std::async(std::launch::async, [this, seqNum, timestamp]() -> void {
-    const auto& lastBlockid = getLastBlockIdCb_();
+uint64_t DbCheckpointManager::createDbCheckpointAsync(const SeqNum& seqNum, const std::optional<Timestamp>& timestamp) {
+  const auto lastBlockid = getLastBlockIdCb_();
+  auto ret = std::async(std::launch::async, [this, seqNum, timestamp, lastBlockid]() -> void {
     createDbCheckpoint(lastBlockid, lastBlockid, seqNum, timestamp);
   });
+  dbCreateCheckPtFuture_.emplace(std::make_pair(lastBlockid, std::move(ret)));
+  return lastBlockid;
 }
```
dartdart26, line 198: I think it is not directly related to this PR, but I was wondering what is going to happen if the caller calls this method twice, in succession, such that `lastBlockid` is the same in the two calls? Would that be a problem?

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -236,8 +223,12 @@ void DbCheckpointManager::removeOldestDbCheckpoint() {
     dbCheckptMetadata_.dbCheckPoints_.erase(it);
     LOG_INFO(getLogger(), "removed db checkpoint, id: " << it->second.checkPointId_);
     updateDbCheckpointMetadata();
+    removeDbCheckpointFuture(it->second.checkPointId_);
   }
 }
+void DbCheckpointManager::removeDbCheckpointFuture(uint64_t id) {
+  if (auto it = dbCreateCheckPtFuture_.find(id); it != dbCreateCheckPtFuture_.end()) dbCreateCheckPtFuture_.erase(it);
```
dartdart26, line 230: We could use `dbCreateCheckPtFuture_.erase(id)` by using the `std::map::erase(Key)` overload.

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -199,15 +184,17 @@ void DbCheckpointManager::initializeDbCheckpointManager(std::shared_ptr<concord:
...
 
-void DbCheckpointManager::createDbCheckpointAsync(const SeqNum& seqNum, const std::optional<Timestamp>& timestamp) {
-  auto ret = std::async(std::launch::async, [this, seqNum, timestamp]() -> void {
-    const auto& lastBlockid = getLastBlockIdCb_();
+uint64_t DbCheckpointManager::createDbCheckpointAsync(const SeqNum& seqNum, const std::optional<Timestamp>& timestamp) {
+  const auto lastBlockid = getLastBlockIdCb_();
+  auto ret = std::async(std::launch::async, [this, seqNum, timestamp, lastBlockid]() -> void {
     createDbCheckpoint(lastBlockid, lastBlockid, seqNum, timestamp);
   });
+  dbCreateCheckPtFuture_.emplace(std::make_pair(lastBlockid, std::move(ret)));
```
dartdart26, line 196: Shouldn't we hold the `lock` here, for updating `dbCreateCheckPtFuture_`? We call `removeOldestDbCheckpoint()` in `createDbCheckpoint()` and that is in a separate thread.

bftengine/src/bftengine/DbCheckpointManager.cpp
nkumar04: May be we can skip db checkpoint creation if last created/in-prog db-checkpoint has same lastBlockId as the current one(returned from the recent lastBlockId api call). I can add this check. wdyt?

dartdart26: We could do that, but I think the user needs to somehow be aware that this has happened, because, otherwise, he/she calls `create` and nothing happens. It needs to be explicit, I think. Or maybe we can add an overload that might or might not create and one that always creates.

dartdart26: What I mean is that is an inherent race between the callers and the `createDbCheckpoint()` call and we might want to make the interface such that it cannot be misused.

bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -32,30 +32,43 @@
...
-  void createDbCheckpointAsync(const SeqNum& seqNum, const std::optional<Timestamp>& timestamp);
+  /***************************************************
+   *@Input parameter1: Request sequnce number
+   *@Input parameter2: Request timestamp
+   *@Description: Creates rocksdb checkpoint asynchronously if database has new state.
+   *              If we already have a checkpoint with same state as lastBlock in the DB, we reject the request.
+   *              Creating another db checkpoint with the same state as the previous one has no use.
+   *              Note: if ReplicaConfig::maxNumberOfDbCheckpoints is set to zero, then also we do not create rocksDb
+   *              and return std::nullopt in that case
+   *@Return: returns the last blockId at which create db checkpoint request is accepted. Else return std::nullopt
```
dartdart26, line 55: I think it is better to not be explicit about what the returned checkpoint ID is. We might say `an unique checkpoint ID`.

bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -32,30 +32,43 @@
...
+   ***************************************************/
+  std::optional<CheckpointId> createDbCheckpointAsync(const SeqNum& seqNum, const std::optional<Timestamp>& timestamp);
+
+  /***************************************************
+   *@Description: Creates last created db checkpoint metadata. This function blocks if db-checkpoint creation is running
```
dartdart26, line 60: Just a suggestion - we could consider returning the last already created checkpoint, in order to not block.

dartdart26: Or this lock is just for persisting to the metadata and not checkpoint creation?

dartdart26: The name of the function is `lastCreated`, so from that perspective we might consider not blocking.

dartdart26 (approved):

PR #2206, opened 2022-01-10
Add State Streaming Interface b/w TRC/S and placeholder methods for assoc. methods

shgandhi:
This PR adds a state streaming proto file to interface b/w TRC/S to allow gRPC state snapshot streaming. Alongside, placeholder methods have been added in clientservice, concord client and TRC. These methods need to be updated once the implementation is ready to be added.

### DIFF
--- a/client/clientservice/include/client/clientservice/client_service.hpp
+++ b/client/clientservice/include/client/clientservice/client_service.hpp
@@ -9,6 +9,8 @@
 // these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
 // file.
 
+#pragma once
+
 #include <memory>
 #include <string>
 
--- a/client/clientservice/include/client/clientservice/client_service.hpp
+++ b/client/clientservice/include/client/clientservice/client_service.hpp
@@ -14,6 +14,7 @@
 
 #include "event_service.hpp"
 #include "request_service.hpp"
+#include "state_snapshot_service.hpp"
 #include "Logger.hpp"
 #include "client/concordclient/concord_client.hpp"
 
@@ -25,18 +26,21 @@ class ClientService {
       : logger_(logging::getLogger("concord.client.clientservice")),
         client_(std::move(client)),
         event_service_(std::make_unique<EventServiceImpl>(client_)),
-        request_service_(std::make_unique<RequestServiceImpl>(client_)){};
+        request_service_(std::make_unique<RequestServiceImpl>(client_)),
+        state_snapshot_service_(std::make_unique<StateSnapshotServiceImpl>(client_)){};
 
   void start(const std::string& addr, uint64_t max_receive_msg_size);
 
   const std::string kRequestService{"vmware.concord.client.request.v1.RequestService"};
   const std::string kEventService{"vmware.concord.client.event.v1.EventService"};
+  const std::string kStateSnapshotService{"vmware.concord.client.statesnapshot.v1.StateSnapshotService"};
 
  private:
   logging::Logger logger_;
   std::shared_ptr<concord::client::concordclient::ConcordClient> client_;
   std::unique_ptr<EventServiceImpl> event_service_;
   std::unique_ptr<RequestServiceImpl> request_service_;
+  std::unique_ptr<StateSnapshotServiceImpl> state_snapshot_service_;
 };
 
 }  // namespace concord::client::clientservice
--- a/client/clientservice/include/client/clientservice/state_snapshot_service.hpp
+++ b/client/clientservice/include/client/clientservice/state_snapshot_service.hpp
@@ -1,6 +1,6 @@
 // Concord
 //
-// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
 //
 // This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
 // compliance with the Apache 2.0 License.
@@ -8,6 +8,8 @@
 // This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
 // these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
 // file.
+
+#pragma once
 
 #include <grpcpp/grpcpp.h>
 #include "state_snapshot.grpc.pb.h"
--- a/client/clientservice/include/client/clientservice/state_snapshot_service.hpp
+++ b/client/clientservice/include/client/clientservice/state_snapshot_service.hpp
@@ -1,0 +1,42 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include <grpcpp/grpcpp.h>
+#include "state_snapshot.grpc.pb.h"
+
+#include "Logger.hpp"
+#include "client/concordclient/concord_client.hpp"
+
+namespace concord::client::clientservice {
+
+class StateSnapshotServiceImpl final
+    : public vmware::concord::client::statesnapshot::v1::StateSnapshotService::Service {
+ public:
+  StateSnapshotServiceImpl(std::shared_ptr<concord::client::concordclient::ConcordClient> client)
+      : logger_(logging::getLogger("concord.client.clientservice.statesnapshot")), client_(client){};
+  grpc::Status GetRecentSnapshot(
+      grpc::ServerContext* context,
+      const vmware::concord::client::statesnapshot::v1::GetRecentSnapshotRequest* request,
+      vmware::concord::client::statesnapshot::v1::GetRecentSnapshotResponse* response) override;
+  grpc::Status StreamSnapshot(
+      grpc::ServerContext* context,
+      const vmware::concord::client::statesnapshot::v1::StreamSnapshotRequest* request,
+      grpc::ServerWriter<vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse>* stream) override;
+  grpc::Status ReadAsOf(grpc::ServerContext* context,
+                        const vmware::concord::client::statesnapshot::v1::ReadAsOfRequest* request,
+                        vmware::concord::client::statesnapshot::v1::ReadAsOfResponse* response) override;
+
+ private:
+  logging::Logger logger_;
+  std::shared_ptr<concord::client::concordclient::ConcordClient> client_;
+};
+
+}  // namespace concord::client::clientservice
--- a/client/clientservice/src/state_snapshot_service.cpp
+++ b/client/clientservice/src/state_snapshot_service.cpp
@@ -1,0 +1,48 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include "client/clientservice/state_snapshot_service.hpp"
+
+using grpc::Status;
+using grpc::ServerContext;
+using grpc::ServerWriter;
+
+using vmware::concord::client::statesnapshot::v1::GetRecentSnapshotRequest;
+using vmware::concord::client::statesnapshot::v1::GetRecentSnapshotResponse;
+using vmware::concord::client::statesnapshot::v1::StreamSnapshotRequest;
+using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
+using vmware::concord::client::statesnapshot::v1::ReadAsOfRequest;
+using vmware::concord::client::statesnapshot::v1::ReadAsOfResponse;
+
+namespace concord::client::clientservice {
+
+Status StateSnapshotServiceImpl::GetRecentSnapshot(ServerContext* context,
+                                                   const GetRecentSnapshotRequest* proto_request,
+                                                   GetRecentSnapshotResponse* response) {
+  // TODO: Add implementation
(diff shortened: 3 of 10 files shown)


client/clientservice/include/client/clientservice/client_service.hpp
```diff
@@ -14,6 +14,7 @@
 
 #include "event_service.hpp"
 #include "request_service.hpp"
+#include "state_snapshot_service.hpp"
```
dartdart26, line 17: Not related to this PR, but we could add `#pragma once` to this header file.

client/clientservice/include/client/clientservice/state_snapshot_service.hpp
```diff
@@ -0,0 +1,42 @@
...
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
```
dartdart26, line 11: We need to add `#pragma once` (or similar) to this header file.

client/clientservice/include/client/clientservice/state_snapshot_service.hpp
```diff
@@ -0,0 +1,42 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
```
dartdart26, line 3: Nit: Could change the year to 2022.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -0,0 +1,59 @@
+// Concord
+//
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
```
dartdart26, line 3: Nit: 2022

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -122,6 +122,21 @@ struct SubscribeRequest {
   std::variant<EventGroupRequest, LegacyEventRequest> request;
 };
 
+struct GetRecentSnapshotRequest {
```
dartdart26, line 125: Logically, maybe this data structure should be called a response. We have the same piece of data in the `GetRecentSnapshotResponse` protobuf message. Please correct me if I am misunderstanding the intention.

The request itself doesn't have any data members, as of now.

client/thin-replica-client/include/client/thin-replica-client/trs_connection.hpp
```diff
@@ -147,6 +148,21 @@ class TrsConnection {
   virtual Result readStateHash(const com::vmware::concord::thin_replica::ReadStateHashRequest& request,
                                com::vmware::concord::thin_replica::Hash* hash);
 
+  // Open a state snapshot subscription stream (connection has to be established before).
+  // A state snapshot subscription subscription stream will be open after
```
dartdart26, line 152: Typo: subscription mentioned twice.

thin-replica-server/proto/state_snapshot_concord.proto
```diff
@@ -0,0 +1,54 @@
+// Copyright 2021 VMware, all rights reserved
```
dartdart26, line 1: Nit: 2022

thin-replica-server/proto/state_snapshot_concord.proto
```diff
@@ -0,0 +1,54 @@
...
+// The Concord StateSnapshot service can be used by the concord client to 
+// retrieve state snapshot from a concord cluster
+// See `client/proto/state_snapshot.proto` for definition of state snapshot.
+
+service StateSnapshot {
+  // Stream a specific state snapshot in a resumable fashion as a finite stream of key-values.
+  // Key-values are streamed with lexicographic order on keys.
+  // Errors:
+  // NOT_FOUND: if a state snapshot with the requested ID is not (or no longer) available.
+  //            Applications are advised to retry initialization by fetching a new snapshot ID.
```
dartdart26, line 19: We could change `application` here, because it is the client that connects to the replica.

thin-replica-server/proto/state_snapshot_concord.proto
```diff
@@ -0,0 +1,54 @@
...
+  // INVALID_ARGUMENT: if resuming a stream using `from_key` and that key is not part of the
+  //                   state snapshot.
+  // UNAVAILABLE: if Concord Client is not ready yet to process requests.
```
dartdart26, line 22: That return code might be useful for the case where there is no available state snapshot yet on the replica. We can change `Concord Client` to `Concord Replica` here and maybe mention that the state snapshot is not available yet, but will be shortly.

dartdart26:
Looks great, thanks Shruti! The stubs will be extremely useful to whoever implements it.

I have left some minor comments only.

thin-replica-server/proto/state_snapshot_concord.proto
```diff
@@ -0,0 +1,54 @@
...
+  // FAILED_PRECONDITION: if a precondition in Concord Client is not satisfied. For example,
+  //                      the state might be corrupted or invalid and, in that case, there
+  //                      would be no point in proceeding with streaming.
+  // UNKNOWN: exact cause is unknown.
+  rpc StreamSnapshot(StreamSnapshotRequest) returns (stream StreamSnapshotResponse);
```
arc-vmware, line 27: If its a bidirectional stream, then the request should also have a stream keyword. A stream represents a sequence of message exchange in GRPC, it's generally a good practice not to name the Request or Response as a stream. Can we use Replay instead of Stream.
So if its a server-side streaming then we might say:
rpc ReplaySnapshot(SnapshotRequest) returns (stream SnapshotResponse);

And if its a bidirectional stream we can use:
rpc ReplaySnapshot(stream SnapshotRequest) returns (stream SnapshotResponse);

arc-vmware (commented):
Thanks @shgandhi , nice changes.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -0,0 +1,59 @@
...
+using namespace std::chrono_literals;
+
+namespace concord::client::clientservice {
+
+Status StateSnapshotServiceImpl::GetRecentSnapshot(ServerContext* context,
+                                                   const GetRecentSnapshotRequest* proto_request,
+                                                   GetRecentSnapshotResponse* response) {
+  // TODO: Add implementation
+  auto status = grpc::Status::OK;
+  return status;
```
robem, line 40: There is specifically a gRPC code you can use - `UNIMPLEMENTED`

thin-replica-server/proto/state_snapshot_concord.proto
```diff
@@ -0,0 +1,54 @@
...
+// Concord Client's StateSnapshot Service
+
+syntax = "proto3";
+package com.vmware.concord.state_snapshot_concord;
+
+option java_package = "com.vmware.concord.state_snapshot_concord";
+
+// The Concord StateSnapshot service can be used by the concord client to 
+// retrieve state snapshot from a concord cluster
+// See `client/proto/state_snapshot.proto` for definition of state snapshot.
```
robem, line 12: The path is not correct. Instead, I suggest `client/proto/state_snapshot` to point to the directory instead of a specific version of the interface.

thin-replica-server/proto/state_snapshot_concord.proto
robem: The API is most likely modeled after `state_snapshot.proto`.
It is not a bi-directional stream, instead, you send one request and as a response you get multiple (finite amount of) answers.

If we want to get rid of the word `Stream` then I suggest the following:
`rpc GetSnapshot(GetSnapshotRequest) returns (stream GetSnapshotResponse)` or
`rpc GetSnapshot(GetSnapshotRequest) returns (stream KeyValuePairs)`

If we change the naming here then we should also change it in `state_snapshot.proto`. @dartdart26

thin-replica-server/proto/state_snapshot_concord.proto
```diff
@@ -0,0 +1,54 @@
+// Copyright 2021 VMware, all rights reserved
+//
+// Concord Client's StateSnapshot Service
+
+syntax = "proto3";
+package com.vmware.concord.state_snapshot_concord;
```
robem, line 6: `com.vmware.concord.state_snapshot_concord` vs `vmware.concord.client.statesnapshot`

Let's be consistent, I suggest `vmware.concord.statesnapshot`.

FYI: https://cloud.google.com/apis/design/naming_convention#package_names

client/thin-replica-client/src/trs_connection.cpp
```diff
@@ -356,4 +356,24 @@ TrsConnection::Result TrsConnection::readHash(Hash* hash) {
   return Result::kFailure;
 }
 
+TrsConnection::Result TrsConnection::openStateSnapshotStream(
+    const com::vmware::concord::state_snapshot_concord::StreamSnapshotRequest& request) {
+  // TODO: Add implementation
+  return Result::kSuccess;
```
robem, line 362: Would you mind adding a `LOG_ERROR` message that prints "unimplemented" or "todo" or something similar - here and in all the other "todo" functions? Meaning, if the function gets called without an implementation then that must be a mistake. Alternatively, we can even add an ASSERT because it would be a programming error to call those functions.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -0,0 +1,59 @@
...
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include <chrono>
+#include <iostream>
+#include <opentracing/tracer.h>
+#include <thread>
```
robem, line 15: Those definitions aren't used.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -0,0 +1,59 @@
...
+using grpc::ServerWriter;
+
+using vmware::concord::client::statesnapshot::v1::GetRecentSnapshotRequest;
+using vmware::concord::client::statesnapshot::v1::GetRecentSnapshotResponse;
+using vmware::concord::client::statesnapshot::v1::StreamSnapshotRequest;
+using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
+using vmware::concord::client::statesnapshot::v1::ReadAsOfRequest;
+using vmware::concord::client::statesnapshot::v1::ReadAsOfResponse;
+
+using namespace std::chrono_literals;
```
robem, line 31: Not used.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -0,0 +1,59 @@
...
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include <chrono>
+#include <iostream>
+#include <opentracing/tracer.h>
+#include <thread>
+
+#include "client/clientservice/state_snapshot_service.hpp"
+#include "client/concordclient/concord_client.hpp"
```
robem, line 18: Not used.

robem (changes requested):

thin-replica-server/proto/state_snapshot_concord.proto
dartdart26: I prefer we don't change it, because this interface is used in many places, affecting different teams.

thin-replica-server/proto/state_snapshot_concord.proto
shgandhi: Thanks for catching, updated.

thin-replica-server/proto/state_snapshot_concord.proto
shgandhi: We will have two proto files with the same name then, I wanted to avoid that, i.e., why I went the inconsistent route. Will update.

shgandhi: I was getting naming conflicts when using same name for both files. I have renamed the TR `state_snapshot.proto` to `replica_state_snapshot.proto`.

client/thin-replica-client/include/client/thin-replica-client/trs_connection.hpp
```diff
@@ -147,6 +150,21 @@ class TrsConnection {
   virtual Result readStateHash(const com::vmware::concord::thin_replica::ReadStateHashRequest& request,
                                com::vmware::concord::thin_replica::Hash* hash);
 
+  // Open a state snapshot subscription stream (connection has to be established before).
+  // A state snapshot subscription stream will be open after
+  // openStateSnapshotStream returns, if and only if openStateSnapshotStream returns
+  // Result::kSuccess; a stream will not be open in the failure and timeout cases.
```
robem, line 156: I think the word "subscription" is misplaced here. From my point of view, you subscribe to all updates that exist now and **in the future**. For state snapshots, the moment you call the API, the updates already exists and are ready to be consumed. Meaning, there are no updates that you're waiting on to be produced.
Therefore, I would just call it a stream, no subscription (same comment below).

thin-replica-server/proto/replica_state_snapshot.proto
```diff
@@ -0,0 +1,56 @@
+// Copyright 2022 VMware, all rights reserved
```
robem, line 1: You're missing
```
// Concord
//
```

thin-replica-server/proto/replica_state_snapshot.proto
```diff
@@ -0,0 +1,56 @@
...
+//
+// Concord's StateSnapshotService
+
+syntax = "proto3";
+package com.vmware.concord.replica_state_snapshot;
```
robem, line 6: No `com.`

robem (approved):

thin-replica-server/proto/replica_state_snapshot.proto
shgandhi: Updated.

thin-replica-server/proto/replica_state_snapshot.proto
shgandhi: Updated.

client/thin-replica-client/include/client/thin-replica-client/trs_connection.hpp
shgandhi: Updated.

shgandhi:
Thanks for the review @dartdart26, @robem and @arc-vmware.

PR #2208, opened 2022-01-11
Fix a malformed JSON generated by getReplicaLastStableSeqNum()

tdimitrov:
getReplicaLastStableSeqNum() in ReplicaImp generates a string which is
returned as a result of "status get replica" diagnostics commands.

The function used kvContainerToJson() and kContainerToJson() functions
from concordUtils, which however are not intended to work with nested
containers.

The code is replaced with a call to nlohmann JSON library which does the
serialisation.


teoparvanov (approved):
LGTM!

PR #2209, opened 2022-01-11
Adding the CMF and key for Public State key set

arc-vmware:
The State keys which are categorized with Public will be stored in internal column families


dartdart26 (approved):
Thanks Amitava!

PR #2213, opened 2022-01-12
Replace PartialProofsSet with CollectorOfThresholdSignatures

evdzhurov:
This PR removes the use of PartialProofsSet in the fast commit path and replaces it with instances of CollectorOfThresholdSignatures to handle commit proof messages similar to the slow path.

Major changes:
* PartialProofsSet is removed - all the functionality is moved to SeqNumInfo and mostly handled by CollectorOfThresholdSignatures.
* onMessage(FullCommitProofMsg) becomes simplified since CollectorOfThresholdSignatures handles the verification now.
* FullCommitProofMsg is no longer passed as an internal replica message when created. Instead the corresponding callback onFastPathCommitVerifyCombinedSigResult handles the execution if the validation succeeds.
* Some methods are now parametrized with a CommitPath variable.
* FullCommitProofMsg and PartialCommitProofMsg methods thresholSignatureLength() and thresholSignature() have been renamed same as other proof messages to work with the CollectorOfThresholdSignatures  template class.


toly-kournik:
@evdzhurov , it would be also a good idea to subclass PartialCommitProofMsg from SignedShareBase

evdzhurov:
> @evdzhurov , it would be also a good idea to subclass PartialCommitProofMsg from SignedShareBase

Agreed, but it's probably better to be addressed in a separate PR. I'll check with @teoparvanov

HristoStaykov (changes requested):

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1549,11 +1522,6 @@ void ReplicaImp::onInternalMsg(InternalMessage &&msg) {
     return finishExecutePrePrepareMsg(t->prePrepareMsg, t->pAccumulatedRequests);
   }
 
-  // Handle a full commit proof sent by self
-  if (auto *fcp = std::get_if<FullCommitProofMsg *>(&msg)) {
-    return onInternalMsg(*fcp);
-  }
```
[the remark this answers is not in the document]
evdzhurov: This seems like a leftover, I should remove it.

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -266,7 +289,135 @@ Time SeqNumInfo::getTimeOfFirstRelevantInfoFromPrimary() const { return firstSee
...
+  auto result = false;
+  switch (cPath) {
+    case CommitPath::OPTIMISTIC_FAST:
+      result = fastPathOptimisticCollector->addMsgWithPartialSignature(m, repId);
+      break;
+    case CommitPath::FAST_WITH_THRESHOLD:
+      result = fastPathThresholdCollector->addMsgWithPartialSignature(m, repId);
+      break;
+    default:
+      ConcordAssert(false && "Unhandled CommitPath!");
```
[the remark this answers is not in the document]
evdzhurov: It seems it does, however what error handling do you suggest in this case? There are other examples where something already validated in the Validate method is a ConcordAssert - for example:
ConcordAssert(repsInfo->isIdOfPeerReplica(msgSender));
ConcordAssert(repsInfo->isCollectorForPartialProofs(msgView, msgSeqNum));
in void ReplicaImp::onMessage<PartialCommitProofMsg>(PartialCommitProofMsg *msg)

Do we:
- Rely entirely on the message validate method and omit additional error checking? 
- Keep additional error checking - either an assert or error log and if it occurs we improve the validate method?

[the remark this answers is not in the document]
HristoStaykov: Basically the convention is to assert only for things that cannot escape the validate method of the message. Is it possible for a malicious replica to put a value there that could pass the validate method and trigger the assert?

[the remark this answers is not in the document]
evdzhurov: After some consideration the validate method of PartialCommitProofMsg can in fact allow invalid values for the commit path in the message that would trigger this assert. This will be addressed as part of the PR.

HristoStaykov (approved):
LGTM, thanks, @evdzhurov !

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1628,12 +1596,28 @@ void ReplicaImp::onInternalMsg(InternalMessage &&msg) {
...
+    return onCommitCombinedSigFailed(ccsf->seqNumber, ccsf->view, CommitPath::SLOW, ccsf->replicasWithBadSigs);
   }
   if (auto *vccs = std::get_if<VerifyCombinedCommitSigResultInternalMsg>(&msg)) {
     return onCommitVerifyCombinedSigResult(vccs->seqNumber, vccs->view, vccs->isValid);
   }
 
+  // Handle Fast Path Commit related internal messages
+  if (auto *ccss = std::get_if<FastPathCombinedCommitSigSucceededInternalMsg>(&msg)) {
+    return onFastPathCommitCombinedSigSucceeded(ccss->seqNumber,
+                                                ccss->view,
```
arc-vmware, line 1608: Should we continue calling this function if currentView() != ccss->view?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2174,6 +2153,131 @@ void ReplicaImp::onCommitVerifyCombinedSigResult(SeqNum seqNumber, ViewNum view,
...
+                                                      const concordUtils::SpanContext &span_context) {
+  SCOPED_MDC_PRIMARY(std::to_string(currentPrimary()));
+  SCOPED_MDC_SEQ_NUM(std::to_string(seqNumber));
+  SCOPED_MDC_PATH(CommitPathToMDCString(cPath));
+  LOG_TRACE(THRESHSIGN_LOG, KVLOG(seqNumber, view, combinedSigLen));
+
+  if (isCollectingState() && mainLog->insideActiveWindow(seqNumber)) {
+    mainLog->get(seqNumber).resetCommitSignatures(cPath);
+    LOG_INFO(CNSUS, "Collecting state, reset commit signatures");
+    return;
```
arc-vmware, line 2170: Should we delete FullCommitProofMsg for the seqNumber here? It may lead to a mem leak.

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -266,7 +289,135 @@ Time SeqNumInfo::getTimeOfFirstRelevantInfoFromPrimary() const { return firstSee
...
+    case CommitPath::OPTIMISTIC_FAST:
+      result = directAdd ? fastPathOptimisticCollector->initMsgWithCombinedSignature(m)
+                         : fastPathOptimisticCollector->addMsgWithCombinedSignature(m);
+      break;
+    case CommitPath::FAST_WITH_THRESHOLD:
+      result = directAdd ? fastPathThresholdCollector->initMsgWithCombinedSignature(m)
+                         : fastPathThresholdCollector->addMsgWithCombinedSignature(m);
+      break;
+    default:
+      ConcordAssert(false && "Unhandled CommitPath!");
```
arc-vmware, line 416: Better to use LOG_ERROR and ConcordAssert(false)

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -405,6 +618,127 @@ IncomingMsgsStorage& SeqNumInfo::ExFuncForCommitCollector::incomingMsgsStorage(v
...
+FullCommitProofMsg* SeqNumInfo::ExFuncForFastPathOptimisticCollector::createCombinedSignatureMsg(
+    void* context,
+    SeqNum seqNumber,
+    ViewNum viewNumber,
+    const char* const combinedSig,
+    uint16_t combinedSigLen,
+    const concordUtils::SpanContext& span_context) {
+  InternalReplicaApi* r = (InternalReplicaApi*)context;
+
+  return new FullCommitProofMsg(
```
arc-vmware, line 634: it would be good to check if this message is deleted by the caller.

bftengine/src/bftengine/SeqNumInfo.cpp
```diff
@@ -405,6 +618,127 @@ IncomingMsgsStorage& SeqNumInfo::ExFuncForCommitCollector::incomingMsgsStorage(v
...
+FullCommitProofMsg* SeqNumInfo::ExFuncForFastPathThresholdCollector::createCombinedSignatureMsg(
+    void* context,
+    SeqNum seqNumber,
+    ViewNum viewNumber,
+    const char* const combinedSig,
+    uint16_t combinedSigLen,
+    const concordUtils::SpanContext& span_context) {
+  InternalReplicaApi* r = (InternalReplicaApi*)context;
+
+  return new FullCommitProofMsg(
```
arc-vmware, line 694: It would be good to check if this message is deleted by the caller of createCombinedSignatureMsg() function.

bftengine/src/bftengine/SeqNumInfo.hpp
```diff
@@ -173,14 +186,79 @@ class SeqNumInfo {
...
+    static InternalMessage createInterVerifyCombinedSigResult(SeqNum seqNumber, ViewNum viewNumber, bool isValid);
+
+    // from the ReplicaImp object
+    static uint16_t numberOfRequiredSignatures(void* context);
+    static std::shared_ptr<IThresholdVerifier> thresholdVerifier(SeqNum seqNumber);
+    static concord::util::SimpleThreadPool& threadPool(void* context);
+    static IncomingMsgsStorage& incomingMsgsStorage(void* context);
+  };
+
+  class ExFuncForFastPathThresholdCollector {
```
arc-vmware, line 217: ExFuncForFastPathOptimisticCollector and ExFuncForFastPathThresholdCollector looks same. So is it possible to create a single class Collector and use to using  <type traits1> = Collector<templates for type trait1> and using <type traits2> = Collector<templates for type trait 2>
And use the type traits instead of separate classes. Please check if this is feasible?
Another idea is to use a single collector interface and use the interface instead if type traits.

arc-vmware: If it's too much of a design change then ignore the comment, we can do the cleanup later.

arc-vmware (approved):
Thanks @evdzhurov , the changes LGTM
I have very few minor comments.

bftengine/src/bftengine/SeqNumInfo.cpp
evdzhurov: This method is called internally by CollectorOfTrhesholdSignatures and the caller becomes the owner of the message.

bftengine/src/bftengine/SeqNumInfo.cpp
evdzhurov: This method is called internally by CollectorOfTrhesholdSignatures and the caller becomes the owner of the message.

bftengine/src/bftengine/ReplicaImp.cpp
evdzhurov: CollectorOfThresholdSignatures takes ownership of the combinedValidSignatureMsg or candidateCombinedSignatureMsg (the FullCommitProofMsg) depending if it created it or received it for verification from another replica. In both cases CollectorOfThresholdSignatures will delete it either when resetCommitSignatures or the CollectorOfThresholdSignatures is destroyed. The only case we delete a FullCommitProofMsg manually is if we receive it but fail to add it to CollectorOfThresholdSignatures.

bftengine/src/bftengine/ReplicaImp.cpp
evdzhurov: This check is done inside onFastPathCommitCombinedSigSucceeded similar to onCommitCombinedSigSucceeded. Should we handle it differently?

bftengine/src/bftengine/SeqNumInfo.cpp
evdzhurov: I've split the ConcordAsserts as suggested since this is the more common usage. Also the log now includes the invalid commit path value that caused the error.

PR #2214, opened 2022-01-12
Complete TLS exchange improvements

yontyon:
In this PR we complete the improvements for TLS exchange procedure

- remove the restriction for no more than f exchanges at once
- allow master key based signed certificates in the client side
- refactor and improve the basic CRE mechanism
- refactor and fix the Apollo tests

### DIFF
--- a/bftclient/include/bftclient/StateControl.hpp
+++ b/bftclient/include/bftclient/StateControl.hpp
@@ -1,6 +1,6 @@
 // Concord
 //
-// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2021 VMware, Inc. All Rights Reserved.
 //
 // This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
 // compliance with the Apache 2.0 License.
--- a/bftclient/include/bftclient/StateControl.hpp
+++ b/bftclient/include/bftclient/StateControl.hpp
@@ -1,0 +1,30 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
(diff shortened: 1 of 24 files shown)


bftclient/include/bftclient/StateControl.hpp
```diff
@@ -0,0 +1,30 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
```
nkumar04, line 3: nit: 2022

bftclient/include/bftclient/config.h
```diff
@@ -68,6 +68,7 @@ struct ClientConfig {
   RetryTimeoutConfig retry_timeout_config;
   std::optional<std::string> transaction_signing_private_key_file_path = std::nullopt;
   std::optional<concord::secretsmanager::SecretData> secrets_manager_config = std::nullopt;
+  std::optional<std::string> replicas_master_key_folder_path = "./replicas_rsa_keys";
```
nkumar04, line 71: nit: Should we name path as: "./replicas_keys" or "./replicas_master_keys"

bftclient/src/bft_client.cpp
```diff
@@ -57,6 +58,13 @@ Client::Client(std::unique_ptr<bft::communication::ICommunication> comm, const C
   }
   communication_->setReceiver(config_.id.val, &receiver_);
   communication_->start();
+  if (config_.replicas_master_key_folder_path.has_value()) {
+    bft::communication::StateControl::instance().setGetPeerPubKeyMethod([&](uint32_t rid) {
+      concord::secretsmanager::SecretsManagerPlain psm_;
+      std::string key_path = config_.replicas_master_key_folder_path.value() + "/" + std::to_string(rid) + "/pub_key";
```
nkumar04, line 64: Do we exchange replica main public key if a replica is down in the initial start.
Do we need to add replica main key exchange here?

nkumar04 (approved):
LGTM!
Thanks!

bftclient/include/bftclient/config.h
yontyon: we should, but as long as we work with rsa keys I prefer to specify this (to ease debugging and so)

bftclient/src/bft_client.cpp
yontyon: right, thanks

PR #2215, opened 2022-01-12
Removing stable state check in checkpoint messgae validation

nkumar04:
Replicas mark checkpoint to be stable when it receives 2f+1 checkpoint messages. Some of the messages may have stable flag set to 0. Removing this check in checkpoint message validation rocksdb checkpoints


itsmeakashgoyal (approved):
LGTM.
Thanks @nkumar04

PR #2221, opened 2022-01-13
Enable concurrent delete of categories when deleting a block

cloudnoize:
A block is composed of several categories, where each category writes to its own set of column families.
Deleting a block iterates over those categories, and per-category performs the dedicated logic that deletes a block for that category.

The Merkle and the Versioned category perform some non-trivial operations that make the deletion of a block a slow operation.

An important observation is that deleting a category is a local event i.e., each category is isolated, and its set of updates 
affects only its column families. Therefore we can concurrently perform the internal category deletion.

Since the NativeWriteSet is not thread-safe, this PR introduces a LocalWriteSet, that has the same put and delete interface as the NativeWriteSet, it knows to keep the order of the operations and can be instantiated per category and then merged into the single NativeWriteSet that eventually performs the write to storage.

When deleting a genesis block, if the Merkle or Version categories have some amount of keys that exceed a threshold, their deletion will be executed as a job on a dedicated thread pool. 

Measurements show that, on average, it improves the deletion by 15%.

### DIFF
--- a/kvbc/include/categorization/details.h
+++ b/kvbc/include/categorization/details.h
@@ -23,6 +23,7 @@
 #include <string>
 #include <type_traits>
 #include <vector>
+#include <queue>
 
 namespace concord::kvbc::categorization::detail {
 
@@ -96,5 +97,57 @@ inline void sortAndRemoveDuplicates(std::vector<std::string> &vec) {
   std::sort(vec.begin(), vec.end());
   vec.erase(std::unique(vec.begin(), vec.end()), vec.end());
 }
+// A LocalWriteBatch is used in a multithreaded context, where several jobs need to collect updates.
+// Each job gets its LocalWriteBatch.
+// then all updates are added to the Global NativeWriteBatch, which is not thread safe.
+struct LocalWriteBatch {
+  struct Put {
+    std::string cFamily;
+    std::string key;
+    std::string value;
+  };
+
+  struct Delete {
+    std::string cFamily;
+    std::string key;
+  };
+
+  template <typename KeySpan, typename ValueSpan>
+  void put(const std::string &cFamily, const KeySpan &key, const ValueSpan &value) {
+    puts_.push({cFamily,
+                std::string(reinterpret_cast<const char *>(key.data()), key.size()),
+                std::string(reinterpret_cast<const char *>(value.data()), value.size())});
+    operations_.push(put_op_);
+  }
+
+  void moveToBatch(storage::rocksdb::NativeWriteBatch &batch) noexcept {
+    while (!operations_.empty()) {
+      if (operations_.front() == put_op_) {
+        const auto &put = puts_.front();
+        batch.put(put.cFamily, put.key, put.value);
+        puts_.pop();
+      } else {
+        const auto &del = dels_.front();
+        batch.del(del.cFamily, del.key);
+        dels_.pop();
+      }
+      operations_.pop();
+    }
+    ConcordAssert(puts_.empty());
+    ConcordAssert(dels_.empty());
+  }
+
+  template <typename KeySpan>
+  void del(const std::string &cFamily, const KeySpan &key) {
+    dels_.push(Delete{cFamily, std::string(reinterpret_cast<const char *>(key.data()), key.size())});
+    operations_.push(delete_op_);
+  }
+
+  std::queue<Put> puts_;
+  std::queue<Delete> dels_;
+  std::queue<uint8_t> operations_;
+  const uint8_t delete_op_{0};
+  const uint8_t put_op_{1};
+};  // namespace concord::kvbc::categorization::detail
 
 }  // namespace concord::kvbc::categorization::detail
--- a/kvbc/include/categorization/kv_blockchain.h
+++ b/kvbc/include/categorization/kv_blockchain.h
@@ -159,17 +159,17 @@ class KeyValueBlockchain {
   void deleteGenesisBlock(BlockId block_id,
                           const std::string& category_id,
                           const ImmutableOutput& updates_info,
-                          storage::rocksdb::NativeWriteBatch&);
+                          detail::LocalWriteBatch&);
 
   void deleteGenesisBlock(BlockId block_id,
                           const std::string& category_id,
                           const VersionedOutput& updates_info,
-                          storage::rocksdb::NativeWriteBatch&);
+                          detail::LocalWriteBatch&);
 
   void deleteGenesisBlock(BlockId block_id,
                           const std::string& category_id,
                           const BlockMerkleOutput& updates_info,
-                          storage::rocksdb::NativeWriteBatch&);
+                          detail::LocalWriteBatch&);
 
   void deleteLastReachableBlock(BlockId block_id,
                                 const std::string& category_id,
@@ -220,6 +220,8 @@ class KeyValueBlockchain {
 
   // currently we are operating with single thread
   util::ThreadPool thread_pool_{1};
+  // For concurrent deletion of the categories inside a block.
+  util::ThreadPool prunning_thread_pool_{2};
 
   // metrics
   std::shared_ptr<concordMetrics::Aggregator> aggregator_;
--- a/kvbc/src/categorization/block_merkle_category.cpp
+++ b/kvbc/src/categorization/block_merkle_category.cpp
@@ -192,14 +192,15 @@ void putKeys(NativeWriteBatch& batch,
     ++deletes_it;
   }
 }
-
-void putLatestTreeVersion(uint64_t tree_version, NativeWriteBatch& batch) {
+template <typename Batch>
+void putLatestTreeVersion(uint64_t tree_version, Batch& batch) {
   batch.put(BLOCK_MERKLE_INTERNAL_NODES_CF, rootKey(0), rootKey(tree_version));
 }
 
 // Write serialized stale keys at the given *tree* version. These keys are safe to delete when the
 // tree version is pruned.
-void putStaleKeys(NativeWriteBatch& batch, sparse_merkle::StaleNodeIndexes&& stale_batch) {
+template <typename Batch>
+void putStaleKeys(Batch& batch, sparse_merkle::StaleNodeIndexes&& stale_batch) {
   auto stale = StaleKeys{};
   for (auto& key : stale_batch.internal_keys) {
     stale.internal_keys.push_back(serialize(toBatchedInternalNodeKey(key)));
@@ -211,7 +212,8 @@ void putStaleKeys(NativeWriteBatch& batch, sparse_merkle::StaleNodeIndexes&& sta
   batch.put(BLOCK_MERKLE_STALE_CF, tree_version_key, serialize(stale));
 }
 
-void putMerkleNodes(NativeWriteBatch& batch, sparse_merkle::UpdateBatch&& update_batch) {
+template <typename Batch>
+void putMerkleNodes(Batch& batch, sparse_merkle::UpdateBatch&& update_batch) {
   for (const auto& [leaf_key, leaf_val] : update_batch.leaf_nodes) {
     auto ser_key = serialize(leafKeyToVersionedKey(leaf_key));
     batch.put(BLOCK_MERKLE_LEAF_NODES_CF, ser_key, leaf_val.value.string_view());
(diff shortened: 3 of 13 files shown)


kvbc/include/categorization/kv_blockchain.h
```diff
@@ -220,6 +220,8 @@ class KeyValueBlockchain {
 
   // currently we are operating with single thread
   util::ThreadPool thread_pool_{1};
+  // For concurrent deletion of the categories inside a block.
+  util::ThreadPool prunning_thread_pool_{2};
```
yontyon, line 224: wouldn't it be better to make it configurable?

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -453,6 +453,9 @@ void KeyValueBlockchain::deleteStateTransferBlock(const BlockId block_id) {
 void KeyValueBlockchain::deleteGenesisBlock() {
   // We assume there are blocks in the system.
   auto genesis_id = block_chain_.getGenesisBlockId();
+  // If a versioned/Merkle category contain more keys than concurrent_threshold
+  // It will be execute in a seperate thread.
+  const auto concurrent_threshold = 10;
```
yontyon, line 458: configuration?

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -470,13 +473,58 @@ void KeyValueBlockchain::deleteGenesisBlock() {
...
+                std::ref(write_batches.back())));
+
+          } else {
+            deleteGenesisBlock(genesis_id, category_id, update_info, write_batches.back());
+          }
+        },
+        update_info);
+  }
+  // wait for jobs to finish
+  for (std::vector<std::future<void>>::size_type i = 0; i < futures.size(); ++i) {
```
yontyon, line 515: I wonder, wouldn't it be better not to wait for the actual delete?
Instead, we can simply change the current genesis block id (increase it by one) and let the actual delete happen in its own time.
It is a bit risky, but maybe it is worth a try, WDYT?

yontyon (approved):
LGTM!
Thanks

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: I don't think it can work with our current implementation.
For example, if we delete a range, it will allow deleting more than one block in parallel, and categories share data across blocks e.g., the Merkle active keys column family.

kvbc/include/categorization/kv_blockchain.h
cloudnoize: I thought of making it configurable, but because we currently have only two categories that require running in their thread I decided to hard code it. But maybe it does worth the extra effort.

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: Same as above, I don't think there is a lot of room to tune it, so I decided to hard code it, but maybe it does worth the extra effort.

storage/include/rocksdb/native_write_batch.ipp
```diff
@@ -55,6 +55,12 @@ void NativeWriteBatch::del(const std::string &cFamily, const KeySpan &key) {
   detail::throwOnError("batch del failed"sv, std::move(s));
 }
 
+template <>
+inline void NativeWriteBatch::del<::rocksdb::Slice>(const std::string &cFamily, const ::rocksdb::Slice &key) {
```
dartdart26, line 59: Nit: Wouldn't `detail::toSlice()` work on a `::rocksdb::Slice` too? If yes, we can consider removing this specialization.

kvbc/include/categorization/details.h
```diff
@@ -96,5 +97,47 @@ inline void sortAndRemoveDuplicates(std::vector<std::string> &vec) {
...
+    ConcordAssert(dels_.empty());
+  }
+
+  template <typename KeySpan>
+  void del(const std::string &cFamily, const KeySpan &key) {
+    dels_.push({cFamily, std::string(reinterpret_cast<const char *>(key.data()), key.size())});
+    order_.push(delete_op_);
+  }
+
+  std::string cFamily_;
```
dartdart26, line 135: I think this member is unused.

kvbc/include/categorization/details.h
```diff
@@ -96,5 +97,47 @@ inline void sortAndRemoveDuplicates(std::vector<std::string> &vec) {
...
+  std::queue<std::pair<std::string, std::pair<std::string, std::string>>> puts_;
+  std::queue<std::pair<std::string, std::string>> dels_;
+  std::queue<uint8_t> order_;
```
dartdart26, line 138: Nit: Might consider naming that member something like `operations_` or similar.

kvbc/include/categorization/details.h
```diff
@@ -96,5 +97,47 @@ inline void sortAndRemoveDuplicates(std::vector<std::string> &vec) {
...
+  }
+
+  template <typename KeySpan>
+  void del(const std::string &cFamily, const KeySpan &key) {
+    dels_.push({cFamily, std::string(reinterpret_cast<const char *>(key.data()), key.size())});
+    order_.push(delete_op_);
+  }
+
+  std::string cFamily_;
+  std::queue<std::pair<std::string, std::pair<std::string, std::string>>> puts_;
```
dartdart26, line 136: Nit: Might consider using a `struct` instead of pairs for readability, i.e.:

```
struct Put {
    std::string cFamily;
    std::string key;
    std::string value;
};

struct Delete {
    std::string cFamily;
    std::string key;
}
```

kvbc/include/categorization/details.h
```diff
@@ -96,5 +97,47 @@ inline void sortAndRemoveDuplicates(std::vector<std::string> &vec) {
...
+struct LocalWriteBatch {
+  template <typename KeySpan, typename ValueSpan>
+  void put(const std::string &cFamily, const KeySpan &key, const ValueSpan &value) {
+    puts_.push({cFamily,
+                {std::string(reinterpret_cast<const char *>(key.data()), key.size()),
+                 std::string(reinterpret_cast<const char *>(value.data()), value.size())}});
+    order_.push(put_op_);
+  }
+
+  void moveToBatch(storage::rocksdb::NativeWriteBatch &batch) {
```
dartdart26, line 112: Nit: I wish std::queue::pop() was noexcept, making this method thread-safe. One way to "achieve" it is to make the whole method `noexcept`, denoting that it either succceeds or we cannot continue safely.

kvbc/include/categorization/details.h
```diff
@@ -96,5 +97,47 @@ inline void sortAndRemoveDuplicates(std::vector<std::string> &vec) {
   std::sort(vec.begin(), vec.end());
   vec.erase(std::unique(vec.begin(), vec.end()), vec.end());
 }
+// A LocalWriteBatch is used in a multithreaded context, where several jobs need to collect updates.
+// Each job gets its LocalWriteBatch.
+// then all updates are added to the Global NativeWriteBatch, which is not thread safe.
+struct LocalWriteBatch {
```
dartdart26, line 103: Nice idea and a simple inteface! Allows us to have minimal changes in other places.

kvbc/src/categorization/block_merkle_category.cpp
```diff
@@ -303,7 +305,8 @@ std::vector<KeyHash> remainingActiveKeys(std::vector<KeyHash>& previously_active
   return active_keys;
 }
 
-void addStaleKeysToDeleteBatch(const ::rocksdb::PinnableSlice& slice, uint64_t tree_version, NativeWriteBatch& batch) {
+template <typename Batch>
+void addStaleKeysToDeleteBatch(const ::rocksdb::PinnableSlice& slice, uint64_t tree_version, Batch&& batch) {
```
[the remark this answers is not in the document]
cloudnoize: it should be a reference, like the others, thanks for the catch

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -470,13 +473,58 @@ void KeyValueBlockchain::deleteGenesisBlock() {
...
+                std::ref(update_info),
+                std::ref(write_batches.back())));
+
+          } else {
+            deleteGenesisBlock(genesis_id, category_id, update_info, write_batches.back());
+          }
+        },
+        update_info);
+  }
+  // wait for jobs to finish
```
dartdart26, line 514: Nit: Could use below for simplicity:

```
for (auto &f : futures) {
    f.wait();
}
```

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -470,13 +473,58 @@ void KeyValueBlockchain::deleteGenesisBlock() {
...
+  for (std::vector<std::future<void>>::size_type i = 0; i < futures.size(); ++i) {
+    futures[i].get();
+  }
+  // Move the local batches into the global batch
+  for (std::vector<detail::LocalWriteBatch>::size_type i = 0; i < write_batches.size(); ++i) {
```
dartdart26, line 519: Nit: Same as above, a range-for might be easier to read.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -470,13 +473,58 @@ void KeyValueBlockchain::deleteGenesisBlock() {
...
+    uint64_t num_of_keys = 0;
+    // Decide whether to perform the deletion on a separate thread based on the number of keys
+    // Immutable category, should always be sequential as its deletion is fast.
+    if (std::holds_alternative<VersionedOutput>(update_info)) {
+      num_of_keys = std::get<VersionedOutput>(update_info).keys.size();
+    } else if (std::holds_alternative<BlockMerkleOutput>(update_info)) {
+      num_of_keys = std::get<BlockMerkleOutput>(update_info).keys.size();
+    }
+    std::visit(
+        [&genesis_id, category_id = category_id, &write_batches, &futures, &num_of_keys, this](
```
dartdart26, line 491: Do we need to capture `genesis_id` by reference? If not, it might be easier to maintain if we copy it.

kvbc/src/categorization/kv_blockchain.cpp
```diff
@@ -470,13 +473,58 @@ void KeyValueBlockchain::deleteGenesisBlock() {
...
+            const auto& update_info) {
+          write_batches.push_back(detail::LocalWriteBatch());
+          if (num_of_keys > concurrent_threshold) {
+            futures.push_back(prunning_thread_pool_.async(
+                [&](BlockId genesis_id,
+                    std::string category_id,
+                    const auto& update_info,
+                    detail::LocalWriteBatch& write_batch) {
+                  LOG_DEBUG(CAT_BLOCK_LOG, "Deletion of " << category_id << " will be performed in a seperate thread");
+                  deleteGenesisBlock(genesis_id, std::ref(category_id), std::ref(update_info), std::ref(write_batch));
```
dartdart26, line 501: What is the rationale for using `std:ref()` here?

kvbc/test/categorization/block_merkle_category_unit_test.cpp
```diff
@@ -160,7 +160,9 @@ class block_merkle_category : public Test {
 
   void deleteGenesisBlock(BlockId block_id, const BlockMerkleOutput &out) {
     auto batch = db->getBatch();
-    cat.deleteGenesisBlock(block_id, out, batch);
+    concord::kvbc::categorization::detail::LocalWriteBatch loclBatch;
```
dartdart26, line 163: Nit: I think it should be named `loc_batch`.

kvbc/test/categorization/versioned_kv_category_unit_test.cpp
```diff
@@ -590,7 +590,9 @@ TEST_F(versioned_kv_category, delete_genesis_with_update) {
   // Delete genesis block 1.
   {
     auto batch = db->getBatch();
-    cat.deleteGenesisBlock(1, out1, batch);
+    concord::kvbc::categorization::detail::LocalWriteBatch loclBatch;
```
dartdart26, line 593: Nit: I think it should be named `loc_batch`. Applies to other places too.

dartdart26 (approved):
Great job and performance improvement! I've left minor comments only.

kvbc/src/categorization/kv_blockchain.cpp
cloudnoize: no rationale :) probably got confused with the use of the std::ref needed for the thread pool.

kvbc/include/categorization/details.h
cloudnoize: thanks!

PR #2222, opened 2022-01-13
Add RSA verification code for db checkpoint interity

nkumar04:
On stable bft-checkpoint, replicas exchange checkpoint messages which are signed by their rsa private keys. Add rsa verification code for db-integrity check

### DIFF
--- a/bftengine/src/bftengine/messages/CheckpointMsg.hpp
+++ b/bftengine/src/bftengine/messages/CheckpointMsg.hpp
@@ -60,6 +60,7 @@ class CheckpointMsg : public MessageBase {
     return (a->seqNumber() == b->seqNumber()) && (a->digestOfState() == b->digestOfState()) &&
            (a->otherDigest() == b->otherDigest()) && (a->state() == b->state());
   }
+  inline size_t getHeaderLen() const { return sizeof(Header); }
 
  protected:
   template <typename MessageT>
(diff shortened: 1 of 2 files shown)


toly-kournik:
@nkumar04 , please consider verifying the checkpoint descriptor using the same mechanism ROReplica and IntegrityChecker do:

`  // no self certificate; static to avoid destruction of CheckpointMsg's during destruction of CheckpointInfo
  static std::map<SeqNum, bftEngine::impl::CheckpointInfo<false>> checkpointsInfo;
  checkpointsInfo.clear();
  for (auto m : desc.checkpointMsgs) 
    m->validate(*params_.repsInfo);
  checkpointsInfo[m->seqNumber()].addCheckpointMsg(m, m->idOfGeneratedReplica());
  if (checkpointsInfo[m->seqNumber()].isCheckpointCertificateComplete()) {
      LOG_INFO(GL, "Checkpoint descriptor is valid for block " << m->state());
      return;
  }
  throw std::runtime_error("Checkpoint descriptor is not valid for block " +
                           std::to_string(desc.checkpointMsgs[0]->state()));`

yontyon (approved):
LGTM!
Thanks

itsmeakashgoyal (approved):
LGTM.
Thanks Nikhil

PR #2223, opened 2022-01-13
Fix ASAN heap-use-after-free on post-exec handle of deferred messages

Liorgal28:


HristoStaykov (approved):
LGTM, thanks, @Liorgal28 !

PR #2225, opened 2022-01-17
Add support for the StateSnapshotRequest

dartdart26:
The StateSnapshotRequest is sent by Clientservice to the replicas to
create a state snapshot and get:
 * the snapshot identifier
 * the event group ID at which the state snapshot is created

On receiving StateSnapshotRequest, we first check if there is an
existing checkpoint. If yes, we use it. If not, we create a new one
asynchronously.

When creating a checkpoint asynchronously, the StateSnapshotRequest
might arrive at block ID = P. However, since the snapshot is done in a
separate thread, the actual block ID might be block ID = Q, where Q > P.
In order to make the snapshots equivalent on all replicas, we introduce
a callback in DbCheckpointManager that prepares the checkpoint for use.
In that case, it will delete all last reachable blocks from Q back to P,
leaving all replicas' checkpoints with last reachable block ID = P.

Currently, we only support snapshots for new participants, meaning that
we will only focus on streaming public state.

### DIFF
--- a/bftengine/include/bftengine/DbCheckpointManager.hpp
+++ b/bftengine/include/bftengine/DbCheckpointManager.hpp
@@ -64,11 +64,19 @@ class DbCheckpointManager {
    ***************************************************/
   std::optional<DbCheckpointMetadata::DbCheckPointDescriptor> getLastCreatedDbCheckpointMetadata();
 
+  // A callback that is called after creating a checkpoint. Its purpose is to prepare it for use. It receives as
+  // parameters:
+  //  * the block ID at which the checkpoint was created (it might be less than the last block ID in the checkpoint
+  //  itself)
+  //  * the path to the checkpoint
+  using PrepareCheckpointCallback = std::function<void(BlockId, const std::string&)>;
+
   Seconds getLastCheckpointCreationTime() const { return lastCheckpointCreationTime_; }
   void initializeDbCheckpointManager(std::shared_ptr<concord::storage::IDBClient> dbClient,
                                      std::shared_ptr<bftEngine::impl::PersistentStorage> p,
                                      std::shared_ptr<concordMetrics::Aggregator> aggregator,
-                                     const std::function<BlockId()>& getLastBlockIdCb);
+                                     const std::function<BlockId()>& getLastBlockIdCb,
+                                     const PrepareCheckpointCallback& prepareCheckpointCb);
   std::map<CheckpointId, DbCheckpointMetadata::DbCheckPointDescriptor> getListOfDbCheckpoints() const {
     return dbCheckptMetadata_.dbCheckPoints_;
   }
@@ -117,6 +125,7 @@ class DbCheckpointManager {
   void removeAllCheckpoints() const;
   void cleanUp();
   std::function<BlockId()> getLastBlockIdCb_;
+  PrepareCheckpointCallback prepareCheckpointCb_;
   // get total size recursively
   uint64_t directorySize(const _fs::path& directory, const bool& excludeHardLinks, bool recursive);
   // get checkpoint metadata
--- a/kvbc/src/Replica.cpp
+++ b/kvbc/src/Replica.cpp
@@ -329,9 +329,9 @@ void Replica::createReplicaAndSyncState() {
       m_replicaPtr->persistentStorage(),
       aggregator_,
       [this]() -> uint64_t { return getLastBlockId(); },
-      [](BlockId blockIdAtCheckpoint, const std::string &path) {
-        if (blockIdAtCheckpoint < INITIAL_GENESIS_BLOCK_ID) {
-          LOG_INFO(GL, "DB checkpoint done at block ID = " << blockIdAtCheckpoint << ", no need to prepare it");
+      [](BlockId block_id_at_checkpoint, const std::string &path) {
+        if (block_id_at_checkpoint < INITIAL_GENESIS_BLOCK_ID) {
+          LOG_INFO(GL, "DB checkpoint done at block ID = " << block_id_at_checkpoint << ", no need to prepare it");
           return;
         }
         const auto read_only = false;
@@ -339,7 +339,7 @@ void Replica::createReplicaAndSyncState() {
         auto db = storage::rocksdb::NativeClient::newClient(
             path, read_only, storage::rocksdb::NativeClient::DefaultOptions{});
         auto kvbc = categorization::KeyValueBlockchain{db, link_st_chain};
-        while (blockIdAtCheckpoint < kvbc.getLastReachableBlockId()) {
+        while (block_id_at_checkpoint < kvbc.getLastReachableBlockId()) {
           LOG_INFO(
               GL, "Deleting last reachable block = " << kvbc.getLastReachableBlockId() << ", DB checkpoint = " << path);
           kvbc.deleteLastReachableBlock();
--- a/kvbc/src/Replica.cpp
+++ b/kvbc/src/Replica.cpp
@@ -325,8 +325,25 @@ void Replica::createReplicaAndSyncState() {
   handleNewEpochEvent();
   handleWedgeEvent();
   bftEngine::impl::DbCheckpointManager::instance().initializeDbCheckpointManager(
-      m_dbSet.dataDBClient, m_replicaPtr->persistentStorage(), aggregator_, [this]() -> uint64_t {
-        return getLastBlockId();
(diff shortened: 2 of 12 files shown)


bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -64,11 +64,19 @@ class DbCheckpointManager {
...
+  //  itself)
+  //  * the path to the checkpoint
+  using PrepareCheckpointCallback = std::function<void(BlockId, const std::string&)>;
+
   Seconds getLastCheckpointCreationTime() const { return lastCheckpointCreationTime_; }
   void initializeDbCheckpointManager(std::shared_ptr<concord::storage::IDBClient> dbClient,
                                      std::shared_ptr<bftEngine::impl::PersistentStorage> p,
                                      std::shared_ptr<concordMetrics::Aggregator> aggregator,
-                                     const std::function<BlockId()>& getLastBlockIdCb);
+                                     const std::function<BlockId()>& getLastBlockIdCb,
```
arc-vmware, line 78: Passing lambda is a good idea, but for the same class, a private member function is preferable. Lambda for lambda calculus is better. If we have a thread that will use the lambda and execute later or if we require lambda to build concepts (c++20) then putting lambda is good. I am not saying to remove the lambda,  instead to see if we will require this function only once or whether this function will be required to be used at multiple places. If there is a chance to use the function at multiple places, then we can move the same as a private function.

kvbc/src/Replica.cpp
```diff
@@ -323,8 +323,25 @@ void Replica::createReplicaAndSyncState() {
   handleNewEpochEvent();
   handleWedgeEvent();
   bftEngine::impl::DbCheckpointManager::instance().initializeDbCheckpointManager(
-      m_dbSet.dataDBClient, m_replicaPtr->persistentStorage(), aggregator_, [this]() -> uint64_t {
-        return getLastBlockId();
+      m_dbSet.dataDBClient,
+      m_replicaPtr->persistentStorage(),
+      aggregator_,
+      [this]() -> uint64_t { return getLastBlockId(); },
+      [](BlockId blockIdAtCheckpoint, const std::string &path) {
```
arc-vmware, line 330: blockIdAtCheckpoint name should be block_id_at_checkpoint

kvbc/src/Replica.cpp
```diff
@@ -323,8 +323,25 @@ void Replica::createReplicaAndSyncState() {
...
+        if (blockIdAtCheckpoint < INITIAL_GENESIS_BLOCK_ID) {
```
arc-vmware, line 331: The logic of checking the block id is given in deleteLastReachableBlock() function. If the block id is wrong then the function is throwing logic_error exception. It's better to catch the exception and return, instead of rechecking all the conditions before calling the function.

storage/include/storage/db_interface.h
```diff
@@ -76,6 +77,7 @@ class IDBClient {
   virtual std::string getPath() const { return ""; };
   virtual void setCheckpointPath(const std::string&){};
   virtual std::string getCheckpointPath() const { return "NotSupported"; };
+  virtual std::string getPathForCheckpoint(std::uint64_t checkpointId) const { ConcordAssert(false); };
```
arc-vmware, line 80: checkpoint_id

storage/src/rocksdb_client.cpp
```diff
@@ -578,6 +578,10 @@ Status Client::rangeDel(const Sliver &_beginKey, const Sliver &_endKey) {
   return Status::OK();
 }
 
+std::string Client::getPathForCheckpoint(std::uint64_t checkpointId) const {
+  return (fs::path{dbCheckpointPath_} / std::to_string(checkpointId)).string();
```
arc-vmware, line 582: db_checkpoint_path

storage/src/rocksdb_client.cpp
```diff
@@ -592,7 +596,7 @@ Status Client::createCheckpoint(const uint64_t &checkPointId) {
     if (!fs::exists(path)) {
       fs::create_directory(path);
     }
-    fs::path chkptDirPath = path / std::to_string(checkPointId);
+    fs::path chkptDirPath = getPathForCheckpoint(checkPointId);
```
arc-vmware, line 599: chkpt_dir_path

arc-vmware (commented):
Thanks, @dartdart26 for the changes. I have very few comments.

kvbc/src/Replica.cpp
nkumar04, line 331: I am not sure if understood this. we call prepareCheckpointCb_ with lastReachableBlock in the blockchain. So, in which case, we may hit this condition? won't this if condition always  be false. may be i am missing something

reconfiguration/src/reconfiguration_handler.cpp
```diff
@@ -237,6 +237,52 @@ bool ReconfigurationHandler::handle(const concord::messages::CreateDbCheckpointC
...
+  // the first one, namely 0.
+  if (last_checkpoint_desc) {
+    resp.data.emplace();
+    resp.data->snapshot_id = last_checkpoint_desc->checkPointId_;
+    resp.data->event_group_id = 0;
+    LOG_INFO(getLogger(),
+             "StateSnapshotRequest(participant ID = " << cmd.participant_id << "): using existing last checkpoint ID: "
+                                                      << last_checkpoint_desc->checkPointId_);
+  } else {
+    const auto checkpoint_id = DbCheckpointManager::instance().createDbCheckpointAsync(sequence_number, timestamp);
```
nkumar04, line 265: note: DbCheckpointManager does not create db checkpoint if `ReplicaConfig::maxNumberOfDbCheckpoints == 0`
We should check this at the start along with feature flag and return error in response

nkumar04 (approved):
LGTM!
only few minor comments.

Thanks @dartdart26

bftengine/include/bftengine/DbCheckpointManager.hpp
dartdart26: The idea behind the callback is that the code inside of it requires access to the `kvbc` library. Yet, we don't want to make the `corebft` library depend on `kvbc`. Using a callback allows us to remove this dependency.

kvbc/src/Replica.cpp
dartdart26: Thanks, fill fix.

kvbc/src/Replica.cpp
dartdart26: You are right, but we are checking the input only here, not any current kvbc state. Does that make sense?

storage/include/storage/db_interface.h
dartdart26: You are right, but I have intentionally kept camelCase, because this whole file uses it. I think it would be better to fix it for the whole file in a separate commit.

storage/src/rocksdb_client.cpp
dartdart26: Same as above - will keep for consistency in that file for now.

storage/src/rocksdb_client.cpp
dartdart26: Same as above - will keep for consistency in that file for now.

kvbc/src/Replica.cpp
dartdart26: INITIAL_GENESIS_BLOCK_ID is a constant that is currently set to 1. The check handles a case where a checkpoint is created at block 0 (if done manually). Does that make sense?

reconfiguration/src/reconfiguration_handler.cpp
dartdart26: I see, will do so, thanks. I am wondering if we need both the flag and the number of checkpoints in that case, though. Do we have a use case for setting the max number to a non-zero value, yet having the flag set to false?

arc-vmware (approved):
Thanks, @dartdart26 , LGTM

reconfiguration/src/reconfiguration_handler.cpp
nkumar04: feature_flag = True && num_of_db_checkpoints = 0; //possible, for example we create db checkpoints on (f+1) replicas only. But all or at-least 2f+1 replicas participate in consensus to decide the seq num to initiate creation of db checkpoint at and that will be based on the flag

dartdart26: Got it, thanks!

PR #2227, opened 2022-01-18
Introduce RawMemoryPool for PreProcessor

yuliasherman:
Use a memory pool in PreProcessor instead of buffers pre-allocated per client request. The buffers could be re-used by any client and not connected to any specific one. The pool minimum/maximum size and the size of the allocated chunks are configurable. The pool is "breathing" - a minimum number of chunks is allocated in the beginning and it could grow up to the maximum. In case the chunks are not in use, they get deleted.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -29,7 +29,7 @@ using namespace std;
 using namespace std::placeholders;
 using namespace concordUtil;
 
-uint8_t RequestState::reqProcessingHistoryHeight = 10;
+uint8_t RequestState::reqProcessingHistoryHeight = 1;
 
 //**************** Class RequestsBatch ****************//
 
@@ -272,7 +272,10 @@ void PreProcessor::addNewPreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunic
 
 void PreProcessor::setAggregator(std::shared_ptr<concordMetrics::Aggregator> aggregator) {
   if (ReplicaConfig::instance().getpreExecutionFeatureEnabled() && aggregator) {
-    for (const auto &elem : preProcessors_) elem->metricsComponent_.SetAggregator(aggregator);
+    for (const auto &elem : preProcessors_) {
+      elem->metricsComponent_.SetAggregator(aggregator);
+      elem->memoryPool_.setAggregator(aggregator);
+    }
   }
 }
 
@@ -305,6 +308,7 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
       numOfReplicas_(myReplica.getReplicaConfig().numReplicas + myReplica.getReplicaConfig().numRoReplicas),
       numOfInternalClients_(myReplica.getReplicaConfig().numOfClientProxies),
       clientBatchingEnabled_(myReplica.getReplicaConfig().clientBatchingEnabled),
+      memoryPool_(myReplica.getReplicaConfig().maxExternalMessageSize, timers),
       metricsComponent_{concordMetrics::Component("preProcessor", std::make_shared<concordMetrics::Aggregator>())},
       metricsLastDumpTime_(0),
       metricsDumpIntervalInSec_{myReplica_.getReplicaConfig().metricsDumpIntervalSeconds},
@@ -336,8 +340,10 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
   const uint16_t numOfReqEntries = numOfExternalClients * clientMaxBatchSize_;
   for (uint16_t i = 0; i < numOfReqEntries; i++) {
     // Placeholders for all clients including batches
-    preProcessResultBuffers_.emplace_back(std::make_pair(false, Sliver()));
+    preProcessResultBuffers_.emplace_back(make_shared<SafeResultBuffer>());
   }
+  // Initially, allocate a memory for all batches of one client (clientMaxBatchSize_)
+  memoryPool_.allocatePool(clientMaxBatchSize_, numOfReqEntries);
   const uint16_t firstClientId = numOfReplicas_ + numOfInternalClients_;
   for (uint16_t i = 0; i < numOfExternalClients; i++) {
     // Placeholders for all client batches
@@ -1344,6 +1350,7 @@ void PreProcessor::releaseClientPreProcessRequest(const RequestStateSharedPtr &r
     if (!myReplica_.isCurrentPrimary()) {
       preProcessorMetrics_.preProcInFlyRequestsNum--;
     }
+    releasePreProcessResultBuffer(clientId, reqSeqNum, reqOffsetInBatch);
   }
 }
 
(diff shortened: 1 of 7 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1457,30 +1464,34 @@ void PreProcessor::registerAndHandleClientPreProcessReqOnNonPrimary(const string
   }
 }
 
+uint32_t PreProcessor::getBufferOffset(uint16_t clientId, ReqId reqSeqNum, uint16_t reqOffsetInBatch) const {
+  return (clientId - numOfReplicas_ - numOfInternalClients_) * clientMaxBatchSize_ + reqOffsetInBatch;
```
cloudnoize, line 1468: maybe it's worth asserting that ```(clientId - numOfReplicas_ - numOfInternalClients_)``` is not negative

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1457,30 +1464,34 @@ void PreProcessor::registerAndHandleClientPreProcessReqOnNonPrimary(const string
...
-    {
-      std::unique_lock lock(resultBufferLock_);
-      if (!preProcessResultBuffers_[bufferOffset].first) {
-        preProcessResultBuffers_[bufferOffset].second = Sliver(buf, maxPreExecResultSize_);
-        preProcessResultBuffers_[bufferOffset].first = true;
-      } else
-        delete[] buf;
-    }
+  const auto bufferOffset = getBufferOffset(clientId, reqSeqNum, reqOffsetInBatch);
+  std::unique_lock lock(preProcessResultBuffers_[bufferOffset]->mutex);
```
cloudnoize, line 1479: If I understand, the mutex protects the buffer, but on the other hand, the method returns a pointer to the buffer i.e. the buffer is not protected and can be accessed concurrently, no?

util/src/RawMemoryPool.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+char* RawMemoryPool::allocateChunk() {
+  char* chunk = nullptr;
+  if (numOfAllocatedChunks_ < maxChunksNum_) {
+    chunk = new char[chunkSize_];
+    increaseNumOfAllocatedChunks();
+    increaseNumOfAvailableChunks();
+  }
+  if (chunk) LOG_DEBUG(logger(), "A chunk has been allocated" << KVLOG(numOfAllocatedChunks_, numOfAvailableChunks_));
```
cloudnoize, line 86: I think it's nice also to have log in case maxChunksNum_ is reached and a chunk wasn't allocated

util/src/RawMemoryPool.cpp
```diff
@@ -0,0 +1,142 @@
...
+  ConcordAssert(pool_.get() != nullptr);
+  char* chunk = nullptr;
+  if (!pool_->pop(chunk)) {
+    // No available chunks => allocate a new one, if permitted
+    chunk = allocateChunk();
+    if (!chunk) {
+      // Pool size limit has been reached; wait until some chunk gets released
+      unique_lock<mutex> lock(waitForAvailChunkLock_);
+      while (!stopWorking_ && !numOfAvailableChunks_)
+        waitForAvailChunkCond_.wait_until(lock, steady_clock::now() + milliseconds(WAIT_TIMEOUT_MILLI));
```
cloudnoize, line 118: shouldn't this thread need to pop again after the notification? 
maybe the while should be something like this
```
while (!stopWorking_ && !chunk){
        waitForAvailChunkCond_.wait_until(lock, steady_clock::now() + milliseconds(WAIT_TIMEOUT_MILLI));
         pool_->pop(chunk);
}
````
wdyt?

File: util/include/RawMemoryPool.hpp [hunk 5]
```diff
@@ -0,0 +1,115 @@
...
+  RawMemoryPool(uint32_t chunkSize, Timers& timers);
+  virtual ~RawMemoryPool();
+
+  void allocatePool(int32_t initialChunksNum, int32_t maxChunksNum);
+  int32_t getNumOfAvailableChunks() const { return numOfAvailableChunks_; }
+  int32_t getNumOfAllocatedChunks() const { return numOfAllocatedChunks_; }
+  bool isPoolFull() const { return numOfAllocatedChunks_ == maxChunksNum_; }
+  char* getChunk();
+  void returnChunk(char*);
+  void stopPool() { stopWorking_ = true; }
```
cloudnoize, line 74: who can call this? under which scenario?

util/src/RawMemoryPool.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+void RawMemoryPool::returnChunk(char* chunk) {
+  if (numOfAvailableChunks_ >= maxChunksNum_) {
+    stringstream err;
+    err << "Returned chunk overflows the pool size limit: numOfAvailableChunks_=" << numOfAvailableChunks_
+        << ", maxChunksNum_=" << maxChunksNum_;
+    LOG_ERROR(logger(), KVLOG(err.str()));
+    throw std::runtime_error(__PRETTY_FUNCTION__ + err.str());
+  }
+  if (!isPoolPruningRequired())
```
cloudnoize, line 136: nice logic:)

util/include/RawMemoryPool.hpp
yuliasherman: Needs to be called from PreProcessor before going down. Will fix. Nice catch!

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: OK

util/src/RawMemoryPool.cpp
yuliasherman: OK

util/src/RawMemoryPool.cpp
yuliasherman: Agree

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: We need to protect only the buffer allocation. I cannot think of a scenario when we modify the same buffer from different threads.

yuliasherman:
Eran, thanks for the great review!

util/include/RawMemoryPool.hpp
```diff
@@ -0,0 +1,114 @@
...
+// This thread-safe memory pool implementation is intended for the raw memory chunks operations. During the pool
+// initialization, a pre-defined number (minChunksNum_) of raw memory chunks of size chunkSize_ gets allocated.
+// In case there is no available chunk, a new one gets dynamically allocated - up to a maximum number (maxChunksNum_).
+// No support for different memory sizes provided.
+
+namespace concordUtil {
+
+class RawMemoryPoolMetrics {
+ public:
+  RawMemoryPoolMetrics(Timers& timers)
```
Liorgal28, line 32: Maybe you can add also a metric to measure the time the chunk was used, it will help to analyze and optimize WAIT_TIMEOUT_MILLI.

I know it can be tricky so it's not a must metric.

util/include/RawMemoryPool.hpp
```diff
@@ -0,0 +1,114 @@
...
+  void increaseNumOfAllocatedChunks();
+  void increaseNumOfAvailableChunks();
+  void decreaseNumOfAllocatedChunks();
+  void decreaseNumOfAvailableChunks();
+  char* allocateChunk();
+  void returnChunkToThePool(char* chunk);
+  void deleteChunk(char*& chunk);
+
+ private:
+  const uint32_t WAIT_TIMEOUT_MILLI = 5;
```
Liorgal28, line 100: Feels too short on the assumption that only the preprocessing takes 40ms.

util/src/RawMemoryPool.cpp
```diff
@@ -0,0 +1,148 @@
...
+           (numOfAvailableChunks_ >= numOfAllocatedChunks_ * maxAvailableChunksPercentage_)));
+}
+
+char* RawMemoryPool::getChunk() {
+  ConcordAssert(pool_.get() != nullptr);
+  char* chunk = nullptr;
+  if (!pool_->pop(chunk)) {
+    // No available chunks => allocate a new one, if permitted
+    chunk = allocateChunk();
+    if (!chunk) {
```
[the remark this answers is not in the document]
yuliasherman: As Eran asked for, I've added a log when we reach the maximum.

util/src/RawMemoryPool.cpp
```diff
@@ -0,0 +1,148 @@
...
+    stringstream err;
+    err << "Returned chunk overflows the pool size limit: numOfAvailableChunks_=" << numOfAvailableChunks_
+        << ", maxChunksNum_=" << maxChunksNum_;
+    LOG_ERROR(logger(), KVLOG(err.str()));
+    throw std::runtime_error(__PRETTY_FUNCTION__ + err.str());
+  }
+  if (!isPoolPruningRequired())
+    returnChunkToThePool(chunk);
+  else  // Too many chunks are not in use => delete
+    deleteChunk(chunk);
```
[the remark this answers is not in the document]
yuliasherman: Yes, but I intentionally don't like to take strong decisions based on just one check (maybe we are going to have a peak starting from the next request?). Besides that, it could be time-consuming doing many delete operations before returning the chunk to the caller.

util/include/RawMemoryPool.hpp
yuliasherman: I think it's OK because we are not waiting for any specific request to complete. There are many requests on the fly - we use the same pool for all PreProcessor requests and any moment one of them could complete.

util/include/RawMemoryPool.hpp
yuliasherman: The pool should be really lightweight. Even the pruning logic I made a very basic to avoid additional locks. And I also don't think this metric will help us to tune WAIT_TIMEOUT_MILLI.

cloudnoize (approved):
Looks great, thanks @yuliasherman !

Liorgal28 (approved):

PR #2228, opened 2022-01-19
Wait for stable seq number to create db checkpoint via operator cmd

nkumar04:
When db checkpoint request is initiated by operator, we create db checkpoint as soon as we reach stable seqNum boundary. But we need to wait for seqNum to be stable so that, (n-f) checkpoint messages are stored in persistence, these messages are used to prove the integrity of db checkpoint directory
Improvements in current apollo tests


itsmeakashgoyal (approved):
LGTM.
Thanks Nikhil

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5293,15 +5293,18 @@ void ReplicaImp::onExecutionFinish() {
...
       DbCheckpointManager::instance().isCreateDbCheckPtSeqNumSet(lastExecutedSeqNum)) {
-    bool noop =
-        (DbCheckpointManager::instance().getNextStableSeqNumToCreateSnapshot().value() != lastExecutedSeqNum + 1);
-    DbCheckpointManager::instance().sendInternalCreateDbCheckpointMsg(
-        lastExecutedSeqNum + 1,
-        noop);  // send noop till we reach stable checkpoint
-    if (!noop) DbCheckpointManager::instance().setNextStableSeqNumToCreateSnapshot(std::nullopt);
-    LOG_INFO(GL, "sendInternalCreateDbCheckpointMsg noop command ");
+    DbCheckpointManager::instance().sendInternalCreateDbCheckpointMsg(lastExecutedSeqNum + 1, true);  // noop=true
+    if (DbCheckpointManager::instance().getNextStableSeqNumToCreateSnapshot().value() == lastExecutedSeqNum + 1) {
```
arc-vmware, line 5297: I think we should check if this should be lastExecutedSeqNum + 1 or it should be lastExecutedSeqNum + activeExecutions_, since these many post executions might be happening in parallel.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5293,15 +5293,18 @@ void ReplicaImp::onExecutionFinish() {
...
+      auto seq_num_to_create_dbcheckpoint = lastExecutedSeqNum + 1;
```
arc-vmware, line 5298: Same here

arc-vmware (approved):
LGTM!!!

PR #2231, opened 2022-01-19
EventGroups: Fix record time forwarding

robem:
When we filter events, we create a new event group object but forget to also
populate the record time. This change fixes it.


shgandhi (approved):
Thanks for catching and fixing!

PR #2233, opened 2022-01-20
Stability improvement for apollo test_skvbc_fast_path

evdzhurov:
* Fix test_fast_to_slow_path_transition to crash C+1 replicas instead of 1 replica to test the transition to slow path.
* test_fast_path_resilience_to_crashes doesn't need to test for the slow path to be prevalent as an intermediate step - the fast/slow path controller doesn't switch logically to the slow path but uses the slow path as a mechanism to downgrade from optimistic fast (n of n replicas) to fast with threshold commit path. This might be improved in the future and should not be captured by a test if it's not needed. 
* Remove dead code in test_skvbc_fast_path that was copy-pasted around in other tests.

### DIFF
--- a/tests/apollo/test_skvbc_fast_path.py
+++ b/tests/apollo/test_skvbc_fast_path.py
@@ -40,12 +40,6 @@ class SkvbcFastPathTest(unittest.TestCase):
 
     __test__ = False  # so that PyTest ignores this test scenario
 
-    def setUp(self):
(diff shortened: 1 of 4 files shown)


teoparvanov (approved):
LGTM, thanks @evdzhurov !

KTeofanidis (approved):
LGTM

PR #2234, opened 2022-01-20
Compute the public state hash on DB checkpoint

dartdart26:
When a DB checkpoint is created, compute the public state hash as part
of its preparation. Do that in the separate checkpoint thread,
sequentially, after the checkpoint itself is created. If it takes too
long, a future commit can change that behaviour such that the checkpoint
is ready before the hash computation.

Add the DbCheckpointManager::getCheckpointState() method that allows
users to check the state of a checkpoint by its ID.

Introduce EBFTSubtype::PublicStateHashAtDbCheckpoint - a key that is
used to persist the current public state hash in the default column
family of the checkpoint.

Move checkpoint preparation code to KeyValueBlockchain and add unit
tests for its behaviour.

### DIFF
--- a/bftengine/src/bftengine/DbCheckpointManager.cpp
+++ b/bftengine/src/bftengine/DbCheckpointManager.cpp
@@ -181,10 +181,30 @@ void DbCheckpointManager::initializeDbCheckpointManager(std::shared_ptr<concord:
     cleanUpFuture_ = std::async(std::launch::async, [this]() { cleanUp(); });
   }
 }
+
 std::optional<DbCheckpointMetadata::DbCheckPointDescriptor> DbCheckpointManager::getLastCreatedDbCheckpointMetadata() {
   std::scoped_lock lock(lockLastDbCheckpointDesc_);
   return lastCreatedCheckpointMetadata_;
 }
+
+DbCheckpointManager::CheckpointState DbCheckpointManager::getCheckpointState(CheckpointId id) const {
+  {
+    auto lock = std::scoped_lock{lockDbCheckPtFuture_};
+    if (dbCreateCheckPtFuture_.count(id)) {
+      return CheckpointState::kPending;
+    }
+  }
+
+  {
+    auto lock = std::scoped_lock{lock_};
+    if (dbCheckptMetadata_.dbCheckPoints_.count(id)) {
+      return CheckpointState::kCreated;
+    }
+  }
+
+  return CheckpointState::kNonExistent;
+}
+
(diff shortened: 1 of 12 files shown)


bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -181,10 +181,30 @@ void DbCheckpointManager::initializeDbCheckpointManager(std::shared_ptr<concord:
...
 }
+
 std::optional<DbCheckpointMetadata::DbCheckPointDescriptor> DbCheckpointManager::getLastCreatedDbCheckpointMetadata() {
   std::scoped_lock lock(lockLastDbCheckpointDesc_);
   return lastCreatedCheckpointMetadata_;
 }
+
+DbCheckpointManager::CheckpointState DbCheckpointManager::getCheckpointState(CheckpointId id) const {
+  {
+    auto lock = std::scoped_lock{lockDbCheckPtFuture_};
```
nkumar04, line 192: Should we check for if "(dbCheckptMetadata_.dbCheckPoints_.count(id))" first. Because "(dbCreateCheckPtFuture_.count(id))" will be equal to 1 in both cases pending and created

nkumar04 (approved):
LGTM!!

bftengine/src/bftengine/DbCheckpointManager.cpp
dartdart26: Yes, you are right. I was assuming we remove the future when we complete creating the checkpoint, but we actually remove it when we remove the oldest checkpoint. Let me fix it.

dartdart26: I am wondering whether I should just change the order of locks or whether I should keep `lock_` while locking `lockDbCheckPtFuture_? Based on the usage, it seems when we change the futures map, we always hold both locks anyway.

storage/include/storage/merkle_tree_key_manipulator.h
```diff
@@ -14,6 +14,9 @@
 #pragma once
 
 #include "key_manipulator_interface.h"
+#include "string.hpp"
```
arc-vmware, line 17: remove this as string is included

arc-vmware (approved):
LGTM, with a minor comment.

storage/include/storage/merkle_tree_key_manipulator.h
dartdart26: It is needed for `toChar()`.

PR #2236, opened 2022-01-21
Docker update to version 0.38 for clientservice support to Apollo

shgandhi:
This PR updates concord-bft docker image to version 0.38. This is because install_deps.sh has been updated with gRPC, protobuf and python3-dev dependencies. These dependencies are required to allow apollo test cases to communicate with clientservice over gRPC.
This PR also adds python bindings for clientservice to Apollo, which will be used by the test cases to invoke request and event service APIs. Apollo framework changes (bft.py) and basic clientservice apollo tests will be added in a separate PR.


teoparvanov (changes requested):
Hi Shruti, I went quickly through the proposed changes. At first sight I noticed quite a bit of code duplication where the client service is being introduced (bft configs, decorators, etc). I think this is because the current approach doesn't fit very well with the existing Apollo abstractions. I'll ping you on Slack a bit later, so that we discuss the options on your Friday morning if it's OK with you.

teoparvanov (approved):
LGTM, thanks Shruti!

PR #2238, opened 2022-01-21
Create db snapshot on replicas wedge with apollo tests

nkumar04:
There are two changes in this PR:
1. We create a db-snapshot when we wedge the replicas. Reason: wedge is performed for some maintainance, scaling, reconfiguration etc. So, we create a snapshot of the database, when we stop/wedge the replicas

2. On stable checkpoint, primary replicas send internal bft request to create db-checkpoint. However, other replicas may not have reached stable checkpoint when they get the request from primary. So, if a replica is yet to reach the sable checkpoint, it just adds a callback to create db-checkpoint on stable sequence number, This ensures that, we capture bft proof of n-f checkpoint messages in the rocksdb checkpoint

### DIFF
--- a/bftengine/include/bftengine/DbCheckpointManager.hpp
+++ b/bftengine/include/bftengine/DbCheckpointManager.hpp
@@ -101,6 +101,11 @@ class DbCheckpointManager {
     if (monitorThread_.joinable()) monitorThread_.join();
   }
   void sendInternalCreateDbCheckpointMsg(const SeqNum& seqNum, bool noop);
+  SeqNum getLastStableSeqNum() const;
+  void setOnStableSeqNumCb_(std::function<void(SeqNum)> cb) { onStableSeqNumCb_ = cb; }
+  void onStableSeqNum(SeqNum s) {
+    if (onStableSeqNumCb_) onStableSeqNumCb_(s);
+  }
 
  private:
   logging::Logger getLogger() {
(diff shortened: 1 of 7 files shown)


bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -115,6 +115,11 @@ class DbCheckpointManager {
     if (monitorThread_.joinable()) monitorThread_.join();
   }
   void sendInternalCreateDbCheckpointMsg(const SeqNum& seqNum, bool noop);
+  SeqNum getLastStableSeqNum() const;
+  void setOnStableSeqNumCb_(std::function<void(SeqNum)> cb) { onStableSeqNumCb_ = cb; }
+  void onStableSeqNum(SeqNum s) {
```
dartdart26, line 120: Do we call this method (I don't see it)?

dartdart26: Actually, we only call it in the constructor. Could you please explain the rationale?

dartdart26: I was looking in the wrong place, please ignore.

dartdart26 (approved):

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -117,8 +117,24 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
...
+        const auto& lastStableSeqNum = DbCheckpointManager::instance().getLastStableSeqNum();
+        if (lastStableSeqNum == static_cast<SeqNum>(createDbChkPtMsg.seqNum)) {
+          DbCheckpointManager::instance().createDbCheckpointAsync(createDbChkPtMsg.seqNum, timestamp, std::nullopt);
+        } else {
+          // this replica has not reached stable seqNum yet to create snapshot at requested seqNum
+          // add a callback to be called when seqNum is stable. We need to create snapshot on stable
+          // seq num because checkpoint msg certificate is stored on stable seq num and is used for intergrity
+          // check of db snapshots
+          const auto& seqNumToCreateSanpshot = createDbChkPtMsg.seqNum;
+          std::optional snapshotId(DbCheckpointManager::instance().getLastReachableBlock());
```
dartdart26, line 130: I think this variable can be called `block_id` instead.

dartdart26:
Looks good, thanks!

bftengine/src/bftengine/RequestHandler.cpp
nkumar04: fixed. Thanks

PR #2239, opened 2022-01-23
concord apollo tests failure fix

situ-s:
Apollo tests were failing as the response size was 0. The req.outExecution status was passed as success from client. As the value was already success, the request was not executed, hence sending reply size zero, causing test failure.

### DIFF
--- a/bftclient/src/bft_client.cpp
+++ b/bftclient/src/bft_client.cpp
@@ -98,6 +98,9 @@ Msg Client::createClientMsg(const RequestConfig& config, Msg&& request, bool rea
   header->requestLength = request.size();
   header->timeoutMilli = config.timeout.count();
   header->cidLength = config.correlation_id.size();
+  header->result = 1;  // UNKNOWN
+  header->reqSignatureLength = 0;
+  header->extraDataLength = 0;
 
   auto* position = msg.data() + header_size;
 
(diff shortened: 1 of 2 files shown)


bftclient/src/bft_client.cpp
```diff
@@ -98,6 +98,7 @@ Msg Client::createClientMsg(const RequestConfig& config, Msg&& request, bool rea
   header->requestLength = request.size();
   header->timeoutMilli = config.timeout.count();
   header->cidLength = config.correlation_id.size();
+  header->result = 1;  // UNKNOWN
```
yuliasherman, line 101: In case it is set to 1 in the ClientRequestMsgHeader, I think there is no need to set it again here.

situ-s: Yeah, I thought by default it should be UNKNOWN, but it was always taking SUCCESS. So had to change it here also.

yuliasherman: The basic code is written in an error-prone way. Because of:
ClientRequestMsgHeader* header = reinterpret_cast<ClientRequestMsgHeader*>(msg.data());
the header object does not get created and thus its initialization does not happen. Actually, we need to initialize here also all other header fields.

yuliasherman (approved):
LGTM!

PR #2240, opened 2022-01-23
Implement the resource entity interface

cloudnoize:
The ReplicaResourceEntity class measures some metrics that should give us an indication of the rate the system can perform pruning.
A user of this class should use those measurements to define a strategy.
The implemented measurements:

- Pruning avg time - the avg time in micro-sec it takes to prune a block.
- Pruning utilization - the percentage that the system was performing pruning from a certain duration.
- Post execution utilization -  the percentage that the system was performing post-execution from a certain duration.
- bft queue size - the avg size of the pending requests.
- Block addition
- Count of consensus reached

In order to measure utilization, a helper class is implemented as part of this PR.
its algorithm is to operate on a vector of duration time measurements, where each measurement is a pair of {start,end}
then it calculates from the total time i.e. last_measure.end - first_measure.start, how much time the operation was busy, 
and return the percentage,


kvbc/src/replica_resources.cpp
```diff
@@ -0,0 +1,59 @@
...
+    ++pruned_blocks_;
+  }
+  if (m.type == ISystemResourceEntity::type::consensus_throughput) {
+    numOfConsensus_ += m.count;
+  }
+  numOfExternalMsgs_ += IncomingMsgsStorageImp::num_of_external_msgs_;
+  ++numOfExternalMsgsCalls_;
+}
+
+uint64_t ReplicaResourceEntity::getMeasurement(const ISystemResourceEntity::type type) const {
```
titusen, line 31: This switch should be replaced with polymorphism.

resources-manager/include/ISystemResourceEntity.hpp
```diff
@@ -14,18 +14,41 @@
 
 #include <cstdint>
 #include <string>
+#include <vector>
+#include "utilization.hpp"
 
 namespace concord::performance {
 class ISystemResourceEntity {
  public:
+  // ps - per second
```
titusen, line 23: Implementations of ISystemResourceEntity should cover these scenarios and return measurements.

kvbc/include/replica_resources.h
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 VMware, all rights reserved
+
+#pragma once
+
+#include "ISystemResourceEntity.hpp"
+#include "IncomingMsgsStorageImp.hpp"
+
+class ReplicaResourceEntity : public concord::performance::ISystemResourceEntity {
```
titusen, line 8: This ReplicaResourceEntity implementation is hijacking the responsibility of IResourceManager. Implementation of that class should take in a constructor all necessary measurements providers (IResourceEntity) and operate with them the way creator intents to. If needs e.g. distincted NetworkMonitorResourceEntity class A : public IResourseManager { A(IResourceEntity* networkMonitorResourceEntity , ... ) }

kvbc/src/replica_resources.cpp
cloudnoize: In my opinion, using polymorphism, in this case, will introduce a lot of boilerplate code and complexity

kvbc/include/replica_resources.h
cloudnoize: Sorry, I don't understand this sentence. 
As we discussed, I see it as a raw source for measurements that the user e.g. ResourceMansger (the strategy) can call and use as it likes

kvbc/src/replica_resources.cpp
cloudnoize: also, it makes the interface less vulnarable to changes, since adding a measurement type doesn't require changing the interface and all it's implementations.

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -149,7 +150,14 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
       req.outActualReplySize = 1;
     }
   }
-  if (userRequestsHandler_) return userRequestsHandler_->execute(requests, timestamp, batchCid, parent_span);
+  if (userRequestsHandler_) {
+    auto isPost = (requests.size() > 0 && !(requests.back().flags & bftEngine::PRE_PROCESS_FLAG));
+    ISystemResourceEntity::scopedDurMeasurment m(
```
yontyon, line 155: This is not accurate.
(1) you don't include here reconfiguration, key exchange, and ticks.
(2) you do include here read requests which are not going through a consensus (you need to filter them out) see https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L4638

yontyon (changes requested):
One comment on the actual code, please address.

Another general comment:
The calculation shouldn't happen in the hot paths of the code, as far as I could see this is the case, but just wanted to emphasize this point.
Also, I think that **all** measurements collections should be conditional with a flag in case we want to disable them to test if they hurt performance.

bftengine/src/bftengine/RequestHandler.cpp
cloudnoize: regarding 1 - I think that the internal requests are neglectable relative to the application requests, though I can make it more general.
Regarding 2- good point, I'll filter them.

cloudnoize: regarding the calculations - they are being triggered by the getMeasurement method, which will be called from another thread than the main thread. 
I'll add a flag on the addMeasurement in order to control measurement addition for performance evaluation.

yontyon (approved):

bftengine/src/bftengine/RequestHandler.cpp
yontyon: Thanks

PR #2245, opened 2022-01-24
Clientservice: Async request service

robem:
This change converts the request service handling from synchronous to
asynchronous. Meaning, instead of relying on the gRPC thread pool to handle
each incoming request synchronously, we now have each thread handle multiple
requests concurrently.
Each executing thread listens on its own "completion queue" which the gRPC
server fills with events - an incoming `Send` request is such an event. Each
request will walk through a state machine whereby the handling of each state is
triggered by an event from the completion queue. In between each state, the
executing thread will listen for further events and is thereby able to work on
multiple requests concurrently. Based on the gRPC documentation, one thread
handles exactly one completion queue. With more than one thread (which is
configurable) we can handle multiple requests in parallel.

### DIFF
--- a/client/clientservice/include/client/clientservice/client_service.hpp
+++ b/client/clientservice/include/client/clientservice/client_service.hpp
@@ -28,21 +28,30 @@ class ClientService {
       : logger_(logging::getLogger("concord.client.clientservice")),
         client_(std::move(client)),
         event_service_(std::make_unique<EventServiceImpl>(client_)),
-        request_service_(std::make_unique<RequestServiceImpl>(client_)),
         state_snapshot_service_(std::make_unique<StateSnapshotServiceImpl>(client_)){};
 
-  void start(const std::string& addr, uint64_t max_receive_msg_size);
+  void start(const std::string& addr, int num_async_threads, uint64_t max_receive_msg_size);
 
   const std::string kRequestService{"vmware.concord.client.request.v1.RequestService"};
   const std::string kEventService{"vmware.concord.client.event.v1.EventService"};
   const std::string kStateSnapshotService{"vmware.concord.client.statesnapshot.v1.StateSnapshotService"};
 
  private:
+  // Handler for asynchronous services
+  void handleRpcs(int thread_idx);
+
   logging::Logger logger_;
   std::shared_ptr<concord::client::concordclient::ConcordClient> client_;
+
+  // Synchronous services
   std::unique_ptr<EventServiceImpl> event_service_;
-  std::unique_ptr<RequestServiceImpl> request_service_;
   std::unique_ptr<StateSnapshotServiceImpl> state_snapshot_service_;
+
+  // Asynchronous services
+  vmware::concord::client::request::v1::RequestService::AsyncService request_service_;
+
+  std::vector<std::unique_ptr<grpc::ServerCompletionQueue>> cqs_;
+  std::vector<std::thread> server_threads_;
 };
 
 }  // namespace concord::client::clientservice
--- a/client/clientservice/src/client_service.cpp
+++ b/client/clientservice/src/client_service.cpp
@@ -15,22 +15,66 @@
 
 namespace concord::client::clientservice {
 
-void ClientService::start(const std::string& addr, uint64_t max_receive_msg_size) {
+void ClientService::start(const std::string& addr, int num_async_threads, uint64_t max_receive_msg_size) {
   grpc::EnableDefaultHealthCheckService(true);
 
   grpc::ServerBuilder builder;
   builder.AddListeningPort(addr, grpc::InsecureServerCredentials());
   builder.SetMaxReceiveMessageSize(max_receive_msg_size);
-  builder.RegisterService(request_service_.get());
+
+  // Register synchronous services
   builder.RegisterService(event_service_.get());
 
+  // Register asynchronous services
+  builder.RegisterService(&request_service_);
+
+  for (int i = 0; i < num_async_threads; ++i) {
+    cqs_.emplace_back(builder.AddCompletionQueue());
+  }
+
   auto clientservice_server = std::unique_ptr<grpc::Server>(builder.BuildAndStart());
+
+  // From the "C++ Performance Notes" in the gRPC documentation:
+  // "Right now, the best performance trade-off is having numcpu's threads and one completion queue per thread."
+  for (int i = 0; i < num_async_threads; ++i) {
+    server_threads_.emplace_back(std::thread([this, i] { this->handleRpcs(i); }));
+  }
 
   auto health = clientservice_server->GetHealthCheckService();
   health->SetServingStatus(kRequestService, true);
   health->SetServingStatus(kEventService, true);
 
   clientservice_server->Wait();
+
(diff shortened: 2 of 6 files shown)


client/clientservice/src/request_service.cpp
```diff
@@ -10,83 +10,105 @@
...
+namespace requestservice {
 
-namespace cc = concord::client::concordclient;
+void RequestServiceCallData::proceed() {
+  if (state_ == CREATE) {
+    state_ = SEND_TO_CONCORDCLIENT;
+    // Request to handle an incoming `Send` RPC -> will put an event on the cq if ready
+    service_->RequestSend(&ctx_, &request_, &responder_, cq_, cq_, this);
+  } else if (state_ == SEND_TO_CONCORDCLIENT) {
+    // We are handling an in coming `Send` right now, let's make sure we handle the next one too
```
yuliasherman, line 28: typo: in coming

yuliasherman (approved):
LGTM

client/clientservice/include/client/clientservice/client_service.hpp
```diff
@@ -28,21 +28,30 @@ class ClientService {
...
+
+  // Synchronous services
   std::unique_ptr<EventServiceImpl> event_service_;
-  std::unique_ptr<RequestServiceImpl> request_service_;
   std::unique_ptr<StateSnapshotServiceImpl> state_snapshot_service_;
+
+  // Asynchronous services
+  vmware::concord::client::request::v1::RequestService::AsyncService request_service_;
+
+  std::vector<std::unique_ptr<grpc::ServerCompletionQueue>> cq_;
```
shgandhi, line 53: Should we rename this to something like `cq_vector_`? because this is not a single completion queue.

shgandhi (approved):
I have a couple of minor comments.
Additionally, since the documentation calls `to shutdown the server prior to shutting down the returned completion queue.`, and we don't explictly shut down the clientservice, do you think that is something we need to add?

client/clientservice/src/main.cpp
```diff
@@ -51,6 +50,7 @@ po::variables_map parseCmdLine(int argc, char** argv) {
     ("config", po::value<std::string>()->required(), "YAML configuration file for the RequestService")
     ("host", po::value<std::string>()->default_value("0.0.0.0"), "Clientservice gRPC service host")
     ("port", po::value<int>()->default_value(50505), "Clientservice gRPC service port")
+    ("num-async-threads", po::value<int>()->default_value(1), "Number of threads for async gRPC services")
```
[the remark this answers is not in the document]
robem: No. The recommendation is to go with the number of CPUs.
The default of `1` still processes requests concurrently without issues in small setups. Only when we go on beefier infrastructure this will matter.

client/clientservice/include/client/clientservice/client_service.hpp
robem: I'm not a huge fan of including the type in the variable name but I see your point. How about `cqs_`?

robem:
> Thanks for the change @robem. LGTM! I have a couple of minor comments. Additionally, since the documentation calls `to shutdown the server prior to shutting down the returned completion queue.`, and we don't explictly shut down the clientservice, do you think that is something we need to add?

Good point, we can add signal handlers to shutdown gracefully. I will add a new task for this.

client/clientservice/src/main.cpp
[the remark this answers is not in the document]
shgandhi: Got it, thanks!

client/clientservice/include/client/clientservice/client_service.hpp
```diff
@@ -28,21 +28,30 @@ class ClientService {
       : logger_(logging::getLogger("concord.client.clientservice")),
         client_(std::move(client)),
         event_service_(std::make_unique<EventServiceImpl>(client_)),
-        request_service_(std::make_unique<RequestServiceImpl>(client_)),
         state_snapshot_service_(std::make_unique<StateSnapshotServiceImpl>(client_)){};
 
-  void start(const std::string& addr, uint64_t max_receive_msg_size);
+  void start(const std::string& addr, int num_async_threads, uint64_t max_receive_msg_size);
```
dartdart26, line 33: Nit: I think `num_async_threads` can be an unsigned type.

client/clientservice/src/client_service.cpp
```diff
@@ -15,22 +15,66 @@
...
+
+  LOG_INFO(logger_, "Wait for async gRPC threads to finish");
+  for (auto& t : server_threads_) {
+    t.join();
+  }
+}
+
+void ClientService::handleRpcs(int thread_idx) {
+  // Note: Memory is freed in `proceed`
+  new requestservice::RequestServiceCallData(&request_service_, cqs_[thread_idx].get(), client_);
```
dartdart26, line 57: How is the `RequestServiceCallData` object itself freed? It seems we never assign the pointer to the allocated memory. Do I miss something?

client/clientservice/src/client_service.cpp
```diff
@@ -15,22 +15,66 @@
...
+
+  void* tag;  // uniquely identifies a request.
```
dartdart26, line 59: Nit: I think the tag can be of type `requestservice::RequestServiceCallData*` and then implicitly converted to `void*`, avoiding the static_cast below.

client/clientservice/src/client_service.cpp
dartdart26: I saw the `delete this` in proceed(). I am wondering if this pattern is exception-safe? The way I understand it is that the pointer is passed around until we reach the end state, but I was thinking if we never do, due to an exception.

In my mind, it is confusing and risky, but I might be missing something and I haven't used a similar one before. What if we use a shared_ptr to pass it around?

dartdart26: One more case I can think of is when `handleRpcs()` returns early - then there will be a memory leak, if I am not mistaken.

dartdart26:
@robem I need to wrap my head around it, but it is getting late. I will continue tomorrow.

client/clientservice/include/client/clientservice/client_service.hpp
robem: Thanks, will fix in a follow-up.

client/clientservice/src/client_service.cpp
robem: I'm following the gRPC example applications and have looked into others as well. This seems to be the pattern for async services. I will add a try-catch block and schedule an alarm to reach the finish state.
With shared pointers, we could have the clientservice own the calldata for all ongoing requests. Meaning, we maintain a list/map of pointers that we have to protect from concurrent access or we make them thread-local 🤔

client/clientservice/src/client_service.cpp
robem: This is from the example code and I prefer to leave it that way. It emphasizes that the completion queue just deals with `void*` which can represent anything. This will later be used if we decide to make more services asynchronous. In addition, to my knowledge the static_cast doesn't add any performance overhead.

PR #2246, opened 2022-01-25
Helen tests failure fix

situ-s:
HelenapiTests were failing as the test was unable to parse the response.

These tests use kvb_client.cpp. Dueing create of ClientRequestmsg was passed with SUCCESS result variable, hence request was not executed, causing zero size response as in logs.

Passed UNKNOWN during clientRequestMsg creation to fix the issue.


yuliasherman (approved):
LGTM

PR #2253, opened 2022-01-26
API return value change to PruneInfo

titusen:
change from uint64_t getPruneBlocksPerSecond(), to PruneInfo getPruneInfo()

### DIFF
--- a/resources-manager/include/IResourceManager.hpp
+++ b/resources-manager/include/IResourceManager.hpp
@@ -15,13 +15,19 @@
 #include <cstdint>
 
 namespace concord::performance {
+
+struct PruneInfo {
+  uint64_t blocksPerSecond;
+  uint64_t batchSize;
+};
+
 class IResourceManager {
  public:
   virtual ~IResourceManager() = default;
   /*
-    getPruneBlocksPerSecond implementation computes proper pruning pace on its measuremnts.
+    getPruneBlocksPerSecond implementation computes proper pruning pace based on its measuremnts.
   */
-  virtual uint64_t getPruneBlocksPerSecond() const = 0;
+  virtual PruneInfo getPruneInfo() = 0;
 };
 
 }  // namespace concord::performance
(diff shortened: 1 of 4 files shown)


resources-manager/include/IResourceManager.hpp
```diff
@@ -15,13 +15,19 @@
...
+
 class IResourceManager {
  public:
   virtual ~IResourceManager() = default;
   /*
-    getPruneBlocksPerSecond implementation computes proper pruning pace on its measuremnts.
+    getPruneBlocksPerSecond implementation computes proper pruning pace based on its measuremnts.
   */
-  virtual uint64_t getPruneBlocksPerSecond() const = 0;
+  virtual PruneInfo getPruneInfo() = 0;
```
dartdart26, line 30: Is `IResourceManager` intended to be a generic interface or pruning-specific? If it is pruning-specific, above change is ok. Otherwise, I think it enforces it to be pruning-specific.

titusen: Correct ResourceManger becomes pruningSpecific, so we thought of changin the name into sth like PruningStrategy

titusen: Not sure, it should be done in this PR

resources-manager/include/IResourceManager.hpp
```diff
@@ -15,13 +15,19 @@
 #include <cstdint>
 
 namespace concord::performance {
+
+struct PruneInfo {
+  uint64_t blocksPerSecond;
```
cloudnoize, line 20: we need to have a lower resolution than per second e.g. 1 block per 5 seconds.
If it will be float we can achieve it, to any other way you think

titusen: Will change to float

cloudnoize (approved):
LGTM

PR #2254, opened 2022-01-26
Support state snapshot public keys count estimate

dartdart26:
Add the CMF StateSnapshotData.key_value_count_estimate field that
provides an estimate (with reasonable accuracy) of the count of
key-values contained in the state snapshot. If we are using an existing
snapshot at the time of the StateSnapshotRequest, use the actual public
key-value count in the snapshot. If creating a snapshot asynchronously
at the time of the request, use the current public key-value count in
the blockchain. In order to achieve it, move StateSnapshotRequest
handling to the KVBC reconfiguration handler instead of the generic one
that doesn't have access to KVBC.

Augment the Apollo tests such that we test all scenarios - with and
without a snapshot and with and without public keys. To achieve that,
add support for the --add-all-keys-as-public TesterReplica option that
allows us to add (or not) public keys.

### DIFF
--- a/kvbc/include/Replica.h
+++ b/kvbc/include/Replica.h
@@ -156,6 +156,8 @@ class Replica : public IReplica,
     return nullptr;
   }
 
+  std::optional<categorization::KeyValueBlockchain> &kvBlockchain() { return m_kvBlockchain; }
+
   ~Replica() override;
 
  protected:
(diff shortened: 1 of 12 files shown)


tdimitrov (approved):
LGTM :)

arc-vmware (approved):
LGTM thanks @dartdart26

PR #2259, opened 2022-01-26
Add prune stop message + implement basic KVBC reconfiguration handlers…

yontyon:
… for the existing commands

In this PR we add a PruneStop CMF message to handle the prune-stop request.
In addition, we also added a basic implementation for the KVBC reconfiguration handler (writing the new pruning commands to the blockchain)


cloudnoize (approved):
LGTM

PR #2262, opened 2022-01-27
This fix has been added for the delay happening during socket connect.

ananpal:
when the peer is unresponsive async_connect is taking 30+ seconds to execute.
I have added a deadline timer(currently 2-sec timeout)before each call to async_connect,
within 2 seconds if async_connect is not successful it will be aborted, and we already
have a retry mechanism to connect in place.

### DIFF
--- a/communication/src/TlsConnectionManager.cpp
+++ b/communication/src/TlsConnectionManager.cpp
@@ -20,6 +20,8 @@
 #include "AsyncTlsConnection.h"
 #include "communication/StateControl.hpp"
 
+using namespace std::placeholders;
+
 namespace bft::communication::tls {
 
 void setSocketOptions(asio::ip::tcp::socket& socket) { socket.set_option(asio::ip::tcp::no_delay(true)); }
@@ -59,7 +61,7 @@ void ConnectionManager::stop() {
   acceptor_.close();
   for (auto& [_, sock] : connecting_) {
     (void)_;  // unused variable hack
-    sock.close();
+    sock.first.close();
   }
   for (auto& [id, conn] : connected_waiting_for_handshake_) {
     LOG_DEBUG(logger_, "Closing connection from: " << config_.selfId_ << ", to: " << id);
@@ -311,24 +313,33 @@ void ConnectionManager::resolve(NodeNum i) {
 }
 
 void ConnectionManager::connect(NodeNum i, const asio::ip::tcp::endpoint& endpoint) {
-  auto [it, inserted] = connecting_.emplace(i, asio::ip::tcp::socket(io_context_));
+  auto [it, inserted] =
+      connecting_.emplace(i, std::make_pair(asio::ip::tcp::socket(io_context_), asio::steady_timer(io_context_)));
   ConcordAssert(inserted);
   status_->num_connecting = connecting_.size();
-  it->second.async_connect(
+  LOG_DEBUG(logger_, "connecting to node : " << i);
+  // If async_connect takes too long to finish
+  // the timeout will happen after given duration
+  connecting_.at(i).second.expires_after(CONNECT_DEADLINE);
+  it->second.first.async_connect(
       endpoint, asio::bind_executor(strand_, [this, i, endpoint](const auto& error_code) {
         if (error_code) {
           LOG_WARN(logger_, "Failed to connect to node " << i << ": " << endpoint << " : " << error_code.message());
-          connecting_.at(i).close();
-          connecting_.erase(i);
-          status_->num_connecting = connecting_.size();
+          if (connecting_.find(i) != connecting_.end()) {
+            connecting_.at(i).first.close();
+            LOG_WARN(logger_, "socket closed ");
+            connecting_.at(i).second.cancel();
+          }
           return;
         }
         LOG_INFO(logger_, "Connected to node " << i << ": " << endpoint);
-        auto connected_socket = std::move(connecting_.at(i));
-        connecting_.erase(i);
-        status_->num_connecting = connecting_.size();
+        auto connected_socket = std::move(connecting_.at(i).first);
+        // cleanup of entry from connecting map is being done in callback of timer
+        connecting_.at(i).second.cancel();
         startClientSSLHandshake(std::move(connected_socket), i);
       }));
+  connecting_.at(i).second.async_wait(
+      asio::bind_executor(strand_, std::bind(&ConnectionManager::handleConnectTimeout, this, i, _1)));
 }
 
 void ConnectionManager::connect() {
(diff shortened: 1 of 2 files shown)


communication/src/TlsConnectionManager.cpp
```diff
@@ -341,6 +355,17 @@ void ConnectionManager::connect() {
...
 
+void ConnectionManager::handleConnectTimeout(NodeNum i, const std::error_code& error_code) {
+  asio::steady_timer& deadline_timer_ = connecting_.at(i).second;
+  if (deadline_timer_.expiry() <= asio::steady_timer::clock_type::now()) {
+    LOG_WARN(logger_, "Timeout during connecting to node : " << i);
+    connecting_.at(i).first.close();
+    status_->num_connecting = connecting_.size();
+  }
+  LOG_TRACE(logger_, "Deadline Timer end");
+  connecting_.erase(i);
```
yuliasherman, line 366: Why connecting_.erase(i) is called also in case the timeout is not expired (sporadic call)?
Don't you need to call async_wait again in this case?

yuliasherman (changes requested):

communication/src/TlsConnectionManager.cpp
```diff
@@ -311,24 +313,36 @@ void ConnectionManager::resolve(NodeNum i) {
...
           status_->num_connecting = connecting_.size();
           return;
         }
         LOG_INFO(logger_, "Connected to node " << i << ": " << endpoint);
-        auto connected_socket = std::move(connecting_.at(i));
-        connecting_.erase(i);
+        auto connected_socket = std::move(connecting_.at(i).first);
+        // cancelling the timer
+        connecting_.at(i).second.cancel();
+
```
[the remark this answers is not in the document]
ananpal: erase is there in the callback fun of timer expiry so when timer cancel the callback will erase from the map

communication/src/TlsConnectionManager.cpp
ananpal: when timeout is not expired it will be called until we call in timer.cancel()

communication/src/TlsConnectionManager.cpp
```diff
@@ -341,6 +355,17 @@ void ConnectionManager::connect() {
   }
 }
 
+void ConnectionManager::handleConnectTimeout(NodeNum i, const std::error_code& error_code) {
+  asio::steady_timer& deadline_timer_ = connecting_.at(i).second;
```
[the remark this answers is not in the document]
ananpal: sorry I missed that somehow, I will change this.

communication/src/TlsConnectionManager.cpp
```diff
@@ -341,6 +355,17 @@ void ConnectionManager::connect() {
...
+  if (deadline_timer_.expiry() <= asio::steady_timer::clock_type::now()) {
+    LOG_WARN(logger_, "Timeout during connecting to node : " << i);
+    connecting_.at(i).first.close();
+    status_->num_connecting = connecting_.size();
+  }
+  LOG_TRACE(logger_, "Deadline Timer end");
```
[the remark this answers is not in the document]
ananpal: I have put trace so that these don't get dumped by default, if someone wants extra debugging by making changing log level to trace they could use it.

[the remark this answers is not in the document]
yuliasherman: In our system, we use DEBUG level for this purpose.

communication/src/TlsConnectionManager.cpp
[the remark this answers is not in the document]
yuliasherman: I don't think it is the correct approach. In case the connection is successfully established, we need to cancel the timer and remove the connection from connecting_ list.

communication/src/TlsConnectionManager.cpp
yuliasherman: According to the documentation, it is possible that the callback is called even when timeout is not expired yet, that's why we have 'if' in this function. You don't distinguish between this case and timer cancelation.

yuliasherman: You did not answer my second question (regarding async_wait).

communication/src/TlsConnectionManager.cpp
```diff
@@ -311,24 +313,35 @@ void ConnectionManager::resolve(NodeNum i) {
...
-  auto [it, inserted] = connecting_.emplace(i, asio::ip::tcp::socket(io_context_));
+  auto [it, inserted] =
+      connecting_.emplace(i, std::make_pair(asio::ip::tcp::socket(io_context_), asio::steady_timer(io_context_)));
   ConcordAssert(inserted);
   status_->num_connecting = connecting_.size();
-  it->second.async_connect(
+  LOG_DEBUG(logger_, "connecting to node : " << i);
+  // If async_connect takes too long to finish
+  // the timeout will happen after given duration
+  connecting_.at(i).second.expires_after(CONNECT_DEADLINE);
```
cjn, line 323: Shouldnt the async_wait call for the timer be here

communication/src/TlsConnectionManager.h
```diff
@@ -137,7 +141,7 @@ class ConnectionManager {
   // Sockets that are in progress of connecting.
   // When these connections complete, an AsyncTlsConnection will be created and moved into
   // `connected_waiting_for_handshake_`.
-  std::unordered_map<NodeNum, asio::ip::tcp::socket> connecting_;
+  std::unordered_map<NodeNum, std::pair<asio::ip::tcp::socket, asio::steady_timer>> connecting_;
```
cjn, line 144: Please update comment to mention about the timer also store in the map

communication/src/TlsConnectionManager.cpp
ananpal: its there at line 343 in code

ananpal: it is there at line 343 in code

cjn: Yes. But shouldnt that be moved to here logically

ananpal: it works in the fashion where at first timer initialization, then the async call on which timeout is needed then async_await .

communication/src/TlsConnectionManager.cpp
```diff
@@ -311,24 +313,35 @@ void ConnectionManager::resolve(NodeNum i) {
...
-          connecting_.erase(i);
+          if (connecting_.find(i) != connecting_.end()) {
+            connecting_.at(i).first.close();
+            LOG_WARN(logger_, "socket closed ");
+            connecting_.at(i).second.cancel();
+          }
           status_->num_connecting = connecting_.size();
           return;
         }
         LOG_INFO(logger_, "Connected to node " << i << ": " << endpoint);
```
yuliasherman, line 336: status_->num_connecting = connecting_.size(); is needed only in case we erase the socket from connecting_.

communication/src/TlsConnectionManager.cpp
```diff
@@ -341,6 +354,17 @@ void ConnectionManager::connect() {
   }
 }
 
+void ConnectionManager::handleConnectTimeout(NodeNum i, const std::error_code& error_code) {
+  asio::steady_timer& deadline_timer = connecting_.at(i).second;
+  if (deadline_timer.expiry() <= asio::steady_timer::clock_type::now()) {
+    LOG_WARN(logger_, "Timeout during connecting to node : " << i << ": " << error_code.message());
+    connecting_.at(i).first.close();
+    status_->num_connecting = connecting_.size();
```
yuliasherman, line 362: As we talked yesterday, status_->num_connecting = connecting_.size(); should be called after erase

yuliasherman (approved):
LGTM, but please verify that concord CI tests pass before merging this PR.

teoparvanov:
Hi @ananpal , as discussed on Slack, please use a fork repo to create PRs. Otherwise the CI jobs get triggered twice and waste unnecessary capacity. I'll close this one - please just re-publish the same content, but from a fork repo and merge from there.

Thank you,
Teo

PR #2265, opened 2022-01-27
Implement the resource entity interface

cloudnoize:
The ReplicaResourceEntity class measures some metrics that should give us an indication of the rate the system can perform pruning.
A user of this class should use those measurements to define a strategy.
The implemented measurements:

Pruning avg time - the avg time in micro-sec it takes to prune a block.
Pruning utilization - the percentage that the system was performing pruning from a certain duration.
Post execution utilization - the percentage that the system was performing post-execution from a certain duration.
bft queue size - the avg size of the pending requests.
Block addition
Count of consensus reached
In order to measure utilization, a helper class is implemented as part of this PR.
its algorithm is to operate on a vector of duration time measurements, where each measurement is a pair of {start,end}
then it calculates from the total time i.e. last_measure.end - first_measure.start, how much time the operation was busy,
and return the percentage,


kvbc/src/replica_resources.cpp
```diff
@@ -0,0 +1,113 @@
...
+      break;
+    default:
+      break;
+  }
+  unlock(reading_measurements_thread);
+  return ret;
+}
+
+// based on "The Peterson Lock"
+bool ReplicaResourceEntity::lock(uint8_t i, bool wait) const {
```
yontyon, line 85: Why not simply use a regular mutex with try_lock?

kvbc/src/replica_resources.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+using namespace concord::performance;
+using namespace std::chrono_literals;
+
+void ReplicaResourceEntity::addMeasurement(const ISystemResourceEntity::measurement &m) {
+  if (is_stopped) {
+    return;
+  }
+  // don't wait for the lock if not available, it's ok to loose small amount of measurements
+  if (!lock(adding_measurements_thread, dont_wait)) return;
```
yontyon, line 15: you also need to unlock here, otherwise, the other thread won't have the chance to "grab" the lock until this thread will mark itself as the victim again

yontyon (approved):
LGTM!
Thanks

PR #2268, opened 2022-01-27
Validate pre-processor reply messages of a batch

Efrat1:
Call validate() for each message in PreProcessBatchReplyMsg.
Also, validate number of messages in PreProcessBatchReplyMsg.


yuliasherman (approved):
LGTM

PR #2269, opened 2022-01-27
Present num of client service parameter

Liorgal28:
As part of our work to unify all client service entities into 1 communication layer, we need to have a unique identity for every client service.
The id of the client service will be after all the other principles in the system.


yuliasherman (approved):
LGTM

PR #2270, opened 2022-01-27
Update clang-format in docker

WildFireFlum:
Updated the clang-format version in the docker image
so that format issues display the file name and line
number.


toly-kournik (approved):
LGTM!
Thanks @WildFireFlum !

PR #2271, opened 2022-01-27
TS:Don't allow requests to execute when time is incorrect.

ukg19:
When a non-primary detects that a PrePrepare has a timestamp which
exceeds the hard limit, it doesn't continue with consensus.
However as part of RFMD it gets the missing data from other replicas who
themseleves had detected incorrect time in Pre-prepare. Thus the request makes through
consensus despite having wrong time.
We simply make a check to not send consensus messages as part of RFMD if
time was not correct for that sequence number.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -3685,7 +3685,7 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
       sendAndIncrementMetric(fcp, msgSender, metric_sent_fullCommitProof_msg_due_to_reqMissingData_);
     }
 
-    if (msg->getPartialProofIsMissing()) {
+    if (msg->getPartialProofIsMissing() && seqNumInfo.isTimeCorrect()) {
       // TODO(GG): consider not to send if msgSender is not a collector
       PartialCommitProofMsg *pcf = seqNumInfo.getFastPathSelfPartialCommitProofMsg();
 
@@ -3695,7 +3695,7 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
       }
     }
 
-    if (msg->getPartialPrepareIsMissing() && (currentPrimary() == msgSender)) {
+    if (msg->getPartialPrepareIsMissing() && (currentPrimary() == msgSender) && seqNumInfo.isTimeCorrect()) {
       PreparePartialMsg *pr = seqNumInfo.getSelfPreparePartialMsg();
 
       if (pr != nullptr) {
@@ -3713,7 +3713,7 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
       }
     }
 
-    if (msg->getPartialCommitIsMissing() && (currentPrimary() == msgSender)) {
+    if (msg->getPartialCommitIsMissing() && (currentPrimary() == msgSender) && seqNumInfo.isTimeCorrect()) {
       CommitPartialMsg *c = seqNumInfo.getSelfCommitPartialMsg();
       if (c != nullptr) {
         LOG_INFO(CNSUS, "Sending PartialCommit message as a response of RFMD" << KVLOG(msgSender, msgSeqNum));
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3671,21 +3671,21 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
...
+    if (msg->getFullCommitIsMissing() && seqNumInfo.isTimeCorrect()) {
       CommitFullMsg *c = seqNumInfo.getValidCommitFullMsg();
       if (c != nullptr) {
         LOG_INFO(CNSUS, "Sending FullCommit message as a response of RFMD" << KVLOG(msgSender, msgSeqNum));
         sendAndIncrementMetric(c, msgSender, metric_sent_commitFull_msg_due_to_reqMissingData_);
       }
     }
 
-    if (msg->getFullCommitProofIsMissing() && seqNumInfo.hasFastPathFullCommitProof()) {
+    if (msg->getFullCommitProofIsMissing() && seqNumInfo.hasFastPathFullCommitProof() && seqNumInfo.isTimeCorrect()) {
```
arc-vmware, line 3682: if time is not correct then why seqNumInfo.hasFastPathFullCommitProof() is returning true?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3704,7 +3704,7 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
       }
     }
 
-    if (msg->getFullPrepareIsMissing()) {
+    if (msg->getFullPrepareIsMissing() && seqNumInfo.isTimeCorrect()) {
```
arc-vmware, line 3707: Why should we not send the preprepare message if time is not correct?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3671,21 +3671,21 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
...
+    if (msg->getFullCommitProofIsMissing() && seqNumInfo.hasFastPathFullCommitProof() && seqNumInfo.isTimeCorrect()) {
       FullCommitProofMsg *fcp = seqNumInfo.getFastPathFullCommitProofMsg();
       LOG_INFO(CNSUS, "Sending FullCommitProof message as a response of RFMD" << KVLOG(msgSender, msgSeqNum));
       sendAndIncrementMetric(fcp, msgSender, metric_sent_fullCommitProof_msg_due_to_reqMissingData_);
     }
 
-    if (msg->getPartialProofIsMissing()) {
+    if (msg->getPartialProofIsMissing() && seqNumInfo.isTimeCorrect()) {
       // TODO(GG): consider not to send if msgSender is not a collector
       PartialCommitProofMsg *pcf = seqNumInfo.getFastPathSelfPartialCommitProofMsg();
```
arc-vmware, line 3690: should we expect partial commit proof msg as nullptr if time is incorrect ?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3671,21 +3671,21 @@ void ReplicaImp::onMessage<ReqMissingDataMsg>(ReqMissingDataMsg *msg) {
       }
     }
 
-    if (msg->getFullCommitIsMissing()) {
+    if (msg->getFullCommitIsMissing() && seqNumInfo.isTimeCorrect()) {
       CommitFullMsg *c = seqNumInfo.getValidCommitFullMsg();
```
arc-vmware, line 3675: the commit full message should be nullptr in the seq num info if time is not correct.

arc-vmware (commented):
Thanks, @ukg19 the changes look good, but I have some specific questions that we should try to understand before check-in. Let's try to check these together.

bftengine/src/bftengine/ReplicaImp.cpp
ukg19: We only need to avoid sending PreparePartialMsg and PartialCommitProofMsg. Have removed this check.

bftengine/src/bftengine/ReplicaImp.cpp
ukg19: Since we are creating next messages in consensus but not sending them only, we are not getting partialproof as nullptr.

arc-vmware (approved):
LGTM!!!

PR #2274, opened 2022-01-28
Avoid replica crash when the system attempts to recover with just a single block on the chain.

evdzhurov:
As per this PR https://github.com/vmware/concord-bft/pull/1174 the case where a replica needs to recover when only one block has been created on the chain is an unhandled edge case causing termination on the recovering replica.

The persistence test test_read_written_data_after_restart_of_all_nodes writes a single block to the chain but waits for a 2f + c + 1 quorum of replies before shutting down the cluster, this means that f + c replicas race to execute the block concurrently with the shutdown. This may trigger the edge case where some replica failed to write the block and needs to recover but the system has a chain with a single block - the unhandled edge case.

This PR avoids the above situation by forcing an N of N quorum of replies when sending the request to ensure all replicas have executed it before shutdown.

### DIFF
--- a/tests/apollo/test_skvbc_persistence.py
+++ b/tests/apollo/test_skvbc_persistence.py
@@ -24,6 +24,7 @@ from math import inf
 from util.bft import KEY_FILE_PREFIX, with_trio, with_bft_network, skip_for_tls
 from util import eliot_logging as log
 
+from bft_client import MofNQuorum
 
 def start_replica_cmd(builddir, replica_id):
     """
@@ -115,7 +116,17 @@ class SkvbcPersistenceTest(unittest.TestCase):
         """
         bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
-        (key, val) = await skvbc.send_write_kv_set()
+
+        # Ensure execution happens on all replicas before shutdown
+        # by requiring an N of N quorum of replies.
+        # Since we're creating a single block on the chain we try to 
+        # avoid the state where some replica failed to execute the block
+        # and needs to recover by starting from the genesis block. A recovery with less than 2
+        # total blocks on the chain is unhandled by the system.
+        # For more info see BC-17607.
+
+        full_quorum = MofNQuorum.All(bft_network.config, bft_network.all_replicas())
+        (key, val) = await skvbc.send_write_kv_set(quorum=full_quorum)
 
         bft_network.stop_all_replicas()
         bft_network.start_all_replicas()
--- a/tests/apollo/util/skvbc.py
+++ b/tests/apollo/util/skvbc.py
@@ -401,7 +401,7 @@ class SimpleKVBCProtocol:
             assert kv2 == dict(kv)
             action.log(message_type=f'[READ-YOUR-WRITES] OK.')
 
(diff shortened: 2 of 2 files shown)


tests/apollo/util/skvbc.py
```diff
@@ -424,22 +424,22 @@ async def send_write_kv_set(self, client=None, kv=None, max_set_size=None, long_
...
-            reply = await self.send_kv_set(client, readset, writeset, read_version, long_exec, assert_reply, raise_slowErrorIfAny)
+            reply = await self.send_kv_set(client, readset, writeset, read_version, long_exec, assert_reply, raise_slowErrorIfAny, m_of_n_quorum)
             action.log(message_type="[send_write_kv_set] OK")
             if kv_input is True:
                 return reply
             else:
                 return writeset[0][0],writeset[0][1]
 
-    async def send_kv_set(self, client, readset, writeset, read_version, long_exec=False, reply_assert=True, raise_slowErrorIfAny=True):
+    async def send_kv_set(self, client, readset, writeset, read_version, long_exec=False, reply_assert=True, raise_slowErrorIfAny=True, m_of_n_quorum=None):
```
teoparvanov, line 434: I'd simply call this parameter `quorum`, to highlight it accepts various types of quorums.

teoparvanov (approved):
LGTM, thanks for the fix @evdzhurov !

tests/apollo/util/skvbc.py
evdzhurov: I matched the already existing name for it. Do you want the change everywhere?

tests/apollo/test_skvbc_persistence.py
```diff
@@ -115,7 +116,17 @@ async def test_read_written_data_after_restart_of_all_nodes(self, bft_network, t
...
+        # Ensure execution happens on all replicas before shutdown
+        # by requiring an N of N quorum of replies.
+        # Since we're creating a single block on the chain we try to 
+        # avoid the state where some replica failed to execute the block
+        # and needs to recover by starting from the genesis block. A recovery with less than 2
+        # total blocks on the chain is unhandled by the system.
+        # For more info see BC-17607.
+
+        quorum = MofNQuorum.All(bft_network.config, bft_network.all_replicas())
+        (key, val) = await skvbc.send_write_kv_set(m_of_n_quorum=quorum)
```
teoparvanov, line 129: If you addressed my previous comment, this would read more naturally:
`send_write_kv_set(quorum=quorum)`

tests/apollo/test_skvbc_persistence.py
```diff
@@ -115,7 +116,17 @@ async def test_read_written_data_after_restart_of_all_nodes(self, bft_network, t
...
+
+        # Ensure execution happens on all replicas before shutdown
+        # by requiring an N of N quorum of replies.
+        # Since we're creating a single block on the chain we try to 
+        # avoid the state where some replica failed to execute the block
+        # and needs to recover by starting from the genesis block. A recovery with less than 2
+        # total blocks on the chain is unhandled by the system.
+        # For more info see BC-17607.
+
+        quorum = MofNQuorum.All(bft_network.config, bft_network.all_replicas())
```
teoparvanov, line 128: `full_quorum` instead of `quorum`?

tests/apollo/util/skvbc.py
teoparvanov: No, it's up to you to decide, as I said it's a minor thing.

PR #2275, opened 2022-01-28
Adaptive pruning manager first version

titusen:
Adaptive pruning manager first version

### DIFF
--- a/resources-manager/include/AdaptivePruningManager.hpp
+++ b/resources-manager/include/AdaptivePruningManager.hpp
@@ -1,0 +1,59 @@
+#pragma once
+
+#include "IRequestHandler.hpp"
+#include "IResourceManager.hpp"
+#include "InternalBFTClient.hpp"
+
+#include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <thread>
+
+enum PruningMode { LEGACY, ADAPTIVE };
+
+namespace concord::performance {
+class AdaptivePruningManager {
+ public:
+  AdaptivePruningManager(const std::shared_ptr<concord::performance::IResourceManager> &resourceManager,
+                         const std::chrono::duration<double, std::milli> &interval,
+                         const std::shared_ptr<bftEngine::impl::IInternalBFTClient> &bftClient);
+  virtual ~AdaptivePruningManager();
+
+  void switchMode(PruningMode newMode) {
+    mode = newMode;
+    conditionVar.notify_one();
+  }
+  PruningMode getCurrentMode() { return mode.load(); }
+
+  void youArePrimary() {  // Is there a listener for becoming a primary replica?
+    amIPrimary = true;
+    conditionVar.notify_one();
+  }
+
+  void youAreNotPrimary() {  // Is there a listener for not being a primary anymore?
+    amIPrimary = false;
+  }
+  void start();
+  void stop();
+
+ protected:
+  void notifyReplicas(const long double &rate, const uint64_t batchSize);
+
+ private:
+  void threadFunction();
+
+ private:
+  const std::uint16_t repId;
+  std::shared_ptr<concord::performance::IResourceManager> resourceManager;
+  std::atomic<PruningMode> mode;
+  std::chrono::duration<double, std::milli> interval;
+  std::shared_ptr<bftEngine::impl::IInternalBFTClient> bftClient;
+  std::thread workThread;
+  std::atomic<bool> isRunning, amIPrimary;
+  std::condition_variable conditionVar;
+  std::mutex conditionLock;
+};
+}  // namespace concord::performance
--- a/resources-manager/include/AdaptivePruningManager.hpp
+++ b/resources-manager/include/AdaptivePruningManager.hpp
@@ -28,14 +28,21 @@ class AdaptivePruningManager {
   }
   PruningMode getCurrentMode() { return mode.load(); }
 
-  void youArePrimary() {  // Is there a listener for becoming a primary replica?
+  void youArePrimary() {
     amIPrimary = true;
     conditionVar.notify_one();
   }
 
-  void youAreNotPrimary() {  // Is there a listener for not being a primary anymore?
-    amIPrimary = false;
+  void youAreNotPrimary() { amIPrimary = false; }
+
+  void setResourceManager(const std::shared_ptr<concord::performance::IResourceManager> &resourceManagerNew) {
+    {
+      std::unique_lock<std::mutex> lk(conditionLock);
+      resourceManager = resourceManagerNew;
+    }
+    start();
   }
+
   void start();
   void stop();
 
--- a/resources-manager/src/AdaptivePruningManager.cpp
+++ b/resources-manager/src/AdaptivePruningManager.cpp
@@ -2,6 +2,7 @@
 #include "Bitmap.hpp"
 #include "ReplicaConfig.hpp"
 #include "concord.cmf.hpp"
+#include "SigManager.hpp"
 
 using namespace concord::performance;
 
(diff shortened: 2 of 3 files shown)


resources-manager/include/AdaptivePruningManager.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <thread>
+
+enum PruningMode { LEGACY, ADAPTIVE };
```
teoparvanov, line 15: What about MANUAL mode, which I remember from the design document? Was that dropped?

resources-manager/include/AdaptivePruningManager.hpp
```diff
@@ -0,0 +1,59 @@
...
+                         const std::shared_ptr<bftEngine::impl::IInternalBFTClient> &bftClient);
+  virtual ~AdaptivePruningManager();
+
+  void switchMode(PruningMode newMode) {
+    mode = newMode;
+    conditionVar.notify_one();
+  }
+  PruningMode getCurrentMode() { return mode.load(); }
+
+  void youArePrimary() {  // Is there a listener for becoming a primary replica?
```
yontyon, line 31: You have it in the replicaImp file. Basically, the primary is always the current_view % num_replicas.
I would suggest that replicaImp will call this method once the view is changed

resources-manager/src/AdaptivePruningManager.cpp
```diff
@@ -0,0 +1,76 @@
...
+  pruneRequest.batch_blocks_num = batchSize;
+
+  rreq.command = pruneRequest;
+  rreq.sender = bftClient->getClientId();
+
+  std::vector<uint8_t> serialized_req;
+  rreq.signature = {};
+  concord::messages::serialize(serialized_req, rreq);
+
+  // auto sig = signer_->sign(std::string(serialized_req.begin(), serialized_req.end()));
```
titusen, line 38: Let me know, how to obtain certificate here, cause I don't belive that it is not avialable somewhere.

resources-manager/include/AdaptivePruningManager.hpp
titusen: It is very similiar to the legacy mode, so maybe it will not be needed at all.

resources-manager/include/AdaptivePruningManager.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <chrono>
+#include <condition_variable>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <thread>
+
+enum PruningMode { LEGACY, ADAPTIVE };
+
+namespace concord::performance {
```
cloudnoize, line 17: nit: this namespace does not reflect the functionality

resources-manager/include/AdaptivePruningManager.hpp
```diff
@@ -0,0 +1,59 @@
...
+class AdaptivePruningManager {
+ public:
+  AdaptivePruningManager(const std::shared_ptr<concord::performance::IResourceManager> &resourceManager,
```
cloudnoize, line 20: which entity will instantiate the IResourceManager and will handle it to the AdaptivePruningManager? 
Which entity is responsible for the configuration of the feature? 
by the design, I thought that this class, for example, what happens when the operator sends a new configuration for the mapping interval, with this interface, it meant that the AdaptivePruningManager will need to be destroyed and created with a new IResourceManager, instead of managing it..

resources-manager/src/AdaptivePruningManager.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+AdaptivePruningManager::~AdaptivePruningManager() { stop(); }
+
+void AdaptivePruningManager::notifyReplicas(const long double &rate, const uint64_t batchSize) {
+  concord::messages::ReconfigurationRequest rreq;
+
+  concord::messages::PruneTicksChangeRequest pruneRequest;
+  pruneRequest.sender_id = bftClient->getClientId();
+  // numOfConnectedReplicas requires cluster size. How can I get it?
```
cloudnoize, line 27: I think that cluster size should be a parameter in the constructor.

resources-manager/src/AdaptivePruningManager.cpp
```diff
@@ -0,0 +1,76 @@
...
+  while (isRunning.load()) {
+    {
+      std::unique_lock<std::mutex> lk(conditionLock);
+      conditionVar.wait(lk, [this]() { return mode.load() == ADAPTIVE && amIPrimary.load(); });
+    }
+    if (isRunning.load()) {
+      auto info = resourceManager->getPruneInfo();
+      notifyReplicas(info.blocksPerSecond, info.batchSize);
+      std::unique_lock<std::mutex> lk(conditionLock);
+      conditionVar.wait_for(lk, interval);
```
yontyon, line 58: If you use the condition variable for notifications only (and not for mutual exclusion), I think it makes sense to use internal messages instead

yontyon: And then you also don't need to use a different thread for the adaptive manager

resources-manager/src/AdaptivePruningManager.cpp
yontyon: Take a look here https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicaImp.cpp#L5287

resources-manager/include/AdaptivePruningManager.hpp
```diff
@@ -0,0 +1,66 @@
...
+  void notifyReplicas(const long double &rate, const uint64_t batchSize);
+
+ private:
+  void threadFunction();
+
+ private:
+  const std::uint16_t repId;
+  std::shared_ptr<concord::performance::IResourceManager> resourceManager;
+  std::atomic<PruningMode> mode;
+  std::chrono::duration<double, std::milli> interval;
```
cloudnoize, line 59: interval is a configuration option? we should prohibit intervals of less than a second.

resources-manager/include/AdaptivePruningManager.hpp
```diff
@@ -0,0 +1,66 @@
...
+  PruningMode getCurrentMode() { return mode.load(); }
+
+  void youArePrimary() {
+    amIPrimary = true;
+    conditionVar.notify_one();
+  }
+
+  void youAreNotPrimary() { amIPrimary = false; }
+
+  void setResourceManager(const std::shared_ptr<concord::performance::IResourceManager> &resourceManagerNew) {
```
cloudnoize, line 38: how is the update flow going to happen, who deserialize the configuration and  instantiates the correct instance of IResourceManager?

titusen: I guess ReconfigurationHandler will parse requests config and initiate the proper object. @yontyon correct?

resources-manager/src/AdaptivePruningManager.cpp
```diff
@@ -0,0 +1,81 @@
...
+AdaptivePruningManager::~AdaptivePruningManager() { stop(); }
+
+void AdaptivePruningManager::notifyReplicas(const long double &rate, const uint64_t batchSize) {
+  concord::messages::ReconfigurationRequest rreq;
+
+  concord::messages::PruneTicksChangeRequest pruneRequest;
+  pruneRequest.sender_id = bftClient->getClientId();
+  pruneRequest.tick_period_seconds =
+      rate / bftClient->numOfConnectedReplicas(bftEngine::ReplicaConfig::instance().numReplicas);
+  pruneRequest.batch_blocks_num = batchSize;
```
cloudnoize, line 30: if the resolution is in seconds, it means that at most a replica can generate 1 tick per second? 
I think that because the resource manager doesn't know about the ticks resolution and cluster size.
it's better for it to return only blocks per second, and the manager should convert it to ticks resolution and batch size. 
for example, if the resource manager calculates 20 blocks per second, it can't calculate the batch size as it is dependent on the cluster size and tick resolution.

resources-manager/src/AdaptivePruningManager.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+AdaptivePruningManager::~AdaptivePruningManager() { stop(); }
+
+void AdaptivePruningManager::notifyReplicas(const long double &rate, const uint64_t batchSize) {
+  concord::messages::ReconfigurationRequest rreq;
+
+  concord::messages::PruneTicksChangeRequest pruneRequest;
+  pruneRequest.sender_id = bftClient->getClientId();
+  pruneRequest.tick_period_seconds =
```
cloudnoize, line 28: in case the rate is lower than cluster size it will result in 0 rate.
the logic should be more sophisticated, to know how to "play" with the rate in order to support different bps options.

titusen: Totally agree, but I think we should have it simple for now. Currently, the minimum supported tick per second is 1 and it can't go lower.

resources-manager/src/AdaptivePruningManager.cpp
titusen: As below 1 tick per second is the minimum not maximum. About, blocks per second - it was this way, but it was agreed that we change it to PruneInfo (ticks per second, batchSzie). I can change it back to what it was, but we are running out of time.

resources-manager/include/AdaptivePruningManager.hpp
cloudnoize: Besides enabling the thread, do those flags have any other meaning?

cloudnoize (approved):
LGTM, thanks!

PR #2276, opened 2022-01-28
Add support for the StateSnapshotReadAsOfRequest

dartdart26:
Introduce the `StateSnapshotReadAsOfRequest` command that allows users
to read a list of state key-values.

Add support for state value conversion by a way of an optional
user-provided converter function. It allows users to convert state
values to any format that is appropriate. The default converter function
doesn't change the state value in any way, just returning it as is.

Fix TesterReplica's behaviour regarding public state - namely, do not
allow duplicate public state keys.

Make sure we add unique keys in DB snapshots tests for which it matters.
In order to do so, use skvbc.unique_random_key() instead of
skvbc.random_key().

Add Apollo tests for the new behaviour.


nkumar04 (approved):
LGTM!
Thanks!

PR #2277, opened 2022-01-28
TS: Do not send PreparePartial and PartialCommitProof messages if Time is incorrect

ukg19:
This PR adds a fix where we want to avoid sending of PreparePartial and PartialCommitProof messages (so that request doesn't commit with incorrect time) for which time was detected as incorrect during preprepare message processing.


arc-vmware (approved):
LGTM!
thanks @ukg19

PR #2278, opened 2022-01-28
Consolidate fast and slow path tests into test_skvbc_commit_path.

evdzhurov:
This PR aims to consolidate all fast and slow path tests into a single suite to avoid duplication and perform tests consistently.

The tests from fast path have been changed as follows:
* test_fast_path_only renamed to test_fast_path_is_default
* test_fast_to_slow_path_transition renamed to test_commit_path_transitions

The tests from slow path have been changed as follows:
* test_persistent_slow_path - tests fast to slow path transition, folded into test_commit_path_transitions
* test_slow_to_fast_path_transition - tests slow to fast path transition folded into test_commit_path_transitions
* test_slow_path_view_change - this test has been revamped as test_fast_path_after_view_change. The rationale is that the slow path is only an intermediate step before the system eventually reaches the fast path. This test now ensures that if a view change occurs the fast path is restored. Whether the slow path is used as an intermediate step is an implementation detail that may change as implementations evolve. This is consistent with how the fast path tests were changed previously to ignore the slow path as a transient state (for example when the controller downgrades from FAST_OPTIMISTIC to OPTIMISTIC_WITH_THRESHOLD).

test_fast_path_is_default and test_fast_to_slow_path are disabled because they are part of test_commit_path_transitions but are left as an option for manual test invocation.

test_fast_path_after_view_change now properly sends a write request to trigger view change, previously test_slow_path_view_change would time out because the requests had a 1 in 400 chance to be all reads and would not trigger a view change.


teoparvanov (approved):
LGTM!

HristoStaykov (approved):
LGTM, thanks @evdzhurov !

PR #2279, opened 2022-01-28
Clientservice: Async threads improvments

robem:
Various improvements to async service handling. See individual commits for details.

### DIFF
--- a/client/clientservice/src/client_service.cpp
+++ b/client/clientservice/src/client_service.cpp
@@ -46,10 +46,17 @@ void ClientService::start(const std::string& addr, unsigned num_async_threads, u
 
   clientservice_server->Wait();
 
-  LOG_INFO(logger_, "Wait for async gRPC threads to finish");
-  for (auto& t : server_threads_) {
-    t.join();
+  LOG_INFO(logger_, "Shutting down and emptying completion queues");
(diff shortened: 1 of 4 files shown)


client/clientservice/src/client_service.cpp
```diff
@@ -46,10 +46,17 @@ void ClientService::start(const std::string& addr, unsigned num_async_threads, u
 
   clientservice_server->Wait();
 
-  LOG_INFO(logger_, "Wait for async gRPC threads to finish");
-  for (auto& t : server_threads_) {
-    t.join();
+  LOG_INFO(logger_, "Shutting down and emptying completion queues");
+  std::for_each(cqs_.begin(), cqs_.end(), [](auto& cq) { cq->Shutdown(); });
```
shgandhi, line 50: Acc. to the [documentation](https://grpc.io/docs/languages/cpp/async/#shutting-down-the-server) we should be shutting down the server before the completion queues. Don't we need to do that here?

shgandhi (approved):
I just have one comment.

client/clientservice/src/client_service.cpp
robem: `Wait()` only returns when that has happened. https://grpc.github.io/grpc/cpp/classgrpc_1_1_server_interface.html

PR #2281, opened 2022-01-31
Persist quorum of complaints

HristoStaykov:
We persist in DescriptorOfLastExitFromView the quorum of complaints
we have accumulated and will later insert in our View Change message
to serve as proof quorum was reached to move to the higher view.
This way after restarts, we are going to recover exactly the same
View Change message as we had prior to the restart.

### DIFF
--- a/bftengine/src/bftengine/PersistentStorageImp.hpp
+++ b/bftengine/src/bftengine/PersistentStorageImp.hpp
@@ -51,6 +51,7 @@ namespace impl {
 // Make a reservation for future params
 const uint16_t reservedSimpleParamsNum = 499;
 const uint16_t reservedWindowParamsNum = 3000;
+const uint16_t reservedComplaintsNum = 300;
 
 const uint16_t MAX_METADATA_PARAMS_NUM = 10000;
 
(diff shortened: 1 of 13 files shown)


bftengine/src/bftengine/PersistentStorageImp.hpp
```diff
@@ -94,7 +95,8 @@ const uint16_t numOfLastExitFromViewDescObjs = kWorkWindowSize + 1;
 // objects plus one - for simple descriptor parameters.
 enum DescMetadataParameterIds {
   LAST_EXIT_FROM_VIEW_DESC = WIN_PARAMETERS_NUM + reservedWindowParamsNum,
-  LAST_EXEC_DESC = LAST_EXIT_FROM_VIEW_DESC + numOfLastExitFromViewDescObjs,
+  LAST_COMPLAINTS_DESC = LAST_EXIT_FROM_VIEW_DESC + numOfLastExitFromViewDescObjs,
+  LAST_EXEC_DESC = LAST_COMPLAINTS_DESC + reservedComplaintsNum + 1,
```
evdzhurov, line 99: I don't know if this is important, but LAST_COMPLAINTS_DESC is now the id of the first complaint written to db while previously it was the id of DescriptorOfLastExecution, this seems to break binary compatibility if data written prior to this code update needs to be read.

evdzhurov (changes requested):

yontyon (approved):
LGTM!
Thanks

bftengine/src/bftengine/PersistentStorageImp.hpp
HristoStaykov: I think the only situation in which this could happen is after update, but currently when we update versions we clear the metadata. Therefore the shifting should be OK, and it preserves the context of the parameter groups.

evdzhurov (approved):
LGTM

teoparvanov (approved):
LGTM, thanks Hristo!

PR #2283, opened 2022-01-31
Implement the interval mapping resource manager

cloudnoize:
This class stores a mapping of transactions per second to how many blocks can be deleted per second.
It calculates the TPS from the last call and returns the mapped result when receiving a query.


resources-manager/src/IntervalMappingResourceManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+  auto pruningAvgTimeMicro = replicaResources_->getMeasurement(ISystemResourceEntity::type::pruning_avg_time_micro);
+  auto tps = transactions / duration;
+  // in order to give an up to date result in the next invocation
+  replicaResources_->reset();
+  auto it = std::upper_bound(intervalMapping_.begin(), intervalMapping_.end(), std::make_pair(tps, (u_int64_t)0));
+
+  PruneInfo ret;
+  if (it != intervalMapping_.end()) {
+    ret.blocksPerSecond = static_cast<long double>(it->second);
+    ret.batchSize = 1;
```
titusen, line 37: One option is that instead of mapping tps only to bps, but to pair (ticks per second, block in batch)

titusen (approved):
For sure, it will change for now it is feasible.

PR #2285, opened 2022-01-31
Create db-snapshot with wedge and stablising apollo tests

nkumar04:
When we create a snapshot on wedge, we need to make sure that, we trigger create of db snapshot after executing all other callbacks like set flag to erase metadata, start new epoch etc. Changes in the PR addresses this.
Also, we stabilises some of the apollo tests


tests/apollo/test_skvbc_dbsnapshot.py
```diff
@@ -652,9 +654,8 @@ async def test_scale_and_restart_blockchain_with_db_snapshot(self, bft_network):
                                                          "Gauges", "lastDbCheckpointBlockId", component="rocksdbCheckpoint")
             await self.wait_for_snapshot(bft_network, replica_id, last_block_id)
         bft_network.stop_all_replicas()
-        for r in bft_network.all_replicas():
-            #restore all replicas using snapshot created in replica id=0
-            self.restore_form_older_snapshot(bft_network, 0, last_block_id)
+        #restore all replicas using snapshot created in replica id=0
+        self.restore_form_older_snapshot(bft_network, last_block_id, src_replica = 0, dest_replicas = bft_network.all_replicas())
         #replicas will clean metadata and start a new blockchain
```
itsmeakashgoyal, line 659: Query: How replica's metadata will clean.

itsmeakashgoyal (approved):
LGTM.
Thanks Nikhil

tests/apollo/test_skvbc_dbsnapshot.py
nkumar04: When add_remove_with_wedge is executed, we wedge on stable checkpoint and set a flag in db clean metadata. So, on restart replicas check this flag and erase metadata

itsmeakashgoyal: got it. Thanks

yontyon (approved):
LGTM!
Thanks!

PR #2287, opened 2022-02-01
By default, run Apollo tests with c=0 (except fast/slow path scenarios)

samcramer:
Speed up Apollo tests by limiting configurations.

In our real environments we don't support deployments where c > 0. On the other hand, Apollo tests by default currently run for both n=7 (c=0) and n=6 (c=1).


teoparvanov (approved):
Hi @samcramer, the code change looks good. Can you please also add a description to your PR, documenting the reasoning behind the change?

Thank you!

samcramer:
> Hi @samcramer, the code change looks good. Can you please also add a description to your PR, documenting the reasoning behind the change?
> 
> Thank you!

Will do.

PR #2290, opened 2022-02-01
Fix skvbc_persistance test generating a FATAL error.

evdzhurov:
Fix test_read_written_data_after_restart_of_all_nodes generating a FАTAL error by avoiding the situation where only a single block is created on the chain and some replica that failed to execute it must recover. This is currently an unhandled edge case.


teoparvanov (approved):
LGTM!

PR #2291, opened 2022-02-01
Metrics and logs addition

titusen:
Metrics and logs for Adaptive Pruning visibility.


yontyon (changes requested):
@titusen, you need to assign the operator of the kvbc/replica in order to propagate the metrics to the higher levels (such as metrics server, wavefront, and so on.
I would suggest having it in the constructor.

resources-manager/src/AdaptivePruningManager.cpp
```diff
@@ -25,11 +28,19 @@ void AdaptivePruningManager::notifyReplicas(const long double &rate, const uint6
...
+  pruneRequest.ticks_per_second = 1;
 
   pruneRequest.batch_blocks_num =
       rate / bftClient->numOfConnectedReplicas(bftEngine::ReplicaConfig::instance().numReplicas);
 
+  // Is this going to register all send values or just update current
+  ticksPerSecondMetric.Get().Set(pruneRequest.ticks_per_second);
+  batchSizeMetric.Get().Set(pruneRequest.batch_blocks_num);
+
+  LOG_DEBUG(ADPTV_PRUNING,
```
cloudnoize, line 40: Since frequency is very low, we can set it as INFO

kvbc/src/resources-manager/AdaptivePruningManager.cpp
```diff
@@ -26,10 +32,18 @@ void AdaptivePruningManager::notifyReplicas(const long double &rate, const uint6
...
 
   pruneRequest.sender_id = bftClient->getClientId();
-  pruneRequest.tick_period_seconds = 1;
+  pruneRequest.ticks_per_second = 1;
 
   pruneRequest.batch_blocks_num = rate / bftEngine::ReplicaConfig::instance().numReplicas;
 
+  // Is this going to register all send values or just update current
+  ticksPerSecondMetric.Get().Set(pruneRequest.ticks_per_second);
+  batchSizeMetric.Get().Set(pruneRequest.batch_blocks_num);
```
cloudnoize, line 41: we also need to send metrics from the resource manager e.g. post-execution utilization, pruning utilization, pruning avg time etc

cloudnoize: For simplicity, the pruning info struct can be extended to include those values and reported here

kvbc/include/resources-manager/IResourceManager.hpp
```diff
@@ -19,6 +19,9 @@ namespace concord::performance {
 struct PruneInfo {
   long double blocksPerSecond{0};
   uint64_t batchSize{0};
+  uint64_t transactionsPerSecond{0};
```
cloudnoize, line 22: also pruning utilization

kvbc/src/resources-manager/AdaptivePruningManager.cpp
```diff
@@ -11,31 +11,51 @@ using namespace concord::performance;
...
     LOG_ERROR(ADPTV_PRUNING, "BFT client is not set");
     return;
   }
   concord::messages::ReconfigurationRequest rreq;
 
   concord::messages::PruneTicksChangeRequest pruneRequest;
-
+  /*  */
   pruneRequest.sender_id = bftClient->getClientId();
-  pruneRequest.tick_period_seconds = 1;
```
titusen, line 36: @dartdart26, does ticks per second in concurrent pruning means that n=5 ticks are generated by each replica every second? Is the frequency 1/n then?

dartdart26: Ticks period means that each replica will generate a tick every `tick_period_seconds`. For example, if you configure `tick_period_seconds = 1` and there are 7 replicas, there will be 7 ticks per second in the system.

cloudnoize (approved):
LGTM

yontyon (approved):
LGTM!

PR #2292, opened 2022-02-01
ThinReplicaServer: Reduce log spam

robem:
Although the log messages are on debug level already, they get printed pretty
much non-stop. This makes it hard to read/debug logs. If the TRS successfully
sends updates to the TRC then there will be a debug message that shows it so we
don't need to print every attempt.


shgandhi (approved):
LGTM!

PR #2297, opened 2022-02-02
Unskip test_fast_path_is_default because it's used internally

evdzhurov:
This test should be running by default since it's needed by internal testing.


teoparvanov (approved):
LGTM, thanks!

PR #2298, opened 2022-02-02
Added test suite for restart recovery.

HristoStaykov:
In this test suite we will perform scenarios with replica restarts
during different states of the system to verify correct recovery of
replicas. This test suite is not meant to be part of the CI, because
it requires running over longer periods. It might be used for
daily/weekly runs.
Currently we have 3 scenarios:
1. Restart a replica while the system is processing client requests
   to verify the restarted replica recovers correctly and the system
   returns to Fast Commit path.
2. Perform multiple View Changes and when the replicas start to initiate
   View change restart 1 random non-Primary replica. Then after the view
   change is complete verify the system restores Fast Commit path.
3. Perform multiple View Changes with the current and next Primaries down. 
    Wait for replicas start initiating the second View Change and restart 1
    random non-Primary replica. Then after the view change is complete
    verify the system restores Fast Commit path.


teoparvanov (approved):
This test is just for manual/nightly runs, right? It should be disabled in CI.

PR #2299, opened 2022-02-02
Expose state snapshot reconf commands to clients

dartdart26:
Introduce the `StateSnapshotReconfigurationHandler` class and move state
snapshot handle() methods to it. Furthermore, override the
verifySignature() method such that it allows requests from the
Operator (via a `BftReconfigurationHandler` instance) and from Clients
(via a `ClientReconfigurationHandler` instance).

We choose composition over inheritance for the handlers, because:
 *  we'd like to avoid complex multiple virtual inheritance
 * `ClientReconfigurationHandler` has a handle() method that would be
   called multiple times if we inherited from it

Finally, add support for state snapshot value conversion in
`kvbc::Replica` and `KeyValueBlockchain`.

### DIFF
--- a/kvbc/include/Replica.h
+++ b/kvbc/include/Replica.h
@@ -159,6 +159,9 @@ class Replica : public IReplica,
   }
 
   std::optional<categorization::KeyValueBlockchain> &kvBlockchain() { return m_kvBlockchain; }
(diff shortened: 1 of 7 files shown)


arc-vmware (approved):
LGTM!!!
thanks @dartdart26

PR #2300, opened 2022-02-02
S3 connectivity retry logic

toly-kournik:
S3 client connectivity logic to infinitely retry in case S3 Object Store is down.
S3 client iterator bug fix.
Change State Transfer to run in main thread to apply back pressure if S3 Object Store is not accessible.
Add two configuration files for s3_integrity_check utility.


Liorgal28 (approved):
🚀

PR #2301, opened 2022-02-03
This fix has been added for the delay happening during socket connect.

ananpal:
when a peer is unresponsive async_connect is taking 30+ seconds to execute
I have added a deadline timer(currently 2-sec timeout)before each call to async_connect,
within 2 seconds if async_connect is not successful it will be aborted, and we already
have a retry mechanism to connect in place.

This is a duplicated pull request created in the forked branch.
https://github.com/vmware/concord-bft/pull/2262/

### DIFF
--- a/communication/src/TlsConnectionManager.cpp
+++ b/communication/src/TlsConnectionManager.cpp
@@ -20,6 +20,8 @@
 #include "AsyncTlsConnection.h"
 #include "communication/StateControl.hpp"
 
+using namespace std::placeholders;
+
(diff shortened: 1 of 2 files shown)


teoparvanov (approved):
Hi @ananpal , thanks for addressing my comment about using a forked repo.

As previously suggested by Yulia, the next step is to create a private Concord build with these changes and run it with Maestro. This would ensure system performance is preserved.

yuliasherman (approved):
LGTM

PR #2302, opened 2022-02-03
Add `ledger_time` to `GetRecentSnapshotResponse`

dartdart26:
The ledger time at which the snapshot was taken.


arc-vmware (approved):
LGTM!!!
thanks @dartdart26

PR #2304, opened 2022-02-03
Change Github workflow so clang debug builds use s3 storage only.

evdzhurov:
This is done to reduce the CI workload by omitting building and testing with the option turned off.


teoparvanov (approved):
LGTM, thanks @evdzhurov !

PR #2305, opened 2022-02-03
Implement adaptive pruning reconfiguration handlers

yontyon:
In this PR we implement the adaptive pruning reconfiguration handlers.
We took the approach of having as much logic in concordbft in order to ease the implementation and the migration of concurrent pruning to concortdbft.
In addition, we load the latest tick configuration and the latest pruning mode configuration on startup and on state transfer completion.


titusen (approved):
LGTM

kvbc/include/st_reconfiguraion_sm.hpp
```diff
@@ -68,10 +73,18 @@ class StReconfigurationHandler {
   bool handle(const concord::messages::RestartCommand&, uint64_t, uint64_t, uint64_t);
   bool handle(const concord::messages::PruneRequest&, uint64_t, uint64_t, uint64_t);
   bool handle(const concord::messages::InstallCommand&, uint64_t, uint64_t, uint64_t);
+  bool handle(const concord::messages::PruneSwitchModeRequest&, uint64_t, uint64_t, uint64_t);
+
+  logging::Logger& getLogger() const {
```
dartdart26, line 78: Nit: Can make it `static` instead of ` const`.

dartdart26 (approved):
Looks good, thanks!

PR #2306, opened 2022-02-03
Fix path transition test

WildFireFlum:
This branch adds a test for the transition from the FAST_OPTIMISTIC commit path to the FAST_WITH_THRESHOLD path.
It changes the way metrics are counted as they previously did not account for three commit paths.
The full (trinary) path information is not propagated through the consensus protocol, but is propagated through the execution path, and is therefore used.
This branch also adds optional prints and seeded test runs for easier debugging process.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -928,7 +928,6 @@ void ReplicaImp::startConsensusProcess(PrePrepareMsg *pp, bool isCreatedEarlier)
 
   if (firstPath == CommitPath::SLOW) {
     seqNumInfo.startSlowPath();
-    metric_slow_path_count_++;
     TimeRecorder scoped_timer1(*histograms_.sendPreparePartialToSelf);
     sendPreparePartial(seqNumInfo);
   } else {
@@ -1135,7 +1134,6 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
         sendPartialProof(seqNumInfo);
       } else {
         seqNumInfo.startSlowPath();
-        metric_slow_path_count_++;
         sendPreparePartial(seqNumInfo);
       }
       // if time is not ok, we do not continue with consensus flow
@@ -1194,7 +1192,6 @@ void ReplicaImp::tryToStartSlowPaths() {
     controller->onStartingSlowCommit(i);
 
     seqNumInfo.startSlowPath();
-    metric_slow_path_count_++;
 
     if (ps_) {
       ps_->beginWriteTran();
@@ -1203,7 +1200,6 @@ void ReplicaImp::tryToStartSlowPaths() {
     }
 
     // send StartSlowCommitMsg to all replicas
-
     StartSlowCommitMsg *startSlow = new StartSlowCommitMsg(config_.getreplicaId(), getCurrentView(), i);
 
     if (!retransmissionsLogicEnabled) {
@@ -1301,7 +1297,6 @@ void ReplicaImp::onMessage<StartSlowCommitMsg>(StartSlowCommitMsg *msg) {
       LOG_INFO(CNSUS, "Start slow path.");
 
       seqNumInfo.startSlowPath();
-      metric_slow_path_count_++;
 
       if (ps_) {
         ps_->beginWriteTran();
@@ -3205,7 +3200,6 @@ void ReplicaImp::onNewView(const std::vector<PrePrepareMsg *> &prePreparesForNew
       seqNumInfo.addMsg(pp);
 
     seqNumInfo.startSlowPath();
-    metric_slow_path_count_++;
   }
 
   if (ps_) ps_->endWriteTran(config_.getsyncOnUpdateOfMetadata());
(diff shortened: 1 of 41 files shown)


teoparvanov:
Hi @WildFireFlum , can you please provide some description? What is the motivation of this change, what is a summary of the proposed changes?

evdzhurov (commented):
I guess the main purpose of the PR is to add a test for the transition between the two fast path modes - optimistic and with threshold, however I have couple of observations:
* There are a lot of changes here, some of which I'm not sure if they are related to fixing the path transition test, like the addition of a random seed.
*  The PR adds a new method wait_for_consensus_path to detect if some execution path transition happened.  Since we already have _wait_for_consensus_path_to_be_prevalent how are we going to differentiate between the two if someone needs to decide what to use in a test? Is wait_for_consensus_path meant to detect the switch but doesn't guarantee prevalence? Further, wait_for_consensus_path says it counts operation executions but gathers metrics per pre-prepare which may contain a number of operations.
* Lastly, the change to how metrics are generated is now solely based on the firstPath parameter of a pre-prepare, which is the desired commit path but not necessarily the commited path since the slow path will begin competing with the fast path if consensus slows down for some sequence. This might still work in the case when we want to detect a switch in the state of the controller (what is the desired commit path) but I'm not sure how it will affect the stability of the alternative logic in _wait_for_consensus_path_to_be_prevalent, which wasn't ideal to begin with.

WildFireFlum:
> Hi @WildFireFlum , can you please provide some description to the PR? What is the motivation of this change, what is a summary of the proposed changes?

I added a description.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4696,6 +4680,7 @@ void ReplicaImp::setConflictDetectionBlockId(const ClientRequestMsg &clientReqMs
 // have some committed seq numbers)
 // TODO(GG): move the "logic related to requestMissingInfo" to the caller of this method (e.g., by returning a value
 // to the caller)
+// Pushes execution requests to another threa
```
yontyon, line 4683: Nit: threa -> thread

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5184,6 +5169,26 @@ void ReplicaImp::finalizeExecution() {
...
+      metric_fast_threshold_path_count_++;
+      break;
+    case CommitPath::SLOW:
+      metric_total_slowPath_++;
+      metric_total_slowPath_requests_ += numOfRequests;
+      break;
+    case CommitPath::NA:
+      break;
+  }
+  LOG_INFO(GL, "Commit path metric update " << KVLOG(std::to_string((int8_t)executionPath)));
```
yontyon, line 5189: move to DEBUG, we have the commit path as part of the MDC

bftengine/src/preprocessor/GlobalData.cpp
```diff
@@ -4,7 +4,4 @@ namespace preprocessor {
 
 std::atomic_uint64_t GlobalData::current_block_id = 0;
 std::atomic_uint64_t GlobalData::block_delta = 0;
-std::atomic_uint64_t GlobalData::delta_factor = 60;
-std::atomic_bool GlobalData::increment_step = false;
-std::atomic_bool GlobalData::decrement_step = false;
```
yontyon, line 9: ???

bftengine/src/preprocessor/GlobalData.hpp
```diff
@@ -6,9 +6,6 @@ struct GlobalData {
   // Used for the conflict detection optimization.
   static std::atomic_uint64_t current_block_id;
   static std::atomic_uint64_t block_delta;
-  static std::atomic_uint64_t delta_factor;
-  static std::atomic_bool increment_step;
-  static std::atomic_bool decrement_step;
```
yontyon, line 11: ???

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1605,7 +1605,6 @@ OperationResult PreProcessor::launchReqPreProcessing(const PreProcessRequestMsgS
              "Primary block [" << preProcessReqMsg->primaryBlockId()
                                << "] is too advanced for conflict detection optimization, replica block id ["
                                << GlobalData::current_block_id << "] delta [" << GlobalData::block_delta << "]");
-    GlobalData::increment_step = true;
```
yontyon, line 1608: ???

yontyon (commented):
LGTM!
One comment, I would like to have a separate PR for the Apollo improvements, is it possible?

bftengine/src/bftengine/ReplicaImp.cpp
WildFireFlum: This wasn't supposed to be committed, I'll remove it.

bftengine/src/preprocessor/GlobalData.cpp
WildFireFlum: These are not used anywhere in the code.

bftengine/src/preprocessor/GlobalData.hpp
WildFireFlum: Unused

bftengine/src/preprocessor/PreProcessor.cpp
WildFireFlum: unused

bftengine/src/preprocessor/GlobalData.cpp
yontyon: so move them to the other PR (the one with the overall fixes and changes)

PR #2309, opened 2022-02-04
State Snapshot Service - through ClientService

arc-vmware:
The state snapshot service is materialized through a set of 3 APIs.
These APIs are based on the service provided by Replica and Thin Replica
server. Major changes done are as below:
1) Initially Reconfiguration requests were not sendable through the concord client (which internally uses the concord client pool). This was never required as historically all the reconfiguration requests were usually sent through the operator or Client of CRE. But Replica is exposing the Snapshot Service APIs as a Reconfiguration Request and Response. So Concord client now had to add this feature as State Snapshot Service has to happen on ClientService. So this commit contains the addition of this capability on Concord Client.

2) The Thin Replica client was implemented such that only a single subscription request can happen. As subscription happens on infinite streams, and it allows the subscribed client to get future updates, we currently support a single subscription at a time. But state snapshot is a stream that represents historical states and thus it's a finite stream. So the current thin replica client is required to share the channel so that multiple parallel state snapshot streams can be read. This commit contains the addition of this capability in the Thin Replica Client.

3) The Thin Replica Client abstracts the GRPC constructs using TrcConnection. So TrcConnection should be considered as a tuple of a channel and a set of pairs of context and streams. As we are adding new streams, TrcConnection is not just related to TRC, instead, it becomes a new GRPC connection. So it's renamed to GrpcConnection. In the future, we might also keep it in a separate namespace. but in this commit, the  GrpcConnection is within the same thin replica namespace. A separate PR will be created for updating the namespace if required.

4) After adding the above capabilities, the State Snapshot Service is created and all the APIs of the state snapshot service are provided based on the concord pool and replica state snapshot service.


bftengine/include/bftengine/SimpleClient.hpp
```diff
@@ -45,6 +45,8 @@ enum ClientMsgFlag : uint8_t {
   HAS_PRE_PROCESSED_REQ = 0x4,
   KEY_EXCHANGE_REQ = 0x8,
   EMPTY_CLIENT_REQ = 0x10,
+  RECONFIG_FLAG_REQ = 0x20,       // Same as WRITE_ONLY_REQ | RECONFIG_FLAG_REQ
```
dartdart26, line 48: I think it might be a bit misleading stating `WRITE_ONLY_REQ` as it is non-existent. Also, the `FLAG` part is not consistent with names of other flags. What about:

- RECONFIG_WRITE_REQ
- RECONFIG_READ_REQ

What do you think?

client/client_pool/include/client/client_pool/concord_client_pool.hpp
```diff
@@ -168,6 +168,8 @@ class ConcordClientPool {
                          const concord::config_pool::ConcordClientPoolConfig&);
   void CreatePool(concord::config_pool::ConcordClientPoolConfig&);
 
+  void AddMissingReconfigurationFields(std::vector<uint8_t>& request, const ClientPtr& chosenClient);
```
dartdart26, line 171: Can we maybe state what we add? Is it the `signature`?

client/clientservice/include/client/clientservice/state_snapshot_service.hpp
```diff
@@ -37,6 +37,11 @@ class StateSnapshotServiceImpl final
                         vmware::concord::client::statesnapshot::v1::ReadAsOfResponse* response) override;
 
  private:
+  bool isHashValid(uint64_t snapshot_id,
```
dartdart26, line 40: Can we make this method `const`?

client/clientservice/include/client/clientservice/state_snapshot_service.hpp
```diff
@@ -37,6 +37,11 @@ class StateSnapshotServiceImpl final
...
+                   concord::util::SHA3_256::Digest& final_hash,
```
dartdart26, line 41: Can we make `final_hash` a const ref?

client/clientservice/include/client/clientservice/state_snapshot_service.hpp
dartdart26, line 40: Does it make sense to return the gRPC status directly, instead of using an output param?

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +27,405 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
 }
 
 Status StateSnapshotServiceImpl::StreamSnapshot(ServerContext* context,
                                                 const StreamSnapshotRequest* proto_request,
                                                 ServerWriter<StreamSnapshotResponse>* stream) {
-  // TODO: Add implementation
-  return grpc::Status(grpc::StatusCode::UNIMPLEMENTED, "StreamSnapshot");
+  Status status;
+
+  if ((proto_request != nullptr) && (proto_request->has_last_received_key())) {
```
dartdart26, line 197: I was wondering about two things:

1. Can `proto_request` ever be nullptr? If yes, what does that mean? If it cannot be, maybe we can assert instead?
2. The client might not send `last_received_key`. If I am not missing something, we need to allow that.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +27,405 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
+          return_status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "Snapshot not ready");
+          break;
+        case SnapshotResponseStatus::Success:
+          if ((res->data).snapshot_id != snapshot_id) {
+            return_status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "Snapshot id mismatch");
+          }
+          concord::util::SHA3_256::Digest replica_hash;
+          for (size_t i = 0; i < (res->data).hash.size(); ++i) {
+            replica_hash[i] = ((res->data).hash)[i];
+          }
```
dartdart26, line 319: I was wondering about the following here:

1. Aren't read-only reconfiguration responses containing data only in the replica-specific portion part (even without the signatures)? Or if the responses are the same (due to lack of signature), reconfiguration would put them in the common part?
2. Nit: Can we avoid copying the hash just to compare it?

client/concordclient/include/client/concordclient/concord_client_exceptions.hpp
```diff
@@ -40,4 +40,17 @@ class InternalError : public std::runtime_error {
  public:
   InternalError() : std::runtime_error("an internal error occurred"){};
 };
+
+// The stream ended
+class EndOfStream : public std::runtime_error {
```
dartdart26, line 45: I am thinking that using exceptions for control flow is not ideal - for example, here `EndOfStream` is a successful condition, yet it inherits from `runtime_error`. Is there a way for us to avoid using exceptions for signalling these conditions?

client/concordclient/include/client/concordclient/remote_update_recipient.hpp
```diff
@@ -46,6 +46,11 @@ struct Update {
   std::string span_context;
 };
 
-typedef std::variant<Update, EventGroup> EventVariant;
+struct SnapshotKVPair {
+  std::string key;
+  std::string val;
+};
+
+typedef std::variant<Update, EventGroup, SnapshotKVPair> RemoteRecepient;
```
dartdart26, line 54: I am thinking that this variant contains the types of data we transfer. Maybe we can rename it to `ClientData` or similar. What do you think?

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -111,6 +111,15 @@ struct ConcordClientConfig {
   SubscribeConfig subscribe_config;
 };
 
+struct StreamSnapshotRequest {
+  uint64_t snapshot_id;
+  std::string last_received_key;
+};
+struct StreamRequest {
+  // Using a veriant for type subsetting.
+  std::variant<StreamSnapshotRequest> request;
```
dartdart26, line 120: Could you please explain the rationale for using a variant?

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -159,18 +157,16 @@ class ConcordClient {
...
-
   // Stream a specific state snapshot in a resumable fashion as a finite stream of key-values.
   // Key-values are streamed with lexicographic order on keys.
-  void streamSnapshot(const StreamSnapshotRequest& request);
+  void readStream(const StreamRequest& request, std::shared_ptr<UpdateQueue>& remote_queue);
 
-  // Read the values of the given keys as of a specific state snapshot.
-  void readAsOf(const ReadAsOfRequest& request);
+  // Get participant_id
+  std::string getTRId() { return config_.subscribe_config.id; }
```
dartdart26, line 165: Nit: method can be `const`.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -159,18 +157,16 @@ class ConcordClient {
...
 
  private:
   config_pool::ConcordClientPoolConfig createClientPoolStruct(const ConcordClientConfig& config);
+  void CheckAndreConnectGrpcConnections();
```
dartdart26, line 169: Nit: Typo - `checkAndReConnectGrpcConnections()`.

client/concordclient/include/client/concordclient/remote_update_recipient.hpp
arc-vmware: There is an Inter Thread Communication (ITC) container. The name of that container is BasicUpdateQueue. That container, unlike vector, is not defined for all kinds of types. Instead, it's defined for a subset of types. This type represents those data types that the ITC can contain. The ITC contains a push method that blocks until the thread on the other end pops the data. So it's like a pipe abstraction. Now assume that there are 2 threads, the producer thread produces the data for the thread of the other end and the consumer will consume the data created in the other end. So this data acts as a recipient of remote data produced/consumed at a remote end. So it's named as RemoteRecepient, but if this name is not apt, then a better name should be RemoteData (if we want to view it as Data). But we should not attach it to the client as it's related to an ITC.

client/concordclient/include/client/concordclient/concord_client.hpp
arc-vmware: ConcordClient is a layer that formalizes the service provided by various clients. It provides an abstraction using Send(), Subscribe(), Unsubscribe() and readStream() APIs. Now let's assume that we have more clients which will provide a subscribe functionality on some endless stream. For all those subsets of types, SubscibeRequest will be used. Similarly, all those clients which provide Stream service (for finite streams) will be provided under the readStream() API. The StreamRequest will represent the subset of all those types. This variant is used for type subsetting of all such future usecases.

bftengine/include/bftengine/SimpleClient.hpp
arc-vmware: This enum represents the states of an 8-bit tuple. Each bit can represent 2 states. So there are 2^8 (256) states. 
Now in the current implementation, we have used ClientMsgFlag to represent those states and not an uint8_t.

The states are of 3 types:
1) Level - 0 states -- This has only one element which is represented by 0x0. This represents that LSB is 0 and it corresponds to WRITE ONLY request and it's mentioned here as EMPTY FLAG.
2) Level - 1 states -- those states which have only one bit 1 and the rest are all zeros.
3) Level - n states -- these states which have more than one place 1. e.g. 00100001 -- This represents ORed state of few of the Level - 1 or Level - 0 states. 
As we are using the enum type to represent the flag and the flag is not uint8_t, we have to enumerate that subset of states that we are going to use. We may not have to add all 256, as most of those we are not using.

Now here as 0x1 is READ ONLY state, 0x0 should represent WRITE ONLY state. So any Level - 1 state is an ORed state of Level - 0.
So as level - 1 is a vanilla state, RECONFIG_FLAG_REQ is used, but essentially it's RECONFIG WRITE REQ. This is true for all like KEY_EXCHANGE_REQ is actually KEY_EXCHANGE_WRITE_REQ. If in the future we have to enumerate the KEY_EXCHANGE_READ_REQ (which will be 0x9) then we should add a similar comment around KEY_EXCHANGE_REQ.  
That's why RECONFIG_WRITE_REQ will be written as a level-1 state and its name will be RECONFIG_FLAG_REQ, but RECONFIG_FLAG_REQ with READ is a level - 2 state and it will be named as RECONFIG_READ_ONLY_REQ.

client/client_pool/include/client/client_pool/concord_client_pool.hpp
arc-vmware: Its sender id and signature. So this method represents all those which need to be populated when the client is decided by CLient Pool.

client/concordclient/include/client/concordclient/concord_client_exceptions.hpp
arc-vmware: This is defined by the container created for ITC. The data is returned using the RemoteRecipient and the state is sent through Exception. So in the current scenario, this is as per implementation. I thought of changing that using a different way by adding a new status enum type variant in RemoteRecipient, but that will require major changes. We can take this in a separate PR.

client/clientservice/include/client/clientservice/state_snapshot_service.hpp
arc-vmware: Yes, I will add this in the next commit.

client/clientservice/include/client/clientservice/state_snapshot_service.hpp
arc-vmware: Thanks @dartdart26 , I will add this in next commit.

client/clientservice/src/state_snapshot_service.cpp
arc-vmware: Thanks, @dartdart26, yes this needs to be changed, I will update this in the next commit.

client/concordclient/include/client/concordclient/concord_client.hpp
arc-vmware: Thanks @dartdart26

client/concordclient/include/client/concordclient/concord_client.hpp
arc-vmware: thanks @dartdart26

client/thin-replica-client/test/replica_stream_snapshot_client_test.cpp
```diff
@@ -0,0 +1,290 @@
...
+              break;
+            } catch (const concord::client::concordclient::InternalError& e) {
+              break;
+            } catch (const concord::client::concordclient::StreamUnavailable& e) {
+              break;
+            } catch (const concord::client::concordclient::EndOfStream& e) {
+              break;
+            } catch (...) {
+              break;
+            }
```
dartdart26, line 256: Do we need to explicitly handle all types?

client/client_pool/include/client/client_pool/concord_client_pool.hpp
dartdart26: I see. We could still name it such that it indicates the fields it adds.

client/clientservice/src/state_snapshot_service.cpp
dartdart26: Thank you!

client/concordclient/include/client/concordclient/concord_client_exceptions.hpp
dartdart26: Cool, let's address separately. Thanks!

client/concordclient/include/client/concordclient/remote_update_recipient.hpp
dartdart26: I understand, thanks for the clarification. What I am suggesting is that we are sending some data in the variant and the value is not the recipient itself - instead, it is the data.

client/concordclient/include/client/concordclient/concord_client.hpp
dartdart26: Can we use function overloading and define `StreamSnapshotRequest` and future types? I think it might be easier to use and understand.

client/thin-replica-client/test/replica_stream_snapshot_client_test.cpp
dartdart26: How can it know which exceptions are about to be thrown? It seems like a lot of boilerplate to me, would be nice to avoid it , IMO.

dartdart26:
Looks very good, thanks @arc-vmware! I am approving it, but since I am not a gRPC/ThinReplica expert, I think it would make sense for @shgandhi or @robem to have a look.

client/concordclient/include/client/concordclient/remote_update_recipient.hpp
arc-vmware: Thanks, @dartdart26 I got it, let me rename this as RemoteData.

client/concordclient/include/client/concordclient/concord_client.hpp
arc-vmware: Thanks, @dartdart26, let me do this with function overloading. Actually, function overloading has a drawback, that the boilerplate code either needs to be re-written for each type, but using variant type subsetting there will be a single boilerplate code in the dispatcher module and the dispatcher will call the different functions. I am fine with either way as this might reduce the complexity of each new overload.

client/client_pool/include/client/client_pool/concord_client_pool.hpp
arc-vmware: Thanks @dartdart26 , I changed the function name to AddSenderAndSignature.

client/concordclient/include/client/concordclient/remote_update_data.hpp
```diff
@@ -46,6 +46,11 @@ struct Update {
   std::string span_context;
 };
 
-typedef std::variant<Update, EventGroup> EventVariant;
+struct SnapshotKVPair {
+  std::string key;
+  std::string val;
+};
+
+typedef std::variant<Update, EventGroup, SnapshotKVPair> RemoteData;
```
robem, line 54: Please **do not** mix any data structures with what is used for the TRC.
The only shared code between the snapshot client and the TRC should be the gRPC connection.
Concord client will initialize both objects separately and expose their APIs separately. Meaning, the types should be separate as well. Otherwise, this will become a maintenance burden if we change one or the other.

If you choose to use the same update mechanism then I suggest to copy the types and adjust them to the state snapshot needs.
However, I was under the impression that we can simply forward the incoming data from the replicas and hand it over to the clientservice via a stream object without even looking at the data. Why do we need `SnapshotKVPair` in the first place?

Also, the TRC and the snapshot client would never share the same queue. Meaning, there is no need to have a distinction between the types of the elements at runtime.

client/concordclient/src/concord_client.cpp
```diff
@@ -35,6 +38,20 @@ ConcordClient::ConcordClient(const ConcordClientConfig& config, std::shared_ptr<
     LOG_INFO(logger_, "Waiting for client pool to connect");
     std::this_thread::sleep_for(std::chrono::seconds(2));
   }
+  for (const auto& replica : config_.topology.replicas) {
+    auto addr = replica.host + ":" + std::to_string(replica.event_port);
+    auto grpc_conn =
+        std::make_shared<GrpcConnection>(addr, config_.subscribe_config.id, /* TODO */ 3, /* TODO */ 3, /* TODO */ 5);
+
+    // TODO: Adapt TRC API to support PEM buffers
+    auto trsc_config = std::make_unique<TrsConnectionConfig>(config_.subscribe_config.use_tls,
```
robem, line 47: It doesn't make sense to initialize a GrpcConnection with a TrsConnectionConfig. Please rename the latter as well.

client/concordclient/src/concord_client.cpp
```diff
@@ -117,6 +134,18 @@ void ConcordClient::send(const bft::client::WriteConfig& config,
   client_pool_->SendRequest(config, std::forward<bft::client::Msg>(msg), callback);
 }
 
+void ConcordClient::CheckAndReConnectGrpcConnections() {
```
robem, line 137: Lower-case function names.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -59,35 +62,46 @@ const int kGrpcMaxInboundMsgSizeInBytes = 1 << 24;
...
-  virtual ~TrsConnection() { this->disconnect(); }
+        hash_timeout_(std::chrono::seconds(hash_operation_timeout_seconds)),
+        snapshot_timeout_(std::chrono::seconds(snapshot_operation_timeout_seconds)) {}
+  // If the snapshot operation is not needed, then it will not be provided and default of
+  // 5 sec will be taken.
+  GrpcConnection(const std::string& address,
+                 const std::string& client_id,
+                 uint16_t data_operation_timeout_seconds,
+                 uint16_t hash_operation_timeout_seconds)
+      : GrpcConnection(address, client_id, data_operation_timeout_seconds, hash_operation_timeout_seconds, 5) {}
```
robem, line 97: Why do we need a new constructor? Instead of setting `snapshot_operation_timeout_seconds=5`?

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -59,35 +62,46 @@ const int kGrpcMaxInboundMsgSizeInBytes = 1 << 24;
...
+
+  virtual ~GrpcConnection() { this->disconnect(); }
 
   // Connect & disconnect from the TRS
   virtual void connect(std::unique_ptr<TrsConnectionConfig>& config);
```
robem, line 102: As mentioned above, a GrpcConnection should connect with the help of a GrpcConnectionConfig.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -154,24 +168,29 @@ class TrsConnection {
...
-  friend std::ostream& operator<<(std::ostream& os, const TrsConnection& trsc) {
+  friend std::ostream& operator<<(std::ostream& os, const GrpcConnection& trsc) {
     return os << trsc.client_id_ << " (" << trsc.address_ << ")";
   }
 
  protected:
   // Helper function to deal with gRPC
-  void createStub();
+  void createTRCStub();
+  void createRSSStub();
```
robem, line 192: According to the coding standard, the names should be `createTrcStub` and `createRssStub`.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -187,16 +206,25 @@ class TrsConnection {
...
+           std::pair<std::unique_ptr<grpc::ClientContext>,
+                     std::unique_ptr<
+                         grpc::ClientReaderInterface<vmware::concord::replicastatesnapshot::StreamSnapshotResponse>>>>
+      rss_streams_;
+  std::mutex rss_streams_mutex_;
+  std::atomic_uint64_t current_req_id_{1};
 
   // gRPC connection
   std::shared_ptr<grpc::Channel> channel_;
   std::unique_ptr<com::vmware::concord::thin_replica::ThinReplica::StubInterface> stub_;
```
robem, line 219: I suggest to rename this to `trc_stub_`.

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
```diff
@@ -0,0 +1,79 @@
...
+        threadpool_(config_->concurrency_level) {}
+  void readSnapshotStream(const SnapshotRequest& request,
+                          std::shared_ptr<concord::client::concordclient::UpdateQueue> remote_queue);
+
+ private:
+  // Thread function to start subscription_thread_ with with snapshot.
+  void receiveSnapshot(const SnapshotRequest& request,
+                       std::shared_ptr<concord::client::concordclient::UpdateQueue> remote_queue);
+
+  void pushFinalStateToRemoteQueue(const thin_replica_client::GrpcConnection::Result& result);
```
robem, line 69: Maybe move the GrpcConnection object into the `client::concordclient` namespace.

client/thin-replica-client/src/grpc_connection.cpp
```diff
@@ -356,26 +374,131 @@ TrsConnection::Result TrsConnection::readHash(Hash* hash) {
...
 }
-TrsConnection::Result TrsConnection::readStateSnapshot(vmware::concord::replicastatesnapshot::KeyValuePair* key_value) {
-  // TODO: Add implementation
-  ConcordAssert("readStateSnapshot should not be called. It is unimplemented." && false);
+
+GrpcConnection::Result GrpcConnection::readStateSnapshot(
+    RequestId request_id, vmware::concord::replicastatesnapshot::StreamSnapshotResponse* snapshot_response) {
+  auto result = async(launch::async, [this, snapshot_response, request_id] {
+    std::lock_guard lock(this->rss_streams_mutex_);
+    return (((this->rss_streams_)[request_id]).second)->Read(snapshot_response);
```
robem, line 469: We are holding the lock on **ALL** stream objects for the `Read` call on a single stream object. If the call times out then we were holding the lock for default 5 seconds. Obviously, we need to prevent anyone from cleaning up the stream object but isn't there a better way? Maybe it won't be a practical issue but holding a lock for 5s on all streams doesn't seem Ok to me.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -30,6 +31,8 @@ using namespace std::chrono_literals;
 
 namespace client::thin_replica_client {
 
+using RequestId = uint64_t;
```
robem, line 34: This has nothing to do with the TRC and hence should be defined outside now. Similar to the GrpcConnection object, I suggest to put it under the `client::concordclient` namespace.

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
```diff
@@ -0,0 +1,79 @@
...
+};
+
+class ReplicaStreamSnapshotClient {
+ public:
+  ReplicaStreamSnapshotClient(std::unique_ptr<ReplicaStateSnapshotClientConfig> config)
+      : logger_(logging::getLogger("concord.client.replica_stream_snapshot")),
+        config_(std::move(config)),
+        threadpool_(config_->concurrency_level) {}
+  void readSnapshotStream(const SnapshotRequest& request,
+                          std::shared_ptr<concord::client::concordclient::UpdateQueue> remote_queue);
```
robem, line 62: Why are we using an update queue here and not a stream of state snapshot responses?
If you decide to use a queue then it should be a queue of snapshot responses and not a queue of variants.

robem (dismissed):
Did you test your change in an end-to-end deployment?

client/client_pool/src/concord_client_pool.cpp
```diff
@@ -67,13 +68,20 @@ SubmitResult ConcordClientPool::SendRequest(std::vector<uint8_t> &&request,
     }
     if (0 == seq_num) {
       seq_num = client->generateClientSeqNum();
+      if (flags & ClientMsgFlag::RECONFIG_FLAG_REQ) {
+        correlation_id += ("-" + std::to_string(seq_num));
```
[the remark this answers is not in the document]
arc-vmware: This is done as it's in operator code. I am not giving here the reference from GitLab, but I will ping you this specific comment and discuss a better way.

client/concordclient/include/client/concordclient/remote_update_data.hpp
arc-vmware: Thanks @robem for setting this discussion. My view is a little different here. I want to explain my reasoning for this using the concept of Type Subset.
Consider that I am building a container (like vector) for some reason. Normally all the STL containers are created for all kinds of types. These containers are close. So we can create compositions of containers of any type once we have a generic container. These containers are generally created as template,
template<typename T>
class Container{...};
So above the container is a generic container.
But sometimes for some reason (which i don't know in this case), containers are created for a subset of types. In such cases the container is not for everything. There are many ways to define such a container. One way is to define the container on top of a variant as below:
typedef variant<T1, T2> DataType;
class Container {
DataType dt;
...
};
Now, if for some valid reason, this Container needs to use the third type then adding the same in the variant is needed as this Container is designed on the subset of the type.
For Inter Thread Communication (ITC) we have a subset type container called BasicUpdateQueue. This container is exactly required for the ITC. So this data type is used here.
Also, every snapshot request is creating a new queue and provides the same to read the snapshot, so the queue object is not shared with TRC and Snapshot. Its the Container that is shared.

client/concordclient/src/concord_client.cpp
arc-vmware: Thanks @robem ,I will change this type aswell.

client/concordclient/src/concord_client.cpp
arc-vmware: Thanks, @robem , I will do this in my next commit.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
arc-vmware: thanks, @robem I will do this and keep one ctor.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
arc-vmware: Thanks, @robem, this I will change in my next commit.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
arc-vmware: thanks @robem, I will change this in next commit.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
arc-vmware: thanks @robem, I will change this in next commit.

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
arc-vmware: thanks @robem, I will change this in the next commit.

client/thin-replica-client/src/grpc_connection.cpp
arc-vmware: This will depend on the load. If we will see that there are many clients accessing and there are many timeouts then this will be a common scenario. But as for most of the call Read is a read of one datum, it will not be 5s and there should not be an issue in practice.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
arc-vmware: thanks @robem, I will change this in next commit.

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
arc-vmware: It's a queue of RemoteData and this is done for the subset of types. So I think there should be no issues with this. Can you tell me any problem that you foresee if we reuse this queue as this queue is purely a data structure for Inter Thread Communication which is applicable for a subset of types? And it should be fine if that subset of type contains different data structures.

client/concordclient/include/client/concordclient/remote_update_data.hpp
arc-vmware: Thanks, @robem for this comment. I have removed the type subsetting in the container and then removed SnapshotKVPair from EventVariant. Now everything is consistent.

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
arc-vmware: Thanks, @robem for this comment. I have removed the type subsetting in the container and then removed SnapshotKVPair from EventVariant. Now everything is consistent.

bftengine/include/bftengine/SimpleClient.hpp
yuliasherman: Reference to a non-existing flag WRITE_ONLY_REQ confuses me, too.

client/client_pool/src/concord_client_pool.cpp
```diff
@@ -221,8 +237,10 @@ SubmitResult ConcordClientPool::SendRequest(const bft::client::ReadConfig &confi
     callback(bftEngine::SendResult{static_cast<uint32_t>(OperationResult::INVALID_REQUEST)});
     return SubmitResult::Overloaded;
   }
+  auto request_flag = ClientMsgFlag::READ_ONLY_REQ;
+  if (config.request.reconfiguration) request_flag = ClientMsgFlag::RECONFIG_READ_ONLY_REQ;
   return SendRequest(std::forward<std::vector<uint8_t>>(request),
-                     ClientMsgFlag::READ_ONLY_REQ,
+                     request_flag,
```
yuliasherman, line 243: Simplify: 
auto request_flag = config.request.reconfiguration ? ClientMsgFlag::RECONFIG_READ_ONLY_REQ : ClientMsgFlag::READ_ONLY_REQ;

client/client_pool/src/external_client.cpp
```diff
@@ -316,6 +317,8 @@ void ConcordClient::setDelayFlagForTest(bool delay) { ConcordClient::delayed_beh
 
 OperationResult ConcordClient::getRequestExecutionResult() { return clientRequestExecutionResult_; }
 
+std::string ConcordClient::messageSignature(bft::client::Msg& message) { return new_client_->signMessage(message); }
```
yuliasherman, line 320: We have already signed client requests. How is this different from the current approach?

client/client_pool/src/concord_client_pool.cpp
```diff
@@ -361,6 +379,18 @@ void ConcordClientPool::CreatePool(concord::config_pool::ConcordClientPoolConfig
   jobs_queue_max_size_ = config.external_requests_queue_size;
 }
 
+void ConcordClientPool::AddSenderAndSignature(std::vector<uint8_t> &request, const ClientPtr &chosenClient) {
+  concord::messages::ReconfigurationRequest rreq;
+  concord::messages::deserialize(request, rreq);
+  rreq.sender = static_cast<decltype(rreq.sender)>(chosenClient->getClientId());
```
yuliasherman, line 385: The sender-id is a part of the ClientRequestMsgHeader, why do you need it inside the request, too?

bftengine/include/bftengine/SimpleClient.hpp
arc-vmware: Ok @yuliasherman, I will remove that from the comment. But all the level-1 flags are having WRITE_ONLY_REQ flag embedded in them by default.

client/client_pool/src/concord_client_pool.cpp
arc-vmware: Thanks @yuliasherman , I will add this in next commit.

client/client_pool/src/external_client.cpp
arc-vmware: In https://github.com/vmware/concord-bft/blob/0bb1834b467342c5f449a3508975c2cc05ca8d28/reconfiguration/cmf/concord.cmf#L419
We can see the following declaration (snipped for easy reference)
Msg ReconfigurationRequest 1 {
  uint64 sender
  bytes signature
  oneof {
   <many reconfiguration commands>
  } command
  bytes additional_data
}

The above request is added with a jacket of ClientRequestMsg (or the above reconfiguraton request is a payload to ClientRequestMsg). Currently when creating ClientRequestMsg, one signature is added by default. But the dispatcher of Reconfiguration message gets the ReconfigurationRequest after the jacket is peeled off in ReplicaImp.  This dispatcher expects the signature of this payload in the signature field and thus we have to add this signature.
Dispatcher: https://github.com/vmware/concord-bft/blob/0bb1834b467342c5f449a3508975c2cc05ca8d28/reconfiguration/src/dispatcher.cpp#L38
So we have to add signature in the AddSenderAndSignature() function for ReconfigurationRequest when the actual client is decided by the client_pool.

client/client_pool/src/concord_client_pool.cpp
arc-vmware: In

[concord-bft/reconfiguration/cmf/concord.cmf](https://github.com/vmware/concord-bft/blob/0bb1834b467342c5f449a3508975c2cc05ca8d28/reconfiguration/cmf/concord.cmf#L419)

Line 419 in [0bb1834](https://github.com/vmware/concord-bft/commit/0bb1834b467342c5f449a3508975c2cc05ca8d28)

 Msg ReconfigurationRequest 1 { 

We can see the following declaration (snipped for easy reference)
Msg ReconfigurationRequest 1 {
uint64 sender
bytes signature
oneof {
\<many reconfiguration commands\>
} command
bytes additional_data
}

Reconfiguration request contains an additional sender id field. So we had to add signature and sender at that point when client_pool picks the actual client (external client --> bft client)

client/client_pool/src/concord_client_pool.cpp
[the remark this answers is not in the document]
robem: I still don't see why we should make a special case for the reconfiguration request here. However, if @yuliasherman is OK with it then I'm fine too.

robem (commented):
I still struggle with the queue. Why are we using a separate queue to forward the snapshot updates/kv-pairs? The snapshot client doesn't do anything with the kv-pairs but put it onto the queue. Meaning, why do we go with this intermittent step/memory instead of forwarding it straight to the caller?
We could define a thin "stream" object around the underlying `StreamSnapshot()` invocation that represents a single active subscription stream and hand it over to the caller. In this case, we don't have to read from the stream ourselves but only the caller.

arc-vmware:
thanks @robem for your comments. We are using the queue as the abstraction is available. I agree that we should use the Stream approach that you explained. I will raise a JIRA and work on that, for now, we will go ahead with this approach of using the queue as it's easier to implement. We will definitely create the stream abstraction in the next phase.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -159,18 +153,16 @@ class ConcordClient {
...
-
   // Stream a specific state snapshot in a resumable fashion as a finite stream of key-values.
   // Key-values are streamed with lexicographic order on keys.
-  void streamSnapshot(const StreamSnapshotRequest& request);
+  void readStream(const StreamSnapshotRequest& request, std::shared_ptr<StreamUpdateQueue>& remote_queue);
 
-  // Read the values of the given keys as of a specific state snapshot.
-  void readAsOf(const ReadAsOfRequest& request);
+  // Get participant_id
+  std::string getTRId() const { return config_.subscribe_config.id; }
```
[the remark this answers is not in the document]
arc-vmware: Thanks, @robem , let me change this.

client/concordclient/include/client/concordclient/remote_update_data.hpp
```diff
@@ -48,4 +48,9 @@ struct Update {
 
 typedef std::variant<Update, EventGroup> EventVariant;
 
+struct SnapshotKVPair {
```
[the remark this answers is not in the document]
arc-vmware: thanks, @robem , I will create snapshot_update.hpp and keep the event_update.hpp as it is since it doesn't require any change.

client/concordclient/include/client/concordclient/remote_update_queue.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include "assertUtils.hpp"
+
+#include "client/concordclient/remote_update_data.hpp"
+
+namespace concord::client::concordclient {
+
+// An implementation of RemoteUpdateQueue should guarantee that ReleaseConsumers, Clear, Push, Pop, and TryPop are all
+// mutually thread safe.
+template <typename T>
+class RemoteUpdateQueue {
```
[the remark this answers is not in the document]
arc-vmware: Thanks, @robem, as UpdateQueue is a data structure required for Inter Thread Communication, this name is changed. A better name for this queue should be LockLess or ThreadSafeQueue and not UpdateQueue. 
RemoteUpdateQueue is used since it signifies that a remotely running thread can Update this Queue and synchronize with other threads.
Please let me know what you think, can I name it as ThreadSafeQueue?

client/concordclient/include/client/concordclient/remote_update_queue.hpp
```diff
@@ -0,0 +1,129 @@
...
+  virtual void releaseConsumers() override;
+  virtual void clear() override;
+  virtual void push(std::unique_ptr<T> update) override;
+  virtual std::unique_ptr<T> pop() override;
+  virtual std::unique_ptr<T> tryPop() override;
+  virtual uint64_t size() override;
+  virtual void setException(std::exception_ptr e) override;
+};
+
+using StreamUpdateQueue = RemoteUpdateQueue<SnapshotKVPair>;
```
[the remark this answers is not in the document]
arc-vmware: Thanks @robem , I will update the name.

client/concordclient/src/concord_client.cpp
```diff
@@ -117,6 +134,18 @@ void ConcordClient::send(const bft::client::WriteConfig& config,
...
+                                                                config_.subscribe_config.pem_private_key,
+                                                                config_.subscribe_config.pem_cert_chain,
+                                                                config_.transport.event_pem_certs);
+      (grpc_connections_[con_offset])->reConnect(trsc_config);
+    }
+  }
+}
+
 void ConcordClient::subscribe(const SubscribeRequest& sub_req,
                               std::shared_ptr<UpdateQueue>& queue,
```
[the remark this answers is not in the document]
arc-vmware: thanks @robem , I will refactor.

client/concordclient/include/client/concordclient/remote_update_queue.hpp
```diff
@@ -0,0 +1,129 @@
...
+using UpdateQueue = RemoteUpdateQueue<EventVariant>;
+
+using BasicStreamUpdateQueue = BasicRemoteUpdateQueue<SnapshotKVPair>;
```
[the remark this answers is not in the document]
arc-vmware: Thanks, @robem, I will keep separate files for each of these.

client/concordclient/include/client/concordclient/remote_update_queue.ipp
```diff
@@ -1,6 +1,6 @@
 // Concord
 //
-// Copyright (c) 2020-2021 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2021-2022 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
arc-vmware: thanks, yes I will remove this change.

client/concordclient/src/remote_update_queue.cpp
```diff
@@ -0,0 +1,16 @@
...
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include "client/concordclient/remote_update_queue.hpp"
+
+// Empty file created for concordclient-event-api
```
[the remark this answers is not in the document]
arc-vmware: Yes, but these will be in the header file I guess as they are some type-computations. Let me see if this empty cpp file can be filled with something other than the include.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -159,18 +153,16 @@ class ConcordClient {
   // will be occupied forever.
   void unsubscribe();
 
-  // Get the ID of a recent and available state snapshot
-  void getRecentSnapshot(const GetRecentSnapshotRequest& request);
-
   // Stream a specific state snapshot in a resumable fashion as a finite stream of key-values.
   // Key-values are streamed with lexicographic order on keys.
-  void streamSnapshot(const StreamSnapshotRequest& request);
+  void readStream(const StreamSnapshotRequest& request, std::shared_ptr<StreamUpdateQueue>& remote_queue);
```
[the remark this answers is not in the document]
arc-vmware: I will change this to getSnapshot()

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+// TODO: Add metrics
+// TODO: Add more comments
+
+struct SnapshotRequest {
+  uint64_t snapshot_id;
+  std::string last_received_key;
+};
+
+class ReplicaStreamSnapshotClient {
```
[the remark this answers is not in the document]
arc-vmware: Yes I will change it. thanks @robem

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include "Metrics.hpp"
+
+#include <opentracing/span.h>
+#include <condition_variable>
+#include <thread>
+#include "Logger.hpp"
+#include "client/concordclient/remote_update_queue.hpp"
+#include "client/thin-replica-client/grpc_connection.hpp"
+
+namespace client::replica_state_snapshot_client {
```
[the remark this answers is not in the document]
arc-vmware: ok I will change the filename, thanks @robem

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
```diff
@@ -0,0 +1,78 @@
...
+  uint32_t concurrency_level;
+  ReplicaStateSnapshotClientConfig(std::vector<std::shared_ptr<client::concordclient::GrpcConnection>>& rss_conns_,
+                                   uint32_t concurrency_level_)
+      : rss_conns(rss_conns_), concurrency_level(concurrency_level_) {}
+};
+
+// TODO: Add metrics
+// TODO: Add more comments
+
+struct SnapshotRequest {
```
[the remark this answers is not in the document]
arc-vmware: Ok, this is a good suggestion, I will remove this type. Thanks @robem

client/thin-replica-client/src/replica_stream_snapshot_client.cpp
```diff
@@ -0,0 +1,122 @@
...
+    replica_id++;
+  }
+  pushFinalStateToRemoteQueue(result);
+}
+void ReplicaStreamSnapshotClient::pushDatumToRemoteQueue(const StreamSnapshotResponse& datum,
+                                                         std::shared_ptr<StreamUpdateQueue> remote_queue,
+                                                         std::string& last_key) {
+  if (datum.has_key_value()) {
+    last_key.assign(datum.key_value().key());
+    auto snapshot_datum = std::unique_ptr<SnapshotKVPair>{new SnapshotKVPair{last_key, datum.key_value().value()}};
```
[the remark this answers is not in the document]
arc-vmware: As datum is const ref, we are not reassigning the ownership in this function. This API will just push one datum and return it to the caller.

client/thin-replica-client/include/client/thin-replica-client/replica_stream_snapshot_client.hpp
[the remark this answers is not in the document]
arc-vmware: Concord client depends on replica state snapshot, but this will lead to a reverse dependency. So I am not doing this now, we will take this later. Thanks for the comment.

client/concordclient/src/remote_update_queue.cpp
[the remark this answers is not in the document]
arc-vmware: I am getting the following error when I remove this file: add_library(concordclient-event-api STATIC)
CMake Error at client/concordclient/CMakeLists.txt:12 (add_library):
  No SOURCES given to target: concordclient-event-api
So for now I am keeping this file and generating the event update interface.
I tried with 
add_library(concordclient-event-api INTERFACE)
But in that case its giving me the following error:
CMake Error at client/concordclient/CMakeLists.txt:13 (target_include_directories):
  target_include_directories may only set INTERFACE properties on INTERFACE
  targets


CMake Error at client/concordclient/CMakeLists.txt:14 (target_link_libraries):
  INTERFACE library can only be used with the INTERFACE keyword of
  target_link_libraries

So it seems with a cmake target with STATIC library, a source file is mandatory.

client/concordclient/include/client/concordclient/remote_update_queue.hpp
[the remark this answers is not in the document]
arc-vmware: This is a good point, what is a good name for this class, that you suggest. In my opinion, UpdateQueue is also not nice, but I will go with your suggestion. In the interface comments, we can see that it's saying that the Implementation should guarantee thread safety as its purpose is to use it in concurrently running threads. So I thought ThreadSafeQueue might be better so that the implementor can always make a thread-safe implementation. But again it would be good if you can provide a better name for this with a second thought.

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -179,7 +171,7 @@ class ConcordClient {
   // TODO: Allow multiple subscriptions
   std::atomic_bool active_subscription_{false};
 
-  std::shared_ptr<BasicUpdateQueue> trc_queue_;
```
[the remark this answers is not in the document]
robem: Oh, wow, ok, thanks!

client/concordclient/include/client/concordclient/remote_update_queue.hpp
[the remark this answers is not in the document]
arc-vmware: thanks, @robem , I am keeping the name of the interface as IQueue.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +28,388 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
+  cmd.participant_id = client_->getSubscriptionId();
+  for (int i = 0; i < proto_request->keys_size(); ++i) {
+    cmd.keys.push_back(proto_request->keys(i));
+  }
+  rreq.command = cmd;
+  bft::client::Msg message;
+  concord::messages::serialize(message, rreq);
+  client_->send(read_config, std::move(message), callback);
+  for (const auto& res : read_as_of_responses) {
+    if (return_status.ok()) {
```
robem, line 362: The following code gets nested into 8 levels!! This is hard to keep track of when reading the code. Given that we `break` if the return status is not ok, I recommend to do this right here to get rid of one level of nesting.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +28,388 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
+        case SnapshotResponseStatus::InternalError:
+          return_status = grpc::Status(grpc::StatusCode::UNKNOWN, "Internal error");
+          break;
+        case SnapshotResponseStatus::SnapshotNonExistent:
+          return_status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "Snapshot doesn't exist");
+          break;
+        case SnapshotResponseStatus::SnapshotPending:
+          return_status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "Snapshot not ready");
+          break;
+        case SnapshotResponseStatus::Success: {
```
robem, line 373: Why is there a `{` at the end?

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +28,388 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
+        break;
+    }
+    return;
+  }
+  return_status = grpc::Status::OK;
+  auto reply = std::get<bft::client::Reply>(send_result);
+  if (is_rsi) {
+    ReconfigurationResponse res;
+    concord::messages::deserialize(reply.matched_data, res);
+    response.response.template emplace<ResponseT>(std::get<ResponseT>(res.response));
```
robem, line 132: Interesting, I haven't seen this before. I assume you're calling `emplace` on the variant but `template` looks odd to me. How does this work?

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +28,388 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
+              if (set_values) {
+                auto* resp_val = response->add_values();
+                if (v.has_value()) {
+                  resp_val->set_value(v.value());
+                }
+              } else {
+                if (v.has_value()) {
+                  if (v.value() != response->values(pos).value()) {
+                    return_status =
+                        grpc::Status(grpc::StatusCode::UNKNOWN, "Got different values from different replicas");
```
robem, line 391: The whole for-loop is a bit magic to me given that the variable names are as generic as it gets. I understand that we loop over all responses and then check if we already have a response "at that position". What is the position? Is it the number of the kv-pairs? Also, how can you tell that if we already have a value at that position that the current response value must be from a different replica?

client/thin-replica-client/include/client/thin-replica-client/replica_state_snapshot_client.hpp
```diff
@@ -0,0 +1,78 @@
...
+ public:
+  ReplicaStateSnapshotClient(std::unique_ptr<ReplicaStateSnapshotClientConfig> config)
+      : logger_(logging::getLogger("concord.client.replica_stream_snapshot")),
+        config_(std::move(config)),
+        threadpool_(config_->concurrency_level) {}
+  void readSnapshotStream(const SnapshotRequest& request,
+                          std::shared_ptr<concord::client::concordclient::SnapshotQueue> remote_queue);
+
+ private:
+  // Thread function to start subscription_thread_ with with snapshot.
```
robem, line 64: "with with"

client/thin-replica-client/src/grpc_connection.cpp
```diff
@@ -356,27 +374,148 @@ TrsConnection::Result TrsConnection::readHash(Hash* hash) {
...
+      (it->second).second.reset();
+    }
+  }
+  {
+    WriteLock write_lock(rss_streams_mutex_);
+    rss_streams_.erase(request_id);
+  }
+}
+
+void GrpcConnection::cancelStateSnapshotStream() {
```
robem, line 446: You are cancelling all snapshot streams. I recommend to reflect that in the function name via `all` or using `Stream*s*`

client/concordclient/include/client/concordclient/snapshot_update.hpp
```diff
@@ -0,0 +1,30 @@
+// Concord
+//
+// Copyright (c) 2021-2022 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
arc-vmware: thanks @robem

client/concordclient/src/thread_safe_queue.cpp
```diff
@@ -0,0 +1,16 @@
...
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include "client/concordclient/event_update.hpp"
+
+// Empty file created for concordclient-event-api
```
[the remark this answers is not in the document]
arc-vmware: Yes @robem , I have answered this comment earlier.
I am getting the following error when I remove this file: add_library(concordclient-event-api STATIC)
CMake Error at client/concordclient/CMakeLists.txt:12 (add_library):
No SOURCES given to target: concordclient-event-api
So for now I am keeping this file and generating the event update interface.
I tried with
add_library(concordclient-event-api INTERFACE)
But in that case its giving me the following error:
CMake Error at client/concordclient/CMakeLists.txt:13 (target_include_directories):
target_include_directories may only set INTERFACE properties on INTERFACE
targets
CMake Error at client/concordclient/CMakeLists.txt:14 (target_link_libraries):
INTERFACE library can only be used with the INTERFACE keyword of
target_link_libraries
So it seems with a cmake target with STATIC library, a source file is mandatory.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +28,388 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
+  client_->getSnapshot(request, update_queue);
+
+  Status status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "Service not available");
+  bool is_end_of_stream = false;
+  auto accumulated_hash = singleHash(std::string{});
+  while (!context->IsCancelled()) {
+    StreamSnapshotResponse response;
+    std::unique_ptr<SnapshotKVPair> update;
+    try {
+      update = update_queue->pop();
```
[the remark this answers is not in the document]
arc-vmware: This will block until the timeout, since the GrpcConnection object times out and gives a timeout result. Trypop is better for updates and pop is better for stream reads.

client/thin-replica-client/src/grpc_connection.cpp
```diff
@@ -356,27 +374,148 @@ TrsConnection::Result TrsConnection::readHash(Hash* hash) {
...
+    auto it = rss_streams_.find(request_id);
+    if (it != rss_streams_.end()) {
+      if (!(it->second).second) {
+        ConcordAssertEQ((it->second).first, nullptr);
+        return;
+      }
+      ConcordAssertNE((it->second).first, nullptr);
+      (it->second).first->TryCancel();
+      (it->second).first.reset();
+      (it->second).second.reset();
```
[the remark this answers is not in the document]
arc-vmware: These streams are for each thread. The read and write locks are for the rss_streams object. So if we do any read and write operation on the rss_streams map object, then that needs to be synchronized. But once we get the object until we do rss_streams_ ->erase(it) or something like emplace() we don't have this issue.

client/thin-replica-client/src/grpc_connection.cpp
```diff
@@ -356,27 +374,148 @@ TrsConnection::Result TrsConnection::readHash(Hash* hash) {
...
+    ReadLock read_lock(rss_streams_mutex_);
+    for (auto& c : rss_streams_) {
+      if (!c.second.second) {
+        ConcordAssertEQ(c.second.first, nullptr);
+        continue;
+      }
+      ConcordAssertNE(c.second.first, nullptr);
+      c.second.first->TryCancel();
+      c.second.first.reset();
+      c.second.second.reset();
```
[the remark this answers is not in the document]
arc-vmware: Yes this is not an issue as streams are not shares across the threads, its the rss_streams_ map that is shared.

client/thin-replica-client/src/grpc_connection.cpp
```diff
@@ -356,27 +374,148 @@ TrsConnection::Result TrsConnection::readHash(Hash* hash) {
...
+  ConcordAssert(status == future_status::ready);
+  if (result.get()) {
+    return Result::kSuccess;
+  }
+
+  {
+    ReadLock read_lock(rss_streams_mutex_);
+    auto it = rss_streams_.find(request_id);
+    if (it != rss_streams_.end()) {
+      auto grpc_status = ((it->second).second)->Finish();
```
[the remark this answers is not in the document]
arc-vmware: This is the same reason, operation on individual streams need not be requiring a write lock, since we have one object per thread.

client/concordclient/include/client/concordclient/thread_safe_queue.hpp
```diff
@@ -0,0 +1,121 @@
+// Concord
+//
+// Copyright (c) 2020-2021 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
arc-vmware: for all code that we touch, I think we don't change the year, we only change if we create new files. So i am not changing the years here.

client/clientservice/src/state_snapshot_service.cpp
arc-vmware: Thanks @robem , I removed the brace.

client/clientservice/src/state_snapshot_service.cpp
arc-vmware: Hey @robem, As this is a template function all the types are not deduced. So we use template type computation here to call the correct emplace. You need to read template metaprogramming or template functional programming book (i don't remember the exact name). This concept should also be available on the internet and in courses of advanced C++.

client/thin-replica-client/src/grpc_connection.cpp
arc-vmware: thanks @robem, very good comments, let me change this.

client/clientservice/src/state_snapshot_service.cpp
arc-vmware: Thanks, @robem, for the question. Here we get the rsi from the bft client. Each replica sends its response. This response contains a signature. The response is defined in the cmf file and @dartdart26 can explain the reason for the structure of the response.
In this loop, there are 2 parts. One part assigns the first value from the rsi. If the bool set_value is true, it means we have to set the values as the value size for response is 0. So this is the first time. So we set the values. After that set_value will be false as the values are already set. Then we iterate to check if the values set from the first entry of rsi are matching with the other rsis. rsi == Replica Specific Information.
I can see the reason for your confusion, the set_value means we have to set the values. So I am changing the name of this variable to have_to_set_value. This logic is based on the response from the reconfiguration request. I am adding more comments as well here.
Thanks for this comment.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -22,27 +28,388 @@ using vmware::concord::client::statesnapshot::v1::StreamSnapshotResponse;
...
+      timeout = curr_timeout;
+    }
+  }
+
+  WriteConfig write_config{RequestConfig{false, 0, 1024 * 1024, timeout, "snapshotreq", "", false, true},
+                           bft::client::LinearizableQuorum{}};
+  StateSnapshotResponse snapshot_response;
+  grpc::Status return_status;
+  auto callback =
+      [&snapshot_response, &return_status, &write_config](concord::client::concordclient::SendResult&& send_result) {
```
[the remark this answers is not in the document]
arc-vmware: thanks, @robem, yes this is a valid concern. Since we are always sending a reconfiguration request we are not doing any client batching. So for our case client batching is always false. So we select the client and send the request then and there. But in the future reconfiguration requests might also be batched. So as you suggested, I will change the callback to async mode and receive the response in the callback, in the next check-in. Thanks for your meticulous review and comments.

arc-vmware:
@robem, thanks for reviewing this deeply, I request you to kindly continue your review, I have incorporated all your review comments (and expecting more). As our timezones are not matching, let's as usual use the git comments to convey ideas and chat to make better code. 🥇

arc-vmware:
@robem, This message I got twice, so I want to request you to please provide a way to create a source less cmake target. 
I tried the following 2 ways and got error, so any other idea will be very helpful:
Method (1) 
add_library(concordclient-event-api STATIC )
target_include_directories(concordclient-event-api PUBLIC include)
target_link_libraries(concordclient-event-api PRIVATE
  protobuf::libprotobuf # Needed for the timestamp in EventGroup
  util
)
Using the above I got the below error:
CMake Error at client/concordclient/CMakeLists.txt:12 (add_library):
No SOURCES given to target: concordclient-event-api

Method(2)
add_library(concordclient-event-api INTERFACE)
target_include_directories(concordclient-event-api PUBLIC include)
target_link_libraries(concordclient-event-api PRIVATE
  protobuf::libprotobuf # Needed for the timestamp in EventGroup
  util
)

Using the above I got the following error:
CMake Error at client/concordclient/CMakeLists.txt:13 (target_include_directories):
target_include_directories may only set INTERFACE properties on INTERFACE
targets
CMake Error at client/concordclient/CMakeLists.txt:14 (target_link_libraries):
INTERFACE library can only be used with the INTERFACE keyword of
target_link_libraries

This is in the CMakeLists.txt of concordclient. I will be grateful if you can kindly help me in creating a header-only target for concordclient-event-api. Thanks a lot in advance.

client/clientservice/src/state_snapshot_service.cpp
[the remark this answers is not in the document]
robem: That is an implementation detail and an assumption you are putting into this line of code. The code can block forever. The assumption is that if the underlying gRPC connection to the replica times out, you will push a timeout update/exception. However, what if the underlying code changes? What if the gRPC connection times out but we forget to populate that information? Or, ...
Bottom line, this code can block forever because there is no timeout specified. In addition, what if the caller decides to cancel the gRPC stream? This will never be checked unless we get an update/exception.

client/concordclient/include/client/concordclient/thread_safe_queue.hpp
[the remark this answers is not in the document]
robem: Every time you touch a source file, you should change the year in the copyright message if it has changed since the last update.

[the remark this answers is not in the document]
arc-vmware: thanks, @robem I am changing this.

client/clientservice/src/state_snapshot_service.cpp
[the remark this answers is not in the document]
arc-vmware: ok I will change this to tryPop(). Thanks @robem

client/concordclient/src/thread_safe_queue.cpp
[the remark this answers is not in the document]
arc-vmware: thanks, @robem, I am doing the following (http://mariobadr.com/creating-a-header-only-library-with-cmake.html)
find_package(Protobuf REQUIRED)
add_library(concordclient-event-api INTERFACE)
target_include_directories(concordclient-event-api INTERFACE include)
target_link_libraries(concordclient-event-api INTERFACE
  protobuf::libprotobuf # Needed for the timestamp in EventGroup
  util
)
---------------------------
The above is working so I am going with this.

teoparvanov (approved):
LGTM, thanks Amitava!

PR #2311, opened 2022-02-04
NOT AN ACTUAL PULL REQUEST

samcramer:
This pull request is intended to invoke Github CI in order to see if we're building with the proper number of CPUs.  It will not do an actual build or test, so impact on the CI pipeline should be minimal.


samcramer:
This was a "fake" pull request used to determine the number of CPUs we specify when building. I'm closing it so it doesn't clutter up the list of legitimate pull requests.

PR #2317, opened 2022-02-08
TRS: Allow last_known + 1 update requests

shgandhi:
This PR updates TRS s.t., last_known + 1 update requests are allowed, and are no longer considered as OUT_OF_RANGE requests. TRS waits in a while loop on the live update queue until the requested update is available to read.

### DIFF
--- a/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
+++ b/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
@@ -274,6 +274,31 @@ class ThinReplicaImpl {
       return subscribe_status;
     }
 
+    // If last_known + 1 update requested keep waiting until we have at least one live update
+    // Note that TRS considers last_known + 1 update request as valid, even if the requested
+    // update doesn't exist in storage yet.
+    if (request->has_events()) {
+      auto last_block_id = (config_->rostorage)->getLastBlockId();
+      if (request->events().block_id() == last_block_id + 1) {
+        while (not live_updates->waitUntilNonEmpty(kWaitForUpdateTimeout)) {
+          if (context->IsCancelled()) {
+            LOG_INFO(logger_, "StreamCancelled while waiting for the next live update.");
+            return grpc::Status::CANCELLED;
+          }
+        }
+      }
+    } else {
+      auto last_eg_id = kvb_filter->newestTagSpecificPublicEventGroupId();
+      if (request->event_groups().event_group_id() == last_eg_id + 1) {
+        while (not live_updates->waitForEventGroupUntilNonEmpty(kWaitForUpdateTimeout)) {
+          if (context->IsCancelled()) {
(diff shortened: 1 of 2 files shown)


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -274,6 +274,34 @@ class ThinReplicaImpl {
...
     }
 
+    // If last_known + 1 update requested keep waiting until we have at least one live update
+    // Note that TRS considers last_known + 1 update request as valid, even if the requested
+    // update doesn't exist in storage yet.
+    bool next_update_available = false;
+    while (not next_update_available) {
+      if (request->has_events()) {
+        // Determine latest block available
+        auto last_block_id = (config_->rostorage)->getLastBlockId();
```
robem, line 284: For every iteration of the while-loop, we are asking storage for the last block id. Don't we only care about the last block id once? If the last block id changes then we are good to go and so we wouldn't even need the live update queue.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -274,6 +274,34 @@ class ThinReplicaImpl {
       return subscribe_status;
     }
 
+    // If last_known + 1 update requested keep waiting until we have at least one live update
+    // Note that TRS considers last_known + 1 update request as valid, even if the requested
+    // update doesn't exist in storage yet.
+    bool next_update_available = false;
```
robem, line 280: Suggestion: No need for the extra variable. As soon as we get an update we could `break` from the loop.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -274,6 +274,34 @@ class ThinReplicaImpl {
...
+    while (not next_update_available) {
+      if (request->has_events()) {
+        // Determine latest block available
+        auto last_block_id = (config_->rostorage)->getLastBlockId();
+        if (request->events().block_id() == last_block_id + 1) {
+          next_update_available = live_updates->waitUntilNonEmpty(kWaitForUpdateTimeout);
+        } else {
+          break;
+        }
```
robem, line 289: Maybe everyone who doesn't subscribe into the future shouldn't enter the while-loop?

robem (approved):

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: I agree, updated.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: Updated.

PR #2318, opened 2022-02-08
Modified the logic of db-checkpoint window size check

nkumar04:
At present, we only check for db checkpoint create condition on the seqNum window size boundary. It may happen that, we complete the time interval required to create next db checkpoint much before reaching next window size boundary, This PR addresses this, now on every stable sequence number, we check if seqNum executed since last db checkpoint creation is greater or equal to the window size and time interval between two db checkpoints is expired.


itsmeakashgoyal (approved):
LGTM !

PR #2322, opened 2022-02-08
Make ValueWithTrids the default state snapshot value format

dartdart26:
Change the default state snapshot converter from the noop one (that
doesn't change the value) to one that parses a ValueWithTrids protobuf
type and extracts the value from it. Make it default for:
 * StateSnapshotReconfigurationHandler - when we execute ReadAsOf
 * computeAndPersistPublicStateHash() - when we compute the public state
   hash
 * ReplicaStateSnapshotServiceImpl - when we stream state

Make sure tests still use the noop converter, for simplicity.


arc-vmware (approved):
LGTM!!!
thanks @dartdart26

PR #2324, opened 2022-02-09
KvbAppFilter: Get TagTableValue from tag table

shgandhi:
A value in the tag table is now of the form -
<global_eg_id> + kTagTableSeparator + <external_tag_eg_id>.
Where external_tag_eg_id is the eg ID from participant's
perspective. The change is required to allow for a more
scalable read path solution for consumption of event groups.
This commit updates the KvbAppFilter to read the updated
value from the tag table.

### DIFF
--- a/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
+++ b/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
@@ -293,7 +293,7 @@ uint64_t KvbAppFilter::getValueFromLatestTable(const std::string &key) const {
   return concordUtils::fromBigEndianBuffer<uint64_t>(val->data.data());
 }
 
-uint64_t KvbAppFilter::getValueFromTagTable(const std::string &key) const {
+TagTableValue KvbAppFilter::getValueFromTagTable(const std::string &key) const {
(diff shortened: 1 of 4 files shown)


kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -308,11 +308,16 @@ uint64_t KvbAppFilter::getValueFromTagTable(const std::string &key) const {
     LOG_ERROR(logger_, msg.str());
     throw std::runtime_error(msg.str());
   }
-  auto result = concordUtils::fromBigEndianBuffer<uint64_t>(val->data.data());
+  std::string delim = kTagTableKeySeparator;
+  std::string result = val->data.data();
+  size_t delim_pos = result.find(delim);
```
robem, line 313: Effectively, we want to store <uint64>#<uint64>. If we use `toBigEndianStringBuffer` then we know the exact position of each component because this function will always return the same length string for a uint64. Meaning, instead of "finding" which involves more computation, we should be able to index directly into that string. Offset 0 is the global event group id, offset 8 is the separator, and offset 9 is the external event group id.

robem (changes requested):

PR #2326, opened 2022-02-09
Add last app-level txn time in StateSnapshotResponse

dartdart26:
Add the `StateSnapshotData.last_application_transaction_time` field
that contains the time of the last application-level transaction in the
snapshot in the form of seconds since epoch.

Additionally, since getting the time of the latest application-level
transaction is specific to the application, add a callback that knows
how to return it. The default callback returns the epoch (0).


arc-vmware (approved):
LGTM!!!

PR #2327, opened 2022-02-09
Skip unstable apollo test test_reconfiguration_with_ror

evdzhurov:
Failure is linked to previously found instability when testing with RO replicas.


teoparvanov (approved):
LGTM, thanks.

PR #2332, opened 2022-02-10
Refactoring BFT Client Code

agrkushvm:
Replacing the if/else block with ISecretsManagerImpl usage


glevkovich (approved):
LGTM!
Thanks @agrkushvm

PR #2333, opened 2022-02-11
Fix malformed JSON with nested objects produced in various places

samcramer:
BC-16805

Adds a `BuildJson` class which correctly composes JSON for nested objects.  The existing functions in `json-output.hpp` produce malformed JSON for nested object.

Testing:
- ran `echo "status get -replica-state | nc localhost 6888 | jq -S`
- ran `echo "status get state-transfer" | nc localhost  6888 | jq -S`
- compared the output of both commands to the output produced by the same commands run on master
- I needed to manually fix the output from master by removing the extraneous quotes around curly brackets, 
   since `jq` refuses to process invalid JSON
- compared the output produced by my fix with the output produced by master

I still need to figure out how to test `status get-replica-state` against a read-only replica; I had some trouble setting up a configuration which includes a read-only replica.  However the change in this case is trivial and I'm pretty confident I got it right.

Also, there should be unit tests for these cases.  I plan to include such tests in a future MR.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -36,10 +36,7 @@
 #include "client/reconfiguration/client_reconfiguration_engine.hpp"
 #include "client/reconfiguration/poll_based_state_client.hpp"
 
-#define STRPAIR(var) toPair(#var, var)
-
 using std::tie;
-using concordUtils::toPair;
 using std::chrono::steady_clock;
 using std::chrono::duration_cast;
 using std::chrono::milliseconds;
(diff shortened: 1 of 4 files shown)


glevkovich:
@samcramer , I'm not sure I am fully understand what hands-on test you've done.
`echo "status get -replica-state | nc localhost 6888 | jq -S`
How did you run this?
From my knowledge we can run similar command from concord-ctl.

util/include/json_output.hpp
```diff
@@ -22,6 +22,8 @@
 
 namespace concordUtils {
 
+// Note: these templated functions are deprecated.  Use the BuildJson class below for new code.
```
glevkovich, line 25: nit: if they deprecated why are you leaving them? Havn't you replaced them everywhere?

util/include/json_output.hpp
```diff
@@ -78,4 +80,56 @@ inline std::pair<std::string, std::string> toPair(const std::string &key, const
   return std::make_pair(key, value);
 }
 
+// Build a JSON object.  Supports nested objects.
+class BuildJson {
+ public:
+  inline BuildJson() {}
+
+  // Start a root JSON object.
+  inline void startJson() { str += "\n{\n"; }
```
glevkovich, line 89: c++ ISO standard:
`A function defined within a class definition is an inline function.`

I think that all the explicit inline keywords are unneeded in a c++17 modern compiler.
The compiler knows it is an inline function since you've declared and defined it in the class declaration (header).
same for all functions below.

util/include/json_output.hpp
```diff
@@ -78,4 +80,56 @@ inline std::pair<std::string, std::string> toPair(const std::string &key, const
   return std::make_pair(key, value);
 }
 
+// Build a JSON object.  Supports nested objects.
+class BuildJson {
+ public:
+  inline BuildJson() {}
```
glevkovich, line 86: nit:
`BuildJson() = default;

question:
why do you need to explciitly declare it, when you don't have any other costume ctor?
The compiler builds an implcit ctor anyway..

util/include/json_output.hpp
```diff
@@ -78,4 +80,56 @@ inline std::pair<std::string, std::string> toPair(const std::string &key, const
...
+
+  // Start a root JSON object.
+  inline void startJson() { str += "\n{\n"; }
+
+  // End the root JSON object.
+  inline void endJson() {
+    if (str.size() >= 2 && str[str.size() - 2] == ',') {
```
glevkovich, line 93: nit: is it possible to put all expressions in commas? I know that some people don't mind about it, but for some it really reduces readibility of the code.
example to the above:
`if ((str.size() >= 2) && (str[str.size() - 2] == ',')) {`

same comment for the rest of the code

util/include/json_output.hpp
```diff
@@ -78,4 +80,56 @@ inline std::pair<std::string, std::string> toPair(const std::string &key, const
...
+
+  // Add a nested JSON object.  The "json" parameter is expected to be a
+  // quoted JSON object in braces with no key prefix.
+  inline void addNestedJson(const std::string &key, const std::string &json) {
+    str += "\"" + key + "\" : ";
+    str += json + ",\n";
+  }
+
+  // Return the generated JSON.  Should be called after endJson().
+  inline std::string getJson() { return str; }
```
glevkovich, line 129: How about returning `const std::string&` ?
Here we have a copy, which is less efficient.

util/include/json_output.hpp
```diff
@@ -78,4 +80,56 @@ inline std::pair<std::string, std::string> toPair(const std::string &key, const
...
+
+ private:
```
glevkovich, line 131: nit: protected for future inheritance?

glevkovich (approved):
@samcramer LGTM, thanks for the hard work :)

Added some few comments/questions.

Liorgal28 (approved):
LGTM

toly-kournik:
@samcramer , FIY we are going to replace all JSON utils as well as nlohmann JSON library by boost.JSON in a near future.

samcramer:
> @samcramer , FIY we are going to replace all JSON utils as well as nlohmann JSON library by boost.JSON in a near future.

If so, is it worth merging this change?  I would like to, but if it is going to be pulled out soon maybe there's no point.

samcramer:
@glevkovich asked

> I'm not sure I am fully understand what hands-on test you've done.
> echo "status get -replica-state | nc localhost 6888 | jq -S
> How did you run this?
> From my knowledge we can run similar command from concord-ctl.

I logged into the docker container, installed `netcat` and ran the command show above, which was used in the ticket which reported this problem.  I will look into `concord-ctl`.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2333,56 +2329,57 @@ void BCStateTran::reportCollectingStatus(const uint64_t firstRequiredBlock,
...
+  bj.addKv("lastStored", psd_->getLastStoredCheckpoint());
+
+  bj.startNested("beingFetched");
+  bj.addKv("checkpointNum", fetched_cp.checkpointNum);
+  bj.addKv("lastBlock", fetched_cp.lastBlock);
+  bj.endNested();
+
+  bj.endNested();
+
+  bj.endJson();
```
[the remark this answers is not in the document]
samcramer: I think of the `startJson()`/`startNested()` and `endJson()`/`endNested()` as the "curly braces" for the JSON clauses.  If the developer forgets to add `endNested()` in the middle of the JSON construction, the nested JSON clause will (incorrectly) continue.

I don't think we can get rid of `endJson()` and do the equivalent in the destructor, since `getJson()` needs to be called before the `BuildJson` destructor is called.

util/include/json_output.hpp
samcramer: @glevkovich No, there are many other places where the templated functions are called.  I fixed the places where nested JSON objects were created, which were incorrect.  In the other places where they are called they produce correct JSON.

I agree that leaving the old ones in place is ugly. I considered replacing them all, but that turns out to be a pretty big job and I was concerned that doing that might introduce bugs.  I asked Gadi about the wisdom of changing everything at once, and he advised not doing it.  I filed a "technical debt" ticket to remove all of them "later".

util/include/json_output.hpp
samcramer: I've always seen ctors which do nothing declared this way.  That said, I did some googling and your position on this is well-supported.

util/include/json_output.hpp
samcramer: Makes sense.

util/include/json_output.hpp
samcramer: If we did so, wouldn't we run the risk of having a reference to a member of an object which has been destroyed if the `BuildJson` object went out of scope?

util/include/json_output.hpp
samcramer: OK.

PR #2334, opened 2022-02-11
TRS: Fix last_known+1 request allowed for event group transition

shgandhi:
If last_known+1 request is the first event group request after
legacy event requests, TRS must wait on the event group queue
and not the event queue. This commit fixes this issue.


robem (approved):
Thanks for the fix! 💪

PR #2337, opened 2022-02-11
Fix unstable skvbc_reply tc

situ-s:
Fixed the below issues:
1. The reply length was zero, causing fatal error when debugging was enabled.
2. The client was not accepting the reply as reply data was different from different replicas.

### DIFF
--- a/tests/apollo/test_skvbc_reply.py
+++ b/tests/apollo/test_skvbc_reply.py
@@ -12,7 +12,6 @@
 
 import sys
 import os.path
-import random
 import unittest
 import trio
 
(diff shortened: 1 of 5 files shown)


tests/apollo/test_skvbc_reply.py
```diff
@@ -138,7 +141,7 @@ async def test_conflict_detected_from_replicas(self, bft_network):
             writeset=[(key_prime, skvbc.random_value())],
             block_id=last_block_id - 1)
 
-        reply = await client.write_with_result(conflicting_write, result=bft_msgs.OperationResult.CONFLICT_DETECTED);
+        reply = await client.write_with_result(conflicting_write, pre_process=True, result=bft_msgs.OperationResult.CONFLICT_DETECTED);
         assert reply[1] == bft_msgs.OperationResult.CONFLICT_DETECTED, \
                         f"Expected Reply={bft_msgs.OperationResult.CONFLICT_DETECTED}; actual={reply[1]}"
```
yuliasherman, line 146: Remove ';'

yuliasherman (approved):
LGTM

PR #2343, opened 2022-02-14
Add support for block IDs in GetRecentSnapshotResponse

dartdart26:
Add support for both event group IDs and block IDs when returning the
point in the blockchain's history at which the state snapshot was taken.

For more details, see the description of the `taken_at` field in
StateSnapshotService.GetRecentSnapshotResponse.


arc-vmware (approved):
LGTM !!!
thanks @dartdart26

PR #2346, opened 2022-02-14
Fix event group transition

robem:
With the recent introduction of waiting for the X+1 event group, we also introduced a couple of bugs. Please see commit messages for details.


shgandhi (approved):
Thanks for finding and fixing these issues! LGTM!

PR #2354, opened 2022-02-16
EventGroups: Improve read-path computation

robem:
This PR exchanges the read-path algorithm. With the most recent tag table format change by @shgandhi we can now leverage the stored external event group ids to improve our search for an external-event-group-id-to-global-event-group-id mapping. Each commit describes the change and I suggest to go through them one-by-one.

In the previous approach, we read every single event group from the beginning because we needed to establish an external event group counter. We didn't store external event group ids in the database and therefore, we had to start at the beginning to properly determine the correct external event group id for a global event group id. This doesn't scale well :)
Now, external event group ids are stored with private event groups and hence we can do binary search over private event groups. If we don't find the correct external event group in the private event groups then we know that it has to be a public event group. In that case, we can compute the proper index without searching.

All other changes in the PR are refactoring based on top of this new implementation.

I will do performance tests with this change before anything gets merged.


kvbc/include/kvbc_app_filter/kvbc_app_filter.h
```diff
@@ -57,54 +57,15 @@ struct KvbFilteredEventGroupUpdate {
...
-  // be read from the tag table
-  uint64_t public_offset;
-  uint64_t private_offset;
-  // current tag-specific event_group_id
-  uint64_t curr_trid_event_group_id;
+struct FindGlobalEgIdResult {
+  // Global event group id matching the given external id
+  uint64_t global_id;
+  // Whether the event group is private to the client or public
+  bool is_private;
```
dartdart26, line 64: Nit: I think we might consider calling this `is_public`, instead. Reason is we use `private` in other places to mean no client can see an event or a state key-value. When an event or state key-value can only seen by a set of clients, we call it `restricted`. Please correct me if I miss something.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,236 +354,197 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
   return public_newest + private_newest;
 }
 
-std::optional<uint64_t> KvbAppFilter::getNextEventGroupId(std::shared_ptr<EventGroupClientState> &eg_state) {
-  LOG_DEBUG(logger_, "Public_offset: " << eg_state->public_offset << ", private_offset: " << eg_state->private_offset);
-  LOG_DEBUG(logger_, "Event group id list batch size: " << eg_state->event_group_id_batch.size());
+FindGlobalEgIdResult KvbAppFilter::findGlobalEventGroupId(uint64_t external_event_group_id) const {
+  auto external_oldest = oldestTagSpecificPublicEventGroupId();
+  auto external_newest = newestTagSpecificPublicEventGroupId();
+  ConcordAssertNE(external_oldest, 0);  // Everything pruned or was never created
```
dartdart26, line 360: Since `oldestTagSpecificPublicEventGroupId()` can return 0, maybe we should be explicit and put a precondition on `findGlobalEventGroupId()` so that it is clear when the use can call it.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -598,79 +598,96 @@ std::optional<uint64_t> KvbAppFilter::getNextEventGroupId(std::shared_ptr<EventG
...
+    // The lesser global event group id is the next update for the client
+    if (pvt_global_id < pub_global_id) {
+      global_eg_id = pvt_global_id;
+      is_previous_private = true;
+      ConcordAssertEQ(ext_eg_id + 1, pvt_external_id);
+      next_pvt_key = client_id_ + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(++next_pvt_eg_id);
+    } else {
+      global_eg_id = pub_global_id;
+      is_previous_private = false;
+      next_pub_key = kPublicEgId + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(++next_pub_eg_id);
```
shgandhi, line 529: If there is a scenario wherein `pvt_global_id` == `pub_global_id` == INT_MAX, i.e., none of the next updates exist yet in storage, we will still enter this else block and increment `next_pub_eg_id`? Can we skip an update here, or am I missing something?

shgandhi (commented):
Thanks for the change @robem! 🚀 
I just have a few comments.

kvbc/include/kvbc_app_filter/kvbc_app_filter.h
```diff
@@ -228,9 +181,8 @@ class KvbAppFilter {
 
   uint64_t newestTagSpecificPublicEventGroupId() const;
 
-  // Generate event group ids in batches from storage
-  // We do not want to process in memory, all the event group ids generated in a pruning window
-  std::optional<uint64_t> getNextEventGroupId(std::shared_ptr<EventGroupClientState> &eg_state);
+  // Given a tag-specific public (external) event group id, return the corresponding global event group id
```
dartdart26, line 184: Is `public` correct for this comment? Aren't we passing a tag-specific event group ID?

kvbc/test/kvbc_app_filter/kvbc_filter_test.cpp
```diff
@@ -986,6 +1032,172 @@ TEST(kvbc_filter_test, no_newest_public_event_group) {
...
+  ASSERT_EQ(kvb_filter.findGlobalEventGroupId(3).global_id, 4);
+
+  storage.fillWithEventGroupData(5, kPublicEgIdKey);
+  storage.fillWithEventGroupData(1, "A");
+  storage.fillWithEventGroupData(2, "B");
+
+  ASSERT_EQ(kvb_filter.findGlobalEventGroupId(9).global_id, 11);
+}
+
+TEST(kvbc_filter_test, find_external_eg_check_result) {
```
dartdart26, line 1141: Great tests, thank you!

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,236 +354,197 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
...
+  std::string prefix = client_id_ + kTagTableKeySeparator;
+
+  // Binary search in private event groups
+  uint64_t window_size;
+  auto window_start = private_start;
+  auto window_end = private_end;
+
+  // Cursors inside the search window; All point to the same entry
+  // client_id_ # current_pvt_eg_id => current_global_eg_id # current_ext_eg_id
+  auto [current_pvt_eg_id, current_global_eg_id, current_ext_eg_id] = std::make_tuple(0ull, 0ull, 0ull);
```
dartdart26, line 392: I was wondering why we need to create a tuple and then bind the variables to it. Can we just create 3 separate variables instead?

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,236 +354,197 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
...
+    uint64_t global_id;
+    std::tie(global_id, std::ignore) = getValueFromTagTable(
+        client_id_ + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(external_event_group_id));
+    return {global_id, true, external_event_group_id, public_end ? public_end - 1 : 0};
+  }
+
+  std::string prefix = client_id_ + kTagTableKeySeparator;
+
+  // Binary search in private event groups
+  uint64_t window_size;
```
dartdart26, line 386: Nit: Does it make sense to compute the `window_size` here and then update it at the end of the `while` loop below (instead of the beginning)? Doing it that way, we will ensure it is always initialized.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,236 +354,197 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
...
+  // At this point, we exhausted all private entries => it has to be a public event group
+  uint64_t global_eg_id;
 
-  ConcordAssertLE(eg_state->event_group_id_batch.size(), kBatchSize);
-  LOG_DEBUG(logger_, "Updated event group id list batch size: " << eg_state->event_group_id_batch.size());
-  return *eg_state->it++;
+  if (external_event_group_id < current_ext_eg_id) {
+    if (current_pvt_eg_id == private_start) {
+      std::tie(global_eg_id, std::ignore) = getValueFromTagTable(
+          kPublicEgId + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(external_event_group_id));
```
dartdart26, line 421: I think it would be better if we have a method/function that creates the tag such that we don't need to manually concatenate strings. Can be done in a separate commit or PR, just an idea.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,236 +354,197 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
...
+      return {global_eg_id, false, current_pvt_eg_id - 1, external_event_group_id};
+    }
+    auto num_pub_egs = current_ext_eg_id - current_pvt_eg_id;
+    auto pub_eg_id = num_pub_egs - (current_ext_eg_id - external_event_group_id - 1);
```
dartdart26, line 425: I am not sure I understand this expression. Could you please explain why we subtract `current_ext_eg_id` from `num_pub_egs`? I am sure it is correct, just couldn't parse it in my mind.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,236 +354,197 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
...
     throw std::runtime_error(msg.str());
   }
-  // update the offsets if we now have corresponding public/private event groups in storage
-  if (eg_data_state_->public_offset == 0) eg_data_state_->public_offset = public_start;
-  if (eg_data_state_->private_offset == 0) eg_data_state_->private_offset = private_start;
 
-  uint64_t event_group_id_end = private_end + public_end;
+  if (external_eg_id_start == 0) {
+    throw InvalidEventGroupId(external_eg_id_start);
+  }
```
dartdart26, line 461: We could move this as the first check in the method as it depends on user input and we don't need to do any work if we are throwing. Another approach is to assert on 0 input.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,236 +354,197 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
...
-    throw InvalidEventGroupId(event_group_id_start);
+  // The next public or private event group might not exist or got pruned
+  // In this case, set to max so that the comparion will be lost later
+  uint64_t pvt_external_id;
+  uint64_t pvt_global_id;
+  try {
+    std::tie(pvt_global_id, pvt_external_id) = getValueFromTagTable(next_pvt_key);
+  } catch (const std::exception &e) {
+    pvt_global_id = std::numeric_limits<uint64_t>::max();
   }
```
dartdart26, line 483: I would suggest changing the interface of `getValueFromTagTable()` such that we can distinguish between storage errors and lack of a key-value. That way we could maybe assert in places where it makes sense and not swallow  potentially critical errors. Maybe we can consider making `getValueFromTagTable()` return an optional.

On a side note, we can make `getValueFromTagTable()` assert if the value is not of type `ImmutableValue`. That will further simplify the interface and allow us to test less.

dartdart26: We could do above (if at all) in a separate PR/commit, definitely not critical now.

dartdart26 (approved):
Looks good, thanks @robem! Tests were extremely helpful!

I ran the code locally and debugged the tests, going line by line. Also, I made sure all returns in `findGlobalEventGroupId()` are covered. Code seems correct to me.

I have added some minor comments that you might choose to ignore or address later.

kvbc/include/kvbc_app_filter/kvbc_app_filter.h
robem: I follow your reasoning and it makes sense to me - will make the change. Thanks!

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: I will add a comment to the function. Thanks!

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: The idea is to only have one place in which we compute the `window_size` which makes reasoning a tiny bit easier IMHO. The variable is only used inside the while loop and the first statement sets the value. Let me know if it this doesn't make sense to you and we can change it.

robem: The idea is to only have one place in which we compute the `window_size` which makes reasoning a tiny bit easier IMHO. The variable is only used inside the while loop and the first statement sets the value. Let me know if it this doesn't make sense to you and we can change it.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: Good idea. Given that the concatenation is only used when we query the tag-table, we can overload `getValueFromTagTable` with `getValueFromTagTable(std::string tag, uint64_t event_group_id)`.
I will add this as a follow-up todo.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: I will add an assert to the beginning. Thanks!

robem:
@dartdart26 I very much appreciated your detailed review. I know this took quite some time. Thank you! 🚀

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: Thanks for this comment and I agree. Unfortunately, I cannot express "catch all but storage issues". I can list the current possible exceptions but as soon as this changes we will forget to update this try-catch. Asserting on the type makes sense to me - meaning, if the assert fires something is seriously wrong. I will add the assert.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -598,79 +598,96 @@ std::optional<uint64_t> KvbAppFilter::getNextEventGroupId(std::shared_ptr<EventG
...
+  auto [global_eg_id, is_previous_private, private_eg_id, public_eg_id] = findGlobalEventGroupId(external_eg_id_start);
+  uint64_t ext_eg_id = external_eg_id_start;
+
+  uint64_t next_pvt_eg_id = private_eg_id + 1;
+  auto next_pvt_key = client_id_ + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(next_pvt_eg_id);
+  uint64_t next_pub_eg_id = public_eg_id + 1;
+  auto next_pub_key = kPublicEgId + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(next_pub_eg_id);
+
+  // The next public or private event group might not exist or got pruned
+  // In this case, set to max so that the comparion will be lost later
```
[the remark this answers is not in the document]
robem: Thanks!

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: You're right, we can skip this whole if-else block and `break` early if both are MAX. Thank you!

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -355,35 +355,35 @@ uint64_t KvbAppFilter::newestTagSpecificPublicEventGroupId() const {
...
-    // We assume that we only get called if the event group exists
-    ConcordAssert(false);
-  } else if (not private_start) {
+  if (not private_start) {
     // requested external event group id == public event group id
     uint64_t global_id;
     std::tie(global_id, std::ignore) = getValueFromTagTable(
         kPublicEgId + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(external_event_group_id));
-    return global_id;
+    return {global_id, false, private_end ? private_end - 1 : 0, external_event_group_id};
```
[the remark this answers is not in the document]
robem: Good catch! I think this is wrong. The return value corresponds to the following comment in the header file:
```
// Either the matching private/public id or the newest up to this external id
// Note: The one that is not matching could be pruned (not queryable)
```
In this case, all private event groups are pruned. That means that the `private_end` event group was pruned and therefore must have occurred before. So, `private_end + 1` will be the next private event group that hasn't happened yet. Simply returning `private_end` will suffice. Thanks!

shgandhi (approved):

shgandhi:
Thanks for addressing the comments @robem. LGTM! 💯

robem:
@shgandhi I appreciate your review. Thanks 💪

robem:
Note, I just rebased and added a comment I forgot in the latest push.

PR #2355, opened 2022-02-17
VSS: Adding ledger time, todos and integration fix

arc-vmware:
This commit contains the following:
1) Ledger time is added.
2) Added fix for optional last received key in the request of StateSnapshotRequest as this usecase was failing.
3) Added a non blocking and non busy wait version of pop (called popTill) in BasicThreadSafeQueue for better performance.
4) Some configurations were hard coded and were not taken from the file and now made configurable. This was a TODO.

### DIFF
--- a/client/concordclient/include/client/concordclient/concord_client.hpp
+++ b/client/concordclient/include/client/concordclient/concord_client.hpp
@@ -108,6 +108,13 @@ struct SubscribeConfig {
   std::string pem_private_key;
 };
 
+struct StateSnapshotConfig {
+  // Number of threads available for state snapshot
+  uint32_t num_threads;
+  // Timeout till grpc connection with replicas will wait.
+  uint16_t timeout_in_sec;
+};
+
 struct ConcordClientConfig {
   // Description of the replica network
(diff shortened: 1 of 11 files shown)


client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -108,6 +108,13 @@ struct SubscribeConfig {
   std::string pem_private_key;
 };
 
+struct StateSnapshotConfig {
+  // Number of threads available for state snapshot
+  uint32_t num_threads;
+  // Timeout till grpc connection with replicas will wait.
+  uint16_t timeout_in_sec;
```
dartdart26, line 115: It might make sense to have default values defined here for the members. We can do it in a separate PR.

dartdart26 (approved):
Looks good, thanks!

client/concordclient/include/client/concordclient/concord_client.hpp
arc-vmware: thanks @dartdart26 , I will take this in a separate PR.

PR #2359, opened 2022-02-17
Add support for legacy events in StateSnapshotService

dartdart26:
Add support for legacy events in StateSnapshotService by introducing the
StateSnapshotData.blockchain_height field. The blockchain height can be
either:
 * EventGroupId for when event groups are enabled
 * BlockId for when legacy events are used

Introduce the ReplicaConfig.enableEventGroups option that specifies
whether event groups are enabled or not. If event groups are enabled,
return the newest public event group ID in the StateSnapshotResponse
message. If event groups are disabled, return the last block ID.

Change the Apollo test to accommodate the new fields.

### DIFF
--- a/client/clientservice/src/state_snapshot_service.cpp
+++ b/client/clientservice/src/state_snapshot_service.cpp
@@ -251,7 +251,14 @@ Status StateSnapshotServiceImpl::GetRecentSnapshot(ServerContext* context,
 
   if (return_status.ok() && snapshot_response.data.has_value()) {
     response->set_snapshot_id(snapshot_response.data->snapshot_id);
-    response->set_event_group_id(snapshot_response.data->event_group_id);
+    switch (snapshot_response.data->blockchain_height_type) {
+      case concord::messages::BlockchainHeightType::EventGroupId:
+        response->set_event_group_id(snapshot_response.data->blockchain_height);
+        break;
+      case concord::messages::BlockchainHeightType::BlockId:
+        response->set_block_id(snapshot_response.data->blockchain_height);
+        break;
+    }
     response->set_key_value_count_estimate(snapshot_response.data->key_value_count_estimate);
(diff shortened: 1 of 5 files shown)


client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -251,7 +251,14 @@ Status StateSnapshotServiceImpl::GetRecentSnapshot(ServerContext* context,
...
   if (return_status.ok() && snapshot_response.data.has_value()) {
     response->set_snapshot_id(snapshot_response.data->snapshot_id);
-    response->set_event_group_id(snapshot_response.data->event_group_id);
+    switch (snapshot_response.data->blockchain_height_type) {
+      case concord::messages::BlockchainHeightType::EventGroupId:
+        response->set_event_group_id(snapshot_response.data->blockchain_height);
+        break;
+      case concord::messages::BlockchainHeightType::BlockId:
+        response->set_block_id(snapshot_response.data->blockchain_height);
+        break;
```
arc-vmware, line 260: nit: we can add default: with a LOG_FATAL() and ConcordAssert(false). This will help us identify the issue if we add any new height type. But lets take that in a later PR.

arc-vmware (approved):
Thanks @dartdart26 ,  changes looks good.

client/clientservice/src/state_snapshot_service.cpp
dartdart26: I rely on the compiler not allowing us to compile without handling all enum variants. That is why I prefer to omit `default`:

```
/concord-bft/client/clientservice/src/state_snapshot_service.cpp:254:13: error: enumeration value 'BlockId' not handled in switch [-Werror,-Wswitch]
    switch (snapshot_response.data->blockchain_height_type) {
```

PR #2360, opened 2022-02-17
Implement S3 multiput by uploading objects in parallel.

toly-kournik:
Tuning retry logic.
Apply SHA256 on key, for uniform length (not to exceed 1024 length).
Add toHexString() to EVPHash.
Revive 3 flaky ro replica tests
Few S3 client logging and cosmetic changes.


yontyon (approved):
LGTM
Thanks

kvbc/test/CMakeLists.txt
```diff
@@ -153,18 +153,6 @@ target_link_libraries(kv_blockchain_db_editor_test PUBLIC
 )
 endif(BUILD_ROCKSDB_STORAGE)
 
-if(USE_S3_OBJECT_STORE)
```
WildFireFlum, line 156: Why was this test removed?

storage/include/storage/test/s3_test_common.h
```diff
@@ -127,7 +127,7 @@ class S3Test : public ::testing::Test {
   struct shared_state {
     concord::storage::s3::StoreConfig config;
     fs::path test_dir;
-    std::unique_ptr<concord::storage::s3::Client> client;
+    std::shared_ptr<concord::storage::s3::Client> client;
```
WildFireFlum, line 130: Wasn't the binary which used this test removed from CMakeLists?
Assuming that this is still being used, why the switch to a shared_ptr?

tests/apollo/test_skvbc_ro_replica.py
```diff
@@ -152,7 +152,6 @@ async def test_ro_replica_start_simultaneously (self, bft_network, tracker):
                                 log.log_message(message_type="Replica" + str(ro_replica_id) + " : lastExecutedSeqNum:" + str(lastExecutedSeqNum))
                                 nursery.cancel_scope.cancel()
 
-    @unittest.skip("Unstable test - BC-17828")
```
WildFireFlum, line 155: Didn't you say that you were using retries for these tests?
I can't find it in the code.

util/include/evp_hash.hpp
```diff
@@ -97,6 +98,12 @@ class EVPHash {
     updating_ = false;
     return digest;
   }
+  static std::string toHexString(const Digest& digest) {
```
WildFireFlum, line 101: This seems like something that can be used in many places throughout the code.
Consider making it a template so that every collection of bytes (vector for instance) can be converted to a hex string as well.

WildFireFlum (commented):
Got a few short questions

kvbc/test/CMakeLists.txt
toly-kournik: This test claimed to test a storage factory, but essentially it didn't

storage/include/storage/test/s3_test_common.h
toly-kournik: shared_ptr to s3::Client is held by s3::Client::Trnsaction objects

tests/apollo/test_skvbc_ro_replica.py
toly-kournik: No, I just uncommented the skip pragmas. Now the tests pass

util/include/evp_hash.hpp
toly-kournik: You are right, we have it in a Sliver, Digest and  STDigest classes as well

storage/include/storage/test/s3_test_common.h
WildFireFlum: So this previously  being a unique ptr was a bug?

PR #2361, opened 2022-02-17
Add support for legacy events in StateSnapshotService

dartdart26:
Add support for legacy events in StateSnapshotService by introducing the
StateSnapshotData.blockchain_height field. The blockchain height can be
either:
 * EventGroupId for when event groups are enabled
 * BlockId for when legacy events are used

Introduce the ReplicaConfig.enableEventGroups option that specifies
whether event groups are enabled or not. If event groups are enabled,
return the newest public event group ID in the StateSnapshotResponse
message. If event groups are disabled, return the last block ID.

Change the Apollo test to accommodate the new fields.


arc-vmware (approved):
Thanks @dartdart26 , LGTM!!!

PR #2363, opened 2022-02-18
Adding low disk threshold factor as a config parameter

nkumar04:
Rocksdb removes old sst files and merge them to new one during compaction, if these removed files are refered by rocksdb checkpoints then they contributes toward increase in the disk storage requirement.To address this we have a monitor threads keeps checking on the available disk space. If it is below threshold then we remove oldest checkpoints to avoid any failure in compaction process. This PR, adds a configuration parameter that is used as theshold factor in reference to current rocksdb size.
if( free_disk_space < (threshold_factor * size_of_rocksdb)) then remove_oldest_dbcheckpoint.

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -233,6 +233,10 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
                std::chrono::seconds,
                std::chrono::seconds{60},
                "Time interval in seconds to monitor disk usage by db checkpoints");
+  CONFIG_PARAM(dbCheckpointDiskSpaceThreshold,
+               std::string,
+               "0.5",
+               "Free disk space threshold for db checkpoint cleanup");
   CONFIG_PARAM(enablePostExecutionSeparation, bool, true, "Post-execution thread separation feature flag");
   CONFIG_PARAM(postExecutionQueuesSize, uint16_t, 50, "Post-execution deferred message queues size");
 
@@ -373,6 +377,7 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
     serialize(outStream, dbCheckpointDirPath);
     serialize(outStream, dbSnapshotIntervalSeconds);
     serialize(outStream, dbCheckpointMonitorIntervalSeconds);
+    serialize(outStream, dbCheckpointDiskSpaceThreshold);
     serialize(outStream, enablePostExecutionSeparation);
     serialize(outStream, postExecutionQueuesSize);
     serialize(outStream, stateIterationMultiGetBatchSize);
@@ -467,6 +472,7 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
     deserialize(inStream, dbCheckpointDirPath);
     deserialize(inStream, dbSnapshotIntervalSeconds);
     deserialize(inStream, dbCheckpointMonitorIntervalSeconds);
+    deserialize(inStream, dbCheckpointDiskSpaceThreshold);
     deserialize(inStream, enablePostExecutionSeparation);
     deserialize(inStream, postExecutionQueuesSize);
     deserialize(inStream, stateIterationMultiGetBatchSize);
@@ -561,6 +567,7 @@ inline std::ostream& operator<<(std::ostream& os, const ReplicaConfig& rc) {
   os << ",";
   os << KVLOG(rc.dbSnapshotIntervalSeconds.count(),
               rc.dbCheckpointMonitorIntervalSeconds.count(),
+              rc.dbCheckpointDiskSpaceThreshold,
               rc.enableMultiplexChannel,
               rc.enableEventGroups);
   for (auto& [param, value] : rc.config_params_) os << param << ": " << value << "\n";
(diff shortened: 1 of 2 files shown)


bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -233,6 +233,10 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
                std::chrono::seconds,
                std::chrono::seconds{60},
                "Time interval in seconds to monitor disk usage by db checkpoints");
+  CONFIG_PARAM(dbCheckpointDiskSpaceThreshold,
+               std::string,
+               "0.5",
```
arc-vmware, line 238: This threshold must be less than 1. It should be good if this is kept below 0.8  (which will be a stricter upper bound on the threshold).
Any value greater than this upper bound will be clipped. This can be done one time instead of doing it every time we use this variable. So I prefer that we should take it as a double or float and then add a data validator in the MR. Taking it as a string and converting it to a double everytime can be avoided.

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -233,6 +233,10 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
                std::chrono::seconds,
                std::chrono::seconds{60},
                "Time interval in seconds to monitor disk usage by db checkpoints");
+  CONFIG_PARAM(dbCheckpointDiskSpaceThreshold,
+               std::string,
```
dartdart26, line 237: Can we use a floating point type here?

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -261,9 +261,17 @@ void DbCheckpointManager::checkAndRemove() {
   const auto& dbCurrentSize = directorySize(dbPath, false, false);
   try {
     const _fs::space_info diskSpace = _fs::space(dbPath);
+    // Threshold for available disk space is kept at (0.5 * curr_sizeof(rocksDb)). Assumption
+    // is that this is enough to run compaction in rocksdb
+    // Before starting a compaction, RocksDB will check if there is enough room to create the output SST files.
+    // This is done by calling SstFileManager::EnoughRoomForCompaction(). This function conservatively estimates
+    // the output size as the sum of the sizes of all the input SST files to the compaction.
+    // Assumption:  SstFileManager::EnoughRoomForCompaction() < (0.5 * curr_sizeof(rocksDb)) all the time
```
dartdart26, line 269: Why do we assume 0.5, in particular? Just asking out of curiosity as I don't know the details of `EnoughRoomForCompaction()`.

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -261,9 +261,17 @@ void DbCheckpointManager::checkAndRemove() {
...
+    const double lowDiskThresholdFactor =
+        stod(bftEngine::ReplicaConfig::instance().getdbCheckpointDiskSpaceThreshold());
     // make sure that we have at least equal amount of free storage available
-    // all the time to avoid any failure dure to low disk space
-    if (diskSpace.available < dbCurrentSize) {
+    // all the time to avoid any failure dur to low disk space
+    if (diskSpace.available < floor(dbCurrentSize * lowDiskThresholdFactor)) {
```
dartdart26, line 274: Why do we need to use `floor()` here?

bftengine/src/bftengine/DbCheckpointManager.cpp
nkumar04: 0.5 is just an approximation. We use level style compaction, where compaction process takes a subset of sst files at level N and merge them to files at level (N+1). This is per Column family.
As per the documents, the ratio of size_at(Level N) : size_at(Level N+1) so 1:10 (approximately). 
So, even when we take a subset of files at one level before the last level in a CF, it should be much smaller than the half of the total Database size on the disk.

bftengine/include/bftengine/ReplicaConfig.hpp
nkumar04: these config parameters need to be serializable and we do not have floating point number serialization support in current utility. So keeping it this way for now.

bftengine/src/bftengine/DbCheckpointManager.cpp
nkumar04: fixed

bftengine/include/bftengine/ReplicaConfig.hpp
nkumar04: these config parameters need to be serializable and we do not have floating point number serialization support in current utility. So keeping it this way for now.

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -261,9 +261,17 @@ void DbCheckpointManager::checkAndRemove() {
...
+    // Before starting a compaction, RocksDB will check if there is enough room to create the output SST files.
+    // This is done by calling SstFileManager::EnoughRoomForCompaction(). This function conservatively estimates
+    // the output size as the sum of the sizes of all the input SST files to the compaction.
+    // Assumption:  SstFileManager::EnoughRoomForCompaction() < (0.5 * curr_sizeof(rocksDb)) all the time
+    const double lowDiskThresholdFactor =
+        stod(bftEngine::ReplicaConfig::instance().getdbCheckpointDiskSpaceThreshold());
     // make sure that we have at least equal amount of free storage available
-    // all the time to avoid any failure dure to low disk space
-    if (diskSpace.available < dbCurrentSize) {
+    // all the time to avoid any failure dur to low disk space
```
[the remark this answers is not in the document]
nkumar04: fixed

arc-vmware (approved):
LGTM!!!

PR #2364, opened 2022-02-18
Clientpool should use the reply message size with fix for callback

arc-vmware:
Client service uses client pool to send the BFT request using BFT client. But after receiving the request the client pool ignores the reply size and calls the bft client with the default reply size (which is 64KB). This may not be correct for all clients of the client pool and as the max_reply_buffe_size is available in the request config, the client pool is expected to honor the same. Any API like ReadAsOf, which might return a bigger size response will get a timeout operation result. So in this PR, we are making sure that the client pool uses the max reply message buffer from the request config.

This is however not taken care of in case of batch requests. The batch request will ignore the request config's the max reply buffer size. It will use the same message size for each message in a batch as it's happening now.

Along with this, the requests that are kept in the external queue are not keeping the callback function. In this PR the callback is added if available.

### DIFF
--- a/client/client_pool/src/concord_client_pool.cpp
+++ b/client/client_pool/src/concord_client_pool.cpp
@@ -49,7 +49,6 @@ SubmitResult ConcordClientPool::SendRequest(std::vector<uint8_t> &&request,
     callback(bftEngine::SendResult{static_cast<uint32_t>(OperationResult::INVALID_REQUEST)});
     return SubmitResult::Overloaded;
   }
-  externalRequest external_request;
   std::unique_lock<std::mutex> lock(clients_queue_lock_);
   metricsComponent_.UpdateAggregator();
   auto serving_candidates = clients_.size();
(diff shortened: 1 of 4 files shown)


client/client_pool/src/concord_client_pool.cpp
```diff
@@ -146,8 +145,10 @@ SubmitResult ConcordClientPool::SendRequest(std::vector<uint8_t> &&request,
     }
   }
 
-  // Request hasn't been processed yet
-  if (external_requests_queue_.size() < jobs_queue_max_size_) {
+  // Request hasn't been processed yet and this is not a reconfiguration request then only allow the request to wait in
+  // the external request queue. Reconfigution requests are unbatched and in general require immediate response. So
+  // waiting is not acceptable.
+  if ((external_requests_queue_.size() < jobs_queue_max_size_) && (!(flags & ClientMsgFlag::RECONFIG_FLAG_REQ))) {
```
yuliasherman, line 151: Please talk to Lior regarding the external requests queue. AFAIR, it has to be removed.

arc-vmware: Thanks, @yuliasherman , yes i am discussing with @Liorgal28.

Liorgal28 (approved):
Thanks for fixing my comments.
LGTM

PR #2375, opened 2022-02-22
Plain TCP compilation fix

situ-s:
Compilation was failing when CONCORD_BFT_CMAKE_TRANSPORT is set to TCP. This PR handles that.


yuliasherman (approved):
LGTM

PR #2376, opened 2022-02-22
Fixing issues during TRC integration test

arc-vmware:
The reconnection usecase of GRPC Connection was not thread safe. When we
raised the number of parallel connections, and stopped the replicas,
reconnection happened and it was stopping the client service abruptly.
So adding the changes to create a thread safe and performant
GrpcConnection abstraction.


client/concordclient/src/concord_client.cpp
```diff
@@ -143,13 +143,20 @@ void ConcordClient::createGrpcConnections() {
 }
 
 void ConcordClient::checkAndReConnectGrpcConnections() {
+  bool needReconnection = false;
```
dartdart26, line 146: Nit: Code style - `need_reconnection`.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -209,6 +212,7 @@ class GrpcConnection {
 
   // gRPC connection
   std::shared_ptr<grpc::Channel> channel_;
+  std::shared_mutex channel_protector_;
```
dartdart26, line 215: Nit: It might be more explicit if we name it `channel_mutex`.

dartdart26:
Would it make sense to describe how a `GrpcConnection` can be used by multiple threads? For example, which methods can be called concurrently (if it can be done), any special considerations, etc. Also, I think we should change this comment:

```
// GrpcConnection provides NO thread safety guarantees in the event more than one
// call to a GrpcConnection function executes concurrently on the same
// GrpcConnection object.
```

client/thin-replica-client/src/grpc_connection.cpp
```diff
@@ -107,16 +112,27 @@ void GrpcConnection::disconnect() {
...
+void GrpcConnection::checkAndReConnect(std::unique_ptr<GrpcConnectionConfig>& config) {
+  WriteLock write_lock(channel_protector_);
+  if (!isConnectedNoLock()) {
+    disconnect();
+    connect(config);
+  }
+}
+
 GrpcConnection::Result GrpcConnection::openDataStream(const SubscriptionRequest& request) {
+  ReadLock read_lock(channel_protector_);
```
dartdart26, line 124: Could you please explain the rationale for using a `ReadLock` here?

client/concordclient/src/concord_client.cpp
arc-vmware: Thanks, @dartdart26, I will change this.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
arc-vmware: Thanks, @dartdart26, I will update the name of this variable.

client/thin-replica-client/src/grpc_connection.cpp
arc-vmware: Thanks, @dartdart26 for the comments. In GrpcConnection the following 2 kinds of members are related and are protected by the channel_mutex_:
1) Channel -- This member is used by connect, reconnect, disconnect and isconnected family of functions.
2) trs_stub and rss_tub --> These stubs are created on top of the channel. They share the channel, so if we want to change the channel then we should make sure that stubs are not accessing the channel. The GRPC is implemented in such a manner that stubs can share a single channel.

Now all the functions which modify the channel directly will have to take a write lock. But when the stubs are used, we should make sure that no one modifies the channel. So we take read lock whenever we use the stub and write lock whenever we update channel. 

We should also note that we need to take write lock when we create the stub.

In all the open family of functions, we use the stub to call the API. So we take a read lock and we assume that channel_ is not nullptr.

arc-vmware:
thanks, @dartdart26, I changed the comment, after these changes GRPC connection should be thread-safe.

client/thin-replica-client/src/grpc_connection.cpp
```diff
@@ -133,13 +149,12 @@ GrpcConnection::Result GrpcConnection::openDataStream(const SubscriptionRequest&
 
   ConcordAssert(status == future_status::ready);
   data_stream_ = stream.get();
-  if (data_stream_) return Result::kSuccess;
+  if (!data_stream_) {
+    data_context_.reset();
+    return Result::kFailure;
+  }
```
shgandhi, line 155: If `data_stream_` exists, I think we should call `Finish()` on the stream to get the rpc status. If the status is ok, we return `Result::kSuccess`, otherwise, we must check for `OUT_OF_RANGE` and `NOT_FOUND` errors, and return the corresponding result, else return `Result::kFailure`. This is because we must catch and relay `OUT_OF_RANGE` and `NOT_FOUND` errors from the TRS to the concord client.
Same comment for `openHashStream` method.

client/concordclient/src/concord_client.cpp
```diff
@@ -143,13 +143,20 @@ void ConcordClient::createGrpcConnections() {
...
+    }
+  }
+  if (need_reconnection) {
+    for (size_t con_offset = 0; con_offset < config_.topology.replicas.size(); ++con_offset) {
       auto trsc_config = std::make_unique<GrpcConnectionConfig>(config_.subscribe_config.use_tls,
                                                                 config_.subscribe_config.pem_private_key,
                                                                 config_.subscribe_config.pem_cert_chain,
                                                                 config_.transport.event_pem_certs);
-      (grpc_connections_[con_offset])->reConnect(trsc_config);
+      (grpc_connections_[con_offset])->checkAndReConnect(trsc_config);
```
shgandhi, line 159: If connection at a `con_offset` is not connected, we reconnect to all the replicas? Why don't we reconnect to the replica which was disconnected?

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -94,7 +94,7 @@ class GrpcConnection {
   // Connect & disconnect from the TRS
   virtual void connect(std::unique_ptr<GrpcConnectionConfig>& config);
   virtual bool isConnected();
-  virtual void reConnect(std::unique_ptr<GrpcConnectionConfig>& config);
+  virtual void checkAndReConnect(std::unique_ptr<GrpcConnectionConfig>& config);
```
shgandhi, line 97: Nit: In my opinion, `reconnect` is a single word, we can do without the camel-case here.

shgandhi (changes requested):
Thanks for the change @arc-vmware. I have added a few comments.

client/thin-replica-client/src/grpc_connection.cpp
arc-vmware: thanks, @shgandhi. will do this change.

client/concordclient/src/concord_client.cpp
arc-vmware: Thanks, @shgandhi, we don't reconnect all replicas, we only reconnect that replica that is disconnected. That's why we check again within the write lock before disconnecting. If it's not disconnected, then inside checkAndReConnect we don't reconnect.

Another question will be why do we have 2 loops?
Ans: The first loop is for isConnected() function which takes a read lock and quickly determines whether a reconnection is needed or not. This happens 99% of the time. But sometimes we actually need to reconnect. But when we reconnect we should check and then only reconnect as it might be possible that some other thread got the write lock and reconnect. That's why we have 2 loops and we check always before reconnection, so we essentially never reconnect any already connected replica.

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
arc-vmware: Thanks, @shgandhi , let me change this in next commit.

arc-vmware: I just realized that after we have the notion of checkAndReconnect, the reconnect is not used and its a private function. So I am removing this function. Thanks for the comment.

client/thin-replica-client/src/grpc_connection.cpp
shgandhi: It looks like you reverted your change. 🤔  
Maybe we can fix this separately.

shgandhi:
Thanks for addressing the comments @arc-vmware. LGTM!

shgandhi (approved):

PR #2378, opened 2022-02-22
Batch size 0 as stop pruning fix

titusen:
Stop command has batch size set to 0 which earlier was disallowed. COndition on throw runtime exception is removed.

### DIFF
--- a/kvbc/include/resources-manager/AdaptivePruningManager.hpp
+++ b/kvbc/include/resources-manager/AdaptivePruningManager.hpp
@@ -80,7 +80,7 @@ class AdaptivePruningManager {
   uint32_t current_pruning_pace_;
   uint64_t current_batch_size_;
   concordMetrics::Component metricComponent;
-  concordMetrics::Component::Handle<concordMetrics::AtomicGauge> ticksPerSecondMetric, batchSizeMetric,
+  concordMetrics::Component::Handle<concordMetrics::AtomicGauge> intervalBetweenTicksSecondsMetric, batchSizeMetric,
       transactionsPerSecondMetric, postExecUtilizationMetric, pruningAvgTimeMicroMetric, pruningUtilizationMetric;
 };
 }  // namespace concord::performance
--- a/kvbc/include/resources-manager/AdaptivePruningManager.hpp
+++ b/kvbc/include/resources-manager/AdaptivePruningManager.hpp
@@ -80,7 +80,7 @@ class AdaptivePruningManager {
   uint32_t current_pruning_pace_;
   uint64_t current_batch_size_;
   concordMetrics::Component metricComponent;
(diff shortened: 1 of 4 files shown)


kvbc/src/pruning_reserved_pages_client.cpp
```diff
@@ -23,7 +23,7 @@ static char* data(std::vector<std::uint8_t>& v) { return reinterpret_cast<char*>
 static const char* cdata(const std::vector<std::uint8_t>& v) { return reinterpret_cast<const char*>(v.data()); }
 
 static std::uint32_t toSecondsCount(const std::chrono::seconds& s) {
-  if (s.count() <= 0 || s.count() > std::numeric_limits<std::uint32_t>::max()) {
+  if (s.count() > std::numeric_limits<std::uint32_t>::max()) {
```
dartdart26, line 26: Does it make sense to document that `batch_blocks_num = 0` and `s.count() <= 0` means pruning stop? Also, do we need to handle `s.count() == 0` somewhere else in a special way?

I think, in the future, it might make more sense to persist a different data structure (maybe an `oneof` or have an explicit flag that says pruning is stopped) instead of relying on these special values.

kvbc/include/resources-manager/AdaptivePruningManager.hpp
```diff
@@ -80,7 +80,7 @@ class AdaptivePruningManager {
   uint32_t current_pruning_pace_;
   uint64_t current_batch_size_;
   concordMetrics::Component metricComponent;
-  concordMetrics::Component::Handle<concordMetrics::AtomicGauge> ticksPerSecondMetric, batchSizeMetric,
+  concordMetrics::Component::Handle<concordMetrics::AtomicGauge> blocksPerSecondMetric, batchSizeMetric,
```
dartdart26, line 83: Does it make sense to leave `ticksPerSecondMetric` and add the `blocksPerSecondMetric` one? I guess it would depend if we need `ticksPerSecondMetric`.

kvbc/src/pruning_reserved_pages_client.cpp
titusen: @yontyon, what do u think? Is this better to have it as 0 or another flag?

kvbc/include/resources-manager/AdaptivePruningManager.hpp
titusen: IMO ticks per second is misleading, because it's blocks. Besides the metric "intervalBetweenTicks" already exists.

teoparvanov (approved):
Hi @titusen , let's merge this, and address the additional comments in a separate PR if considered critical.

Thank you,
Teo

PR #2379, opened 2022-02-22
create an order on the clientIds

yontyon:
In this PR we create an order on the client's ids in concordbft.
The order is: [replicas, read-only replica, proxies clients, external clients, clientservice, operator, internal clients]

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -147,6 +147,7 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
 
   // Reconfiguration credentials
   CONFIG_PARAM(pathToOperatorPublicKey_, std::string, "", "Path to the operator public key pem file");
+  CONFIG_PARAM(operatorEnabled_, bool, true, "true if operator is enabled");
   // Pruning parameters
   CONFIG_PARAM(pruningEnabled_, bool, false, "Enable pruning");
   CONFIG_PARAM(numBlocksToKeep_, uint64_t, 0, "how much blocks to keep while pruning");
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicasInfo.cpp
```diff
@@ -63,6 +63,11 @@ ReplicasInfo::ReplicasInfo(const ReplicaConfig& config,
       _dynamicCollectorForPartialProofs{dynamicCollectorForPartialProofs},
       _dynamicCollectorForExecutionProofs{dynamicCollectorForExecutionProofs},
 
+      /*
+       * Ids order is: [replicas, ro-replicas, clientProxies, client-service external clients, client-service ID,
+       * operator, internal-clients]. Notice, that in case we have an operator, it is included in numOfExternalClients,
```
yuliasherman, line 68: -> client-service clients (aka external clients)
Please mention also cre client. 
BTW, there is a configuration parameter for cre presence, should we take it into account or remove it if not relevant anymore?

yontyon: the cre configuration parameter is only for the client not for replica (because the client needs to know not to use the last client)

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -438,6 +440,7 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
     deserialize(inStream, thresholdVerificationKeys_);
 
     deserialize(inStream, pathToOperatorPublicKey_);
+    deserialize(inStream, operatorEnabled_);
```
yuliasherman, line 443: Don't we like to show the value in the log, like for other parameters?

yuliasherman (approved):
LGTM

PR #2381, opened 2022-02-23
Clientpool should use the reply message size with fix for callback

arc-vmware:
Client service uses client pool to send the bft request using bft
client. But after receiving the request the client pool ignores the
reply size and calls the bft client with default reply size (which is
64KB). This may not be correct for all clients of client pool and as the
max_reply_buffe_size is available in the request config, client pool is
expected to honor the same. Any API like ReadAsOf, which might return a
bigger size response will get a timeout operation result. So in this PR
we are making sure that client pool uses the max reply message buffer
from the request config.

This is however not taken care in case of batch request. The batch
request will ignore the request config's max reply buffer size. It will
use same message size for each message in a batch as its happening now.

Alongwith this, the requests sent by client service are not keeping the
callback for longer time untill its called. So mechanism for keeping the
callback and cleanup later is added.


teoparvanov (approved):
LGTM!

PR #2383, opened 2022-02-23
Removing the external_queue of requests from client pool

arc-vmware:
Client Pool (CP) batches the requests till batch size or batch timeout in a pending request queue inside a client in the pool. Either of these conditions (timer off or size overflow) leads to creation of a batch job which sends the requests and then fills the client with some requests which were in the external request queue. This client will then wait till it fills up or timeout and then the process repeats.

During the SendRequest if there are no serving clients or available clients then the request is dumped into the external request queue (ERQ).

As we can see this is a good mechanism if batching is not available. But if batching is available, we should deprecate the ERQ and this PR is doing the same.


yuliasherman (approved):
LGTM! Thanks for simplifying the code.

Liorgal28 (changes requested):
Don't forget that this change should also involve changes in concord side (removing the configuration param in configuration manager).

Liorgal28 (approved):
Amazing cleanup 👏

PR #2384, opened 2022-02-23
Initial implementation for byzantine_behavior (faulty) replica.

HristoStaykov:
In this implementation we instantiate the faulty replicas at the
creation of our distributed system. Even though the distributed system
is aware who are the faulty nodes, the replicas themselves don't have
access to this knowledge.

### DIFF
--- a/docs/sbft-formal-model/proof.dfy
+++ b/docs/sbft-formal-model/proof.dfy
@@ -365,7 +365,7 @@ module Proof {
 
   lemma CountHonestReplicas(c: Constants)
     requires c.WF()
-    ensures |HonestReplicas(c)| == c.clusterConfig.AgreementQuorum()
+    ensures |HonestReplicas(c)| >= c.clusterConfig.AgreementQuorum()
   {
     var count := 0;
     var f := c.clusterConfig.F();
--- a/docs/sbft-formal-model/proof.dfy
+++ b/docs/sbft-formal-model/proof.dfy
@@ -32,7 +32,7 @@ module Proof {
   predicate IsHonestReplica(c:Constants, hostId:HostId) 
   {
     && c.WF()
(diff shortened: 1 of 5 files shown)


docs/sbft-formal-model/proof.dfy
```diff
@@ -347,14 +346,44 @@ module Proof {
...
 
+  function HonestReplicas(c: Constants) : (honest:set<HostId>) 
+    requires c.WF()
+  {
+    set sender | 0 <= sender < c.clusterConfig.N() && IsHonestReplica(c, sender)
+  }
+
+  lemma CountHonestReplicas(c: Constants)
+    requires c.WF()
+    ensures |HonestReplicas(c)| == c.clusterConfig.AgreementQuorum()
```
teoparvanov, line 368: shouldn't this rather be `|HonestReplicas(c)| >= c.clusterConfig.AgreementQuorum()` ?

teoparvanov (approved):
Brilliant work, thanks @HristoStaykov !

docs/sbft-formal-model/proof.dfy
HristoStaykov: Currently we set the faulty replicas to be the maximum safety preserving value and the honest are exactly = 2F+1, but it is more accurate to make this >=, because we might want to run the verifier with less than F faulty also. I will change this accordingly.

teoparvanov: Thanks, sounds good! I hope it doesn't break the existing proofs...

HristoStaykov: I checked locally, the proof passes correctly. This reminds me that we need to add the check for the proof (and syntactic check for all .dfy files) in the CI soon :)

PR #2385, opened 2022-02-23
WIP Added initial version for executable reference model.

HristoStaykov:
The idea behind the executable reference model is for us to have
an experiment environment where we can pass sequence of messages,
mutate the state of the replica and examine the consequences.
This way we could in the future implement tests with sequence of
messages for our executable reference model and then the same
set of messages for the real implementation replica to verify
the same behavior is observed.
The reference executable model is kept as close as possible to
the Dafny model code.


shgandhi:
Thanks for adding this @HristoStaykov! LGTM!

shgandhi (approved):

teoparvanov (changes requested):

PR #2386, opened 2022-02-23
Don't create replies for internal clients

yontyon:
(1) add clientservice to tests (to make sure we don't have bugs)
(2) don't create replies for internal clients


yuliasherman (approved):
LGTM

PR #2388, opened 2022-02-24
Fix an error with clientservice ids

yontyon:
Client service ids need to consider the operator's existence on their "start" position and not on their "end" position


yuliasherman (approved):
LGTM

PR #2390, opened 2022-02-24
Add debug logs to RawMemoryPool and correct numOfAvailableChunks_ counter

yuliasherman:
1. Added DEBUG logs printing memory chunk addresses to allow debugging of RawMemoryPool issues.
2. Increase and decrease numOfAvailableChunks_ only in case the chunk gets allocated/returned from/to the pool.

### DIFF
--- a/util/src/RawMemoryPool.cpp
+++ b/util/src/RawMemoryPool.cpp
@@ -41,7 +41,9 @@ void RawMemoryPool::allocatePool(int32_t initialChunksNum, int32_t maxChunksNum)
   maxChunksNum_ = maxChunksNum;
   pool_ = make_shared<boost::lockfree::queue<char*, boost::lockfree::fixed_sized<true>>>(maxChunksNum_);
   for (int32_t i = 0; i < initialChunksNum_; ++i) {
-    pool_->push(new char[chunkSize_]);
+    char* chunk = new char[chunkSize_];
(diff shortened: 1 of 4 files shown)


util/src/RawMemoryPool.cpp
```diff
@@ -41,7 +41,9 @@ void RawMemoryPool::allocatePool(int32_t initialChunksNum, int32_t maxChunksNum)
   maxChunksNum_ = maxChunksNum;
   pool_ = make_shared<boost::lockfree::queue<char*, boost::lockfree::fixed_sized<true>>>(maxChunksNum_);
   for (int32_t i = 0; i < initialChunksNum_; ++i) {
-    pool_->push(new char[chunkSize_]);
+    char* chunk = new char[chunkSize_];
```
ukg19, line 44: should we check if new returned Null here (in case)? We would be pushing empty memory in pool in that case.

ukg19 (changes requested):

util/src/RawMemoryPool.cpp
yuliasherman: Operator new throws std::bad_alloc exception in case it fails allocating memory

ukg19 (approved):
LGTM !!
Thanks @yuliasherman for adding more logging information in memory pool.

PR #2393, opened 2022-02-26
Remove view change desc in resetMetadata

nkumar04:
On resetMetadata - we do not remove these two items: descriptorOfLastExitFromView, descriptorOfLastNewView
in case of cross-node restore, these descriptors are not relevant and can be removed. After restore, replica can sync its view from other replica via ReplicaStatus message


toly-kournik (approved):
LGTM!
Thanks

HristoStaykov (approved):
LGTM, thanks @nkumar04 !

PR #2396, opened 2022-02-27
Fixing the signature verification for Reconfiguration requests of Snapshot Request family

arc-vmware:
The signature verification is not needed for StateSnapshotRequest, SignedPublicStateHashRequest and StateSnapshotReadAsOfRequest if transaction signing is not enabled.

Adding the fix for this issue.
This is a temporary fix and I will add the permanent fix after discussion with @yontyon in details.


teoparvanov (approved):
LGTM!

PR #2410, opened 2022-03-02
Deleting all dbcheckpoints when dbcheckpoint feature is disable

itsmeakashgoyal:
This PR addresses the issue when customer disable dbcheckpoint feature later in future, than all the old dbcheckpoints will be deleted from disk storage as well as persistence.

### DIFF
--- a/bftengine/src/bftengine/DbCheckpointManager.cpp
+++ b/bftengine/src/bftengine/DbCheckpointManager.cpp
@@ -133,7 +133,7 @@ void DbCheckpointManager::cleanUp() {
   // this gets called when db checkpoint is disabled
   // check if there is chkpt data in persistence
   loadCheckpointDataFromPersistence();
-  if (!maxNumOfCheckpoints_) {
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -133,7 +133,7 @@ void DbCheckpointManager::cleanUp() {
   // this gets called when db checkpoint is disabled
   // check if there is chkpt data in persistence
   loadCheckpointDataFromPersistence();
-  if (!maxNumOfCheckpoints_) {
+  if (!maxNumOfCheckpoints_ || !ReplicaConfig::instance().dbCheckpointFeatureEnabled) {
```
nkumar04, line 136: nit: 
its better to keep the order like this:
if ( !ReplicaConfig::instance().dbCheckpointFeatureEnabled || !maxNumOfCheckpoints_ )

nkumar04 (approved):
LGTM!
Thanks!

PR #2411, opened 2022-03-02
Fix for data race in ClientsManager.

HristoStaykov:
executeRequests method is now being called from threads
different than the message processing thread. This causes
multi threaded read/write access to the requestsInfo data
structures in ClientsManager, which eventually leads to
data corruption.
The proposed fix posts an internal message to the message
processing thread to clean up the given client request
thus maintaining the read/write access to 1 thread.


nkgeorgiev (approved):

teoparvanov (approved):
LGTM, thanks @HristoStaykov !

yuliasherman (approved):
LGTM

PR #2413, opened 2022-03-03
Downgrade hash stream closed message to WARN from ERROR

shgandhi:
When TRC forces resubscription on one of the f+1 open streams
to not miss an update while receiving updates, it closes all
hash streams, this doesn't warrant an ERROR message on the TRS,
hence downgrading to WARN.

### DIFF
--- a/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
+++ b/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
@@ -364,6 +364,10 @@ class ThinReplicaImpl {
         LOG_WARN(logger_, "Requested update pruned in syncAndSend: " << error.what());
         CLEANUP_SUBSCRIPTION();
         return grpc::Status(grpc::StatusCode::NOT_FOUND, error.what());
+      } catch (StreamClosed& error) {
+        LOG_INFO(logger_, "StreamClosed in syncAndSend: " << error.what());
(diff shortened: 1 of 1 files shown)


robem:
We only emit an ERROR message when those functions are called as part of `syncAndSend` and `syncAndSendEventGroups`. Otherwise, the exceptions are being logged as INFO messages. Pretty much all try-catch blocks are in the outer-most layer (the RPC handle function) and I suggest we keep it that way because it makes reading/reasoning about the code easier if everything is in one place. However, we can use try-catch if we need special handling in lower layers but for this PR we just log a message an rethrow.

shgandhi:
@robem I have updated the commit s.t., we catch StreamClosed error from `syncAndSend` and `syncAndSendEventGroups` methods when stream closed exception is thrown, and print a warning.
gRPC status CANCELLED is returned because I think it is the only status code that closely corresponds to the given scenario.

robem (approved):
Lgtm. I wouldn't be opposed to even downgrade it to INFO given that this is part of the normal workflow of the TRC and the TRS has actually no information why the stream was closed.

shgandhi:
> Lgtm. I wouldn't be opposed to even downgrade it to INFO given that this is part of the normal workflow of the TRC and the TRS has actually no information why the stream was closed.

I agree, downgraded to INFO.

PR #2414, opened 2022-03-03
Json whitespace

samcramer:
This fixes the problem you were seeing with extra white-space in the generated JSON.

### DIFF
--- a/util/include/json_output.hpp
+++ b/util/include/json_output.hpp
@@ -86,42 +86,46 @@ class BuildJson {
   inline BuildJson() {}
 
   // Start a root JSON object.
-  inline void startJson() { str += "\n{\n"; }
+  inline void startJson() { str += "{\n"; }
 
   // End the root JSON object.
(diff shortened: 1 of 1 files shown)


util/include/json_output.hpp
```diff
@@ -86,42 +86,46 @@ class BuildJson {
   inline BuildJson() {}
 
   // Start a root JSON object.
-  inline void startJson() { str += "\n{\n"; }
+  inline void startJson() { str += "{\n"; }
```
glevkovich, line 89: nit: I think I commented before, why do you need the inline keyword
Iso C++:
`A function defined within a class definition is an inline function.`

glevkovich (approved):
@samcramer , Thank you, small comment.

util/include/json_output.hpp
samcramer: Yes, you're right.  I will fix that and other review comments in an upcoming PR.

PR #2419, opened 2022-03-03
Fix error propagation issue between external_client and client_pool

yuliasherman:
Due to two different approaches existing in bft client error reporting, external_client missed one of them.

### DIFF
--- a/client/client_pool/src/external_client.cpp
+++ b/client/client_pool/src/external_client.cpp
@@ -64,6 +64,9 @@ ConcordClient::ConcordClient(int client_id)
 ConcordClient::~ConcordClient() noexcept = default;
 
 bft::client::Reply ConcordClient::SendRequest(const bft::client::WriteConfig& config, bft::client::Msg&& request) {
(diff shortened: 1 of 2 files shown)


client/client_pool/src/external_client.cpp
```diff
@@ -99,9 +109,13 @@ bft::client::Reply ConcordClient::SendRequest(const bft::client::ReadConfig& con
                            << " has failed to invoke, timeout=" << config.request.timeout.count()
                            << " ms has been reached");
   }
-  LOG_DEBUG(logger_,
-            "Request has completed processing"
-                << KVLOG(client_id_, config.request.sequence_number, config.request.correlation_id));
+  if (res.result != static_cast<uint32_t>(OperationResult::SUCCESS))
```
ukg19, line 112: same as above.

ukg19 (changes requested):

ukg19 (approved):
LGTM !!
Thanks @yuliasherman

PR #2426, opened 2022-03-04
Clean up suggested in review feedback

samcramer:
Remove unnecessary inline keywords, remove constructor.


glevkovich (approved):
LGTM, Thanks!

PR #2427, opened 2022-03-04
Adding new google.protobuf.Any type to concord client

cjn:
This commit consists of introducing google.protobuf.Any request/response for application request/response.
The current bytes request/response will co-exist for the time being to support both types.

This is with reference to [PR](https://github.com/vmware/concord-bft/pull/1961)

### DIFF
--- a/client/clientservice/src/request_service.cpp
+++ b/client/clientservice/src/request_service.cpp
@@ -65,8 +65,13 @@ void RequestServiceCallData::populateResult(grpc::Status status) {
 }
 
 void RequestServiceCallData::sendToConcordClient() {
-  bft::client::Msg msg(request_.request().begin(), request_.request().end());
-
+  bft::client::Msg msg;
+  if (request_.has_typed_request()) {
+    std::string str = request_.typed_request().SerializeAsString();
+    msg = bft::client::Msg(str.begin(), str.end());
+  } else {
+    msg = bft::client::Msg(request_.raw_request().begin(), request_.raw_request().end());
+  }
   auto seconds = std::chrono::seconds{request_.timeout().seconds()};
   auto nanos = std::chrono::nanoseconds{request_.timeout().nanos()};
   auto timeout = std::chrono::duration_cast<std::chrono::milliseconds>(seconds + nanos);
@@ -120,7 +125,13 @@ void RequestServiceCallData::sendToConcordClient() {
     auto reply = std::get<bft::client::Reply>(send_result);
     // We need to copy because there is no implicit conversion between vector<uint8> and std::string
     std::string data(reply.matched_data.begin(), reply.matched_data.end());
-    this->response_.set_response(std::move(data));
+
+    // Check if the application response is of Any Type then set it to Any response.
+    google::protobuf::Any* anyResponse = this->response_.mutable_typed_response();
+    if (anyResponse->ParseFromString(data) == false) {
+      this->response_.set_raw_response(std::move(data));
+      this->response_.release_typed_response();
+    }
     this->populateResult(grpc::Status::OK);
   };
 
--- a/client/proto/request/v1/request.proto
+++ b/client/proto/request/v1/request.proto
@@ -6,6 +6,7 @@ syntax = "proto3";
 package vmware.concord.client.request.v1;
 
 import "google/protobuf/duration.proto";
+import "google/protobuf/any.proto";
 
 // Specifies Java package name, using the standard prefix "com."
 option java_package = "com.vmware.concord.client.request.v1";
@@ -34,7 +35,10 @@ service RequestService {
 
 message Request {
   // Required application request which gets evaluated by the execution engine.
-  bytes request = 1;
+  oneof application_request {
+    bytes raw_request = 1;
+    google.protobuf.Any typed_request = 6;
+  }
 
   // Required timeout which defines the maximum amount of time the caller is
   // willing to wait for the request to be processed by a quorum of replicas.
@@ -65,5 +69,8 @@ message Request {
 
 message Response {
   // Application data returned by the execution engine.
-  bytes response = 1;
+  oneof application_response {
+    bytes raw_response = 1;
+    google.protobuf.Any typed_response = 6;
+  } 
 }


client/proto/request/v2/request.proto
```diff
@@ -0,0 +1,76 @@
...
+  // INVALID_ARGUMENT if `read_only` and `pre_execute` are set.
+  optional bool pre_execute = 4;
+
+  // Optional request identifier. At runtime, Concord Client maps the
+  // `correlation_id` to a sequence number which can be used to track the request
+  // in log messages in the blockchain network.
+  optional string correlation_id = 5;
+  
+  //This will be the new application request which will co-exist with request during transition
+  optional google.protobuf.Any new_request = 6;
```
miklos-da, line 67: I'd strongly suggest not using the word 'new' for fields:
```suggestion
  optional google.protobuf.Any application_request = 6;
```

client/proto/request/v2/request.proto
```diff
@@ -0,0 +1,76 @@
...
+  // RESOURCE_EXHAUSTED: if Concord Client is overloaded. The caller should retry with a backoff.
+  // UNAVAILABLE: if Concord Client is currently unable to process any requests. The caller should retry with a backoff.
+  // ABORTED: if Concord has a contention between concurrently running requests. The caller should retry.
+  // INTERNAL: if Concord Client cannot progress independent of the request.
+  rpc Send(Request) returns (Response);
+}
+
+message Request {
+  // Required application request which gets evaluated by the execution engine.
+  bytes request = 1;
```
miklos-da, line 38: We can still [keep wire compatibility](https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility_issues) by introducing a `oneof` for the typed vs. non-typed request:
```suggestion
  oneof application_request {
    bytes raw_request = 1 [deprecated = true];
    google.protobuf.Any typed_request = 6;
  }
```
The same may be applied to the `Response` message. I.e., this way we make it explicit that only one of these fields will be ever set.
I have annotated the old field as deprecated above as well (in a subsequent service version it may be removed).

client/proto/request/v2/request.proto
```diff
@@ -0,0 +1,76 @@
...
+  // in log messages in the blockchain network.
+  optional string correlation_id = 5;
+  
+  //This will be the new application request which will co-exist with request during transition
+  optional google.protobuf.Any new_request = 6;
+}
+
+message Response {
+  // Application data returned by the execution engine.
+  bytes response = 1;
```
miklos-da, line 72: Following the above suggestion for `Request`, here we may introduce a `oneof` for the two response types.

client/proto/request/v2/request.proto
cjn: Agree. Updated changes

client/proto/request/v2/request.proto
cjn: This could be tricky which is the reason why OneOf was not introduced. The document does mention about backward compatibility but that would be when we are moving existing fields to a new OneOf. But here we are creating a new Oneof and also introducing a new field typed_request  into the OneOf which  is not the same case.
cc: @robem

miklos-da: @cjn Introducing a new `oneof` that encapsulates an existing and a new field has the same wire-compatibility guarantees as moving more than one existing fields into a `oneof`. I.e., old clients that do not know about the new field just won't be able to interpret requests that have the `oneof` field populated that they are not aware of.
@robem I suggested using `oneof` + `[deprecated = true]` exactly for this reason, i.e., this way we don't have to introduce a new service version as the suggested changes are wire-compatible.

cjn: Thanks @miklos-da  and @robem. I will take these inputs into consideration for existing service.

teoparvanov (approved):
LGTM, thanks Celine! Let's move forward with this.

client/proto/request/v2/request.proto
cjn: Updated changes to PR

client/proto/request/v1/request.proto
```diff
@@ -34,7 +35,10 @@ service RequestService {
 
 message Request {
   // Required application request which gets evaluated by the execution engine.
-  bytes request = 1;
+  oneof application_request {
+    bytes raw_request = 1 [deprecated = false];
```
miklos-da, line 39: If you don't want to deprecate the previous field yet, then please remove the flag as it's `false` by default. See the effects of this flag [here](https://developers.google.com/protocol-buffers/docs/proto3#options), i.e., it just adds an annotation to generated classes.
```suggestion
    bytes raw_request = 1;
```

client/proto/request/v1/request.proto
```diff
@@ -65,5 +69,8 @@ message Request {
 
 message Response {
   // Application data returned by the execution engine.
-  bytes response = 1;
+  oneof application_response {
+    bytes raw_response = 1 [deprecated = false];
```
miklos-da, line 73: Same as for `raw_request` above:
```suggestion
    bytes raw_response = 1;
```

client/proto/request/v2/request.proto
cjn: Updated changes

client/proto/request/v1/request.proto
cjn: Updated changes

client/clientservice/src/request_service.cpp
```diff
@@ -65,8 +65,13 @@ void RequestServiceCallData::populateResult(grpc::Status status) {
 }
 
 void RequestServiceCallData::sendToConcordClient() {
-  bft::client::Msg msg(request_.request().begin(), request_.request().end());
-
+  bft::client::Msg msg;
+  if (request_.has_typed_request()) {
+    std::string str = request_.typed_request().SerializeAsString();
```
miklos-da, line 70: As we don't have the replica side implemented here, I'd rather suggest forwarding the embedded message. I.e., we can assume for now that the embedded type is always going to be a `ConcordRequest` or `DamlRequest`.
```suggestion
    std::string str = request_.typed_request().value();
```

client/clientservice/src/request_service.cpp
```diff
@@ -120,7 +125,13 @@ void RequestServiceCallData::sendToConcordClient() {
     auto reply = std::get<bft::client::Reply>(send_result);
     // We need to copy because there is no implicit conversion between vector<uint8> and std::string
     std::string data(reply.matched_data.begin(), reply.matched_data.end());
-    this->response_.set_response(std::move(data));
+
+    // Check if the application response is of Any Type then set it to Any response.
+    google::protobuf::Any* anyResponse = this->response_.mutable_typed_response();
+    if (anyResponse->ParseFromString(data) == false) {
```
miklos-da, line 131: This is not a reliable way to determine a type of a message. We either need to switch to serializing a container message (such as `google.protobuf.Any`) or assume for now that the response is always going to be of a certain type (such as `ConcordResponse` or `DamlResponse`).
I'd strongly suggest doing the latter first, and then changing the protocol between the BFT client and the replicas to contain the actual type of the messages traveling.

client/clientservice/src/request_service.cpp
```diff
@@ -65,8 +65,13 @@ void RequestServiceCallData::populateResult(grpc::Status status) {
 }
 
 void RequestServiceCallData::sendToConcordClient() {
-  bft::client::Msg msg(request_.request().begin(), request_.request().end());
-
+  bft::client::Msg msg;
+  if (request_.has_typed_request()) {
```
miklos-da, line 69: Can we add tests for this, please?

miklos-da (commented):
The proto changes look good, thank you! Added a few comments on the C++ code.

PR #2429, opened 2022-03-05
Fix additional error propagation issue

yuliasherman:
Client request service expects to receive either result or reply as a request execution result. In case it has both (in case of failure, bft_client has the reply filled with error info), it ignores the error and returns a successful result to the ledger_api.

### DIFF
--- a/client/client_pool/src/external_client.cpp
+++ b/client/client_pool/src/external_client.cpp
@@ -66,7 +66,7 @@ ConcordClient::~ConcordClient() noexcept = default;
 bft::client::Reply ConcordClient::SendRequest(const bft::client::WriteConfig& config, bft::client::Msg&& request) {
   LOG_INFO(
       logger_,
-      "Start request processing" << KVLOG(client_id_, config.request.sequence_number, config.request.correlation_id));
+      "Request processing started" << KVLOG(client_id_, config.request.sequence_number, config.request.correlation_id));
   bft::client::Reply res;
   clientRequestExecutionResult_ = OperationResult::SUCCESS;
   try {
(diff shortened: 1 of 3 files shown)


client/client_pool/src/external_client.cpp
```diff
@@ -177,25 +177,33 @@ std::pair<int32_t, ConcordClient::PendingReplies> ConcordClient::SendPendingRequ
...
-      auto cid = seq_num_to_cid.find(rep.first)->second;
-      auto data_size = rep.second.matched_data.size();
-      for (auto& reply : pending_replies_) {
-        if (reply.cid != cid) continue;
-        if (reply.cb) {
-          auto result = bftEngine::SendResult{rep.second};
-          reply.cb(std::move(result));
-          LOG_DEBUG(logger_, "Request processing via callback completed " << KVLOG(client_id_, batch_cid, reply.cid));
+    LOG_INFO(logger_, "Batch processing started" << KVLOG(client_id_, batch_cid));
+    auto receivedRepliesMap = new_client_->sendBatch(request_queue, batch_cid);
```
ukg19, line 181: Nit: Variables naming. received_replies_map and received_reply_entry

client/client_pool/src/external_client.cpp
```diff
@@ -177,25 +177,33 @@ std::pair<int32_t, ConcordClient::PendingReplies> ConcordClient::SendPendingRequ
...
+    for (const auto& receivedReplyEntry : receivedRepliesMap) {
+      auto cid = seq_num_to_cid.find(receivedReplyEntry.first)->second;
```
ukg19, line 183: May result in a crash if seq_num to cid entry is not found. 
Should we first check if entry is found and then access the value (to be on safe side)?

client/client_pool/src/external_client.cpp
yuliasherman: fixed

client/client_pool/src/external_client.cpp
yuliasherman: OK

client/client_pool/src/external_client.cpp
```diff
@@ -177,25 +177,41 @@ std::pair<int32_t, ConcordClient::PendingReplies> ConcordClient::SendPendingRequ
...
-        if (reply.cb) {
-          auto result = bftEngine::SendResult{rep.second};
-          reply.cb(std::move(result));
-          LOG_DEBUG(logger_, "Request processing via callback completed " << KVLOG(client_id_, batch_cid, reply.cid));
+    LOG_INFO(logger_, "Batch processing started" << KVLOG(client_id_, batch_cid));
+    auto received_replies_map = new_client_->sendBatch(request_queue, batch_cid);
+    for (const auto& received_reply_entry : received_replies_map) {
+      const auto received_reply_seq_num = received_reply_entry.first;
+      const auto& pending_seq_num_to_cid_entry = seq_num_to_cid.find(received_reply_seq_num);
+      if (pending_seq_num_to_cid_entry == seq_num_to_cid.end()) {
```
ukg19, line 185: Thanks for the check @yuliasherman 
Let's not call find two times on map. 
auto cid = pending_seq_num_to_cid_entry->second

client/client_pool/src/external_client.cpp
```diff
@@ -177,25 +177,41 @@ std::pair<int32_t, ConcordClient::PendingReplies> ConcordClient::SendPendingRequ
...
+        LOG_ERROR(logger_,
+                  "Received reply for a non pending for reply request"
+                      << KVLOG(client_id_, batch_cid, received_reply_seq_num));
+        continue;
+      }
+      auto cid = seq_num_to_cid.find(received_reply_entry.first)->second;
```
ukg19, line 191: Thanks for the check @yuliasherman
Let's not call find two times on map.
auto cid = pending_seq_num_to_cid_entry->second

yuliasherman: Ooops. Thanks for paying attention.

yontyon (approved):
LGTM

ukg19 (approved):
LGTM !!
Thanks @yuliasherman

PR #2432, opened 2022-03-06
Pass batch cid to the functions where it is missing

yuliasherman:
Correct batch cid is a condition for the batch object to be released. In case it is empty, the batch does not get released and the client gets stuck with it forever.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.hpp
+++ b/bftengine/src/preprocessor/PreProcessor.hpp
@@ -179,7 +179,7 @@ class PreProcessor {
                                  bool isReadOnly,
                                  uint16_t clientId,
                                  NodeIdType senderId,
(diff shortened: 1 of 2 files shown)


bftengine/src/preprocessor/PreProcessor.hpp
```diff
@@ -274,12 +274,12 @@ class PreProcessor {
                                         bool arrivedInBatch,
                                         uint16_t msgOffsetInBatch,
                                         PreProcessRequestMsgSharedPtr &preProcessRequestMsg,
-                                        const std::string &batchCid = "",
-                                        uint32_t batchSize = 1);
+                                        const std::string &batchCif,
```
ukg19, line 277: small typo, batchCid

ukg19 (approved):

PR #2434, opened 2022-03-07
Fixes for data races reported from Apollo TSAN run.

HristoStaykov:
Because of concurrent access to the counters we need
to make them atomic.

### DIFF
--- a/bftengine/src/bftengine/DebugStatistics.hpp
+++ b/bftengine/src/bftengine/DebugStatistics.hpp
@@ -14,6 +14,7 @@
 #include <cstddef>
 #include <stdint.h>
 #include "TimeUtils.hpp"
+#include <atomic>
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/DebugStatistics.hpp
```diff
@@ -44,7 +45,7 @@ class DebugStatistics {
     Time lastCycleTime;
 
     size_t receivedMessages;
-    size_t sendMessages;
+    std::atomic<size_t> sendMessages;
```
nkgeorgiev, line 48: Why only this variable is made atomic?
What about the other?

glevkovich (approved):
LGTM!

nkgeorgiev (approved):

PR #2435, opened 2022-03-07
Pass batch cid to the functions where it is missing

yuliasherman:
Cherry-pick for 8f3d53a11647ce5a036df0aafd2b9e0d618ce5bb
Correct batch cid is a condition for the batch object to be released. In case it is empty, the batch does not get released and the client gets stuck with it forever.


ukg19 (approved):
LGTM !!

PR #2436, opened 2022-03-08
TS: Fix an issue with time check

ukg19:
Removes the check  seqNumInfo.isTimeCorrect() during arrival of StartSlowCommitMsg. Because of it the replica will not be able to quickly recover from a temporary time difference, and it may lead to unnecessary view changes.


arc-vmware (approved):
LGTM!!!

PR #2438, opened 2022-03-08
Fix RSI indexing v0.14

yontyon:
RSI manager assumes order on the external client ids. Now that we have clientservice ids added, we need to adapt this index
In this PR we simply add the clientservice ids into account


arc-vmware (approved):
Thanks @yontyon , LGTM!!!

PR #2439, opened 2022-03-08
Add client nodes ids to the RSI manager

yontyon:
See https://github.com/vmware/concord-bft/pull/2438


arc-vmware (approved):
Thanks @yontyon , LGTM!!!

PR #2441, opened 2022-03-08
Fix apollo dbsnapshot test

Efrat1:
Fixed test_restore_from_snapshot_of_other

### DIFF
--- a/tests/apollo/test_skvbc_dbsnapshot.py
+++ b/tests/apollo/test_skvbc_dbsnapshot.py
@@ -24,7 +24,7 @@ from util import bft
 from util import skvbc as kvbc
 from util.skvbc import SimpleKVBCProtocol
 from util.skvbc_history_tracker import verify_linearizability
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc_dbsnapshot.py
```diff
@@ -965,6 +960,19 @@ async def wait_for_created_snapshots_metric(self, bft_network, replicas, expecte
                 else:
                     break
 
+    async def wait_for_fast_paths_metric_to_update(self, bft_network, replica_id, old_num_of_fast_paths):
```
toly-kournik, line 963: Don't we have an existing method to check that replica has caught up?

Efrat1: I used wait_for_consensus_path with path_type OPTIMISTIC_FAST.
Let me know if you thought about something else.

toly-kournik: I don't know why whis method was defined in test_skvbc_dbsnapshot.py, when it's a bft network functionality.
There's wait_for_fast_path_to_be_prevalent() defined in bft

toly-kournik (approved):
LGTM! thanks Efrat

PR #2443, opened 2022-03-08
Perform scale cmd restart after db-checkpoint is created

nkumar04:
Create db checkpoint is an async operation. We need to wait for it to complete before restarting replica if scale operation was done with restart option true. This PR addresses this issue


yontyon (approved):
LGTM!
Thanks

PR #2447, opened 2022-03-09
Replace boost::filysytem by std::filesystem

toly-kournik:
Since std::filesystem became a standard, we want to replace all existing boost::filesystem usages by std::filesystem.
This will also allow not to build boost::filesystem library

### DIFF
--- a/bftclient/test/CMakeLists.txt
+++ b/bftclient/test/CMakeLists.txt
@@ -9,6 +9,7 @@ target_link_libraries(bft_client_test PRIVATE
   bftclient_new
   threshsign
   secretsmanager
+  stdc++fs
 )
 
 add_executable(bft_client_api_tests bft_client_api_tests.cpp)
--- a/util/include/util/filesystem.hpp
+++ b/util/include/util/filesystem.hpp
@@ -1,0 +1,22 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the sub-component's license, as noted in the LICENSE
+// file.
+
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace fs = std::filesystem;
+#elif __has_include(<experimental/filesystem>)
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#else
+#error "Missing filesystem support"
+#endif
--- a/bftengine/include/bftengine/DbCheckpointManager.hpp
+++ b/bftengine/include/bftengine/DbCheckpointManager.hpp
@@ -32,8 +32,8 @@
 #include "Metrics.hpp"
 #include "InternalBFTClient.hpp"
 #include "storage/db_interface.h"
-#include <boost/filesystem.hpp>
-namespace _fs = boost::filesystem;
+#include "util/filesystem.hpp"
+namespace _fs = fs;
 namespace bftEngine::impl {
 using std::chrono::duration_cast;
 using Status = concordUtils::Status;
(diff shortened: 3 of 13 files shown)


yuliasherman:
Where is the description?

toly-kournik:
The description is in the title

yuliasherman:
Not good enough. What is the reason for such a replacement?

util/include/util/filesystem.hpp
```diff
@@ -0,0 +1,22 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
```
WildFireFlum, line 3: 2022?

bftclient/test/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ target_link_libraries(bft_client_test PRIVATE
   bftclient_new
   threshsign
   secretsmanager
+  stdc++fs
```
WildFireFlum, line 12: https://releases.llvm.org/9.0.0/projects/libcxx/docs/UsingLibcxx.html#using-filesystem

"Starting with LLVM 9.0, support for <filesystem> is provided in the main library and nothing special is required to use <filesystem>."
Why is this needed?

util/include/util/filesystem.hpp
```diff
@@ -0,0 +1,22 @@
...
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the sub-component's license, as noted in the LICENSE
+// file.
+
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace fs = std::filesystem;
```
WildFireFlum, line 16: Two questions:
1. Do we use a compiler on which the filesystem library is still experimental?
2. Shouldn't the specific files who want this alias do it? A global two character namespace seems risky to me.

bftclient/test/CMakeLists.txt
toly-kournik: what about gcc?

util/include/util/filesystem.hpp
toly-kournik: no idea

yuliasherman: Needs an update to 2020-2022

util/include/util/filesystem.hpp
toly-kournik: 1. it doesn't hurt
2. although it's a common namespace alias, we can replace it by a longer one.

bftclient/test/CMakeLists.txt
eyalrund: what tests did you do to make sure there is no degradation and everything is still working?

toly-kournik: what kind of degradation? We already use std::filesystem in different modules, also it's not used in a performance critical paths, but rarely, mainly during startup.

eyalrund: I ment functionality degradation. making sure everything is still working.

WildFireFlum: > what about gcc?
From https://gcc.gnu.org/gcc-9/changes.html:

Runtime Library (libstdc++)
Improved support for C++17, including:
The C++17 implementation is no longer experimental.
Parallel algorithms and <execution> (requires [Thread Building Blocks](https://github.com/oneapi-src/oneTBB) 2018 or newer).
<memory_resource>.
Using the types and functions in <filesystem> does not require linking with -lstdc++fs now.

What version of gcc is supposed to be supported?
If this is the case, it should be documented in the header.

util/include/util/filesystem.hpp
WildFireFlum: 1. There might be incompatibilities between experimental and final version, which will result is difficult bugs, so If this is not necessary, I wouldn't do it.
2. bft::fs is probably best if we are using our own header.

bftclient/test/CMakeLists.txt
toly-kournik: gcc won't compile without experimental and without -stdc++fs linkage.
right now I'm not changing the existing include/linkage behavior

WildFireFlum (approved):

PR #2448, opened 2022-03-09
Fix pruning with async execution bug

yontyon:
In case we pruning is enabled, we would like to reject any further async execution (within the same batch or another batch)
We also need to mark it as failed.
This PR handles the above


glevkovich (approved):
LGTM, Thanks @yontyon !

PR #2450, opened 2022-03-09
Make use of RawMemoryPool configurable

yuliasherman:
Use RawMemoryPool only when enableMemoryPool configuration flag is set.


teoparvanov (approved):
LGTM, thanks Yulia!

PR #2454, opened 2022-03-10
Fix test_state_transfer_with_multiple_clients for v0.14.x

WildFireFlum:
Increased the timeout for sending messages so that state transfer
is initiated correctly


teoparvanov (approved):
LGTM, thanks Ynon!

PR #2459, opened 2022-03-13
Consider client service id in the client manager and signature manager

yontyon:
Currently, the client service id is part of the client group when transaction singing is enabled, but this fact is not reflected in the code.
This PR is to fix this problem in v0.14.x


yuliasherman (approved):
LGTM. Awaits a fix for the corresponding Hermes test.

PR #2462, opened 2022-03-14
Make RequestsBatch class more thread-safe

yuliasherman:
Check for atomic variables is not thread-safe in case we need to modify a few of them atomically.


situ-s (approved):
LGTM! Thanks.

PR #2465, opened 2022-03-15
Fix a race between primary replica request retry and its cancellation

yuliasherman:
In case the primary replica pre-execution result hash is different from that non-primaries agreed on, it retries the pre-execution. Practically it never helps as the blockchain state changes all the time and the primary most probably will generate different results than others. I think the better approach is to cancel the request and wait for a client retry to give another chance for all replicas to re-generate the result.


ukg19 (approved):
Makes sense. LGTM !!
Thanks @yuliasherman

PR #2469, opened 2022-03-16
Fix misleading/disordered logs

yuliasherman:
Organize comments and the logs related to the order of principle-ids.


HristoStaykov (approved):
LGTM, thanks @yuliasherman !

PR #2470, opened 2022-03-17
Sync with 1.6

yuliasherman:
Fix a race between primary replica request retry and its cancellation 5023b7f446099c30ffb5486b03eae7ecc157cc30
Make RequestsBatch class more thread-safe 153b937d56a8b860d046bc501782830ddb19e7ab


teoparvanov (approved):
Hi Yulia, thanks for not forgetting poor master 😄

PR #2473, opened 2022-03-18
Gracefully shutdown clientservice

shgandhi:
This commit updates clientservice to handle signals such as
SIGTERM and SIGINT.

### DIFF
--- a/client/clientservice/include/client/clientservice/client_service.hpp
+++ b/client/clientservice/include/client/clientservice/client_service.hpp
@@ -32,6 +32,17 @@ class ClientService {
 
   void start(const std::string& addr, unsigned num_async_threads, uint64_t max_receive_msg_size);
 
+  // Blocks waiting for all work to complete
+  // Assumes that the caller has started the clientservice prior to this call
+  void wait();
+
+  // This method does the following -
+  // 1. Shuts down the clientservice gRPC server without a deadline and with forced cancellation
+  // 2. Shuts down all the completion queues associated with the server, and drains them using repeated next() calls
+  // 3. Waits for async gRPC server threads to return
+  // The above order is important. Reference: https://grpc.github.io/grpc/cpp/classgrpc_1_1_server_interface.html
+  void shutdown();
+
   const std::string kRequestService{"vmware.concord.client.request.v1.RequestService"};
   const std::string kEventService{"vmware.concord.client.event.v1.EventService"};
   const std::string kStateSnapshotService{"vmware.concord.client.statesnapshot.v1.StateSnapshotService"};
@@ -52,6 +63,7 @@ class ClientService {
 
   std::vector<std::unique_ptr<grpc::ServerCompletionQueue>> cqs_;
   std::vector<std::thread> server_threads_;
+  std::unique_ptr<grpc::Server> clientservice_server_;
 };
 
 }  // namespace concord::client::clientservice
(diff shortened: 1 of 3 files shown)


client/clientservice/src/client_service.cpp
```diff
@@ -33,19 +33,24 @@ void ClientService::start(const std::string& addr, unsigned num_async_threads, u
...
   health->SetServingStatus(kRequestService, true);
   health->SetServingStatus(kEventService, true);
 
-  clientservice_server->Wait();
+  clientservice_server_->Wait();
+}
+
+void ClientService::shutdown() {
+  LOG_INFO(logger_, "Shutting down clientservice");
+  clientservice_server_->Shutdown();
```
upshaw-alex, line 53: It looks like, in `client/clientservice/src/main.cpp`, we are starting the thread for signal handling, which can call `ClientService::shutdown()`, before we call `ClientService::start` (correct me if I'm misinterpreting here).

Do we not therefore need to handle the case where `ClientService::shutdown()` begins running before `ClientService::start` has made it to the line `clientservice_server_->Wait();`? (I do imagine this would likely have to involve someone interrupting the Client Service fairly soon after starting it, though it may also be technically possible if the scheduler arbitrarily delays the main thread somewhere between the point where the interrupt handler is set up and where ` clientservice_server_->Wait();` is reached). For starters, it looks like `clientservice_server_` could technically still be null here in that case. Furthermore, it looks like `ClientService::shutdown` uses `cqs_` and `server_threads_` but `ClientService::start` at some point before ` clientservice_server_->Wait();` modifies these.

client/clientservice/include/client/clientservice/client_service.hpp
```diff
@@ -32,6 +32,13 @@ class ClientService {
 
   void start(const std::string& addr, unsigned num_async_threads, uint64_t max_receive_msg_size);
 
+  // The method does the following -
```
samcramer, line 35: Nit: "This method"

client/clientservice/src/main.cpp
```diff
@@ -113,10 +115,25 @@ void initJaeger(const std::string& addr, const std::string& id) {
...
+  auto logger = logging::getLogger("concord.client.clientservice.main");
+  try {
+    if (error) {
+      LOG_ERROR(logger, error.message());
+      return;
+    }
+    LOG_INFO(logger, "Signal received (" << signum << ")");
+    clientservice->shutdown();
+  } catch (std::exception& e) {
+    std::cout << "Exception in signal handler: " << e.what() << std::endl;
```
samcramer, line 128: Should we LOG_ERROR here?

samcramer (commented):
A few nits.

client/clientservice/src/client_service.cpp
shgandhi: Thanks for the catch @upshaw-alex. I have updated the commit to fix the issue.
Note that since the call to `builder.BuildAndStart()` effectively starts the server, I decided to factor out the call to `wait` into another method called `wait`.

client/clientservice/src/main.cpp
shgandhi: Thanks @samcramer, updated.

shgandhi:
Thanks for the review @upshaw-alex and @samcramer. Updated the PR acc. to your comments.

upshaw-alex (approved):

client/clientservice/src/client_service.cpp
upshaw-alex: Thank you for fixing this.

samcramer (approved):
Looks good!

PR #2475, opened 2022-03-19
Makefile: format help, introduce new test and clean targets

glevkovich:
- Makefile: format help and add new targets for testing and cleaning
- bft.py: modify repeat options, allow breaking on 1st success/failure

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,7 +247,9 @@ test-single-suite: ## Run a single test `make test-single-suite TEST_NAME=<test 
 		ctest ${CONCORD_BFT_ADDITIONAL_CTEST_RUN_PARAMS} -V -R ${TEST_NAME} --timeout ${CONCORD_BFT_CTEST_TIMEOUT} --output-on-failure"
 
 .PHONY: test-single-apollo-case
-test-single-apollo-case: ## Run a single Apollo test case: `make test-single-apollo-case TEST_FILE_NAME=<test suite file name> TEST_CASE_NAME=<test case name> NUM_REPEATS=<number of repeats, default=1, optional> BREAK_ON_FAILURE=<TRUE|FALSE, optional>`. Test suite file name should come without *.py. Test case is expected without a class name, and must be unique.
+test-single-apollo-case: ## Run a single Apollo test case: `make test-single-apollo-case TEST_FILE_NAME=<test file name> TEST_CASE_NAME=<test case name> NUM_REPEATS=<number of repeats,default=1,optional> BREAK_ON_FAILURE=<TRUE|FALSE,optional>`. Test suite file name should come without *.py. Test case is expected without a class name, and must be unique. Example: `make test-single-apollo-case BREAK_ON_FAILURE NUM_REPEATS=100 TEST_FILE_NAME=test_skvbc_reconfiguration TEST_CASE_NAME=test_tls_exchange_client_replica_with_st`
+	@if [ -z ${TEST_FILE_NAME} ]; then echo "Error: TEST_FILE_NAME is mandatory"; exit 1; fi
+	@if [ -z ${TEST_CASE_NAME} ]; then echo "Error: TEST_CASE_NAME is mandatory"; exit 1; fi
 	$(eval PREFIX := $(shell docker run ${BASIC_RUN_PARAMS} \
 		${CONCORD_BFT_CONTAINER_SHELL} -c \
 		"mkdir -p ${CONCORD_BFT_CORE_DIR} && \
@@ -257,7 +259,6 @@ test-single-apollo-case: ## Run a single Apollo test case: `make test-single-apo
 		${CONCORD_BFT_CONTAINER_SHELL} -c \
 		"python3 scripts/apollo_list_tests.py \
 		${CONCORD_BFT_TARGET_SOURCE_PATH}/tests/apollo/ f | grep ${TEST_FILE_NAME} | grep -w ${TEST_CASE_NAME}"))
-	@echo PREFIX='$(PREFIX)'
 	@if [ -z "${PREFIX}" ] || [ -z "${POSTFIX}" ]; then \
 		echo "Error: Failed to start test, check if TEST_FILE_NAME=${TEST_FILE_NAME}" \
 			"or TEST_CASE_NAME=${TEST_CASE_NAME} exist."; exit 1;\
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -213,6 +205,23 @@ tidy-check: gen_cmake ## Run clang-tidy
...
 
+.PHONY: list-tests
+list-tests: gen_cmake ## List all tests. This one is helpful to choose which test to run when calling `make single-test TEST_NAME=<test name>`
+	docker run  ${CONCORD_BFT_USER_GROUP} ${BASIC_RUN_PARAMS} \
+		${CONCORD_BFT_CONTAINER_SHELL} -c \
+		"cd ${CONCORD_BFT_BUILD_DIR} && \
+		ctest -N"
+
+# Test targets
+NUM_REPEATS?=1
```
WildFireFlum, line 216: I already implemented repeats for apollo tests, you better use the implemented mechanism to avoid rerunning tests more times than you originally intended.
Just use the NUM_REPEATS environment variable in the with_bft_network function.
Gtest also supports repeats using the command line argument --gtest_repeat.

glevkovich: I'm trying to remember why I went this way, as I've tried to use the internal capabilities of both testing framework and backed off (It's an old branch I forgot to push).

For gtest: you are correct, thank you for the comment: it would be beter to use --gtest_repeat, there is another flag  --gtest_break_on_failure. I will replace them both.

As for Apollo tests, there is no option currently to break on failure + the parameter is not visable in the Makefile.
I think it would be good to have them both - an option to break on failure when running inside container, and keep what i did. I don't think there could be a confusion as your parameter was intended to use form inside the container +  it is not reflected in the makefile, so it can never happen. Even if yes, we will get a multiplication, it's developer responsibility.

What do you think?

glevkovich: Adding to the Apollo issue:
As I've said the repeat option you’ve added doesn’t support break on failure + it is not visible for the outside. Not all developers/simple users know how to run from inside the container.
We can have duplicated mechanisms or add this option with your loop + reflect it all the way to the Makefile, but I suggest doing this in another commit.

tests/apollo/util/test_base.py
```diff
@@ -51,22 +51,26 @@ async def wrapper(*args, **kwargs):
 
     return decorator
 
-def retry_test(max_retries: int):
+def repeat_test(max_repeats: int, break_on_first_failure: bool, break_on_first_success: bool):
     """
     Runs a test multiple times until a run succeeds
```
WildFireFlum, line 56: Change documentation and add new parameters

tests/apollo/util/test_base.py
```diff
@@ -51,22 +51,26 @@ async def wrapper(*args, **kwargs):
...
+                    if break_on_first_success:
+                        break
                 except Exception as e:
-                    log.log_message(message_type='Test attempt failed', run=i, max_retries=max_retries)
-                    if i == max_retries:
+                    log.log_message(message_type='Test attempt failed', run=i, max_repeats=max_repeats)
+                    if i == max_repeats or break_on_first_failure:
                         raise e
-
+                    print(f"Exception caught: {e}, continuing to next iteration (break_on_first_failure is False)..")
```
WildFireFlum, line 74: Maybe just add the new parameters to the log message?
you can print the logs to stdout by setting the APOLLO_LOG_STDOUT environment variable (also works with the makefile).

tests/apollo/util/test_base.py
```diff
@@ -51,22 +51,26 @@ async def wrapper(*args, **kwargs):
...
     return decorator
 
-def retry_test(max_retries: int):
+def repeat_test(max_repeats: int, break_on_first_failure: bool, break_on_first_success: bool):
     """
     Runs a test multiple times until a run succeeds
     """
-
+    assert not (break_on_first_failure and break_on_first_success), \
+        "both flags break_on_first_failure and break_on_first_success cannot be enable at the same time!"
```
WildFireFlum, line 59: *enabled

tests/apollo/util/bft.py
```diff
@@ -153,7 +153,8 @@ async def background_sender(arg1, arg2, *, task_status=trio.TASK_STATUS_IGNORED)
 
 def with_bft_network(start_replica_cmd, selected_configs=None, num_clients=None, num_ro_replicas=0,
                      rotate_keys=False, bft_configs=None, with_cre=False, publish_master_keys=False,
-                     retries=1):
+                     num_repeats=int(os.getenv("NUM_REPEATS", 1)),
+                     break_on_first_failure=os.getenv('BREAK_ON_FAILURE') == 'TRUE', break_on_first_success=False):
```
WildFireFlum, line 157: This will raise a keyerror if the environment variable is not set, which will be the case for most people.
You can use the second parameter of getenv for a default value.

tests/apollo/util/bft.py
```diff
@@ -165,6 +166,8 @@ def start_replica_cmd(builddir, replica_id, config)
     third parameter named 'config' (the exact name is important!).
     If you don't need this configuration - use two parameters callback with any names you want.
     """
+    assert not (break_on_first_failure and break_on_first_success), \
```
WildFireFlum, line 169: You are checking this condition twice, remove this redundant check and keep the one in repeat_test

tests/apollo/util/test_base.py
```diff
@@ -51,22 +51,26 @@ async def wrapper(*args, **kwargs):
...
     def decorator(async_fn):
         @wraps(async_fn)
         async def wrapper(*args, **kwargs):
-            for i in range(1, max_retries + 1):
+            for i in range(1, max_repeats + 1):
                 try:
+                    print(f"Running iteration {i}/{max_repeats}, (break_on_first_failure={break_on_first_failure},"
```
WildFireFlum, line 65: Log instead of printing, but I would not print this at all (just on failure)

tests/apollo/util/bft.py
glevkovich: That is incorrect.
Ir returns 'None' if nothing is set as default, and 'None' does not equal 'True'.

tests/apollo/util/bft.py
glevkovich: I prefer doubl checking, since I have no idea how th decorator will be used in the future.
If we have another with_bft_network1 function, and it is not checking in the decorator, we might get bugs.
In general, double checking in 2 different places is never a bad thing when therei. good reason for it.

tests/apollo/util/test_base.py
glevkovich: The main use if it will be done from the command line on the make file when someone wants to run a test 100 times - they want to see this printing to understand how many runs are left.
Since we embed the repeats in tests, it makes sense to print here and not to logs, as looking at the logs is not user-friendly - no one who chooses to run a test case 100 times goes to look at the logs when they want to know how many tests are left + there I might be issued with flashing which I don't want to deal with. 

Bottom line - I think it is better to leave it, but with one correction - print it only when num_repeats > 1.

tests/apollo/util/bft.py
WildFireFlum: Since you pass it to a function whose parameter type is a bool, i think that this:
bool(os.getenv('BREAK_ON_FAILURE', False)) makes more sense

WildFireFlum: bool(os.getenv('BREAK_ON_FAILURE', False)) Makes more sense since you pass to a function which expects a bool

tests/apollo/util/bft.py
WildFireFlum: This is exactly the point, it doesn't matter how the decorator (repeat_test) is used, because it validates its inputs.
if you implement with_bft_network__1 and use the repeat_test decorator, you'll be validated, and if you won't use it, then check for yourself like you did.

tests/apollo/util/test_base.py
```diff
@@ -51,22 +51,31 @@ async def wrapper(*args, **kwargs):
...
+    assert not (break_on_first_failure and break_on_first_success), \
+        "both flags break_on_first_failure and break_on_first_success cannot be enabled at the same time!"
     def decorator(async_fn):
         @wraps(async_fn)
         async def wrapper(*args, **kwargs):
-            for i in range(1, max_retries + 1):
+            for i in range(1, max_repeats + 1):
                 try:
+                    if (max_repeats > 1):
+                        print(f"Running iteration {i}/{max_repeats}, (break_on_first_failure={break_on_first_failure},"
```
WildFireFlum, line 70: Idk why, but I can't comment on the previous discussion.
If the reason for printing is that the logs are not easily readable, this is no longer the case if you use the APOLLO_LOG_STDOUT environment variable.
I'm generally against mixing different types of reports as it means the searching for messages becomes much harder.

tests/apollo/util/bft.py
glevkovich: We are setting BREAK_ON_FAILURE=TRUE, but for readbility reasons one might choose to set  BREAK_ON_FAILURE=FALSE.
Sothis is not a good idea...

WildFireFlum (approved):

PR #2477, opened 2022-03-20
Exclude operator id from sig manager (v0.14.x)

yontyon:
When transaction signing is enabled, the operator id is not part of any client group but it is still considered as an external client.
This causes the replicas to crash in case transaction signing is enabled.

### DIFF
--- a/bftengine/src/bftengine/ReplicasInfo.cpp
+++ b/bftengine/src/bftengine/ReplicasInfo.cpp
@@ -112,7 +112,7 @@ ReplicasInfo::ReplicasInfo(const ReplicaConfig& config,
         for (auto i = start; i < (end - ((uint16_t)config.operatorEnabled_)); ++i) {
           ret.insert(i);
         }
-        ret.insert(end + config.numOfClientServices - 1);
+        if (config.operatorEnabled_) ret.insert(end + config.numOfClientServices - 1);
         if (start != end)
           LOG_INFO(GL,
                    "Principal ids in _idsOfExternalClients: " << start << " to "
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicasInfo.hpp
```diff
@@ -107,6 +108,7 @@ class ReplicasInfo {
   const std::set<PrincipalId> _idsOfExternalClients;
   const std::set<PrincipalId> _idsOfClientServices;
   const std::set<PrincipalId> _idsOfInternalClients;
+  const std::set<PrincipalId> _idsOfOperators;
```
yuliasherman, line 111: Let's keep also the order of initializations correct. _idsOfOperators should appear between _idsOfClientServices and _idsOfInternalClients.

bftengine/src/bftengine/ReplicasInfo.cpp
```diff
@@ -143,6 +143,13 @@ ReplicasInfo::ReplicasInfo(const ReplicaConfig& config,
         }
         if (start != end) LOG_INFO(GL, "Principal ids in _idsOfInternalClients: " << start << " to " << end - 1);
         return ret;
+      }()},
+      _idsOfOperators{[&config]() {
```
yuliasherman, line 147: Let's keep also the order of initialisations correct. _idsOfOperators should appear between _idsOfClientServices and _idsOfInternalClients. Please move the log 
if (config.operatorEnabled_) LOG_INFO(GL, "Operator id: " << end); from _idsOfClientServices to here.

bftengine/src/bftengine/ReplicasInfo.cpp
```diff
@@ -112,7 +112,7 @@ ReplicasInfo::ReplicasInfo(const ReplicaConfig& config,
         for (auto i = start; i < (end - ((uint16_t)config.operatorEnabled_)); ++i) {
           ret.insert(i);
         }
-        ret.insert(end + config.numOfClientServices - 1);
+        if (config.operatorEnabled_) ret.insert(end + config.numOfClientServices - 1);
```
yuliasherman, line 115: Please remove log
if (config.operatorEnabled_) LOG_INFO(GL, "Operator id: " << end);
from _idsOfClientServices{[&config]()
to not be printed twice.

teoparvanov:
Hi @yontyon , thank you for the fix attempt. What testing have you done for these changes? Are there unit, Apollo tests? Hermes? Manual tests? Please clarify in the description,

CC: @eyalrund

yuliasherman (approved):
LGTM

PR #2483, opened 2022-03-22
Makefile, test_base: minor fixes on 5a393f7

glevkovich:
1) Makefile - fix type
2) test_base: print traceback when exception thrown and ignored since
we continue into next test iteration.


StefanPalashev (approved):
Thanks, @glevkovich.

PR #2486, opened 2022-03-22
Refactor time service logic

arc-vmware:
Instead of having the time request as a client request inside the prePrepare message, we add it as a metadata in the prePrepare message

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -165,10 +165,11 @@ void ReplicaImp::registerMsgHandlers() {
 template <typename T>
 void ReplicaImp::messageHandler(MessageBase *msg) {
   T *trueTypeObj = new T(msg);
-  validateMessage(trueTypeObj);
   if (isCollectingState()) {
     // Extract only required information and handover to ST thread
-    peekConsensusMessage<T>(msg);
+    if (validateMessage(trueTypeObj)) {
+      peekConsensusMessage<T>(msg);
+    }
   }
   delete msg;
 
@@ -639,7 +640,7 @@ std::pair<PrePrepareMsg *, bool> ReplicaImp::buildPrePrepareMsgBatchByOverallSiz
 std::pair<PrePrepareMsg *, bool> ReplicaImp::buildPrePrepareMsgBatchByRequestsNum(uint32_t requiredRequestsNum) {
   ConcordAssertGT(requiredRequestsNum, 0);
 
-  if (requestsQueueOfPrimary.size()) {
+  if (requestsQueueOfPrimary.size() < requiredRequestsNum) {
     LOG_DEBUG(GL,
               "Not enough messages in the primary replica queue to fill a batch"
                   << KVLOG(requestsQueueOfPrimary.size(), requiredRequestsNum));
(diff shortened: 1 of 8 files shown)


bftengine/src/bftengine/TimeServiceManager.hpp
```diff
@@ -81,56 +81,12 @@ class TimeServiceManager {
...
-  [[nodiscard]] bool isPrimarysTimeWithinBounds(const impl::PrePrepareMsg& msg) const {
-    ConcordAssertGE(msg.numberOfRequests(), 1);
-
-    auto it = impl::RequestsIterator(&msg);
-    char* requestBody = nullptr;
-    ConcordAssertEQ(it.getCurrent(requestBody), true);
-
-    ClientRequestMsg req((ClientRequestMsgHeader*)requestBody);
-    return isPrimarysTimeWithinBounds(req);
+  [[nodiscard]] ConsensusTickRep getTimePoint() {
```
yontyon, line 84: NIT: I know I gave the name of this method, but timeService already has a getTime method, so it worth thinking on a more distinguished name

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
```diff
@@ -38,6 +38,7 @@ class PrePrepareMsg : public MessageBase {
     EpochNum epochNum;
     uint16_t flags;
     uint64_t batchCidLength;
+    int64_t time;
```
yontyon, line 41: Please set it to 0 in the PrePrepare constructor (we rely on the 0 value to indicate that there is not time record)

yontyon (changes requested):
Great PR.
I have one comment (nullify the time in the prePrepare constructor).
Thanks

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
arc-vmware: Thanks @yontyon, fixing this.

bftengine/src/bftengine/TimeServiceManager.hpp
arc-vmware: I think the getTime is a good name since that is returning the time of the time service. 
But getTimePoint can be renamed to getClockTimePoint() as it actually gets the time of the clock of primary and returns the same. @yontyon please let me know if this is a good name for this function?

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
arc-vmware: Another question is, the time is int64_t. So it can take negative values also. Although negative time makes sense in the Minkowski space-time metric in concord it will always be positive. So should we also change all the references:
 if (msg->getTime() == 0)  to  if (msg->getTime() <= 0) 
Since  time_since_epoch() of the clock may return negative values.
https://stackoverflow.com/questions/29436297/c-chrono-giving-negative-epoch-time

arc-vmware: Discussed offline with @yontyon , he said he is fine with the check for the negative range. So resolving the comment.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1100,14 +1091,14 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsg *msg) {
     const bool slowStarted = (msg->firstPath() == CommitPath::SLOW || seqNumInfo.slowPathStarted());
 
     bool time_is_ok = true;
-    if (config_.timeServiceEnabled && msg->numberOfRequests() > 0) {
-      if (!time_service_manager_->hasTimeRequest(*msg)) {
-        LOG_WARN(CNSUS, "PrePrepare will be ignored");
+    if (config_.timeServiceEnabled) {
+      if (msg->getTime() <= 0) {
+        LOG_WARN(CNSUS, "No timePoint in the prePrepare, PrePrepare will be ignored");
```
ukg19, line 1096: nit: Can we also print the value of msg->getTime(), will be useful in debugging in case any issues.

arc-vmware: Thanks @ukg19 , I will add this in the next commit.

yontyon (approved):
LGTM!
Thanks

PR #2487, opened 2022-03-22
Prevent PreProcessor from sending an empty PreProcessBatchReplyMsg

yuliasherman:
Fix a race in the batch completion process when two batch requests could cause sending a good and empty reply to the primary replica.

### DIFF
--- a/bftengine/src/preprocessor/messages/PreProcessBatchReplyMsg.cpp
+++ b/bftengine/src/preprocessor/messages/PreProcessBatchReplyMsg.cpp
@@ -56,8 +56,13 @@ void PreProcessBatchReplyMsg::validate(const ReplicasInfo& repInfo) const {
   }
 
   const auto numOfMessagesInBatch = msgBody()->numOfMessagesInBatch;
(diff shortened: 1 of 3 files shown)


bftengine/src/preprocessor/messages/PreProcessBatchReplyMsg.cpp
```diff
@@ -56,8 +56,13 @@ void PreProcessBatchReplyMsg::validate(const ReplicasInfo& repInfo) const {
   }
 
   const auto numOfMessagesInBatch = msgBody()->numOfMessagesInBatch;
-  if (!numOfMessagesInBatch || (numOfMessagesInBatch > MAX_BATCH_SIZE)) {
-    LOG_WARN(logger(), "Too many messages in batch" << KVLOG(numOfMessagesInBatch));
```
ukg19, line 60: Thanks for fixing this misleading log.

ukg19:
Thanks for fixing the race condition @yuliasherman . 
LGTM !!

ukg19 (approved):

PR #2490, opened 2022-03-23
Add support to recover prune requests

cloudnoize:
If a replica terminates while performing pruning, on restart, it will re-execute the requests that were part of the last sequence number that was executed.
Requests that were already executed before the crash should be revered prior to the re-execution or should not have an additional side effect if re-executed.

Concurrent pruning is triggered by tick commands, when those commands are being re-executed, the concurrent pruning handler re-executes them regardless of whether they have been executed before resulting in diverge in state from other replicas as double execution might occur.

This PR will extend the information that is saved in the reserved pages before each invocation to include also the bft sequence number and the internal tick order within the sequence number.
the internal order inside a preprepare msg is deterministic therefore on re-execution the same request must get the same sequence number and internal order.

This information will be used in order to total order the ticks.
A corresponding MR will include this change and will implement the handler logic that checks whether a received tick is newer than the saved tick.

### DIFF
--- a/ccron/include/ccron/tick.hpp
+++ b/ccron/include/ccron/tick.hpp
@@ -28,11 +28,14 @@ struct Tick {
   // Time that is the same across replicas after being agreed upon.
   // It is optional as the time service that provides agreement across replicas is itself optional.
   std::optional<std::chrono::milliseconds> ms_since_epoch;
+
+  // a sequnce number, may contain several ticks, this field is used in order to order them.
+  std::uint64_t internal_order{};
 };
 
 inline bool operator==(const Tick& l, const Tick& r) {
   return (l.component_id == r.component_id && l.bft_sequence_num == r.bft_sequence_num &&
-          l.ms_since_epoch == r.ms_since_epoch);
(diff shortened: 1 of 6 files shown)


ccron/include/ccron/tick.hpp
```diff
@@ -28,11 +28,14 @@ struct Tick {
   // Time that is the same across replicas after being agreed upon.
   // It is optional as the time service that provides agreement across replicas is itself optional.
   std::optional<std::chrono::milliseconds> ms_since_epoch;
+
+  // a sequnce number, may contain several ticks, this field is used in order to order them.
+  std::uint64_t internal_order{};
 };
 
 inline bool operator==(const Tick& l, const Tick& r) {
```
nkgeorgiev, line 36: You mentioned in the description that we will total order the ticks. Don't we need some form of operator< for this?

cloudnoize: the ordering is done against the Batch cmf msg that is saved in the reserved pages and saves in addition to the sequence number and internal order also the target block id.

PR #2492, opened 2022-03-23
Apollo: slow down ST by adding a DelayStateTransferMsgStrategy

glevkovich:
New command line parameter added to TesterReplica, long-form only:
--delay-state-transfer-messages-millisec MS, while MS is the time in ms
to delay each sent ST message.

To be able to monitor ST better, we add a delay in 3 test suites which
use ST extensively:
test_skvbc_persistence,test_skvbc_reconfiguration
and test_skvbc_state_transfer.

Reduce all tests which open 10 checkpoints gap of 10 into 4, due to
the fact that this was initially done to slow down the ST cycle.

### DIFF
--- a/tests/simpleKVBC/TesterReplica/setup.cpp
+++ b/tests/simpleKVBC/TesterReplica/setup.cpp
@@ -14,6 +14,7 @@
 #include <thread>
 #include <sys/param.h>
 #include <string>
+#include <sstream>
 #include <cstring>
 #include <getopt.h>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


tests/simpleKVBC/TesterReplica/setup.cpp
```diff
@@ -94,8 +98,14 @@ std::unique_ptr<TestSetup> TestSetup::ParseArgs(int argc, char** argv) {
     std::string byzantineStrategies;
     bool is_separate_communication_mode = false;
     int addAllKeysAsPublic = 0;
+    int stateTransferMsgDelayMs = 0;
 
+    // do not change order of the next options, as some might use an option index!
     static struct option longOptions[] = {
+        // long format only options should be put here
+        {"delay-state-transfer-messages-millisec", required_argument, 0, 0},
```
StefanPalashev, line 106: I think this should be an optional argument, since the byzantine strategies are. Also since it is a long format parameter, maybe the `stateTransferMsgDelayMs` variable can be used with a flag, which will save all the parsing.

StefanPalashev (approved):
I think this will be a valuable asset to our byzantine strategies.
Thanks, @glevkovich!

arc-vmware (approved):
LGTM!!!

PR #2498, opened 2022-03-25
Address parse response array any issue

cjn:
Added changes to update response based on request type being any or buffer.


arc-vmware (approved):
Thanks @cjn !!!
LGTM!!!

PR #2499, opened 2022-03-25
Request Service: Populate response if it can be parsed from array

shgandhi:
Fixes the bug introduced a couple of weeks back wherein setting
of concord-client send request's response via callback was getting
incorrectly skipped.


samcramer (approved):
Looks good.

shgandhi:
Closing the PR. @cjn is already working on a fix - see https://github.com/vmware/concord-bft/pull/2498.

PR #2501, opened 2022-03-28
Verify that Apollo client's communication protocol matches replica's

WildFireFlum:
Added log messages to Apollo
Ensured that all log messages of an Apollo run are saved in one
directory
Changed CommFactory so that the communication protocol can be uniquely
identified from the log
Added log messages with metadata of binaries which Apollo
invokes (hash, path)


yontyon (approved):
LGTM!
Thanks

PR #2502, opened 2022-03-28
Handle NULL cases for ps

harsht1737:
Handling NULL cases for persistent storage ptr in ReplicaSpecificInfoManager.cpp
This handling is already present for ReplicaImp.cpp, ExecutionEngine.cpp and other consumers.
SimpleTest currently passes persistent storage pointer as NULL as metadata storage parameter default value is NULL


arc-vmware:
@harsht1737 please add description to the PR.

arc-vmware (approved):
LGTM.

PR #2503, opened 2022-03-28
Fix batch release process

yuliasherman:
Avoid a race between increaseNumOfCompletedReqs and releaseReqsAndSendBatchedReplyIfCompleted functions.


situ-s (approved):
LGTM! Thanks for fixing @yuliasherman.

PR #2504, opened 2022-03-28
Added changes to fix compatibility issues for using request.proto raw and typed requests

cjn:
Added changes to fix compatibility issue for modules using raw request/response and typed_request/response

1> Rename raw request to old attribute name
2> Removed one of constraint for backward compatibility for referring to raw request type.
3> Replaced changes in code during request and response processing


teoparvanov (approved):
LGTM!

PR #2507, opened 2022-03-28
S3 path prefix default value.

toly-kournik:
Change S3 Object Store path prefix default value to an empty string.
Apollo tests generate random value for s3-path-prefix and there's no need to generate a random value in the s3 configuration parsing code.


yontyon (approved):
LGTM!
Thanks

PR #2511, opened 2022-03-29
Less frequent client request retransmission to primary

mdarade:
Reduce frequency of retransmission of client request msg received
by non primary replica from every 1 sec to every 4 sec


yuliasherman:
5*x > 3*x. It looks like you are going to retransmit more frequently.

yontyon (approved):
LGTM!

mdarade:
> 5_x > 3_x. It looks like you are going to retransmit more frequently.

Earlier 5x was valid only once after that every 1 sec retransmission was done.
Now every 4 sec retransmission would be done as after every timeout we reset to current time.

PR #2513, opened 2022-03-29
Less frequent client request retransmission to primary

mdarade:
Reduce frequency of retransmission of client request msg received
by non primary replica from every 1 sec to every 4 sec


yontyon (approved):
LGTM!

PR #2518, opened 2022-03-31
RVBManager: Fix bug while restarted before adding a single RVB

glevkovich:
In that scenario, with current code assert will fail when entering
addRvbDataOnBlockRange. Here we change some of the code while calling to
addRvbDataOnBlockRange i(2 places).


mdarade (approved):
Thanks Gil.
Looks good.

StefanPalashev (approved):
LGTM, thanks @glevkovich.

PR #2521, opened 2022-03-31
Adaptive pruning computation change

titusen:


titusen:
New one is easier with resolved conflicts by cherry-pick

PR #2526, opened 2022-04-01
Call findGlobalEventGroupId() only once per subscription

shgandhi:
This PR updates syncAndSendEventGroups method s.t.,
computation heavy method findGlobalEventGroupId() is no longer used
to get the global event group ID corresponding to the requested
external event group ID when syncing live update queue with storage.
Instead we calculate global event group ID corresponding to
global event group ID of the first live update to be read by the
client on the fly.

### DIFF
--- a/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
+++ b/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
@@ -539,8 +539,7 @@ void KvbAppFilter::readEventGroups(EventGroupId external_eg_id_start,
     }
     ext_eg_id += 1;
   }
-  setNextExternalEgIdToRead(++ext_eg_id);
-  setLastGlobalEgIdReadForClient(global_eg_id);
+  setLastEgIdsRead(ext_eg_id, global_eg_id);
 }
 
 void KvbAppFilter::readEventGroupRange(EventGroupId external_eg_id_start,
@@ -599,8 +598,7 @@ string KvbAppFilter::readEventGroupHash(EventGroupId external_eg_id) {
     throw KvbReadError(msg.str());
   }
   KvbFilteredEventGroupUpdate filtered_update{external_eg_id, filterEventsInEventGroup(result.global_id, event_group)};
-  setLastGlobalEgIdReadForClient(result.global_id);
-  setNextExternalEgIdToRead(++external_eg_id);
+  setLastEgIdsRead(external_eg_id, result.global_id);
   return hashEventGroupUpdate(filtered_update);
 }
 
@@ -690,13 +688,11 @@ kvbc::categorization::EventGroup KvbAppFilter::getEventGroup(kvbc::EventGroupId 
   return event_group_out;
 }
 
-void KvbAppFilter::setNextExternalEgIdToRead(uint64_t ext_eg_id) { next_ext_eg_id_to_read_ = ext_eg_id; }
+void KvbAppFilter::setLastEgIdsRead(uint64_t last_ext_eg_id_read, uint64_t last_global_eg_id_read) {
+  last_ext_and_global_eg_id_read_ = std::make_pair(last_ext_eg_id_read, last_global_eg_id_read);
+}
 
-uint64_t KvbAppFilter::getNextExternalEgIdToRead() const { return next_ext_eg_id_to_read_; }
-
-void KvbAppFilter::setLastGlobalEgIdReadForClient(uint64_t global_eg_id) { last_global_eg_id_read_ = global_eg_id; }
-
-uint64_t KvbAppFilter::getLastGlobalEgIdReadForClient() const { return last_global_eg_id_read_; }
+std::pair<uint64_t, uint64_t> KvbAppFilter::getLastEgIdsRead() { return last_ext_and_global_eg_id_read_; }
 
 }  // namespace kvbc
 }  // namespace concord
--- a/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
+++ b/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
@@ -842,7 +842,7 @@ class ThinReplicaImpl {
   // returns when the next update can be taken from the given live updates.
   // We assume that the Commands handler is already filling the queue. Also, we
   // don't care if the queue fills up. In that case, the caller won't be able to
-  // use the queue as soon as he consumes it.
+  // use the queue as soon as they consume it.
   template <typename ServerContextT, typename ServerWriterT, typename DataT>
   void syncAndSend(ServerContextT* context,
                    kvbc::BlockId start,
@@ -912,7 +912,7 @@ class ThinReplicaImpl {
   // returns when the next update can be taken from the given live updates.
   // We assume that the Commands handler is already filling the queue. Also, we
   // don't care if the queue fills up. In that case, the caller won't be able to
-  // use the queue as soon as he consumes it.
+  // use the queue as soon as they consume it.
   template <typename ServerContextT, typename ServerWriterT, typename DataT>
   void syncAndSendEventGroups(ServerContextT* context,
                               kvbc::EventGroupId start,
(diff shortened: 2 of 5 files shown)


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
+      }
+
+      auto ext_eg_id = end + 1;  // next external event group ID to be read
+
+      // live_updates in sync with KVB i.e., event group ID associated with the oldest live update is the
+      // next_global_eg_id_to_read.
+      if (live_updates->oldestEventGroupId() == next_global_eg_id_to_read) {
+        auto filtered_eg_update = kvb_filter->filterEventGroupUpdate(live_updates->oldestEventGroup());
+        // If the oldest live update is not for the requesting client, keep reading from the live update queue
+        // until the first relevant live update is reached. Drop all non-relevant live updates read along the way.
```
robem, line 963: This comment should sit on top of the if-condition.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
+        live_updates->PopEventGroup(sub_eg_update);
+        LOG_INFO(logger_, "Sync dropping " << sub_eg_update.event_group_id);
+        is_update_available = false;
+        continue;
+      }
+
+      auto ext_eg_id = end + 1;  // next external event group ID to be read
+
+      // live_updates in sync with KVB i.e., event group ID associated with the oldest live update is the
+      // next_global_eg_id_to_read.
```
robem, line 959:
```suggestion

      // At this point, all event groups in the live update queue are newer than anything we read from storage earlier.
      
```

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
+      if (live_updates->oldestEventGroupId() == next_global_eg_id_to_read) {
+        auto filtered_eg_update = kvb_filter->filterEventGroupUpdate(live_updates->oldestEventGroup());
```
robem, line 961: We will read and filter the event group twice if this an event group for the client. Once here, and once after we return from this function.
Let's add a comment or TODO so we can take a look at potentially optimizing this later.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
-      ConcordAssertGT(kvb_filter->newestExternalEventGroupId(), end + 1);
-      start = end + 1;
-      end = kvb_filter->newestExternalEventGroupId();
+      // fill the gap
+      LOG_INFO(logger_,
+               "Sync filling gap [" << next_global_eg_id_to_read << ", " << live_updates->oldestEventGroupId() << "]");
+      while (live_updates->oldestEventGroupId() > next_global_eg_id_to_read) {
+        // gap exists between next_global_eg_id_to_read and live_updates->oldestEventGroupId()
+        // Let's read from data table directly, filter and send updates, until next_global_eg_id_to_read equals
+        // live_updates->oldestEventGroupId()
```
robem, line 982: I suggest to put this comment with the "fill the gap" comment or more specifically, above the while loop.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
+          throw kvbc::KvbReadError(msg.str());
+        }
+        auto eg_update_from_storage = kvbc::EgUpdate{event_group.id, std::move(event_group)};
+        auto filtered_eg_update = kvb_filter->filterEventGroupUpdate(eg_update_from_storage);
+        if (!filtered_eg_update) {
+          next_global_eg_id_to_read++;
+          continue;
+        }
+        // Overwrite event group ID in the filtered update to external event group ID
+        // We don't want to expose the global event group ID to the client
```
robem, line 996: More important, every client has their own event group id counter. We don't want to expose the global one but it is also incorrect.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
+        filtered_eg_update.value().event_group_id = ext_eg_id;
 
-      LOG_INFO(logger_, "Sync filling gap [" << start << ", " << end << "]");
-      readAndSendEventGroups<ServerContextT, ServerWriterT, DataT>(logger_, context, stream, start, end, kvb_filter);
-    }
+        kvb_filter->setNextExternalEgIdToRead(++ext_eg_id);
+        kvb_filter->setLastGlobalEgIdReadForClient(next_global_eg_id_to_read++);
```
robem, line 1000: Optional: Add a comment which explains that those two will hold the ext-to-global-eg-id-mapping as long as the subscription is active so that we don't have to recompute it.

Now, this makes me think, given that we should always update both at the same time, maybe a `std::pair<uint64,uint64>` would be a better representation.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
-    // Overlap:
-    // If we read updates from KVB that were added to the live updates already
-    // then we just need to drop the overlap and return
-    ConcordAssert(live_updates->oldestEventGroupId() <= last_global_eg_id_read);
-    SubEventGroupUpdate update;
-    do {
-      live_updates->PopEventGroup(update);
-      LOG_INFO(logger_, "Sync dropping " << update.event_group_id);
-    } while (update.event_group_id < last_global_eg_id_read);
+        // send update
```
robem, line 1002: Given the clear name of the following functions, I believe this comment adds no value.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -933,72 +933,84 @@ class ThinReplicaImpl {
...
+        if (!filtered_eg_update) {
           SubEventGroupUpdate sub_eg_update;
           live_updates->PopEventGroup(sub_eg_update);
+          LOG_INFO(logger_, "Sync dropping upon filtering " << sub_eg_update.event_group_id);
+          num_updates_filtered_out++;
           is_update_available = false;
+          next_global_eg_id_to_read += 1;
+          continue;
         }
+        break;
```
robem, line 973:
```suggestion
        // The next event group in the live update queue has events for this client.
        return; 
```

robem (approved):
This looks good to me 👍  Thanks for the taking this on 💪  and thanks for the great commit messages and tests!!

shgandhi:
Thanks so much for the review @robem. I have updated the PR according to your comments.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -690,13 +688,11 @@ kvbc::categorization::EventGroup KvbAppFilter::getEventGroup(kvbc::EventGroupId
   return event_group_out;
 }
 
-void KvbAppFilter::setNextExternalEgIdToRead(uint64_t ext_eg_id) { next_ext_eg_id_to_read_ = ext_eg_id; }
-
-uint64_t KvbAppFilter::getNextExternalEgIdToRead() const { return next_ext_eg_id_to_read_; }
-
-void KvbAppFilter::setLastGlobalEgIdReadForClient(uint64_t global_eg_id) { last_global_eg_id_read_ = global_eg_id; }
+void KvbAppFilter::setLastEgIdsRead(std::pair<uint64_t, uint64_t> last_ext_and_global_eg_id_read) {
```
robem, line 691: Nit/Optional: The fact that the storage format is a pair doesn't need to be part of the function signature. Simply `setLastEgIdsRead(uint64_t external, uint64_t global)` is sufficient.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -690,13 +688,11 @@ kvbc::categorization::EventGroup KvbAppFilter::getEventGroup(kvbc::EventGroupId
...
+  last_ext_and_global_eg_id_read_ = last_ext_and_global_eg_id_read;
+}
 
-uint64_t KvbAppFilter::getLastGlobalEgIdReadForClient() const { return last_global_eg_id_read_; }
+std::pair<uint64_t, uint64_t> KvbAppFilter::getLastEgIdsRead() { return last_ext_and_global_eg_id_read_; }
```
robem, line 695: The function should document what the first and second value is semantically.

robem (approved):

PR #2527, opened 2022-04-03
Fix a bug where status is resolved only on the first funciton call

WildFireFlum:


samcramer:
Please explain how you tested this fix.

Also, I'm curious as to the motivation for making the initial change. As far as I know, the prior version worked properly.  I find the prior version simpler and easier to read than this version.  Is there an advantage to the change that I'm not seeing?

WildFireFlum:
> Please explain how you tested this fix.
> 
> Also, I'm curious as to the motivation for making the initial change. As far as I know, the prior version worked properly. I find the prior version simpler and easier to read than this version. Is there an advantage to the change that I'm not seeing?

I am trying to find a way of testing it inside apollo at the moment, as this code was not tested at all, if you have any ideas as to how to test them they would be welcome.

The motivation is 24 lines of code for the previous implementation vs 17 in the current one, and 3 lines per new status for the previous implementation vs a single line in the newer one (depending of course on the logic behind it).

There was also a redundancy with the calls to status.output.set_value, but this could have been changed without using a map.

You might not think that line of code count is a big deal when the total amount saved ins 7, but in a file with 5K+ lines of code, whenever I can save some, I would.

samcramer:
I would test it by hand for now.  The JIRA ticket explains how to do that.  Obviously it is better if there is an automated test; at this point there isn't.  I don't think it is a good idea to issue an MR without testing the code that will be merged.

We disagree regarding the importance of the number of lines in the function.  I think that correctness, avoidance of unnecessary work (opportunity cost), simplicity and readability are more important than saving 7 lines of code.   If the file has 5K+ lines of code, at some point we can address that in a broader way.  Right now, I don't see the need to rewrite code that is simple and that works.

WildFireFlum:
> I would test it by hand for now. The JIRA ticket explains how to do that. Obviously it is better if there is an automated test; at this point there isn't. I don't think it is a good idea to issue an MR without testing the code that will be merged.
> 
> We disagree regarding the importance of the number of lines in the function. I think that correctness, avoidance of unnecessary work (opportunity cost), simplicity and readability are more important than saving 7 lines of code. If the file has 5K+ lines of code, at some point we can address that in a broader way. Right now, I don't see the need to rewrite code that is simple and that works.

I added an apollo test suite which tests this logic in concord-bft.
I disagree with you, I think that the previous implementation did not scale well for the amount of people working on this codebase, as adding new statuses would add to the existing code bloat. 
I do agree that I did not save many code lines myself in this case, but my mindset is set to cleaning up as I go through the code. I count on myself, our extensive test suites and my reviewers to catch errors when making said changes, but I do believe that they are necessary. 
In this case, there were no tests, so I implemented them.

cloudnoize (approved):
LGTM!

tests/apollo/test_skvbc_diagnostics.py
```diff
@@ -0,0 +1,122 @@
...
+    def validate_status_describe_command(self, command_args: Sequence[str], replica_id:int = 0):
+        invalid_response = '*--DESCRIPTION_NOT_FOUND--*'
+        output = self.run_ctl_command(['status', 'describe'] + list(command_args), replica_id)
+        assert invalid_response not in output
+        return output
+
+    @with_trio
+    @parameterize(is_status_valid=[True, False])
+    @with_bft_network(start_replica_cmd, bft_configs=[BFTConfig(n=4)], rotate_keys=ROTATE_KEYS)
+    @verify_linearizability()
```
teoparvanov, line 96: you don't need linearizability checks for diagnostic status tests

WildFireFlum: I do, since I need the skvbc object which needs a tracker...

teoparvanov (approved):
LGTM, thanks for making it possible to test the diagnostic server from Apollo!

tests/apollo/test_skvbc_diagnostics.py
teoparvanov: @WildFireFlum actually, the tracker is optional in the skvbc object:
https://github.com/vmware/concord-bft/blob/master/tests/apollo/test_skvbc.py#L64

This is to enable testing with/without linearizability checks.

WildFireFlum: Changed, thank you.

samcramer (changes requested):
Thank you for adding the test.

Please revert the change you have made previously to `ReplicaImp::onInternalMsg()` (commit f2e16e24) and please remove the changes to that method in this change set. It is not necessary, since there is no evidence that this method causes performance problems.

The change to `ReplicaImp::onInternalMsg()` replaces simple and easy to read code with an implementation which is considerably more complicated.  The previous implementation used simple `if` and `return` statements.  The new code adds a non-static `std::unordered_map` and a number of lambdas, one of which differs from the others because it does not capture `this`. It is undoubtedly more complicated than the code it replaces, but provides no offsetting actual advantage.

In the very unlikely event that we find that the previous implementation of this method causes performance problems we can revisit this issue.  In the meantime, this part of the change is unnecessary code churn which has already resulted in a P1 bug.

teoparvanov:
> Thank you for adding the test.
> 
> Please revert the change you have made previously to `ReplicaImp::onInternalMsg()` (commit [f2e16e2](https://github.com/vmware/concord-bft/commit/f2e16e24246b2bb5bdc7bde392206af96e6e0188)) and please remove the changes to that method in this change set. It is not necessary, since there is no evidence that this method causes performance problems.
> 
> The change to `ReplicaImp::onInternalMsg()` replaces simple and easy to read code with an implementation which is considerably more complicated. The previous implementation used simple `if` and `return` statements. The new code adds a non-static `std::unordered_map` and a number of lambdas, one of which differs from the others because it does not capture `this`. It is undoubtedly more complicated than the code it replaces, but provides no offsetting actual advantage.
> 
> In the very unlikely event that we find that the previous implementation of this method causes performance problems we can revisit this issue. In the meantime, this part of the change is unnecessary code churn which has already resulted in a P1 bug.

Hi @samcramer and @WildFireFlum , as a compromise solution, my suggestion is to merge the Apollo test, but without the change in `ReplicaImp::onInternalMsg()`. The ability to test the diagnostics server in the open source code is valuable IMO.

samcramer:
Yes, @teoparvanov I think that is a good plan.  The test is very useful and is a nice piece of work. Thanks @WildFireFlum for implementing it.

samcramer:
Note that commit f2e16e24246b2bb5bdc7 will need to be backed out if the test is merged.

teoparvanov (approved):

samcramer:
Thanks very much @WildFireFlum !  It's really great to have the test in place.

PR #2534, opened 2022-04-04
Sleep in async execution until pruning is completed

yontyon:
This is to fix the following bug with async execution.
Assume two replicas, r1, r2, and assume 2 sequence numbers s1 and s2.
Also, let's assume that one of s1's requests is a pruning request and that pruning is happening in an async mode.
The following inconsistency can happen:
1. r1, schedule s1 and s2 for execution
2. s1 is executed and s2 is rejected (here https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/RequestHandler.cpp#L175)

In the meantime, r2 does the following:
1. schedule s1 for execution
2. execute s1
3. schedule s2 for execution
In r2 case, s2 will be executed once we did with pruning.
The end result is that s2 was executed in r2 but not in r1.

To fix the above, we let the execution thread sleep in the sequence number that is scheduled right after the pruning request (s2 in our case) instead of ignoring it.

### DIFF
--- a/bftengine/include/bftengine/ControlStateManager.hpp
+++ b/bftengine/include/bftengine/ControlStateManager.hpp
@@ -20,6 +20,7 @@
 #include "Serializable.h"
 #include "SysConsts.hpp"
 #include "callback_registry.hpp"
+#include "assertUtils.hpp"
 
 namespace bftEngine {
 class ControlStateManager {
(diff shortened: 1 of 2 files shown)


bftengine/include/bftengine/ControlStateManager.hpp
```diff
@@ -34,8 +35,12 @@ class ControlStateManager {
   void wedge() { wedged = true; }
   void unwedge() { wedged = false; }
   void markRemoveMetadata(bool include_st = true) { removeMetadataCbRegistry_.invokeAll(include_st); }
-  void setPruningProcess(bool onPruningProcess) { onPruningProcess_ = onPruningProcess; }
+  void setPruningProcess(bool onPruningProcess) {
```
toly-kournik, line 38: what happens if setPrunningProcess(true) is called multiple times in a row?

yontyon: right, I have added an assert

toly-kournik (approved):
LGTM!

PR #2536, opened 2022-04-04
Fix data race on cid_arrival_map_.

HristoStaykov:
This is a fix for a reported corruption in cid_arrival_map_
from an ASAN run.

### DIFF
--- a/client/client_pool/src/concord_client_pool.cpp
+++ b/client/client_pool/src/concord_client_pool.cpp
@@ -506,18 +506,21 @@ void ConcordClientPool::processReplies(ClientPtr &client,
   ClientPoolMetrics_.average_req_dur_gauge.Get().Set((uint64_t)average_req_dur_.avg());
   if (average_req_dur_.numOfElements() == 1000) average_req_dur_.reset();  // reset the average every 1000 samples
   ClientPoolMetrics_.executed_requests_counter++;
-  metricsComponent_.UpdateAggregator();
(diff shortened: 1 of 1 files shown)


client/client_pool/src/concord_client_pool.cpp
```diff
@@ -506,18 +506,21 @@ void ConcordClientPool::processReplies(ClientPtr &client,
...
-    cid_arrival_map_.erase(reply.cid);
-    duration = std::chrono::duration_cast<std::chrono::milliseconds>(before_send - arrival_time).count();
-    average_cid_receive_dur_.add(duration);
-
-    auto after_send = client->getAndDeleteCidResponseTime(reply.cid);
-    duration = std::chrono::duration_cast<std::chrono::milliseconds>(finish - after_send).count();
-    average_cid_close_dur_.add(duration);
+  {
+    std::unique_lock<std::mutex> lock(clients_queue_lock_);
+    metricsComponent_.UpdateAggregator();
```
yuliasherman, line 511: Why does this need to be under a lock?

HristoStaykov: I agree, I will move it out of the lock scope. Metrics have their own sync mechanism.

teoparvanov (approved):
LGTM, thanks Hristo!

yuliasherman (approved):
LGTM

PR #2537, opened 2022-04-04
ClientPoolTimer: Fix busy-wait loop

robem:
When there are no timeouts that need to be processed then the timer thread spins in a busy-wait loop. This PR is fixing the problem. Please see commit messages for details.

### DIFF
--- a/client/client_pool/include/client/client_pool/client_pool_timer.hpp
+++ b/client/client_pool/include/client/client_pool/client_pool_timer.hpp
@@ -15,12 +15,9 @@
 
 #include <chrono>
 #include <future>
-#include <stdexcept>
 
-#include <boost/asio/deadline_timer.hpp>
-#include <boost/asio/io_service.hpp>
-#include <boost/date_time/posix_time/posix_time_duration.hpp>
-#include <boost/system/error_code.hpp>
+#include <asio/steady_timer.hpp>
+#include <asio/io_context.hpp>
 
 #include "Logger.hpp"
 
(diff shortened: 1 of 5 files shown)


client/client_pool/include/client/client_pool/client_pool_timer.hpp
```diff
@@ -15,49 +15,45 @@
...
-    timer_.expires_from_now(timeout);
-    auto handler = [this](const boost::system::error_code& error) {
-      if (error != boost::asio::error::operation_aborted) {
+    auto handler = [this](const asio::error_code& error) {
+      if (error != asio::error::operation_aborted) {
         on_timeout_(std::move(client_));
       }
     };
+
+    start_timer_ = std::chrono::steady_clock::now();
```
yuliasherman, line 55: Why did you move timer_ new settings to appear after a timeout handler?

client/client_pool/include/client/client_pool/client_pool_timer.hpp
```diff
@@ -15,49 +15,45 @@
...
-    stop_ = true;
-    io_service_.stop();
+    io_context_.stop();
     timer_thread_future_.wait();
   }
 
-  void set(const ClientT& client) {
-    if (timeout_.count() == 0) {
+  void start(const ClientT& client) {
+    if (timeout_.count() == 0 || not timer_thread_future_.valid() || io_context_.stopped()) {
```
yuliasherman, line 45: I would add a log here.

client/client_pool/include/client/client_pool/client_pool_timer.hpp
```diff
@@ -15,49 +15,45 @@
...
+  ~Timer() { stopTimerThread(); }
 
-  void stop() {
+  void stopTimerThread() {
     if (timeout_.count() == 0) {
       return;
     }
-    stop_ = true;
-    io_service_.stop();
+    io_context_.stop();
```
yuliasherman, line 40: Who will re-run io_context_ after client_pool will call stopTimerThread function?

robem: No one. The point of `stopTimerThread()` is to stop the timer thread and hence the io_context. If you want to restart the thread (and start the io_context) then you need to create a new `Timer` object. That's how the implementation was done and I didn't change it. I agree that this could be improved but I didn't want to change the implementation for no reason so I would leave this to a future task if there is a need.

client/client_pool/include/client/client_pool/client_pool_timer.hpp
robem: No strong reasoning, I wanted to group the code logically. Meaning, all timer and time value statements are clustered and separate from the handler and the client.

client/client_pool/include/client/client_pool/client_pool_timer.hpp
yuliasherman: That's not what I see in the code. batch_timer_ object is created only in ConcordClientPool ctor.

client/client_pool/include/client/client_pool/client_pool_timer.hpp
robem: The implementation didn't use logging before and I don't see it urgent to do so. However, I agree with you that a debug log would be helpful. To me this is a future refactoring task as well (together with re-thinking the lifecycle of the timer object, the constructor, dynamic timeout, ...) but I can add it if you feel it adds value now.

yuliasherman (approved):
LGTM. Thanks for taking care of it!

client/client_pool/include/client/client_pool/client_pool_timer.hpp
robem: `start()` to start the timeout and restart it for a new client. `cancel()` to cancel the current timeout. `stopTimerThread()` to stop the timer thread.

client/client_pool/include/client/client_pool/client_pool_timer.hpp
robem: I saw that you added logging a couple of hours ago. I will add the message.

PR #2538, opened 2022-04-05
Downgrade hash stream closed message to WARN from ERROR

shgandhi:
When TRC forces resubscription on one of the f+1 open streams
to not miss an update while receiving updates, it closes all
hash streams, this doesn't warrant an ERROR message on the TRS,
hence downgrading to WARN.


cjn (approved):

cjn:
LGTM!

PR #2382, opened 2022-02-23
Refactor concord-bft execution phase

yontyon:
This PR is about refactoring and enhancing the execution phase in concord-bft.
Until today , we had 5 modes of operations:
1. not accumulated and synchronized
2. accumulated and synchronized
3. not accumulates and async
4. accumulated and async
5. recovery
Each of these modes has its own conditions and code path - creating very complicated code paths.

In this PR we refactor this phase by introducing a base class that encapsulates all functionality.
Depending on the configuration and specific scenario, we can implement this base class in 6 different ways:
1. not accumulated and synchronized
2. accumulated and synchronized
3. not accumulates and async
4. accumulated and async
5. with a pre-defined set of requests for execution (for recovery)
6. with only sending responses to the client (for sending responses for already executed requests)

In addition, we fixed some issues and bugs during the refactoring.
1. no special requests execution
2. we can execute up to min(150, concurrency-level) sequence numbers in the execution thread, even by the leading quorum
3. on recovery, we are able to recover multiple PP messages (was a bug)

PR #2540, opened 2022-04-05
Lock reserved pages access (on inmemory datastore)

yontyon:
As a result of executing the "special requests" in a separate thread (see https://github.com/vmware/concord-bft/pull/2382) we may access the reserved pages from two different threads - the execution thread and the consensus thread.
To lock the read/write operation (as this is the only contention that can happen), we used mutex on the read/write level which is the inmemory datastore.
Note that in modern cpp, as this contention is very rare, using mutex is cheap as using spinlock with compare&swap.


toly-kournik (approved):
LGTM!

PR #2542, opened 2022-04-05
test_state_transfer_with_multiple_clients: fix an issue and enable back

glevkovich:
In some rare cases, after run_concurrent_ops, some replicas need more
time to send RFMI messages and get back into consensus.
We enable test back, and sleep 2 more seconds before calling
network_wait_for_checkpoint.


StefanPalashev (approved):
LGTM, thanks @glevkovich.

teoparvanov (changes requested):

PR #2544, opened 2022-04-06
Perf imprv for latest version CF migration tool

arc-vmware:
We saw that for a 700GB database the latest version CF migration tool, migration is not getting completed even after 12 hours. So in this PR, we are improving the performance by reading in parallel for a batch of blocks. And then do a serialized write. The order of writing is preserved. After the fix, the migration is completed within 2 hours. So this tool will result in a migration time of 2 hours for a DB whose size is around 700 GB

### DIFF
--- a/kvbc/src/migrations/block_merkle_latest_ver_cf_migration.cpp
+++ b/kvbc/src/migrations/block_merkle_latest_ver_cf_migration.cpp
@@ -13,6 +13,7 @@
 #include "migrations/block_merkle_latest_ver_cf_migration.h"
 
 #include "assertUtils.hpp"
+#include "thread_pool.hpp"
 #include "categorization/blockchain.h"
 #include "categorization/column_families.h"
 #include "categorization/db_categories.h"
@@ -39,6 +40,7 @@ using kvbc::categorization::detail::BLOCK_MERKLE_LATEST_KEY_VERSION_CF;
 using kvbc::categorization::detail::hash;
 using kvbc::categorization::Hash;
 using util::toChar;
+using util::ThreadPool;
 
 using ::rocksdb::Checkpoint;
 using ::rocksdb::ColumnFamilyDescriptor;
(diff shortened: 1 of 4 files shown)


kvbc/src/migrations/block_merkle_latest_ver_cf_migration.cpp
```diff
@@ -106,42 +111,55 @@ void BlockMerkleLatestVerCfMigration::clearExistingLatestVerCf() {
...
-        if (status.ok()) {
-          batch.put(BLOCK_MERKLE_LATEST_KEY_VERSION_CF, key_it->first, value);
-        } else {
-          // We expect that no pruning has occurred. Therefore, the latest version of the key cannot be missing.
-          throw std::runtime_error{"multiGet() failed, reason: " + status.ToString()};
+  ThreadPool tp(iteration_batch_size_);
+  for (auto block_id = INITIAL_GENESIS_BLOCK_ID; block_id <= blockchain.getLastReachableBlockId();
+       block_id += iteration_batch_size_) {
+    std::vector<std::future<void>> tasks;
+    std::vector<NativeWriteBatch> batches(iteration_batch_size_, NativeWriteBatch(db_));
```
cloudnoize, line 118: Given NativeWriteBatch takes the db_ as its state, may it suggest that NativeWriteBatch is not thread safe?

cloudnoize, line 118: Given NativeWriteBatch takes the db_ as its state, may it suggest that NativeWriteBatch is not thread safe?

cloudnoize (approved):
Beautiful algorithm @arc-vmware , great job!

kvbc/src/migrations/block_merkle_latest_ver_cf_migration.cpp
arc-vmware: It may be possible that NativeWriteBatch is not threadsafe. So here we are creating a separate batch for each thread and then using that batch to hold all the keys read from DB using multiget. Each thread is writing in its own allocated batch as the batches vector is preallocated.

arc-vmware: thanks, @cloudnoize, for the review.
It may be possible that NativeWriteBatch is not threadsafe. So here we are creating a separate batch for each thread and then using that batch to hold all the keys read from DB using multiget. Each thread is writing in its own allocated batch as the batches vector is preallocated.

PR #2548, opened 2022-04-06
Clientservice: Shutdown gRPC server with deadline

robem:
During runtime analysis, we wanted to shut down clientservice gracefully but
the server was hanging on the `Shutdown()` call. Let's not wait forever and
release resources properly. If the deadline is reached, the gRPC server will
cancel all ongoing requests.


shgandhi (approved):
LGTM! Thanks for the fix.

PR #2549, opened 2022-04-07
Perf imprv for latest version CF migration tool

arc-vmware:
We saw that for a 700GB database the latest version CF migration tool, migration is not getting completed even after 12 hours. So in this PR, we are improving the performance by reading in parallel for a batch of blocks. And then do a serialized write. The order of writing is preserved. After the fix, the migration is completed within 2 hours. So this tool will result in a migration time of 2 hours for a DB whose size is around 700 GB
This is cherry-picked into v0.14.x for https://github.com/vmware/concord-bft/pull/2544


teoparvanov (approved):
LGTM! It's awesome that you have a thorough unit test for this, it helps with both understanding the changes, and being confident about the correctness.

PR #2551, opened 2022-04-07
Decoupling kv blockchain and its usage

arc-vmware:
This PR adds an adapter layer that can be used to decouple the KV blockchain with the rest of the codebase. Later when the real implementation happens, changing the constructors and functions in the adapter layer can easily help us to choose the correct implementation. Further, this will also help us in doing A/B testing and comparing any improvements in newer versions of the blockchain.

### DIFF
--- a/kvbc/include/blockchain_type.hpp
+++ b/kvbc/include/blockchain_type.hpp
@@ -1,0 +1,26 @@
+// Concord
+//
+// Copyright (c) 2022-2023 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <variant>
+#include <optional>
+#include <memory>
+#include "categorization/kv_blockchain.h"
+
+namespace concord::kvbc {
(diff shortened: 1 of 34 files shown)


toly-kournik:
@arc-vmware , can you please provide a document with a more detailed motivation and design?

kvbc/include/blockchain_type.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+#pragma once
+
+#include <variant>
+#include <optional>
+#include <memory>
+#include "categorization/kv_blockchain.h"
+
+namespace concord::kvbc {
+enum BLOCKCHAIN_VERSION { CATEGORIZED_BLOCKCHAIN = 1, NATURAL_BLOCKCHAIN = 4 };
```
toly-kournik, line 22: Version or type?

arc-vmware: I think this is the version. For the same type of blockchain we will have different versions and they will coexist. 
A future use case can be said to implement categorized blockchain on top of Redis or aerospike as they may provide some feature that is better than rocksdb. Or implement a new type of blockchain on cockroach DB and mongo DB. One is a new-SQL DB and another is a no-SQL DB. In such scenarios for better performance, we might create a different KV Blockchain version for the same type and do A/B testing to finalize on default. 
There will be use case where these different versions co-exist, say vendor 1 likes to have blockchain in MongoDB and vendor 2 want it to be categorized as they definitely need Merkel tree. In such cases, we might provide the same type of KV blockchain with a new version.

arc-vmware:
We are moving to the strategy of creating a feature branch and then collaborating on that, so closing this PR.

PR #2552, opened 2022-04-07
Allow recovery from multiple executions at once

yontyon:
Fix some bugs with async execution

1. Fix the recovery method - while recovering we need to allow at most maxBatch size * checkpoint size blocks removal as this is the new number of queued executions
2. Don't start execution in case view is not active
3. Fix chaotic tests (not part of CI)


teoparvanov:
Hi Yoni, thanks for the fix. However I think the cleaner approach would be to revert this change and introduce it more gradually:
https://github.com/vmware/concord-bft/pull/2382

teoparvanov (changes requested):

PR #2553, opened 2022-04-07
Consider the latestCollectedSn in various of situations

yontyon:
There are some situations in which we need to consider the latestCollectedSn and not the latestExecutedSn.
This PR is for fixing this


robem (commented):
@yontyon do you mind elaborating on "There are some situations"? What situations did you encounter/consider?

teoparvanov (changes requested):

PR #2554, opened 2022-04-07
Fix issues of SimpleTest

harsht1737:
Issue 1: viewChangeCallback should be passed as an empty function wrapper rather than than nullptr as this function is called as no-op later in the code

Issue 2: aggregator has to be set for ReplicaBase at the time of Replica creation in SimpleTest Replica creation code


teoparvanov:
Hi @harsht1737 , thanks for those fixes! Does it now work as described in the readme?
https://github.com/vmware/concord-bft/blob/master/tests/simpleTest/README.md

CC: @arc-vmware

harsht1737:
Hi @teoparvanov, these are few of the fixes and some more of them are being analysed/fixed. Will drop a note to you when simpletest works as described in readme file.

arc-vmware (approved):
@harsht1737 , good work.
LGTM.

PR #2555, opened 2022-04-07
TRS: Wait for last_block_id + 1 live update in SubscribeToUpdates

shgandhi:
This commit fixes a bug in TRS, and adds minor refactoring.

### DIFF
--- a/client/concordclient/include/client/concordclient/thread_safe_queue.ipp
+++ b/client/concordclient/include/client/concordclient/thread_safe_queue.ipp
@@ -59,7 +59,7 @@ std::unique_ptr<T> BasicThreadSafeQueue<T>::pop() {
   if (release_consumers_) {
     return std::unique_ptr<T>(nullptr);
   }
-  ConcordAssert(queue_data_.size() > 0);
+  ConcordAssertGT(queue_data_.size(), 0);
   std::unique_ptr<T> ret = move(queue_data_.front());
   queue_data_.pop_front();
   return ret;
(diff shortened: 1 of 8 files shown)


robem (commented):
Thanks for the fix! It doesn't look like we have a test for this scenario. The existing `SubscribeToNextBlockNotInStorage` test only checks that we properly timeout. But the logic that you are changing should succeed eventually and hence we should have a test case for that.
Maybe rename `SubscribeToNextBlockNotInStorage` => `SubscribeToNextBlockNotInStorageTimeout`. And introduce a new `SubscribeToNextBlockNotInStorage` test? What do you think?

robem:
This change should be crossported to `v0.14.x`

shgandhi:
@robem thanks for the review. 
I was able to reproduce the issue with the timeout test, because live_update_queue had block 5, i.e., it wasn't empty, we continue waiting in the while loop until we see block 6 in the live_update_queue, which we never do, hence the timeout. Without this fix, TRS would have moved on to SyncAndSend and not timed out, because it found a live update in the live update queue.
However, I agree we should have a test for the logic that moves on to SyncAndSend, and not fail at the assert. I have added a test for that as you suggested. 
I will cherry-pick this change to v0.14.x when this gets merged in master.

robem (approved):
Thank you 🚀

rastogiasr:
Questions answered by Shruti offline:

1. Is this a regression or the defect has always been there: This defect always existed, it is an edge case that we hadn’t encountered before this test.
2. If not a regression, any idea on why it could be triggered now - something in the private build for memory testing: To add more context, if the last block ID in storage is last_block_id, and the client requests last_block_id + 1 , TRS allows this request to the next block even if it is not in storage. The failure is seen only when a new subscription is started with a request for last_block_id + 1 , and the live update queue has an update with the block id less than last_block_id + 1 . Both the conditions coming into play together is rare, i.e., why we hadn’t seen this error before. This doesn’t have anything to with the specific nature of the build and the test that was run, we hadn’t seen this code path being traversed before.
3. It looks like a concord crash, how likely this would be seen in a performance run: Since we have not yet seen this in any other LRTs, afaik, I believe it is rare, but still a possibility as seen in this private build.
4. Are there any side effects of the change - trying to gauge the risk level: I don’t think so.

PR #2557, opened 2022-04-08
TRS cherry-pick: Wait for last_block_id + 1 live update in SubscribeToUpdates

shgandhi:


samcramer (approved):
I checked against the change set that's being cherry-picked; looks good.

PR #2558, opened 2022-04-11
Change log level in getDirectorySize api

nkumar04:
Disk space monitoring thread tracks size of the database,
we use boost::filesystem::recursive_directory_iterator for this.
It enumerates all the files in folder and then calculate size.
If rocksdb delete a sst files in between for compaction then we may see an error
Changing log level of this event to WARN


itsmeakashgoyal (approved):
LGTM
Thanks Nikhil

PR #2559, opened 2022-04-11
Initial Commit to introduce replica for a example/demo

itsmeakashgoyal:
This PR focus on open source phase 1 implementation to introduce a replica for the example/demo.

A replica part for the open source phase 1 part is introduced in this commit.
These updates are responsible for assisting and guiding many concord-bft users throughout the world, such as those who want to learn more about concord-bft, its usage, and how it operates.
This will be a simple example/demo that we can provide to any user who wants to learn how concord-bft works and how they may use it.
So far, we're intending to implement this using an open source replica kvbc, similar to how we do it in Apollo tests, and then, after the setup is complete, we'll include other open source execution engines for this showcase, such as Ethereum, WASM, etc.


robem (changes requested):
You're introducing an example application with ~2k LoC in a single commit with a single PR message, and no commit message or readme. Please tell the reviewer what is going on in this change. What does the example show? How do you use the example? Is this the simplest example possible? Did you make any trade-offs?

arc-vmware (commented):
Thanks, @itsmeakashgoyal for adding the example folder. The code changes are very good. This is a code that will help and guide many many users of concord-bft worldwide. So it would be great if we can do something to help future users. So, before the review, I think adding the following will be very helpful :
1) Ammend the commit message with the details of the change. It would be great if you add a 1 liner for each class added by this PR. Please add any future plans to the example, like adding some sample execution engine.
2) Add a readme.md file which should contain the details of this example folder. You can refer to a good readme file like: https://github.com/sylabs/examples/blob/master/message-broker/rabbitmq/README.md
3) Add an install.md file to explain how we can run the example.
4) For every function please add the following comment (if we can add a doxy comment then that will help us in the future):
 /* @brief <one or two-line details of the function
 *  @param <explaination of each of the input parameter
 *  @return <one liner to explain the return>
 */
5) Every class should have comments with /* @brief */ annotation or other doxy annotation if you like.
This will be a long-running code so we won't try to do everything in one go, lets's keep doing things in steps with some minimum work as asked above.

example/README.md
```diff
@@ -0,0 +1,87 @@
+# Overview
+This directory contains a simple example/demo that we can provide to any user who wants to learn how concord-bft works and how they may use it.
```
cjn, line 2: Eventually a bit more elaboration can be added here

teoparvanov:
Hi @itsmeakashgoyal, it's great to see that an easy to use open source example is taking shape! I was wondering, do you plan to have a dedicated CI workflow which makes sure this demo doesn't get regressed by ongoing code changes in concord-bft? Of course, this can be done in a separate PR.

CC: @arc-vmware

example/scripts/test_example.sh
```diff
@@ -0,0 +1,41 @@
...
+
+#trap 'cleanup' SIGINT
+
+cleanup
+
+scriptdir=$(cd $(dirname $0); pwd -P)
+echo $scriptdir
+
+echo "Generating new keys..."
+$scriptdir/../../tools/GenerateConcordKeys -f 1 -n 4 -o replica_keys_
```
cjn, line 18: Are we considering number of replicas to be configurable in future?

arc-vmware:
> Hi @itsmeakashgoyal, it's great to see that an easy to use open source example is taking shape! I was wondering, do you plan to have a dedicated CI workflow which makes sure this demo doesn't get regressed by ongoing code changes in concord-bft? Of course, such an enhancement can be made in a separate PR.
> 
> CC: @arc-vmware

Thanks @teoparvanov,
Yes, we will add this to the CI path. And eventually, it will contain a real execution engine (we are thinking of adding WASM). After that, we will work on an application that will create the WASM contracts and send those. This will give a full-fledged end-to-end example.

teoparvanov:
> > Hi @itsmeakashgoyal, it's great to see that an easy to use open source example is taking shape! I was wondering, do you plan to have a dedicated CI workflow which makes sure this demo doesn't get regressed by ongoing code changes in concord-bft? Of course, such an enhancement can be made in a separate PR.
> > CC: @arc-vmware
> 
> Thanks @teoparvanov, Yes, we will add this to the CI path. And eventually, it will contain a real execution engine (we are thinking of adding WASM). After that, we will work on an application that will create the WASM contracts and send those. This will give a full-fledged end-to-end example.

Awesome! WASM sounds like the way to go even for our product long term IMO :)

example/README.md
itsmeakashgoyal: I have added some more details about the replica part. This will be an ongoing task and I will update the documentation when I will add client and other execution engines.

example/scripts/test_example.sh
itsmeakashgoyal: So as an initial commit, I have added an test_example.sh script to run client and replica but it will not be like this and will add a python script which will responsible to run this example/demo and user can easily configure the parameters using that script.

itsmeakashgoyal:
@teoparvanov, To exchange the kvbc messages, I used apollo's internal commands handler as an open source replica command handler. This will be the basic command handler for the time being, and when the client is ready, I will adapt it to reflect the simple use case.
Our primary goal is to introduce some real execution engines, such as WASM and others, so that they can be used widely.

example/opensourcecmf/opensource_replica_msgs.cmf
```diff
@@ -0,0 +1,62 @@
+# Definition of client-server interface for the opensource example test application.
```
teoparvanov, line 1: This is not a client-server interface. These are the messages/commands for the open source state machine example (effectively, a byzantine fault-tolerant K/V store).

example/opensourcecmf/opensource_replica_msgs.cmf
```diff
@@ -0,0 +1,62 @@
...
+#
+# Note OSRKVBC effectively implements a versioned key-value store supporting
+# only keys and values that are byte strings.
+
+Msg OSRKVBCReadRequest 3 {
```
teoparvanov, line 6: The "OSRKVBC" prefix is a bit cryptic IMO... Assuming "OSR" means open source, I think we can drop it as it is obvious this is an open source project :) Can we simplify the message names further, to something like `KVReadRequest`, etc.?

CC: @arc-vmware

example/opensourcecmf/opensource_replica_msgs.cmf
```diff
@@ -0,0 +1,62 @@
...
+
+    # Note an OSRKVBCReadRequest with read version 0 indicates that the latest
+    # version should be read.
+    uint64 read_version
+
+    list bytes keys
+}
+
+Msg OSRKVBCWriteRequest 4 {
```
teoparvanov, line 15: KVWriteRequest?

example/opensourcecmf/opensource_replica_msgs.cmf
```diff
@@ -0,0 +1,62 @@
...
+    } request
+}
+
+Msg OSRKVBCReply 2 {
+    oneof {
+        OSRKVBCReadReply
+        OSRKVBCWriteReply
+        OSRKVBCGetLastBlockReply
+    } reply
+}
```
teoparvanov, line 62: same here - let's consider replacing the cumbersome `OSRKVBC` prefix with simply `KV`, or `KVBC` - or something better?

example/replica/include/OSReplicaCommandHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+#include <thread>
+#include "opensource_replica_msgs.cmf.hpp"
+#include "SharedTypes.hpp"
+#include "categorization/db_categories.h"
+#include "categorization/kv_blockchain.h"
+
+static const std::string VERSIONED_KV_CAT_ID{concord::kvbc::categorization::kExecutionPrivateCategory};
+static const std::string BLOCK_MERKLE_CAT_ID{concord::kvbc::categorization::kExecutionProvableCategory};
+
+class OSReplicaCommandHandler : public concord::kvbc::ICommandsHandler {
```
teoparvanov, line 34: How about `KVReplicaComandsHandler`, or even simpler - `KVCommandsHandler`?

example/README.md
```diff
@@ -0,0 +1,87 @@
+# Overview
+This directory contains a simple open source example/demo that we can provide to any user who wants to learn how concord-bft works and how they may use it. 
+
+This example is responsible for assisting and guiding many concord-bft users throughout the world, including those who want to understand more about concord-bft, its usage, and how it works. This open source example/demo will provide end to end functionality of Concord byzantine fault tolerant state machine replication library about its usage and how we can use this as a library to implement a highly scalable and energy-efficient distributed trust infrastructure for consensus and smart contract execution.
+So far, we're intending to implement this using an open source replica kvbc, similar to how we do it in Apollo tests, and then, after the setup is complete, we'll include other open source execution engines for this showcase, such as Ethereum, WASM, etc.
```
teoparvanov, line 5: This sounds like a historical reference, rather than a description of the current project. I think you should rather emphasize on the fact that:

- this example is a byzantine fault tolerant K/V-store implementation, based on Concord-BFT consensus
- it is a fully wired system, that is self-service and ready for demoing
- other important topics to highlight @itsmeakashgoyal and @arc-vmware ?

example/opensourcecmf/opensource_replica_msgs.cmf
itsmeakashgoyal: @teoparvanov we can consider prefix KV for all message commands for this. Thanks for the suggestion .

toly-kournik:
How the proposed replica differ from a simpleKVBC?
We could improve/enhance the existing replica instead of duplicating so many codes of line, which by themselves could be refactored/improved.

teoparvanov (changes requested):

itsmeakashgoyal:
Closing this PR as a separate PR #2686 is opened for both client and replica.

PR #2561, opened 2022-04-11
Fix a race condition between PreProcessor and consensus threads

yuliasherman:
**Problem Overview**
PreProcessor uses two ClientsManager functions - hasReply and isClientRequestInProcess to determine whether to process or reject client request. As ClientsManager is not thread-safe, we have a race between the consensus thread that modifies required information and the PreProcessor functions that query for.
**Testing Done**
Ran a load of 700 PTS on a reserved system. Gradual performance degradation to 640 TPS has been observed after 5 days of the run; not different from a regular run.
The second commit is a cherry-pick of https://github.com/vmware/concord-bft/pull/2561/commits/c7f351e9b21351c31387903c3f787873ed0544d1

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -24,6 +24,158 @@
 using namespace std::chrono;
 using namespace concord::serialize;
 namespace bftEngine::impl {
+
+/*************************** Class ClientsManager::RequestsInfo ***************************/
+
+void ClientsManager::RequestsInfo::emplaceSafe(NodeIdType clientId, ReqId reqSeqNum, const std::string& cid) {
+  const lock_guard<mutex> lock(requestsMapMutex_);
+  if (requestsMap_.find(reqSeqNum) != requestsMap_.end()) {
+    LOG_WARN(CL_MNGR, "The request already exists - skip adding" << KVLOG(clientId, reqSeqNum));
+    return;
+  }
+  requestsMap_.emplace(reqSeqNum, RequestInfo{getMonotonicTime(), cid});
+  LOG_DEBUG(CL_MNGR, "Added request" << KVLOG(clientId, reqSeqNum, requestsMap_.size()));
+}
+
+bool ClientsManager::RequestsInfo::findSafe(ReqId reqSeqNum) const {
+  const lock_guard<mutex> lock(requestsMapMutex_);
+  return (requestsMap_.find(reqSeqNum) != requestsMap_.end());
+}
+
+bool ClientsManager::RequestsInfo::removeRequestsOutOfBatchBoundsSafe(NodeIdType clientId, ReqId reqSequenceNum) {
+  ReqId maxReqId{0};
+  const lock_guard<mutex> lock(requestsMapMutex_);
+  if (requestsMap_.find(reqSequenceNum) != requestsMap_.end()) return false;
+  for (const auto& reqInfo : requestsMap_)
+    if (reqInfo.first > maxReqId) maxReqId = reqInfo.first;
+
+  if (requestsMap_.size() == maxNumOfRequestsInBatch && maxReqId > reqSequenceNum) {
+    // If we don't have room for the sequence number, and we see that the highest sequence number is greater
+    // than the given one, it means that the highest sequence number is out of the boundaries and can be safely removed
+    requestsMap_.erase(maxReqId);
+    return true;
+  }
+  return false;
+}
+
+void ClientsManager::RequestsInfo::removeOldPendingReqsSafe(NodeIdType clientId, ReqId reqSeqNum) {
+  const lock_guard<mutex> lock(requestsMapMutex_);
+  for (auto it = requestsMap_.begin(); it != requestsMap_.end();) {
+    if (it->first <= reqSeqNum) {
+      it = requestsMap_.erase(it);
+      LOG_INFO(CL_MNGR, "Remove old pending request" << KVLOG(clientId, reqSeqNum));
+    } else
+      it++;
+  }
+}
+
+void ClientsManager::RequestsInfo::removePendingForExecutionRequestSafe(NodeIdType clientId, ReqId reqSeqNum) {
+  const lock_guard<mutex> lock(requestsMapMutex_);
+  const auto& reqIt = requestsMap_.find(reqSeqNum);
+  if (reqIt != requestsMap_.end()) {
+    requestsMap_.erase(reqIt);
+    LOG_DEBUG(CL_MNGR, "Removed request" << KVLOG(clientId, reqSeqNum, requestsMap_.size()));
+  }
+}
+
+void ClientsManager::RequestsInfo::clearSafe() {
+  const std::lock_guard<std::mutex> lock(requestsMapMutex_);
+  requestsMap_.clear();
+}
+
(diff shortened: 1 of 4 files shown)


teoparvanov:
As discussed on Slack, I believe this fix should go in master as a first step. Merging in 1.6 should be considered separately, right @eyalrund ?

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -218,9 +219,45 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
...
+    void clearSafe();
+    void removeOldPendingReqsSafe(NodeIdType clientId, ReqId reqSeqNum);
+    void removePendingForExecutionRequestSafe(NodeIdType clientId, ReqId reqSeqNum);
+
+    size_t size() const { return requestsMap_.size(); }
+    bool find(ReqId reqSeqNum) const;
+    bool isPending(ReqId reqSeqNum) const;
+    void markRequestAsCommitted(NodeIdType clientId, ReqId reqSeqNum);
+    void infoOfEarliestPendingRequest(Time& earliestTime, RequestInfo& earliestPendingReqInfo) const;
+    void logAllPendingRequestsExceedingThreshold(const int64_t threshold,
```
HristoStaykov, line 236: If I understand correctly we use the non locking functions when we designate a given entry form **clientsInfo_** to a thread and we don't access it from others?

yuliasherman: Not sure I understand the question. Doesn't the "Problem overview" part answer your question?

HristoStaykov: Yes it does, but because the actual usage is something I am not in-depth familiar with, wanted to make sure. If my general understanding is correct we need to have a mechanism that blocks other threads from modifying the map **requestsMap_** while we are operating on it with the non-sync alternatives of the methods?

yuliasherman: Correct

HristoStaykov (approved):
LGTM, thanks @yuliasherman ! All the extracted functions look good and equal to the previous code but with the added sync where necessary.

yuliasherman:
Will be merged into the master only.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -24,6 +24,158 @@
...
+  requestsMap_.emplace(reqSeqNum, RequestInfo{getMonotonicTime(), cid});
+  LOG_DEBUG(CL_MNGR, "Added request" << KVLOG(clientId, reqSeqNum, requestsMap_.size()));
+}
+
+bool ClientsManager::RequestsInfo::findSafe(ReqId reqSeqNum) {
+  const lock_guard<mutex> lock(requestsMapMutex_);
+  return (requestsMap_.find(reqSeqNum) != requestsMap_.end());
+}
+
+bool ClientsManager::RequestsInfo::removeRequestsOutOfBatchBoundsSafe(NodeIdType clientId, ReqId reqSequenceNum) {
```
ukg19, line 45: Thanks Yulia for creating safe methods for the thread-unsafe methods. Any reasons for not calling the unsafe methods by putting the lock in safe method. That would avoid code duplication?

ukg19: Oh, I notice you have removed unsafe methods. This is fine. Thanks.

ukg19 (approved):
LGTM !!

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -24,6 +24,158 @@
...
+  if (requestsMap_.find(reqSeqNum) != requestsMap_.end()) {
+    LOG_WARN(CL_MNGR, "The request already exists - skip adding" << KVLOG(clientId, reqSeqNum));
+    return;
+  }
+  requestsMap_.emplace(reqSeqNum, RequestInfo{getMonotonicTime(), cid});
+  LOG_DEBUG(CL_MNGR, "Added request" << KVLOG(clientId, reqSeqNum, requestsMap_.size()));
+}
+
+bool ClientsManager::RequestsInfo::findSafe(ReqId reqSeqNum) {
+  const lock_guard<mutex> lock(requestsMapMutex_);
```
arc-vmware, line 41: Taking the write-only mutex is a good idea. But here we can see that we have separate read and write operations on requestMap_. So a read and write lock will be better for better performance:
If we use the read and write lock as below:
using WriteLock = std::unique_lock<std::shared_mutex>;
using ReadLock = std::shared_lock<std::shared_mutex>;

Then it will keep the throughput better.
So in emplaceSafe() we should use WriteLock(requestsMapMutex_);
And in findSafe() we should use ReadLock(requestsMapMutex_);
This way multiple findSafes can happen together.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -24,6 +24,158 @@
...
+  return (requestsMap_.find(reqSeqNum) != requestsMap_.end());
+}
+
+bool ClientsManager::RequestsInfo::removeRequestsOutOfBatchBoundsSafe(NodeIdType clientId, ReqId reqSequenceNum) {
+  ReqId maxReqId{0};
+  const lock_guard<mutex> lock(requestsMapMutex_);
+  if (requestsMap_.find(reqSequenceNum) != requestsMap_.end()) return false;
```
arc-vmware, line 48: If we use read / write lock, then in this function we will take read lock till Line 53 and take write lock only for requestsMap_.erase(maxReqId);.

arc-vmware: But in other safe functions where we are doing erase in a loop, we should take a write lock only.

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -218,9 +219,45 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
...
+  class RepliesInfo {
+   public:
+    void deleteOldestReplyIfNeededSafe(NodeIdType clientId, uint16_t maxNumOfReqsPerClient);
+    bool insertOrAssignSafe(ReqId reqSeqNum, Time time);
+    bool findSafe(ReqId reqSeqNum);
+
+    bool find(ReqId reqSeqNum) const;
+
+   public:
+    std::mutex repliesMapMutex_;
```
arc-vmware, line 254: nit: It's better to declare mutex as mutable so that functions that should be const remain const. I think the hasReply is non-const now due to this mutex.
mutable std::mutex repliesMapMutex_;

arc-vmware (commented):
thanks, @yuliasherman, I have very few comments. The code looks great.

bftengine/src/bftengine/ClientsManager.cpp
yuliasherman: Read/write lock approach is efficient only in case there are significantly more reads than writes. In case the difference is not big, it is more expensive than a mutex.

bftengine/src/bftengine/ClientsManager.hpp
yuliasherman: Good point.

arc-vmware (approved):
LGTM

PR #2562, opened 2022-04-11
Changes to support separate error for out of time bound requests

cjn:
Added changes to support separate error for out of time bound requests to be send back to the application as gRPC error - ABORT.

### DIFF
--- a/client/clientservice/src/request_service.cpp
+++ b/client/clientservice/src/request_service.cpp
@@ -99,39 +99,55 @@ void RequestServiceCallData::sendToConcordClient() {
       switch (std::get<uint32_t>(send_result)) {
         case (static_cast<uint32_t>(bftEngine::OperationResult::INVALID_REQUEST)):
           LOG_INFO(logger, "Request failed with INVALID_ARGUMENT error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::INVALID_ARGUMENT, "Invalid argument");
+          status = grpc::Status(
+              grpc::StatusCode::INVALID_ARGUMENT,
+              ConcordErrorMessage_Name(vmware::concord::client::request::v1::CONCORD_ERROR_INVALID_REQUEST));
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::NOT_READY)):
           LOG_INFO(logger, "Request failed with NOT_READY error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "No clients connected to the replicas");
+          status =
+              grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                           ConcordErrorMessage_Name(vmware::concord::client::request::v1::CONCORD_ERROR_NOT_READY));
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::TIMEOUT)):
           LOG_INFO(logger, "Request failed with TIMEOUT error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::DEADLINE_EXCEEDED, "Timeout");
+          status = grpc::Status(grpc::StatusCode::DEADLINE_EXCEEDED,
+                                ConcordErrorMessage_Name(vmware::concord::client::request::v1::CONCORD_ERROR_TIMEOUT));
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::EXEC_DATA_TOO_LARGE)):
           LOG_INFO(logger, "Request failed with EXEC_DATA_TOO_LARGE error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::INTERNAL, "Execution data too large");
+          status = grpc::Status(
+              grpc::StatusCode::INTERNAL,
+              ConcordErrorMessage_Name(vmware::concord::client::request::v1::CONCORD_ERROR_EXEC_DATA_TOO_LARGE));
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::EXEC_DATA_EMPTY)):
           LOG_INFO(logger, "Request failed with EXEC_DATA_EMPTY error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::INTERNAL, "Execution data is empty");
+          status = grpc::Status(
+              grpc::StatusCode::INTERNAL,
+              ConcordErrorMessage_Name(vmware::concord::client::request::v1::CONCORD_ERROR_EXEC_DATA_EMPTY));
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::CONFLICT_DETECTED)):
           LOG_INFO(logger, "Request failed with CONFLICT_DETECTED error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted due to conflict detected");
+          status = grpc::Status(
+              grpc::StatusCode::ABORTED,
+              ConcordErrorMessage_Name(vmware::concord::client::request::v1::CONCORD_ERROR_CONFLICT_DETECTED));
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::OVERLOADED)):
           LOG_INFO(logger, "Request failed with OVERLOADED error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::RESOURCE_EXHAUSTED, "All clients occupied");
+          status =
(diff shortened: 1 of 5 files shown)


client/clientservice/src/request_service.cpp
```diff
@@ -119,12 +119,16 @@ void RequestServiceCallData::sendToConcordClient() {
...
-          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted");
+          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted due to conflict detected");
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::OVERLOADED)):
           LOG_INFO(logger, "Request failed with OVERLOADED error for cid=" << req_config.correlation_id);
           status = grpc::Status(grpc::StatusCode::RESOURCE_EXHAUSTED, "All clients occupied");
           break;
+        case (static_cast<uint32_t>(bftEngine::OperationResult::OUT_OF_TIME_BOUND_ERROR)):
+          LOG_INFO(logger, "Request failed with OUT_OF_TIME_BOUND_ERROR error for cid=" << req_config.correlation_id);
+          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted due to out of time bound error");
```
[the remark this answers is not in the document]
yuliasherman: I've asked Celine this question offline and proposed to sync with you regarding it.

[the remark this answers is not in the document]
cjn: Timeout error is different and it is was request time out.
OUT_OF_TIME_BOUND_ERROR is an error from commandhandler like CONFLICT_DETECTED. Previously explicit write sets were provided by the execution engine if the recording time is not within time bounds. Now this is a new change in behaviour where errors have to be send back in such cases which will propagate as ABORT to the application so that they can choose to act based on this error.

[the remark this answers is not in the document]
robem: To my knowledge, Concord does not check or enforce a time span in which the result of a request can be persisted. This is execution engine specific. Therefore, I don't see a reason to introduce a new error type which doesn't relate to Concord's behavior. If the execution engine doesn't want to persist a result then it can signal this via an application specific response.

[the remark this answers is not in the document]
cjn: Post discussion, there is requirement to distinguish the ABORTED error due to conflict detected and rejection during processing. 
As discussed, I have updated changes to make it generic rejection error instead of OUT_OF_TIME_BOUND_ERROR

client/clientservice/src/request_service.cpp
```diff
@@ -119,12 +119,16 @@ void RequestServiceCallData::sendToConcordClient() {
...
-          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted");
+          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted due to conflict detected");
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::OVERLOADED)):
           LOG_INFO(logger, "Request failed with OVERLOADED error for cid=" << req_config.correlation_id);
           status = grpc::Status(grpc::StatusCode::RESOURCE_EXHAUSTED, "All clients occupied");
           break;
+        case (static_cast<uint32_t>(bftEngine::OperationResult::EXEC_ENGINE_REJECT_ERROR)):
+          LOG_INFO(logger, "Request failed with EXEC_ENGINE_REJECT_ERROR error for cid=" << req_config.correlation_id);
+          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted due to request rejected error");
```
robem, line 130: Rejected by the execution engine.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -112,6 +112,10 @@ static void getResponseSetStatus(concord::client::concordclient::SendResult&& se
         LOG_INFO(logger, "Request failed with OVERLOADED error for cid=" << correlation_id);
         return_status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "All clients occupied");
         break;
+      case (static_cast<uint32_t>(bftEngine::OperationResult::EXEC_ENGINE_REJECT_ERROR)):
+        LOG_INFO(logger, "Request failed with EXEC_ENGINE_REJECT_ERROR error for cid=" << correlation_id);
+        return_status = grpc::Status(grpc::StatusCode::UNAVAILABLE, "Aborted");
```
robem, line 117: Why do we return UNAVAILABLE here? @arc-vmware Does this make sense to you?
The [documentation](https://github.com/vmware/concord-bft/blob/master/client/proto/state_snapshot/v1/state_snapshot.proto) says that Concord is not ready to process a request in this case.

client/clientservice/src/request_service.cpp
```diff
@@ -119,12 +119,16 @@ void RequestServiceCallData::sendToConcordClient() {
           break;
         case (static_cast<uint32_t>(bftEngine::OperationResult::CONFLICT_DETECTED)):
           LOG_INFO(logger, "Request failed with CONFLICT_DETECTED error for cid=" << req_config.correlation_id);
-          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted");
+          status = grpc::Status(grpc::StatusCode::ABORTED, "Aborted due to conflict detected");
```
robem, line 122: Please adapt the [API documentation](https://github.com/vmware/concord-bft/blob/master/client/proto/request/v1/request.proto#L31) to reflect this change.

robem (commented):
I still don't understand why the execution engine cannot return an error message on its own layer which is then read by the application and delivered via Concord and clientservice.
For example, if you shop online and you enter the wrong credit card credentials then your ISP won't reject the request but the shop (the web application) will.
Similar, from Concord's perspective, why would we want to get into the business of application error codes?

client/clientservice/src/state_snapshot_service.cpp
arc-vmware: I think making it UNAVAILABLE doesn't make sense, since CS is available here. So we should change this to:
return_status = grpc::Status(grpc::StatusCode::UNKNOWN, "Execution rejected this request");
For the UNAVAILABLE case, the caller might retry after some time with the same req. But in this case, the request itself is a problem, so instead of retry, they should crash. So I agree with @robem that this should be changed.

cjn: @robem - Sure. I can change the error. But for CONFLICT_DETECTED also we return UNAVAILABLE here. So wanted to confirm if that also needs to be UNKNOWN?

robem: We should be consistent and return `ABORTED` as well, IMHO. If the same error on two different services returns two different errors then this is somewhat confusing.

arc-vmware: @cjn why EXEC_ENGINE_REJECT_ERROR and CONFLICT_DETECTED are related?

arc-vmware: We can return UNKNOWN and UNAVAILABLE. 
UNAVAILABLE should be returned in cases where retrying the same message can succeed. UNKNOWN means the client of CS should crash and seek help from SRE.

client/clientservice/src/request_service.cpp
cjn: Updated with enums

client/clientservice/src/state_snapshot_service.cpp
cjn: As discussed changed to UNKNOWN

client/clientservice/src/request_service.cpp
cjn: Updated comments for error message strings

cjn:
> I still don't understand why the execution engine cannot return an error message on its own layer which is then read by the application and delivered via Concord and clientservice. For example, if you shop online and you enter the wrong credit card credentials then your ISP won't reject the request but the shop (the web application) will. Similar, from Concord's perspective, why would we want to get into the business of application error codes?

> I still don't understand why the execution engine cannot return an error message on its own layer which is then read by the application and delivered via Concord and clientservice. For example, if you shop online and you enter the wrong credit card credentials then your ISP won't reject the request but the shop (the web application) will. Similar, from Concord's perspective, why would we want to get into the business of application error codes?

Thanks @robem and @arc-vmware for valuable feedback. Currently I have added generic error to mitigate immediate integration issue which is generic and can be used by any execution engine in future. However I will continue to investigate on possibility of return such errors from application layer in a separate task.

teoparvanov (approved):
LGTM, thanks Celine! I think this approach works for now. Before making it more generic, we should also consider the requirements for future EE integrations. This can indeed be a separate investigation, tracked as tech debt for the time being.

PR #2566, opened 2022-04-12
BCStateTran: memset BlockIOContext::blockData to claim ownership

glevkovich:
When allocating only, memory is not marked as used, and other processed
might capture it later on. Some time can pass until replica may enter
into St as a source or as a destination. We would like to make sure the
memory is owned on startup, since the amount of memory acquired here
is very large. Failing in allocation on startup is better then failing
later on.
Solution: memset all memory in BlockIOContext::blockData immediately
after allocation in order marked it by OS as used and reserved for ST.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.hpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.hpp
@@ -464,7 +464,9 @@ class BCStateTran : public IStateTransfer {
     static size_t sizeOfBlockData;
     BlockIOContext() {
       ConcordAssert(sizeOfBlockData != 0);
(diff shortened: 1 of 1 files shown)


robem (commented):
@glevkovich "When allocating only, memory is not marked as used, and other processed
might capture it later on." Do you have a reference? I might learn something here. If memory is allocated by a process then how could anyone obtain the exact same memory via a subsequent allocation? Maybe I don't understand what you mean by "capture".

robem:
Thanks @glevkovich, overcommitting rings a bell.

teoparvanov (changes requested):

PR #2570, opened 2022-04-12
Revert new execution phase

yontyon:
* **Problem Overview**  
Reverting the new execution phase refactoring and enhancement.
The system is not stable enough so we try to revert some of the new changes (even if they were not proved to be related) and first stabilize the system. Once the system is stabilized again we will re-enter these changes (see https://github.com/vmware/concord-bft/pull/2382)
* **Testing Done**  
  Regular GHA tests


teoparvanov (approved):
Thanks Yoni, I think this is the right approach.

teoparvanov (approved):
Thanks Yoni - can you also trigger a bump MR, to propagate these changes to Concord?

PR #2571, opened 2022-04-13
Fix a race condition between PreProcessor and consensus threads

yuliasherman:
* **Problem Overview**  
  PreProcessor uses two ClientsManager functions - hasReply and isClientRequestInProcess to determine whether to 
  process or reject client request. As ClientsManager is not thread-safe, we have a race between the consensus thread that 
  modifies required information and the PreProcessor functions that query for.
* **Testing Done**  
  Ran a load of 700 PTS on a reserved system for 1.6. After 4 days of the run, the performance was degraded to 650 TPS; no 
  difference from a regular run.

### DIFF
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -24,6 +24,158 @@
 using namespace std::chrono;
 using namespace concord::serialize;
 namespace bftEngine::impl {
+
+/*************************** Class ClientsManager::RequestsInfo ***************************/
+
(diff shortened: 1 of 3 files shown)


HristoStaykov (approved):
LGTM, thanks @yuliasherman !

teoparvanov:
Note that @HristoStaykov has made a thorough review in the PR which was targeted to the `v0.14.x` branch:
https://github.com/vmware/concord-bft/pull/2561

For this PR he simply made sure the code changes are identical.

PR #2572, opened 2022-04-13
Import concord-ctl directly

WildFireFlum:
When the diagnostics tests were added, the concord-ctl utility was imported indirectly by extracting its content to
a file compatible to python's naming conventions. 
The concord-ctl script is expected to be an independent copiable file, and this PR makes it so again.


toly-kournik (approved):
LGTM!

PR #2573, opened 2022-04-13
Fix faulty replica model

HristoStaykov:
* **Problem Overview**  
 In `distributed_system.s.dfy` we had omitted the Faulty Replica step. We have added it accordingly and taken it in consideration in the proof.
* **Testing Done**  
Dafny verifies after the additions to `distributed_system.s.dfy` and `proof.dfy`

### DIFF
--- a/docs/sbft-formal-model/proof.dfy
+++ b/docs/sbft-formal-model/proof.dfy
@@ -27,7 +27,7 @@ module Proof {
   import Replica
   import opened DistributedSystem
   import opened SafetySpec
-  import Library
+  import opened Library
 
   predicate IsHonestReplica(c:Constants, hostId:HostId) 
   {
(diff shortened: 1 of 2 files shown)


docs/sbft-formal-model/proof.dfy
```diff
@@ -434,24 +434,24 @@ module Proof {
...
+  //                                       :: QuorumOfPreparesInNetwork(c, v', view, seqID, clientOp))
+  // {
+  //   forall view, seqID, clientOp | QuorumOfPreparesInNetwork(c, v, view, seqID, clientOp)
+  //                                 ensures QuorumOfPreparesInNetwork(c, v', view, seqID, clientOp)
+  //   {
+  //     var senders := Messages.sendersOf(sentPreparesForSeqID(c, v, view, seqID, clientOp));
+  //     var senders' := Messages.sendersOf(sentPreparesForSeqID(c, v', view, seqID, clientOp));
+  //     Library.SubsetCardinality(senders, senders');
+  //   }
+  // }
```
teoparvanov, line 454: why not delete this lemma, instead of commenting out?

HristoStaykov: It was very useful during the initial phase of the proof and we left it for future debugging purposes. I will fix it to take in consideration the FaultyReplica's steps.

PR #2576, opened 2022-04-14
calculate timestamp even when there are no special requests

yontyon:
* **Problem Overview**  
  We calculate the timestamp only in case we have special requests in the prePrepare message. As time service requests have been removed, it means that in a regular prePrepare message there are no special requests. Hence, the timestamp won't be calcualted
* **Testing Done**  
  Regular GHA tests


nkgeorgiev (approved):

teoparvanov (approved):
LGTM, thanks Yoni!

PR #2579, opened 2022-04-14
EventGroups: Add convenient getValueFromTagTable API

robem:
This change overloads the `getValueFromTagTable` function so that the caller
doesn't need to care about the separator or the string concatenation. As a
result, the code becomes more readable for the caller.
This is a follow-up from a discussion on #2354 - thanks @dartdart26!

### DIFF
--- a/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
+++ b/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
@@ -58,8 +58,7 @@ std::optional<kvbc::categorization::EventGroup> KvbAppFilter::getNewestPublicEve
     // No public event group ID.
     return std::nullopt;
   }
-  const auto [global_eg_id, _] = getValueFromTagTable(kPublicEgId + kTagTableKeySeparator +
-                                                      concordUtils::toBigEndianStringBuffer(newest_pub_eg_id));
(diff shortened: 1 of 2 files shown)


kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -309,6 +308,10 @@ uint64_t KvbAppFilter::getValueFromLatestTable(const std::string &key) const {
   return concordUtils::fromBigEndianBuffer<uint64_t>(val->data.data());
 }
 
+TagTableValue KvbAppFilter::getValueFromTagTable(const std::string &tag, uint64_t pvt_eg_id) const {
+  return getValueFromTagTable(tag + kTagTableKeySeparator + concordUtils::toBigEndianStringBuffer(pvt_eg_id));
```
shgandhi, line 312: Why don't we update the function currently accepting `key` to accept `tag` and `pvt_eg_id`, and concatenate the parameters inside it? i.e., I am trying to understand why do we need to overload the function.

shgandhi (approved):
Thanks for the refactoring @robem! LGTM. 
I just have one comment.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: You're right, there is actually no use-case for the one-parameter function. I will change it. Thanks for your review!

PR #2581, opened 2022-04-18
Prohibit request cancellation when a primary replica has not completed the pre-processing

yuliasherman:
**Problem Overview**
In case the primary replica is still busy with a request pre-processing, but other replicas have already responded with the replies, which are not equal, we cancel the request. This operation also releases a memory allocated from the memory pool. Then when this request pre-execution completes and we try to write the result into the reply buffer, the replica crashes, as the buffer is not available anymore.
**Testing Done**
This behavior is hard to reproduce. I've run the system under a load of 700 TPS for two hours to verify the fix does not harm.


ukg19 (approved):
LGTM !!
Thanks @yuliasherman

PR #2582, opened 2022-04-18
EventService: Conditional sleep on trc queue

robem:
Instead of "just" sleeping, the event service should process updates as soon as
they are available. Recently, we introduced `popTill` as part of introducing
the snapshot client. With this change, we wake up as soon as an update is
available but also when the timeout expires. Therefore, we can check regularly
whether the subscription was cancelled but also don't miss out on processing
new updates immediately.


shgandhi:

shgandhi (approved):

PR #2585, opened 2022-04-20
EventService: Move events instead of copying

robem:
There is no need to copy events to the protobuf data structure given that its
API offers to move elements into the repeated field 'events'.


shgandhi (approved):
Thanks for the change. LGTM!

PR #2586, opened 2022-04-21
Add more metrics to TRC and EventService for Performance Analysis

shgandhi:
This commit adds more metrics to TRC, and introduces metrics to EventService. Please see individual commits for more detail.

### DIFF
--- a/client/clientservice/include/client/clientservice/event_service.hpp
+++ b/client/clientservice/include/client/clientservice/event_service.hpp
@@ -18,10 +18,42 @@
 
 namespace concord::client::clientservice {
 
+// EventService metrics
+struct EventServiceMetrics {
+  EventServiceMetrics()
+      : metrics_component_{"EventService", std::make_shared<concordMetrics::Aggregator>()},
+        num_writes_per_second{metrics_component_.RegisterGauge("num_writes_per_second", 0)},
+        update_processing_dur{metrics_component_.RegisterGauge("update_processing_dur", 0)},
+        write_dur{metrics_component_.RegisterGauge("write_dur", 0)} {
+    metrics_component_.Register();
+  }
+
+  void setAggregator(const std::shared_ptr<concordMetrics::Aggregator>& aggregator) {
+    metrics_component_.SetAggregator(aggregator);
+  }
+
+  void updateAggregator() { metrics_component_.UpdateAggregator(); }
+
+ private:
+  concordMetrics::Component metrics_component_;
+
+ public:
+  // number of updates written to the gRPC stream per second
+  concordMetrics::GaugeHandle num_writes_per_second;
+  // time taken to process an update (time between when requested is received by the event service until it is written
+  // to the stream)
+  concordMetrics::GaugeHandle update_processing_dur;
+  // time taken to write an update to the gRPC stream
+  concordMetrics::GaugeHandle write_dur;
+};
+
 class EventServiceImpl final : public vmware::concord::client::event::v1::EventService::Service {
  public:
-  EventServiceImpl(std::shared_ptr<concord::client::concordclient::ConcordClient> client)
-      : logger_(logging::getLogger("concord.client.clientservice.event")), client_(client){};
+  EventServiceImpl(std::shared_ptr<concord::client::concordclient::ConcordClient> client,
+                   std::shared_ptr<concordMetrics::Aggregator> aggregator)
+      : logger_(logging::getLogger("concord.client.clientservice.event")), client_(client), metrics_() {
+    metrics_.setAggregator(aggregator);
+  };
   grpc::Status Subscribe(grpc::ServerContext* context,
                          const vmware::concord::client::event::v1::SubscribeRequest* request,
                          grpc::ServerWriter<vmware::concord::client::event::v1::SubscribeResponse>* stream) override;
@@ -29,6 +61,7 @@ class EventServiceImpl final : public vmware::concord::client::event::v1::EventS
  private:
   logging::Logger logger_;
   std::shared_ptr<concord::client::concordclient::ConcordClient> client_;
+  EventServiceMetrics metrics_;
 };
 
 }  // namespace concord::client::clientservice
--- a/client/clientservice/include/client/clientservice/event_service.hpp
+++ b/client/clientservice/include/client/clientservice/event_service.hpp
@@ -22,7 +22,7 @@ namespace concord::client::clientservice {
 struct EventServiceMetrics {
   EventServiceMetrics()
       : metrics_component_{"EventService", std::make_shared<concordMetrics::Aggregator>()},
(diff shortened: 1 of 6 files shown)


client/clientservice/src/event_service.cpp
```diff
@@ -119,11 +127,28 @@ Status EventServiceImpl::Subscribe(ServerContext* context,
...
+    metrics_.update_processing_dur.Get().Set(update_duration_overall.count());
+
+    // update writes/sec metric
+    auto duration_overall = std::chrono::duration_cast<std::chrono::seconds>(end_overall - start_writes_per_second);
+    if (duration_overall >= std::chrono::seconds(1)) {
+      metrics_.num_writes_per_second.Get().Set(num_writes / duration_overall.count());
+      num_writes = 0;
+      start_writes_per_second = std::chrono::steady_clock::now();
+    }
+    metrics_.updateAggregator();
```
robem, line 151: When we update the values inside the aggregator we acquire a lock. I wonder if this function might be called too frequently and therefore lead to unnecessary waiting. Wouldn't it be sufficient to call it once every second? Meaning, call it whenever we calculate writes per second. @yontyon Might have suggestions.

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -741,8 +741,8 @@ void ThinReplicaClient::pushUpdateToUpdateQueue(std::unique_ptr<EventVariant> up
 
   // update metrics
   std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();
-  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
-  metrics_.update_dur_ms.Get().Set((uint64_t)(duration.count()));
+  auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+  metrics_.update_dur_ms.Get().Set(duration.count());
```
robem, line 745: Please change the variable name accordingly `_ms` -> `_us`.

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -501,6 +501,10 @@ void ThinReplicaClient::receiveUpdates() {
     LOG_WARN(logger_, "Need to stop receiving updates");
     return;
   }
+  // start time for metrics updates_received_per_second and updates_processed_per_second
+  std::chrono::steady_clock::time_point start_time = std::chrono::steady_clock::now();
```
robem, line 505: Suggestion: Rename to `start_per_second` because we already have a `start` time variable for the while loop.

client/thin-replica-client/include/client/thin-replica-client/thin_replica_client.hpp
```diff
@@ -164,6 +166,10 @@ struct ThinReplicaClientMetrics {
   // the TRC, to when it is pushed to the update queue for consumption by the
   // application using TRC
   concordMetrics::GaugeHandle update_dur_ms;
+  // number of updates received by TRC per second
+  concordMetrics::GaugeHandle updates_received_per_second;
```
robem, line 170: Specifically, this is the number of data updates (no hash updates) per second and they could come from different replicas. However, it doesn't tell us how many updates we actually processed which is why we have the other value below. So, I'm pondering how do we use this value during analysis? I guess, this is only interesting if the two `updates_received` and `updates_processed` diverge but the system doesn't get stuck. Then we can conclude that we find an f+1 agreement amongst all replicas but we get "unlucky" in how we choose the data stream!?

client/clientservice/src/event_service.cpp
yontyon: Yes, the metrics are eventually consistent, so calling the update aggregator once in a while is fine.
Notice, however, that as the metrics themselves are not thread-safe, you must call the updateAggregator from the same thread in which you update the metrics.
Note that the aggregator lock is for mutual exclusion between the metrics server thread (e.g. Prometheus HTTP server) and this should happen only once in a few seconds. With modern implementation, grabbing mutex when there is no contention is fast as test&set (see https://preshing.com/20111118/locks-arent-slow-lock-contention-is/).

client/clientservice/include/client/clientservice/event_service.hpp
```diff
@@ -18,17 +18,50 @@
...
+
+  void updateAggregator() { metrics_component_.UpdateAggregator(); }
+
+ private:
+  concordMetrics::Component metrics_component_;
+
+ public:
+  // number of updates written to the gRPC stream per second
+  concordMetrics::GaugeHandle num_writes_per_second;
+  // time taken to process an update (time between when requested is received by the event service until it is written
```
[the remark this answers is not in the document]
shgandhi: Updated.

client/clientservice/src/event_service.cpp
```diff
@@ -90,6 +92,7 @@ Status EventServiceImpl::Subscribe(ServerContext* context,
     if (not update) {
       continue;
     }
+    std::chrono::steady_clock::time_point start_overall = std::chrono::steady_clock::now(), end_overall;
```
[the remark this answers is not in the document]
shgandhi: Updated.

client/thin-replica-client/src/thin_replica_client.cpp
shgandhi: I have removed `start_per_second`, as it is no longer needed, since I have changed `updates_processed_per_second` -> `num_updates_processed` i.e., a counter.

client/thin-replica-client/include/client/thin-replica-client/thin_replica_client.hpp
shgandhi: I agree, I have removed `updates_received_per_second`.

client/clientservice/src/event_service.cpp
shgandhi: Thanks @yontyon updateAggregator will be called by the same thread that would update the metrics.

client/clientservice/include/client/clientservice/event_service.hpp
```diff
@@ -18,17 +18,50 @@
...
+    metrics_component_.SetAggregator(aggregator);
+  }
+
+  void updateAggregator() { metrics_component_.UpdateAggregator(); }
+
+ private:
+  concordMetrics::Component metrics_component_;
+
+ public:
+  // number of updates written to the gRPC stream per second
```
robem, line 41: Needs an update

robem (approved):
Lgtm. Thanks!

client/clientservice/include/client/clientservice/event_service.hpp
shgandhi: Thanks for catching, updated.

shgandhi:
Thanks for the review @robem!

PR #2587, opened 2022-04-22
Add TRC/TRS fixes

shgandhi:
This PR adds the following changes -
1. Adds missing asserts in thin_replica_impl.hpp
2. Fixes handling of requests for pruned event groups
3. Adds unit tests for pruning scenarios
4. Fixes a bug in findGlobalEventGroupID for a pruning scenario, identified via the unit tests.

### DIFF
--- a/client/proto/event/v1/event.proto
+++ b/client/proto/event/v1/event.proto
@@ -38,7 +38,7 @@ service EventService {
 
 // Events come in two formats - Events (old API) and EventGroups (new API).
 // An EventsRequest will result in a stream of EventGroups or a stream of Events which turns into a stream of EventGroups.
-// If the block requested by EventsRequest contains EventGroups only then the stream will start at EventGroup id 1.
+// If the block requested by EventsRequest contains EventGroups only then the stream will start at the first EventGroup.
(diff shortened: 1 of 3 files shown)


kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -371,12 +371,12 @@ FindGlobalEgIdResult KvbAppFilter::findGlobalEventGroupId(uint64_t external_even
   uint64_t private_start = getValueFromLatestTable(client_id_ + "_oldest");
   uint64_t private_end = getValueFromLatestTable(client_id_ + "_newest");
 
-  if (not private_start) {
+  if (not private_start and not private_end) {
```
robem, line 374: This is a good find!! Now, the code below might throw :/ I need to think about this a bit more.

robem: I added this change as part of #2592

shgandhi: Thanks, removed.

PR #2591, opened 2022-04-25
Take read operations out of the lock

yuliasherman:
* **Problem Overview**  
Combined read/write operations in ClientsManager were put under the mutex lock by https://github.com/vmware/concord-bft/pull/2571 PR. But actually, the read part of those operations could be done without any lock, as we need to protect the read operations from the PreProcessor threads only from the add/delete/modify operations done by the consensus threads.
* **Testing Done**  
Apollo regression tests.


yontyon (approved):
LGTM

PR #2592, opened 2022-04-25
KvbAppFilter: External event group ids and pruning

robem:
Recent system testing with pruning found bugs in the external event group id calculation. This MR fixes the issue and adds new unit tests for those scenarios. See commit messages for details.

### DIFF
--- a/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
+++ b/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
@@ -343,8 +343,11 @@ uint64_t KvbAppFilter::oldestExternalEventGroupId() const {
   uint64_t public_oldest = getValueFromLatestTable(kPublicEgIdKeyOldest);
   uint64_t private_oldest = getValueFromLatestTable(client_id_ + "_oldest");
   if (!public_oldest && !private_oldest) return 0;
-  if (!public_oldest) return private_oldest;
-  if (!private_oldest) return public_oldest;
+
+  // If public or private was fully pruned then we have to account for those event groups as well
+  if (!public_oldest) return private_oldest + getValueFromLatestTable(kPublicEgIdKeyNewest);
+  if (!private_oldest) return public_oldest + getValueFromLatestTable(client_id_ + "_newest");
+
   // Adding public and private results in an external event group id including two query-able event groups
   // (the oldest private and the oldest public).
   // However, we are only interested in the oldest external and not the second oldest. Hence, we have to subtract 1.
--- a/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
+++ b/kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
@@ -374,12 +374,12 @@ FindGlobalEgIdResult KvbAppFilter::findGlobalEventGroupId(uint64_t external_even
   uint64_t private_start = getValueFromLatestTable(client_id_ + "_oldest");
   uint64_t private_end = getValueFromLatestTable(client_id_ + "_newest");
 
-  if (not private_start) {
+  if (not private_start and not private_end) {
     // requested external event group id == public event group id
     uint64_t global_id;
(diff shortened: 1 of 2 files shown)


kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -418,24 +418,21 @@ FindGlobalEgIdResult KvbAppFilter::findGlobalEventGroupId(uint64_t external_even
...
-      std::tie(global_eg_id, std::ignore) = getValueFromTagTable(kPublicEgId, external_event_group_id);
-      return {global_eg_id, true, current_pvt_eg_id - 1, external_event_group_id};
-    }
     auto num_pub_egs = current_ext_eg_id - current_pvt_eg_id;
     auto pub_eg_id = num_pub_egs - (current_ext_eg_id - external_event_group_id - 1);
     std::tie(global_eg_id, std::ignore) = getValueFromTagTable(kPublicEgId, pub_eg_id);
     return {global_eg_id, true, current_pvt_eg_id - 1, pub_eg_id};
   }
 
   ConcordAssertGT(external_event_group_id, current_ext_eg_id);
```
shgandhi, line 435: Shouldn't this be `ConcordAssertGE`? Since we skip the while loop we don't check for `current_ext_eg_id == external_event_group_id` in case of pruning.

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -418,24 +418,21 @@ FindGlobalEgIdResult KvbAppFilter::findGlobalEventGroupId(uint64_t external_even
...
 
+  // If all private event groups were pruned then we need to adjust current_* to point to the last private event group
+  if (private_start == 0) {
+    ConcordAssertNE(private_end, 0);
+    current_pvt_eg_id = private_end;
+    current_ext_eg_id = private_end + public_start - 1;
+  }
+
   if (external_event_group_id < current_ext_eg_id) {
-    if (current_pvt_eg_id == private_start) {
```
shgandhi, line 419: Isn't this condition still a valid optimization for non-pruning scenarios, i.e., if `private_start != 0`?

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -418,24 +418,21 @@ FindGlobalEgIdResult KvbAppFilter::findGlobalEventGroupId(uint64_t external_even
...
-      std::tie(global_eg_id, std::ignore) = getValueFromTagTable(kPublicEgId, external_event_group_id);
-      return {global_eg_id, true, current_pvt_eg_id - 1, external_event_group_id};
-    }
     auto num_pub_egs = current_ext_eg_id - current_pvt_eg_id;
```
shgandhi, line 429: Consider a scenario, wherein -
private_start = 0, private_end = 5, public_start = 1, public_end = 5, i.e., all private event groups have been pruned. Requested external_event_group_id = 1.
Now, acc. to this change, current_pvt_eg_id = 5, current_ext_eg_id = 5.
and, num_pub_egs = 0 and pub_eg_id = -3, is that expected, or am I missing something?

shgandhi:
Thanks for the fixes @robem and clear test cases. I just have a minor comment. LGTM.

shgandhi (approved):

kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
robem: This part of the code is actually wrong and the pruning scenarios unveiled it. Just because our current pointer points to private_start doesn't mean that there were no private event groups before that point. In fact, if we pruned partially then there will be private event groups before private_start but they aren't queryable. Therefore, we cannot simply use the given external event group id but instead need to do proper calculation.
If we do that right here then we just duplicating the same logic for the "non-optimized" case and with that we end up similar to the code below which is redundant.

robem (commented):
Thanks for your review @shgandhi

PR #2595, opened 2022-04-26
Allow s3 integrity checker to recieve only required public keys configuration

toly-kournik:
* **Problem Overview**  
 Previously, due to code reuse, keys configuration file should have matched the one of the RO replica,
which included private key.
Now only replicas public keys are required.

- enhance yaml config file parser
- minor changes related to passing string_view by value.
 
* **Testing Done**  
- integrity apollo tests
- manual tests

### DIFF
--- a/kvbc/tools/db_integrity_check/integrity_checker.cpp
+++ b/kvbc/tools/db_integrity_check/integrity_checker.cpp
@@ -69,12 +69,27 @@ void IntegrityChecker::parseCLIArgs(int argc, char** argv) {
 void IntegrityChecker::initKeysConfig(const fs::path& keys_file) {
   LOG_DEBUG(logger_, keys_file);
   auto& config = bftEngine::ReplicaConfig::instance();
-  auto sys = inputReplicaKeyfileMultisig(keys_file, config);
-  (void)sys;  // currently for ro replica cryptosys is null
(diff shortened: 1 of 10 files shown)


kvbc/tools/db_integrity_check/integrity_checker.cpp
```diff
@@ -69,12 +69,27 @@ void IntegrityChecker::parseCLIArgs(int argc, char** argv) {
...
+  for (size_t i = 0; i < config.numReplicas; ++i)
+    config.publicKeysOfReplicas.insert(std::pair<uint16_t, std::string>(i, rsaPublicKeys[i]));
+
+  config.replicaId = config.numReplicas;  // "my" replica id shouldn't match one of the regular replicas
+
   repsInfo_ = new ReplicasInfo(config, true, false);
 
   bftEngine::impl::SigManager::init(config.replicaId,
-                                    config.replicaPrivateKey,
+                                    "", /*private key*/
```
yontyon, line 92: don't you need to pass the private key here?

yontyon (approved):
lgtm
only one comment

kvbc/tools/db_integrity_check/integrity_checker.cpp
toly-kournik: That's the whole PR idea - private key shouldn't be loaded

PR #2596, opened 2022-04-26
Allow batch cancellation only when completed

yuliasherman:
* **Problem Overview**  
  In case one of the requests in the batch has not reached the pre-execution consensus, we cancel the whole batch.
  It is a good approach because we can succeed only in case all the requests in the batch are completed successfully. 
  But when one of the batch requests is in the middle of pre-processing, the buffer allocated for it from the memory pool
  could not be released, otherwise, concord will crash. I've changed the approach. Now we wait until all requests complete 
  pre-processing and only then cancel the batch.
* **Testing Done**  
  Ran a one-day test on a reserved system.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -207,18 +207,39 @@ void RequestsBatch::releaseReqsAndSendBatchedReplyIfCompleted(PreProcessReplyMsg
 }
 
 // On a primary replica
+void RequestsBatch::cancelRequestAndBatchIfCompleted(const string &reqBatchCid,
+                                                     uint16_t reqOffsetInBatch,
+                                                     PreProcessingResult status) {
+  const lock_guard<mutex> lock(batchMutex_);
+  if (batchCid_ != reqBatchCid) {
+    LOG_INFO(preProcessor_.logger(),
+             "The batch has been cancelled/completed earlier; do nothing" << KVLOG(clientId_, reqBatchCid, batchCid_));
+    return;
+  }
+  const auto &reqEntry = requestsMap_[reqOffsetInBatch];
+  if (reqEntry) {
+    if (status == CANCEL) preProcessor_.preProcessorMetrics_.preProcConsensusNotReached++;
+    preProcessor_.releaseClientPreProcessRequestSafe(clientId_, reqEntry, status);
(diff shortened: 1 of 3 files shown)


ukg19:
Thanks for fix Yulia.
Have couple of questions regarding this approach?
1. Will this approach not affect the performance? As in this case, we are unnecessarily waiting for pre-processing to complete while we know that batch will be cancelled later on.
2. Did we do an LRT run to see impact on performance?
3. Can we think of another solution where we safely delete the memory for an ongoing pre-processing request?

ukg19:
Thanks for fix Yulia.
Have couple of questions regarding this approach?
1. Will this approach not affect the performance? As in this case, we are unnecessarily waiting for pre-processing to complete while we know that batch will be cancelled later on.
2. Did we do an LRT run to see impact on performance?
3. Can we think of another solution where we safely delete the memory for an ongoing pre-processing request?

yuliasherman:
The scenario when we don't reach a pre-processing consensus is very rare. It could happen once in 3-5 days of running or even less, so I don't think that the performance aspect is important here.

ukg19 (approved):
LGTM !!
Thanks @yuliasherman

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -207,18 +207,39 @@ void RequestsBatch::releaseReqsAndSendBatchedReplyIfCompleted(PreProcessReplyMsg
...
   uint32_t numOfCompletedReqs = 0;
-  {
-    const lock_guard<mutex> lock(batchMutex_);
-    if (numOfCompletedReqs_ < batchSize_) return;
-    batchCid = batchCid_;
-    batchSize = batchSize_;
-    numOfCompletedReqs = numOfCompletedReqs_;
-    resetBatchParams();
-  }
+  if (numOfCompletedReqs_ < batchSize_) return;
```
cloudnoize, line 238: if we in the cancellation flow, and ```numOfCompletedReqs_ < batchSize_```
What is the future on this batch?

yuliasherman: Does not matter how the request completes - successfully or has been canceled, numOfCompletedReqs_ increases.

cloudnoize (approved):

PR #2597, opened 2022-04-27
Allowing single replica cluster

yontyon:
* **Problem Overview**  
  We are not able to run the configuration where n=1 and f = c = 0. This PR is for fixing this
* **Testing Done**  
  GHA tests + simple single replica cluster test

### DIFF
--- a/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
+++ b/bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
@@ -182,7 +182,6 @@ class CollectorOfThresholdSignatures {
   void initExpected(SeqNum seqNumber, ViewNum view, Digest& digest) {
     ConcordAssert(seqNumber != 0);
     ConcordAssert(expectedSeqNumber == 0);
-    ConcordAssert(!processingSignaturesInTheBackground);
 
     expectedSeqNumber = seqNumber;
     expectedView = view;
@@ -193,7 +192,6 @@ class CollectorOfThresholdSignatures {
   bool initMsgWithPartialSignature(PART* partialSigMsg, ReplicaId repId) {
     ConcordAssert(partialSigMsg != nullptr);
 
-    ConcordAssert(!processingSignaturesInTheBackground);
     ConcordAssert(expectedSeqNumber != 0);
     ConcordAssert(combinedValidSignatureMsg == nullptr);
     ConcordAssert(candidateCombinedSignatureMsg == nullptr);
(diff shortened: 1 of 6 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2177,14 +2177,20 @@ void ReplicaImp::onFastPathCommitCombinedSigSucceeded(SeqNum seqNumber,
 
   SeqNumInfo &seqNumInfo = mainLog->get(seqNumber);
 
+  FullCommitProofMsg *fcp = seqNumInfo.getFastPathFullCommitProofMsg();
+  if (fcp != nullptr) {
+    LOG_WARN(CNSUS, "FullCommitProofMsg is already exist (valid only if n == 1)");
```
toly-kournik, line 2182: maybe assert on n==1?

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
```diff
@@ -182,7 +182,6 @@ class CollectorOfThresholdSignatures {
   void initExpected(SeqNum seqNumber, ViewNum view, Digest& digest) {
     ConcordAssert(seqNumber != 0);
     ConcordAssert(expectedSeqNumber == 0);
-    ConcordAssert(!processingSignaturesInTheBackground);
```
toly-kournik, line 185: why it's not valid anymore?

toly-kournik (approved):
LGTM with minor questions

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
yontyon: (a) If we directly put the signature/digest we should be caring if there is another processing in the background.
(b) Specifically, with a single replica cluster, when it recover, once it pushes the paritailCommit message (of itself) it is already have enough data to start and process the multi signature, however, it is not needed because immidetly after it directly put the fullCommit message signature

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: will do

teoparvanov (commented):
Hi Yoni, we need to be careful about possible regressions. Could you create a private build, and check that all Hermes reconfiguration tests pass? (before merging this PR)

Thanks for you diligence, 
Teo

teoparvanov (changes requested):

PR #2598, opened 2022-04-27
Exclude sub-directory iteration for db size calculation

nkumar04:
**Problem Overview**
Calculation of rocksdb dir size does not need to include the subdirectories.
rocksdbdata directory also has checkpoint folder for db-checkpoints, and
those should be excluded. To fix this we need to use directory_iterator
instead of recursive_directory_iterator.
Added support to test this apollo
**Testing Done**
Tested in apollo test for SkvbcDbSnapshotTest


itsmeakashgoyal (approved):
LGTM
Thanks Nikhil

PR #2600, opened 2022-04-27
Add another metric to TRS and allow tags for counter metrics

shgandhi:
This PR makes the following changes -
1. Metrics: Update counter metrics to allow for tags
2. TRS: Add num_skipped_event_groups metric and refactor


robem (approved):
Looks good. Thanks!

PR #2603, opened 2022-04-28
Extract trace context from http request and connect with replicas

bayarcher:
Extract OpenTracing trace context from http request and connect with trace sent to replicas

### DIFF
--- a/client/clientservice/src/request_service.cpp
+++ b/client/clientservice/src/request_service.cpp
@@ -15,54 +15,14 @@
 #include "client/clientservice/request_service.hpp"
 #include "client/concordclient/concord_client.hpp"
 #include "concord_client_request.pb.h"
+#include "client/thin-replica-client/trace_contexts.hpp"
 
+using namespace client::thin_replica_client;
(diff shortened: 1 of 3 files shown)


client/clientservice/src/request_service.cpp
```diff
@@ -22,6 +22,48 @@ namespace concord::client::clientservice {
 
 namespace requestservice {
 
+struct GrpcMetadataCarrierForReading : opentracing::TextMapReader {
```
robem, line 25: This class doesn't seem to be specific to the request service. Can it be used by other components? If so then we should put this code into a common/utility space. Can you maybe re-use or refactor the code in `client/thin-replica-client/src/trace_contexts.cpp`?

bayarcher: Refactored the code to make it reusable as suggested by the above comment.

robem (approved):
Lgtm. Ideally we can refactor `trace_contexts.cpp` to be part of a tracing utils library instead of being part of the thin replica client.

PR #2605, opened 2022-04-29
CMF C++: Allow string as serialization buffer

robem:
This change adds the ability to serialize to and deserialize from `std::string`.
The user should always prefer `std::vector<uint8_t>` over a string because it
more accurately describes serialization and when would we ever need the
terminating NULL-byte?
However, `std::string` can be used to stay compatible with libraries that use
this type for byte buffers. Currently, C++ doesn't allow us to convert between
those types without copying. Therefore, with this change we can use CMF with
those libraries without performance overhead. However, it is considered a
*workaround* because byte buffers should not be strings in the first place.

---

This change introduces new tests (which will be auto generated) to make sure serialization and deserialization still works. In addition, this change has been tested manually with closed source code that takes advantage of CMF.

### DIFF
--- a/messages/compiler/cpp/cpp_visitor.py
+++ b/messages/compiler/cpp/cpp_visitor.py
@@ -21,15 +21,26 @@ struct {name} {{
 """
 
 
-serialize_fn = "void serialize(std::vector<uint8_t>& output, const {name}& t)"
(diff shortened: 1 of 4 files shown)


shgandhi (approved):

PR #2606, opened 2022-04-30
Fix for correct timeout calculation in state snapshot service

arc-vmware:
Normally GRPC client sets a deadline for the request. If the request takes more time than the deadline, then the client times out. As State Snapshot service is a GRPC service, it also calculates a timeout based on the deadline set. But the deadline was calculated in a reverse manner and this PR is fixing that issue. There are 2 changes
with this PR:
1) The deadline calculations are correctly done.
2) In case there is no deadline, the default timeout will become 30 seconds
3) The maximum deadline that can be set from the client-side is 10 mins

So if the deadline is past (due to clock skew) or if the deadline is not set, we will assume 30 sec as timeout. Any deadline between 1 millisecond to 600000ms is considered valid. If the deadline is invalid then 30s will be choosen as the default deadline.

* **Problem Overview**  
  The timeout calculation is wrong in StateSnapshotService. This fix will be needed for 1.6
* **Testing Done**  
  Manual testing is done for this scenario.
2022-04-30T19:28:33,237,+0000|INFO ||concord.client.clientservice.statesnapshot||||state_snapshot_service.cpp:218|virtual grpc::Status concord::client::clientservice::StateSnapshotServiceImpl::GetRecentSnapshot(grpc::ServerContext *, const vmware::concord::client::statesnapshot::v1::GetRecentSnapshotRequest *, vmware::concord::client::statesnapshot::v1::GetRecentSnapshotResponse *)|Received a GetRecentSnapshotRequest with timeout : 30000ms|{"primary":""}|
^
 |
We are able to see that the correct timeout is set.

### DIFF
--- a/client/clientservice/src/state_snapshot_service.cpp
+++ b/client/clientservice/src/state_snapshot_service.cpp
@@ -202,19 +202,27 @@ static std::shared_ptr<bftEngine::RequestCallBack> getCallbackLambda(const bftEn
   return std::make_shared<bftEngine::RequestCallBack>(callback);
 }
 
+std::chrono::milliseconds StateSnapshotServiceImpl::setTimeoutFromDeadline(ServerContext* context) {
+  std::chrono::milliseconds timeout = 120s;  // This is default timeout
+  if (context != nullptr) {
+    auto end_time = context->deadline();
+    auto curr_timeout =
+        std::chrono::duration_cast<std::chrono::milliseconds>(end_time - std::chrono::system_clock::now());
+    if ((curr_timeout.count() > 0) && (curr_timeout.count() <= MAX_TIMEOUT_MS)) {
+      timeout = curr_timeout;
+    } else {
+      LOG_WARN(logger_,
+               "Extreme deadline set by client of client service "
+                   << KVLOG(curr_timeout.count(), end_time.time_since_epoch().count()));
+    }
+  }
+  return timeout;
+}
+
 Status StateSnapshotServiceImpl::GetRecentSnapshot(ServerContext* context,
                                                    const GetRecentSnapshotRequest* proto_request,
                                                    GetRecentSnapshotResponse* response) {
-  std::chrono::milliseconds timeout = 5s;  // This is default timeout
-  if (context != nullptr) {
-    auto time_left =
-        gpr_time_to_millis(gpr_time_sub(gpr_now(context->raw_deadline().clock_type), context->raw_deadline()));
-    std::chrono::milliseconds curr_timeout{time_left};
-    if (curr_timeout.count() > 0) {
-      timeout = curr_timeout;
-    }
-  }
-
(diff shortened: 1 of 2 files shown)


client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -202,19 +202,23 @@ static std::shared_ptr<bftEngine::RequestCallBack> getCallbackLambda(const bftEn
...
+    } else {
+      LOG_WARN(logger_, "Extreme deadline set by client of client service " << KVLOG(curr_timeout.count()));
     }
   }
+}
 
+Status StateSnapshotServiceImpl::GetRecentSnapshot(ServerContext* context,
+                                                   const GetRecentSnapshotRequest* proto_request,
+                                                   GetRecentSnapshotResponse* response) {
+  std::chrono::milliseconds timeout = 30s;  // This is default timeout
```
yuliasherman, line 220: Nit: you can define a static data member for this.

client/clientservice/src/state_snapshot_service.cpp
```diff
@@ -202,19 +202,23 @@ static std::shared_ptr<bftEngine::RequestCallBack> getCallbackLambda(const bftEn
   return std::make_shared<bftEngine::RequestCallBack>(callback);
 }
 
-Status StateSnapshotServiceImpl::GetRecentSnapshot(ServerContext* context,
-                                                   const GetRecentSnapshotRequest* proto_request,
-                                                   GetRecentSnapshotResponse* response) {
-  std::chrono::milliseconds timeout = 5s;  // This is default timeout
+void StateSnapshotServiceImpl::setTimeoutFromDeadline(ServerContext* context, std::chrono::milliseconds& timeout) {
```
yuliasherman, line 205: Nit: I would leave the default timeout setting in the scope of this function and not set it from outside:
std::chrono::milliseconds StateSnapshotServiceImpl::setTimeoutFromDeadline(ServerContext* context) {
  std::chrono::milliseconds timeout = 30s;
  .....
  return timeout;
}
...
Caller:
std::chrono::milliseconds  timeout = setTimeoutFromDeadline(...);

yuliasherman (approved):
LGTM

client/clientservice/src/state_snapshot_service.cpp
arc-vmware: Thanks, @yuliasherman , I will avoid static here. But definitely, i will move this inside setTimeoutFromDeadline as you explained in the next comment.

client/clientservice/src/state_snapshot_service.cpp
arc-vmware: Thanks, @yuliasherman will update the PR with this. This is really nice.

yuliasherman (approved):
LGTM

PR #2610, opened 2022-05-03
Fix typo

shgandhi:
Remove the line that incorrectly sets the last expected event group ID to 11, instead it should be set only once to `end` + 1.


arc-vmware (approved):
LGTM!!!

PR #2611, opened 2022-05-04
Added restart recovery test for View Change with F crashed replicas.

HristoStaykov:
In this test we initiate View Change multiple times by restarting F replicas
including the current Primary, but excluding the next one. In this way
we perform a View Change with only 2F+1 replicas and in general we expect
only one View increment to happen per View Change, even though if a replica
initiates State Transfer this is not necessarily guaranteed and the test is
able to handle such scenarios correctly.

* **Problem Overview**  
  We need a test where we perform View Changes multiple times with F restarted replicas. In this way we can verify the system recovers correctly with only 2F+1 quorum during the VC.
* **Testing Done**  
  N/A

### DIFF
--- a/tests/apollo/test_skvbc_restart_recovery.py
+++ b/tests/apollo/test_skvbc_restart_recovery.py
@@ -242,3 +242,109 @@ class SkvbcRestartRecoveryTest(ApolloTest):
         """
 
         await self._restarting_replica_during_system_is_in_view_change(bft_network, True)
+
+    @with_trio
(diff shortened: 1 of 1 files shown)


teoparvanov:
Hi @HristoStaykov , is this supposed to be running in CI, or is it a manual test?

HristoStaykov:
> Hi @HristoStaykov , is this supposed to be running in CI, or is it a manual test?

All tests under **test_skvbc_restart_recovery.py** are not part of the CI. They are activated with a build variable specifying **ENABLE_RESTART_RECOVERY_TESTS=true**, which by default is **false**.

teoparvanov (approved):
Thanks, looks good!

PR #2617, opened 2022-05-08
Configure number of active clients

yuliasherman:
* **Problem Overview**  
  < a few sentences about what problem we want to solve with our change(e.g. description of the bug we are fixing) >
* **Testing Done**  
  < describe automated and/or manual test scenarios >


HristoStaykov (approved):

yuliasherman:
As discussed offline, the parameter allows shrinking a client pool size on a deployed system, which is not possible with calculated operator/cre/clientservice ids. if we don't specify the number of active clients or specify it as equal to zero, all clients in the pool will be active.

PR #2618, opened 2022-05-08
Log RO replica state during boot up

mdarade:
Read Only replica to log state of S3 bucket.
This will be useful in case existing bucket gets assigned to container.

### DIFF
--- a/kvbc/src/Replica.cpp
+++ b/kvbc/src/Replica.cpp
@@ -65,7 +65,8 @@ Status Replica::initInternals() {
   m_currentRepStatus = RepStatus::Starting;
 
   if (replicaConfig_.isReadOnly) {
-    LOG_INFO(logger, "ReadOnly mode");
+    LOG_INFO(logger,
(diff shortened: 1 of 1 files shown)


kvbc/src/Replica.cpp
```diff
@@ -65,7 +65,8 @@ Status Replica::initInternals() {
   m_currentRepStatus = RepStatus::Starting;
 
   if (replicaConfig_.isReadOnly) {
-    LOG_INFO(logger, "ReadOnly mode");
+    LOG_INFO(logger,
```
glevkovich, line 68: nit: is there a reasib not to print it (with some change in style) outside for any replica in line 64?

glevkovich (approved):
@mdarade  thanks, consider my comment.

kvbc/src/Replica.cpp
mdarade: For backup replica's such information is printed just before this function call.

PR #2620, opened 2022-05-12
Added kvb tool to hash state keys filtered with ACL

nkumar04:
* **Problem Overview**  
This tool is added as part of POC work for client verifiability. It uses
rocks-db snapshot to hash state keys and value for a client id from an
offset=L(define by event group id). Such that L < M (latest event group
id).

* **Testing Done**  
Tool is tested with command line argument on a rocksdb database.


teoparvanov (changes requested):

PR #2621, opened 2022-05-12
Moving helper functions in db-checkpoint test suite to BftTestNetwork

nkumar04:
* **Problem Overview**  
 Moving common utility functions used db-checkpoint apollo test suite
to BftTestNetwork for better code re-usability
* **Testing Done**  
  Changes will be tested as part of apollo test run in CI


toly-kournik:
Nit: since we are not using a separate "db_snapshot_helper" it would be more intuitive to use `db_snapshot` instead of just a `snapshot` in function names

toly-kournik (approved):
Thank you @nkumar04 
LGTM!

PR #2625, opened 2022-05-13
Bayarcher/client service trace tags

bayarcher:
Remove ClientService instance ID from application name in OpenTracing traces and add the Instance ID as Tags


robem (approved):
Thanks for the fix 💪

PR #2628, opened 2022-05-16
Add logs for TRS/TRC communication

situ-s:
Added log to get server address and client address during TRS and TRC communication.


yuliasherman (approved):
LGTM

PR #2629, opened 2022-05-16
Apollo test. View change when primary is behind

titusen:
The Apollo test, which should be part of the test_skvbc_restart_recovery suite needs to implement the following steps:

Start all Replicas.
Stop the expected next Primary.
Advance the system 5 Checkpoints.
Stop the current Primary.
Send Client operations in order to trigger View Change.
Bring back up all the previously stopped replicas.
Verify View Change was successful.
Goto Step 2.

### DIFF
--- a/tests/apollo/test_skvbc_restart_recovery.py
+++ b/tests/apollo/test_skvbc_restart_recovery.py
@@ -348,3 +348,86 @@ class SkvbcRestartRecoveryTest(ApolloTest):
         await bft_network.wait_for_fast_path_to_be_prevalent(
             run_ops=lambda: skvbc.run_concurrent_ops(num_ops=20, write_weight=1), threshold=20)
 
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: c == 0, rotate_keys=True)
+    @verify_linearizability()
+    async def test_recovering_of_primary_with_initiated_view_change(self, bft_network, tracker):
+        """
+        The Apollo test, which should be part of the test_skvbc_restart_recovery suite needs to implement the following steps:
+        1) Start all Replicas.
+        2) Stop the expected next Primary.
+        3) Advance the system 5 Checkpoints.
+        4) Stop the current Primary.
+        5) Send Client operations in order to trigger View Change.
+        6) Bring back up all the previously stopped replicas.
+        7) Verify View Change was successful.
+        8) Goto Step 2.
+        """
+
+        # start replicas
+        [bft_network.start_replica(i) for i in bft_network.all_replicas()]
+
+        loop_counter = 0
+        while (loop_counter < 100):
+            loop_counter = loop_counter + 1
+            log.log_message(f"Loop run {loop_counter}")
+            # loop start
+            view = await bft_network.get_current_view()
+
+            primary = await bft_network.get_current_primary()
+            next_primary = (primary + 1) % bft_network.config.n
+            # Stop the expected next Primary.
+            bft_network.stop_replica(next_primary)
+
+            # Advance the system 5 Checkpoints.
+            skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
+            await skvbc.fill_and_wait_for_checkpoint(
+                initial_nodes=bft_network.all_replicas(without = {next_primary}),
+                num_of_checkpoints_to_add=5,
+                verify_checkpoint_persistency=False,
+                assert_state_transfer_not_started=False
+            )
+
+            # Stop the current Primary
+            bft_network.stop_replica(primary)
+
+            # Send Client operations in order to trigger View Change.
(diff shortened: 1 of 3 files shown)


tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -348,3 +348,42 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
         await bft_network.wait_for_fast_path_to_be_prevalent(
             run_ops=lambda: skvbc.run_concurrent_ops(num_ops=20, write_weight=1), threshold=20)
 
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: c == 0, rotate_keys=True)
```
teoparvanov, line 352: `c==0` is the default one now, so no need to specify it:
https://github.com/vmware/concord-bft/blob/master/tests/apollo/util/bft.py#L95

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -348,3 +348,42 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
...
+
+        # Send Client operations in order to trigger View Change.
+        await skvbc.run_concurrent_ops(10)
+
+        #Bring back up all the previously stopped replicas.
+        bft_network.start_replica(0)
+        bft_network.start_replica(1)
+
+        #Verify View Change was successful.
+        await bft_network.wait_for_replicas_to_reach_view(bft_network.all_replicas(), 2)
```
teoparvanov, line 389: As per step 8) defined above - isn't this sequence supposed to run in a loop?

teoparvanov:
Hi @titusen, in addition to a successful test outcome, please also carefully analyze the logs, to make sure the protocols are indeed triggered as expected.

tests/apollo/test_skvbc_restart_recovery.py
StefanPalashev: Just as a note to Teo's comment, please bear in mind that if this sequence runs in a loop the next expected primary will no longer be 1.
If this is the case, calling `await bft_network.get_current_primary()` might be helpful.

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -348,3 +348,42 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
...
+
+        # start replicas
+        [bft_network.start_replica(i) for i in bft_network.all_replicas()]
+
+        # Stop the expected next Primary.
+        bft_network.stop_replica(1)
+
+        # Advance the system 5 Checkpoints.
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
+        await skvbc.run_concurrent_ops(150 * 5)
```
StefanPalashev, line 376: I'd suggest using skvbc's `fill_and_wait_for_checkpoint` method, as it guarantees that the checkpoints have been reached.

tests/apollo/test_skvbc_restart_recovery.py
titusen: how many times it should run?

tests/apollo/test_skvbc_restart_recovery.py
titusen: done

tests/apollo/util/bft.py
```diff
@@ -1766,7 +1766,7 @@ async def expected_view_to_be_reached():
                 key = ['replica', 'Gauges', 'view']
                 replica_view = await self.retrieve_metric(replica_id, *key)
 
-                if replica_view is not None and replica_view == expected_view:
+                if replica_view is not None and replica_view >= expected_view:
```
StefanPalashev, line 1769: I don't think this statement has to be modified as it aims to prove that the given replica has reached exactly the specified view. The function `async def wait_for_replica_to_reach_view` was created as a helper of the function `async def wait_for_replicas_to_reach_view`, which opens a nursery where all of the replicas are tested to be in the same view. Changing this line would break/invalidate the guarantees that we were relying on in the tests that were using the function `async def wait_for_replicas_to_reach_view` before.

My suggestion is to create a separate function, which checks whether _the expected view or a higher one_ is reached or not.

HristoStaykov: I agree with Stefan, we have many test cases in CI that for example arrange the situation to only do a single View Change with 1 View increment. Might be it will be best to create another function that loosens up the expectations for the longer running tests.

HristoStaykov (changes requested):
Great work, @titusen ! I only have 1 recommendation that coincides with Stephan's suggestion.

HristoStaykov (approved):

teoparvanov (approved):
LGTM, thanks for the great work on this test scenario!

PR #2630, opened 2022-05-16
Add state transfer callback for publish client key request

yontyon:
* **Problem Overview**  
  When a lagged replica has done with state transfer it needs to update the published client keys such that it won't execute the request (executing the request may cause a deviation in the reserved pages hash).
This PR is about adding a state transfer callback for this request
* **Testing Done**  
  GHA tests


cloudnoize (approved):
Looks great, thanks @yontyon !!

PR #2643, opened 2022-05-19
Update pruning metrics when pruning is done after state transfer

yontyon:
* **Problem Overview**  
  When pruning is done after state transfer, we use a code path that doesn't update the pruning metrics
This PR is for fixing this bug
* **Testing Done**  
  CI tests + enabling buck the test_pruning_with_failures_test and update it to also test that the lagged replica updates its metrics


eyalrund:
please add bug id and short RCA and fix description

cloudnoize (approved):
LGTM

teoparvanov:
As discussed offline, @yontyon will test this specific scenario locally, and describe the outcome in the "Testing done" section.

Thanks for your diligence, @yontyon !
Teo

eyalrund (approved):

PR #2644, opened 2022-05-19
Added utility to parse subject field from tls cert bundle

nkumar04:
* **Problem Overview**  
  Client service user inputs like trs-tls cert path is taken from command
line argument and passed directly into a call to popen. This could allow
aribitrary command injection attack through command line argument. This
PR addresses the issue by using openssl api call to parse the
certificate instead of using OS command directly. The utility is added
to CryptoUtil class. 
* **Testing Done**  
  GHA test (gtests are added to test the api in the same PR)


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -734,30 +716,8 @@ class ThinReplicaImpl {
   }
 
   static std::string getClientIdFromCertificate(const std::string& client_cert_path, bool use_unified_certs = false) {
-    std::array<char, certSubjectLength> buffer;
```
situ-s, line 737: Please remove certSubjectLength static variable as its not used anymore.

situ-s, line 737: Please remove certSubjectLength static variable as its not used anymore.

situ-s (approved):
Given a minor comment.
LGTM!

toly-kournik (approved):
LGTM!
I don't believe we did it using a command line utility

nityavasudevan (approved):

PR #2649, opened 2022-05-22
Fix max message size

yontyon:
* **Problem Overview**  
MessageBase considers the span size outside of the regular "size" parameter. Hence, in some cases, we may end up letting a message be larger than the allowed size.  For example, if a prePrepare message span size is smaller than the max span size, we may end up with a message that exceeds a bit from the maximal size. This change is for fixing this issue
* **Testing Done**  
  CI tests

### DIFF
--- a/bftengine/src/bftengine/messages/PrePrepareMsg.cpp
+++ b/bftengine/src/bftengine/messages/PrePrepareMsg.cpp
@@ -151,12 +151,13 @@ PrePrepareMsg::PrePrepareMsg(ReplicaId sender,
                              const concordUtils::SpanContext& spanContext,
                              const std::string& batchCid,
                              size_t size)
-    : MessageBase(sender,
-                  MsgCode::PrePrepare,
-                  spanContext.data().size(),
-                  (((size + sizeof(Header) + batchCid.size()) < maxMessageSize<PrePrepareMsg>())
-                       ? (size + sizeof(Header) + batchCid.size())
-                       : maxMessageSize<PrePrepareMsg>() - spanContext.data().size())) {
+    : MessageBase(
+          sender,
+          MsgCode::PrePrepare,
+          spanContext.data().size(),
+          (((size + sizeof(Header) + batchCid.size()) + spanContext.data().size() < maxMessageSize<PrePrepareMsg>())
+               ? (size + sizeof(Header) + batchCid.size())
+               : maxMessageSize<PrePrepareMsg>() - spanContext.data().size())) {
   bool ready = size == 0;  // if null, then message is ready
   if (!ready) {
     b()->digestOfRequests.makeZero();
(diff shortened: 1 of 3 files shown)


eyalrund (changes requested):
how is this change solved the mentioned issue?

eyalrund:
please more details about the problem (what it caused) and about the proposed fix (how it addresses the problem)

yontyon:
@eyalrund, by mistake I didn't commit the correct changes.
Please review again

bftengine/src/bftengine/messages/PrePrepareMsg.hpp
```diff
@@ -155,7 +155,7 @@ class RequestsIterator {
 
 template <>
 inline MsgSize maxMessageSize<PrePrepareMsg>() {
-  return ReplicaConfig::instance().getmaxExternalMessageSize() + MessageBase::SPAN_CONTEXT_MAX_SIZE;
+  return ReplicaConfig::instance().getmaxExternalMessageSize() - MessageBase::SPAN_CONTEXT_MAX_SIZE;
```
ukg19, line 158: looks good.
but why do we need to subtract SPAN_CONTEXT_MAX_SIZE? 
If SPAN_CONTEXT is added after the PPM is framed, then it looks fine and can you pls mention the same in the PR description? Thanks.

yontyon: Because if we exceeds the maxExternalMessagesSize, the messages will be dropped by the communication layer

yontyon: So spans has to be part of the message itself

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -156,7 +156,7 @@ PrePrepareMsg::PrePrepareMsg(ReplicaId sender,
                   spanContext.data().size(),
                   (((size + sizeof(Header) + batchCid.size()) < maxMessageSize<PrePrepareMsg>())
                        ? (size + sizeof(Header) + batchCid.size())
-                       : maxMessageSize<PrePrepareMsg>() - spanContext.data().size())) {
+                       : maxMessageSize<PrePrepareMsg>() - MessageBase::SPAN_CONTEXT_MAX_SIZE)) {
```
ukg19, line 159: Thanks for offline clarification about changing this logic.
There are multiple conditions here and some calculation being done, If possible can we add a unit test in PrePrepareMsg_test.cpp to make sure that sizeof PPM doesn't go beyond max_external_msg_size?

ukg19: Thanks for test.

eyalrund (approved):

PR #2652, opened 2022-05-25
Fix races between pruning, execution and main thread

yontyon:
* **Problem Overview**  
  In case we run in async pruning mode and we have a prePrepare message that contains pruning request + regular post-execution request we can hit the following scenario:
  1. The pruning thread starts working
  2. The main thread schedule the regular request for execution in the post-execution thread
  3. The post-execution thread adds block while pruning (causing a state deviation between the replicas)

The solution provided before (see https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/RequestHandler.cpp#L175) was to return from execution if we had a pruning request in the prePrepeare message. This solution won't work when we have a separated thread for post-execution because pruning is executed in synchronous mode as part of the execution of the special requests. 
On the other hand, we cannot simply abort the execution in case async pruning is running because pruning is a local operation in the replica.
(If you notice carefully, this solution is not even correct in the regular case, because pruning may take different time in different replicas)

The solution suggested in this PR is to use a mutex to block the post-execution thread in case pruning has started. 
__Correctness__: When we have a pruning request in a prePrepare message we first execute it synchronously in the main thread as part of the special requests. Hence, every "regular" request will be executed after pruning has started.
On the other hand, in case of pruning was not done when the regular request was scheduled for execution, the post-execution thread will simply wait for pruning to be finished.
This means that all replicas will execute the other requests in the prePrepare only after the pruning thread releases the lock and hence the state will be identical.
__Efficiency__: As long as there is no contention, grabbing modern mutex is cheap as testAndSet operation. Hence, this should not affect performance (see https://preshing.com/20111118/locks-arent-slow-lock-contention-is/)
* **Testing Done**  
  CI tests (unable to reproduce as this is a non-deterministic test)

### DIFF
--- a/bftengine/src/bftengine/RequestHandler.cpp
+++ b/bftengine/src/bftengine/RequestHandler.cpp
@@ -34,6 +34,7 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
                              std::optional<Timestamp> timestamp,
                              const std::string& batchCid,
                              concordUtils::SpanWrapper& parent_span) {
+  bool has_pruning_request = false;
   for (auto& req : requests) {
     if (req.flags & KEY_EXCHANGE_FLAG) {
       KeyExchangeMsg ke = KeyExchangeMsg::deserializeMsg(req.request, req.requestSize);
@@ -51,6 +52,7 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
     } else if (req.flags & MsgFlag::RECONFIG_FLAG) {
       ReconfigurationRequest rreq;
       deserialize(std::vector<std::uint8_t>(req.request, req.request + req.requestSize), rreq);
+      has_pruning_request = std::holds_alternative<concord::messages::PruneRequest>(rreq.command);
       ReconfigurationResponse rsi_res = reconfig_dispatcher_.dispatch(rreq, req.executionSequenceNum, timestamp);
       // in case of read request return only a success part of and replica specific info in the response
       // and the rest as additional data, since it may differ between replicas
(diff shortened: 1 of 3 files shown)


eyalrund:
We need to have a maestro performance run for multiple hours before we can merge this change

teoparvanov (commented):
Thanks @yontyon, I'd suggest the following additional verification steps:
- code review by @cloudnoize 
- running all Hermes tests (including PruningTests suite) on a private build including these changes
- **NB!** update the PR description with a detailed explanation about the additional testing done

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -168,11 +170,15 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
...
-    if (bftEngine::ControlStateManager::instance().getPruningProcessStatus()) {
+    if (has_pruning_request) {
+      // Stop pricessing requests after pruning has started (relevant for both, sync and async execution)
       req.outActualReplySize = 0;
     }
   }
-  if (bftEngine::ControlStateManager::instance().getPruningProcessStatus()) {
+  if (has_pruning_request) {
+    // Dont waste your time in calling to the user command handler. In case of sync execution all requests were marked
+    // as not for execution. In xase of async execution, the set of requests contains only special requests and not user
```
HristoStaykov, line 180: typo: In case

HristoStaykov (approved):
Thanks @yontyon . The fix looks correct to me, we need a sync mechanism that will not allow an execution job to interfere with pruning, thus we need to use a lock, an atomic variable won't by itself work correctly, because we also need to block and wait, which should be correct now.

yontyon:
@teoparvanov , yes, the run has just been completed successfully. We can safely merge this change.
Can you approve plz?

PR #2653, opened 2022-05-26
Assert because of two transactions on checkpoint

HristoStaykov:
* **Problem Overview**  
 Because we store the checkpoint in one transaction and update 
the sequence ID in another we can potentially recover in a situation 
where we have already created a checkpoint for the current sequence ID.
This is something that can only happen due to restarts.
We need to resolve the situation in the recovery procedure, because
the storage mechanisms used in ReplicaImp and State Transfer are different
in order to improve decoupling and stacking the 2 transactions in 1 is not
feasible. This bug was found by the Apollo's restart recovery suite test
test_restarting_replica_with_client_load.
* **Testing Done**  
Apollo's test test_restarting_replica_with_client_load was used to both discover this problem and to verify correct recovery after the fix.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -563,6 +563,13 @@ void BCStateTran::createCheckpointOfCurrentState(uint64_t checkpointNumber) {
   ConcordAssert(running_);
   ConcordAssert(!isFetching());
   ConcordAssertGT(checkpointNumber, 0);
+  if (checkpointNumber == lastStoredCheckpointNumber) {
+    // We persist the lastStoredCheckpointNumber in a separate transaction from the actual
+    // update of the lastExecutedSeqNum. Since we have no mechanism to batch the 2 transactions
+    // together we need to handle this rare recovery situation.
+    LOG_WARN(logger_, "checkpointNumber == lastStoredCheckpointNumber" << KVLOG(checkpointNumber));
+    return;
+  }
   ConcordAssertGT(checkpointNumber, lastStoredCheckpointNumber);
 
   metrics_.create_checkpoint_++;


CMakeLists.txt
```diff
@@ -80,9 +81,13 @@ string(APPEND CMAKE_CXX_FLAGS " -Wbuiltin-macro-redefined")
...
 string(APPEND CMAKE_CXX_FLAGS " -Werror")
 string(APPEND CMAKE_CXX_FLAGS " -fno-omit-frame-pointer")
-if(LEAKCHECK AND THREADCHECK)
+if((LEAKCHECK AND THREADCHECK) OR (ADDRESSCHECK AND THREADCHECK))
     message(FATAL_ERROR "Cannot have both Thread and Address/Leak Sanitizers Enabled")
 endif()
+if(ADDRESSCHECK)
+    string(APPEND CMAKE_CXX_FLAGS " -fsanitize=address")
+    message("-- Address Sanitizers Enabled")
+endif()
```
teoparvanov, line 90: Is this change intentionally included in the PR?

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -79,11 +79,11 @@ async def test_restarting_replica_with_client_load(self, bft_network):
...
-        # class foo:
-        #     def log_message(self, var):
-        #         print(f"{var}")
-        #
-        # log = foo()
+        class foo:
+            def log_message(self, var):
+                print(f"{var}")
+
+        log = foo()
```
teoparvanov, line 86: Outside of local debugging, maybe this should remain commented out?

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -259,11 +259,11 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
...
```
teoparvanov, line 266: same as above

tests/apollo/util/bft.py
```diff
@@ -764,11 +764,11 @@ def start_replica(self, replica_id):
...
             self.test_dir = os.path.join(self.builddir, "tests", "apollo", "logs", test_name, self.current_test)
             replica_test_log_path = os.path.join(self.test_dir, f"stdout_{replica_id}.log")
-
+            replica_test_log_path_err = os.path.join(self.test_dir, f"stdout_{replica_id}_err.log")
             os.makedirs(self.test_dir, exist_ok=True)
 
-            stdout_file = open(replica_test_log_path, 'w+')
-            stderr_file = open(replica_test_log_path, 'w+')
+            stdout_file = open(replica_test_log_path, 'a+')
+            stderr_file = open(replica_test_log_path_err, 'a+')
```
teoparvanov, line 771: local debug code?

teoparvanov (commented):
It looks like this PR should be trimmed down to just the `.cpp` file changes?

CMakeLists.txt
HristoStaykov: Unintentionally included :) will revert

teoparvanov (approved):
Thanks Hristo, looks great!

PR #2657, opened 2022-05-31
BCStateTran: initialize session and avoid closing an already closed one

glevkovich:
startTime_ is initialized in ctor. Print warning if trying to close
and already closed session.

* **Problem Overview**  
This is a minor fix to prevent wrong close nd log print when St starts. There is no actual bug fixed.
* **Testing Done**  
no testing is needed for such a minor fix. We relay on current CI.


StefanPalashev (approved):
LGTM, thanks @glevkovich.

PR #2658, opened 2022-05-31
RangeValidationTree: change id_to_node_ into a map

glevkovich:
id_to_node_ is an unordered map. This is wrong when trying to serialize,
as the checkpoints, data, and hash representation will all be
mismatched. here we convert it into an ordered map.

* **Problem Overview**  
Ask for checkpoint summary message mismatch. The reason is the use of unordered_map in the RVT. We must use an ordered map.
* **Testing Done**  
Trying to reproduce locally. In addition, RVT conflict detection is committed soon - this should detect such issues much earlier.


HristoStaykov (approved):
Good catch @glevkovich , thanks!

PR #2661, opened 2022-06-01
Persist the lastBlockId in metadata on unwedge

yontyon:
* **Problem Overview**  
  For recovery, we persist the last reachable block id after every execution. However, on unwedge (without restart) we add a block due to a **read request**, hence we do not persist the last reachable block id. If we do restart right after the wedge, the recovery mechanism will delete the new epoch block and the system will be stayed wedged.
The solution is to write the latest reachable block id to the metadata after successful unwedge.

* **Testing Done**  
The test_unwedge_command Apollo test was modified a bit to also test that we are not wedged even if we restart the replicas after the wedge

### DIFF
--- a/bftengine/include/bftengine/IRequestHandler.hpp
+++ b/bftengine/include/bftengine/IRequestHandler.hpp
@@ -23,6 +23,7 @@
 #include "OpenTracing.hpp"
 #include "TimeService.hpp"
 #include "ISystemResourceEntity.hpp"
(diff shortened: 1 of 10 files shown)


nkumar04 (approved):
LGTM!
Thanks!

eyalrund:
Please also describe the solution and we need some specific tests for this scenrio

eyalrund (approved):

PR #2662, opened 2022-06-02
St rvt conflict detection

glevkovich:
* **Problem Overview**  
RVT conflict detection is Introduced here. We would like to detect diverging of the State Transfer RVB data as part of the normal consensus, during checkpointing. The hash of the EVT data is now sent as part of a checkpoint message and is being compared in each replica as part of identical 2f+1 message collection processes. In case there is a divergence (mismatch) in the RVB data, consensus will halt.

* **Testing Done**  

1. Added one unit-test in bcstatetransfer_tests.
2. Added 5 new apollo test cases (under test_skvbc_checkpoints.py)
3. manual tests for stability on a Maestro deployment.


eyalrund (approved):

StefanPalashev (approved):
LGTM, Great job on further improving some of the code! Thanks, @glevkovich.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -375,6 +397,146 @@ async def test_checkpoint_propagation_after_f_nodes_including_primary_isolated(s
...
+            checkpoint_of_another_replica_id = await bft_network.wait_for_checkpoint(replica_id)
+
+            assert checkpoint_of_another_replica_id == target_checkpoint, \
+                f"Mismatch on target checkpoint: target_checkpoint={target_checkpoint} \
+                    checkpoint_of_another_replica_id={checkpoint_of_another_replica_id}"
+
+            if replica_id not in byzantine_replica_ids:
+                another_number_of_checkpoint_mismatch = await bft_network.metrics.get(replica_id, *key)
+                assert checkpoint_of_another_replica_id == target_checkpoint, \
+                    f"Mismatch on mismatch checkpoints metrics: \
```
StefanPalashev, line 536: Nit: This message seems a bit strange.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -375,6 +397,146 @@ async def test_checkpoint_propagation_after_f_nodes_including_primary_isolated(s
...
+        for replica_id in bft_network.all_replicas(without={0}) :
+            checkpoint_of_another_replica_id = await bft_network.wait_for_checkpoint(replica_id)
+
+            assert checkpoint_of_another_replica_id == target_checkpoint, \
+                f"Mismatch on target checkpoint: target_checkpoint={target_checkpoint} \
+                    checkpoint_of_another_replica_id={checkpoint_of_another_replica_id}"
+
+            if replica_id not in byzantine_replica_ids:
+                another_number_of_checkpoint_mismatch = await bft_network.metrics.get(replica_id, *key)
+                assert checkpoint_of_another_replica_id == target_checkpoint, \
```
StefanPalashev, line 535: I think `another_number_of_checkpoint_mismatch` should have been used in this comparison.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -375,6 +397,146 @@ async def test_checkpoint_propagation_after_f_nodes_including_primary_isolated(s
...
+
+        await skvbc.fill_and_wait_for_checkpoint(
+            initial_nodes=bft_network.all_replicas(without={current_primary}),
+            num_of_checkpoints_to_add=4,
+            verify_checkpoint_persistency=False
+        )
+
+        key1 = ['checkpoint_msg', 'Counters', 'number_of_checkpoint_mismatch']
+        key2 = ['replica', 'Gauges', 'lastExecutedSeqNum']
+        for replica_id in bft_network.all_replicas(without={current_primary}) :
```
StefanPalashev, line 486: The current primary has to be included in here in order to verify -> 4) Verify that all replicas reached checkpoint 4, including the primary.

StefanPalashev (changes requested):
Please, take a look at my comments related to the newly added tests and apply the necessary changes before merging, thanks.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -609,20 +625,34 @@ void BCStateTran::getDigestOfCheckpointImpl(uint64_t checkpointNumber,
   ConcordAssert(psd_->hasCheckpointDesc(checkpointNumber));
 
   DataStore::CheckpointDesc desc = psd_->getCheckpointDesc(checkpointNumber);
-  LOG_INFO(logger_,
-           KVLOG(desc.checkpointNum, desc.maxBlockId, desc.digestOfMaxBlockId, desc.digestOfResPagesDescriptor));
+  auto rvbDataSize = desc.rvbData.size();
+  size_t copysize = std::min((uint16_t)sizeof(Digest), sizeOfDigestBuffer);
```
arc-vmware, line 629: why do we need to downcast the sizeof Digest. Instead we can upcast sizeOfDigestBuffer
auto copysize = std::min(sizeof(Digest), static_cast<size_t>(sizeOfDigestBuffer));

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -596,11 +596,27 @@ void BCStateTran::createCheckpointOfCurrentStateImpl(uint64_t checkpointNumber)
...
+  });
+  return defaultRvbDataDigest;
+}
+
 void BCStateTran::getDigestOfCheckpointImpl(uint64_t checkpointNumber,
                                             uint16_t sizeOfDigestBuffer,
                                             uint64_t &outBlockId,
                                             char *outStateDigest,
-                                            char *outOtherDigest) {
+                                            char *outResPagesDigest,
```
arc-vmware, line 618: For digest we have a type called Digest. Why are we not using Digest& instead of char* ptr.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2301,7 +2302,8 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
                                                               msgIsStable,
                                                               msgState,
                                                               stateDigest,
-                                                              otherDigest));
+                                                              reservedPagesDigest,
+                                                              rvbDataDigest));
   LOG_INFO(GL, "My " << KVLOG(lastStableSeqNum, lastExecutedSeqNum, getSelfEpochNumber()));
```
arc-vmware, line 2307: Good to remove the word "My" from the info log. We can use "Current" or leave that blank.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2343,18 +2345,22 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
       // <= to allow repeating checkpoint message since state transfer may not kick in when we are inside active
       // window
       if (pos != tableOfStableCheckpoints.end()) delete pos->second;
-      CheckpointMsg *x = new CheckpointMsg(msgGenReplicaId, msgSeqNum, msgState, stateDigest, otherDigest, msgIsStable);
+      CheckpointMsg *x = new CheckpointMsg(
```
arc-vmware, line 2348: should we rename x to stableChkPtMsg. 'x' is not giving any information about the variable.

arc-vmware, line 2348: x is created using new and then assigned to tableOfStableCheckpoints.
tableOfStableCheckpoints is std::map<ReplicaId, CheckpointMsg*>
Should we use std::map<ReplicaId, std::shared_ptr<CheckpointMsg>>

create 'x' using std::make_shared. 
This will help us not to worry about the memory leak as I am not sure how is the delete called for each of these new.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5190,14 +5204,17 @@ void ReplicaImp::finalizeExecution() {
...
-        checkpointNum, sizeof(Digest), checkState, (char *)&checkDigest, (char *)&otherDigest);
-    CheckpointMsg *checkMsg =
-        new CheckpointMsg(config_.getreplicaId(), lastExecutedSeqNum, checkState, checkDigest, otherDigest, false);
+    stateTransfer->getDigestOfCheckpoint(checkpointNum,
+                                         sizeof(Digest),
+                                         checkState,
+                                         stateDigest.content(),
+                                         reservedPagesDigest.content(),
+                                         rvbDataDigest.content());
+    CheckpointMsg *checkMsg = new CheckpointMsg(
```
arc-vmware, line 5216: when is checkMsg is deleted. I think as new and delete callers are different, better we use std::shared_ptr to avoid any memory leak.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5511,14 +5528,17 @@ void ReplicaImp::executeRequestsInPrePrepareMsg(concordUtils::SpanWrapper &paren
...
-        checkpointNum, sizeof(Digest), state, (char *)&stateDigest, (char *)&otherDigest);
-    CheckpointMsg *checkMsg =
-        new CheckpointMsg(config_.getreplicaId(), lastExecutedSeqNum, state, stateDigest, otherDigest, false);
+    stateTransfer->getDigestOfCheckpoint(checkpointNum,
+                                         sizeof(Digest),
+                                         state,
+                                         stateDigest.content(),
+                                         reservedPagesDigest.content(),
+                                         rvbDataDigest.content());
+    CheckpointMsg *checkMsg = new CheckpointMsg(
```
arc-vmware, line 5540: Same here, use of std::shared_ptr is recommended.

bftengine/src/bftengine/messages/CheckpointMsg.cpp
```diff
@@ -68,5 +70,45 @@ void CheckpointMsg::validate(const ReplicasInfo& repInfo) const {
   // TODO(GG): consider to protect against messages that are larger than needed (here and in other messages)
 }
 
+bool CheckpointMsg::equivalent(const CheckpointMsg* a, const CheckpointMsg* b) {
+  bool equal = (a->seqNumber() == b->seqNumber()) && (a->stateDigest() == b->stateDigest()) &&
```
arc-vmware, line 74: it will be good to add a one-liner comment defining the equivalent Checkpoint message.

bftengine/src/simplestatetransfer/SimpleStateTran.cpp
```diff
@@ -61,8 +61,8 @@ class SimpleStateTran : public ISimpleInMemoryStateTransfer {
                              uint16_t sizeOfDigestBuffer,
                              uint64_t& outBlockId,
                              char* outStateDigest,
-                             char* outFullStateDigest) override;
-
+                             char* outResPagesDigest,
```
arc-vmware, line 64: Same comment for all the functions taking char* digest to convert the type to Digest& if that's better.

arc-vmware (approved):
very good code @glevkovich, enjoyed reading the same. I have very few minor comments from a code maintenance perspective as the fix is addressing conflict detection very well. I liked the test cases and the very well-crafted strategies.

I am approving to avoid cycles of review as the code LGTM!!!, which we will discuss in the comments.

tests/apollo/test_skvbc_checkpoints.py
glevkovich: fixed

tests/apollo/test_skvbc_checkpoints.py
glevkovich: good catch, thanks..fixed.

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -375,6 +397,146 @@ async def test_checkpoint_propagation_after_f_nodes_including_primary_isolated(s
...
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+
+        await skvbc.fill_and_wait_for_checkpoint(
+            initial_nodes=bft_network.all_replicas(without=byzantine_replica_ids),
+            num_of_checkpoints_to_add=2,
+            verify_checkpoint_persistency=False
+        )
+
+        assert 0 not in byzantine_replica_ids, \
+            f"replica id 0 should not be in \byzantine_replica_ids={byzantine_replica_ids}"
```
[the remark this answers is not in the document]
glevkovich: Thats a typo, fixed.

tests/apollo/test_skvbc_checkpoints.py
glevkovich: I'm going to fix things differently:

- fill_and_wait_for_checkpoint will wait to al 7 backup replicas
- In the loop verify mismatches for not primary, and minimal  seqno for all

bftengine/src/bcstatetransfer/BCStateTran.cpp
glevkovich: That's fine, I agree this one is better. Thanks.

bftengine/src/bcstatetransfer/BCStateTran.cpp
glevkovich: Well, this is more of an API design discussion.
I didn't write the original API but I can tell you what I think, from my past conversations with, and from how we deal similar code inside ST - the idea is to be able to handle an arbitrary type fo data, although it is not really practical currently, as all calls work on Digest types. 

The same question could have been asked also for the need of sizeOfDigestBuffer.

bftengine/src/bftengine/ReplicaImp.cpp
glevkovich: I did, thanks

glevkovich: I think you are right. I think we should not to approve any PRs which use raw pointer allocations, unless there is a very good reason for that. Since we are hardly deal with low level programming, I don't see any reason for raw pointer allocations at all.
This one belongs to do original code.
I'll fix it.

glevkovich: Specifically here, I need to see how hard it will be to make the change, as there a whole logic around deleting the pointer, and I don't want to get into here. If I can't do it here, I might add a new ticket and assign to one of the managers.

glevkovich: @arc-vmware , I was able to convert it to a shared pointer (and removed some code accordingly)

bftengine/src/bftengine/ReplicaImp.cpp
glevkovich: I'm afraid that this is out of the scope for his PR, as this requires a large change.

This checkpoint is transffered to multiple modules as a raw pointer. The big change here is to hold the internal pointer as shared one (since the object is pointed by multiple modules), and not the one outside.

bftengine/src/bftengine/ReplicaImp.cpp
glevkovich: See my comment above.

bftengine/src/bftengine/messages/CheckpointMsg.cpp
glevkovich: @arc-vmware , I don't undertstand, plase clarify.

bftengine/src/simplestatetransfer/SimpleStateTran.cpp
glevkovich: See my answer above, I think it was intentionally made to be char*, although not really used for now.

StefanPalashev (approved):

StefanPalashev (commented):
@arc-vmware, thanks for the constructive feedback and ideas on how to improve the code quality.
@glevkovich, thanks for addressing the previous comments and applying the suggested changes.

PR #2664, opened 2022-06-06
As part of the effort to remove dependencies installation duplications:

toly-kournik:
- Modify concord-bft Dockerfile not to remove tools needed for future steps Docker successors (git, python, wget).
 - Replace using of python libs in format-code.sh for obtaining absolute path by "realpath".
 - Add environment variable CI="on" for a newer version of a minio server, which allows to use a root partition for a minio data dir.
 - Increment concord-bft image version to 0.42

* **Problem Overview**  
 Update concord-bft image to allow for upper layer dockers to use it as a parent image 
* **Testing Done**  
  The whole set of Apollo test


eyalrund (approved):

WildFireFlum:
My only comment is that we have multiple instances of the same setup (the environment variables) which could maybe be extracted to some configuration file (MINIO_ACCESS_KEY, MINIO_SECRET_KEY, CI) because both apollo and the native tests have the same setup.

PR #2665, opened 2022-06-06
RVBManager: fix a bug in updateRvbDataDuringCheckpoint

glevkovich:
* **Problem Overview** 
RVB failed on assertion while didn't find a block digest during real deployment. The search range given to the RVB during updateRvbDataDuringCheckpoint was incorrect due to multiple pruning commands.

1) Reverse order of RVT update: first prune the blocks, then add blocks.
both are optional of course. The reason is that the whole logic checks
the RVT min/max which should be most updated before adding blocks (
  in rare cases) for the most simplified code. Therefore, we would like
  to first trim the RVT.
2) Calculate add_range_min_block_id while taking into consideration
the pruned_blocks_digests_. We never want to define a range for blocks
that were already pruned.

* **Testing Done***
test bkpCheckpointingWithPruning is added as a unit test to
reproduce the exact scenario for this fix. After the fix is applied,
the test won't fail.

### DIFF
--- a/bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
+++ b/bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
@@ -558,6 +558,9 @@ class BcStTest : public ::testing::Test {
   // Infra member functions
   void compareAppStateblocks(uint64_t minBlockId, uint64_t maxBlockId) const;
 
+  // Target/Product - backup replica, during checkpointing
+  void bkpPrune(uint64_t maxBlockIdToDelete);
+
   // Target/Product ST - destination API & assertions
   void dstStartRunningAndCollecting(FetchingState expectedState = FetchingState::NotFetching);
   void dstStartCollecting();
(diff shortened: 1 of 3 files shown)


bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
```diff
@@ -610,8 +613,8 @@ DataGenerator::DataGenerator(const Config& targetConfig, const TestConfig& testC
  */
 void DataGenerator::generateBlocks(TestAppState& appState, uint64_t fromBlockId, uint64_t toBlockId) {
   std::unique_ptr<char[]> buff = std::make_unique<char[]>(Block::getMaxTotalBlockSize());
-  ConcordAssertEQ(toBlockId % testConfig_.checkpointWindowSize, 0);
-  ConcordAssertGT(fromBlockId, 1);
+  // ConcordAssertEQ(toBlockId % testConfig_.checkpointWindowSize, 0);
```
mdarade, line 616: nit: Is this comment required?

mdarade (approved):
Looks good.
Thanks Gil.

eyalrund (approved):

bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
glevkovich: I should completely remove it, we are not restricted to this condition..
Thanks.

StefanPalashev (approved):
LGTM, thanks @glevkovich. Great way to reproduce the problematic scenario!

PR #2668, opened 2022-06-08
Primary recover with network interruptions

titusen:
Implement a test with constant client load that is running under the Packet Dropping Adversary
with 50% chance of dropping a packet from all replicas and trigger View Changes periodically.
The test should run for 10 minutes and in the end we drop the adversarial packet dropping
activity and verify that we will regain Fast Commit path in the system.

Step by step scenario:
1. Create a packet dropping adversary with 50% chance of dropping any packet.
2. Introduce constant client load.
3. Restart the current primary to trigger View Change.
4. Wait for View Change to complete (because of the packet dropping activity this could take significantly more time than usual and a non-deterministic view increments can happen).
5. Loop to step 3.

After multiple (for example 50) iterations of the above steps we need to stop the packet dropping adversary and wait for Fast Commit Path to be recovered in the system.

### DIFF
--- a/tests/apollo/test_skvbc_restart_recovery.py
+++ b/tests/apollo/test_skvbc_restart_recovery.py
@@ -497,25 +497,23 @@ class SkvbcRestartRecoveryTest(ApolloTest):
         # start replicas
         [bft_network.start_replica(i) for i in bft_network.all_replicas()]
 
-        loop_count = 0
-        log.log_message(f"Loop run: {loop_count}")
+        loop_count_outer = 0
 
         #skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
 
-        while (loop_count < 5):
-            loop_count = loop_count + 1
-            log.log_message(f"Loop run: {loop_count}")
-            with net.PacketDroppingAdversary(bft_network, drop_rate_percentage=50) as adversary:
-                adversary.interfere()
+        while (loop_count_outer < 20):
+            loop_count_outer = loop_count_outer + 1
+            loop_count = 0
+            while (loop_count < 5):
+                loop_count = loop_count + 1
+                log.log_message(f"Loop run: {loop_count}")
+                with net.PacketDroppingAdversary(bft_network, drop_rate_percentage=50) as adversary:
+                    adversary.interfere()
 
-                view = await bft_network.get_current_view()
+                    primary = await bft_network.get_current_primary()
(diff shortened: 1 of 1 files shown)


tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -55,6 +54,7 @@ def _advance_current_next_primary(new_view, num_replicas):
         _next_primary = (_current_primary + 1) % num_replicas
         return _current_primary, _next_primary
 
+    @unittest.skip("Unstable test - BC-17831")
```
titusen, line 57: Will remove these

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -470,6 +470,56 @@ async def test_view_change_with_non_primary_replica_in_state_transfer(self, bft_
...
+            loop_count_outer = loop_count_outer + 1
+            loop_count = 0
+            primary = 0
+            while (loop_count < 5):
+                loop_count = loop_count + 1
+                log.log_message(f"Loop run: {loop_count}")
+                with net.PacketDroppingAdversary(bft_network, drop_rate_percentage=50) as adversary:
+                    adversary.interfere()
+
+                    primary = await bft_network.get_current_primary()
```
HristoStaykov, line 512: I would suggest to move the getting of the primary before we enter the section with the packet dropping adversary. This way we will get it on a stable point and will restart it right after we enter the scope with the packet dropping adversary to make sure View Change happens with certainty.

HristoStaykov (approved):
Thanks @titusen, great work! I only have 2 minor comments to limit the nondeterminism in the scenario and focus it to the place we want to test over a longer running periods.

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -470,6 +470,56 @@ async def test_view_change_with_non_primary_replica_in_state_transfer(self, bft_
...
+                    bft_network.stop_replica(primary)
+                    bft_network.start_replica(primary)
+
+                    await trio.sleep(seconds=20)
+
+                primary = (primary + 1) % bft_network.config.n
+                await self._await_replicas_in_state_transfer(log, bft_network, skvbc, primary)
```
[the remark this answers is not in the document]
titusen: Done

PR #2670, opened 2022-06-09
Fix getBlockStaleKeys method

yontyon:
* **Problem Overview**  
Consider the following BC state:
activeKeys = [key1], Block(x) = [key1], Block(x+1) = [key1], Block(x+2)=[key1]
On an actual pruning, we prune 3 keys:
1. In block(x), key1 is pruned from **active keys** and from **block(x)**.
2. In block (x+1) key1 is pruned from **block(x+1)**
3. In block(x+2) key1 is not stale and hence won't be pruned.

However, when calculating the stale keys, because we don't do the actual pruning, we calculate 5 stale keys:
1. In block(x), key1 is counted as stale from **active keys** and from **block(x)**. 
2. In block(x + 1), key1 is counted as stale from **active keys** and from **block(x + 1)**. 
3. In block(x + 1), key1 is counted as stale from **active keys**.

The solution is to separate the two kinds of keys. 
First, we will calculate only the keys that are staled due to a newer version.
Then, we rotate over the blocks again and count (without duplications) the stale active keys --> each active key may be counted only once.
**Why it is ok and we don't need to actually simulate pruning?**
If during pruning, a key is added to active keys, it means that it doesn't have a newer version. Hence no new version will make it counted as stale in the future.

In addition, I've fixed some minor issues and refactored in kv_blockchain_editor tool
* **Testing Done**  
  CI + new unittest to test this fix

### DIFF
--- a/kvbc/include/categorization/block_merkle_category.h
+++ b/kvbc/include/categorization/block_merkle_category.h
@@ -69,6 +69,7 @@ class BlockMerkleCategory {
                              std::vector<std::optional<TaggedVersion>>& versions) const;
 
   std::vector<std::string> getBlockStaleKeys(BlockId, const BlockMerkleOutput&) const;
+  std::set<std::string> getStaleActiveKeys(BlockId, const BlockMerkleOutput&) const;
   // Delete the given block ID as a genesis one.
   // Precondition: The given block ID must be the genesis one.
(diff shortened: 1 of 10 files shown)


arc-vmware (approved):
Thanks @yontyon , the changes looks great. 
I have only one minor comment.

eyalrund (approved):

PR #2671, opened 2022-06-09
state transfer apollo internal cycle validation test

mdarade:
Internal cycle validation would be done using newly added metric

* **Problem Overview**  
This apollo test was failing intermittently. Fixed it to validate internal cycle with metric counter.
* **Testing Done**  
Ran 50 times locally with configuration as same as CI.

### DIFF
--- a/tests/apollo/test_skvbc_state_transfer.py
+++ b/tests/apollo/test_skvbc_state_transfer.py
@@ -16,6 +16,7 @@ from os import environ
 
 from util.consts import STATE_TRANSFER_WINDOW
 from util.test_base import ApolloTest
+from util import bft_network_partitioning as net
 from util import bft_network_traffic_control as ntc
 import trio
 
@@ -113,8 +114,8 @@ class SkvbcStateTransferTest(ApolloTest):
         await skvbc.assert_successful_put_get()
         await bft_network.force_quorum_including_replica(stale_node)
         await skvbc.assert_successful_put_get()
+    
 
-    # This test should never run with TLS/TCP, only UDP
     @skip_for_tls
     @with_trio
     @with_bft_network(start_replica_cmd,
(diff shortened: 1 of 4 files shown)


tests/apollo/test_skvbc_state_transfer.py
```diff
@@ -113,56 +114,54 @@ async def test_state_transfer_with_multiple_clients(self, bft_network,exchange_k
...
             bft_network.start_replica(stale_node)
-            with trio.fail_after(30):
-                while(True):
-                    fetchingState = await bft_network.get_metric(stale_node, bft_network,
-                                                             "Statuses", "fetching_state", "bc_state_transfer")
-                    if(fetchingState == "GettingMissingBlocks"):
-                        break
-                    await trio.sleep(0.1)
-        await skvbc.send_concurrent_ops_with_isolated_replica(isolated_replica={stale_node}, run_duration=30)
+            adversary.interfere(10)
```
glevkovich, line 148: 1)
Can you please elaborate? I think you should consider adding a comment to explain what you are doing here in this whole scope?
It is not clear when one try to read fast, spending minimal time and avoid digging into the code.
Will you be able to understand in 5 seconds what is done here 6 months from now?

2)
adversary.interfere(10) means? I can't see a single interfere() call in all tests with a number. Why this can even run?

3)
In addition, How can we be sure that replica reaches getting missing blocks when interfere is called?
I understanf that you ran it 50 times, but i'm not sure if it was ran 50 times on remote. In addition, It might fail once in 150 times (for example) in avarage , as the methodology looks opportunistic.

tests/apollo/util/bft.py
```diff
@@ -1162,6 +1162,29 @@ async def force_quorum_including_replica(self, replica_id):
             self.stop_replicas(self.random_set_of_replicas(
                 len(self.procs) - (2 * self.config.f + self.config.c + 1), without={primary, replica_id}))
 
+
+    async def wait_for_getting_missing_blocks_state(self, replica_id):
```
glevkovich, line 1166: What if cycle is ended already?
this is an issue in all out tests I think.
I think that the idea is to have such a long cycle to make the chance for the cycle to end when entering this function == 0.
One of the way to be sure about this is convert cycleCounter_ to a metric called something like "current_cycle_id" and have another metric called last_cycle_id.
Id  current_cycle_id is 1 and last_cycle_id is 0 -> means we are still in the middle of cycle 1. if both are equal - it means that cycle is done.

glevkovich (changes requested):
Hi, I've raise few general comments to better understand why our tests failing sporadically (rare, but still fail sometime).
I think that some of the issues raised relate to other tests as well.

PR #2673, opened 2022-06-11
Fix race conditions in thin_replica_server_test.

upshaw-alex:
This PR is intended to fix a couple race conditions previously observed in the `thin_replica_server_test` unit test suite in cases with names of the form `SubscribeTo*EventGroup*WithGapTwoClients`. The root cause of these race conditions was determined to be that these cases make calls to the test suite's `addMoreEventGroups` helper function in such a way that the addition of these event groups races with the `TestStateMachine` helper object's logic for detecting that all event group updates of interest to the test case have been streamed in order to end the subscription.

This PR makes a couple changes to prevent the `TestStateMachine` from incorrectly detecting the end of a subscription that may be running concurrently with `addMoreEventGroups` calls. First, `addMoreEventGroups`'s existing behavior included alerting the `TestStateMachine` that there are no more event groups to add with a call to `TestStateMachine::toggle_more_event_groups_to_add(false)`. This PR adds a new optional `bool` parameter `even_more_event_groups_to_add_after_these` to `addMoreEventGroups` in order to allow test cases to avoid incorrectly notifying the `TestStateMachine` that there are no more event groups to add in cases where the test case will be making further calls to `addMoreEventGroups` after the current one. Second, this PR shifts the responsibility for calling `TestStateMachine::set_expected_last_event_group_to_send` from `addMoreEventGroups` to the actual test cases that call it so that the expected last event group to be sent can be correctly set strictly before the Thin Replica subscription begins running in order to avoid races between the final expected event group ID getting set correctly and the `TestStateMachine` checking updates the `ThinReplicaImpl` writes against this value.

Additionally, this PR reactivates the affected test cases, as they were previously deactivated to avoid blocking other changes to improve Concord-BFT's and its CI's stability.

**Problem Overview**  
 Previously, test cases with names of the form `SubscribeTo*EventGroup*WithGapTwoClients` in the `thin_replica_server_test` unit test suite were observed to fail intermittently; such failures were noted to occur significantly more frequently for `DEBUG` builds. The root cause of the intermittent failures were effectively race conditions between `TestStateMachine` receiving updates produced by the `ThinReplicaImpl` then trying to check whether all updates relevant to the test case have been received and concurrent calls to `addMoreEventGroups` updating the point at which the current test case ends.

In one mode of failure, the `TestStateMachine` could enter the conditional branch at [`thin_replica_server_test.cpp:574`](https://github.com/vmware/concord-bft/blob/88cb7a23191cd082351c209c133c442f3909e840/thin-replica-server/test/thin_replica_server_test.cpp#L574) early. In this branch, the `TestStateMachine` generates and pushes a new Event Group update with Event Group ID `current_event_group_to_send_ + storage_.total_egs_for_other_clients + 1` to the live update buffers for the active subscription. The condition for entering this branch is `((current_event_group_to_send_ + storage_.total_egs_for_other_clients == (last_event_group_to_send_ - 1)) && not more_event_groups_to_add_)`. If an active subscription is racing with multiple calls to `addMoreEventGroups`, it is possible this condition could become `true` before all calls to `addMoreEventGroups` have completed since (before this PR), _every_ call to `addMoreEventGroups` set `more_event_groups_to_add_` to `false` and updated `last_event_group_to_send_` to reflect only the Event Group(s) the current call is adding. In such cases, two (different) Event Groups with the same global Event Group ID could get pushed to the live update buffers, one by the `TestStateMachine` by entering this branch early, and another from one of the concurrent calls to `addMoreEventGroups` that had not finished at the time of entering this branch; this could ultimately trigger an assertion failure at [thin_replica_impl.hpp:513](https://github.com/vmware/concord-bft/blob/88cb7a23191cd082351c209c133c442f3909e840/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp#L513) when the `ThinReplicaImpl` noticed the duplicated global Event Group ID.

In another mode of failure, the `TestStateMachine` could enter the conditional branch at [thin_replica_server_test.cpp:569](https://github.com/vmware/concord-bft/blob/88cb7a23191cd082351c209c133c442f3909e840/thin-replica-server/test/thin_replica_server_test.cpp#L569) early. In this case, this would cause the `TestStateMachine` to return `false` to a `TestServerWriter::Write` call, causing the current subscription to end early, causing the test case to fail when it checks that the number of updates actually sent matches its expectations. The condition for entering this branch is `(current_event_group_to_send_ + storage_.total_egs_for_other_clients == last_event_group_to_send_)`, and, again, this could be `true` early if the subscription is racing with call(s) to `addMoreEventGroups` since (prior to this PR) `last_event_group_to_send_` would not hold a value accurately reflecting all the Event Groups these call(s) add until after the last such call updates it.

**Testing Done**  
Using a build with my changes for this PR and with `-DCMAKE_BUILD_TYPE=DEBUG`, I ran `sudo make test-single-gtest-case TEST_NAME=thin_replica_server_test TEST_CASE_FILTER=SubscribeTo*EventGroup*WithGapTwoClients NUM_REPEATS=128 BREAK_ON_FAILURE=TRUE` once  and `sudo make test-single-suite TEST_NAME=thin_replica_server_test` 32 times in a loop in `bash`, and saw no failures. I also repeated those tests for a build with my changes for this PR and with `-DCMAKE_BUILD_TYPE=RELEASE`, and also saw no failures. I have not tried any other tests manually, though I would assume these changes should not affect anything other than this particular test suite since the changes are limited to the source file for this test suite, `thin-replica-servertest/thin_replica_server_test.cpp`.


thin-replica-server/test/thin_replica_server_test.cpp
```diff
@@ -574,7 +574,9 @@ class TestStateMachine {
         if ((current_event_group_to_send_ + storage_.total_egs_for_other_clients == (last_event_group_to_send_ - 1)) &&
             not more_event_groups_to_add_) {
           on_finished_dropping_event_groups();
-        } else if (live_buffer_->Full() || live_buffer_->oldestEventGroupId() > (storage_.getLastEventGroupId() + 1)) {
+        } else if (live_buffer_->Full() ||
+                   (!(live_buffer_->EmptyEventGroupQueue()) &&
```
shgandhi, line 578: Thanks for adding this check. I think we should also add a similar check i.e., `live_buffer_->Size() > 0` for legacy events [here](https://github.com/vmware/concord-bft/blob/88cb7a23191cd082351c209c133c442f3909e840/thin-replica-server/test/thin_replica_server_test.cpp#L554). -
`else if (live_buffer_->Full() || live_buffer_->Size() > 0 && live_buffer_->oldestBlockId() > (storage_.getLastBlockId() + 1))`

shgandhi (approved):
Thanks for fixing this issue and the detailed explanation! 💪 
I just have one comment. LGTM.

upshaw-alex:
I have force-pushed a rebase of my commit for this PR off of `master` in the interest of keeping this PR up-to-date.

upshaw-alex:
I have force-pushed an amendment to my commit for this PR that makes revisions in light of @shgandhi's review feedback.

thin-replica-server/test/thin_replica_server_test.cpp
upshaw-alex: I have added this check for the legacy events case in recent revisions to this PR.

teoparvanov (approved):
Hi @upshaw-alex , thank you for the high-quality work, as well as for the detailed PR description, and testing done. Looks great!

PR #2675, opened 2022-06-12
Yf/replace bls

WildFireFlum:
* **Problem Overview**  
We are currently using the relic library to implement threshold signatures using BLS for the slow, fast and fast_with_threshold commit paths.
The BLS threshold implementation is exposed to ReplicaImp via the threshsign library.
This branch adds a new multisig implementation which relies on the EdDSA algorithm, and uses OpenSSL's implementation of EdDSA.
We decided to replace BLS with EdDSA because EdDSA's OpenSSL implementation is better tested and is considered safer.
Both libraries are still available to use via compilation flags (USE_RELIC, USE_EDDSA_OPENSSL).
The new implementation introduces a computational penalty when verifying with a large threshold, which was measured with a new threshsign benchmark added by this branch.
The benchmark uses the opensource picobench (single header) library, which was added by this branch to the repository.
* **Testing Done**  
UnitTests: Tested both the threshsign interface and the OpenSSL interface (need to add these tests again after a refactor) for signature generation and (positive, negative) verification. Tested key generation, serialization and deserialization.
see TestEdDSAMultisig.cpp for the full implementation.


threshsign/src/CMakeLists.txt
```diff
@@ -8,7 +8,12 @@ set(common_source_files
     IThresholdVerifier.cpp
     VectorOfShares.cpp
     ThresholdSignaturesTypes.cpp
-)
+    # TODO: create a separate target
+    eddsa/EdDSAMultisigFactory.cpp
```
toly-kournik: These files are in the eddsa directory, and should be namespaced correspondingly.
No need to duplicate it in the filename, i.e. `eddsa/MultisigFactory.cpp`

threshsign/src/eddsa/EdDSAMultisigFactory.cpp
```diff
@@ -0,0 +1,91 @@
+//
```
toly-kournik, line 2: Use a formal header

threshsign/src/Utils.h
```diff
@@ -22,6 +22,13 @@
 
 #include "XAssert.h"
 
+#ifndef UNUSED
+#define UNUSED(expr) \
+  do {               \
+    (void)(expr);    \
```
toly-kournik: why `(void)expr` isn't enough?

threshsign/test/TestEdDSAMultisig.cpp
```diff
@@ -0,0 +1,183 @@
...
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the
+// LICENSE file.
+
+#include <openssl/evp.h>
+#include <random>
+#include "gtest/gtest.h"
+#include "../include/threshsign/eddsa/EdDSAMultisigFactory.h"
```
toly-kournik, line 16: let's not use relative paths

logging/include/Logging.hpp
```diff
@@ -111,9 +111,9 @@ class Logger {
 
 }  // namespace logging
 
-#define LOG_COMMON(logger, level, s)                              \
-  if ((logger).getLogLevel() <= level) {                          \
-    (logger).print(level, __PRETTY_FUNCTION__) << s << std::endl; \
+#define LOG_COMMON(logger, level, s)                                                                              \
+  if ((logger).getLogLevel() <= level) {                                                                          \
+    (logger).print(level, __PRETTY_FUNCTION__) << s << " | [SQ:" << std::to_string(getSeq()) << "]" << std::endl; \
```
toly-kournik, line 116: It breaks log4cplus formatting configuration, we shouldn't do it hardcoded.

threshsign/src/ThresholdSignaturesTypes.cpp
```diff
@@ -15,11 +15,18 @@
...
+#include "threshsign/IThresholdFactory.h"
 #include "yaml_utils.hpp"
 #include "Logger.hpp"
+
+#ifdef USE_RELIC
+#include "threshsign/bls/relic/BlsThresholdFactory.h"
+#include "threshsign/bls/relic/PublicParametersFactory.h"
+#endif
+
+#ifdef USE_EDDSA_OPENSSL
```
toly-kournik, line 27: `#elseif` ?

toly-kournik: I know, it's on the cmake level, here is a source code level

WildFireFlum: I added a header which hides this #ifdef
The general problem is that there needs either be a file which knows both implementations (and future implementations)
or an inheritance hierarchy for cryptosystem.
I chose to think about cryptosystem as a factory, which might be the wrong thing to do, but it was quickest.

threshsign/bench/BenchmarkThreshsign.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include "include/threshsign/bls/relic/BlsThresholdFactory.h"
+#include "include/threshsign/bls/relic/PublicParametersFactory.h"
+
+static std::unique_ptr<std::vector<uint8_t>> s_data{nullptr};
+
+/**
+ * Initializes randomDataBytes random bytes
+ * @param randomDataBytes
+ */
+void initRandomData(size_t randomDataBytes) {
```
pkthapa, line 44: There is an existing function that generates randomData. `generateRandomData()`.
Can we use it?

WildFireFlum: If we had a library of test utilities then yes. 
Since this is benchmark code and not production code, I didn't think that it matters.

tools/KeyfileIOUtils.cpp
```diff
@@ -20,12 +20,15 @@
...
 #include "KeyfileIOUtils.hpp"
 #include "yaml_utils.hpp"
+#include "util/filesystem.hpp"
 
 void outputReplicaKeyfile(uint16_t numReplicas,
                           uint16_t numRoReplicas,
                           bftEngine::ReplicaConfig& config,
                           const std::string& outputFilename,
                           Cryptosystem* commonSys) {
+  std::cout << "Outputting key file: " << fs::absolute(outputFilename) << std::endl;
```
yontyon, line 30: Adding prints to stdout may break tests that assume a specific output.
I would remove this line.

util/include/memory.hpp
```diff
@@ -1,6 +1,6 @@
 // Concord
 //
-// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2018-2022 VMware, Inc. All Rights Reserved.
```
yontyon, line 3: The name of this file is misleading, I would define its definitions where you use them (as far as I can see it is only in util/include/openssl_crypto.hpp

util/include/string.hpp
```diff
@@ -104,5 +104,9 @@ std::string toString(const Container& container, const char* const separator = "
   return out.str();
 }
 
+inline bool isValidHexString(const std::string& str) {
```
yontyon, line 107: Do you sure we need this? If the e signature is not a valid hex, then verification will fail in any case, right?

util/pyclient/bft_config.py
```diff
@@ -42,8 +42,8 @@ def __init__(self, n: int, f: int = 1, c: int = 0, clients=10):
         self.f : int = f
         self.c : int = c
         self.clients : int = clients
-        assert self.n == 3 * self.f + 2 * self.c + 1, \
-            f"Invariant breached. Expected: n = 3f + 2c + 1"
+        assert self.n >= 3 * self.f + 2 * self.c + 1, \
```
yontyon, line 45: Each replica has to have a role of either f or c.
Otherwise, this replica only increases the burden of the bft protocol without contributing to its resilience. Unless you have a good reason for this change I would remove it

logging/include/Logging.hpp
WildFireFlum: This is not what this "sequence number" is about, it is a thread local counter which allows checking if log messages are missing.

bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -12,6 +12,7 @@
 #include <bftengine/Replica.hpp>
 #include <optional>
 #include <functional>
+#include <bitset>
```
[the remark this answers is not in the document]
WildFireFlum: I changed includes of other files which made this include necessary.

logging/src/Logger.cpp
```diff
@@ -21,11 +21,15 @@ ScopedMdc::~ScopedMdc() { MDC_REMOVE(key_); }
 
 }  // namespace logging
 
+thread_local uint64_t bft_seq = 0;
+uint64_t getSeq() { return bft_seq++; }
```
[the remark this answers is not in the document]
WildFireFlum: See previous comments

util/include/crypto/eddsa/EdDSAVerifier.hpp
```diff
@@ -0,0 +1,52 @@
...
+#pragma once
+#include "EdDSA.h"
+#include "openssl_crypto.hpp"
+
+/**
+ * @tparam PublicKeyType The type of the public key - expected to be a SerializableByteArray
+ *
+ */
+template <typename PublicKeyType>
+class EdDSAVerifier {
```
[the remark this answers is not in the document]
WildFireFlum: This code is to be merged with a version which does so.

util/include/crypto/eddsa/EdDSASigner.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "EdDSA.h"
+#include "openssl_crypto.hpp"
+
+/**
+ *
+ * @tparam PrivateKeyType The type of the private key, expected to be a SerializableByteArray
+ *
+ */
+template <typename PrivateKeyType>
+class EdDSASigner {
```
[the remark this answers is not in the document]
WildFireFlum: This code is to be merged with a version which does so.

tools/KeyfileIOUtils.cpp
WildFireFlum: I added this because key generation is voodoo.
ConcordKeyGenerator is a separate process, who parses its output?

util/include/memory.hpp
WildFireFlum: It is an extension of std::memory, which lets you define unique pointers with custom deleters with relative ease.
It is used only once because it was implemented in this branch, the other crypto related branch will use it as well, and many places in our code can be refactored to use it and save up code lines.
openssl_crypto.hpp is a user of this generic code, Pankaj's branch will likely use it too.

util/include/string.hpp
WildFireFlum: What you're saying does not sound very security oriented.
Regardless, I did not invent the sanity check but rather inherited it :)

util/pyclient/bft_config.py
WildFireFlum: Since this code is used for testing purposes, I don't really care about the resiliency of the network, but rather if 5 replicas have a bug which 6 don't.

thirdparty/include/picobench.h
```diff
@@ -0,0 +1,1455 @@
+// clang-format off
```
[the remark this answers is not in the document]
WildFireFlum: The library had a bug which I fixed manually (it is no longer maintained), 
I still chose to use it in order to not introduce a large dependency.
It is only used for presenting benchmark results..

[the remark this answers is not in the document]
yontyon: One more issue is the license. 
Concordbft is Apache 2 and this picobench is MIT license.
I'm not sure what are the implications, but we should address this issue (maybe @guyg8 can help here)

tools/KeyfileIOUtils.cpp
yontyon: Maybe no one, but we had issues before where other parts of the product parsed the output (maybe config service?).
Hence I would prefer to avoid printing to the stdout unless it is absolutely needed

thirdparty/include/picobench.h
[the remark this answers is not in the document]
WildFireFlum: @eyalrund ?

threshsign/src/ThresholdSignaturesTypes.cpp
```diff
@@ -42,11 +43,13 @@ Cryptosystem::Cryptosystem(const std::string& sysType,
 
 // Helper function to generateNewPseudorandomKeys.
 IThresholdFactory* Cryptosystem::createThresholdFactory() {
-  if (type_ == MULTISIG_BLS_SCHEME || forceMultisig_) {
+  if (type_ == MULTISIG_BLS_SCHEME && forceMultisig_) {
```
toly-kournik, line 47: at the time it was intentionally ||

threshsign/src/ThresholdSignaturesTypes.cpp
```diff
@@ -59,28 +62,32 @@ IThresholdFactory* Cryptosystem::createThresholdFactory() {
 
 void Cryptosystem::generateNewPseudorandomKeys() {
   std::unique_ptr<IThresholdFactory> factory(createThresholdFactory());
+  std::cout << (uint64_t)factory.get() << std::endl;
```
toly-kournik: cout?

CMakeLists.txt
```diff
@@ -42,11 +43,25 @@ option(USE_JSON "Enable use of JSON library" ON)
 option(USE_HTTPLIB "Enable use of httplib" ON)
 option(USE_GRPC "Enable GRPC and Protobuf" ON)
 option(USE_OPENSSL "Enable use of OpenSSL" ON)
+option(USE_EDDSA_OPENSSL "Enable use of EdDSA's OpenSSL implementation for signature generation and verification" ON)
 option(BUILD_THIRDPARTY "Wheter to build third party librarie or use preinstalled ones" OFF)
 option(CODECOVERAGE "Enable Code Coverage Metrics in Clang" OFF)
 option(ENABLE_RESTART_RECOVERY_TESTS "Enable tests for restart recovery" OFF)
 
-if(USE_OPENSSL AND NOT BUILD_THIRDPARTY)
+if((NOT (USE_EDDSA_OPENSSL OR USE_RELIC)) OR (USE_EDDSA_OPENSSL AND USE_RELIC))
```
toly-kournik: maybe put this logic down to the threshold CMakeLists.txt?
there's no mention of threshold in these options, currently we still can use both relic and openssl eddsa instead of rsa

threshsign/include/threshsign/bench/picobench.h
```diff
@@ -0,0 +1,1453 @@
+// picobench v2.00
```
toly-kournik: third party?

toly-kournik (approved):
A couple of minor comments, LGTM, thanks very good job!!!

threshsign/include/threshsign/bench/picobench.h
WildFireFlum: it was moved, i don't know which commit you are looking at

threshsign/src/ThresholdSignaturesTypes.cpp
WildFireFlum: removed a while ago :)

threshsign/src/ThresholdSignaturesTypes.cpp
WildFireFlum: I know, but "multisig" is no longer unique to bls.
Should this still be reverted?

tools/KeyfileIOUtils.cpp
WildFireFlum: I switched over to use the info log.
I tend to disagree, if the code passes CI.

PR #2676, opened 2022-06-12
bcstatetransfer_tests: Add test case bkpRandomCheckpointingWithPruning

glevkovich:
* **Problem Overview**  
This test case is similar to bkpCheckpointingWithPruning, but performs
<numIterations> iterations. Every iteration it decided is blocks are
added, pruned according to a given probability. If blocks are addled
it gets a random number of added blocks. If blocks are pruned, it gets
a random percentage of the total blockchain to be pruned.
After both operations are completed, checkpoint is created,
lastRechableBlockId and genesisBlockId are saved and the next iteration
starts.
* **Testing Done**  
NA


eyalrund (approved):

StefanPalashev (approved):
Looks good, thanks @glevkovich.

PR #2677, opened 2022-06-13
Stop adaptive pruning manager on wedge and start it on unwedge (witho…

yontyon:
…ut restart)

* **Problem Overview**  
 Adaptive pruning is a component that runs in an async thread and continuously sends bft requests. 
It makes sense to stop it on the wedge because we block all consensuses in any case.
We want to start it on unwedge (but not with restart) because if we do activate it with restart as well, we may end up restarting the system while adding blocks which is not recommended and may lead to an unstable situation.
* **Testing Done**  
  CI + fixing specific test that was arbitrarily failing due to this error (test_remove_nodes_with_unwedge)

### DIFF
--- a/kvbc/src/reconfiguration_kvbc_handler.cpp
+++ b/kvbc/src/reconfiguration_kvbc_handler.cpp
@@ -619,6 +619,8 @@ bool ReconfigurationHandler::handle(const concord::messages::WedgeCommand& comma
   auto blockId = persistReconfigurationBlock(
       serialized_command, bft_seq_num, std::string{kvbc::keyTypes::reconfiguration_wedge_key}, ts, false);
   LOG_INFO(getLogger(), "WedgeCommand block is " << blockId);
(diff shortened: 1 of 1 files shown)


kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -710,7 +712,8 @@ bool ReconfigurationHandler::handle(const concord::messages::AddRemoveWithWedgeC
                           std::string(serialized_cmd_data.begin(), serialized_cmd_data.end()));
   }
   auto blockId = persistReconfigurationBlock(ver_updates, sequence_number, ts, true);
-
+  LOG_INFO(getLogger(), "Stoppint adaptive pruning resource manager");
```
nkumar04, line 715: nit: spell check - Stopping

nkumar04 (approved):
LGTM!
Thanks!

eyalrund (approved):

PR #2679, opened 2022-06-14
State Transfer : Destination replica to resume from last-stored-checkpoint instead of last-stored-checkpoint+1

mdarade:
Abruptly restarted destination replica to ask for state from network with last-stored-checkpoint as minimum required instead of earlier last-stored-checkpoint+1.
* **Problem Overview**  
Apollo test validating state transfer by abruptly restarting fetcher and sender nodes was failing intermittently in CI runs. After simplifying apollo test code, it also exposed a bug in fetcher replica where it was asking for last-stored-checkpoint + 1 from stopped consensus network having same last-stored-checkpoint.
* **Testing Done**  
Ran test locally 25 times with config matching with CI runs.
Reproduced issue and validated fix by writing Gtest.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -1215,7 +1215,7 @@ bool BCStateTran::checkStructureOfVirtualBlock(char *virtualBlock,
     char *p = virtualBlock + sizeof(HeaderOfVirtualBlock) + (i * elementSize);
     ElementOfVirtualBlock *e = reinterpret_cast<ElementOfVirtualBlock *>(p);
 
-    if (e->checkpointNumber <= h->lastCheckpointKnownToRequester) {
+    if (e->checkpointNumber < h->lastCheckpointKnownToRequester) {
       LOG_ERROR(logger, KVLOG(e->checkpointNumber, h->lastCheckpointKnownToRequester, i, e->pageId));
       return false;
     }
@@ -1489,7 +1489,10 @@ void BCStateTran::sendAskForCheckpointSummariesMsg() {
   SCOPED_MDC_SEQ_NUM(getScopedMdcStr(config_.myReplicaId, lastMsgSeqNum_));
 
   msg.msgSeqNum = lastMsgSeqNum_;
-  msg.minRelevantCheckpointNum = psd_->getLastStoredCheckpoint() + 1;
+  // Abruptly restarted fetcher replica might have last-stored-checkpoint same as what *stopped* network would have.
+  // In case fetcher asks for last-stored-checkpoint + 1, consensus network may reject such request thinking it doesn't
+  // have requested checkpoint.
+  msg.minRelevantCheckpointNum = psd_->getLastStoredCheckpoint() ? psd_->getLastStoredCheckpoint() : 1;
 
   LOG_INFO(logger_, KVLOG(lastMsgSeqNum_, msg.minRelevantCheckpointNum));
 
@@ -1664,7 +1667,7 @@ bool BCStateTran::onMessage(const CheckpointSummaryMsg *m, uint32_t msgLen, uint
   }
 
   // if msg is not relevant
-  if (m->requestMsgSeqNum != lastMsgSeqNum_ || m->checkpointNum <= psd_->getLastStoredCheckpoint()) {
+  if (m->requestMsgSeqNum != lastMsgSeqNum_ || m->checkpointNum < psd_->getLastStoredCheckpoint()) {
     LOG_WARN(logger_,
              "Msg is irrelevant: " << KVLOG(
                  replicaId, m->requestMsgSeqNum, lastMsgSeqNum_, m->checkpointNum, psd_->getLastStoredCheckpoint()));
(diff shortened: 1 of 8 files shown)


bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -321,6 +321,9 @@ class BCStateTran : public IStateTransfer {
     SendingBatch,
   };
 
+ public:
```
glevkovich, line 324: Please remove the public, should be protected, and use a delegator in any UT.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -1345,6 +1345,11 @@ inline bool BCStateTran::BlocksBatchDesc::isValid() const {
 // State
 //////////////////////////////////////////////////////////////////////////////
 
+void BCStateTran::enableFetchingState(void) {
```
glevkovich, line 1348: move to delegator, we still need to be convinced, that even after you move to delegator - this is a must.
changing target state, is not a good idea usually.

bftengine/src/bcstatetransfer/InMemoryDataStore.cpp
```diff
@@ -88,8 +88,6 @@ std::vector<std::pair<BlockId, Digest>> InMemoryDataStore::getPrunedBlocksDigest
 
 void InMemoryDataStore::setCheckpointDesc(uint64_t checkpoint, const CheckpointDesc& desc) {
   ConcordAssert(checkpoint == desc.checkpointNum);
-  ConcordAssert(descMap.count(checkpoint) == 0);
```
glevkovich, line 91: its better to keep the assert, and find a way from the caller to avoid it from failing:

- new flag - less preferable (flag set to false by default, and if true , jump the assert)
- delete the descripto (zero it?) before the call - better way.

bftengine/src/bcstatetransfer/InMemoryDataStore.cpp
```diff
@@ -237,9 +235,10 @@ void InMemoryDataStore::setResPage(uint32_t inPageId,
                                    const char* inPage) {
   auto lock = std::unique_lock(reservedPagesLock_);
   LOG_DEBUG(logger(), "pageId: " << inPageId << " checkpoint: " << inCheckpoint);
-  // create key, and make sure that we don't already have this element
+  // create key
+  // pages may already contain key when restarted replica asks and gets
+  // last stored checkpoint from network which is also stopped at same checkpoint
   ResPageKey key = {inPageId, inCheckpoint};
-  ConcordAssert(pages.count(key) == 0);
```
glevkovich, line 242: same idea like previous comment. lets try to keep it, and zero the res page from outside, or (less preferable) tell to jump assert using a flag.

glevkovich (approved):

bftengine/src/bcstatetransfer/BCStateTran.hpp
mdarade: Thanks. Addressed.

bftengine/src/bcstatetransfer/InMemoryDataStore.cpp
mdarade: I went with new flag as it was more convenient compared to bringing in whole new API.

bftengine/src/bcstatetransfer/BCStateTran.cpp
glevkovich: you left it here, i dont see it addressed

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -322,6 +322,7 @@ class BCStateTran : public IStateTransfer {
   };
 
  protected:
+  void enableFetchingState(void);
```
glevkovich, line 325: move to unittest?

bftengine/src/bcstatetransfer/InMemoryDataStore.cpp
```diff
@@ -86,10 +86,13 @@ void InMemoryDataStore::setPrunedBlocksDigests(const std::vector<std::pair<Block
...
 std::vector<std::pair<BlockId, Digest>> InMemoryDataStore::getPrunedBlocksDigests() { return prunedBlocksDigests; }
 
-void InMemoryDataStore::setCheckpointDesc(uint64_t checkpoint, const CheckpointDesc& desc) {
+void InMemoryDataStore::setCheckpointDesc(uint64_t checkpoint,
+                                          const CheckpointDesc& desc,
+                                          const bool checkIfAlreadyExists) {
   ConcordAssert(checkpoint == desc.checkpointNum);
-  ConcordAssert(descMap.count(checkpoint) == 0);
-
+  if (checkIfAlreadyExists) {
```
glevkovich, line 93: nit: 
`concordAssertOR(!checkIfAlreadyExists, descMap.count(checkpoint) == 0);`
same for all other similar places.

bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
```diff
@@ -417,6 +417,7 @@ class BcStTestDelegator {
     return stateTransfer_->checkStructureOfVirtualBlock(virtualBlock, virtualBlockSize, pageSize, logger);
   }
   const Digest& computeDefaultRvbDataDigest() const { return stateTransfer_->computeDefaultRvbDataDigest(); }
+  void enableFetchingState(void) { return stateTransfer_->enableFetchingState(); }
```
glevkovich, line 420: I think you didnt understand me.
enableFetchingState should be completely removed from priduct code.
define here:
```

void enableFetchingState(void) { 
   DataStoreTransaction::Guard g(stateTransfer_->psd_->beginTransaction());
  g.txn()->setIsFetchingState(true);
 }
```

glevkovich (approved):

bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
mdarade: Thanks Gil.
I've worked on all your latest comments and updated the PR.

bftengine/src/bcstatetransfer/BCStateTran.hpp
mdarade: Done.

bftengine/src/bcstatetransfer/BCStateTran.cpp
mdarade: I misunderstood last time. I've addressed it now.

eyalrund (approved):

PR #2682, opened 2022-06-16
RangeValidationTree: modify printToLog

glevkovich:
* **Problem Overview**  
RVT may be huge in production and can reach tens of thousands of nodes.
We always need the basic summary to be printed, but at this level,
it doesn't make sense to print the tree structure at all.

1) Always print the tree header (basic info).
2) If verbosity is DETAILED, print structure as well.
3) If verbosity is SUMMARY, or if there are more than 100 nodes -
print only basic header. This overrides section 2.
4) Change all calls to print SUMMARY level only.

* **Testing Done**  
NA


StefanPalashev (approved):
LGTM, thanks @glevkovich.

PR #2683, opened 2022-06-16
BCStateTran: fix wrong state of is_fetching_ metrics

glevkovich:
* **Problem Overview**  
is_fetching_ should be set to 1only if current replica is an active dest.
The way code written now, set to 1 also when current replica is an
active source.

* **Testing Done**  
NA


StefanPalashev (approved):
LGTM, thanks @glevkovich.

eyalrund (approved):

PR #2688, opened 2022-06-19
New KVBC implementation aka v4

cloudnoize:
This implementation of the blockchain storage has the following properties:

1. Simplified structure over the currently categorized implementation.
2. Far fewer column families and by that reduced compaction overhead and complexity.
3. immediate pruning.
4. no change in the storage KVBC interfaces and supporting both categorized and v4 storage by configuration.

**Simplified structure**
 The blockchain is composed of two logical units, the blockchain itself and the latest keys column family i.e the active keys.
the block in v4 storage is a byte array containing the input updates and the prev block digest, contrary to the categorized storage where a block was just a header containing metadata.
the blockchain is a chain of those blocks where the key is the block id and the value is the byte array.
it supports fast retrieval in contrast to the categorized that had to recompose the block.
and pruning is trivial as just the block needs to be deleted.

the other unit is the latest keys column family that holds the state of the blockchain. on each block, its keys are being added to that column family with their version in the postfix value.
if a key is marked for deletion, then it is deleted immediately and not waiting for pruning in contrast to the categorized storage.
updating a key overrides its previous version, which is still accessible by performing a get with an explicit version that gets the value from the blockchain itself (source of truth).
those properties ensure that the latest CF contains all keys that are latest regardless if their block was pruned or not, just as the active keys of the categorized storage but without managing and performing complex bookkeeping while pruning.

**no change in the storage KVBC interfaces and supporting both categorized and v4 storage by configuration**
We went over the code and replaced any concrete usage of storage with the db_interfaces i.e. block adder, reader, delete etc.
a simple hierarchy of an adapter class is used to switch between blockchain implementations.

**recovery**
Recovery is needed when crashing while executing requests preprepare requests or when a DB checkpoint is taken.
Recovery is the process of deleting the last reachable blocks i.e the newly added blocks.
it's done by using RocksDB snapshot at a stable version i.e. the blockchain is at a state where recovery is guaranteed not to be needed.
on recovery, we take the block that we want to delete, for each key we take its previous version and write it into a write batch.
this batch when applied reverts the blockchain to block X-1.
The recovery itself is two-phase operation:
1 - open the blockchain in read-only mode in order to use rocksdb snapshot without losing information that can happen on compaction, open a recovery db instance , for each block that we want to revert, read it's update batch from the readonly instant and write it to the recovery db.
2- open the blockchain in write mode and apply the changes from the first step.

test:
- UT on all functionality
- All Apollo tests are passing as they all use the kvbc layer.


arc-vmware (approved):
LGTM !!!

cloudnoize:
> Hi @cloudnoize - are the Apollo tests in this PR running on the v4 storage layer? If not, we are missing a significant opportunity for regression coverage...
> 
> Thanks for the clarification, Teo

@teoparvanov yes they are running on v4 and some run on both storages

bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -33,6 +33,7 @@
 #include "InternalBFTClient.hpp"
 #include "storage/db_interface.h"
 #include "util/filesystem.hpp"
+#include "kv_types.hpp"
```
toly-kournik, line 36: bft engine shouldn't include kvbc as it creates a cyclic dependency

bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -122,6 +124,7 @@ class DbCheckpointManager {
   void sendInternalCreateDbCheckpointMsg(const SeqNum& seqNum, bool noop);
   BlockId getLastReachableBlock() const;
   SeqNum getLastStableSeqNum() const;
+  void setCheckpointInProcess(bool, concord::kvbc::BlockId) const;
```
toly-kournik, line 127: same

bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -165,6 +168,7 @@ class DbCheckpointManager {
   void cleanUp();
   std::function<BlockId()> getLastBlockIdCb_;
   PrepareCheckpointCallback prepareCheckpointCb_;
+  std::function<void(bool, concord::kvbc::BlockId)> checkpointInProcessCb_;
```
toly-kournik, line 171: same

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -79,6 +83,7 @@ Status DbCheckpointManager::createDbCheckpoint(const CheckpointId& checkPointId,
     for (auto& cb : onDbCheckpointCreated_) {
       if (cb) cb(seqNum);
     }
+    checkpointInProcessCb_(false, lastBlockId);
```
toly-kournik, line 86: same

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -163,11 +168,13 @@ uint64_t DbCheckpointManager::directorySize(const _fs::path& directory, const bo
...
-void DbCheckpointManager::initializeDbCheckpointManager(std::shared_ptr<concord::storage::IDBClient> dbClient,
-                                                        std::shared_ptr<bftEngine::impl::PersistentStorage> p,
-                                                        std::shared_ptr<concordMetrics::Aggregator> aggregator,
-                                                        const std::function<BlockId()>& getLastBlockIdCb,
-                                                        const PrepareCheckpointCallback& prepareCheckpointCb) {
+void DbCheckpointManager::initializeDbCheckpointManager(
+    std::shared_ptr<concord::storage::IDBClient> dbClient,
+    std::shared_ptr<bftEngine::impl::PersistentStorage> p,
+    std::shared_ptr<concordMetrics::Aggregator> aggregator,
+    const std::function<BlockId()>& getLastBlockIdCb,
```
toly-kournik, line 175: same

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -163,11 +168,13 @@ uint64_t DbCheckpointManager::directorySize(const _fs::path& directory, const bo
...
+    const PrepareCheckpointCallback& prepareCheckpointCb,
```
toly-kournik, line 176: same

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -132,7 +133,7 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
           // seq num because checkpoint msg certificate is stored on stable seq num and is used for intergrity
           // check of db snapshots
           const auto& seqNumToCreateSanpshot = createDbChkPtMsg.seqNum;
-          std::optional blockId(DbCheckpointManager::instance().getLastReachableBlock());
+          DbCheckpointManager::instance().setCheckpointInProcess(true, *blockId);
```
toly-kournik, line 136: same

toly-kournik:
@cloudnoize, as discussed offline, there's a code which can become obsolete if dbsnapshotting happens in ROReplica and there will be no need in trimming snapshot blocks which were added during async snapshotting.
In any case there should be no dependencies on kvbc in bftEngine.

bftengine/include/bftengine/DbCheckpointManager.hpp
cloudnoize: thanks for the catch @toly-kournik ! will change it retroactively after the merge

PR #2689, opened 2022-06-20
Titusen/slow fast path alternately

titusen:
The Apollo test, which should be part of the test_skvbc_restart_recovery suite needs to implement the following steps:

Start all Replicas and Introduce constantly running Client requests in the background.
Verify that the system is processing client requests on Fast Path.
Restart the Primary to initiate View Change.
Verify the View Change succeeded.
Wait for the system to recover Fast commit path again (Here we will have to set a time limit for which we expect the system to recover to Fast path. We can start with 1 minute interval.)
Stop 1 Non Primary replica to transition the system to Slow path.
Verify the system is making progress on Slow path.
Restart the Primary to initiate View Change.
Start the Replica we stopped in step 6 and verify the View Change succeeded.
Goto Step 2.

### DIFF
--- a/tests/apollo/test_skvbc_restart_recovery.py
+++ b/tests/apollo/test_skvbc_restart_recovery.py
@@ -329,7 +329,7 @@ class SkvbcRestartRecoveryTest(ApolloTest):
         # start replicas
         [bft_network.start_replica(i) for i in bft_network.all_replicas()]
 
-        log = foo()
+        # log = foo()
 
         loop_count = 0
         while (loop_count < 100):


tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -320,9 +318,72 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
...
+
+            await bft_network.wait_for_replicas_to_reach_at_least_view(replicas_ids=bft_network.all_replicas(), expected_view=view + 1, timeout=60)
+
+            bft_network.start_replica(non_primary)
+
     @with_trio
     @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: c == 0, rotate_keys=True)
     @verify_linearizability()
+    @with_constant_load
     async def test_recovering_of_primary_with_initiated_view_change(self, bft_network, tracker):
```
teoparvanov, line 387: Is this change intentional? Because it will modify another already existing test.

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -320,9 +318,72 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
...
+        loop_count = 0
+        while (loop_count < 1):
+            loop_count = loop_count + 1
+
+            log.log_message("view = await bft_network.get_current_view(")
+            view = await bft_network.get_current_view()
+
+            primary = await bft_network.get_current_primary()
+            bft_network.stop_replica(primary)
+            bft_network.start_replica(primary)
```
HristoStaykov, line 352: After the restart we need the metrics server to be started. In many of the runs I performed locally we get a blocking on **_await_replicas_in_state_transfer** waiting to get status of the replica that was restarted and eventually timeouting. One way to solve it is to introduce a waiting period (5 to 10 seconds should be enough I think). We might want to add a note to correct the getting of the metrics to not get stuck if the metric server on the target replica is not live yet and revisit once this is implemented.

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -120,13 +125,6 @@ async def _restarting_replica_during_system_is_in_view_change(self, bft_network,
         next_primary = 1
         view = 0
 
-        # uncomment for live tracking of log messages from the test
```
HristoStaykov, line 123: It will be good to remove this block from all the tests now that we have extracted it to be general for the entire test suite. This is much better for reducing the noise.

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -43,7 +43,12 @@ def start_replica_cmd(builddir, replica_id):
             "-v", viewChangeTimeoutMilli,
             "-e", str(True)
             ]
+# uncomment for live tracking of log messages from the test
```
HristoStaykov, line 46: I would suggest for us to keep this code block commented and leave the logging output to be handled by eliot by default. I think un-commenting it is only useful for live tracking of the tests, which we do on our machines.

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -320,9 +318,72 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
...
+            log.log_message("wait_for_replicas_to_reach_at_least_view(")
+            await bft_network.wait_for_replicas_to_reach_at_least_view(replicas_ids=bft_network.all_replicas(), expected_view=view + 1, timeout=60)
+
+            log.log_message("_await_replicas_in_state_transfer(")
+            await self._await_replicas_in_state_transfer(log, bft_network, skvbc, (primary + 1) % bft_network.config.n)
+
+            await trio.sleep(seconds=20)
+            log.log_message("wait_for_fast_path_to_be_prevalent(")
+            await bft_network.wait_for_fast_path_to_be_prevalent(
+            run_ops=lambda: skvbc.run_concurrent_ops(num_ops=20, write_weight=1), threshold=40)
```
HristoStaykov, line 363: It is better to keep the threshold equal or less to the number of operations. Since we have constant load it might be slightly more operations that will happen, but this will not cause problem as long as they are on the fast path.

tests/apollo/test_skvbc_restart_recovery.py
titusen: Reverted

tests/apollo/test_skvbc_restart_recovery.py
titusen: Done

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -320,9 +318,72 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
...
+
+            view = await bft_network.get_current_view()
+
+            non_primary = primary
```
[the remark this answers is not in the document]
titusen: In this moment the value of primary is no longer primary.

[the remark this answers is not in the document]
titusen: Do you agree?

tests/apollo/test_skvbc_restart_recovery.py
titusen: Done

tests/apollo/test_skvbc_restart_recovery.py
titusen: done

tests/apollo/test_skvbc_restart_recovery.py
titusen: Done

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -320,6 +308,72 @@ async def test_restarting_f_replicas_for_view_change(self, bft_network):
         await bft_network.wait_for_fast_path_to_be_prevalent(
             run_ops=lambda: skvbc.run_concurrent_ops(num_ops=20, write_weight=1), threshold=20)
 
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: c == 0, rotate_keys=True)
+    @with_constant_load
```
teoparvanov, line 313: As discussed on our 1:1, `@with_constant_load` and `@verify_linearizability` don't play along nicely. The reason is, the `@with_constant_load` decorator doesn't use the tracker to record write requests, and therefore the linearizability check would fail.

teoparvanov: PS: if you need `@with_constant_load` for this test scenarion, let's remove the `@verify_linearizability` decorator for now.

PR #2691, opened 2022-06-20
Release pre-processed batch requests out of the batch lock

yuliasherman:
* **Problem Overview**  
   The locking order of different mutexes has to be preserved to avoid dead-locks. An attempt to fix such an ordering issue 
   resulted in a crash. Introducing an alternative issue resolution.
* **Testing Done**  
  CI

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -177,6 +177,7 @@ void RequestsBatch::releaseReqsAndSendBatchedReplyIfCompleted(PreProcessReplyMsg
   string batchCid;
   atomic_uint32_t batchSize = 0;
   PreProcessBatchReplyMsgSharedPtr batchReplyMsg;
+  list<uint16_t> reqOffsetsInBatch;
   {
     unique_lock<mutex> lock(batchMutex_);
     if (!batchInProcess_) {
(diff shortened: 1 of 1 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -192,14 +193,17 @@ void RequestsBatch::releaseReqsAndSendBatchedReplyIfCompleted(PreProcessReplyMsg
     batchCid = batchCid_;
     batchSize = batchSize_;
     // The last batch request has pre-processed => send batched reply message
-    lock.unlock();
-    for (auto const &replyMsg : repliesList_) {
-      replyMsgsSize += replyMsg->size();
-      preProcessor_.releaseClientPreProcessRequestSafe(clientId_, replyMsg->reqOffsetInBatch(), COMPLETE);
+    for (auto const &reply : repliesList_) {
```
eyalrund, line 196: this is only reverting to the previous solution, right?

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -177,6 +177,7 @@ void RequestsBatch::releaseReqsAndSendBatchedReplyIfCompleted(PreProcessReplyMsg
   string batchCid;
   atomic_uint32_t batchSize = 0;
   PreProcessBatchReplyMsgSharedPtr batchReplyMsg;
+  set<uint16_t> reqOffsetsInBatch;
```
ukg19, line 180: Nit: Can we use another container which has better insertion time complexity ,here e.g. a vector? which gives amortized constant time insertion complexity while set will give logarithmic to the size of set time complexity?

ukg19 (approved):

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: I can change that, but for a few tens of requests, it really does not matter.

eyalrund (approved):

eyalrund:
we should create a private build and give ST to verify. Also run maesto on this fix

PR #2695, opened 2022-06-23
view change with isolated replicas

titusen:
A test where we set tings up so that the next F-1 expected next
primaries will not be able to send messages to their peers, but will be able
to accept messages from them, then trigger View Change. In this way we will
be able to test View Changes with multiple View increments, where the
isolated F-1 expected next primaries will not be able to step in as
primaries, but will activate the corresponding view for which it is
theirs turn to become Primary.

Step by step scenario:
1. Use a one way isolating adversary to isolate the F-1 replicas after the current primary in such a way that they cannot send messages to the peers, but can receive messages from them.
2. Stop the current primary.
3. Send Client requests to trigger a View Change.
4. Wait for the system to finish View Change. Note that multiple View increments will happen.
5. Drop the network adversary and verify Fast Commit Path is recovered in the system by introducing client requests.

We can perform this test in a loop multiple times.

### DIFF
--- a/tests/apollo/test_skvbc_restart_recovery.py
+++ b/tests/apollo/test_skvbc_restart_recovery.py
@@ -597,7 +597,7 @@ class SkvbcRestartRecoveryTest(ApolloTest):
 
             view = await bft_network.get_current_view()
 
-            with net.ReplicaOneWayTwoSubsetsIsolatingAdversary(bft_network, [], replicas_to_isolate) as adversary:
+            with net.ReplicaOneWayTwoSubsetsIsolatingAdversary(bft_network, other_replicas, replicas_to_isolate) as adversary:
                 adversary.interfere()
 
                 bft_network.stop_replica(primary)
--- a/tests/apollo/test_skvbc_restart_recovery.py
+++ b/tests/apollo/test_skvbc_restart_recovery.py
@@ -575,7 +575,7 @@ class SkvbcRestartRecoveryTest(ApolloTest):
 
         skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
         loop_count = 0
-        while (loop_count < 5):
+        while (loop_count < 100):
             loop_count = loop_count + 1
 
             primary = await bft_network.get_current_primary()
--- a/tests/apollo/test_skvbc_restart_recovery.py
+++ b/tests/apollo/test_skvbc_restart_recovery.py
@@ -549,6 +549,60 @@ class SkvbcRestartRecoveryTest(ApolloTest):
                 run_ops=lambda: skvbc.run_concurrent_ops(num_ops=20, write_weight=1), threshold=20)
             log.log_message("fast path prevailed")
 
+    @with_trio
+    @with_bft_network(start_replica_cmd, selected_configs=lambda n, f, c: c == 0, rotate_keys=True)
+    @verify_linearizability()
+    async def test_view_change_with_isolated_replicas(self, bft_network, tracker):
+        """
+        test View Changes with multiple View increments, where the
+        isolated F-1 expected next primaries will not be able to step in as
+        primaries, but will activate the corresponding view for which it is
+        theirs turn to become Primary.
+
+        Step by step scenario:
+        1. Use a one way isolating adversary to isolate the F-1 replicas after the current primary in such a way that they cannot send messages to the peers, but can receive messages from them.
+        2. Stop the current primary.
+        3. Send Client requests to trigger a View Change.
+        4. Wait for the system to finish View Change. Note that multiple View increments will happen.
+        5. Drop the network adversary and verify Fast Commit Path is recovered in the system by introducing client requests.
+
+        We can perform this test in a loop multiple times.
+        """
+
+        # start replicas
+        [bft_network.start_replica(i) for i in bft_network.all_replicas()]
+
+        skvbc = kvbc.SimpleKVBCProtocol(bft_network, tracker)
+        loop_count = 0
+        while (loop_count < 5):
+            loop_count = loop_count + 1
+
+            primary = await bft_network.get_current_primary()
+
+            replicas_to_isolate = random.sample(
+                bft_network.all_replicas(without={primary}), bft_network.config.f - 1)
+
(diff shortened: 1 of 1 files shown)


teoparvanov (changes requested):
There seems to be some mix-up with the commits probably... Why are there two new tests in this PR? Which one should I review?

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -549,6 +549,61 @@ async def test_view_change_with_non_primary_replica_in_state_transfer(self, bft_
...
+                for r2 in replicas_to_isolate:
+                    if (r1 == r2):
+                        append = False
+                        break
+                if (append):
+                    other_replicas.append(r1)
+
+            view = await bft_network.get_current_view()
+
+            with net.ReplicaOneWayTwoSubsetsIsolatingAdversary(bft_network, [], replicas_to_isolate) as adversary:
```
HristoStaykov, line 598: I think the ReplicaOneWayTwoSubsetsIsolatingAdversary takes as parameters 2 subsets - the first being the blocked receivers and the second - the blocked senders. Why is the set of blocked receivers empty here?

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -549,6 +549,61 @@ async def test_view_change_with_non_primary_replica_in_state_transfer(self, bft_
...
+                adversary.interfere()
+
+                await skvbc.run_concurrent_ops(10)
+                await bft_network.wait_for_replicas_to_reach_at_least_view(other_replicas, expected_view=view + 1, timeout=60)
```
HristoStaykov, line 602: If we have set things up correctly, we should perform a single View Change with F View increments. Therefore the expected View should be **view + F** if I am not missing something?

tests/apollo/test_skvbc_restart_recovery.py
titusen: "Use a one way isolating adversary to isolate the F-1 replicas after the current primary in such a way that they cannot send messages to the peers, but can receive messages from them."

teoparvanov: @HristoStaykov do you agree with @titusen's approach here?

HristoStaykov: Our goal of using the **ReplicaOneWayTwoSubsetsIsolatingAdversary** is to make it impossible for the next F primaries to step in as such, therefore we should get 1 View Change with F view increments. We want them to be able to receive messages from the peers, but be unable to send to the peers. In this way neither of the F next primaries will be able to send his NewView message to the peers.

titusen: So is it correct now? If no, what's wrong?

titusen: Resolving this conversation, because the matter has been resolved vis zoom.

HristoStaykov (approved):
LGTM, thanks @titusen for adding this test case!

tests/apollo/test_skvbc_restart_recovery.py
```diff
@@ -549,6 +549,59 @@ async def test_view_change_with_non_primary_replica_in_state_transfer(self, bft_
...
+
+            view = await bft_network.get_current_view()
+
+            with net.ReplicaOneWayTwoSubsetsIsolatingAdversary(bft_network, other_replicas, replicas_to_isolate) as adversary:
+                adversary.interfere()
+
+                bft_network.stop_replica(primary)
+                await skvbc.run_concurrent_ops(10)
+
+                await bft_network.wait_for_replicas_to_reach_at_least_view(other_replicas, expected_view=view + bft_network.config.f, timeout=60)
```
teoparvanov, line 598: Hi @titusen, did you double-check in the logs that the expected view is actually reached?

titusen: Yes, I did. It happens. 
0|28-06-2022 13:36:51.651|INFO |concord.bft.viewchange|message-processing|||2||tryToEnterView|L:695|View is activated 2previous view was 0
0|28-06-2022 13:37:12.320|INFO |concord.bft.viewchange|message-processing|||3||tryToEnterView|L:695|View is activated 3previous view was 2
0|28-06-2022 13:37:30.204|INFO |concord.bft.viewchange|message-processing|||4||tryToEnterView|L:695|View is activated 4previous view was 3
0|28-06-2022 13:37:57.901|INFO |concord.bft.viewchange|message-processing|||6||tryToEnterView|L:695|View is activated 6previous view was 4
0|28-06-2022 13:38:16.825|INFO |concord.bft.viewchange|message-processing|||7||tryToEnterView|L:695|View is activated 7previous view was 6
0|28-06-2022 13:38:23.701|INFO |concord.bft.viewchange|message-processing|||8||tryToEnterView|L:695|View is activated 8previous view was 7
0|28-06-2022 13:38:46.449|INFO |concord.bft.viewchange|message-processing|||9||tryToEnterView|L:695|View is activated 9previous view was 8
0|28-06-2022 13:39:10.857|INFO |concord.bft.viewchange|message-processing|||11||tryToEnterView|L:695|View is activated 11previous view was 9
0|28-06-2022 13:39:32.512|INFO |concord.bft.viewchange|message-processing|||12||tryToEnterView|L:695|View is activated 12previous view was 11
0|28-06-2022 13:39:55.685|INFO |concord.bft.viewchange|message-processing|||13||tryToEnterView|L:695|View is activated 13previous view was 12
0|28-06-2022 13:40:14.827|INFO |concord.bft.viewchange|message-processing|||14||tryToEnterView|L:695|View is activated 14previous view was 13
0|28-06-2022 13:40:21.376|INFO |concord.bft.viewchange|message-processing|||15||tryToEnterView|L:695|View is activated 15previous view was 14

teoparvanov: Awesome, thanks for the detailed analysis! It also helps you better understand the system's internals.

teoparvanov (approved):
Good to go, thank you @titusen !

PR #2696, opened 2022-06-23
St fix message free bug during continue send batch

glevkovich:
* **Problem Overview**  
 core dump observed in BCStateTran::continueSendBatch due to rare FetchBlocksMsg rejection. During rejection a buffer is being freed with the wrong function call. ItemDataMsg::free should be called instead of freeStateTransferMsg.
* **Testing Done**  
There is no way to reproduce this, as there are no logs and I have no idea what is the order of events that caused such rejection. I've added a test that test for similar rejection.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -2103,7 +2103,7 @@ void BCStateTran::continueSendBatch() {
       if ((rvbGroupDigestsActualSize == 0) || (sb.rvbGroupDigestsExpectedSize != rvbGroupDigestsActualSize)) {
         auto additionalInfo = "Rejecting message - not holding all requested digests (or some other error)" +
                               KVLOG(sb.rvbGroupDigestsExpectedSize, rvbGroupDigestsActualSize);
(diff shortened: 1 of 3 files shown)


eyalrund (approved):

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -2103,7 +2103,7 @@ void BCStateTran::continueSendBatch() {
       if ((rvbGroupDigestsActualSize == 0) || (sb.rvbGroupDigestsExpectedSize != rvbGroupDigestsActualSize)) {
         auto additionalInfo = "Rejecting message - not holding all requested digests (or some other error)" +
```
mdarade, line 2104: Should getSerializedDigestsOfRvbGroup() be improved to log all possible cases?

mdarade (approved):
Looks good to me.
Thanks Gil.

bftengine/src/bcstatetransfer/BCStateTran.cpp
glevkovich: Sorry, missed this one. Anyway it wouldn't have been fied in this PR.
I will set up a memo on slack to discuss this next week, might create a ticket for it.

PR #2697, opened 2022-06-23
Enable delay strategy in state transfer apollo test 

mdarade:
Enable delay strategy for fetcher-sender-crash test so that
test gets enough time to catch replica in state transfer.

* **Problem Overview**  
State transfer test crashing fetcher and sender node was sometimes failing to catch them in state transfer. We have now increased number of checkpoints and also delaying messages so that test gets enough time to crash both the nodes.
* **Testing Done**  
Ran tests locally 50+ times matching CI config.

### DIFF
--- a/tests/apollo/test_skvbc_checkpoints.py
+++ b/tests/apollo/test_skvbc_checkpoints.py
@@ -459,7 +459,6 @@ class SkvbcCheckpointTest(ApolloTest):
                 assert number_of_checkpoint_mismatch == n, \
                     f"Replica {replica_id} number_of_checkpoint_mismatch={number_of_checkpoint_mismatch} != {n}"
 
-    @unittest.skip("Unstable test")
     @with_trio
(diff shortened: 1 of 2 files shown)


tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -459,7 +459,6 @@ async def test_rvt_conflict_detection_after_corrupting_checkpoint_msg_for_f_plus
                 assert number_of_checkpoint_mismatch == n, \
                     f"Replica {replica_id} number_of_checkpoint_mismatch={number_of_checkpoint_mismatch} != {n}"
 
-    @unittest.skip("Unstable test")
```
glevkovich, line 462: Why did you enable only a single test?
Maybe I misunderstood you.
We have 4 tests disabled by @StefanPalashev  recently. See:
60e1b8a716f8a8553367059ee4a9de2dccce7dc6
"Disable unstable checkpoint tests"

glevkovich (changes requested):

glevkovich (approved):
LGTM

eyalrund (approved):

PR #2698, opened 2022-06-26
Makefile, install_deps: Introduce Cppcheck in concord-bft dev image

glevkovich:
Introduce Cppcheck in concord-bft dev image

Cppcheck (2.8), is a static analaysis tool:
https://cppcheck.sourceforge.io/
https://github.com/danmar/cppcheck
https://en.wikipedia.org/wiki/Cppcheck

"Cppcheck is a static code analysis tool for the C and C++ programming
languages. It is a versatile tool that can check non-standard code.[2]
The creator and lead developer is Daniel Marjamäki. Cppcheck is free
software under the GNU General Public License."

Clang-tidy is a more powerful tool. On the other hand, there is quite
a wide consensus that using 2-3 tools for static analysis is almost
always better, as different issues may arise. Cppcheck is one of the
oldest and most widely used open-source tools for c++ static analysis
(I would say it is among the top 5).

At this stage, Cppcheck is not enforced in CI. It is part of the image,
with version 2.8 installed, It can be invoked natively or dockerize.
For simplicity, I've added a Makefile command 'make cppcheck'
which allows for a manual check of the whole source code under
concord-bft, or part of it by defining a relative path or file(s).

process_cppcheck_results.sh is the actual script that was invoked from
Makefile. One should be able to run the script also from the command line as
long as there is an report file under the target path (script's
input).


glevkovich:
Comment: image 0.43 is not yet pushed to remote. It will be pushed after approval and. beforem merge.

toly-kournik:
We'd like to run cppcheck not via docker as well

.cppcheck/includes.txt
```diff
@@ -0,0 +1,92 @@
+/concord-bft/
```
toly-kournik, line 1: Can't we calculate the work dirs automatically?
We'll forget to update this when additional directories are added.
Isn't it better to use excludes.txt?

install_deps.sh
```diff
@@ -363,6 +363,19 @@ git clone -b v0.7.0 --depth 1 https://github.com/jaegertracing/jaeger-client-cpp
     cd ${HOME} && \
     rm -r jaeger-client-cpp
 
+# Cppcheck 2.8
```
toly-kournik, line 366: apt-get install?

.cppcheck/includes.txt
glevkovich: You are right, I was trying to do so and wrote a script which created an include file according to all hpp/h files. I removed it later, since it didnt work well - as many includes are relative, e.g #include "categorization/db_categories.h"

The way to know if includes are missing is to run --check-config which runs a special run and give comments on all configuration issues, some of them are includes. Tha'ts what I did.
To have an auto discovering script will require much more work (dig into CMakeLists.txt files). I might try to do it on a 2nd iteration. Anyhow, Cppcheck will never fail if include is not find - it just prints an error.

install_deps.sh
glevkovich: apt-get install will install cppCheck v 1.4 which is very old and doesn't even support c++17.
The only way to install the newest version is via ppa or building from the official download site sourceforge.
I havn't found a ppa.

glevkovich:
`We'd like to run cppcheck not via docker as well`
@toly-kournik, this is a compleltely new requierment, as the Makefile supports only dockerized commands, except for build-docker-image.
If there is a need for that, we should open a new ticket to create another Makefile, or refactor this Makefile to support calls out of container for all comamnds, assuming that all packages are installed in host.

WildFireFlum:
> `We'd like to run cppcheck not via docker as well` @toly-kournik, this is a compleltely new requirement, as the Makefile supports only dockerized commands, except for build-docker-image. If there is a need for that, we should open a new ticket to create another Makefile, or refactor this Makefile to support calls out of container for all commands, assuming that all packages are installed in host.

You can easily mimic the install_deps approach: write a script and call it from the makefile..
It probably won't work on windows, but supporting linux/mac should be easy enough to implement.

toly-kournik:
> `We'd like to run cppcheck not via docker as well` @toly-kournik, this is a compleltely new requirement, as the Makefile supports only dockerized commands, except for build-docker-image. If there is a need for that, we should open a new ticket to create another Makefile, or refactor this Makefile to support calls out of container for all commands, assuming that all packages are installed in host.

There was no requirement for cppcheck - it's your initiative, so please make it work for native linux build.
The generic way of doing it is to create a cmake target and if needed run it from a dockerized build.

.cppcheck/includes.txt
WildFireFlum: Doesn't this utility support using cmake's compilation database like clang-tidy?
https://cppcheck.sourceforge.io/manual.pdf
see chapter 4.2.

Makefile
```diff
@@ -208,6 +208,44 @@ tidy-check: gen_cmake ## Run clang-tidy
...
+	@if [ ! -f ${REL_PATH} ] && [ ! -d ${REL_PATH} ]; then echo 'Error: REL_PATH=${REL_PATH} is not a valid file/path!'; exit 1; fi;
+	docker run ${CONCORD_BFT_USER_GROUP} ${BASIC_RUN_PARAMS} \
+		${CONCORD_BFT_CONTAINER_SHELL} -c \
+		"set -eo pipefail && \
+		cd ${CONCORD_BFT_BUILD_DIR} && \
+		rm -rf cppcheck/reports/${REL_PATH}/ && \
+		mkdir -p cppcheck/reports/${REL_PATH}/ && \
+		cppcheck  --template='cppcheck1'  --inconclusive --enable=all --platform=unix64 --std=c++17 -j ${NUM_JOBS} --force --quiet \
```
WildFireFlum, line 226: Please export to a bash script and add comments explaining what you were doing and why.  This is not an easy read.

glevkovich:
```
We'd like to run cppcheck not via docker as well @toly-kournik, this is a compleltely new requirement, as the Makefile supports only dockerized commands, except for build-docker-image. If there is a need for that, we should open a new ticket to create another Makefile, or refactor this Makefile to support calls out of container for all commands, assuming that all packages are installed in host.

There was no requirement for cppcheck - it's your initiative, so please make it work for native linux build.
The generic way of doing it is to create a cmake target and if needed run it from a dockerized build.
Our Makefile is a hack in a cmake build environment. All Makefiles should be generated by cmake.
```
@toly-kournik , do we support clang-tidy in native linux?

glevkovich:
```
We'd like to run cppcheck not via docker as well @toly-kournik, this is a compleltely new requirement, as the Makefile supports only dockerized commands, except for build-docker-image. If there is a need for that, we should open a new ticket to create another Makefile, or refactor this Makefile to support calls out of container for all commands, assuming that all packages are installed in host.

You can easily mimic the install_deps approach: write a script and call it from the makefile..
It probably won't work on windows, but supporting linux/mac should be easy enough to implement.
```

@WildFireFlum , I think I know how to do it, my idea is that many other Makefile commands are not running in native Linux, so a support for native Linux should have been done as part of a seperate task.
I don't mind adding this one as the 1st one to support native Linux, but as a side comment-  current Makefile is used mostly for dockerized runs/checks. I will support Cppcheck only, for a start.

toly-kournik:
https://stackoverflow.com/questions/48625499/cppcheck-support-in-cmake

> ```
> We'd like to run cppcheck not via docker as well @toly-kournik, this is a compleltely new requirement, as the Makefile supports only dockerized commands, except for build-docker-image. If there is a need for that, we should open a new ticket to create another Makefile, or refactor this Makefile to support calls out of container for all commands, assuming that all packages are installed in host.
> 
> There was no requirement for cppcheck - it's your initiative, so please make it work for native linux build.
> The generic way of doing it is to create a cmake target and if needed run it from a dockerized build.
> Our Makefile is a hack in a cmake build environment. All Makefiles should be generated by cmake.
> ```
> 
> @toly-kournik , do we support clang-tidy in native linux?

neither clang-tidy nor cppcheck should be part of the top-level manually edited Makefile - 
cmake has an integrated support for clang-tidy and for cppcheck. please refer to :
https://stackoverflow.com/questions/48625499/cppcheck-support-in-cmake

toly-kournik (changes requested):
Please use cmake integrated support for cppcheck. This will also make the use of the manually edited .cppcheck/includes.txt obsolete

WildFireFlum:
> https://stackoverflow.com/questions/48625499/cppcheck-support-in-cmake

https://ortogonal.github.io/cmake-clang-tidy/
This is an example of how to use clang tidy via cmake (if you are doing that)

.cppcheck/includes.txt
glevkovich: That's a good point. For that, we will have to define target  project folder as well (I need to do more research).
I don't want to loose the other options, since they allow checking single files + wildecards and folders, not part of a project, which are important for manual checks.
I've already changed to this option as the main one. There is one thing that compicates thing a bit, project import does not support ignore files, so I have to generate all messages/cmf and protobufs (1-2 minute of build, if you havn't built already).I will see if the includes are needed - i'm not sure they are needed anymore. thanks.

glevkovich:
```
https://stackoverflow.com/questions/48625499/cppcheck-support-in-cmake

https://ortogonal.github.io/cmake-clang-tidy/
This is an example of how to use clang tidy via cmake (if you are doing that)
```
@WildFireFlum , thanks - I've already manged to do it. i'll take a look at the example to see if there is something done better.

glevkovich:
@toly-kournik @WildFireFlum , just to summarize the above:
I have no issue with the claim that we should support native Linux, as this is stated also in our open source repository. But, the current design of the Makefile to support only dockerized calls creates a status where only Cppcheck is going to be supported after my merge. There is a gap between "what is needed"/ "what should have been done" to "what we have now", that's all I'm saying - and that's not the center of this new feature, as all recent Makefile commands supported only dockerized calls. If the gap is too large I'll have to postpone this task for another iteration. As of now, it looks like it's possible to do it and there is no need to postpone.

I'm going to add the 2 suggested changes:
1) support CMake project generation and import (and probably remove the inclues file)
2) support native Linux run -  I am going to assume that when someone is using native Linux, Cppcheck is already installed. All I'm going to do in the Makefile is to check such install exists + required version.

glevkovich:
@WildFireFlum , an update - adding the compile commands still requires having an includes file. For some considirations, the system includes path is not part of the compile command JSON file and also not part of the Cppcheck default search paths, and therefore some 3rd party library are not found if not explicitly included. (I assume that the same case happen in clang-tidy which also relies on the same JSON file). In addition, some of the relative includes in hpp files are not in the compile commands file as well, and part of the system path search of clang. I will have to add an includes file anyhow.

Makefile
glevkovich: I will.

toly-kournik:
@glevkovich, haven't we agreed to use a standard built in `CMAKE_CXX_CPPCHECK` flag for this check?

glevkovich:
> @glevkovich, haven't we agreed to use a standard built in `CMAKE_CXX_CPPCHECK` flag for this check?

@toly-kournik I've written to you in the slack to explain my choice. This way is insufficient, as the script allows for more flexibility. So after I've wrriten the script, I tihnk that having this another flag force us to support 2 ways (double work to maintain) - script and vs cmake, while the later can run (without the script) only in a basic form.

scripts/run-cppcheck.sh
```diff
@@ -0,0 +1,217 @@
+#!/usr/bin/env bash
+
+# This script runs Cppcheck.
+# Prerequisites:
+# 1) Compile commands JSON file is under ${CONCORD_BFT_BUILD_DIR}/compile_commands.json.
+#    It can be generated by calling 'cmake gen-compile-commands'.
```
toly-kournik, line 6: `make gen-compile-commands` , meaning you cannot run natively without invoking dockerixed make

.cppcheck/includes.txt
toly-kournik, line 1: whether you using cmake -DCMAKE_CXX_CPPCHECK or -DCMAKE_EXPORT_COMPILE_COMMANDS you don't want to maintain this file

glevkovich: The exported file compile_commands.json is limited, it does not point to all includes, some will miss.
It doesn't recognize system and 3rd party includes and some includes will be missing, we get warning by cppcheck pre processor, resuting in a non optimized analaysis. So this include file is more like a hint.
For now, I didn't find any better solution. Omitting this file will result in less optimizaed results.
Also, this file is mandatory if I want to call to for checks like this:
```
make cppcheck CPPCHECK_REL_TARGET_PATH=make cppcheck CPPCHECK_REL_TARGET_PATH=bftengine/src/preprocessor/
```
to check only preprocessor, or
```
make cppcheck CPPCHECK_REL_TARGET_PATH=bftengine/src/preprocessor/PreProcessor.cpp
```
to check a single file
or
```
make cppcheck CPPCHECK_REL_TARGET_PATH=bftengine/src/preprocessor/PreProcessor*
```
to check all files which start with 'PreProcessor'.
In all of these examples I cant use compile comands and I must use the includes.

I agree that solution is not perfect, but cppcheck itself isn't perfect...
BTW, in clang-tidy i think we are not even importing compile comands, it uses a linux search ...

toly-kournik:
And now compare it to the following short and effective solution:
```
scripts/CMakeLists.txt:

if(CPPCHECK)
    find_program(cppcheck cppcheck HINTS "/usr/local/bin/cppcheck" REQUIRED)
    message(STATUS "cppcheck ${cppcheck}")
    if(cppcheck MATCHES "NOTFOUND")
        message(FATAL_ERROR "failed to find cppcheck executable for CPPCHECK option")
    endif()
    set(CXX_CPPCHECK "${cppcheck}"
            "--enable=all"
            "--inconclusive"
            "--inline-suppr"
            "--quiet"
            "--suppress=unmatchedSuppression"
            "--suppress=unusedFunction"
            "--template='{file}:{line}: warning: {id} ({severity}): {message}'"
            "--output-file=${PROJECT_BINARY_DIR}/cppcheck.out" CACHE STRING "cppcheck")
    endif()
endif(CPPCHECK)
```
we can play with options passed to cppcheck here.

For the native build:
`cmake -DCPPCHECK=ON ..; make`
For the dockerized build run the same commands inside the container

glevkovich:
`we can play with options passed to cppcheck here.`

@toly-kournik ,  I wanted to add choices also to the cmake option but havn't pushed yet.
 I agree it is better, as described in next link + your example:
https://hopstorawpointers.blogspot.com/2018/11/integrating-cppcheck-and-cmake.html?m=1
I'll improve the cmake file options.
Thanks.

WildFireFlum:
> `we can play with options passed to cppcheck here.`
> 
> @toly-kournik , I wanted to add choices also to the cmake option but havn't pushed yet. I agree this one is good as well, as described in next link + your example: https://hopstorawpointers.blogspot.com/2018/11/integrating-cppcheck-and-cmake.html?m=1 I'll improve the script calling options. But the thing is that I'm not convinced its a good idea to maintain both. I'll see what I can do.
> 
> My idea is this example is very basic, and doesn't support multiple things supported in the script. It runs on the whole source code, it output different results and it handles different things. The script is supposed to make things simpler.

I tend to agree with Toly's approach, I think that if the cmake works, there is no need for an additional script.

glevkovich:
```
we can play with options passed to cppcheck here.

@toly-kournik , I wanted to add choices also to the cmake option but havn't pushed yet. I agree this one is good as well, as described in next link + your example: https://hopstorawpointers.blogspot.com/2018/11/integrating-cppcheck-and-cmake.html?m=1 I'll improve the script calling options. But the thing is that I'm not convinced its a good idea to maintain both. I'll see what I can do.

My idea is this example is very basic, and doesn't support multiple things supported in the script. It runs on the whole source code, it output different results and it handles different things. The script is supposed to make things simpler.

I tend to agree with Toly's approach, I think that if the cmake works, there is no need for an additional script.
```
The script and the current solution are not the same. I had a similar solution in CMake and removed it, since it is less flexible and user friendly.  The script is longer since it is doing many more things, include post call checks and reports.
At it's core, it is doing the same as the above, running cppcheck, but it's checking, validating and configuring for multiple options and scenarios. I assume that most of it can be converted to CMake, but it will requires days of work.

toly-kournik:
There's no need to maintain anything

glevkovich:
> There's no need to maintain anything

I think we need to zoom, our misunderstanding / disagreement won't be solved here...

eyalrund:
lets schedule a meeting on Sunday.

CMakeLists.txt
```diff
@@ -114,6 +117,41 @@ if(CODECOVERAGE)
...
+    # concord-bft/build > make bcstatetransfer_tests 2> cppcheck/reports/full_report.txt
+    # concord-bft/build > scripts/run-cppcheck.sh -p
+    set(CMAKE_CXX_CPPCHECK
+          "${cppcheck}"
+          "--enable=all"
+          "--inconclusive"
+          "--inline-suppr"
+          "--quiet"
+          "--std=c++17"
+          "--platform=unix64"
```
toly-kournik, line 142: Since we aim to a multiplatform project, I wouldn't mention a platform here

CMakeLists.txt
```diff
@@ -114,6 +117,41 @@ if(CODECOVERAGE)
...
+          "--template=cppcheck1"
+          "--max-configs=1"
+          "--library=boost.cfg"
+          "--library=openssl.cfg"
+          "--library=googletest"
+          "--addon=threadsafety.py"
+          "--cppcheck-build-dir=${PROJECT_BINARY_DIR}/cppcheck/"
+          "--suppressions-list=${CMAKE_CURRENT_SOURCE_DIR}/.cppcheck/suppressions.txt"
+          "--exitcode-suppressions=${CMAKE_CURRENT_SOURCE_DIR}/.cppcheck/exitcode-suppressions.txt"
```
toly-kournik, line 151: we need to add `--error-exitcode=1` to enable compilation failure on cppcheck errors

CMakeLists.txt
```diff
@@ -114,6 +117,41 @@ if(CODECOVERAGE)
...
+    if(cppcheck MATCHES "NOTFOUND")
+      message(FATAL_ERROR "failed to find cppcheck executable for CPPCHECK option")
+    endif()
+    # Create <cppcheck> work folder for whole program analysis, for faster analysis and to store some useful debug information
+    # Add cppcheck work folder and reports folder for cppcheck output.
+    file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/cppcheck/reports/)
+    # max number of threads = number of CPUs
+    include(ProcessorCount)
+    ProcessorCount(CPU_CORES)
+    # Report output is sent to stderr. To have the output in a single report file, run run-cppcheck.sh only to analyse results, for example:
```
toly-kournik, line 132: This comment is confusing here, please move it to the readme of run-cppcheck.sh

CMakeLists.txt
glevkovich: removed here..keeping in script

CMakeLists.txt
glevkovich: Assuming that there is 1 error in every file in a target, and there are N files, this requires 
3........N builds + fixes between.
As of now I think it will not be the best way , since we dont want to enable CI yet, so you cant get the big picture on all issues by looking at all report. I think for now we output all to full_report.txt (see confluence) and decide status there. See confluence page.

CMakeLists.txt
glevkovich: will remove it

.cppcheck/includes.txt
WildFireFlum: @glevkovich Does not using this file affects the tool's results when run via cmake? If so, please document that the script produces a more verbose result. If not, please document that the script is equivalent to running via cmake, but doesn't require compilation.

WildFireFlum (approved):

PR #2700, opened 2022-06-28
Fix test_state_transfer_rvt_root_validation_after_adding_blocks

mdarade:
* **Problem Overview**  
Last-stored-checkpoint metric was being read from just restarted
replica. Increased wait time to 2 sec. Ideally metric should not be
allowed to read while replica is still coming up or going down.

* **Testing Done**  
Ran test locally 50+ times with HW config same as CI.


glevkovich (approved):
LGTM, Thanks @mdarade

eyalrund (approved):

StefanPalashev (approved):
LGTM, thanks @mdarade.

PR #2701, opened 2022-06-28
Asan warning fixes

StefanPalashev:
* **Problem Overview**  
A memory leak in `InMemoryDataStore` was discovered while performing an ASAN run on the `bcstatetransfer_tests` gtest suite.

This Pull Request changes the raw pointers, storing reserved pages, to smart ones (std::shared_ptr).
It adds an additional check to avoid excessive memory allocation and replacement of the same data.

* **Testing Done**  
Ran the relevant gtest and Apollo suites to ensure that the leak is no longer present.


glevkovich (changes requested):

glevkovich (approved):
Thanks @StefanPalashev !

PR #2704, opened 2022-06-29
Fix PreProcessor crash

yuliasherman:
* **Problem Overview**  
  The fix: when a primary replica completes the pre-processed request handling, it should not be interrupted by a received 
  PreProcessReplyMsg message for the same request. Theoretically, this should not harm the processing, but in the two 
  core dumps I've analyzed, I saw those two events competing in handling the same request.
* **Testing Done**  
  4 days of the DAML test on a reserved AWS system

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -1025,7 +1025,7 @@ bool PreProcessor::handleSinglePreProcessRequestMsg(PreProcessRequestMsgSharedPt
     preProcessorMetrics_.preProcInFlyRequestsNum++;  // Increase the metric on non-primary replica
     auto totalPreExecDurationRecorder = TimeRecorder(*totalPreExecDurationRecorder_.get());
     // Pre-process the request, calculate a hash of the result and send a reply message back
-    launchAsyncReqPreProcessingJob(preProcessReqMsg, batchCid, false, false, std::move(totalPreExecDurationRecorder));
+    launchAsyncReqPreProcessingJob(preProcessReqMsg, batchCid, false, std::move(totalPreExecDurationRecorder));
     return true;
   }
   return false;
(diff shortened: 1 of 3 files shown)


bftengine/src/preprocessor/RequestProcessingState.cpp
```diff
@@ -257,15 +264,17 @@ PreProcessingResult RequestProcessingState::definePreProcessingConsensusResult()
                  << KVLOG(batchCid_, reqSeqNum_, reqCid_, numOfReceivedReplies_, numOfRequiredEqualReplies_));
     return CONTINUE;
   }
-  if (primaryPreProcessResultLen_ == 0) {
```
HristoStaykov, line 260: What is a case in which we can have **primaryPreProcessResultLen_ == 0**. I understand with the modification we consider **preprocessingRightNow_** and don't proceed with the logic that follows in the function, but I was wondering how can **primaryPreProcessResultLen_ == 0**. If I understand correctly we now ignore this case?

yuliasherman: We can verify primary pre-processing completion either through primaryPreProcessResultLen_ != 0 or preprocessingRightNow_ = false. The second one looks more native.

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1808,46 +1816,31 @@ ReqId PreProcessor::getOngoingReqIdForClient(uint16_t clientId, uint16_t reqOffs
...
-  setPreprocessingRightNow(clientId, reqOffsetInBatch, false);
-  const PreProcessingResult result =
-      handlePreProcessedReqByPrimaryAndGetConsensusResult(clientId, reqOffsetInBatch, resultBufLen, preProcessResult);
-  if (result != NONE)
-    handlePreProcessReplyMsg(
-        preProcessReqMsg->getCid(), result, clientId, reqOffsetInBatch, preProcessReqMsg->reqSeqNum(), batchCid);
+  PreProcessingResult result = NONE;
+  const auto &reqEntry = ongoingReqBatches_[clientId]->getRequestState(reqOffsetInBatch);
+  {
+    lock_guard<mutex> lock(reqEntry->mutex);
```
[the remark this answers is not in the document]
yuliasherman: Yes. I would like to have holdCompleteOrCancelRequest called under the same lock now.

eyalrund:
I want to test it for performance also in maestro and on high end setup

yuliasherman:
> I want to test it for performance also in maestro and on high end setup

Is 650TPS enough? I'm not sure I can catch AMD env.

PR #2708, opened 2022-06-30
Fixes for 5 State Transfer Apollo tests 

mdarade:
* **Problem Overview**  
Couple of state transfer tests were failing due to below issues.
1. Tests were not waiting enough for replica's to come up before issuing writes.
2. Test was wrongly expecting for 5 checkpoints to go through in presence of f+1 byzantine replica's.
3. Test should have waited for pruning to complete before issuing new writes.

* **Testing Done**  
Ran all tests 50+ times in local environment matching CI HW.

### DIFF
--- a/tests/apollo/test_skvbc_checkpoints.py
+++ b/tests/apollo/test_skvbc_checkpoints.py
@@ -397,7 +397,6 @@ class SkvbcCheckpointTest(ApolloTest):
             isolated_replicas,
             expected_checkpoint_num=lambda ecn: ecn == checkpoint_before + 1)
 
-    @unittest.skip("Unstable test")
     @with_trio
     @with_bft_network(start_replica_cmd_with_corrupted_checkpoint_msgs(corrupt_checkpoints_from_replica_ids={ 1 }),
                       selected_configs=lambda n, f, c: n == 7)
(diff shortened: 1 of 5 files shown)


tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -440,7 +441,7 @@ async def test_rvt_conflict_detection_after_corrupting_checkpoint_msg_for_f_plus
             with self.assertRaises(trio.TooSlowError):
                   await skvbc.fill_and_wait_for_checkpoint(
                     initial_nodes=bft_network.all_replicas(without=byzantine_replica_ids),
-                    num_of_checkpoints_to_add=5,
+                    num_of_checkpoints_to_add=2,
```
glevkovich, line 444: Please correct me on below claim, I might be missing something again...

I am not convinced that this is what we want to check, even with the new behavior.
I think we still want to try to advance 5 checkpoints, but see that network advanced only 2 checkpoints.
Otherwise, what are you testing here? The idea is to show that network does not advance out of the window ,since there are f+1 byzantine. If you write only 2 checkpoints, they will all reach that point, since it's still inside the window. In that case you are loosing the whole idea of the test, test always succeed and nothing was tested.Better way is to:
1) write 5 checkpoints
2) check that network advanced only 2
3) check that number_of_checkpoint_mismatch == (f+1)*5

glevkovich (changes requested):

tests/apollo/test_skvbc_checkpoints.py
mdarade: Thanks Gil.
I agree with first two points.
Third point is not feasible as at max there are going to be only (f+1)*2 mismatched checkpoint.

glevkovich: Oh ye, you are right..

glevkovich (approved):
Thanks Mahveer, really good fixes...

tests/apollo/test_skvbc_checkpoints.py
```diff
@@ -474,6 +425,10 @@ async def test_checkpoint_propagation_after_corrupting_checkpoint_msg_for_primar
         """
         bft_network.start_all_replicas()
         skvbc = kvbc.SimpleKVBCProtocol(bft_network)
+        await trio.sleep(2)
```
glevkovich, line 428: I think ypu should explain here and in line 450 why are you doing this.
It would be very hard for a reader to understand in a short time why we sleep here and why the number of mismatches is expected tp be 3 or 4.
same for line 504.

In general, just think you will need to debug it yourself in 1 year - any chance you will remember yourself the reason you did that?

glevkovich (approved):
Thanks @mdarade , add one comment about commenting.

eyalrund (approved):

PR #2713, opened 2022-06-30
V4 Migration related fixes

arc-vmware:
* **Problem Overview**  
1) The link st link bool is passed as false for read-only DB and true
for the read-write DB.
2) The ending condition is fixed since the migration's storage thread
was getting into an infinite loop.
* **Testing Done**  
Ran the migration tool on v1 DB to check the fix for infinite loop and DB loading.


cloudnoize (approved):
thanks @arc-vmware !

PR #2714, opened 2022-06-30
Fix calculation of connected connections count

ukg19:
* **Problem Overview**  
  While counting number of connected replicas inside `MsgsCommunicator::numOfConnectedReplicas()`, we are incrementing the count when `getCurrentConnectionStatus()` is not Disconnected. But ConnectionStatus can also be `Unknown` as declared [here](https://github.com/vmware/concord-bft/blob/e9488e6d44db412dd568f1f88e3fe8d19c07c75c/communication/include/communication/ICommunication.hpp#L24) . This function will also increment the count in that case. Thus falsely calculating the number of connected connections and can give more number of connected connections than actual. Hence it has the potential to conclude on a false consensus while lesser number of replicas ( < 2f+1) are connected. The problem is not seen currently as `Unknown` state is not being used, but it will be potential bug when this state is used.
In this PR, I've fixed this problem, by adding up count only when getCurrentConnectionStatus() returns connection status as Connected. 
* **Testing Done**  
  Verified using existing apollo test cases and temporary logs which indicates the number of connections are calculated correctly.

### DIFF
--- a/bftengine/src/bftengine/MsgsCommunicator.cpp
+++ b/bftengine/src/bftengine/MsgsCommunicator.cpp
@@ -60,8 +60,9 @@ void MsgsCommunicator::send(std::set<NodeNum> dests, char* message, size_t messa
 uint32_t MsgsCommunicator::numOfConnectedReplicas(uint32_t clusterSize) {
   uint32_t ret{0};
   for (uint32_t i = 0; i < clusterSize; ++i) {
-    if (communication_->getCurrentConnectionStatus(i) == ConnectionStatus::Disconnected) continue;
-    ++ret;
+    if (communication_->getCurrentConnectionStatus(i) == ConnectionStatus::Connected) {
+      ++ret;
+    }
   }
   return ret;
 }


bftengine/src/bftengine/MsgsCommunicator.cpp
```diff
@@ -60,8 +60,9 @@ void MsgsCommunicator::send(std::set<NodeNum> dests, char* message, size_t messa
 uint32_t MsgsCommunicator::numOfConnectedReplicas(uint32_t clusterSize) {
   uint32_t ret{0};
   for (uint32_t i = 0; i < clusterSize; ++i) {
-    if (communication_->getCurrentConnectionStatus(i) == ConnectionStatus::Disconnected) continue;
-    ++ret;
+    if (communication_->getCurrentConnectionStatus(i) == ConnectionStatus::Connected) {
+      ++ret;
+    }
```
teoparvanov, line 65: Hi Utkarsh - did you explicitly verify the count is correct after your change? What could help is for example a temporary  INFO log while running the Apollo tests, to make sure the number is correct?

Or maybe you validated using another method? If so, please provide more details in the **Testing done** section.

Thank you!
Teo

cloudnoize (approved):
nice catch! thanks

teoparvanov (approved):
LGTM, thanks Utkarsh - esp. for the thorough testing!

PR #2715, opened 2022-06-30
Fix initial key exchange issues

yontyon:
* **Problem Overview**  
  An initial key exchange may be invoked from 3 different threads (where we configured to start with n-f replicas):
  1. from the start thread (with sequence number 0)
  2. from the message processing thread (with sequence number 0) if VC happened
  3. from the state transfer thread in case the replica started with state transfer (with the last transferred sequence number)
These three threads may collide between themselves where we get the new key message of ourselves through the consensus (for example, if the state transfer thread  created a new candidate key before the previous one committed)
This PR is to fix these issues by adding some more conditions and constraints on the creation and committing of keys.
* **Testing Done**  
  CI tests

### DIFF
--- a/bftengine/src/bftengine/KeyExchangeManager.cpp
+++ b/bftengine/src/bftengine/KeyExchangeManager.cpp
@@ -237,6 +237,10 @@ void KeyExchangeManager::sendKeyExchange(const SeqNum& sn) {
     LOG_INFO(KEY_EX_LOG, "we already have a candidate for this sequence number, trying to send it again");
     msg.pubkey = candidate_private_keys_.generated.pub;
     msg.repID = repID_;
+    SeqNum new_sn = sn;
+    if ((sn - candidate_private_keys_.generated.sn) / checkpointWindowSize < 2)
+      new_sn = candidate_private_keys_.generated.sn;
+    candidate_private_keys_.generated.sn = new_sn;
     msg.generated_sn = candidate_private_keys_.generated.sn;
     msg.epoch = EpochManager::instance().getSelfEpochNumber();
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -229,9 +232,21 @@ void KeyExchangeManager::sendKeyExchange(const SeqNum& sn) {
...
+                 sn, candidate_private_keys_.generated.sn));
+    return;
+  }
   KeyExchangeMsg msg;
-  if (sn == candidate_private_keys_.generated.sn && !candidate_private_keys_.generated.cid.empty()) {
-    LOG_INFO(KEY_EX_LOG, "we already have a candidate for this sequence number, trying to send it again");
+  // Notice, that a replica always sends candidates in acsending oreder, this means that in every prePrepare message,
+  // even if it consist two KEY-EXCHANGE messages from the same replica, the first one will always be possessed before
+  // the later one
+  if (sn > 0 && sn - candidate_private_keys_.generated.sn / checkpointWindowSize < 2 &&
```
nkumar04, line 245: `if (sn > 0 && (sn - candidate_private_keys_.generated.sn) / checkpointWindowSize < 2 &&
      !candidate_private_keys_.generated.cid.empty()) {`

Did you mean this?

yontyon: good catch!
Thanks

toly-kournik (approved):
LGTM!
Minimalist and elegant change !

eyalrund (approved):

PR #2716, opened 2022-06-30
Fix a bug when reaching the maximal size of pre-prepare

yontyon:
* **Problem Overview**  
  When reaching the limit of prePrepare message size we need to break the loop where we add requests. However, a bug was found where we compare the client message to the overall size before adding the requests causing to deletion of all extra client request messages.
* **Testing Done**  
  CI


yuliasherman (commented):
The fix looks correct, but it has to be tested using a unit test and/or Apollo test.

PR #2719, opened 2022-07-03
Unblock message-processor during initial startup/key-exchange while view change occurs

ukg19:
* **Problem Overview**  
Calls to `KeyExchangeManager::instance().sendInitialKey(currentPrimary());` is made from two different threads, 
(1) from the thread that starts the replica
(2) from the message processing thread right after the justified VC, (part of view change callback)
But `sendInitialKey` method has a mutex which means that this method will be blocked for 60 seconds as inside it, it tries to connect to primary for 60 seconds(starts under assumption that replica 0 will be primary). At the same time, if a view change has occurred and primary has changed, it will still be blocked and keep trying to connect to replica 0 for 60 seconds. Thus message processing thread is not able to enter `sendInitialKey` method, as it has to be protected by mutex and the other thread working inside it will not leave for 60 seconds. 
During this, other replicas might start with their message processing but this replica will be delayed as message processing is blocked. Due to delayed processing, and when time-service enabled, it can cause to additional view changes as PPM message processing in such replicas will be delayed making client request message timeout.
**Fix**: Instead of waiting for a fixed primary, I made provision to get current primary inside sendInitialKey function, so that it knows in real time, which primary to connect.
* **Testing Done**  
Added an apollo test case to test this scenario.
Ran test 100 times, to check the fix and stability after adding the fix.

### DIFF
--- a/bftengine/include/bftengine/KeyExchangeManager.hpp
+++ b/bftengine/include/bftengine/KeyExchangeManager.hpp
@@ -23,7 +23,7 @@
 namespace bftEngine::impl {
 
 class IInternalBFTClient;
-
+class ReplicaImp;
 typedef int64_t SeqNum;  // TODO [TK] redefinition
 
 class KeyExchangeManager {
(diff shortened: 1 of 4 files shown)


yontyon (approved):
LGTM!
Thanks

nkumar04 (commented):
LGTM!!
Thanks!

HristoStaykov (approved):
LGTM! Thanks for fixing this situation @ukg19 ! It is great that we also have it covered with an Apollo test.

nityavasudevan:
Thanks for uncovering this corner case and fixing Utkarsh!

nityavasudevan (approved):

PR #2721, opened 2022-07-06
Restart recovery test workflow stability improvements

titusen:
Nightly restart_recovery suite is unstable. These are minor improvements.

Testing: On local VM reproduction rate in test_restarting_replica_during_system_is_in_view_change_next_primary_down got sufficiently low.


teoparvanov (approved):
LGTM - let's see if this improves the stability of the nightly run. Otherwise, it's maybe an indication of an actual bug in the product?

PR #2723, opened 2022-07-06
Dafny SBFT: model execution & checkpoint protocol

HristoStaykov:
* **Problem Overview**  
  Checkpoint generation/collection implemented in honest replicas. This gives the opportunity for us to limit the Working Window size. All infinite maps we used in the Working Window in Replicas are now converted to finite maps.
* **Testing Done**  
  This change is related to the formal model and does not affect the executable code.

### DIFF
--- a/docs/sbft-formal-model/messages.dfy
+++ b/docs/sbft-formal-model/messages.dfy
@@ -74,7 +74,9 @@ module Messages {
                      | Commit(view:ViewNum, seqID:SequenceID, operationWrapper:OperationWrapper)
                      | ClientRequest(clientOp:ClientOperation)
                      | ViewChangeMsg(newView:ViewNum, certificates:imap<SequenceID, PreparedCertificate>) // omitting last stable because we don't have checkpointing yet.
-                     | NewViewMsg(newView:ViewNum, vcMsgs:ViewChangeMsgsSelectedByPrimary) {
+                     | NewViewMsg(newView:ViewNum, vcMsgs:ViewChangeMsgsSelectedByPrimary) 
+                     | CheckpointMsg(seqIDReached:SequenceID)
+                     {
                        predicate WF() {
                          && (ViewChangeMsg? ==> Library.FullImap(certificates)) //TODO: rename TotalImap
                        }
--- a/docs/sbft-formal-model/messages.dfy
+++ b/docs/sbft-formal-model/messages.dfy
@@ -17,7 +17,7 @@ module Messages {
   import opened HostIdentifiers
   import Network
 
-  type SequenceID = nat
+  type SequenceID = k:nat | 0 < k witness 1
   type ViewNum = nat
 
   datatype ClientOperation = ClientOperation(sender:HostId, timestamp:nat)
--- a/docs/sbft-formal-model/replica.i.dfy
+++ b/docs/sbft-formal-model/replica.i.dfy
@@ -342,13 +342,13 @@ module Replica {
                      :: v.workingWindow.committedClientOperations[seqID].Some?)
   }
 
-  predicate Execute(c:Constants, v:Variables, v':Variables, msgOps:Network.MessageOps<Message>)
+  predicate Execute(c:Constants, v:Variables, v':Variables, msgOps:Network.MessageOps<Message>, seqID:SequenceID)
   {
     && v.WF(c)
     && msgOps.NoSendRecv()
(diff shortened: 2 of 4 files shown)


docs/sbft-formal-model/messages.dfy
```diff
@@ -17,7 +17,7 @@ module Messages {
   import opened HostIdentifiers
   import Network
 
-  type SequenceID = nat
+  type SequenceID = k:nat | 0 < k witness 1
```
teoparvanov, line 20: Does `witness 1` indicate that k=1 is a witness that instantiates this predicate?

docs/sbft-formal-model/replica.i.dfy
```diff
@@ -35,35 +38,45 @@ module Replica {
...
-      && PrePreparesRcvdWF(prePreparesRcvd)
+      && committedClientOperations.Keys == getActiveSequenceIDs(c)
+      && preparesRcvd.Keys == getActiveSequenceIDs(c)
+      && commitsRcvd.Keys == getActiveSequenceIDs(c)
+      && prePreparesRcvd.Keys == getActiveSequenceIDs(c)
+      && (forall x | x in prePreparesRcvd && prePreparesRcvd[x].Some? :: prePreparesRcvd[x].value.payload.PrePrepare?)
       && (forall seqID | seqID in preparesRcvd :: PrepareProofSetWF(c, preparesRcvd[seqID]))
       && (forall seqID | seqID in commitsRcvd :: CommitProofSetWF(c, commitsRcvd[seqID]))
     }
+    function ShiftWorkingWindow<T>(c:Constants, m:map<SequenceID,T>, empty:T) : map<SequenceID,T>
```
teoparvanov, line 72: or simply, `Shift()`, as this function operates on the `workingWindow` structure?

docs/sbft-formal-model/replica.i.dfy
```diff
@@ -473,14 +525,52 @@ module Replica {
...
+    && v' == v.(checkpointMsgsRecvd := v.checkpointMsgsRecvd.(msgs := v.checkpointMsgsRecvd.msgs + {msg}))
+  }
+
+  predicate HasStableCheckpointForSeqID(c:Constants, v:Variables, seqID:SequenceID) {
+    && v.WF(c)
+    && var relevantCheckpointMsgs := set msg | && msg in v.checkpointMsgsRecvd.msgs 
+                                               && msg.payload.CheckpointMsg?
+                                               && msg.payload.seqIDReached == seqID;
+    && |relevantCheckpointMsgs| >= c.clusterConfig.AgreementQuorum()
+  }
```
teoparvanov, line 551: Awesome :) State of the art definition!

docs/sbft-formal-model/replica.i.dfy
```diff
@@ -473,14 +525,52 @@ module Replica {
...
+
+  predicate AdvanceWorkingWindow(c:Constants, v:Variables, v':Variables, msgOps:Network.MessageOps<Message>, seqID:SequenceID) {
+    && v.WF(c)
+    && msgOps.NoSendRecv()
+    && seqID > v.workingWindow.lastStableCheckpoint
```
teoparvanov, line 556: Shouldn't the pre-condition be stricter, and say something like `seqID > ls + win`, like in the paper - where `win` is the size of the active window?

teoparvanov (commented):
Great progress introducing the checkpoint protocol, thanks @HristoStaykov ! 
Do the existing invariant proofs succeed with the Checkpoint protocol enabled?

docs/sbft-formal-model/messages.dfy
HristoStaykov: Yes, exactly. This is a change that we made because in our system we start committing from Sequence ID No:1. Previously in the model the **SequenceID** was **nat**, starting from 0 to infinity. Both ways the invariants are preserved.

docs/sbft-formal-model/replica.i.dfy
HristoStaykov: Yes, I will change it, now when I look at it, it is a bit too wordy in the usage :)

docs/sbft-formal-model/replica.i.dfy
teoparvanov: Ohh I see, good point.

teoparvanov (approved):
Looks great, thanks @HristoStaykov !

PR #2725, opened 2022-07-07
Allow reserved pages contain replies of all requests in a batch

Efrat1:
* **Problem Overview**  
* Until now we had reserved pages for one request per client. That means that when client sends a batch, only one reply of it is saved to reserved pages. If a one reply (or more) of a batch is missing, the client retransmits the batch. In that case, we would like to have the replies we already sent saved to reserved pages to avoid additional consensuses and executions.
So, with this PR, the reserved pages can contain #batch_size replies (when client batching and pre-processing are enabled).
The location of a reply in client's reserved pages is determined according to request's offset in batch.
* **Testing Done** 
* unit tests and Apollo tests
* Tested locally with docker-compose

### DIFF
--- a/bftengine/include/bftengine/ClientMsgs.hpp
+++ b/bftengine/include/bftengine/ClientMsgs.hpp
@@ -42,6 +42,7 @@ struct ClientRequestMsgHeader {
   uint32_t cidLength = 0;
   uint16_t reqSignatureLength = 0;
   uint32_t extraDataLength = 0;
+  uint16_t indexInBatch = 0;
 
   // followed by the request (security information, such as signatures, should be part of the request)
 
--- a/bftengine/include/bftengine/IRequestHandler.hpp
+++ b/bftengine/include/bftengine/IRequestHandler.hpp
@@ -47,6 +47,7 @@ class IRequestsHandler {
     uint32_t maxReplySize = 0;
     char *outReply = nullptr;
     uint64_t requestSequenceNum = executionSequenceNum;
+    uint16_t reqIndexInClientBatch = 0;
     uint32_t outExecutionStatus = 1;  // UNKNOWN
     uint32_t outActualReplySize = 0;
     uint32_t outReplicaSpecificInfoSize = 0;
--- a/bftengine/src/bftengine/ClientsManager.cpp
+++ b/bftengine/src/bftengine/ClientsManager.cpp
@@ -62,9 +62,10 @@ bool ClientsManager::RequestsInfo::removeRequestsOutOfBatchBoundsSafe(NodeIdType
 void ClientsManager::RequestsInfo::removeOldPendingReqsSafe(NodeIdType clientId, ReqId reqSeqNum) {
   const lock_guard<mutex> lock(requestsMapMutex_);
   for (auto it = requestsMap_.begin(); it != requestsMap_.end();) {
-    if (it->first <= reqSeqNum) {
+    const auto oldSeqNum = it->first;
+    if (oldSeqNum <= reqSeqNum) {
+      LOG_INFO(CL_MNGR, "Remove old pending request" << KVLOG(clientId, oldSeqNum, reqSeqNum));
       it = requestsMap_.erase(it);
-      LOG_INFO(CL_MNGR, "Remove old pending request" << KVLOG(clientId, reqSeqNum));
     } else
       it++;
   }
@@ -132,47 +133,48 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
 
 /*************************** Class ClientsManager::RepliesInfo ***************************/
 
-void ClientsManager::RepliesInfo::deleteOldestReplyIfNeededSafe(NodeIdType clientId, uint16_t maxNumOfReqsPerClient) {
-  Time earliestTime = MaxTime;
-  ReqId earliestReplyId = 0;
-  if (repliesMap_.size() < maxNumOfReqsPerClient) return;
-  if (repliesMap_.size() > maxNumOfReqsPerClient)
+void ClientsManager::RepliesInfo::deleteReplyIfNeededSafe(NodeIdType clientId,
+                                                          ReqId reqSeqNum,
+                                                          uint16_t maxNumOfReqsPerClient,
+                                                          uint16_t reqIndex) {
+  ReqId savedReplyId = 0;
+  if (repliesBiMap_.size() > maxNumOfReqsPerClient) {
     LOG_FATAL(CL_MNGR,
               "More than maxNumOfReqsPerClient_ items in repliesInfo"
-                  << KVLOG(repliesMap_.size(), maxNumOfReqsPerClient, clientId));
-  for (auto& reply : repliesMap_) {
-    if (earliestTime > reply.second) {
-      earliestReplyId = reply.first;
-      earliestTime = reply.second;
+                  << KVLOG(repliesBiMap_.size(), maxNumOfReqsPerClient, clientId));
+  }
+
+  if (repliesBiMap_.right.find(reqIndex) != repliesBiMap_.right.end()) {
+    savedReplyId = repliesBiMap_.right.at(reqIndex);
+    if (savedReplyId > reqSeqNum) {
+      LOG_WARN(CL_MNGR, "A newer reply was already saved" << KVLOG(clientId, reqSeqNum, savedReplyId, reqIndex));
     }
   }
-  const lock_guard<mutex> lock(repliesMapMutex_);
-  if (earliestReplyId) {
-    repliesMap_.erase(earliestReplyId);
-  } else if (!repliesMap_.empty()) {
-    // Delete reply arrived through ST
-    auto const& reply = repliesMap_.cbegin();
-    earliestReplyId = reply->first;
-    earliestTime = reply->second;
-    repliesMap_.erase(reply);
+
+  if (savedReplyId) {
+    const lock_guard<mutex> lock(repliesMapMutex_);
+    repliesBiMap_.left.erase(savedReplyId);
+    LOG_DEBUG(CL_MNGR, "Deleted reply message" << KVLOG(clientId, savedReplyId, reqIndex, repliesBiMap_.size()));
   }
-  LOG_DEBUG(CL_MNGR,
-            "Deleted reply message" << KVLOG(
-                clientId, earliestReplyId, earliestTime.time_since_epoch().count(), repliesMap_.size()));
 }
 
-bool ClientsManager::RepliesInfo::insertOrAssignSafe(ReqId reqSeqNum, Time time) {
+void ClientsManager::RepliesInfo::insertOrAssignSafe(ReqId reqSeqNum, uint16_t reqIndexInBatch) {
   const lock_guard<mutex> lock(repliesMapMutex_);
-  return repliesMap_.insert_or_assign(reqSeqNum, time).second;
+  repliesBiMap_.insert(boost::bimaps::bimap<ReqId, uint16_t>::value_type(reqSeqNum, reqIndexInBatch));
 }
 
 bool ClientsManager::RepliesInfo::findSafe(ReqId reqSeqNum) {
   const lock_guard<mutex> lock(repliesMapMutex_);
-  return (repliesMap_.find(reqSeqNum) != repliesMap_.end());
+  return (repliesBiMap_.left.find(reqSeqNum) != repliesBiMap_.left.end());
 }
 
 bool ClientsManager::RepliesInfo::find(ReqId reqSeqNum) const {
-  return (repliesMap_.find(reqSeqNum) != repliesMap_.end());
+  return (repliesBiMap_.left.find(reqSeqNum) != repliesBiMap_.left.end());
+}
+
+uint16_t ClientsManager::RepliesInfo::getIndex(ReqId reqSeqNum) const {
+  const auto& replyIt = repliesBiMap_.left.find(reqSeqNum);
+  return (replyIt != repliesBiMap_.left.end()) ? replyIt->second : 0;
 }
 
 /*************************** Class ClientsManager ***************************/
@@ -204,12 +206,16 @@ ClientsManager::ClientsManager(const std::set<NodeIdType>& proxyClients,
       clientServices_{clientServices},
       internalClients_{internalClients},
       maxReplySize_(ReplicaConfig::instance().getmaxReplyMessageSize()),
-      maxNumOfReqsPerClient_(
-          ReplicaConfig::instance().clientBatchingEnabled ? ReplicaConfig::instance().clientBatchingMaxMsgsNbr : 1),
+      maxNumOfReqsPerClient_(ReplicaConfig::instance().clientBatchingEnabled &&
+                                     ReplicaConfig::instance().preExecutionFeatureEnabled
+                                 ? ReplicaConfig::instance().clientBatchingMaxMsgsNbr
+                                 : 1),
       metrics_(metrics),
       metric_reply_inconsistency_detected_{metrics_.RegisterCounter("totalReplyInconsistenciesDetected")},
       metric_removed_due_to_out_of_boundaries_{metrics_.RegisterCounter("totalRemovedDueToOutOfBoundaries")} {
-  reservedPagesPerClient_ = reservedPagesPerClient(sizeOfReservedPage(), maxReplySize_);
+  ConcordAssert(maxNumOfReqsPerClient_ > 0);
+  reservedPagesPerRequest_ = reservedPagesPerRequest(sizeOfReservedPage(), maxReplySize_);
+  reservedPagesPerClient_ = reservedPagesPerClient(sizeOfReservedPage(), maxReplySize_, maxNumOfReqsPerClient_);
   for (NodeIdType i = 0; i < ReplicaConfig::instance().numReplicas + ReplicaConfig::instance().numRoReplicas; i++) {
     clientIds_.insert(i);
   }
@@ -240,11 +246,19 @@ ClientsManager::ClientsManager(const std::set<NodeIdType>& proxyClients,
                         << KVLOG(sizeOfReservedPage(), reservedPagesPerClient_, maxReplySize_, maxNumOfReqsPerClient_));
 }
 
-uint32_t ClientsManager::reservedPagesPerClient(const uint32_t& sizeOfReservedPage, const uint32_t& maxReplySize) {
-  uint32_t reservedPagesPerClient = maxReplySize / sizeOfReservedPage;
+uint32_t ClientsManager::reservedPagesPerRequest(uint32_t sizeOfReservedPage, uint32_t maxReplySize) {
+  uint32_t reservedPagesPerReq = maxReplySize / sizeOfReservedPage;
   if (maxReplySize % sizeOfReservedPage != 0) {
-    reservedPagesPerClient++;
+    reservedPagesPerReq++;
   }
+  return reservedPagesPerReq;
+}
+
+uint32_t ClientsManager::reservedPagesPerClient(uint32_t sizeOfReservedPage,
+                                                uint32_t maxReplySize,
+                                                uint16_t maxNumReqPerClient) {
+  uint32_t reservedPagesPerReq = reservedPagesPerRequest(sizeOfReservedPage, maxReplySize);
+  uint32_t reservedPagesPerClient = reservedPagesPerReq * maxNumReqPerClient;
   reservedPagesPerClient++;  // for storing client public key
   return reservedPagesPerClient;
 }
@@ -255,34 +269,41 @@ uint32_t ClientsManager::reservedPagesPerClient(const uint32_t& sizeOfReservedPa
 // * fill clientInfo
 // * remove pending request if loaded reply is newer
 void ClientsManager::loadInfoFromReservedPages() {
-  for (auto const& clientId : clientIds_) {
+  for (auto const& [clientId, clientInfo] : clientsInfo_) {
     if (internalClients_.find(clientId) != internalClients_.end()) continue;
     if (loadReservedPage(getKeyPageId(clientId), sizeOfReservedPage(), scratchPage_.data())) {
-      auto& info = clientsInfo_[clientId];
       std::istringstream iss(scratchPage_);
-      concord::serialize::Serializable::deserialize(iss, info.pubKey);
-      ConcordAssertGT(info.pubKey.first.length(), 0);
-      KeyExchangeManager::instance().loadClientPublicKey(info.pubKey.first, info.pubKey.second, clientId, false);
+      concord::serialize::Serializable::deserialize(iss, clientInfo.pubKey);
+      ConcordAssertGT(clientInfo.pubKey.first.length(), 0);
+      KeyExchangeManager::instance().loadClientPublicKey(
+          clientInfo.pubKey.first, clientInfo.pubKey.second, clientId, false);
     }
 
-    if (!loadReservedPage(getReplyFirstPageId(clientId), sizeOfReservedPage(), scratchPage_.data())) continue;
-    ClientReplyMsgHeader* replyHeader = (ClientReplyMsgHeader*)scratchPage_.data();
-    ConcordAssert(replyHeader->msgType == 0 || replyHeader->msgType == MsgCode::ClientReply);
-    ConcordAssert(replyHeader->currentPrimaryId == 0);
-    ConcordAssert(replyHeader->replyLength >= 0);
-    ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
+    for (uint16_t indexInBatch = 0; indexInBatch < maxNumOfReqsPerClient_; indexInBatch++) {
+      const auto pageOffset = getReplyFirstPageId(clientId) + indexInBatch * reservedPagesPerRequest_;
+      if (!loadReservedPage(pageOffset, sizeOfReservedPage(), scratchPage_.data())) continue;
+      ClientReplyMsgHeader* replyHeader = reinterpret_cast<ClientReplyMsgHeader*>(scratchPage_.data());
+      ConcordAssert(replyHeader->msgType == 0 || replyHeader->msgType == MsgCode::ClientReply);
+      ConcordAssert(replyHeader->currentPrimaryId == 0);
+      ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
 
-    clientsInfo_[clientId].repliesInfo->deleteOldestReplyIfNeededSafe(clientId, maxNumOfReqsPerClient_);
-    const auto& res = clientsInfo_[clientId].repliesInfo->insertOrAssignSafe(replyHeader->reqSeqNum, MinTime);
-    LOG_INFO(CL_MNGR, "Added/updated reply message" << KVLOG(clientId, replyHeader->reqSeqNum, res));
-    clientsInfo_[clientId].requestsInfo->removeOldPendingReqsSafe(clientId, replyHeader->reqSeqNum);
+      clientInfo.repliesInfo->deleteReplyIfNeededSafe(
+          clientId, replyHeader->reqSeqNum, maxNumOfReqsPerClient_, indexInBatch);
+      clientInfo.repliesInfo->insertOrAssignSafe(replyHeader->reqSeqNum, indexInBatch);
+
+      LOG_INFO(CL_MNGR,
+               "Added/updated reply message" << KVLOG(clientId, replyHeader->reqSeqNum, indexInBatch, pageOffset));
+      clientInfo.requestsInfo->removeOldPendingReqsSafe(clientId, replyHeader->reqSeqNum);
+    }
   }
 }
 
 bool ClientsManager::hasReply(NodeIdType clientId, ReqId reqSeqNum) {
   try {
     const bool found = clientsInfo_.at(clientId).repliesInfo->findSafe(reqSeqNum);
-    if (found) LOG_DEBUG(CL_MNGR, "Reply found for" << KVLOG(clientId, reqSeqNum));
+    if (found) {
+      LOG_DEBUG(CL_MNGR, "Reply found for" << KVLOG(clientId, reqSeqNum));
+    }
     return found;
   } catch (const std::out_of_range& e) {
     LOG_DEBUG(CL_MNGR, "No info found for client" << KVLOG(clientId, reqSeqNum));
@@ -290,8 +311,13 @@ bool ClientsManager::hasReply(NodeIdType clientId, ReqId reqSeqNum) {
   }
 }
 
-void ClientsManager::deleteOldestReply(NodeIdType clientId) {
-  clientsInfo_[clientId].repliesInfo->deleteOldestReplyIfNeededSafe(clientId, maxNumOfReqsPerClient_);
+void ClientsManager::deleteReplyIfNeeded(NodeIdType clientId, uint16_t indexInBatch, ReqId newReqSeqNum) {
+  try {
+    clientsInfo_[clientId].repliesInfo->deleteReplyIfNeededSafe(
+        clientId, newReqSeqNum, maxNumOfReqsPerClient_, indexInBatch);
+  } catch (const std::out_of_range& e) {
+    LOG_DEBUG(CL_MNGR, "No info found for client" << KVLOG(clientId, newReqSeqNum));
+  }
 }
 
 // Reference the ClientInfo of the corresponding client:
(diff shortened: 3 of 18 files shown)


yuliasherman:
Please create a concord MR pointing to this PR and verify that the concord passes the tests. Then launch a job for 3 hours through Maestro.

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -244,15 +248,22 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
...
+    void deleteReplyIfNeededSafe(NodeIdType clientId,
+                                 ReqId reqSeqNum,
+                                 uint16_t maxNumOfReqsPerClient,
+                                 uint16_t reqOffset);
+    void insertOrAssignSafe(NodeIdType clientId,
+                            ReqId reqSeqNum,
+                            uint16_t maxNumOfReqsPerClient,
+                            uint16_t requestOffset);
     bool findSafe(ReqId reqSeqNum);
+    uint16_t getOffsetSafe(ReqId reqSeqNum);
```
glevkovich, line 260: nit: 
```
uint16_t getOffsetSafe(ReqId reqSeqNum) const;
```

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -244,15 +248,22 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
...
 
     bool find(ReqId reqSeqNum) const;
 
    public:
     std::mutex repliesMapMutex_;
-    std::map<ReqId, Time> repliesMap_;
+    std::map<ReqId, uint16_t> repliesMap_;
```
glevkovich, line 266: See my comment about offset, why using. uint16_t and not uint32_t or better size_t ?

glevkovich (changes requested):

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,45 +132,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
+void ClientsManager::RepliesInfo::deleteReplyIfNeededSafe(NodeIdType clientId,
+                                                          ReqId reqSeqNum,
+                                                          uint16_t maxNumOfReqsPerClient,
+                                                          uint16_t reqOffset) {
+  ReqId oldReplyId = 0;
   if (repliesMap_.size() > maxNumOfReqsPerClient)
     LOG_FATAL(CL_MNGR,
               "More than maxNumOfReqsPerClient_ items in repliesInfo"
                   << KVLOG(repliesMap_.size(), maxNumOfReqsPerClient, clientId));
+  const lock_guard<mutex> lock(repliesMapMutex_);
```
yuliasherman, line 144: Please return the lock back where it was before. I see I forgot to add an explanation for the 'strange' locking model here. The thread safety here is between the PreProcessor thread, which is read-only, and the ReplicaImp main thread, which performs either read or write at a time. So, all write operations should be guarded as well as PrePropessor calls hasReply and isClientRequestInProcess. All other operations are safe. Please add the explaining comment to avoid misunderstanding.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,45 +132,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
+
+  // TODO: create another/additional data structure for efficiency (map offset to seqNum)
```
yuliasherman, line 146: Please don't leave 'TODO's in the code. If you plan anything - do it.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,45 +132,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
   for (auto& reply : repliesMap_) {
-    if (earliestTime > reply.second) {
-      earliestReplyId = reply.first;
-      earliestTime = reply.second;
+    if (reply.second == reqOffset) {
+      if (reply.first > reqSeqNum) LOG_WARN(CL_MNGR, "A newer reply was already saved");
```
yuliasherman, line 149: Such kinds of logs are not informative. Please add more info: both sequence numbers, client_id, etc.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,45 +132,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
+  repliesMap_.insert_or_assign(reqSeqNum, requestOffset);
 }
 
 bool ClientsManager::RepliesInfo::findSafe(ReqId reqSeqNum) {
   const lock_guard<mutex> lock(repliesMapMutex_);
   return (repliesMap_.find(reqSeqNum) != repliesMap_.end());
 }
 
+uint16_t ClientsManager::RepliesInfo::getOffsetSafe(ReqId reqSeqNum) {
+  const lock_guard<mutex> lock(repliesMapMutex_);
```
yuliasherman, line 175: Should not be safe!

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -204,12 +210,16 @@ ClientsManager::ClientsManager(const std::set<NodeIdType>& proxyClients,
       clientServices_{clientServices},
       internalClients_{internalClients},
       maxReplySize_(ReplicaConfig::instance().getmaxReplyMessageSize()),
-      maxNumOfReqsPerClient_(
-          ReplicaConfig::instance().clientBatchingEnabled ? ReplicaConfig::instance().clientBatchingMaxMsgsNbr : 1),
+      maxNumOfReqsPerClient_(ReplicaConfig::instance().clientBatchingEnabled &&
+                                     ReplicaConfig::instance().preExecutionFeatureEnabled
```
yuliasherman, line 214: Why to ask about ReplicaConfig::instance().preExecutionFeatureEnabled?

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -265,17 +283,21 @@ void ClientsManager::loadInfoFromReservedPages() {
...
-    ConcordAssert(replyHeader->currentPrimaryId == 0);
-    ConcordAssert(replyHeader->replyLength >= 0);
-    ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
-
-    clientsInfo_[clientId].repliesInfo->deleteOldestReplyIfNeededSafe(clientId, maxNumOfReqsPerClient_);
-    const auto& res = clientsInfo_[clientId].repliesInfo->insertOrAssignSafe(replyHeader->reqSeqNum, MinTime);
-    LOG_INFO(CL_MNGR, "Added/updated reply message" << KVLOG(clientId, replyHeader->reqSeqNum, res));
-    clientsInfo_[clientId].requestsInfo->removeOldPendingReqsSafe(clientId, replyHeader->reqSeqNum);
+    for (uint32_t offsetInBatch = 0; offsetInBatch < maxNumOfReqsPerClient_; offsetInBatch++) {
+      auto pageOffset = getReplyFirstPageId(clientId) + offsetInBatch * reservedPagesPerRequest_;
```
yuliasherman, line 287: Nit: const auto pageOffset

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -358,8 +385,14 @@ std::unique_ptr<ClientReplyMsg> ClientsManager::allocateNewReplyMsgAndWriteToSto
 std::unique_ptr<ClientReplyMsg> ClientsManager::allocateReplyFromSavedOne(NodeIdType clientId,
                                                                           ReqId requestSeqNum,
                                                                           uint16_t currentPrimaryId) {
-  const uint32_t firstPageId = getReplyFirstPageId(clientId);
-  LOG_DEBUG(CL_MNGR, KVLOG(clientId, requestSeqNum, firstPageId));
+  if (!clientsInfo_[clientId].repliesInfo->findSafe(requestSeqNum)) {
+    LOG_FATAL(CL_MNGR, "Reply is not found for specified request" << KVLOG(clientId, requestSeqNum));
```
yuliasherman, line 389: Not sure whether it should be a fatal error.

bftengine/include/bftengine/ClientMsgs.hpp
```diff
@@ -42,6 +42,7 @@ struct ClientRequestMsgHeader {
   uint32_t cidLength = 0;
   uint16_t reqSignatureLength = 0;
   uint32_t extraDataLength = 0;
+  uint16_t offsetInBatch = 0;
```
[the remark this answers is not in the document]
Efrat1: The max value of this variable should be max batch size. uint16_t should be enough (uint16_t max value is 65,535)

bftengine/src/bftengine/ClientsManager.cpp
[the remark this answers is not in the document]
Efrat1: 1. See Yulia's explanation below.
2. Done, thanks.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,45 +132,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
+                                                          uint16_t reqOffset) {
+  ReqId oldReplyId = 0;
   if (repliesMap_.size() > maxNumOfReqsPerClient)
     LOG_FATAL(CL_MNGR,
               "More than maxNumOfReqsPerClient_ items in repliesInfo"
                   << KVLOG(repliesMap_.size(), maxNumOfReqsPerClient, clientId));
+  const lock_guard<mutex> lock(repliesMapMutex_);
+
+  // TODO: create another/additional data structure for efficiency (map offset to seqNum)
   for (auto& reply : repliesMap_) {
```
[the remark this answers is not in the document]
Efrat1: Thanks

bftengine/src/bftengine/ClientsManager.cpp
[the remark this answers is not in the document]
Efrat1: 1. We shouldn't get to this warning. If we do, it means that something's wrong.
2. The offset (the index of request in batch) determines the reserved page in which the reply will be saved. So when we have new reply to be saved, we should remove from map the old reply that is now saved in the same page and will be overridden by the new one.
There cannot be such case as you described. In case the client retransmits a request, s/he re-sends the whole batch, so the request keeps its index in the batch.
3. We expect that the saved reply would be older than the one we are about to save. That's why we have this warning - for the unexpected case where the saved reply is newer than the one we have now.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,45 +132,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
-            "Deleted reply message" << KVLOG(
-                clientId, earliestReplyId, earliestTime.time_since_epoch().count(), repliesMap_.size()));
 }
 
-bool ClientsManager::RepliesInfo::insertOrAssignSafe(ReqId reqSeqNum, Time time) {
+void ClientsManager::RepliesInfo::insertOrAssignSafe(NodeIdType clientId,
+                                                     ReqId reqSeqNum,
+                                                     uint16_t maxNumOfReqsPerClient,
+                                                     uint16_t requestOffset) {
+  deleteReplyIfNeededSafe(clientId, reqSeqNum, maxNumOfReqsPerClient, requestOffset);
```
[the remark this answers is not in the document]
Efrat1: Thanks, I moved the delete out of insert function, as it was before.

bftengine/include/bftengine/IRequestHandler.hpp
```diff
@@ -48,6 +48,7 @@ class IRequestsHandler {
     uint32_t maxReplySize = 0;
     char *outReply = nullptr;
     uint64_t requestSequenceNum = executionSequenceNum;
+    uint16_t reqOffsetInBatch = 0;
```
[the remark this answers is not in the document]
Efrat1: I renamed the offset fields/args/vars to indexInBatch.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,45 +132,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
 
 bool ClientsManager::RepliesInfo::findSafe(ReqId reqSeqNum) {
   const lock_guard<mutex> lock(repliesMapMutex_);
   return (repliesMap_.find(reqSeqNum) != repliesMap_.end());
 }
 
+uint16_t ClientsManager::RepliesInfo::getOffsetSafe(ReqId reqSeqNum) {
+  const lock_guard<mutex> lock(repliesMapMutex_);
+  const auto& replyIt = repliesMap_.find(reqSeqNum);
+  return (replyIt != repliesMap_.end()) ? replyIt->second : 0;
```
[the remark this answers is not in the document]
Efrat1: Thanks. However, we need to create the replyIt var in order to use it...

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -240,11 +250,19 @@ ClientsManager::ClientsManager(const std::set<NodeIdType>& proxyClients,
                         << KVLOG(sizeOfReservedPage(), reservedPagesPerClient_, maxReplySize_, maxNumOfReqsPerClient_));
 }
 
-uint32_t ClientsManager::reservedPagesPerClient(const uint32_t& sizeOfReservedPage, const uint32_t& maxReplySize) {
-  uint32_t reservedPagesPerClient = maxReplySize / sizeOfReservedPage;
+uint32_t ClientsManager::reservedPagesPerRequest(const uint32_t& sizeOfReservedPage, const uint32_t& maxReplySize) {
```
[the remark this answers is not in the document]
Efrat1: Thanks, done.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -265,17 +283,21 @@ void ClientsManager::loadInfoFromReservedPages() {
...
-    ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
-
-    clientsInfo_[clientId].repliesInfo->deleteOldestReplyIfNeededSafe(clientId, maxNumOfReqsPerClient_);
-    const auto& res = clientsInfo_[clientId].repliesInfo->insertOrAssignSafe(replyHeader->reqSeqNum, MinTime);
-    LOG_INFO(CL_MNGR, "Added/updated reply message" << KVLOG(clientId, replyHeader->reqSeqNum, res));
-    clientsInfo_[clientId].requestsInfo->removeOldPendingReqsSafe(clientId, replyHeader->reqSeqNum);
+    for (uint32_t offsetInBatch = 0; offsetInBatch < maxNumOfReqsPerClient_; offsetInBatch++) {
+      auto pageOffset = getReplyFirstPageId(clientId) + offsetInBatch * reservedPagesPerRequest_;
+      if (!loadReservedPage(pageOffset, sizeOfReservedPage(), scratchPage_.data())) continue;
+      ClientReplyMsgHeader* replyHeader = (ClientReplyMsgHeader*)scratchPage_.data();
```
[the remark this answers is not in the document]
Efrat1: Thanks

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -265,17 +283,21 @@ void ClientsManager::loadInfoFromReservedPages() {
...
+      ConcordAssert(replyHeader->msgType == 0 || replyHeader->msgType == MsgCode::ClientReply);
+      ConcordAssert(replyHeader->currentPrimaryId == 0);
+      ConcordAssert(replyHeader->replyLength >= 0);
```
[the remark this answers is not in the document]
Efrat1: Removed, thanks

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -265,17 +283,21 @@ void ClientsManager::loadInfoFromReservedPages() {
...
+      ConcordAssert(replyHeader->replyLength + sizeof(ClientReplyMsgHeader) <= maxReplySize_);
+
+      clientsInfo_[clientId].repliesInfo->insertOrAssignSafe(
```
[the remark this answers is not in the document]
Efrat1: Done, thanks

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -290,8 +312,9 @@ bool ClientsManager::hasReply(NodeIdType clientId, ReqId reqSeqNum) {
   }
 }
 
-void ClientsManager::deleteOldestReply(NodeIdType clientId) {
-  clientsInfo_[clientId].repliesInfo->deleteOldestReplyIfNeededSafe(clientId, maxNumOfReqsPerClient_);
+void ClientsManager::deleteReplyIfNeeded(NodeIdType clientId, uint16_t offsetInBatch, ReqId newReqSeqNum) {
+  clientsInfo_[clientId].repliesInfo->deleteReplyIfNeededSafe(
```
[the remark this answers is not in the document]
Efrat1: Added try-catches, thanks.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -302,14 +325,18 @@ void ClientsManager::deleteOldestReply(NodeIdType clientId) {
...
                                                                                      char* reply,
                                                                                      uint32_t replyLength,
                                                                                      uint32_t rsiLength,
                                                                                      uint32_t executionResult) {
-  clientsInfo_[clientId].repliesInfo->deleteOldestReplyIfNeededSafe(clientId, maxNumOfReqsPerClient_);
-  clientsInfo_[clientId].repliesInfo->insertOrAssignSafe(requestSeqNum, getMonotonicTime());
-  LOG_DEBUG(CL_MNGR, KVLOG(clientId, requestSeqNum));
+  ConcordAssert(reqOffsetInBatch < maxNumOfReqsPerClient_);
+  uint16_t requestOffset = reqOffsetInBatch * reservedPagesPerRequest_;
+  clientsInfo_[clientId].repliesInfo->insertOrAssignSafe(
```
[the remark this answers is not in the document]
Efrat1: Thanks

bftengine/src/bftengine/ClientsManager.hpp
Efrat1: Done, thanks

bftengine/src/bftengine/ClientsManager.hpp
Efrat1: uint16_t should be good enough, see my comment above.

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -102,6 +102,7 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
   //   ClientsManager's construction.
   std::unique_ptr<ClientReplyMsg> allocateNewReplyMsgAndWriteToStorage(NodeIdType clientId,
                                                                        ReqId requestSeqNum,
+                                                                       uint16_t reqOffsetInBatch,
```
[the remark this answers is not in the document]
Efrat1: Thanks

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -193,7 +193,10 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
   void setClientPublicKey(NodeIdType, const std::string& key, concord::util::crypto::KeyFormat) override;
 
   // General
-  static uint32_t reservedPagesPerClient(const uint32_t& sizeOfReservedPage, const uint32_t& maxReplySize);
+  static uint32_t reservedPagesPerRequest(const uint32_t& sizeOfReservedPage, const uint32_t& maxReplySize);
```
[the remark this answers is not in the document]
Efrat1: Done, thanks.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -63,8 +63,8 @@ void ClientsManager::RequestsInfo::removeOldPendingReqsSafe(NodeIdType clientId,
   const lock_guard<mutex> lock(requestsMapMutex_);
   for (auto it = requestsMap_.begin(); it != requestsMap_.end();) {
     if (it->first <= reqSeqNum) {
+      LOG_INFO(CL_MNGR, "Remove old pending request" << KVLOG(clientId, it->first, reqSeqNum));
```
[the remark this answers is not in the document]
Efrat1: Done, thanks.

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Thanks for the explanation.

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: You're right, done.

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Added info. Thanks.

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Done, thanks!

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: As far as I see, client batch is possible only when pre-processing is enabled. I added this to reduce the number of reserved pages when it's possible.

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Thanks

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Changed to warning, thanks.

bftengine/src/bftengine/ClientsManager.cpp
yuliasherman: Exactly, that's why there is no need to ask if preExecutionFeatureEnabled or not.

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -244,15 +249,19 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
...
+    void insertOrAssignSafe(ReqId reqSeqNum, uint16_t reqIndexInBatch);
     bool findSafe(ReqId reqSeqNum);
 
     bool find(ReqId reqSeqNum) const;
+    uint16_t getIndex(ReqId reqSeqNum) const;
 
    public:
     std::mutex repliesMapMutex_;
-    std::map<ReqId, Time> repliesMap_;
+    boost::bimaps::bimap<ReqId, uint16_t> repliesBiMap_;
```
yuliasherman, line 264: Add a comment explaining what field 'uint16_t' represents here.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -69,95 +69,107 @@
 using concordUtil::Timers;
 using namespace std;
 using namespace std::chrono;
-using namespace std::placeholders;
+// using namespace std::placeholders;
```
yuliasherman, line 72: Remove, if unneeded.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5712,10 +5733,13 @@ void ReplicaImp::sendResponses(PrePrepareMsg *ppMsg, IRequestsHandler::Execution
...
                                                                     req.outActualReplySize,
+                                                                    req.reqIndexInBatch,
                                                                     0,
                                                                     executionResult);
-    send(replyMsg.get(), req.clientId);
-    free(req.outReply);
+    if (replyMsg) {
+      send(replyMsg.get(), req.clientId);
+      free(req.outReply);
+    }
```
yuliasherman, line 5742: If you add the 'if' condition, you need to handle the 'else' one, too. At least, add the log.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,47 +133,51 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
-    // Delete reply arrived through ST
-    auto const& reply = repliesMap_.cbegin();
-    earliestReplyId = reply->first;
-    earliestTime = reply->second;
-    repliesMap_.erase(reply);
+
+  if (savedReplyId) {
+    // The thread safety here is between the PreProcessor thread, which is read-only, and the ReplicaImp main thread,
+    // which performs either read or write at a time. So, all write operations should be guarded as well as PrePropessor
+    // calls hasReply and isClientRequestInProcess. All other operations are safe.
```
yuliasherman, line 157: I would move this comment to the hpp file and put it before 'safe' functions of the RepliesInfo class.

bftengine/include/bftengine/ClientMsgs.hpp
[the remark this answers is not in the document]
glevkovich: Are you planning to change offsetInBatch to something like indexInBatch?
I was thinking the offset is in bytes, which looked dangerouse to me, for 2^16 indexes, I think we are fine.

bftengine/src/bftengine/messages/ClientRequestMsg.hpp
```diff
@@ -42,7 +42,8 @@ class ClientRequestMsg : public MessageBase {
                    const concordUtils::SpanContext& spanContext = concordUtils::SpanContext{},
                    const char* requestSignature = nullptr,
                    uint32_t requestSignatureLen = 0,
-                   const uint32_t extraBufSize = 0);
+                   const uint32_t extraBufSize = 0,
```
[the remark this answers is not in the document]
Efrat1: Done

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -62,9 +62,10 @@ bool ClientsManager::RequestsInfo::removeRequestsOutOfBatchBoundsSafe(NodeIdType
 void ClientsManager::RequestsInfo::removeOldPendingReqsSafe(NodeIdType clientId, ReqId reqSeqNum) {
   const lock_guard<mutex> lock(requestsMapMutex_);
   for (auto it = requestsMap_.begin(); it != requestsMap_.end();) {
-    if (it->first <= reqSeqNum) {
+    auto oldSeqNum = it->first;
```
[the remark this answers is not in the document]
Efrat1: Done

bftengine/src/bftengine/ClientsManager.hpp
Efrat1: ok

bftengine/src/bftengine/ReplicaImp.cpp
Efrat1: Right, removed.

bftengine/src/bftengine/ReplicaImp.cpp
Efrat1: Added logs

bftengine/include/bftengine/IRequestHandler.hpp
```diff
@@ -48,6 +48,7 @@ class IRequestsHandler {
     uint32_t maxReplySize = 0;
     char *outReply = nullptr;
     uint64_t requestSequenceNum = executionSequenceNum;
+    uint16_t reqIndexInBatch = 0;
```
cloudnoize, line 51: if this field is used by the pre-process, please rename it to reflect it.

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,47 +133,48 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
...
-  Time earliestTime = MaxTime;
-  ReqId earliestReplyId = 0;
-  if (repliesMap_.size() < maxNumOfReqsPerClient) return;
-  if (repliesMap_.size() > maxNumOfReqsPerClient)
+void ClientsManager::RepliesInfo::deleteReplyIfNeededSafe(NodeIdType clientId,
+                                                          ReqId reqSeqNum,
+                                                          uint16_t maxNumOfReqsPerClient,
+                                                          uint16_t reqIndex) {
+  ReqId savedReplyId = 0;
+  if (repliesBiMap_.size() > maxNumOfReqsPerClient) {
```
cloudnoize, line 141: what is the meaning of the ```Bi``` in ```repliesBiMap_```

bftengine/src/bftengine/ClientsManager.cpp
```diff
@@ -132,47 +133,48 @@ void ClientsManager::RequestsInfo::logAllPendingRequestsExceedingThreshold(const
 
 /*************************** Class ClientsManager::RepliesInfo ***************************/
 
-void ClientsManager::RepliesInfo::deleteOldestReplyIfNeededSafe(NodeIdType clientId, uint16_t maxNumOfReqsPerClient) {
-  Time earliestTime = MaxTime;
-  ReqId earliestReplyId = 0;
-  if (repliesMap_.size() < maxNumOfReqsPerClient) return;
-  if (repliesMap_.size() > maxNumOfReqsPerClient)
+void ClientsManager::RepliesInfo::deleteReplyIfNeededSafe(NodeIdType clientId,
```
cloudnoize, line 136: do we have a unit test for this?

Efrat1: Yes, in clientsManager_test.cpp

bftengine/src/bftengine/ClientsManager.hpp
```diff
@@ -244,15 +249,22 @@ class ClientsManager : public ResPagesClient<ClientsManager>, public IPendingReq
...
+    void insertOrAssignSafe(ReqId reqSeqNum, uint16_t reqIndexInBatch);
     bool findSafe(ReqId reqSeqNum);
 
     bool find(ReqId reqSeqNum) const;
+    uint16_t getIndex(ReqId reqSeqNum) const;
 
    public:
     std::mutex repliesMapMutex_;
-    std::map<ReqId, Time> repliesMap_;
+    boost::bimaps::bimap<ReqId, uint16_t> repliesBiMap_;  // mapping request sequence number to request index in batch
```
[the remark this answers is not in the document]
Efrat1: Right, it contains unique sequence numbers and unique indices.
The purpose of this structure here is to get the index given the seqNum and vice versa (to avoid map loop when we have the seqNum, as it was before).
Regarding the uniqueness - a sequence number will always have the same index because if a request was sent in a batch, the client will re-send it, if needed, as part of the same batch retransmit, so the request always keeps its index. (If the request was sent without batch, its index is 0 and will be the same when the request is retransmit.)
The indices are also unique within the repliesBiMap. This BiMap holds the currently saved replies in the reserved pages, so it's not possible to have two replies saved on the same page. deleteReplyIfNeededSafe() ensure this by deleting the overridden reply.

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Bi-directional map

bftengine/include/bftengine/IRequestHandler.hpp
Efrat1: It's passed through the consensus and finally used in ReplicaImp as an input to ClientsManager.
Maybe would be clearer if I rename it to reqIndexIn**Client**Batch

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Theoretically it's possible to set in configurations _clientBatchingEnabled=True_ along with _preExecutionFeatureEnabled=False_.
I agree that in this case the _clientBatchingEnabled=True_ doesn't have a reasonable meaning but I don't see that we force anywhere clientBatchingEnabled to be False when pre-executing is disable. 
So this "extra" checking is for that not-smart-but-possible case.

bftengine/include/bftengine/ClientMsgs.hpp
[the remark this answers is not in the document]
Efrat1: Done, thanks

bftengine/src/bftengine/ClientsManager.cpp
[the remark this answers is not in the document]
Efrat1: 1. We don't expect this warning to happen. request should arrive in ascending order
2. There is 1:1 mapping between page id and offset (index in batch).
    A client request will always have the same index in batch because since a retransmit of a batch requests would always be a retransmit of the whole batch.
3. We expect to see an older reply in map

bftengine/src/bftengine/ClientsManager.cpp
Efrat1: Used LOG_DEBUG instead

eyalrund (approved):

glevkovich (approved):
LGTM

PR #2729, opened 2022-07-11
Fix for replica crash in collecting state

situ-s:
* **Problem Overview**  
Replica crashed inside function tryToStartOrFinishExecution.
` FATAL|4|concord.bft|message-processing|||4|slow|ReplicaImp.cpp:4705| +Assert: expression '!isCollectingState()' is false in function tryToStartOrFinishExecution`

When tryToStartOrFinishExecution is called, the replica should not be in collecting state.
The way of finding out if replica is in collecting state causes the issue as we have state transfer in different thread now.
So even if we are checking if replica is not in state stander few sec back we cannot be sure,  that after few sec also it will not be in state transfer.

* **Testing Done**  

Tested on 200 TPS, no crash found.
Tested perf-run on 650 TPS and no crash found.
Ran Manual state transfers on 1000 TPS and no crash was found on replica.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -2416,14 +2416,11 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
     }
   }
 
-  if (askForStateTransfer && !stateTransfer->isCollectingState()) {
+  if (askForStateTransfer && !isCollectingState()) {
     if (activeExecutions_ > 0)
       isStartCollectingState_ = true;
     else {
-      LOG_INFO(GL, "Call to startCollectingState()");
-      time_in_state_transfer_.start();
-      clientsManager->clearAllPendingRequests();  // to avoid entering a new view on old request timeout
-      stateTransfer->startCollectingState();
+      startCollectingState("On receiving checkpoint message");
     }
   } else if (msgSenderId == msgGenReplicaId) {
     if (msgSeqNum > lastStableSeqNum + kWorkWindowSize) {
@@ -3354,6 +3351,7 @@ void ReplicaImp::onTransferringCompleteImp(uint64_t newStateCheckpoint) {
     if (ps_) {
       ps_->endWriteTran(config_.getsyncOnUpdateOfMetadata());
     }
+    setIsCollectingState(false);
     return;
   }
   lastExecutedSeqNum = newCheckpointSeqNum;
@@ -3419,6 +3417,8 @@ void ReplicaImp::onTransferringCompleteImp(uint64_t newStateCheckpoint) {
     LOG_INFO(GL, "tryToEnterView after State Transfer finished ...");
     tryToEnterView();
   }
+
+  setIsCollectingState(false);
 }
 
 void ReplicaImp::onSeqNumIsSuperStable(SeqNum superStableSeqNum) {
@@ -4285,6 +4285,7 @@ ReplicaImp::ReplicaImp(bool firstTime,
                        shared_ptr<PersistentStorage> ps,
                        const std::function<void(bool)> &viewChangeCallBack)
     : ReplicaForStateTransfer(config, requestsHandler, stateTrans, msgsCommunicator, msgHandlers, firstTime, timers),
+      isCollectingState_{stateTransfer->isCollectingState()},
       viewChangeProtocolEnabled{config.viewChangeProtocolEnabled},
       autoPrimaryRotationEnabled{config.autoPrimaryRotationEnabled},
       restarted_{!firstTime},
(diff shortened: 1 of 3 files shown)


eyalrund:
please add more details for problem and solution

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3415,6 +3415,8 @@ void ReplicaImp::onTransferringCompleteImp(uint64_t newStateCheckpoint) {
 
   sendToAllOtherReplicas(checkpointMsg);
 
+  isCollectingState_ = false;
```
glevkovich, line 3418: 1) I recommend, for being able to debug, to add an inline function setIsCollectingState() :
```
void ReplicaImp::setIsCollectingState(bool newState) {
  LOG_INFO(GL,  std::boolalpha << "Setting CollectingState to" << KVLOG(newState));
 isCollectingState_ = newState;
}
```

2) You missed an important place,  see line 3357.

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -368,7 +369,7 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
   std::function<bool(MessageBase*)> getMessageValidator();
 
   // InternalReplicaApi
-  bool isCollectingState() const override { return stateTransfer->isCollectingState(); }
+  bool isCollectingState() const override { return isCollectingState_; }
```
glevkovich, line 372: This one is fine, the wrapping of the function saved you the need from replacing all calls.
But I think you've missed something  we've discussed offline - you only set isCollectingState_ on init, but you do not set isCollectingState_ to true when starting ST, so nothing here will really work.
You need to create a wrapper function ReplicaImp::startCollectingState, call stateTransfer->startCollectingState() and set isCollectingState_=true.
Please contact me if you dont fully/clearly understand my explenation.

glevkovich (changes requested):

bftengine/src/bftengine/ReplicaImp.cpp
situ-s: made the changes. Thanks.

bftengine/src/bftengine/ReplicaImp.hpp
situ-s: made the changes.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3333,6 +3333,11 @@ void ReplicaImp::tryToMarkCheckpointStableForFastPath(const SeqNum &lastCheckpoi
   sendToAllOtherReplicas(checkpointMessage, true);
 }
 
+void ReplicaImp::setIsCollectingState(bool newState) {
```
glevkovich, line 3336: I think it would be better to move this one to header, and use inline specifier.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3415,6 +3421,8 @@ void ReplicaImp::onTransferringCompleteImp(uint64_t newStateCheckpoint) {
 
   sendToAllOtherReplicas(checkpointMsg);
 
+  setIsCollectingState(false);
```
glevkovich, line 3424: You want to set this as last line in the function.
Previous behavior: isCollecting state called directly to ST would return true, right?We want the same behavior so we set to false as last line.
If you put this line here, you change behavior since tryToEnterView execute when collecting state is false.
I dont have good explenation why we run all this when is collecting state is still set to false, need ot check.
Also, it would be better to put a comment that this one should be last in the function.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2423,7 +2423,7 @@ void ReplicaImp::onMessage<CheckpointMsg>(CheckpointMsg *msg) {
       LOG_INFO(GL, "Call to startCollectingState()");
       time_in_state_transfer_.start();
       clientsManager->clearAllPendingRequests();  // to avoid entering a new view on old request timeout
-      stateTransfer->startCollectingState();
+      startCollectingState();
```
glevkovich, line 2426: All the above section is problematic and need ot to refactored and modified:

- what is left should be inserted inside ReplicaImp::startCollectingState, since the same code repeats in few places.
- you already have a member called isStartCollectingState_, need to remove it and modify code accordingly.
- time_in_state_transfer_ probably need to be modified inside startCollectingState
- and more
- startCollectingState should get a string reason, so we know who called it and the reason. Then the above log  "call to startCollectingState()" should be removed. Do this to all calls to startCollectingState().

glevkovich (changes requested):

mdarade (commented):
Please write a test to reproduce assert seen in BC-23868 and validate that this PR fixes it.

mdarade (approved):
LGTM.
Thanks Situ.

bftengine/src/bftengine/ReplicaImp.hpp
```diff
@@ -382,6 +387,12 @@ class ReplicaImp : public InternalReplicaApi, public ReplicaForStateTransfer {
   bool isClientRequestInProcess(NodeIdType clientId, ReqId reqSeqNum) const override {
     return clientsManager->isClientRequestInProcess(clientId, reqSeqNum);
   }
+  inline void setIsCollectingState(bool newState) {
```
glevkovich, line 390: nit:
setIsCollectingState -> setCollectingState ?

glevkovich (approved):
LGTM. wrote 1 small comment, thanks @situ-s!

nityavasudevan:
Thanks for the diligence Situ

eyalrund:
Thanks Situ, Can we have a 5 trading days run with this fix?

eyalrund (approved):

situ-s:
> Thanks Situ, Can we have a 5 trading days run with this fix?

Yes Eyal, we had 5 trading days run with Maestro. Thanks.

PR #2732, opened 2022-07-12
Makefile: modify target test-single-suite to run N times and break

glevkovich:
* **Problem Overview**  
Adding the same options seen in test-single-apollo-case and
test-single-gtest-case, NUM_REPEATS and BREAK_ON_FAILURE.

Logic is implemented using bash in target, and not ctest - ctest options
(--repeat-until-fail, --repeat <mode>:<n>) are not exactly what we need
for current logic.

* **Testing Done**  
Manual testing: checked succesfull scenario and faulty one.


harsht1737 (approved):
LGTM

eyalrund (approved):

PR #2734, opened 2022-07-13
Optimize immutable writes

cloudnoize:
Immutable keys by agreement are not being read by getLatest API but by get with block id.
Therefore saving their value in the latest CF is redundant.
For safety, in case the application tries to read the latest of on immutable, we'll save the version of the immutable in the latest CF, and perform indirection by first reading its version and then reading the value from the block.

A counter metric was added in order to quickly observe the usage of the getLatest API with immutables.
Tests:
UT that use getLatest for immutable pass.
Apollo and Concord CI.
Running long runs:

- Reduciton of ~25% in storage
- no calls for getLatest with immutable

### DIFF
--- a/kvbc/src/v4blockchain/v4_blockchain.cpp
+++ b/kvbc/src/v4blockchain/v4_blockchain.cpp
@@ -38,7 +38,8 @@ KeyValueBlockchain::KeyValueBlockchain(
       v4_metrics_comp_{concordMetrics::Component("v4_blockchain", std::make_shared<concordMetrics::Aggregator>())},
       blocks_deleted_{v4_metrics_comp_.RegisterGauge(
           "numOfBlocksDeleted", block_chain_.getGenesisBlockId() > 0 ? (block_chain_.getGenesisBlockId() - 1) : 0)},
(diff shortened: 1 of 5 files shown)


kvbc/src/v4blockchain/v4_blockchain.cpp
```diff
@@ -578,7 +587,21 @@ void KeyValueBlockchain::multiGet(const std::string &category_id,
...
                                         std::vector<std::optional<categorization::Value>> &values) const {
-  return latest_keys_.multiGetValue(category_id, keys, values);
+  auto category_type = latest_keys_.categoryType(category_id);
+  if (category_type == concord::kvbc::categorization::CATEGORY_TYPE::immutable) {
+    immutables_reads_++;
+    v4_metrics_comp_.UpdateAggregator();
+    std::vector<std::optional<categorization::TaggedVersion>> tagged_versions;
+    latest_keys_.multiGetLatestVersion(category_id, keys, tagged_versions);
+    std::vector<BlockId> versions;
+    std::for_each(
```
arc-vmware, line 597: this is nice !!!

arc-vmware (approved):
thanks @cloudnoize , very good code.

eyalrund (approved):

PR #2738, opened 2022-07-18
Get rid of unneeded memory allocations in the client pool

yuliasherman:
* **Problem Overview**  
  Clientservice memory usage is currently quite high. This is caused by two main factors: 
  1. Static memory buffer allocation per client/batch/request - will be optimized in this PR
  2. TCP buffer allocation per client/replica - will be optimized when we move to the multiplex channel
  After we moved to a new BFT client, we stopped providing memory to it to be filled by the replicas response. 
  Thus we don't need to pre-allocate static memory chunks anymore. In case the memory buffer has been 
  provided by the caller (e.g. for the performance tests), this buffer is used to return the response to the caller, as before.
* **Testing Done**  
  Full 5 days run on a reserved system


toly-kournik (approved):
LGTM

PR #2739, opened 2022-07-19
UTT new interface (WIP)

yontyon:
Introducing a new interface to the UTT library to be used by less expert users.
The main goals of this interface are:
1. Expose as less as functionality as possible
2. Use the simplest input parameters types

The following are worth noticing:
1. This implementation also includes a distributed registration functionality
2. The PKI is still based on IBE
3. Cryptographic curve points are represented by std::vector<uint64_t>
4. Signatures are represented as std::vetor<uint8_t>
5. The implementation exposes the "newutt" static lib
6. This PR includes tests for testing the functionality

### DIFF
--- a/utt/include/clientIdentity.hpp
+++ b/utt/include/clientIdentity.hpp
@@ -1,0 +1,31 @@
+#pragma once
+#include "commitment.hpp"
+#include "details.hpp"
+#include "common.hpp"
+#include <memory>
+namespace libutt {
+class AddrSK;
+class RandSigPK;
+}  // namespace libutt
+namespace libutt::api {
+class ClientIdentity {
+ public:
+  ClientIdentity(const std::string& pid, const std::string& bpk);
+  Commitment generatePartialRCM(Details& d);
+  std::string getPid() const;
+  std::vector<uint64_t> getPRFSecretKey() const;
+  std::vector<uint64_t> getPidHash() const;
+  void setIBEDetails(const std::vector<uint8_t>& sk, const std::vector<uint8_t>& mpk);
+  void setRCM(const Commitment& comm, const std::vector<uint8_t>& sig);
+  void randomizeRCM();
+  std::pair<Commitment, std::vector<uint8_t>> getRcm() const;
+  template <typename T>
+  bool validate(const T&);
+
+ private:
+  std::unique_ptr<libutt::AddrSK> ask_;
+  std::unique_ptr<libutt::RandSigPK> bpk_;
+  Commitment rcm_;
+  std::vector<uint8_t> rcm_sig_;
+};
+}  // namespace libutt::api
--- a/utt/include/coin.hpp
+++ b/utt/include/coin.hpp
@@ -12,7 +12,12 @@ class ClientIdentity;
 class Coin {
  public:
   enum Type { Normal = 0x0, Budget };
-  Coin(Details& d, const std::vector<uint64_t>& sn, const std::vector<uint64_t>& val, Type p, const std::vector<uint64_t>& exp_date, ClientIdentity& cid);
+  Coin(Details& d,
+       const std::vector<uint64_t>& sn,
+       const std::vector<uint64_t>& val,
+       Type p,
+       const std::vector<uint64_t>& exp_date,
+       ClientIdentity& cid);
   const Nullifier& getNullifier() const;
   bool hasSig() const;
   void setSig(const std::vector<uint8_t>& sig);
--- a/utt/include/coin.hpp
+++ b/utt/include/coin.hpp
@@ -1,0 +1,31 @@
+#pragma once
+#include "details.hpp"
+#include "clientIdentity.hpp"
+#include "nullifier.hpp"
+#include "commitment.hpp"
+#include <memory>
+namespace libutt {
+class Coin;
+}
+namespace libutt::api {
+class ClientIdentity;
+class Coin {
+ public:
+  enum Type { Normal = 0x0, Budget };
+  Coin(Details& d, const std::vector<uint64_t>& sn, const std::vector<uint64_t>& val, Type p, const std::vector<uint64_t>& exp_date, ClientIdentity& cid);
+  const Nullifier& getNullifier() const;
+  bool hasSig() const;
+  void setSig(const std::vector<uint8_t>& sig);
+  Type getType() const;
+  std::vector<uint8_t> getSig() const;
+  void randomize();
+
+ private:
+  friend class ClientIdentity;
+  std::unique_ptr<libutt::Coin> coin_;
+  std::unique_ptr<Nullifier> nullifier_;
+  bool has_sig_{false};
+
+  Type type_;
+};
+}  // namespace libutt::api
--- a/utt/include/commitment.hpp
+++ b/utt/include/commitment.hpp
@@ -1,0 +1,38 @@
+#pragma once
+#include "details.hpp"
+#include <vector>
+#include <cstdint>
+namespace libutt {
+class Comm;
+}
+namespace libutt::api {
+class Commitment;
+}
+libutt::api::Commitment operator+(libutt::api::Commitment lhs, const libutt::api::Commitment& rhs);
+namespace libutt::api {
+class RegistratorIdentity;
+class ClientIdentity;
+class Commitment {
+ public:
+  enum Type { REGISTRATION = 0, VALUE, COIN };
+  static std::vector<Commitment> create(Details& d,
+                                        Type t,
+                                        const std::vector<std::vector<uint64_t>>& messages,
+                                        std::vector<std::vector<uint64_t>>& randomizations,
+                                        bool withG2);
+  Commitment(Details& d, Type t, const std::vector<std::vector<uint64_t>>& messages, bool withG2);
+  Commitment(const std::string& comm);
+  Commitment(const Commitment& comm);
+  Commitment();
+  Commitment& operator=(const Commitment&);
+  Commitment& operator+=(const Commitment&);
+  std::vector<uint64_t> randomize(Details& d, Type t);
+  static size_t getCommitmentSn(const Commitment& comm);
+  static std::string getCommitmentHash(const Commitment& comm);
+
+ private:
+  friend class RegistratorIdentity;
+  friend class ClientIdentity;
+  std::shared_ptr<libutt::Comm> comm_;
+};
+}  // namespace libutt::api
--- a/utt/include/common.hpp
+++ b/utt/include/common.hpp
@@ -1,0 +1,25 @@
+#pragma once
+
+#include "commitment.hpp"
+#include "details.hpp"
+#include <memory>
+#include <vector>
+#include <map>
+
+namespace libutt::api {
+struct RegistrationDetails {
+  Commitment rcm_;
+  std::vector<uint8_t> rcm_sig_;
+  std::vector<uint8_t> dsk_pub_;
+  std::vector<uint8_t> dsk_priv_;
+};
+
+class Utils {
+ public:
+  static std::vector<uint8_t> aggregateSigShares(Details& d,
+                                                 Commitment::Type t,
+                                                 uint32_t n,
+                                                 const std::map<uint32_t, std::vector<uint8_t>>& rsigs,
+                                                 std::vector<std::vector<uint64_t>> randomness);
+};
+}  // namespace libutt::api
--- a/utt/include/mint.hpp
+++ b/utt/include/mint.hpp
@@ -1,0 +1,29 @@
+#pragma once
+#include <string>
+#include <memory>
+#include <vector>
+#include "coin.hpp"
+#include "clientIdentity.hpp"
+#include "details.hpp"
+namespace libutt {
+class MintOp;
+}  // namespace libutt
+namespace libutt::api {
+class BankIdentity;
+}
+namespace libutt::api::operations {
+
+class Mint {
+ public:
+  Mint(const std::string& uniqueHash, size_t value, const std::string& recipPID);
+  bool validate(const std::string& uniqueHash, size_t value, const std::string& recipPID) const;
+  Coin claimCoin(Details& d,
+                 ClientIdentity& cid,
+                 uint32_t n,
+                 const std::map<uint32_t, std::vector<uint8_t>>& rsigs) const;
+
+ private:
+  friend class libutt::api::BankIdentity;
+  std::unique_ptr<libutt::MintOp> op_;
+};
+}  // namespace libutt::api::operations
--- a/utt/include/registratorIdentity.hpp
+++ b/utt/include/registratorIdentity.hpp
@@ -1,0 +1,29 @@
+#pragma once
+#include "commitment.hpp"
+#include "common.hpp"
+#include "details.hpp"
+#include <string>
+#include <memory>
+#include <vector>
+namespace libutt {
+class RegAuthShareSK;
+class RegAuthPK;
+}  // namespace libutt
+namespace libutt::api {
+class RegistratorIdentity {
+ public:
+  RegistratorIdentity(const std::string& id, const std::string& rsk, const std::string& rbk);
+  RegistrationDetails registerClient(Details& d,
+                                     const std::string& pid,
+                                     const std::vector<uint64_t>& pid_hash,
+                                     const Commitment& partiail_comm) const;
+  bool validateRCM(const Commitment& comm, const std::vector<uint8_t>& sig);
+
+ private:
+  std::vector<uint8_t> generateSecretForPid(const std::string& pid) const;
+  std::vector<uint8_t> getMPK() const;
+  std::string id_;
+  std::unique_ptr<RegAuthShareSK> rsk_;
+  std::unique_ptr<RegAuthPK> rpk_;
+};
+}  // namespace libutt::api
--- a/utt/libutt/include/utt/Coin.h
+++ b/utt/libutt/include/utt/Coin.h
@@ -75,7 +75,8 @@ class Coin {
  public:
   // WARNING: Used only for deserialization
   Coin() {}
-
+  // Used to create the coin's partial commitment (without nullfier)
+  Coin(const CommKey& ck, const Fr& sn, const Fr& val, const Fr& type, const Fr& exp_date, const Fr& pidHash);
   // Used in TxOut::tryMakeMine and Factory
   Coin(const CommKey& ck,
        const Nullifier::Params& np,
--- a/utt/libutt/src/MintOp.cpp
+++ b/utt/libutt/src/MintOp.cpp
@@ -8,10 +8,10 @@
 #include <utt/Comm.h>
 #include <utt/Params.h>
 #include <utt/MintOp.h>
+#include <utt/RandSig.h>
 
-#include <utt/Serialization.h>
-
-namespace libutt {
+#include <xassert/XAssert.h>
+#include <xutils/Log.h>
 
 std::ostream& operator<<(std::ostream& out, const libutt::MintOp& op) {
   out << op.sn << endl;
@@ -29,17 +29,19 @@ std::istream& operator>>(std::istream& in, libutt::MintOp& op) {
   return in;
 }
 
+namespace libutt {
+
 static Fr createSN(std::string uniqueHash) { return hashToField("new sn|" + uniqueHash); }
 
-MintOp::MintOp(std::string uniqueHash, size_t v, const std::string& recipPID) {
+MintOp::MintOp(const std::string& uniqueHash, size_t v, const std::string& recipPID) {
   testAssertGreaterThanOrEqual(v, 1);
 
   sn = createSN(uniqueHash);
   pidHash = AddrSK::pidHash(recipPID);
   value.set_ulong(static_cast<unsigned long>(v));
 }
-
-bool MintOp::validate(std::string uniqueHash, size_t v, std::string recipPID) const {
+MintOp::MintOp(std::istream& in) { in >> *this; }
+bool MintOp::validate(const std::string& uniqueHash, size_t v, const std::string& recipPID) const {
   Fr snV = createSN(uniqueHash);
   Fr pidHashV = AddrSK::pidHash(recipPID);
   Fr valueV;
--- a/utt/libutt/src/clientIdentity.cpp
+++ b/utt/libutt/src/clientIdentity.cpp
@@ -1,0 +1,56 @@
+#include "clientIdentity.hpp"
+#include "coin.hpp"
+#include <utt/IBE.h>
+#include <utt/Address.h>
+#include <utt/RandSig.h>
+#include <utt/Params.h>
+#include <utt/Serialization.h>
+#include <utt/Coin.h>
+#include <vector>
+#include <sstream>
+namespace libutt::api {
+ClientIdentity::ClientIdentity(const std::string& pid, const std::string& bpk) {
+  ask_.reset(new libutt::AddrSK());
+  ask_->pid = pid;
+  ask_->s = Fr::random_element();
+  ask_->pid_hash = AddrSK::pidHash(pid);
+  bpk_.reset(new libutt::RandSigPK());
+  *bpk_ = libutt::deserialize<libutt::RandSigPK>(bpk);
+}
+Commitment ClientIdentity::generatePartialRCM(Details& d) {
+  std::vector<std::vector<uint64_t>> m = {getPidHash(), ask_->s.to_words(), Fr::zero().to_words()};
+  auto comm = Commitment(d, Commitment::Type::REGISTRATION, m, true);
+  return comm;
+}
+
+std::string ClientIdentity::getPid() const { return ask_->pid; }
+std::vector<uint64_t> ClientIdentity::getPidHash() const { return ask_->getPidHash().to_words(); }
+std::vector<uint64_t> ClientIdentity::getPRFSecretKey() const { return ask_->s.to_words(); }
+
+void ClientIdentity::setIBEDetails(const std::vector<uint8_t>& sk, const std::vector<uint8_t>& mpk) {
+  ask_->e = libutt::deserialize<libutt::IBE::EncSK>(std::string(sk.begin(), sk.end()));
+  ask_->mpk_ = libutt::deserialize<libutt::IBE::MPK>(std::string(mpk.begin(), mpk.end()));
+}
+
+void ClientIdentity::setRCM(const Commitment& comm, const std::vector<uint8_t>& sig) {
+  rcm_ = comm;
+  rcm_sig_ = sig;
+  ask_->rs = libutt::deserialize<libutt::RandSig>(sig);
+}
+
+std::pair<Commitment, std::vector<uint8_t>> ClientIdentity::getRcm() const { return {rcm_, rcm_sig_}; }
+template <>
+bool ClientIdentity::validate<Coin>(const Coin& c) {
+  return c.coin_->hasValidSig(*bpk_);
+}
+void ClientIdentity::randomizeRCM() {
+  auto r_delta = rcm_.randomize(Details::instance(), Commitment::Type::REGISTRATION);
+  Fr fr_r_delta;
+  fr_r_delta.from_words(r_delta);
+  auto rand_sig = libutt::deserialize<libutt::RandSig>(rcm_sig_);
+  rand_sig.rerandomize(fr_r_delta, Fr::random_element());
+  auto tmp_sig = libutt::serialize<libutt::RandSig>(rand_sig);
+  rcm_sig_ = std::vector<uint8_t>(tmp_sig.begin(), tmp_sig.end());
+}
+
+}  // namespace libutt::api
--- a/utt/libutt/src/coin.cpp
+++ b/utt/libutt/src/coin.cpp
@@ -1,0 +1,37 @@
+#include "coin.hpp"
+#include <utt/Coin.h>
+#include <utt/Params.h>
+#include <utt/Serialization.h>
+#include <utt/RandSig.h>
+
+namespace libutt::api {
+Coin::Coin(Details& d, const std::vector<uint64_t>& sn, const std::vector<uint64_t>& val, Type t, const std::vector<uint64_t>& exp_date, ClientIdentity& cid) {
+  Fr fr_sn;
+  fr_sn.from_words(sn);
+  Fr fr_val;
+  fr_val.from_words(val);
+  Fr fr_type = t == Type::Normal ? libutt::Coin::NormalType() : libutt::Coin::BudgetType(); 
+  Fr fr_exp_date;
+  fr_exp_date.from_words(exp_date);
+  Fr pid_hash;
+  pid_hash.from_words(cid.getPidHash());
+  coin_.reset(new libutt::Coin(d.getParams().ck_coin, fr_sn, fr_val, fr_type, fr_exp_date, pid_hash));
+  nullifier_.reset(new Nullifier(d, cid.getPRFSecretKey(), sn, Fr::random_element().to_words()));
+  type_ = t;
+}
+const Nullifier& Coin::getNullifier() const { return *nullifier_; }
+bool Coin::hasSig() const { return has_sig_; }
+void Coin::setSig(const std::vector<uint8_t>& sig) {
+  coin_->sig = libutt::deserialize<libutt::RandSig>(std::string(sig.begin(), sig.end()));
+  has_sig_ = true;
+}
+Coin::Type Coin::getType() const { return type_; }
+std::vector<uint8_t> Coin::getSig() const { 
+  auto str_sig = libutt::serialize<libutt::RandSig>(coin_->sig);
+  return std::vector<uint8_t>(str_sig.begin(), str_sig.end()); 
+  }
+void Coin::randomize() {
+  Fr u_delta = Fr::random_element();
+  coin_->sig.rerandomize(coin_->r, u_delta);
+}
+}  // namespace libutt::api
--- a/utt/libutt/src/coin.cpp
+++ b/utt/libutt/src/coin.cpp
@@ -5,12 +5,17 @@
 #include <utt/RandSig.h>
 
 namespace libutt::api {
-Coin::Coin(Details& d, const std::vector<uint64_t>& sn, const std::vector<uint64_t>& val, Type t, const std::vector<uint64_t>& exp_date, ClientIdentity& cid) {
+Coin::Coin(Details& d,
+           const std::vector<uint64_t>& sn,
+           const std::vector<uint64_t>& val,
+           Type t,
+           const std::vector<uint64_t>& exp_date,
+           ClientIdentity& cid) {
   Fr fr_sn;
   fr_sn.from_words(sn);
   Fr fr_val;
   fr_val.from_words(val);
-  Fr fr_type = t == Type::Normal ? libutt::Coin::NormalType() : libutt::Coin::BudgetType(); 
+  Fr fr_type = t == Type::Normal ? libutt::Coin::NormalType() : libutt::Coin::BudgetType();
   Fr fr_exp_date;
   fr_exp_date.from_words(exp_date);
   Fr pid_hash;
@@ -26,10 +31,10 @@ void Coin::setSig(const std::vector<uint8_t>& sig) {
   has_sig_ = true;
 }
 Coin::Type Coin::getType() const { return type_; }
-std::vector<uint8_t> Coin::getSig() const { 
+std::vector<uint8_t> Coin::getSig() const {
   auto str_sig = libutt::serialize<libutt::RandSig>(coin_->sig);
-  return std::vector<uint8_t>(str_sig.begin(), str_sig.end()); 
-  }
+  return std::vector<uint8_t>(str_sig.begin(), str_sig.end());
+}
 void Coin::randomize() {
   Fr u_delta = Fr::random_element();
   coin_->sig.rerandomize(coin_->r, u_delta);
--- a/utt/libutt/src/commitment.cpp
+++ b/utt/libutt/src/commitment.cpp
@@ -1,0 +1,90 @@
+#include "commitment.hpp"
+#include <utt/Comm.h>
+#include <utt/Params.h>
+#include <utt/RandSig.h>
+#include <utt/Serialization.h>
+#include <utt/PolyCrypto.h>
+
+#include <sstream>
+
+libutt::api::Commitment operator+(libutt::api::Commitment lhs, const libutt::api::Commitment& rhs) {
+  libutt::api::Commitment comm = lhs;
+  comm += rhs;
+  return comm;
+}
+
+namespace libutt::api {
+libutt::CommKey getCommitmentKey(libutt::Params& p, Commitment::Type t) {
+  switch (t) {
+    case Commitment::Type::REGISTRATION:
+      return p.ck_reg;
+    case Commitment::Type::VALUE:
+      return p.ck_val;
+    case Commitment::Type::COIN:
+      return p.ck_coin;
+      return libutt::CommKey();
+  }
+}
+std::vector<Commitment> Commitment::create(Details& d,
+                                           Type t,
+                                           const std::vector<std::vector<uint64_t>>& messages,
+                                           std::vector<std::vector<uint64_t>>& randomizations,
+                                           bool withG2) {
+  std::vector<Fr> fr_messages(messages.size());
+  for (size_t i = 0; i < messages.size(); i++) {
+    fr_messages[i].from_words(messages.at(i));
+  }
+  std::vector<Fr> fr_randomness(randomizations.size());
+  for (size_t i = 0; i < randomizations.size(); i++) {
+    fr_randomness[i].from_words(randomizations.at(i));
+  }
+
+  auto comms = libutt::Comm::create(getCommitmentKey(d.getParams(), t), fr_messages, fr_randomness, withG2);
+  std::vector<Commitment> ret(comms.size());
+  for (size_t i = 0; i < comms.size(); i++) {
+    *(ret[i].comm_) = comms[i];
+  }
+  return ret;
+}
+
+Commitment::Commitment(Details& d, Type t, const std::vector<std::vector<uint64_t>>& messages, bool withG2) {
+  std::vector<Fr> fr_messages(messages.size());
+  for (size_t i = 0; i < messages.size(); i++) {
+    fr_messages[i].from_words(messages.at(i));
+  }
+  comm_.reset(new libutt::Comm());
+  *comm_ = libutt::Comm::create(getCommitmentKey(d.getParams(), t), fr_messages, withG2);
+}
+Commitment::Commitment(const std::string& comm) {
+  comm_.reset(new libutt::Comm());
+  *comm_ = libutt::deserialize<libutt::Comm>(comm);
+}
+Commitment::Commitment() { comm_.reset(new libutt::Comm()); }
+Commitment& Commitment::operator=(const Commitment& comm) {
+  comm_.reset(new libutt::Comm());
+  *comm_ = *comm_ = *comm.comm_;
+  return *this;
+}
+
+Commitment::Commitment(const Commitment& comm) {
+  comm_.reset(new libutt::Comm());
+  *comm_ = *comm.comm_;
+}
+
+size_t Commitment::getCommitmentSn(const Commitment& comm) {
+  return std::hash<std::string>{}(libutt::serialize<libutt::Comm>(*(comm.comm_)));
+}
+std::string Commitment::getCommitmentHash(const Commitment& comm) {
+  return hashToHex("rcm|" + libutt::serialize<libutt::Comm>(*(comm.comm_)));
+}
+Commitment& Commitment::operator+=(const Commitment& comm) {
+  (*comm_) += *(comm.comm_);
+  return *this;
+}
+
+std::vector<uint64_t> Commitment::randomize(Details& d, Type t) {
+  Fr u_delta = Fr::random_element();
+  comm_->rerandomize(getCommitmentKey(d.getParams(), t), u_delta);
+  return u_delta.to_words();
+}
+}  // namespace libutt::api
--- a/utt/libutt/src/common.cpp
+++ b/utt/libutt/src/common.cpp
@@ -1,0 +1,36 @@
+#include "common.hpp"
+#include <utt/RandSig.h>
+#include <utt/Comm.h>
+#include <utt/Serialization.h>
+#include <utt/Params.h>
+namespace libutt::api {
+std::vector<uint8_t> Utils::aggregateSigShares(Details& d,
+                                               Commitment::Type t,
+                                               uint32_t n,
+                                               const std::map<uint32_t, std::vector<uint8_t>>& rsigs,
+                                               std::vector<std::vector<uint64_t>> randomness) {
+  std::vector<libutt::RandSigShare> shares;
+  std::vector<size_t> signers;
+  for (const auto& [sid, rsig] : rsigs) {
+    shares.push_back(libutt::deserialize<libutt::RandSigShare>(rsig));
+    signers.push_back((size_t)sid);
+  }
+  std::vector<Fr> fr_randomness(randomness.size());
+  for (size_t i = 0; i < randomness.size(); i++) fr_randomness[i].from_words(randomness[i]);
+  libutt::CommKey ck;
+  switch (t) {
+    case Commitment::Type::REGISTRATION:
+      ck = d.getParams().ck_reg;
+      break;
+    case Commitment::Type::VALUE:
+      ck = d.getParams().ck_val;
+      break;
+    case Commitment::Type::COIN:
+      ck = d.getParams().ck_coin;
+      break;
+  }
+  auto csig = libutt::RandSigShare::aggregate((size_t)(n), shares, signers, ck, fr_randomness);
+  auto str_csig = libutt::serialize<libutt::RandSig>(csig);
+  return std::vector<uint8_t>(str_csig.begin(), str_csig.end());
+}
+}  // namespace libutt::api
--- a/utt/tests/TestDistributedRegistration.cpp
+++ b/utt/tests/TestDistributedRegistration.cpp
@@ -1,0 +1,34 @@
+#include "details.hpp"
+#include "testUtils.hpp"
+
+#include <memory>
+#include <vector>
+#include <cstdlib>
+#include <ctime>
+
+using namespace libutt;
+using namespace libutt::api;
+
+int main(int argc, char* argv[]) {
+  (void)argc;
+  (void)argv;
+  size_t thresh = 3;
+  size_t n = 4;
+  size_t c = 10;
+  auto [dkg, rc] = testing::init(n, thresh);
+  auto& d = Details::instance();
+  auto registrators = testing::GenerateRegistrators(n, rc);
+  auto banks = testing::GenerateCommitters(n, dkg);
+  auto clients = testing::GenerateClients(c, dkg.getSK().toPK());
+
+  for (auto& c : clients) {
+    testing::registerClient(d, c, registrators, thresh);
+    for (auto& r : registrators) {
+      c.randomizeRCM();
+      auto rcm_data = c.getRcm();
+      assertTrue(r->validateRCM(rcm_data.first, rcm_data.second));
+    }
+  }
+
+  return 0;
+}
--- a/utt/tests/testUtils.hpp
+++ b/utt/tests/testUtils.hpp
@@ -23,7 +23,7 @@ using namespace libutt;
 using namespace libutt::api;
 namespace libutt::api::testing {
 std::vector<uint32_t> getSubGroup(uint32_t n, uint32_t size) {
-    std::srand((unsigned int)std::time(0));
+  std::srand((unsigned int)std::time(0));
   std::map<uint32_t, uint32_t> ret;
   for (uint32_t i = 0; i < n; i++) ret[i] = i;
   for (uint32_t i = 0; i < n - size; i++) {
--- a/utt/tests/testUtils.hpp
+++ b/utt/tests/testUtils.hpp
@@ -1,0 +1,96 @@
+#pragma once
+#include "details.hpp"
+#include "bankIdentity.hpp"
+#include "clientIdentity.hpp"
+#include "registratorIdentity.hpp"
+#include "common.hpp"
+#include "coin.hpp"
+#include <utt/RegAuth.h>
+#include <utt/RandSigDKG.h>
+#include <utt/Serialization.h>
+#include <utt/Params.h>
+#include <utt/IBE.h>
+#include <utt/Serialization.h>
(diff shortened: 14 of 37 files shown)


evdzhurov:
Did a quick pass over the changes and it looks good, but I'll need some time to get into the details and do the review.

evdzhurov (commented):
Looks good so far. Several things I'd like to clear up before I approve.
 - The registration flow is not yet split properly between the client and the replicas. I think I saw on Friday some usage of a commitment hash disguised as a "serial number" which I didn't understand. It looked like something totally different compared to the algorithm.
 - Unnecessary copying and memory allocations should be avoided as early as possible. It's ok to have some inefficient methods initially, but if you know that something can be written better please comment it so it's not forgotten down the road.
 - Interfaces should be const-correct (specify every non-modifiable value as const) - this helps to provide the user with an interface that properly communicates its intentions.
 - If singletons (like Details) will be exposed to users in the client API please plan to make them hard to misuse.

utt/libutt/src/common.cpp
```diff
@@ -0,0 +1,36 @@
...
+    case Commitment::Type::REGISTRATION:
+      ck = d.getParams().ck_reg;
+      break;
+    case Commitment::Type::VALUE:
+      ck = d.getParams().ck_val;
+      break;
+    case Commitment::Type::COIN:
+      ck = d.getParams().ck_coin;
+      break;
+  }
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/commitment.hpp
```diff
@@ -0,0 +1,38 @@
...
+  Commitment& operator=(const Commitment&);
+  Commitment& operator+=(const Commitment&);
+  std::vector<uint64_t> randomize(Details& d, Type t);
+  static size_t getCommitmentSn(const Commitment& comm);
+  static std::string getCommitmentHash(const Commitment& comm);
+
+ private:
+  friend class RegistratorIdentity;
+  friend class ClientIdentity;
+  std::shared_ptr<libutt::Comm> comm_;
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/commitment.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <sstream>
+
+libutt::api::Commitment operator+(libutt::api::Commitment lhs, const libutt::api::Commitment& rhs) {
+  libutt::api::Commitment comm = lhs;
+  comm += rhs;
+  return comm;
+}
+
+namespace libutt::api {
+libutt::CommKey getCommitmentKey(libutt::Params& p, Commitment::Type t) {
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/commitment.cpp
```diff
@@ -0,0 +1,90 @@
...
+  switch (t) {
+    case Commitment::Type::REGISTRATION:
+      return p.ck_reg;
+    case Commitment::Type::VALUE:
+      return p.ck_val;
+    case Commitment::Type::COIN:
+      return p.ck_coin;
+      return libutt::CommKey();
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/commitment.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <utt/Comm.h>
+#include <utt/Params.h>
+#include <utt/RandSig.h>
+#include <utt/Serialization.h>
+#include <utt/PolyCrypto.h>
+
+#include <sstream>
+
+libutt::api::Commitment operator+(libutt::api::Commitment lhs, const libutt::api::Commitment& rhs) {
+  libutt::api::Commitment comm = lhs;
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/commitment.cpp
```diff
@@ -0,0 +1,90 @@
...
+  *comm_ = libutt::Comm::create(getCommitmentKey(d.getParams(), t), fr_messages, withG2);
+}
+Commitment::Commitment(const std::string& comm) {
+  comm_.reset(new libutt::Comm());
+  *comm_ = libutt::deserialize<libutt::Comm>(comm);
+}
+Commitment::Commitment() { comm_.reset(new libutt::Comm()); }
+Commitment& Commitment::operator=(const Commitment& comm) {
+  comm_.reset(new libutt::Comm());
+  *comm_ = *comm_ = *comm.comm_;
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/commitment.cpp
```diff
@@ -0,0 +1,90 @@
...
+  comm_.reset(new libutt::Comm());
+  *comm_ = libutt::Comm::create(getCommitmentKey(d.getParams(), t), fr_messages, withG2);
+}
+Commitment::Commitment(const std::string& comm) {
+  comm_.reset(new libutt::Comm());
+  *comm_ = libutt::deserialize<libutt::Comm>(comm);
+}
+Commitment::Commitment() { comm_.reset(new libutt::Comm()); }
+Commitment& Commitment::operator=(const Commitment& comm) {
+  comm_.reset(new libutt::Comm());
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/clientIdentity.cpp
```diff
@@ -0,0 +1,56 @@
...
+  bpk_.reset(new libutt::RandSigPK());
+  *bpk_ = libutt::deserialize<libutt::RandSigPK>(bpk);
+}
+Commitment ClientIdentity::generatePartialRCM(Details& d) {
+  std::vector<std::vector<uint64_t>> m = {getPidHash(), ask_->s.to_words(), Fr::zero().to_words()};
+  auto comm = Commitment(d, Commitment::Type::REGISTRATION, m, true);
+  return comm;
+}
+
+std::string ClientIdentity::getPid() const { return ask_->pid; }
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/clientIdentity.hpp
```diff
@@ -0,0 +1,31 @@
...
+  Commitment generatePartialRCM(Details& d);
+  std::string getPid() const;
+  std::vector<uint64_t> getPRFSecretKey() const;
+  std::vector<uint64_t> getPidHash() const;
+  void setIBEDetails(const std::vector<uint8_t>& sk, const std::vector<uint8_t>& mpk);
+  void setRCM(const Commitment& comm, const std::vector<uint8_t>& sig);
+  void randomizeRCM();
+  std::pair<Commitment, std::vector<uint8_t>> getRcm() const;
+  template <typename T>
+  bool validate(const T&);
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/common.hpp
```diff
@@ -0,0 +1,25 @@
...
+  std::vector<uint8_t> dsk_priv_;
+};
+
+class Utils {
+ public:
+  static std::vector<uint8_t> aggregateSigShares(Details& d,
+                                                 Commitment::Type t,
+                                                 uint32_t n,
+                                                 const std::map<uint32_t, std::vector<uint8_t>>& rsigs,
+                                                 std::vector<std::vector<uint64_t>> randomness);
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/coin.hpp
```diff
@@ -0,0 +1,36 @@
...
+class ClientIdentity;
+class Coin {
+ public:
+  enum Type { Normal = 0x0, Budget };
+  Coin(Details& d,
+       const std::vector<uint64_t>& sn,
+       const std::vector<uint64_t>& val,
+       Type p,
+       const std::vector<uint64_t>& exp_date,
+       ClientIdentity& cid);
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/coin.cpp
```diff
@@ -0,0 +1,42 @@
...
+  coin_.reset(new libutt::Coin(d.getParams().ck_coin, fr_sn, fr_val, fr_type, fr_exp_date, pid_hash));
+  nullifier_.reset(new Nullifier(d, cid.getPRFSecretKey(), sn, Fr::random_element().to_words()));
+  type_ = t;
+}
+const Nullifier& Coin::getNullifier() const { return *nullifier_; }
+bool Coin::hasSig() const { return has_sig_; }
+void Coin::setSig(const std::vector<uint8_t>& sig) {
+  coin_->sig = libutt::deserialize<libutt::RandSig>(std::string(sig.begin(), sig.end()));
+  has_sig_ = true;
+}
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/coin.cpp
```diff
@@ -0,0 +1,42 @@
...
+Coin::Type Coin::getType() const { return type_; }
+std::vector<uint8_t> Coin::getSig() const {
+  auto str_sig = libutt::serialize<libutt::RandSig>(coin_->sig);
+  return std::vector<uint8_t>(str_sig.begin(), str_sig.end());
+}
+void Coin::randomize() {
+  Fr u_delta = Fr::random_element();
+  coin_->sig.rerandomize(coin_->r, u_delta);
+}
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/mint.hpp
```diff
@@ -0,0 +1,29 @@
...
+class BankIdentity;
+}
+namespace libutt::api::operations {
+
+class Mint {
+ public:
+  Mint(const std::string& uniqueHash, size_t value, const std::string& recipPID);
+  bool validate(const std::string& uniqueHash, size_t value, const std::string& recipPID) const;
+  Coin claimCoin(Details& d,
+                 ClientIdentity& cid,
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/registratorIdentity.hpp
```diff
@@ -0,0 +1,29 @@
...
+}  // namespace libutt
+namespace libutt::api {
+class RegistratorIdentity {
+ public:
+  RegistratorIdentity(const std::string& id, const std::string& rsk, const std::string& rbk);
+  RegistrationDetails registerClient(Details& d,
+                                     const std::string& pid,
+                                     const std::vector<uint64_t>& pid_hash,
+                                     const Commitment& partiail_comm) const;
+  bool validateRCM(const Commitment& comm, const std::vector<uint8_t>& sig);
```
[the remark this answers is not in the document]
yontyon: fixed

utt/tests/testUtils.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <vector>
+#include <memory>
+#include <cstdlib>
+#include <iostream>
+#include <ctime>
+using namespace libutt;
+using namespace libutt::api;
+namespace libutt::api::testing {
+std::vector<uint32_t> getSubGroup(uint32_t n, uint32_t size) {
+  std::srand((unsigned int)std::time(0));
```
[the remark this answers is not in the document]
yontyon: This is only for testing, so it is not that important IMO

utt/libutt/src/transaction.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <utt/RandSig.h>
+#include <utt/RegAuth.h>
+#include <utt/Coin.h>
+#include <utt/Address.h>
+namespace libutt::api {
+Transaction::Transaction(Details& d,
+                         const ClientIdentity& cid,
+                         const std::vector<Coin>& coins,
+                         const std::optional<Coin>& bc,
+                         const std::vector<std::tuple<std::string, uint64_t>>& receipents) {
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/clientIdentity.cpp
```diff
@@ -0,0 +1,55 @@
...
+  ask_.reset(new libutt::AddrSK());
+  ask_->pid = pid;
+  ask_->s = Fr::random_element();
+  ask_->pid_hash = AddrSK::pidHash(pid);
+  bpk_.reset(new libutt::RandSigPK());
+  *bpk_ = libutt::deserialize<libutt::RandSigPK>(bpk);
+  rpk_.reset(new libutt::RegAuthPK());
+  *rpk_ = libutt::deserialize<libutt::RegAuthPK>(rvk);
+}
+Commitment ClientIdentity::generatePartialRCM(Details& d) {
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/coin.cpp
```diff
@@ -0,0 +1,57 @@
...
+  type_ = t;
+}
+Coin::Coin(const Coin& c) {
+  coin_.reset(new libutt::Coin());
+  *(coin_) = *(c.coin_);
+  has_sig_ = c.has_sig_;
+  type_ = c.type_;
+}
+Coin& Coin::operator=(const Coin& c) {
+  coin_.reset(new libutt::Coin());
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/commitment.cpp
```diff
@@ -0,0 +1,90 @@
...
+  }
+  std::vector<Fr> fr_randomness(randomizations.size());
+  for (size_t i = 0; i < randomizations.size(); i++) {
+    fr_randomness[i].from_words(randomizations.at(i));
+  }
+
+  auto comms = libutt::Comm::create(getCommitmentKey(d.getParams(), t), fr_messages, fr_randomness, withG2);
+  std::vector<Commitment> ret(comms.size());
+  for (size_t i = 0; i < comms.size(); i++) {
+    *(ret[i].comm_) = comms[i];
```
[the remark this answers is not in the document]
yontyon: removed

yontyon:
> Looks good so far. Several things I'd like to clear up before I approve.
> 
> * The registration flow is not yet split properly between the client and the replicas. I think I saw on Friday some usage of a commitment hash disguised as a "serial number" which I didn't understand. It looked like something totally different compared to the algorithm.
> * Unnecessary copying and memory allocations should be avoided as early as possible. It's ok to have some inefficient methods initially, but if you know that something can be written better please comment it so it's not forgotten down the road.
> * Interfaces should be const-correct (specify every non-modifiable value as const) - this helps to provide the user with an interface that properly communicates its intentions.
> * If singletons (like Details) will be exposed to users in the client API please plan to make them hard to misuse.

@evdzhurov 
1. I've completed the registration process
2. I tried to address all of your comments, please let me know if I forgot anything.

utt/tests/testUtils.hpp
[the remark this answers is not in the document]
evdzhurov: I would agree with you if this is a temp function, but if it's going to be used in tests going forward I would say it's important. To me it seems that the seeding here is done in case someone forgets to seed before using getSubGroup. Is there any advantage to seed inside the function besides that? IMO you can seed once in each test and name the function "getReandomSubGroup" to hint that it does use randomness. And btw, there are other ways to implement "sampling without replacement", including some new stuff from C++17 (https://stackoverflow.com/questions/33802205/how-to-sample-without-replacement-using-c-uniform-int-distribution) - see the second post.

utt/include/client.hpp
```diff
@@ -0,0 +1,52 @@
...
+         const std::string& rpk,
+         const std::string& csk,
+         const std::string& mpk);
+  Commitment generateInputRCM();
+  void setPRFKey(const types::CurvePoint& s2);
+  const std::string& getPid() const;
+  types::CurvePoint getPRFSecretKey() const;
+  types::CurvePoint getPidHash() const;
+
+  // This method has to be called only after setPRFKey with s2
```
evdzhurov, line 31: Instead of relying on this comment and hope that the caller won't do mistakes can we combine setPRFKey and setRCMSig to provide a single method that does the right thing?

utt/tests/TestDistributedRegistration.cpp
```diff
@@ -0,0 +1,33 @@
...
+  auto banks = testing::GenerateCommitters(n, dkg, rc.toPK());
+  auto clients = testing::GenerateClients(c, dkg.getSK().toPK(), rc.toPK(), rc);
+
+  for (auto& c : clients) {
+    testing::registerClient(d, c, registrators, thresh);
+    for (auto& r : registrators) {
+      auto rcm_data = c.getRcm();
+      assertTrue(r->validateRCM(rcm_data.first, rcm_data.second));
+    }
+  }
```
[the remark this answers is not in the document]
yontyon: It is verified as part of the split proof, see https://github.com/vmware/concord-bft/blob/utt/utt/libutt/src/SplitProof.cpp#L28.
And yes, the bank is responsible for this verification and he needs to know the registration verification key.

utt/libutt/src/transaction.cpp
```diff
@@ -0,0 +1,47 @@
...
+  for (auto& c : coins) input_coins.push_back(*(c.coin_));
+  std::optional<libutt::Coin> budget_coin = std::nullopt;
+  if (bc.has_value()) budget_coin.emplace(*(bc->coin_));
+  std::vector<std::tuple<std::string, Fr>> fr_receips(receipents.size());
+  for (size_t i = 0; i < receipents.size(); i++) {
+    auto& [r_str, r_id] = receipents[i];
+    auto& [id, fr] = fr_receips[i];
+    id = r_str;
+    fr.set_ulong(r_id);
+  }
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/clientIdentity.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <utt/RandSig.h>
+#include <utt/RegAuth.h>
+#include <utt/Params.h>
+#include <utt/Serialization.h>
+#include <utt/Coin.h>
+#include <utt/BurnOp.h>
+#include <vector>
+#include <sstream>
+namespace libutt::api {
+ClientIdentity::ClientIdentity(const std::string& pid, const std::string& bpk, const std::string& rvk) {
```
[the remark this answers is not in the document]
yontyon: fixed (an exception will be thrown)

utt/libutt/src/registrator.cpp
```diff
@@ -0,0 +1,40 @@
...
+  id_ = id;
+  rsk_.reset(new libutt::RegAuthShareSK());
+  *rsk_ = libutt::deserialize<libutt::RegAuthShareSK>(rsk);
+  rpk_.reset(new libutt::RegAuthPK());
+  *rpk_ = libutt::deserialize<libutt::RegAuthPK>(rbk);
+  tmp_.reset(new libutt::RegAuthSK());
+  *tmp_ = tmp;
+}
+
+std::pair<types::CurvePoint, types::Signature> Registrator::ComputeRCM(const types::CurvePoint& pid_hash,
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/client.hpp
yontyon: fixed

utt/tests/TestDistributedRegistration.cpp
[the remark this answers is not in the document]
evdzhurov: Ok, so rcm is "somehow" incleaded in the splitproof. But in TestDistributedRegistration do we test that the verification of a re-randomized rcm holds on its own?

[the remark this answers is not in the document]
yontyon: ok, I've added this to the test
Thanks

utt/libutt/test/TestIBE.cpp
```diff
@@ -65,8 +66,10 @@ int main(int argc, char *argv[]) {
   auto ctxtCopy = ctxt;
   testAssertEqual(ctxt, ctxtCopy);
 
-  testAssertNotEqual(ctxt,
-                     mpk.encrypt(pid, frsToBytes({Fr::random_element(), Fr::random_element(), Fr::random_element()})));
+  auto frs_data = frsToBytes({Fr::random_element(), Fr::random_element(), Fr::random_element()});
+  AutoBuf<unsigned char> auto_buf_data(frs_data.size());
+  memcpy(auto_buf_data.getBuf(), encbuf.data(), encbuf.size());
```
evdzhurov, line 71: This line seems wrong and is failing the IBE test.

evdzhurov:
It seems there are 6 libutt tests failing after the changes - please address the issues so I can continue the review.

utt/include/globalParams.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include <string>
+#include <memory>
+namespace libutt {
+class Params;
+}
+namespace libutt::api {
+class GlobalParams {
+ public:
+  static GlobalParams& instance() {
```
yontyon, line 10: remove singleton

utt/include/commitment.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace libutt::api {
+class Registrator;
+class Client;
+namespace operations {
+class Burn;
+class Transaction;
+}  // namespace operations
+class Commitment {
+ public:
+  enum Type { REGISTRATION = 0, VALUE, COIN };
```
yontyon, line 23: value type can be removed

utt/include/commitment.hpp
```diff
@@ -0,0 +1,39 @@
...
+  static const libutt::CommKey& getCommitmentKey(const GlobalParams& d, Type t);
+  Commitment(const GlobalParams& d, Type t, const std::vector<types::CurvePoint>& messages, bool withG2);
+  Commitment(const Commitment& comm);
+  Commitment();
+  Commitment& operator=(const Commitment&);
+  Commitment& operator+=(const Commitment&);
+  types::CurvePoint rerandomize(const GlobalParams& d, Type t);
```
yontyon, line 30: explain  or allow input parameter for the random number

utt/include/registrator.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <vector>
+namespace libutt {
+class RegAuthShareSK;
+class RegAuthPK;
+class RegAuthSK;
+}  // namespace libutt
+namespace libutt::api {
+class Registrator {
+ public:
+  Registrator(const std::string& id, const std::string& rsk, const std::string& rbk);
```
yontyon, line 15: more elaborated names

yontyon: add comment and explanations

utt/libutt/test/TestIBE.cpp
yontyon: fixed

utt/include/burn.hpp
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @brief This object represent a Burn operation.
+ *
+ */
+class Burn {
+ public:
+  /**
+   * @brief Construct a new Burn object
+   *
+   * @param p The shared global UTT parametrs
```
eyalrund, line 33: typo: parameter

utt/include/client.hpp
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * @brief The Client object represents the client's public and private date and a group of relevant operations.
+   *
+   */
+ public:
+  /**
+   * @brief Construct a new Client object using an IBE based PKI
+   *
+   * @param id The client id
+   * @param bnak_public_key The serialized bank public key
```
eyalrund, line 44: typo: bank

utt/include/client.hpp
```diff
@@ -0,0 +1,166 @@
...
+}  // namespace libutt
+namespace libutt::api {
+class Coin;
+namespace operations {
+class Burn;
+class Transaction;
+}  // namespace operations
+class Client {
+  /**
+   * @brief The Client object represents the client's public and private date and a group of relevant operations.
```
eyalrund, line 36: public and private date?

utt/include/client.hpp
```diff
@@ -0,0 +1,166 @@
...
+   */
+  std::pair<Commitment, types::Signature> getRcm() const;
+
+  /**
+   * @brief Claims the coins owned by this client
+   *
+   * @tparam T one of <operations::Mint, operations::Budget, operations::Transaction>
+   * @param p The shared global UTT parametrs
+   * @param blindedSigs The combined signature. Notice, that even if someone else has collected the signature, only the
+   * client knows the coin's randomness and hance only the client can unblind it
```
eyalrund, line 137: typo: hence

utt/include/coin.hpp
```diff
@@ -0,0 +1,181 @@
...
+class Burn;
+class Mint;
+class Transaction;
+class Budget;
+}  // namespace operations
+
+class Coin {
+  /**
+   * @brief Represent a UTT coin. A coin can be either a normal value coin or a budget coin. A budget coin is a coin
+   * that limits the amount of annonymus transferred tokens.
```
eyalrund, line 37: typo: anonymous

utt/include/coin.hpp
```diff
@@ -0,0 +1,181 @@
...
+   *
+   */
+ public:
+  enum Type { Normal = 0x0, Budget };
+  /**
+   * @brief Construct a new Coin object. This constructor also construct the coin's nulliffier. Hence it can be invoked
```
eyalrund, line 43: also constructs :-)

utt/include/coin.hpp
```diff
@@ -0,0 +1,181 @@
...
+   * only by the client.
+   *
+   * @param p The shared global UTT parametrs
```
eyalrund, line 46: typo - parameters

utt/include/coin.hpp
```diff
@@ -0,0 +1,181 @@
...
+}  // namespace operations
+
+class Coin {
+  /**
+   * @brief Represent a UTT coin. A coin can be either a normal value coin or a budget coin. A budget coin is a coin
+   * that limits the amount of annonymus transferred tokens.
+   *
+   */
+ public:
+  enum Type { Normal = 0x0, Budget };
```
eyalrund, line 41: Is 'Normal' the best term for such a coin?

utt/include/common.hpp
```diff
@@ -0,0 +1,73 @@
...
+   * @brief Aggreate the shared signatures for a given signature
+   *
+   * @param n The total number of coinsSigner replicas in the system
+   * @param rsigs A map of <signer ID, Signature> of a thershold collected signatures
+   * @return types::Signature The combined signature
+   */
+  static types::Signature aggregateSigShares(uint32_t n, const std::map<uint32_t, types::Signature>& rsigs);
+
+  /**
+   * @brief ublineds a signature randomness
```
eyalrund, line 43: typo: unblind

utt/include/common.hpp
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * @param exp_date An input string represents a date in the format of %Y-%m-%d %H:%M:%S. For example, 2022-08-17
+   * 11:28:32
+   * @return uint64_t The encoded date as uint of in the following format yyyymmddHHMMSS for example 2022-08-17 11:28:32
+   * -> 20220817112832
+   */
+  static uint64_t getExpirationDateAsUint(const std::string& exp_date);
+
+  /**
+   * @brief The reverese of getExpirationDateAsUint method
```
eyalrund, line 66: i.e converts a uint date to a string format

utt/include/globalParams.hpp
```diff
@@ -0,0 +1,50 @@
...
+#pragma once
+#include <string>
+#include <memory>
+namespace libutt {
+class Params;
+}
+namespace libutt::api {
+class GlobalParams {
+  /**
+   * @brief Represents a shared known global UTT params. These parameters includes commitment keys, nullifier parametrs
```
eyalrund, line 23: typo: include , parameters

utt/include/mint.hpp
```diff
@@ -0,0 +1,52 @@
...
+}
+namespace libutt::api {
+class CoinsSigner;
+class Client;
+}  // namespace libutt::api
+namespace libutt::api::operations {
+
+class Mint {
+  /**
+   * @brief The mint operation takes public tokens and convert them to a single UTT coin
```
eyalrund, line 31: typo: converts .. into a

utt/include/mint.hpp
```diff
@@ -0,0 +1,52 @@
...
+   *
+   */
+ public:
+  /**
+   * @brief Construct a new Mint object
+   *
+   * @param uniqueHash The mint transaction hash
+   * @param value The required value
+   * @param recipPID The recipent id
```
eyalrund, line 40: typo: recipient

utt/include/transaction.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <vector>
+namespace libutt {
+class Tx;
+class IEncryptor;
+}  // namespace libutt
+
+namespace libutt::api::operations {
+class Transaction {
+  /**
+   * @brief A transaction represents a transfer of UTT coin(s) from a sender accound c1 to other accoun(s) [c2, ...,
```
eyalrund, line 31: typo account(s)

evdzhurov:
@yontyon Do you have plans to change the organization of the source code in the utt folder? I'm starting to notice some things that may cause confusion. For example in libutt/src we now have both a "Client.cpp" and a "client.cpp" files. Can we do something better with the source files in the utt folder? For example - we can move the interface work to an "api" subproject or something similar.

utt/libutt/include/utt/DataUtils.hpp
```diff
@@ -0,0 +1,65 @@
...
+  virtual ~IEncryptor() {}
+};
+
+class IDecryptor {
+ public:
+  virtual std::vector<uint8_t> decrypt(const std::vector<uint8_t>&) const = 0;
+  virtual ~IDecryptor() {}
+};
+
+class IBEEncryptor : public IEncryptor {
```
WildFireFlum, line 28: What does IBE stand for?

utt/libutt/src/RandSig.cpp
```diff
@@ -189,7 +189,29 @@ bool RandSigShare::verify(const std::vector<Comm>& c, const RandSigSharePK& pk)
...
+  RandSig sig;
 
+  // WARNING: Assuming all shares verify and have the same base 'h' is wrong
+  // since individual calls on RandSigShare::verify(sigShares[i], ...) will not guarantee this!
+  // Instead, the caller must ensure this manually, which we do here.
+  sig.s1 = sigShares[0].s1;
+  std::vector<G1> s2;
+  for (auto& ss : sigShares) {
+    if (ss.s1 != sig.s1) {
+      throw std::runtime_error("Expected signature shares with the same 'h'");
```
WildFireFlum, line 204: I have no idea what this h is, but the field name is 's1' and not 'h' - should probably rename it.

utt/tests/TestBurn.cpp
```diff
@@ -0,0 +1,54 @@
...
+    for (size_t i = 0; i < banks.size(); i++) {
+      rsigs.push_back(banks[i]->sign(mint).front());
+    }
+    auto sbs = testing::getSubGroup((uint32_t)n, (uint32_t)thresh);
+    std::map<uint32_t, std::vector<uint8_t>> sigs;
+    for (auto i : sbs) {
+      sigs[i] = rsigs[i];
+    }
+    auto blinded_sig = Utils::aggregateSigShares((uint32_t)n, {sigs});
+    auto coin = c.claimCoins(mint, d, {blinded_sig}).front();
```
evdzhurov, line 46: Does this coin have the same properties as the mint operation - value, expiration, etc.?

utt/include/globalParams.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * @brief Get the libutt::Params object
+   *
+   * @return const libutt::Params&
+   */
+  const libutt::Params& getParams() const;
+  libutt::Params& getParams();
+  GlobalParams(const GlobalParams& other);
+  GlobalParams& operator=(const GlobalParams& other);
```
evdzhurov, line 45: Couple of things worry me about the code design of this "global params" implementation:
1) You use the non-const version of "getParams()" as a form of a setter.
2) The "init" method uses a static class flag to do the library initialization once - while this is fine, init is also responsible for allocating the memory for params. Later params is assumed that it's always allocated.
3) There's a confusing mix of static and non-static data usage.

What Global params should be to be a "clean" implementation IMO:
1) Global params should be a singleton class, where the instance object is created and initialized only once before first use - this is probably necessary because of how libff/NTL are globally initialized.
2) We might need to customize the creation of global params (by passing extra parameters for libff/NTL or seeds), so we might need an explicit static "create" method called by the user of the library. However besides this, we should still rely on a "instance()" method to get the singleton object.
3) Global params object should have a private constructor so the only way to be created is to use the create/instance public methods.

utt/include/coin.hpp
yontyon: This is the term that was used until now, so I would prefer not to change it

utt/libutt/include/utt/DataUtils.hpp
```diff
@@ -0,0 +1,65 @@
...
+ public:
+  RSAEncryptor(const std::map<std::string, std::string>& rsa_public_keys_map);
+  ~RSAEncryptor() override;
+  std::vector<uint8_t> encrypt(const std::string& id, const std::vector<uint8_t>& msg) const override;
+
+ private:
+  mutable std::map<std::string, EVP_PKEY*> encryptors_;
+};
+
+class RSADecryptor : public IDecryptor {
```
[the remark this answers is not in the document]
yontyon: Basically, the utt is independent in concordbft. I would even prefer to have it in a different repository (if it wouldn't be such a bureaucratic process). This is why I prefer not to use anything from concordbft

utt/libutt/include/utt/Coin.h
```diff
@@ -75,6 +75,17 @@ class Coin {
  public:
   // WARNING: Used only for deserialization
```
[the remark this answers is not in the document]
yontyon: This is a legacy approach of the libutt, we can change it of course, but not in the scope of this PR (which is big enough already)

utt/libutt/include/utt/Coin.h
```diff
@@ -75,6 +75,17 @@ class Coin {
...
   Coin() {}
+  // Used to create the coin's partial commitment (without nullfier)
+  Coin(const CommKey& ck, const Fr& sn, const Fr& val, const Fr& type, const Fr& exp_date, const Fr& pidHash);
```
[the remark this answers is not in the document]
yontyon: I'm not sure I understood the comment, can you elaborate?

utt/libutt/include/utt/DataUtils.hpp
yontyon: IBE = identity based encryption. Please see the UTT paper for reference

utt/libutt/src/coinsSigner.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+const std::string& CoinsSigner::getId() const { return bid_; }
+template <>
+std::vector<types::Signature> CoinsSigner::sign<operations::Mint>(operations::Mint& mint) const {
+  auto res = mint.op_->shareSignCoin(*bsk_);
+  auto res_str = libutt::serialize<libutt::RandSigShare>(res);
+  return {types::Signature(res_str.begin(), res_str.end())};
+}
+template <>
+std::vector<types::Signature> CoinsSigner::sign<operations::Transaction>(operations::Transaction& tx) const {
```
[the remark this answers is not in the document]
yontyon: This is because the bank is not necessarily a distributed bank.

utt/tests/TestBudget.cpp
```diff
@@ -0,0 +1,71 @@
...
+    for (size_t i = 0; i < banks.size(); i++) {
+      rsigs.push_back(banks[i]->sign(budget).front());
+    }
+    auto sbs = testing::getSubGroup((uint32_t)n, (uint32_t)thresh);
+    std::map<uint32_t, std::vector<uint8_t>> sigs;
+    for (auto i : sbs) {
+      sigs[i] = rsigs[i];
+    }
+    auto blinded_sig = Utils::aggregateSigShares((uint32_t)n, {sigs});
+    auto coin = c.claimCoins(budget, d, {blinded_sig}).front();
```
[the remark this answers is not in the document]
yontyon: yes, it claims the output coin of the budget operation (the budget operation is basically a mint for budget coin)

utt/tests/TestBurn.cpp
yontyon: yes

utt/libutt/include/utt/Coin.h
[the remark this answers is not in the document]
WildFireFlum: The model mentions two "types" of coins: The private ones - "cash", and the public ones. I was trying to understand to which of the two types this class relates - both, or just a single one of them.

utt/libutt/include/utt/DataUtils.hpp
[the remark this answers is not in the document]
WildFireFlum: The utils part of concord-bft has no other dependencies, and could have been separated from the main repository as well.
Your implementation has similar bugs to ones recently fixed in concord-bft's utils (for example, no use of RAII), we should not maintain both versions.

utt/libutt/include/utt/DataUtils.hpp
WildFireFlum: If this is important, it should probably be explained in a comment..
Unless it is a requirement to read the paper before making any modification to this file, which should also be specified somehow in the source so that people who didn't will stay away.

utt/include/burn.hpp
yontyon: fixed

utt/include/client.hpp
yontyon: fixed

utt/include/client.hpp
yontyon: fixed

utt/include/coin.hpp
yontyon: fixed

utt/include/coin.hpp
yontyon: fixed

utt/include/common.hpp
yontyon: fixed

utt/include/common.hpp
yontyon: fixed

utt/include/globalParams.hpp
yontyon: fixed

utt/include/mint.hpp
yontyon: fixed

utt/include/transaction.hpp
yontyon: fixed

yontyon:
@evdzhurov , I have moved all the api source code to utt/libutt/src/api

utt/include/client.hpp
yontyon: fixed

utt/libutt/src/DataUtils.cpp
```diff
@@ -0,0 +1,94 @@
...
+  auto [succ, dtxt] = esk_.decrypt(ctxt);
+  if (!succ) return {};
+  std::vector<uint8_t> ret(dtxt.size());
+  memcpy(ret.data(), dtxt.getBuf(), dtxt.size());
+  return ret;
+}
+
+RSAEncryptor::RSAEncryptor(const std::map<std::string, std::string>& rsa_public_keys_map) {
+  for (const auto& [id, pub_key_str] : rsa_public_keys_map) {
+    BIO* keybio;
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/RandSig.cpp
yontyon: fixed

utt/include/budget.hpp
```diff
@@ -1,23 +1,64 @@
...
 #pragma once
 #include "coin.hpp"
 #include "client.hpp"
 #include "globalParams.hpp"
 #include "types.hpp"
 #include <string>
 namespace libutt::api::operations {
+/**
+ * @brief
+ * The Budget is used for creating an initial budger coin. This can ber viewed as a mint operation for budget coin.
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/common.cpp
```diff
@@ -3,24 +3,67 @@
...
+  uint16_t hours = (uint16_t)(exp_date / (uint64_t)std::pow(10, 4));
+  auto h_str = month >= 10 ? std::to_string(hours) : "0" + std::to_string(hours);
+  exp_date %= (uint64_t)std::pow(10, 4);
+  uint16_t minutes = (uint16_t)(exp_date / (uint64_t)std::pow(10, 2));
+  auto min_str = month >= 10 ? std::to_string(minutes) : "0" + std::to_string(minutes);
+  exp_date %= (uint64_t)std::pow(10, 2);
+  uint16_t seconds = (uint16_t)(exp_date);
+  auto sec_str = month >= 10 ? std::to_string(seconds) : "0" + std::to_string(seconds);
+  return std::to_string(year) + "-" + m_str + "-" + d_str + " " + h_str + ":" + min_str + ":" + sec_str;
+}
```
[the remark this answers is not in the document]
yontyon: fixed. 
My only concern is, do you know if the time_since_epoch is dependent on the local clock? because, if it does, we cannot use it.

utt/libutt/include/utt/Tx.h
```diff
@@ -117,7 +131,7 @@ class Tx {
 
     return sigShare.verify(getCommVector(txoIdx, H), bpkShare);
   }
-
+  std::unordered_map<size_t, TxOut> getMineTransactions(const IDecryptor& decriptor) const;
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/include/utt/Tx.h
```diff
@@ -135,7 +149,8 @@ class Tx {
                                    size_t n,
                                    const std::vector<RandSigShare>& sigShares,
                                    const std::vector<size_t>& signerIds,
-                                   const RandSigPK& bpk) const;
+                                   const RandSigPK& bpk,
+                                   const IDecryptor& decriptor) const;
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/DataUtils.cpp
```diff
@@ -0,0 +1,94 @@
...
+    RSA* pb_key = NULL;
+    if (!PEM_read_bio_RSA_PUBKEY(keybio, &pb_key, nullptr, nullptr)) {
+      throw std::runtime_error("Failed to create rsa public key");
+    }
+    encryptors_[id] = EVP_PKEY_new();
+    EVP_PKEY_assign_RSA(encryptors_[id], pb_key);
+    BIO_free(keybio);
+    // RSA_free(pb_key);
+  }
+}
```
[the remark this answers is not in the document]
yontyon: fixed (using RAII)

utt/libutt/src/DataUtils.cpp
```diff
@@ -0,0 +1,94 @@
...
+RSAEncryptor::~RSAEncryptor() {
+  for (auto [_, rsa_pub_key] : encryptors_) {
+    if (rsa_pub_key) EVP_PKEY_free(rsa_pub_key);
+  }
+}
+std::vector<uint8_t> RSAEncryptor::encrypt(const std::string& id, const std::vector<uint8_t>& msg) const {
+  if (encryptors_.find(id) == encryptors_.end()) throw std::runtime_error("Unknown id for encryption id: " + (id));
+  EVP_PKEY_CTX* ctx;
+  ctx = EVP_PKEY_CTX_new(encryptors_.at(id), NULL);
```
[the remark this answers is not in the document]
yontyon: fixed (using RAII)

utt/libutt/src/DataUtils.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+RSADecryptor::RSADecryptor(const std::string& rsa_private_key) {
+  BIO* keybio;
+  keybio = BIO_new_mem_buf(rsa_private_key.c_str(), -1);
+  if (!keybio) {
+    throw std::runtime_error("Failed to create key BIO");
+  }
+  RSA* priv_key = NULL;
+  if (!PEM_read_bio_RSAPrivateKey(keybio, &priv_key, nullptr, nullptr)) {
+    throw std::runtime_error("Failed to create rsa public key");
```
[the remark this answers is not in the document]
yontyon: fixed (using RAII)

utt/libutt/src/DataUtils.cpp
```diff
@@ -0,0 +1,94 @@
...
+  }
+  pkey_ = EVP_PKEY_new();
+  EVP_PKEY_assign_RSA(pkey_, priv_key);
+  BIO_free(keybio);
+}
+RSADecryptor::~RSADecryptor() { EVP_PKEY_free(pkey_); }
+std::vector<uint8_t> RSADecryptor::decrypt(const std::vector<uint8_t>& ctxt) const {
+  EVP_PKEY_CTX* ctx;
+  ctx = EVP_PKEY_CTX_new(pkey_, NULL);
```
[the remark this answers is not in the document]
yontyon: fixed (using RAII)

utt/libutt/src/MintOp.cpp
```diff
@@ -8,14 +8,16 @@
...
-
-namespace libutt {
+#include <xassert/XAssert.h>
+#include <xutils/Log.h>
 
 std::ostream& operator<<(std::ostream& out, const libutt::MintOp& op) {
   out << op.sn << endl;
   out << op.pidHash << endl;
+  out << op.clientId.size() << endl;
+  out.write((char*)(op.clientId.data()), (long)op.clientId.size());
```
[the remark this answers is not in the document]
yontyon: for some reason, reading the string from the stream is not working if the string has some characters. so reading the string from the stream has to be done with the read method. Here I wanted to be consistent with the deserialization method

utt/libutt/src/client.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include <utt/BurnOp.h>
+#include <utt/MintOp.h>
+#include <utt/Tx.h>
+#include <utt/DataUtils.hpp>
+#include <vector>
+#include <sstream>
+namespace libutt::api {
+Client::Client(const std::string& pid, const std::string& bpk, const std::string& rvk, const std::string& rsaSk) {
+  if (pid.empty() || bpk.empty() || rvk.empty() || rsaSk.empty())
+    throw std::runtime_error("Invalid paramets for building the client");
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/client.cpp
```diff
@@ -0,0 +1,157 @@
...
+  auto r = rcm_copy.rerandomize(d, Commitment::Type::REGISTRATION, std::nullopt);
+  Fr r_rand;
+  r_rand.from_words(r);
+  auto sig_obj = ask_->rs;
+  sig_obj.rerandomize(r_rand, Fr::random_element());
+  auto tmp = libutt::serialize<libutt::RandSig>(sig_obj);
+  return {rcm_copy, types::Signature(tmp.begin(), tmp.end())};
+}
+template <>
+std::vector<libutt::api::Coin> Client::claimCoins<operations::Mint>(
```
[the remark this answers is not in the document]
yontyon: fixed

utt/libutt/src/client.cpp
```diff
@@ -0,0 +1,157 @@
...
+                      Coin::Type::Normal,
+                      libutt::Coin::DoesNotExpire().to_words());
+  c.setSig(sig);
+  c.rerandomize(std::nullopt);
+  return {c};
+}
+
+template <>
+std::vector<libutt::api::Coin> Client::claimCoins<operations::Budget>(
+    const operations::Budget& budget, const GlobalParams& d, const std::vector<types::Signature>& blindedSigs) const {
```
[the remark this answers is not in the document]
yontyon: fixed

utt/include/globalParams.hpp
yontyon: @guyg7 suggested not using a singleton.
However, I reorganized the code such that one needs to run the following steps before starting:
1. initialize the underlying libraries
2. knowing the coin commitment key after generating the secrets for coin signing
3. knowing the registration commitment key after generating the secrets for registration signing
4. create the global params

evdzhurov: Then why keep the name "Global params" if it's not supposed to be a singleton? Also the initialization step "allocated params once", which made it behave like a singleton, but I see that you changed that.

utt/libutt/src/coinsSigner.cpp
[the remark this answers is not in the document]
evdzhurov: When do we use a non-distributed bank? And how does the caller differentiate that?

utt/tests/TestBurn.cpp
evdzhurov: But does the test verify that?

utt/libutt/src/common.cpp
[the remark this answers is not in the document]
evdzhurov: Well, considering time_since_epoch is calculated from a time_t that was obtained using timegm, we are not using local time. However, why are we even adding support for converting time formatted strings? Isn't the utt library supposed to only encode and store the bytes of a time object instead of parsing times? Why don't we simply accept a 64-bit number or however much bytes we can store in a single curve point?
Also timegm aslo doesn't seem like a standard library function.

utt/include/globalParams.hpp
```diff
@@ -20,27 +20,32 @@ class Params;
...
-   * @brief Represents a shared known global UTT params. These parameters includes commitment keys, nullifier parametrs
+   * @brief Represents a shared known global UTT params. These parametesr include commitment keys, nullifier parameters
    * and more (see uttlib/Params.h)
    *
    */
-  static bool initialized;
 
  public:
+  struct BaseLibsInitData {
+    std::string ntl_finite_field{"21888242871839275222246405745257275088548364400416034343698204186575808495617"};
```
evdzhurov, line 30: ntl_finite_field should not be a variable. This number is the same big prime used with the bn128 curve that libff uses. This number should not be a part of an interface unless we allow customization of the underlying cryptographic math, which we probably wont do for quite some time.

utt/libutt/src/api/client.cpp
```diff
@@ -129,8 +131,9 @@ template <>
 std::vector<libutt::api::Coin> Client::claimCoins<operations::Transaction>(
     const operations::Transaction& tx, const GlobalParams& d, const std::vector<types::Signature>& blindedSigs) const {
   std::vector<libutt::api::Coin> ret;
-  auto mineTransactions = tx.tx_->getMineTransactions(*decryptor_);
+  auto mineTransactions = tx.tx_->getMyTransactions(*decryptor_);
   for (const auto& [txoIdx, txo] : mineTransactions) {
```
evdzhurov, line 135: mineTransactions -> myTransactions

utt/libutt/src/api/client.cpp
```diff
@@ -129,8 +131,9 @@ template <>
...
+    if (blindedSigs.size() <= txoIdx) throw std::runtime_error("Invalid number of blinded signatures");
```
evdzhurov, line 136: Isn't it correct to simply check that the size of output coins matches the size of signatures? If that's how we're expecting them to be passed?

utt/libutt/src/api/globalParams.cpp
```diff
@@ -41,10 +38,19 @@ void GlobalParams::init() {
 #endif
 
   RangeProof::Params::initializeOmegas();
-  params.reset(new libutt::Params());
+}
+
+GlobalParams GlobalParams::create(void* initData) {
```
evdzhurov, line 43: Why do we hide the type of the initial data? How de we guard from errors?

utt/tests/testUtils.hpp
```diff
@@ -40,13 +45,13 @@ std::vector<uint32_t> getSubGroup(uint32_t n, uint32_t size) {
...
+  GlobalParams::BaseLibsInitData base_libs_init_data;
+  GlobalParams::initLibs(base_libs_init_data);
   auto dkg = RandSigDKG(thresh, n, Params::NumMessages);
-  d.getParams() = Params::random(dkg.getCK());
-  auto rc = RegAuthSK::generateKeyAndShares(d.getParams().ck_reg, thresh, n, d.getParams().ibe);
-  d.getParams().ck_reg = rc.ck_reg;
+  auto rc = RegAuthSK::generateKeyAndShares(thresh, n);
+  GpData gp_data{dkg.getCK(), rc.ck_reg};
+  GlobalParams d = GlobalParams::create((void*)(&gp_data));
+  rc.setIBEParams(d.getParams().ibe);
```
evdzhurov, line 54: I don't think this is a "clean" interface to create a UTT instance. We should probably think how to improve it.
1) Are global params really global params? If they are - why aren't we using a singleton? if they're not - why are we calling them global params?
2) RandSigDKG gives no indication that we're creating a "bank authority", for "rc" it's almost the same, although you can probably make a guess from "RegAuthSK".
3) Creation is fragmented into little steps instead of a straightforward higher-level creation step.
4) Creation of global params is unsafe because it erases the type of the argument.
5) We start by creating "rc" but then we circle back to set some addition params for IBE that we need from global params - this is complicated to follow.

evdzhurov (approved):
The memory management issues around usage of the RSA cryptography have been addressed, so I'm going to go ahead and approve. I think in terms of "cleanness" of the interface we can think about how to make the creation of the parameters and keys for a single UTT instance more straightforward and understandable.

utt/libutt/src/api/globalParams.cpp
yontyon: Because we use forward declarations, I prefer to avoid from letting the user to pass the internal libutt types

utt/tests/testUtils.hpp
yontyon: I agree this could be improved, but:
1. Global params are per UTT instance, this is why we can't use a singleton
2. I agree that we can have a better naming, but I prefer not to change the legacy code naming for now
3. I think the higher level creation needs to be in the client library, I want to let the flexibility and the ability to control every step
4. See above
5. right, but I don't think we have any other option. To create the global params we need the registration commitment key (hence we first create rc) but the IBE params are part of the global params.
We can break it into even smaller pieces and initialize the IBE params separately, but I think it would be even more complicated

utt/libutt/src/api/client.cpp
yontyon: Currently, the client simply takes all the transaction's signatures and tries to unblind them according to the txoId (which is the original tx ID in the outTx).
This simplifies the logic because it encapsulates everything in a single place.
of course we can let the client to first filter its output coins and then pickup only relevant signatures (by the tx ID) but it will require the client to understand much better the structure of the transaciton

teoparvanov (approved):
Approving, modulo the following items that still need to be addressed in the current PR:
* Let's not make assumptions about the time format (store it as just a number). Agreement not to deal with platform-specific time APIs/encodings.
* Fix seeding in getRandomSubset()
* Merge the PR with Evgeniy's approval (+Teo as Concord gatekeeper)

yontyon:
Another PR is needed to fix the tidy check tests, will be published in the next few days

PR #2744, opened 2022-07-20
Connect PreExecute and Consensus traces

bayarcher:
Connect PreExecute and Consensus traces to get complete transaction trace


yuliasherman:
Please launch Maestro performance tests (650TPS) for the whole 5 days. We should be cautious with the open trace as in the past it caused performance degradation.

bayarcher:
> Please launch Maestro performance tests (650TPS) for the whole 5 days. We should be cautious with the open trace as in the past it caused performance degradation.

I have sent the performance results in email

PR #2746, opened 2022-07-24
Don't assert if ike was not done after state transfer (#2740)

yontyon:
* **Problem Overview**  
    In case we have a VC before the initial KE, the replicas fill their active window with no-ops. Hence, a late replica may start state transfer. However, this replica will assert when state transfer is done because the initial key exchange has not been completed yet.
The solution is not to assert if the initial key exchange has not been done yet.
* **Testing Done**  
  CI


teoparvanov (approved):
LGTM, thanks Yoni!

PR #2755, opened 2022-07-28
Fix replica restart command handler

agrkushvm:
* **Problem Overview**  
Issue : On receiving the restart command, replicas should wait for stable sequence number if bft_support is false.
        But it was waiting for stable sequence number only. This was causing a timing issue where one of the replicas was not able to collect
        the stable sequence number as other replicas have already restarted.

Fix : Change callback to superstable sequence number

* **Testing Done**  
  Verified by running the test_restart_no_bft_with_restart_flag 100 times.

Signed-off-by: Kush Agrawal <kusha@vmware.com>


nkumar04 (approved):
LGTM!
Thanks!

nityavasudevan (approved):

PR #2756, opened 2022-07-29
Native build corrections

alfredd:
* **Problem Overview**  
  The native build was failing due to incorrect commands in the install_deps.sh. Additionally, the CMakeLists.txt used GCC for compiling  the code and not Clang. Both of these have been corrected. I have also updated the README.txt to reflect that the builds are supported on Ubuntu 18.04. 

* **Testing Done**  
Manual testing was done.

[install_deps.log.txt](https://github.com/vmware/concord-bft/files/9215277/install_deps.log.txt)

[make.log.txt](https://github.com/vmware/concord-bft/files/9215313/make.log.txt)

[make_test.log.txt](https://github.com/vmware/concord-bft/files/9215316/make_test.log.txt)

Failing tests are not compatible with native build execution style; i.e., they assume that docker is being used for running the tests and paths used in the tests are inconsistent with those in the native build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,6 @@
 cmake_minimum_required(VERSION 3.2)
+set(CMAKE_CXX_COMPILER "/usr/bin/clang++")
+set(CMAKE_C_COMPILER "/usr/bin/clang")
 project(concord-bft VERSION 0.1.0.0 LANGUAGES CXX)
 
 #
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ when a backwards incompatible change is made.
 
 ## Install and Build (Ubuntu Linux 18.04)
 
-Concord-BFT supports two kinds of builds: native and docker.
+Concord-BFT supports two kinds of builds: native and docker. At present both builds are supported on Ubuntu Linux 18.04. 
 
 The docker build is **strongly recommended**.
 
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -14,22 +14,22 @@ WGET_FLAGS="--https-only"
 echo ca_certificate=/etc/ssl/certs/ca-certificates.crt > ~/.wgetrc
 
 # Install tools
-apt-get update && apt-get ${APT_GET_FLAGS} install \
+sudo apt-get update && sudo apt-get ${APT_GET_FLAGS} install \
     autoconf \
     automake \
     libtool \
     build-essential \
     ccache \
-    clang-9 \
-    clang-format-10 \
-    clang-tidy-10 \
+    clang \
+    clang-format \
+    clang-tidy \
     curl \
     gdb \
     gdbserver \
     git \
     iptables \
     less \
-    llvm-9 \
+    llvm \
     lzip \
     net-tools \
     parallel \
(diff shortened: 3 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 cmake_minimum_required(VERSION 3.2)
+set(CMAKE_CXX_COMPILER "/usr/bin/clang++")
```
itsmeakashgoyal, line 2: This is not the correct position to set the path.
Do it below project below set(CMAKE_CXX_EXTENSIONS OFF)

CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
...
+set(CMAKE_C_COMPILER "/usr/bin/clang")
```
itsmeakashgoyal, line 3: same with this

README.md
```diff
@@ -38,7 +38,7 @@ when a backwards incompatible change is made.
 
 ## Install and Build (Ubuntu Linux 18.04)
 
-Concord-BFT supports two kinds of builds: native and docker.
+Concord-BFT supports two kinds of builds: native and docker. At present both builds are supported on Ubuntu Linux 18.04.
```
itsmeakashgoyal, line 41: It is already there in the heading section so its not necessary to mention here.

install_deps.sh
```diff
@@ -14,22 +14,22 @@ WGET_FLAGS="--https-only"
...
+sudo apt-get update && sudo apt-get ${APT_GET_FLAGS} install \
     autoconf \
     automake \
     libtool \
     build-essential \
     ccache \
-    clang-9 \
-    clang-format-10 \
-    clang-tidy-10 \
+    clang \
```
itsmeakashgoyal, line 23: We are using a specific version of clang. Are you sure when you install clang package it will install the this specific clang version?

install_deps.sh
```diff
@@ -14,22 +14,22 @@ WGET_FLAGS="--https-only"
...
+    clang-format \
+    clang-tidy \
     curl \
     gdb \
     gdbserver \
     git \
     iptables \
     less \
-    llvm-9 \
+    llvm \
```
itsmeakashgoyal, line 32: We are currently supporting llvm-9 so my question is same like it was before for clang

install_deps.sh
```diff
@@ -43,21 +43,13 @@ apt-get update && apt-get ${APT_GET_FLAGS} install \
     iproute2 \
     wget
 
-update-alternatives --install /usr/bin/clang clang /usr/lib/llvm-9/bin/clang 100
```
itsmeakashgoyal, line 46: If we remove these alternatives it will broke down docker build functionality.
So these are important for docker build.

itsmeakashgoyal: if there are many changes needs to be done in exisiting install_deps file than I will recommend you to create a new install_deps file specific to native build.

itsmeakashgoyal (commented):
I have updated my comments. Can you please check.

toly-kournik:
Hi, I'm not sure this PR is essential:
compiler is not explicitly specified to support compiling with various compilers, just specify:
`CXX=clang CC=clang cmake ..`

Also I don't understand necessity the changes in the `install_deps.sh`

CMakeLists.txt
alfredd: We can't do it below that line. It has to be declared [before the project variables are set](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#how-do-i-use-a-different-compiler). This method is marked to be avoided. The better method is to specify using the CLI option -D. If that's OK, I'll update the README instructions on how to compile via Clang by passing the language variable via the -D option. I will remove this line from the CMakeLists.txt.

install_deps.sh
alfredd: I will revert this change. I checked that clang version 6.0 was being installed on Ubuntu 18.04. However, these commands will have to be changed once we support Ubuntu 20.04 or higher.

shgandhi:
@alfredd let's close this PR, since we no longer need it.

alfredd:
Closing this PR as I created a new PR with only necessary changes. https://github.com/vmware/concord-bft/pull/2763

PR #2760, opened 2022-07-31
Save and load genesis block id from storage

cloudnoize:
* **Problem Overview**  
 After pruning, finding the genesis block using RocksDB iterators took a significant duration since RocksDB iterates over the whole range, which can be millions of blocks, to find the first key that is not deleted.
The fix is that when pruning is performed, save the genesis id in storage, and on load, first look for the key, if it exists, load the genesis from it.

* **Testing Done**  
  Unit test that covers all paths that set and load the genesis

### DIFF
--- a/kvbc/src/v4blockchain/detail/blockchain.cpp
+++ b/kvbc/src/v4blockchain/detail/blockchain.cpp
@@ -24,6 +24,9 @@ Blockchain::Blockchain(const std::shared_ptr<concord::storage::rocksdb::NativeCl
   if (native_client->createColumnFamilyIfNotExisting(v4blockchain::detail::BLOCKS_CF)) {
     LOG_INFO(V4_BLOCK_LOG, "Created [" << v4blockchain::detail::BLOCKS_CF << "] column family for the main blockchain");
   }
+  if (native_client_->createColumnFamilyIfNotExisting(v4blockchain::detail::MISC_CF)) {
+    LOG_INFO(V4_BLOCK_LOG, "Created [" << v4blockchain::detail::MISC_CF << "] column family");
+  }
   auto last_reachable_block_id = loadLastReachableBlockId();
   if (last_reachable_block_id) {
     last_reachable_block_id_ = last_reachable_block_id.value();
@@ -31,6 +34,9 @@ Blockchain::Blockchain(const std::shared_ptr<concord::storage::rocksdb::NativeCl
   }
   auto genesis_blockId = loadGenesisBlockId();
   if (genesis_blockId) {
+    ConcordAssertGT(genesis_blockId.value(), INVALID_BLOCK_ID);
+    ConcordAssertLE(genesis_blockId.value(), last_reachable_block_id_);
+    ConcordAssert(!hasBlock(genesis_blockId.value() - 1));
     setGenesisBlockId(genesis_blockId.value());
     LOG_INFO(V4_BLOCK_LOG, "Genesis block was loaded from storage " << genesis_block_id_);
   }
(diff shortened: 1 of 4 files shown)


kvbc/src/v4blockchain/detail/blockchain.cpp
```diff
@@ -24,6 +24,9 @@ Blockchain::Blockchain(const std::shared_ptr<concord::storage::rocksdb::NativeCl
   if (native_client->createColumnFamilyIfNotExisting(v4blockchain::detail::BLOCKS_CF)) {
     LOG_INFO(V4_BLOCK_LOG, "Created [" << v4blockchain::detail::BLOCKS_CF << "] column family for the main blockchain");
   }
+  if (native_client_->createColumnFamilyIfNotExisting(v4blockchain::detail::MISC_CF)) {
```
glevkovich, line 27: 2 things:
1) nit: consider  to have a more meaningful name? I  can't  really understand much from this name unless I go and search what is stored under it.
e.g METADATA_CF or ADDITIONAL_INFO_CF

2) Why do you LOG_PRINT outside the function? Why not print inside createColumnFamilyIfNotExisting if created? Do you want to avoid to many prints? for that toy can also pass a flag e.g print_on_creation with default set to false.

kvbc/src/v4blockchain/detail/blockchain.cpp
```diff
@@ -197,8 +202,16 @@ std::optional<BlockId> Blockchain::loadLastReachableBlockId() {
   return concordUtils::fromBigEndianBuffer<BlockId>(itr.keyView().data());
 }
 
-// get the closest key to INITIAL_GENESIS_BLOCK_ID
+// if the genesis_block_key is found it means that we had pruning and we saved the genesis id in storage
+// in order to save iteration over a deleted range that takes time.
+// if it wasn't found we get the closest value to INITIAL_GENESIS_BLOCK_ID.
 std::optional<BlockId> Blockchain::loadGenesisBlockId() {
+  auto opt_val = native_client_->get(v4blockchain::detail::MISC_CF, kvbc::keyTypes::genesis_block_key);
```
glevkovich, line 209: Maybe you already do it somewhere else - should we have a sanity check to see that this information is valid?
Try to see that for the blockId fetched from genesis_block_key, ID-1 is not there? This should take no time and makes sure we don't have some DB corruption.

kvbc/test/v4blockchain/blockchain_test.cpp
```diff
@@ -681,6 +681,72 @@ TEST_F(v4_blockchain, block_updates) {
...
 }
 
+TEST(v4_blockchain_genesis, load_genesis) {
+  // 1 - add blocks.
+  // 2 - test genesis
+
+  std::random_device seed;
+  std::mt19937 gen{seed()};                       // seed the generator
+  std::uniform_int_distribution dist{100, 1000};  // set min and max
+  auto rand_number = (uint64_t)dist(gen);         // generate number
```
glevkovich, line 691: nit:
auto rand_number = std::static_cast<uint64_t>(dist(gen));

glevkovich (approved):
Thanks @cloudnoize , left few minor questions/comments

kvbc/src/v4blockchain/detail/blockchain.cpp
cloudnoize: ```createColumnFamilyIfNotExisting``` is an API call; I prefer leaving it as clean as possible.

kvbc/src/v4blockchain/detail/blockchain.cpp
cloudnoize: will add it, thanks!

eyalrund (approved):

PR #2761, opened 2022-08-02
prevent unwedge on an unewedged system

yontyon:
* **Problem Overview**  
  If we unwedge an already unwedged (or never wedged) system, we may end up with an inconsistent state
* **Testing Done**  
  CI


cloudnoize (approved):
LGTM!

PR #2762, opened 2022-08-02
Optimize immutable writes

cloudnoize:
Cherry-pick of - https://github.com/vmware/concord-bft/pull/2734

Immutable keys by agreement are not being read by getLatest API but by get with block id.
Therefore saving their value in the latest CF is redundant.
For safety, in case the application tries to read the latest of on immutable, we'll save the version of the immutable in the latest CF, and perform indirection by first reading its version and then reading the value from the block.

A counter metric was added in order to quickly observe the usage of the getLatest API with immutables.
Tests:
UT that use getLatest for immutable pass.
Apollo and Concord CI.
Running long runs:

Reduciton of ~25% in storage
no calls for getLatest with immutable


arc-vmware (approved):
LGTM!!!

eyalrund (approved):

PR #2763, opened 2022-08-03
Native build update

alfredd:
* **Problem Overview**  
  Updated the README file to show how compilation can be done via command line using clang as the default compiler.. Moved installation scripts in install_deps.sh into functions that are called at the end. This allows the installation scripts to be arranged properly in the script. Turned off Apollo tests from the native build compilation and testing as they are written to be executed using the docker build setup. The Apollo tests are still executed in the docker build and they can also be executed in the native build setup by following instructions in the README file. 
* **Testing Done**  
Manual Testing was done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,11 @@ make
 sudo make test
 ```
 
+By default the build chooses the active compiler on the native build platform. In order to force compilation by clang you can use the following command.
+```sh
+CC=clang CXX=clang++ cmake ..
+```
+
 In order to turn on or off various options, you need to change your cmake configuration. This is
 done by passing arguments to cmake with a `-D` prefix: e.g. `cmake -DBUILD_TESTING=OFF`. Note that
 make must be run afterwards to build according to the configuration. Please see [CMakeLists.txt](CMakeLists.txt) for configurable options.
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -14,326 +14,383 @@ WGET_FLAGS="--https-only"
 echo ca_certificate=/etc/ssl/certs/ca-certificates.crt > ~/.wgetrc
 
 # Install tools
-apt-get update && apt-get ${APT_GET_FLAGS} install \
-    autoconf \
-    automake \
-    libtool \
-    build-essential \
-    ccache \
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -23,8 +23,8 @@ if (NOT DEFINED USE_LOG4CPP)
   option(USE_LOG4CPP "Enable LOG4CPP" ON)
 endif()
 
-option(RUN_APOLLO_TESTS "Enable Apollo tests run" ON)
-option(KEEP_APOLLO_LOGS "Retains logs from replicas in separate folder for each test in build/tests/apollo/logs" ON)
+option(RUN_APOLLO_TESTS "Enable Apollo tests run" OFF)
```
toly-kournik, line 26: why have you disabled these options

alfredd: The Apollo tests in native builds always fails in my testing. This is because during the build they look for certain files in locations that would be available in the docker builds. There are instructions in the README file on how to run the Apollo tests. So instead of running failing tests during native build (via Cmake) which takes a lot of time, users who want to run native build can run Apollo tests via instructions in the README file and save time on running the tests via make test.

toly-kournik: AFAIR there are two scripts addressed by an absolute path docker path `/conncord/*`: one for transaction signing and one for certificates generation. This should be fixed.

alfredd: In that case I will enable this flag for now. I will raise the issue of the failing Apollo tests as a separate issue.

toly-kournik (approved):
LGTM!

shgandhi:
Thanks for the change @alfredd. LGTM.
Nit: please add a description whenever possible to each commit message. For example, for the commit message [Restructuring install_deps.sh script](https://github.com/vmware/concord-bft/pull/2763/commits/36ab1188dbb4f10615d0049ed5e96dac694af243), a summary of what the commit includes, and why the change was needed i.e.,`what and why` (pls see https://cbea.ms/git-commit/).

shgandhi (approved):

PR #2768, opened 2022-08-03
Introduce ccache: a compiler cache

glevkovich:
* **Problem Overview**  

Introduce ccache: a compiler cache

https://ccache.dev/
Ccache is a compiler cache. It speeds up recompilation by caching
previous compilations and detecting when the same compilation is
being done again. Ccache is free software, released under the GNU
General Public License version 3 or later. See also the license page.

Important points to mention:
*) ccache existed on concord-bft docker image but wasn't used
till now, and the version was old. Upgraded to newest version 4.6.1.
*) For now, ccache is configured to work only locally using Makefile
dockerized build, or by using CMake.
*) ccache configuration file is under .ccache/ccache.conf
*) The cache is stored under $HOME/.ccache. It is positioned there
to cache information
from multiple workspaces/projects.
*) ccache is enabled by default. To disable ccache, run build with
CONCORD_BFT_CMAKE_CCACHE=FALSE or set disable in ccache.conf to false
*) There are 2  targets added/modified in Makefile:
create-ccache-folder: is used internally to create ccache folder
clean: will clean build folder by default, or you can clean everything
with the flag CLEAN_ALL set to TRUE. If you like to clean only untracked
files use CLEAN_UNTRACKED=TRUE. If you like to clean ccache cache use 
CLEAN_CACHE=TRUE.
*) For native linux, It is assumed that caller has called cmake and
install_deps.sh.

* **Testing Done**  
Mainly manual tests, to see cache really works.

### DIFF
--- a/.ccache/ccache.conf
+++ b/.ccache/ccache.conf
@@ -1,0 +1,26 @@
+# For complete documentation, see https://ccache.dev/manual/4.6.1.html#_configuration
+# Comments:
+# 1) By default, cache_dir is in $HOME/.ccache
+# 2) Dockerized builds might override parameters in this file, or defaults by using enviorment variables
+#    (which have  precedence)
+
+# When true, ccache will just call the real compiler, bypassing the cache completely. The default is false.
+disable = false
+
+# This option specifies the maximum size of the cache. Use 0 for no limit. The default value is 5G.
+# Available suffixes: k, M, G, T (decimal) and Ki, Mi, Gi, Ti (binary). The default suffix is G
+max_size = 10G
+
+#If true, ccache will compress data it puts in the cache. However, this option has no effect on how files are
+#retrieved from the cache; compressed and uncompressed results will still be usable regardless of this option.
+# The default is true.
+compression = true
+
+# This option determines the level at which ccache will compress object files using the real-time compression
+# algorithm Zstandard. It only has an effect if compression is enabled (which it is by default). In restricted space environments,
+# it is recommended to set the value of this parameter to 19 (or via environment parameter CCACHE_COMPRESSLEVEL)
+# Currently set to 1 to have minimize compression, with priority for speed.
+compression_level = 1
+
+# If true, ccache will not discard the comments before hashing preprocessor output.
+keep_comments_cpp = true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,3 +233,4 @@ if(BUILD_TESTING)
     add_subdirectory(tests)
     add_subdirectory(messages)
 endif()
+
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 CONCORD_BFT_DOCKER_REPO?=concordbft/
 CONCORD_BFT_DOCKER_IMAGE?=concord-bft
-CONCORD_BFT_DOCKER_IMAGE_VERSION?=0.43
+CONCORD_BFT_DOCKER_IMAGE_VERSION?=0.44
 CONCORD_BFT_DOCKER_CONTAINER?=concord-bft
 
 CONCORD_BFT_DOCKERFILE?=Dockerfile
@@ -58,6 +58,7 @@ CONCORD_BFT_CMAKE_ASAN?=FALSE
 CONCORD_BFT_CMAKE_TSAN?=FALSE
 CONCORD_BFT_CMAKE_UBSAN?=FALSE
 CONCORD_BFT_CMAKE_CODECOVERAGE?=FALSE
+CONCORD_BFT_CMAKE_CCACHE?=TRUE
 CONCORD_BFT_CMAKE_USE_FAKE_CLOCK_IN_TIME_SERVICE?=FALSE
 ENABLE_RESTART_RECOVERY_TESTS?=FALSE
 
@@ -71,6 +72,13 @@ endif
 ifeq (${CONCORD_BFT_CMAKE_CODECOVERAGE},TRUE)
 	CONCORD_BFT_CMAKE_CXX_FLAGS_RELEASE='-O0 -g'
 endif
+
+# The consistency parameter makes sense only at MacOS.
+# It is ignored at all other platforms.
+CONCORD_BFT_CONTAINER_MOUNT_CONSISTENCY?=,consistency=cached
+CONCORD_BFT_CTEST_TIMEOUT?=3000 # Default value is 1500 sec. It takes 2500 to run all the tests at my dev station
+CONCORD_BFT_USER_GROUP?=--user `id -u`:`id -g`
+CONCORD_BFT_CORE_DIR?=${CONCORD_BFT_TARGET_SOURCE_PATH}/${CONCORD_BFT_BUILD_DIR}/cores
 
 CONCORD_BFT_CMAKE_FLAGS?= \
 			-DCMAKE_BUILD_TYPE=${CONCORD_BFT_CMAKE_BUILD_TYPE} \
@@ -99,17 +107,19 @@ CONCORD_BFT_CMAKE_FLAGS?= \
 			-DTXN_SIGNING_ENABLED=${CONCORD_BFT_CMAKE_TRANSACTION_SIGNING_ENABLED} \
 			-DENABLE_RESTART_RECOVERY_TESTS=${ENABLE_RESTART_RECOVERY_TESTS}
 
-
-# The consistency parameter makes sense only at MacOS.
-# It is ignored at all other platforms.
-CONCORD_BFT_CONTAINER_MOUNT_CONSISTENCY?=,consistency=cached
-CONCORD_BFT_CTEST_TIMEOUT?=3000 # Default value is 1500 sec. It takes 2500 to run all the tests at my dev station
-CONCORD_BFT_USER_GROUP?=--user `id -u`:`id -g`
-CONCORD_BFT_CORE_DIR?=${CONCORD_BFT_TARGET_SOURCE_PATH}/${CONCORD_BFT_BUILD_DIR}/cores
-
 CONCORD_BFT_ADDITIONAL_RUN_PARAMS?=
 APOLLO_CTEST_RUN_PARAMS?=
 CONCORD_BFT_FORMAT_CMD=make format-check &&
+
+ifeq (${CONCORD_BFT_CMAKE_CCACHE},TRUE)
+	CCACHE_HOST_CACHE_DIR=${HOME}/.ccache/
+	CCACHE_CONTAINER_CACHE_DIR=/mnt/ccache/
+	CONCORD_BFT_CMAKE_FLAGS+=-DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache
+	CONCORD_BFT_ADDITIONAL_RUN_PARAMS+=\
+		--mount type=bind,source=${CCACHE_HOST_CACHE_DIR},target=${CCACHE_CONTAINER_CACHE_DIR}${CONCORD_BFT_CONTAINER_MOUNT_CONSISTENCY} \
+		--env CCACHE_CONFIGPATH=/concord-bft/.ccache/ccache.conf \
+		--env CCACHE_DIR=/mnt/ccache/
+endif
 
 ifneq (${APOLLO_LOG_STDOUT},)
 	CONCORD_BFT_ADDITIONAL_RUN_PARAMS+=--env APOLLO_LOG_STDOUT=TRUE
@@ -157,7 +167,7 @@ login: ## Login to the container. Note: if the container is already running, log
 	fi
 
 .PHONY: gen-cmake
-gen-cmake: ## Generate cmake files, used internally
+gen-cmake: create-ccache-folder ## Generate cmake files, used internally
 	docker run ${CONCORD_BFT_USER_GROUP} ${BASIC_RUN_PARAMS} \
 		${CONCORD_BFT_CONTAINER_SHELL} -c \
 		"mkdir -p ${CONCORD_BFT_TARGET_SOURCE_PATH}/${CONCORD_BFT_BUILD_DIR} && \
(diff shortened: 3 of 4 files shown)


.ccache/ccache.conf
```diff
@@ -0,0 +1,32 @@
+# For complete documentation, see https://ccache.dev/manual/4.6.1.html#_configuration
+
+# This option specifies the maximum size of the cache. Use 0 for no limit. The default value is 5G.
+# Available suffixes: k, M, G, T (decimal) and Ki, Mi, Gi, Ti (binary). The default suffix is G
+max_size = 10G
+
+# This option specifies where ccache will keep its cached compiler outputs.
+cache_dir = /mnt/ccache/
```
WildFireFlum, line 8: shouldn't this be in the build directory or under bft's root?

.ccache/ccache.conf
```diff
@@ -0,0 +1,32 @@
...
+
+#If true, ccache will compress data it puts in the cache. However, this option has no effect on how files are
+#retrieved from the cache; compressed and uncompressed results will still be usable regardless of this option.
+# The default is true.
+compression = true
+
+# This option determines the level at which ccache will compress object files using the real-time compression
+# algorithm Zstandard. It only has an effect if compression is enabled (which it is by default). In restricted space environments,
+# it is recommended to set the value of this parameter to 19 (or via environment parameter CCACHE_COMPRESSLEVEL)
+compression_level = 1
```
WildFireFlum, line 18: why compress if you use the lowest possible compression rate?

.ccache/ccache.conf
```diff
@@ -0,0 +1,32 @@
...
+
+# When true, ccache will just call the real compiler, bypassing the cache completely. The default is false.
+disable = false
+
+# If set to a file path, ccache will write information on what it is doing to the specified file.
+# This is useful for tracking down problems.
+log_file = /mnt/ccache/ccache.log
```
WildFireFlum, line 25: Should probably be local

.ccache/ccache.conf
```diff
@@ -0,0 +1,32 @@
...
+
+# If set to a file path, ccache will write statistics counter updates to the specified file. This is useful
+# for getting statistics for individual builds. To show a summary of the current stats log, use ccache --show-log-stats.
+stats_log = /mnt/ccache/ccache_stats.log
```
WildFireFlum, line 29: should be local

.ccache/ccache.conf
```diff
@@ -0,0 +1,32 @@
...
+
+# If true, ccache will not discard the comments before hashing preprocessor output.
+keep_comments_cpp = true
```
WildFireFlum, line 32: Shouldn't this be set to false?

Makefile
```diff
@@ -157,7 +166,7 @@ login: ## Login to the container. Note: if the container is already running, log
 	fi
 
 .PHONY: gen-cmake
-gen-cmake: ## Generate cmake files, used internally
+gen-cmake: create-ccache-folder ## Generate cmake files, used internally
```
WildFireFlum, line 169: If the cache directory is under bft, you can use mkdir to create it instead of a separate target

Makefile
```diff
@@ -326,20 +335,28 @@ test-single-gtest-case: ## Run a single GoogleTest test case: `make test-single-
...
 .PHONY: clean
-clean: ## Clean Concord-BFT build directory
+clean: create-ccache-folder ## Clean Concord-BFT build directory
 	docker run ${BASIC_RUN_PARAMS} \
 		${CONCORD_BFT_CONTAINER_SHELL} -c \
 		"rm -rf ${CONCORD_BFT_BUILD_DIR}"
 
 .PHONY: clean-all
-clean-all: ## Clean Concord-BFT build directory and any other untracked/ignored files.For a 'dry run' use DRY_RUN=TRUE, for an interactive run use INTER=TRUE.
+clean-all: ## Clean Concord-BFT build directory and any other untracked/ignored files. Does not include ccache cache folder.For a 'dry run' use DRY_RUN=TRUE, for an interactive run use INTER=TRUE.
```
WildFireFlum, line 344: I believe that this should also remove the ccache (same effect as removing the build directory), otherwise the name is misleading.

install_deps.sh
```diff
@@ -41,7 +40,8 @@ apt-get update && apt-get ${APT_GET_FLAGS} install \
     sudo \
     vim \
     iproute2 \
-    wget
+    wget \
+    libhiredis-dev
```
WildFireFlum, line 44: Does ccache use redis?

install_deps.sh
```diff
@@ -363,6 +363,23 @@ git clone -b v0.7.0 --depth 1 https://github.com/jaegertracing/jaeger-client-cpp
...
+cd "$HOME"
+wget https://github.com/ccache/ccache/releases/download/v${CCACHE_VER}/ccache-${CCACHE_VER}.tar.xz && \
+  tar xvf ccache-${CCACHE_VER}.tar.xz && \
+  cd ccache-${CCACHE_VER} && \
+  mkdir build && cd build && \
+  cmake -DCMAKE_BUILD_TYPE=Release .. && make && make install && \
+  cd "${HOME}" && \
+  rm -rf ccache-${CCACHE_VER} ccache-${CCACHE_VER}.tar.xz
+CCACHE_PATH=$(which ccache)
+ln -s "${CCACHE_PATH}" /usr/local/bin/gcc
```
WildFireFlum, line 377: do you have to create these links manually? Isn't there a ccache script that does it?

WildFireFlum (changes requested):

eyalrund (approved):

Makefile
```diff
@@ -72,6 +73,13 @@ ifeq (${CONCORD_BFT_CMAKE_CODECOVERAGE},TRUE)
 	CONCORD_BFT_CMAKE_CXX_FLAGS_RELEASE='-O0 -g'
 endif
 
+# The consistency parameter makes sense only at MacOS.
```
toly-kournik, line 76: How is this related to ccache?

toly-kournik, line 76: What about native build?

toly-kournik:
Just want to ensure ccache works with a native build

.ccache/ccache.conf
glevkovich: No, as his is done on purpose.
First, we need to remember that /mnt/ccache is a folder mount from the host (for containerized).
Also, the default ccache recommended cache is under ${HOME}, see manual.
All the idea of ccache is to boost your compilations time between projects and between builds after 'make clean.
If you put it in the build you are losing most of the idea of the cache. The cache will be deleted when you clean, or when you make a new project or workspace it won't help anything + it will be duplicated in different projects. And I want to share it between projects.

For example, I have 3 vmewathena where I work on vmwathenaX/concord/submodule/concord-bft/, and I also have 3 concord-bft separate workspaces/projects (total 6 projects). When I clean them, I don't want to clean the cache (which is what will happen when calling 'make clean') and I want to share a single cache among 6 projects + I don't want a 'make clean' to clean my cache. So, the idea is that the cache is not part of the build folder. To delete the cache you call make 'clean-ccache-cache'

.ccache/ccache.conf
glevkovich: - Lowest compression rate still compress it more than 80% and the space on local build is not unlimited, so I wanted to have compression still.
- I first had it on 19 (higest for all versions) to check how it will be on CI. The differnce in speed in 5-10% 19 vs 1.
- I prefer speed over space for local, and space over speed on CI. still, 1 gives around 80% compression on very minimal cost 1-3% of the speed.

.ccache/ccache.conf
glevkovich: I'm going to move both log files into concord-bft/build/ccache/

.ccache/ccache.conf
glevkovich: I'm going to move both log files into concord-bft/build/ccache/

.ccache/ccache.conf
glevkovich: The default is false, my idea was that ccache processor shouldn't touch the file content at all, just hash them. Anyway, I don't think it matters much.

Makefile
glevkovich: cach folder (as we discussed) should not be under bft

install_deps.sh
glevkovich: This is for future use - secondary cache.Rethinking - I'm going to remove it for now. If we want to use secondary redis cache in the future, we just add the support.

install_deps.sh
glevkovich: CMAKE_CXX_COMPILER_LAUNCHER/CMAKE_C_COMPILER_LAUNCHER might crate the links but then the image is removed and the linsk too, so if we want permanent links we must do this as part of the image.
Still, without ovverriding the path nothing will work, so it's ok to put the links here even if ccache is disabled.

glevkovich:
It’s a problem to set as ${HOME}/.ccache inside container since we are not running as root when building. we are running as user id 1000 (non root) so we dont have permission to that folder but it is still set as out home folder.

.ccache/ccache.conf
glevkovich: Hi, I've decided to remove this log completely for now for multiple reasons. In case we need to debug i'll add it.

.ccache/ccache.conf
glevkovich: Hi, I've decided to remove this log completely for now for multiple reasons. In case we need to debug i'll add it.

Makefile
glevkovich: That's a bit of a problem. What if I just want to clean the whole local folder, build and any leftovers and not clean ccache? This is usually what I want. I agree name might be a bit misleading but the idea is to clean everything inside the current root project folder, and ccache cache is not under that folder, so it's a matter of interpertation - everything is cleaned under concord-bft, we don't delete /tmp/* as well, right?

ccache deletion should be explicit by running clean-ccache-cache or rm -rf $HOME/.ccache/*

install_deps.sh
glevkovich: removed this installed.

.ccache/ccache.conf
```diff
@@ -0,0 +1,26 @@
+# For complete documentation, see https://ccache.dev/manual/4.6.1.html#_configuration
+# Comments:
+# 1) By default, cache_dir is in $HOME/.ccache
```
cloudnoize, line 3: Should the cache_dir be inside the build dir?

Makefile
```diff
@@ -326,20 +336,28 @@ test-single-gtest-case: ## Run a single GoogleTest test case: `make test-single-
...
 .PHONY: clean
-clean: ## Clean Concord-BFT build directory
+clean: create-ccache-folder ## Clean Concord-BFT build directory
 	docker run ${BASIC_RUN_PARAMS} \
 		${CONCORD_BFT_CONTAINER_SHELL} -c \
 		"rm -rf ${CONCORD_BFT_BUILD_DIR}"
 
 .PHONY: clean-all
-clean-all: ## Clean Concord-BFT build directory and any other untracked/ignored files.For a 'dry run' use DRY_RUN=TRUE, for an interactive run use INTER=TRUE.
+clean-all: ## Clean Concord-BFT build directory and any other untracked/ignored files. Does not include ccache cache folder.For a 'dry run' use DRY_RUN=TRUE, for an interactive run use INTER=TRUE.
```
cloudnoize, line 345: Why it does not include the cache dir?
 When we clean, we want a fresh build without any use of cache and reuse of objects no?

.ccache/ccache.conf
glevkovich: This is done intentionally.They must be inside this dir for few reasons,  (this is also the recommended default by ccache) with the main ones:

- To supshare cache between multiple projects/workspaces. Not everyone work with Git worktrees.
- Even when you clean or when you clone, ccache will go here. ccache is transperent to cleaning the build folder.

Makefile
glevkovich: Well, my idea says there are 3 levels of clean, but I guess there is no single truth here. What makes sense to me might not make sense to others:

- 'make clean' is regular clean of project bin -we dont want to delete ccache since we want to clean local build foder, and keep shared cache with other projects.
- make 'clean-all' - clean build and all repository artifacts. it still cleans things under CURRENT project. ccache is above prohject/folder/concord-bft level, it is a shared (or remote) cache shared by all users (projects, hosts or users)
- make 'clean-ccache-cache' - the only expclit command which will clean the ccache, and for me it makes sensse since cleaning the whole shared cache for everyone shouldn't be on the project level. this one goes to $HOME/.ccache/ by default and delete everything under.

glevkovich: Im going to add explentation in confluence about it.
It think my logic make sense here. Other people might find a different solution more appropriate.

glevkovich: I think all the many cleans created confusion.
Tere going to be a single make clean which cleans only build folder. Using flags it will be possible to clean different things. For more infor see confluene page or make help.

arc-vmware (approved):
LGTM!!!
Thanks @glevkovich for adding this. Will it be a good idea to add this in [Readme](https://github.com/vmware/concord-bft/blob/master/README.md)? If it's available in make help then its fine.

CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ option(TXN_SIGNING_ENABLED "Enable External concord client transcattion signing"
 option(BUILD_COMM_TCP_PLAIN "Enable TCP communication" OFF)
 option(LEAKCHECK "Enable Address and Leak Sanitizers" OFF)
 option(THREADCHECK "Enable Thread Sanitizer" OFF)
+option(USE_CCACHE "Enable ccache to speed up recompilation" ON)
```
toly-kournik, line 32: this doesn't work with a native build in a conjunction with the `if (USE_CCACHE)` later in this file.
I propose not to change `CMakeLists.txt` at all, but rather call it as recommended `cmake -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ..` from an outside `Makefile`

install_deps.sh
```diff
@@ -401,51 +400,70 @@ wget ${WGET_FLAGS} https://archive.apache.org/dist/thrift/0.11.0/thrift-0.11.0.t
...
+    tar xvf ccache-${CCACHE_VER}.tar.xz && \
+    cd ccache-${CCACHE_VER} && \
+    mkdir build && cd build && \
+    cmake -DCMAKE_BUILD_TYPE=Release -DREDIS_STORAGE_BACKEND=OFF .. && \
+    make && \
+    make install && \
+    cd "${HOME}" && \
+    rm -rf ccache-${CCACHE_VER} ccache-${CCACHE_VER}.tar.xz
+  CCACHE_PATH=$(which ccache)
+  ln -s "${CCACHE_PATH}" /usr/local/bin/gcc
```
toly-kournik, line 460: why make this symbolic links?
the compiler command should look like:
`ccache /usr/local/bin/clang++ ...` making it possible to choose between enabling/disabling ccache usage and choosing a compiler

glevkovich: There are 2 ways to run ccache and the symbolic link is a valid one. ccache knows how to find the compiler according to the nvoked sym link name, so there is no way it will not run the right g++/clang++.
To simplify native build, It's better to have it without sym links. I'll check if it's possible to run it the way you suggest..

glevkovich: We don't need to use ccache explcitly. I checked that. the CMake CMAKE_CXX_COMPILER_LAUNCHER definition do the trick, all we need to do is to pass it as CMake definition when building

toly-kournik (approved):
LGTM from the point of view of a native build!

WildFireFlum (approved):

PR #2769, opened 2022-08-03
 Add a unit test for 'no pre-processing consensus reached' situation

yuliasherman:
* **Problem Overview**
  - Remove temporary configuration parameter batchedPreProcessEnabled from PreProcessor.  
  - Add a missing PreProcessor test to verify that all resources are released in case the batch gets canceled.
* **Testing Done**  
  Concord CI


HristoStaykov (commented):
Thanks @yuliasherman for adding this additional unit test. I would recommend adding a brief description before the test to help the reader get familiar quicker with what is happening. Also if possible it will be good to have comments on the most significant steps in the test and what step in the test causes the effect of not reaching consensus.

HristoStaykov (approved):
LGTM, thanks, @yuliasherman !

PR #2770, opened 2022-08-04
State Transfer : Construct RVT in absense of metadata

mdarade:
During upgrade, when metadata is explicitly removed, RVT should
still be able to get constructed from storage.

* **Problem Overview**  
As part of upgrade, metadata is cleaned up before spinning 1.7 blockchain with existing RocksDB data. Whenever replicas gets started, they try to build RVT either from last stored checkpoint descriptors from metadata or from RocksDB storage. Existing checks had assumption that metadata would always be present (to refer to desc.maxBlockId, desc.rvbData) and there was neither Apollo nor unit test present to check such scenario. Failure at not building RVT during boot up would end up building it at first checkpoint. This would result in replicas taking considerable time (based on number of blocks in RocksDB) to commit very first checkpoint after upgrade which can also subsequently end up triggering client request timeouts.

* **Testing Done**  

Added Apollo test to reproduce issue and validated fix.
Added GTest to reproduce issue and validated fix.

### DIFF
--- a/bftengine/src/bcstatetransfer/RVBManager.cpp
+++ b/bftengine/src/bcstatetransfer/RVBManager.cpp
@@ -46,14 +46,18 @@ RVBManager::RVBManager(const Config& config, const IAppState* state_api, const s
 }
 
 void RVBManager::init(bool fetching) {
-  LOG_TRACE(logger_, "");
-  bool loaded_from_data_store = false;
+  LOG_INFO(logger_, std::boolalpha << KVLOG(fetching));
+  bool loaded_from_stored_checkpoint = false;
+  bool hasCheckpointBeingFetched{ds_->hasCheckpointBeingFetched()};
   static constexpr bool print_rvt = true;
   CheckpointDesc desc{0};
+  uint64_t last_stored_cp_num{};
 
-  if (ds_->hasCheckpointBeingFetched()) {
+  if (hasCheckpointBeingFetched) {
     ConcordAssert(fetching);
     desc = ds_->getCheckpointBeingFetched();
+    ConcordAssertGT(desc.maxBlockId, 0);
+    ConcordAssertGT(desc.checkpointNum, 0);
   } else {
     // unknown state for RVBM
     auto last_stored_cp_num = ds_->getLastStoredCheckpoint();
@@ -62,11 +66,19 @@ void RVBManager::init(bool fetching) {
       last_checkpoint_desc_ = desc;
     }
   }
+  LOG_INFO(logger_,
+           std::boolalpha << KVLOG(last_stored_cp_num,
+                                   hasCheckpointBeingFetched,
+                                   desc.checkpointNum,
+                                   desc.maxBlockId,
+                                   desc.digestOfMaxBlockId,
+                                   desc.digestOfResPagesDescriptor));
 
   // Get pruned blocks digests
   pruned_blocks_digests_ = ds_->getPrunedBlocksDigests();
   metrics_.pruning_vector_elements_count_.Get().Set(pruned_blocks_digests_.size());
   metrics_.pruning_vector_size_in_bytes_.Get().Set(pruned_blocks_digests_.size() * (sizeof(BlockId) + sizeof(Digest)));
+  LOG_INFO(logger_, KVLOG(pruned_blocks_digests_.size()));
 
   if (config_.enableStoreRvbDataDuringCheckpointing) {
     // Try to get RVT from persistent storage. Even if the tree exist we need to check if it
@@ -74,8 +86,8 @@ void RVBManager::init(bool fetching) {
     if ((desc.checkpointNum > 0) && (!desc.rvbData.empty())) {
       // There is RVB data in this checkpoint - try to load it
       std::istringstream rvb_data(std::string(reinterpret_cast<const char*>(desc.rvbData.data()), desc.rvbData.size()));
-      loaded_from_data_store = in_mem_rvt_->setSerializedRvbData(rvb_data);
-      if (!loaded_from_data_store) {
+      loaded_from_stored_checkpoint = in_mem_rvt_->setSerializedRvbData(rvb_data);
+      if (!loaded_from_stored_checkpoint) {
         LOG_ERROR(logger_, "Failed to load RVB data from stored checkpoint" << KVLOG(desc.checkpointNum));
         metrics_.failures_while_setting_serialized_rvt_++;
       } else {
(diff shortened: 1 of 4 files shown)


eyalrund (approved):

bftengine/src/bcstatetransfer/RVBManager.cpp
```diff
@@ -95,15 +107,21 @@ void RVBManager::init(bool fetching) {
     }
   }
 
-  if (!loaded_from_data_store && (desc.maxBlockId > 0)) {
+  // Do not attempt to build RVT in case
+  // 1. RVT is already build from RVB data stored in last checkpoint
+  // 2. Max Block Id from last stored checkpoint is 0
+  // 3. last reachable block == default genesis block (eg. 1)
+
+  if ((!loaded_from_data_store) && ((desc.maxBlockId) || (last_reachable_block_num > 1))) {
```
glevkovich, line 115: 1) Why do you put brackets around everything, e.g !loaded_from_data_store ? also you converted (desc.maxBlockId > 0) to (desc.maxBlockId ) which is more of a C style writing, and reduce readibility. It is less clearer in my opinion (you must relay on naming to understand the type, if we write all code like that it might force writer to go and check type, and even cause bugs).
What if i decide to change the type of x to int64_t and we have 'if (x)' on 1000 places (leave a side automatic replace). which ways of coding is better?

2) The above is not accurate.
You relay on the fact that we always have a non-pruned blockchain.
What if the same issue happens in a pruned blockchain?
You should get the genesis block id outside the condition:
```
auto genesis_block_id = as_->getGenesisBlockNum();
if (!loaded_from_data_store && ((desc.maxBlockId > 0) || (last_reachable_block_num > genesis_block_id))) {
...
```

bftengine/src/bcstatetransfer/RVBManager.cpp
```diff
@@ -95,15 +107,21 @@ void RVBManager::init(bool fetching) {
...
     DurationTracker<std::chrono::milliseconds> reconstruct_dt("reconstruct_dt", true);
     auto genesis_block_id = as_->getGenesisBlockNum();
     LOG_INFO(logger_, "Reconstructing RVB data" << KVLOG(loaded_from_data_store, desc.maxBlockId, genesis_block_id));
-    if (genesis_block_id <= desc.maxBlockId) {
-      num_rvbs_added =
-          addRvbDataOnBlockRange(genesis_block_id, desc.maxBlockId, std::optional<Digest>(desc.digestOfMaxBlockId));
+    if ((genesis_block_id <= desc.maxBlockId) || (desc.maxBlockId <= last_reachable_block_num)) {
+      num_rvbs_added = addRvbDataOnBlockRange(genesis_block_id,
+                                              std::max(desc.maxBlockId, last_reachable_block_num),
+                                              std::optional<Digest>(desc.digestOfMaxBlockId));
```
glevkovich, line 124: I'm worried about the distributed correctness of all the above, in rare cases (and we are now dealing with rare cases).
You should give priority always to maxBlockId since it relays on a checkpoint and the checkpoint should be the same for all replicas, hence all init and have the same tree.
constructing from last_reachable_block_num is very tricky, we might get for some time a non-identical RVT, if there was some human error or things we can think about now. We are building a special tree here, not relying on checkpoints. This is very sensitive if the network is standing still immediately after cloning.
If and ONLY if desc.maxBlockId is 0, then you should build the tree according to last_reachable_block_num.
What you did here is a combination of both, by getting the max between both ( last_reachable_block_num will be larger if both are non 0).
I might e wrong here, need to think more..  Let's discuss this offline.

tests/apollo/test_skvbc_persistence.py
```diff
@@ -120,6 +120,44 @@ async def test_view_change_transitions_safely_without_quorum(self, bft_network,
                             # success!
                             nursery.cancel_scope.cancel()
 
+    @with_trio
```
glevkovich, line 123: nice test!

tests/apollo/util/bft.py
```diff
@@ -1834,9 +1834,30 @@ def reset_metadata(self, replica_id):
...
+        with log.start_action(action_type="remove_metadata", replica_id=replica_id) as action:
+            db_editor_path = os.path.join(self.builddir, "kvbc", "tools", "db_editor", "kv_blockchain_db_editor")
+            db_dir = os.path.join(self.testdir, DB_FILE_PREFIX + str(replica_id))
+            remove_md_cmd = [db_editor_path,
+                            db_dir,
+                            "removeMetadata",
+                            str(replica_id),
+                            str(self.config.f),
+                            str(self.config.c),
+                            "37",  # number of principles, defined in setup.cpp
```
glevkovich, line 1853: Do we have a way importing and not mentioning explciit values? This is less maintainable...the 10 and 37.

glevkovich (changes requested):

tests/apollo/util/bft.py
mdarade: Thanks. Removed these 2 as they were not necessary.

bftengine/src/bcstatetransfer/RVBManager.cpp
mdarade: Thanks. Addressed.

bftengine/src/bcstatetransfer/RVBManager.cpp
mdarade: Thanks. Addressed.

bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
```diff
@@ -2712,6 +2713,78 @@ TEST_F(BcStTest, bkpValidateRvbDataInitialSource) {
   testConfig_.productDbDeleteOnEnd = true;
 }
 
+// Validate differenent combinations of scenarios in which RVT construction will happen while replica coming up
```
glevkovich, line 2716: great test!

bftengine/tests/bcstatetransfer/bcstatetransfer_tests.cpp
```diff
@@ -2712,6 +2713,78 @@ TEST_F(BcStTest, bkpValidateRvbDataInitialSource) {
...
+    }
+    ASSERT_NFF(dstRestart(false, FetchingState::NotFetching));
+  }
+
+  auto rvbm = stDelegator_->getRvbManager();
+  if (erase_metadata || do_not_checkpoint) {
+    ASSERT_EQ(rvbm->getRvbDataSource(), RVBManager::RvbDataInitialSource::FROM_STORAGE_RECONSTRUCTION);
+  } else {
+    ASSERT_EQ(rvbm->getRvbDataSource(), RVBManager::RvbDataInitialSource::FROM_STORAGE_CP);
+  }
```
glevkovich, line 2775: nit: consider adding her a check that tree is not empty as well.

glevkovich (approved):
Thanks @mdarade , good tests...
I've added 1 small comment which I'm not sure is really needed..

PR #2772, opened 2022-08-08
Thin replica bugs dev v1.7

glevkovich:
* **Problem Overview**  
This PR aims to improve visibility and debuggability in thin replica client-server and some parts of client service (e.g events service). This is done by fixing, adding improving logs, metrics, and histograms.
In addition, minimal (opportunistic) refactoring is done.

Commits:

1. Thin-Replica, event_service: add additional logs to improve debugging
2. thin_replica_client,event_service,concord_client: rename, add and modify metrics
3. concord_client: fix wrong variable name
4. thin_replica_impl: fix bug in ThinReplicaImpl::sendHash
5. clientservice, trc: refactor (shorten) use of std::chrono members
6. event_service: refactor metrics, and had histograms (snapshots)

* **Testing Done**  
  Maestro manual test to check for stability, and that all logs apear as expected.

### DIFF
--- a/client/clientservice/include/client/clientservice/event_service.hpp
+++ b/client/clientservice/include/client/clientservice/event_service.hpp
@@ -37,6 +37,8 @@ struct EventServiceMetrics {
  private:
   concordMetrics::Component metrics_component_;
 
+  // TODO - GILL - update_processing_duration and write_duration should be removed later, as histogram/snapshots are
+  // preferable
  public:
   // total number of updates written to the gRPC stream
   concordMetrics::CounterHandle total_num_writes;
@@ -45,6 +47,26 @@ struct EventServiceMetrics {
   concordMetrics::GaugeHandle update_processing_duration;
   // time taken to write an update to the gRPC stream
   concordMetrics::GaugeHandle write_duration;
+};
+
+struct Recorders {
+  static constexpr uint64_t MAX_VALUE_MICROSECONDS = 60ULL * 1000ULL * 1000ULL;  // 60 seconds
+
+  Recorders() {
+    auto& registrar = concord::diagnostics::RegistrarSingleton::getInstance();
+    if (!registrar.perf.isRegisteredComponent("clientservice_event_service")) {
+      registrar.perf.registerComponent("clientservice_event_service", {processing_duration, write_duration});
+    }
+  }
+  ~Recorders() {
+    auto& registrar = concord::diagnostics::RegistrarSingleton::getInstance();
+    if (registrar.perf.isRegisteredComponent("clientservice_event_service")) {
+      registrar.perf.unRegisterComponent("clientservice_event_service");
+    }
+  }
+
+  DEFINE_SHARED_RECORDER(processing_duration, 1, MAX_VALUE_MICROSECONDS, 3, concord::diagnostics::Unit::MICROSECONDS);
+  DEFINE_SHARED_RECORDER(write_duration, 1, MAX_VALUE_MICROSECONDS, 3, concord::diagnostics::Unit::MICROSECONDS);
 };
 
 class EventServiceImpl final : public vmware::concord::client::event::v1::EventService::Service {
@@ -62,6 +84,7 @@ class EventServiceImpl final : public vmware::concord::client::event::v1::EventS
   logging::Logger logger_;
   std::shared_ptr<concord::client::concordclient::ConcordClient> client_;
   EventServiceMetrics metrics_;
+  Recorders histograms_;
 };
 
 }  // namespace concord::client::clientservice
--- a/client/clientservice/include/client/clientservice/event_service.hpp
+++ b/client/clientservice/include/client/clientservice/event_service.hpp
@@ -23,8 +23,8 @@ struct EventServiceMetrics {
   EventServiceMetrics()
       : metrics_component_{"EventService", std::make_shared<concordMetrics::Aggregator>()},
         total_num_writes{metrics_component_.RegisterCounter("total_num_writes", 0)},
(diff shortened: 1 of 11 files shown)


client/clientservice/include/client/clientservice/event_service.hpp
```diff
@@ -37,6 +37,8 @@ struct EventServiceMetrics {
  private:
   concordMetrics::Component metrics_component_;
 
+  // TODO - GILL - update_processing_duration and write_duration should be removed later, as histogram/snapshots are
```
mdarade, line 40: Nit: Name needs to be part of TODO?

client/clientservice/src/event_service.cpp
```diff
@@ -47,6 +50,8 @@ Status EventServiceImpl::Subscribe(ServerContext* context,
                                    ServerWriter<SubscribeResponse>* stream) {
   LOG_INFO(logger_, "");
   cc::SubscribeRequest request;
+  static const auto historgram_log_duration_microsec{std::chrono::microseconds(300s).count()};
```
mdarade, line 53: Are these options going to have configurable values in near future?
Is TODO note required here?

client/clientservice/src/event_service.cpp
```diff
@@ -101,64 +107,69 @@ Status EventServiceImpl::Subscribe(ServerContext* context,
...
+        for (auto& event : event_group_in.events) {
+          proto_event_group.add_events(std::move(event));
+        }
+        *proto_event_group.mutable_record_time() = event_group_in.record_time;
+        *proto_event_group.mutable_trace_context() = {event_group_in.trace_context.begin(),
+                                                      event_group_in.trace_context.end()};
+        *response.mutable_event_group() = proto_event_group;
+        DurationTracker<chrono::microseconds> write_dt("write_dt", true);
+        stream->Write(response);
+        metrics_.write_duration.Get().Set(write_dt.totalDuration());
```
mdarade, line 127: On similar lines to legacy event update, do we need LOG_DEBUG statement here?

client/concordclient/include/client/concordclient/concord_client.hpp
```diff
@@ -187,7 +187,7 @@ class ConcordClient {
 
   logging::Logger logger_;
   const ConcordClientConfig& config_;
-  std::shared_ptr<concordMetrics::Aggregator> metrics_;
+  std::shared_ptr<concordMetrics::Aggregator> aggregator_;
```
mdarade, line 190: Nit: Would metrics_aggregator_ be very specific name?

client/thin-replica-client/src/thin_replica_client.cpp
```diff
@@ -957,6 +958,9 @@ void ThinReplicaClient::Subscribe() {
     state.pop_front();
   }
   latest_verified_block_id_ = block_id;
+  metrics_.last_verified_block_id.Get().Set(latest_verified_block_id_);
+  metrics_.current_queue_size.Get().Set(config_->update_queue->size());
```
mdarade, line 965: Would operator= from Gauge work here?

client/clientservice/src/event_service.cpp
```diff
@@ -40,6 +40,7 @@ namespace concord::client::clientservice {
 Status EventServiceImpl::Subscribe(ServerContext* context,
                                    const SubscribeRequest* proto_request,
                                    ServerWriter<SubscribeResponse>* stream) {
+  LOG_INFO(logger_, "");
```
mdarade, line 51: Should anything from input params be logged here?

client/clientservice/include/client/clientservice/event_service.hpp
glevkovich: It's really a matter of style.
Actually, I put it here since I'm not the main owner, I usually don't put if I wrote/modified most of the module.

client/clientservice/src/event_service.cpp
glevkovich: 1) I don't see why we would like to configure this from the outside.
This is a hard-coded configurable value, and currently, things still don't work well (pending task).
It put it here and moved it from the middle of the code to give better visibility, so this is probably the most we should do to configure this.

2) I'm going to move them to be class members, even though we have a single subscribe by design, it is not a good thing to keep it static if by any chance we will have more than a single subscribe in the future.

client/clientservice/src/event_service.cpp
glevkovich: I'll update here and the other place with LOG_DEBUG + the duration.
Thanks.

client/concordclient/include/client/concordclient/concord_client.hpp
glevkovich: 10-20% of aggregator in our code are called <some_name>_aggregator.
but mostly we just used the name aggregator_.
It's more like a convention in our code.
Anyway, the name metrics_ was a total mistake.
We can use the name concord_client_aggregator_
but I don't see how it helps.

client/thin-replica-client/src/thin_replica_client.cpp
glevkovich: We overload operator= for another gauge, but here I try to set a value uint64_t and we don't overload it.

client/clientservice/src/event_service.cpp
glevkovich: context is pretty opaque, used to call APIs (RPC type).
The response is not yet filled.
proto_request holds a variant, so maybe can add info later but not in this line, it depends what mode we work with.
I'll see if I can add some logs later on the if/else.
Thanks.

eyalrund (approved):

PR #2774, opened 2022-08-09
State Transfer : Construct RVT from storage in absence of metadata

mdarade:
During upgrade, when metadata is explicitly removed, RVT should
still be able to get constructed from storage.

* **Problem Overview**  
As part of upgrade, metadata is cleaned up before spinning 1.7 blockchain with existing RocksDB data. Whenever replicas gets started, they try to build RVT either from last stored checkpoint descriptors from metadata or from RocksDB storage. Existing checks had assumption that metadata would always be present (to refer to desc.maxBlockId, desc.rvbData) and there was neither Apollo nor unit test present to check such scenario. Failure at not building RVT during boot up would end up building it at first checkpoint. This would result in replicas taking considerable time (based on number of blocks in RocksDB) to commit very first checkpoint after upgrade which can also subsequently end up triggering client request timeouts.

* **Testing Done**  
Added Apollo test to reproduce issue and validated fix.
Added GTest to reproduce issue and validated fix.

### DIFF
--- a/bftengine/src/bcstatetransfer/RVBManager.cpp
+++ b/bftengine/src/bcstatetransfer/RVBManager.cpp
@@ -46,14 +46,18 @@ RVBManager::RVBManager(const Config& config, const IAppState* state_api, const s
 }
 
 void RVBManager::init(bool fetching) {
-  LOG_TRACE(logger_, "");
-  bool loaded_from_data_store = false;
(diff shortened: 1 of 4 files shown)


glevkovich (approved):
LGTM

nityavasudevan (approved):

PR #2784, opened 2022-08-29
Upgrade gRPC version to 1.48.0

shgandhi:
Over the past ten versions of gRPC multiple optimizations and bug fixes have been introduced. Please see [gRPC release page](https://github.com/grpc/grpc/releases) for the detailed breakdown of updates added from 1.38.0 to 1.48.0.
This upgrade would also allow for batching of updates natively via setting the buffer hint flag for improved performance.
    
This PR updates install_deps.sh, client and TRS CMakeLists files, and adds a few support cmake files to account for the upgrade.

Concord was built manually to test the change. Performance testing is being done in parallel based off of the MR.


client/proto/CMakeLists.txt
```diff
@@ -1,21 +1,105 @@
...
+get_filename_component(ccr_proto "../concordclient/proto/concord_client_request/v1/concord_client_request.proto" ABSOLUTE)
+get_filename_component(rs_proto_path "${rs_proto}" PATH)
+get_filename_component(es_proto_path "${es_proto}" PATH)
+get_filename_component(ss_proto_path "${ss_proto}" PATH)
+get_filename_component(ccr_proto_path "${ccr_proto}" PATH)
+
+get_target_property(grpc_cpp_plugin_location gRPC::grpc_cpp_plugin LOCATION)
+
+# Generated sources
+set(rs_proto_srcs "${CMAKE_CURRENT_BINARY_DIR}/request.pb.cc")
```
vrancurel, line 22: Why can't we use `grpc_generate_cpp()` any more ?

shgandhi: @vrancurel the [FindProtobuf.cmake](https://github.com/Kitware/CMake/blob/master/Modules/FindProtobuf.cmake#L103) file specifies that [protobuf_generate_cpp works properly ](https://github.com/Kitware/CMake/blob/01e95efc34a2a9bcf9910d65d3b38ff9b8a51750/Modules/FindProtobuf.cmake#L103)within the same directory. I am not entirely sure what it means, but with our directory structure, I wasn’t able to compile the proto files using the functions `protobuf_generate_cpp` and `grpc_generate_cpp`.
Currently all the examples in the grpc cpp repo, also compile the proto files using the cmake custom command (for e.g., [this](https://github.com/grpc/grpc/blob/master/examples/cpp/helloworld/CMakeLists.txt))

arc-vmware (approved):
LGTM !!!

cmake/FindPackageHandleStandardArgs.cmake
```diff
@@ -0,0 +1,605 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
toly-kournik, line 1: @shgandhi, this file is a part of a standard cmake installation, there's no need to put it here, also it may change when upgrading cmake version

cmake/FindProtobuf.cmake
```diff
@@ -1,126 +1,685 @@
...
-# Locate and configure the Google Protocol Buffers library
-#
-# Adds the following targets:
-#
-#  protobuf::libprotobuf - Protobuf library
-#  protobuf::libprotobuf-lite - Protobuf lite library
-#  protobuf::libprotoc - Protobuf Protoc Library
-#  protobuf::protoc - protoc executable
-#
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
toly-kournik, line 1: This file is installed together with a protobuf library under `/usr/share/cmake-3.20/Modules/FindProtobuf.cmake`
No reason to have it in our source tree

toly-kournik (changes requested):
@shgandhi , there are `*.cmake` files which shouldn't be installed as part of our source tree as they are either part of a cmake or of a specific library installation

shgandhi:
> @shgandhi , there are `*.cmake` files which shouldn't be installed as part of our source tree as they are either part of a cmake or of a specific library installation

Thanks for the review @toly-kournik, I have removed the redundant files in the latest rebase.

shgandhi:
Thanks for the review @vrancurel, @toly-kournik and @arc-vmware.

install_deps.sh
```diff
@@ -300,36 +300,20 @@ install_openssl() {
...
-            -DgRPC_SSL_PROVIDER=package \
-            -DBUILD_SHARED_LIBS=ON \
-            -DCMAKE_BUILD_TYPE=Release \
-            -DCMAKE_INSTALL_PREFIX=/usr/local \
-            ../.. && \
-        make -j$(nproc) install &&
-        cd ${HOME} && \
-        rm -r ${HOME}/grpc
-
+git clone -b v1.48.x --depth 1 --recurse-submodules https://github.com/grpc/grpc && \
```
toly-kournik, line 303: backslashes are redundant here, since it's a function

cmake/common.cmake
```diff
@@ -0,0 +1,104 @@
+cmake_minimum_required(VERSION 3.5)
```
toly-kournik, line 1: What's the purpose of this file in our case? Where and how it's used?

install_deps.sh
shgandhi: @toly-kournik all the functions in this file use backslashes, I believe this is for readability.

cmake/common.cmake
shgandhi: It defines variables which are used by TRS and clientservice CMakeLists files.

install_deps.sh
toly-kournik: It less readable and it was just copied as is when converted to functions :-(

shgandhi: @toly-kournik I have updated the install_deps file to remove redundant backslashes.

toly-kournik (approved):
LGTM!

PR #2785, opened 2022-08-30
Fix a bug in build db checkpoint metadata from persistence

nkumar04:
* **Problem Overview**  
  When a reconfiguration request like scale is performed, metadata is
erased as a result db checkpoint metdata is also removed. Hence, on restart
we rebuild the metadata from the file system. In this PR, we fix a bug
where latest db checkpoint is set in-correctly on building dbcheckpoint
metadata from filesystem
* **Testing Done**  
  Github CI Test


arc-vmware (approved):
LGTM!!!

PR #2789, opened 2022-09-02
Replace thread_safe_queue with spsc_queue b/w TRC and Event Service

shgandhi:
The queue b/w TRC and event service is expected to have an update at most times. Additionally event service currently doesn't allow concurrent subscription streams from the same subscriber. Hence, currently we do not expect contention for reading from the queue.
Thus, lockfree queue is a more performant alternative to thread_safe_queue which uses locks.
Note the queue size is set to 10000 by default. Without event groups enabled we haven't see > 20 updates per second in the TRC queue, and with event groups we haven't > 1000 updates per second in the TRC queue, hence, 10k should be a sufficiently large queue size.
    
For testing, unit tests have been updated to reflect the change in queue type. Performance testing is being done in parallel.

### DIFF
--- a/client/concordclient/include/client/concordclient/trc_queue.hpp
+++ b/client/concordclient/include/client/concordclient/trc_queue.hpp
@@ -1,0 +1,112 @@
+// Concord
+//
+// Copyright (c) 2020-2022 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License").
+// You may not use this product except in compliance with the Apache 2.0
+// License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include <condition_variable>
+#include <boost/lockfree/spsc_queue.hpp>
+#include <list>
+#include <memory>
+#include <thread>
+#include <mutex>
+#include <exception>
+#include <chrono>
+#include <variant>
+
+#include "assertUtils.hpp"
+#include "event_update.hpp"
+
+namespace concord::client::concordclient {
+
+typedef std::variant<Update, EventGroup> EventVariant;
+
+class TrcQueue {
+ private:
+  boost::lockfree::spsc_queue<EventVariant*> queue_data_;
+  std::exception_ptr exception_;
+  std::mutex mutex_;
+  std::condition_variable condition_;
+  static constexpr size_t kQueueDataSize{10000u};
+
+ public:
+  // Construct a TrcQueue.
+  TrcQueue() : queue_data_(kQueueDataSize) {}
+
+  // Copying or moving a TrcQueue is explicitly disallowed, as we do not know of a compelling use case
+  // requiring copying or moving TrcQueue, we believe semantics for these operations are likely to be
+  // messy in some cases, and we believe implementation may be non-trivial. We may revisit the decision to disallow
+  // these operations should compelling use cases for them be found in the future.
+  TrcQueue(const TrcQueue& other) = delete;
+  TrcQueue(const TrcQueue&& other) = delete;
+  TrcQueue& operator=(const TrcQueue& other) = delete;
+  TrcQueue& operator=(const TrcQueue&& other) = delete;
+
+  void clear() { queue_data_.reset(); }
+
+  void push(EventVariant* update) {
+    queue_data_.push(std::move(update));
+    condition_.notify_one();
+  }
+
+  std::unique_ptr<EventVariant> getElementFromQueue() {
+    if (queue_data_.read_available()) {
+      EventVariant* update = new EventVariant;
+      queue_data_.pop(update);
+      return std::unique_ptr<EventVariant>(update);
+    }
+    if (exception_) {
+      auto e = exception_;
+      exception_ = nullptr;
+      std::rethrow_exception(e);
+    }
+    return nullptr;
+  }
+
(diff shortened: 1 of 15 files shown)


client/concordclient/include/client/concordclient/trc_queue.hpp
```diff
@@ -0,0 +1,106 @@
...
+      }
+      return getElementFromQueue();
+    }
+  }
+
+  std::unique_ptr<EventVariant> tryPop() { return getElementFromQueue(); }
+
+  uint64_t size() { return queue_data_.read_available(); }
+
+  void setException(std::exception_ptr e) { exception_ = e; }
```
bsenthilr, line 103: this is not thread safe

shgandhi: Thanks for catching this @bsenthilr. I have added a fix.

bsenthilr: updates looks good

client/concordclient/include/client/concordclient/trc_queue.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include <variant>
+
+#include "assertUtils.hpp"
+#include "event_update.hpp"
+
+namespace concord::client::concordclient {
+
+typedef std::variant<Update, EventGroup> EventVariant;
+
+class TrcQueue {
```
arc-vmware, line 33: Can we implement this as :
template \<typename T\>
class TrcQueue : public IQueue\<T\> 

So TrcQueue will be another implementation of IQueue

client/concordclient/include/client/concordclient/trc_queue.hpp
```diff
@@ -0,0 +1,112 @@
...
+ private:
+  boost::lockfree::spsc_queue<EventVariant*> queue_data_;
+  std::exception_ptr exception_;
+  std::mutex mutex_;
```
arc-vmware, line 37: if we are using the lockfree queue, then we should remove these mutex and condition variable.

client/concordclient/include/client/concordclient/trc_queue.hpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+    if (exception_) {
+      auto e = exception_;
+      exception_ = nullptr;
+      std::rethrow_exception(e);
+    }
+    return nullptr;
+  }
+
+  std::unique_ptr<EventVariant> popTill(std::chrono::milliseconds timeout) {
```
arc-vmware, line 75: Why do we need this method for TRC queue?

arc-vmware (commented):
Changes are very good. Few comments from my side.

client/concordclient/include/client/concordclient/trc_queue.hpp
shgandhi: @arc-vmware if `TrcQueue` derives from `IQueue`, there will be a type mismatch b/w the `push` method of TrcQueue and IQueue, as the latter requires `std::unique_ptr<T>` as an argument, whereas with spsc queue we can only use raw pointers. 
This is because spsc_queue does not use move semantics (methods to push onto the queue take parameters as const reference and construct a copy in the queue's internal storage), so i can not push a unique_ptr<T> onto the TrcQueue, i.e., why I used a separate implementation.

client/concordclient/include/client/concordclient/trc_queue.hpp
shgandhi: We use mutex and condition variable when popping from the queue if the queue is empty, and in `setException`.

client/concordclient/include/client/concordclient/trc_queue.hpp
shgandhi: Because event service uses popTill ([ref](https://github.com/vmware/concord-bft/blob/d78090cd03393b8ed3db4a515c1c5f9d961bb692/client/clientservice/src/event_service.cpp#L84)).

client/concordclient/include/client/concordclient/trc_queue.hpp
arc-vmware: void push(EventVariant* update) {
    queue_data_.push(std::move(update));
    condition_.notify_one();
  }

Here we can see that we use move semantics. I think it will be better if we refactor and use smart ptrs instead of raw ptr. Also, we can pass the ptr inside a unique_ptr as below:
auto update = std::unique_ptr<EventVariant>{raw_ptr_update};
remote_queue->push(std::move(update));
Internally the queue will use the unique_ptr. I think enforcing the queue to use a smart ptr (unique_ptr in this case) is a better design. It will be good if we give it a thought and try it. If it doesn't work we can do things as it is.

client/concordclient/include/client/concordclient/trc_queue.hpp
arc-vmware: ok, so the mutex is for exception_ variable. thanks.

client/concordclient/include/client/concordclient/trc_queue.hpp
shgandhi: I agree, but boost::lockfree::spsc_queue doesn't allow it. Please see https://www.boost.org/doc/libs/1_59_0/doc/html/boost/lockfree/spsc_queue.html, this page mentions boost::lockfree::spsc_queue of type T has the following requirements -
- T must have a default constructor
- T must be copyable
A unique_ptr type is not copyable.

arc-vmware (approved):
LGTM !!!

shgandhi:
Thanks for the review @arc-vmware and @bsenthilr!

PR #2791, opened 2022-09-05
Introducing UTT's signature processor

yontyon:
In this PR we introduce the basic implementation of the UTT signature processor component.
The signature processor is responsible for computing the partial signature, collecting others' partial signatures, assembling the full signature, and publishing it back through a consensus to the system.

In this implementation, we introduce a leader-based approach that relies on the bft consensus for correctness and safety.
Also, the signature processor is thread-safe such that the entry point method can be called by any thread.

Finally, we also added some simple unittests that demonstrate how to use the signature processor


utt/signature-processor/include/sigData.hpp
```diff
@@ -0,0 +1,68 @@
...
+namespace utt {
+
+struct SigJobEntry {
+  uint64_t job_id;
+  uint64_t start_time{0};
+  uint64_t job_timeout_ms_{1000};
+  std::map<uint32_t, libutt::api::types::Signature> partial_sigs;
+};
+
+class SigJobEntriesMap {
```
evdzhurov, line 28: I find the interface of this class a bit confusing.
* We call "registerEntry" and "put" with a SigJobEntry to actually create an entry? Can we create a valid entry if we call just one of those methods?
* We have an option to call "get" or "getEntries". The first returns a non-const reference, the second returns a copy of the internal map. Supposedly this hints that the "jobs" themselves are the entries, but when we call "registerEntry" we actually add a validation callback.

utt/signature-processor/include/sigProcessor.hpp
```diff
@@ -0,0 +1,181 @@
...
+   * @brief The sigProcessor responsibility is to compute the partial UTT signature, collecting and combining the full
+   * signature.
+   *
+   */
+ public:
+  /**
+   * @brief JobCoordinatorCb is a callback for choosing the coordinator for a given signature job. The only requirement
+   * is that it has to be deterministic among all replicas The default implementation is simply job_id % n
+   */
+  typedef std::function<uint16_t(uint64_t)> JobCoordinatorCb;
```
evdzhurov, line 37: Consider "using" instead of typedef.

utt/signature-processor/include/sigProcessor.hpp
```diff
@@ -0,0 +1,181 @@
...
+   * @param signer the UTT Registrator object
+   * @param job_timeout timeout for this job. If the timeout passes, the replica will broadcast the partial signature to
+   * all. default value is 1000 ms.
+   */
+  void processRCMSignature(uint64_t sigId,
+                           const libutt::api::types::CurvePoint& pidHash,
+                           const libutt::api::Commitment& rcm1,
+                           const libutt::api::types::CurvePoint& s2,
+                           std::shared_ptr<libutt::api::Registrator> signer,
+                           uint64_t job_timeout = 1000);
```
evdzhurov, line 85: job_timeout lacks time resolution, consider "job_timeout_ms" for example.

utt/signature-processor/include/sigProcessor.hpp
```diff
@@ -0,0 +1,181 @@
...
+   *
+   * @tparam T one of <Budget, Mint, Transaction>
+   * @param sigIds a vector of identifiers for the signature computing jobs
+   * @param data the UTT object which we want to sign of (one of <Budget, Mint, Transaction>)
+   * @param signer the UTT's coinsSigner object
+   * @param job_timeout a timeout for this job(s). If the timeout passes, the replica will broadcast the partial
+   * signature to all. The default is 1000 ms
+   */
+  template <typename T>
+  void processSignature(std::vector<uint64_t> sigIds,
```
evdzhurov, line 115: Do we need to copy sigIds?

utt/signature-processor/include/sigProcessor.hpp
```diff
@@ -0,0 +1,181 @@
...
+  concordUtil::Timers& timers_;
+  uint16_t repId_;
+  uint16_t n_;
+  uint16_t threshold_;
+  uint64_t timer_handler_timeout_;
+  JobCoordinatorCb getJobCoordinator = [&](uint64_t sigId) { return sigId % n_; };
+  GenerateAppClientRequestCb generateAppClientRequest = [](uint64_t sig_id, const libutt::api::types::Signature& sig) {
+    std::vector<uint8_t> ret(sizeof(uint64_t) + sig.size(), 0);
+    std::memcpy(ret.data(), &sig_id, sizeof(uint64_t));
+    std::memcpy(ret.data() + sizeof(uint64_t), sig.data(), sig.size());
```
evdzhurov, line 175: Probably need to consider endianness here for the multibyte values

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    for (auto& [id, psig] : jobEntry.partial_sigs) {
+      if (!validator((uint16_t)id, psig)) invalids.push_back((uint16_t)id);
+    }
+    for (auto id : invalids) jobEntry.partial_sigs.erase((uint32_t)id);
+
+    jobEntry.partial_sigs.emplace((uint32_t)repId_, partial_sig);
+    jobEntry.job_id = sigId;
+    jobEntry.start_time = std::chrono::steady_clock::now().time_since_epoch().count();
+    jobEntry.job_timeout_ms_ = job_timeout;
+    jobs_.put(sigId, jobEntry);
```
evdzhurov, line 140: What's the point of creating and modifying a reference of a JobEntry with the "get" method and then completely overwriting this entry with a new one by calling "put"?

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    auto job_coordinator = getJobCoordinator(sigId);
+    uint16_t rep = repId_;
```
evdzhurov, line 142: Why do we copy repId_?

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                            const libutt::api::types::Signature& partial_sig) {
+  SigJobEntry& entry = jobs_.get(sigId);
+  if (entry.partial_sigs.size() == threshold_) return;
+  if (jobs_.isRegistered(sigId)) {
+    auto& validator = jobs_.getJobValidator(sigId);
+    if (!validator(sig_source, partial_sig)) return;
+  }
+  entry.partial_sigs.emplace((uint32_t)sig_source, partial_sig);
+  if (entry.partial_sigs.size() == threshold_) {
+    auto sig = libutt::api::Utils::aggregateSigShares(n_, entry.partial_sigs);
```
evdzhurov, line 159: This aggregation can fail if the sig shares have different bases.

evdzhurov (commented):
As we're moving forward, the utt branch will not be used for development anymore. This PR will need to be resubmitted relative to "master" after we're done merging the utt branch.

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                       const libutt::api::types::CurvePoint& s2,
+                                       std::shared_ptr<libutt::api::Registrator> signer,
+                                       uint64_t job_timeout) {
+  jobs_.registerEntry(sigId, [signer, rcm1, s2, pidHash](uint64_t rid, const libutt::api::types::Signature& sig) {
+    (void)sig;
+    (void)rid;
+    return true;
+  });
+  auto partial_sig = signer->signRCM(pidHash, s2, rcm1);
+  std::vector<uint64_t> sigIds = {sigId};
```
cloudnoize, line 117: nit: redundant line, you can do the same in the call to handleSigJob saving the creation of an l value

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                   }
+                                 });
+}
+void SigProcessor::processRCMSignature(uint64_t sigId,
+                                       const libutt::api::types::CurvePoint& pidHash,
+                                       const libutt::api::Commitment& rcm1,
+                                       const libutt::api::types::CurvePoint& s2,
+                                       std::shared_ptr<libutt::api::Registrator> signer,
+                                       uint64_t job_timeout) {
+  jobs_.registerEntry(sigId, [signer, rcm1, s2, pidHash](uint64_t rid, const libutt::api::types::Signature& sig) {
```
cloudnoize, line 111: why do you set a callback per signature and not have a map of callback per signature type that you set once e.g. sigTypeRCM -> validateRCM  etc

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    uint16_t rep = repId_;
+    messages::PartialSigMsg msg(rep, partial_sig, sigId);
+    msgsCommunicator_->sendAsyncMessage((uint64_t)job_coordinator, msg.body(), msg.size());
+  }
+}
+SigProcessor::~SigProcessor() { timers_.cancel(timeout_handler_); }
+void SigProcessor::onReceivingNewPartialSig(uint64_t sigId,
+                                            uint16_t sig_source,
+                                            const libutt::api::types::Signature& partial_sig) {
+  SigJobEntry& entry = jobs_.get(sigId);
```
cloudnoize, line 151: This get can create a new entry as it uses the [] operator on the map.
Is it intended? if so maybe it's worth implying if in the method name

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    auto job_coordinator = getJobCoordinator(sigId);
+    uint16_t rep = repId_;
+    messages::PartialSigMsg msg(rep, partial_sig, sigId);
+    msgsCommunicator_->sendAsyncMessage((uint64_t)job_coordinator, msg.body(), msg.size());
+  }
+}
+SigProcessor::~SigProcessor() { timers_.cancel(timeout_handler_); }
+void SigProcessor::onReceivingNewPartialSig(uint64_t sigId,
+                                            uint16_t sig_source,
+                                            const libutt::api::types::Signature& partial_sig) {
```
cloudnoize, line 150: Worth adding assert for sigID%cluster_size == rep_id

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                uint64_t job_timeout) {
+  if (sigIds.size() != partial_sigs.size()) return;
+  for (size_t i = 0; i < sigIds.size(); i++) {
+    auto sigId = sigIds.at(i);
+    auto& partial_sig = partial_sigs.at(i);
+    auto& jobEntry = jobs_.get(sigId);
+    auto& validator = jobs_.getJobValidator(sigId);
+    std::vector<uint16_t> invalids;
+    for (auto& [id, psig] : jobEntry.partial_sigs) {
+      if (!validator((uint16_t)id, psig)) invalids.push_back((uint16_t)id);
```
cloudnoize, line 132: Why do you need here to validate the signatures as it's being done on the on message?
what is the scenario that can lead to stored invalid msgs?

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+SigProcessor::~SigProcessor() { timers_.cancel(timeout_handler_); }
+void SigProcessor::onReceivingNewPartialSig(uint64_t sigId,
+                                            uint16_t sig_source,
+                                            const libutt::api::types::Signature& partial_sig) {
+  SigJobEntry& entry = jobs_.get(sigId);
+  if (entry.partial_sigs.size() == threshold_) return;
+  if (jobs_.isRegistered(sigId)) {
+    auto& validator = jobs_.getJobValidator(sigId);
+    if (!validator(sig_source, partial_sig)) return;
```
cloudnoize, line 155: log msg?

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                                     0x0,
+                                                     requestSeqNum,
+                                                     (uint32_t)appClientReq.size(),
+                                                     (const char*)appClientReq.data(),
+                                                     60000,
+                                                     "new-utt-sig-" + std::to_string(sigId));
+    msgsCommunicator_->getIncomingMsgsStorage()->pushExternalMsg(std::unique_ptr<MessageBase>(crm));
+  }
+}
+void SigProcessor::onReceivingNewValidFullSig(uint64_t sigId) { jobs_.erase(sigId); }
```
cloudnoize, line 173: What calls this and when ?

utt/signature-processor/tests/sigProcessorTests.cpp
```diff
@@ -0,0 +1,468 @@
...
+      delete msg;
+    });
+    sig_processors[i] =
+        std::make_shared<utt::SigProcessor>(i, n, t, 1000, msc, msr, ((IncomingMsgsStorageImp*)(ims.get()))->timers());
+    msc->startCommunication(i);
+    msc->startMsgsProcessing(i);
+  }
+
+  LOG_INFO(logger, "register clients...");
+  uint64_t jid_{0};
```
cloudnoize, line 224: nit: why ```_``` postfix ?

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    (void)sig;
+    (void)rid;
+    return true;
+  });
+  auto partial_sig = signer->signRCM(pidHash, s2, rcm1);
+  std::vector<uint64_t> sigIds = {sigId};
+  handleSigJob(sigIds, {partial_sig.second}, job_timeout);
+}
+
+void SigProcessor::handleSigJob(std::vector<uint64_t> sigIds,
```
cloudnoize, line 121: Why is it accepting ```std::vector``` and not a single id:sig ?

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                const std::vector<libutt::api::types::Signature>& partial_sigs,
+                                uint64_t job_timeout) {
+  if (sigIds.size() != partial_sigs.size()) return;
+  for (size_t i = 0; i < sigIds.size(); i++) {
+    auto sigId = sigIds.at(i);
+    auto& partial_sig = partial_sigs.at(i);
+    auto& jobEntry = jobs_.get(sigId);
+    auto& validator = jobs_.getJobValidator(sigId);
```
cloudnoize, line 129: This will create a new entry for a validator if it does not exist for sigId since the ```getJobValidator``` uses ```[]```  , it will return an empty std::function that you'll use in line 132

utt/signature-processor/src/sigProcessor.cpp
```diff
@@ -0,0 +1,181 @@
...
+      repId_{repId},
+      n_{n},
+      threshold_{threshold},
+      timer_handler_timeout_{timer_handler_timeout} {
+  msgHandlers->registerMsgHandler(bftEngine::impl::MsgCode::Reserved, [&](bftEngine::impl::MessageBase* message) {
+    messages::PartialSigMsg* msg = (messages::PartialSigMsg*)message;
+    onReceivingNewPartialSig(msg->getSigId(), msg->idOfGeneratedReplica(), msg->getPartialSig());
+    delete msg;
+  });
+  timeout_handler_ = timers_.add(std::chrono::milliseconds(timer_handler_timeout_),
```
cloudnoize, line 90: I see that in the test, the timer evaluation is called from ```IncomingMsgsStorageImp::dispatchMessages``` , is it going to be the same as in the product? if the job entry is big, can it block the msg dispatcher for a significant time?

utt/signature-processor/include/sigData.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <map>
+#include <unordered_map>
+#include <mutex>
+#include <functional>
+#include "types.hpp"
+namespace utt {
+
+struct SigJobEntry {
+  uint64_t job_id;
```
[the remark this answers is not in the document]
yontyon: fixed

utt/signature-processor/include/sigData.hpp
```diff
@@ -0,0 +1,68 @@
...
+  uint64_t start_time{0};
+  uint64_t job_timeout_ms_{1000};
```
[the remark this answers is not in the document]
yontyon: fixed

utt/signature-processor/include/sigData.hpp
```diff
@@ -0,0 +1,68 @@
...
+ public:
+  SigJobEntry& get(uint64_t id) {
+    std::unique_lock lck(lock_);
+    return jobs_[id];
+  }
+
+  void put(uint64_t id, const SigJobEntry& entry) {
+    std::unique_lock lck(lock_);
+    jobs_[id] = entry;
+  }
```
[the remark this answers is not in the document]
yontyon: fixed

utt/signature-processor/include/sigData.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+  void registerEntry(uint64_t id, std::function<bool(uint16_t, const libutt::api::types::Signature&)> vcb) {
+    std::unique_lock lck(lock_);
+    validation_callbacks_[id] = vcb;
+  }
+  const std::unordered_map<uint64_t, SigJobEntry> getEntries() {
+    std::unique_lock lck(lock_);
+    return jobs_;
+  }
+  auto& getJobValidator(uint64_t jobId) {
```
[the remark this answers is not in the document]
yontyon: fixed

utt/signature-processor/include/sigData.hpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  void registerEntry(uint64_t id, std::function<bool(uint16_t, const libutt::api::types::Signature&)> vcb) {
+    std::unique_lock lck(lock_);
+    validation_callbacks_[id] = vcb;
+  }
+  const std::unordered_map<uint64_t, SigJobEntry> getEntries() {
+    std::unique_lock lck(lock_);
+    return jobs_;
+  }
```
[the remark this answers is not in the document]
yontyon: fixed

utt/signature-processor/include/sigData.hpp
yontyon: fixed

utt/signature-processor/include/sigProcessor.hpp
yontyon: fixed

utt/signature-processor/include/sigProcessor.hpp
yontyon: fixed

utt/signature-processor/include/sigProcessor.hpp
yontyon: fixed

utt/signature-processor/include/sigProcessor.hpp
yontyon: This will be probably used only for testing

utt/signature-processor/src/sigProcessor.cpp
yontyon: fixed

utt/signature-processor/src/sigProcessor.cpp
yontyon: fixed

utt/signature-processor/tests/sigProcessorTests.cpp
```diff
@@ -0,0 +1,468 @@
...
+    ims_->pushExternalMsg(std::move(msg));
+  }
+  void onConnectionStatusChanged(NodeNum, ConnectionStatus) override {}
+  ~TestReceiver() {}
+
+ private:
+  std::shared_ptr<IncomingMsgsStorage> ims_;
+};
+
+struct GpData {
```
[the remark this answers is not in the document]
yontyon: global params

utt/signature-processor/src/sigProcessor.cpp
yontyon: fixed

utt/signature-processor/src/sigProcessor.cpp
yontyon: the validation is per the specific request data, so if we don't want the sigCollector to know about the specific UTT type, we have to pass a callback (created by the caller)

utt/signature-processor/src/sigProcessor.cpp
yontyon: fixed

utt/signature-processor/src/sigProcessor.cpp
yontyon: replicas that are not the leader also may get partial signatures from others (in case they hit the timeout)

utt/signature-processor/src/sigProcessor.cpp
yontyon: If the job leader has started the signature processing a bit later than others, he might get partial signatures that it doesn't have the data to validate.
Hence, once it has the data, it needs to validate everything it has already got

utt/signature-processor/src/sigProcessor.cpp
yontyon: we need to add log messages across all the code, I hope to get into it later (in a different PR)

utt/signature-processor/src/sigProcessor.cpp
yontyon: the user of sigProcessor call it when it gets through a consensus a message with a valid full signature (in the case of Ethereum, it will be called by the EVM on the execution step)

utt/signature-processor/tests/sigProcessorTests.cpp
yontyon: fixed

utt/signature-processor/src/sigProcessor.cpp
yontyon: fixed

utt/signature-processor/src/sigProcessor.cpp
yontyon: fixed

utt/signature-processor/src/sigProcessor.cpp
yontyon: yes, this is the stright forward way we have to implement a thread-safe timer.
You are right that if we have many ongoing signature processing this may block the message thread, but I don't think this is will be a frequent scenario. Recall that once we have done computing the signatures we evacuate the relevant entry.
Anyway, I think we can leave it as it is for now, and lets keep it in mind for the futuer

utt/signature-processor/src/sigProcessor.cpp
cloudnoize: cool ,thanks

utt/signature-processor/src/sigProcessor.cpp
cloudnoize: cool ,thanks

PR #2793, opened 2022-09-06
Updating grpc bindings for python clientservice

alfredd:
* **Problem Overview**  
The grpc stubs located under apollo clientservice directory were based
on older protocol buffers. The state_snapshot.proto bindings were
mising. The following changes were made:
1. Added code in the update_python_bindings.sh to generate python
bindings for state_snapshot.proto.
2. Added the newly generated python grpc stubs to the directory so that
they can be used by the ClientService class in csutil.py
3. Updated the ClientService class in the csutil.py file to use the grpc
bindings for state_snapshot.proto


teoparvanov (approved):
LGTM!

shgandhi:
LGTM!

PR #2794, opened 2022-09-06
Yf/crypto algo alignment

WildFireFlum:
* **Problem Overview**  
Remove relic and cryptopp from our repository and use openssl instead.

There have been multiple previous pull requests which were not directed to master:
https://github.com/arc-vmware/concord-bft/pull/6
https://github.com/arc-vmware/concord-bft/pull/15
https://github.com/arc-vmware/concord-bft/pull/12


util/include/crypto/openssl/crypto.hpp
```diff
@@ -110,6 +127,67 @@ class AsymmetricPublicKey {
   AsymmetricPublicKey() {}
 };
 
+// Wrapper class for OpenSSL Crypto's EVP_PKEY objects implementing
+// AsymmetricPrivateKey.
+class EVPPKEYPrivateKey : public AsymmetricPrivateKey {
```
WildFireFlum, line 132: This class is needed if we want to use existing code to generate ecdsa keys.
We still need to adapt the interface to support PEM files, which are used by 
 ClientTlsKeyExchangeHandler::exchangeTlsKeys.

WildFireFlum:
The purposes of moving all of the crypto utilities are:
1. it is clear which library is used when and for what
2. libraries can be added/removed easily

Currently, cryptopp cannot be removed as it is used for ecdsa key generation and aes encryption. Both should be replaced by openssl so that we can build without cryptopp.
the usage of cryptopp should be limited to factory.cpp (and maybe another factory for digest algorithms).

PR #2795, opened 2022-09-07
Fix unreleased client request batches

yuliasherman:
* **Problem Overview**  
  We observe that after a VC, the system performance degrades. One of the issues that caused this degradation is an 
  expired client batch that does not get released and as a result, prevents from the corresponding client accepting new 
  requests. The issue that caused the batch not to be occasionally released is the use of the batchCid variable, defined
  as a reference to the corresponding request field. According to the function logic, the request gets moved and the
  variable becomes empty. The use of the moved variables is undefined, that's why we observed an unstable batch
  release behavior.  
* **Testing Done**  
  Tested on a reserved system doing interruption tests.


eyalrund:
please explain RCA and fix

HristoStaykov (approved):
LGTM, thanks for the fix, @yuliasherman !

PR #2796, opened 2022-09-07
add compaction message and rocksdb native compaction

vrancurel:
* **Problem Overview**  
We observe that after a prune (delete range) the space is not physically reclaimed. 
We introduce:
  - A new `pruneCompact` message that does nothing (but triggers compaction if there is no other messages).
  - A Rocksdb native `CompactRange()` method which combined with `DeleteRange()` implements the slow blocks deletion. 
  - A Rocksdb native `DeleteFilesInRange()` method which implements fast -- but imperfect -- blocks deletion.
  - A new parameter in the `IBlocksDeleter` interface to optionally enable fast blocks deletion. 
* **Testing Done**  
Extensive manual tests at scale have been done (up to 1,500,000 blocks) at rest and under load, in order to study the behavior and impact on compaction on cpu, I/O, storage and blockchain transactions.


kvbc/test/replica_state_sync_test.cpp
```diff
@@ -261,7 +261,7 @@ TEST_F(replica_state_sync_on_bft_seq_num_test, cannot_delete_only_block_left_wit
   addBlockWithBftSeqNum(2);  // block 1
   addBlockWithBftSeqNum(3);  // block 2
   // Prune block 1.
-  blockchain_->deleteBlocksUntil(2);
+  blockchain_->deleteBlocksUntil(2, false);
```
teoparvanov, line 264: Isn't it `false` by default?

teoparvanov (approved):
The changes look good, thanks @vrancurel! Although I don't see what change from this PR could be having a side effect on state transfer 🤔 (the Apollo test for ST looks unstable)

kvbc/test/replica_state_sync_test.cpp
vrancurel: It is an instance of `ReplicaBlockchain`, which does not have a default parameter. I implemented default parameters only for the `KeyValueBlockchain` and below classes.

kvbc/src/v4blockchain/detail/blockchain.cpp
```diff
@@ -36,7 +36,8 @@ Blockchain::Blockchain(const std::shared_ptr<concord::storage::rocksdb::NativeCl
   if (genesis_blockId) {
     ConcordAssertGT(genesis_blockId.value(), INVALID_BLOCK_ID);
     ConcordAssertLE(genesis_blockId.value(), last_reachable_block_id_);
-    ConcordAssert(!hasBlock(genesis_blockId.value() - 1));
+    // The following is not necessarily true when we use DeleteFilesInRange()
+    // ConcordAssert(!hasBlock(genesis_blockId.value() - 1));
```
cloudnoize, line 40: I guess it's related to the note in the doc ```Leaking keys on the boundaries (logically and physically) - but it should have no impact on the product.``` right?

It seems that the recent introduction of saving the genesis id as a KV , complements this behavior and prevents deviation across replicas.

kvbc/src/v4blockchain/detail/blockchain.cpp
```diff
@@ -92,8 +93,17 @@ BlockId Blockchain::deleteBlocksUntil(BlockId until) {
...
   // Removes the database entries in the range ["begin_key", "end_key"), i.e.,
   // including "begin_key" and excluding "end_key" --> this is why end is (last_deleted_block + 1)
-  write_batch.delRange(
-      v4blockchain::detail::BLOCKS_CF, generateKey(genesis_block_id_), generateKey(last_deleted_block + 1));
+  //
+  // We sometimes want to re-run the prune so it is better to delete from 1 than
+  // from genesis_block_id_.
+  if (delete_files_in_range) {
+    // Allows fast removal of SST files in a range but can "leak" keys
+    native_client_->deleteFilesInRange(
```
cloudnoize, line 101: it can be nice to have a log line that indicates which deletion took place

kvbc/include/kvbc_app_filter/kvbc_key_types.h
```diff
@@ -36,6 +36,7 @@ const char kKvbKeySummarizedTime = 0x22;
 const char kKvbKeyCorrelationId = 0x23;
 const char kKvbKeyLastAgreedPrunableBlockId = 0x24;
 const char kIndexKey = 0x25;
+const char kKvbKeyDeleteFilesInRange = 0x26;
```
cloudnoize, line 39: what is the usage of this key?

kvbc/src/v4blockchain/detail/blockchain.cpp
cloudnoize: Can it fail Unit tests? Where a block is assumed to be deleted? 
Did we think if it can cause deviation across replicas and whether those deviations are harmful?

cloudnoize (approved):
Looks great @vrancurel , left some questions that can be asynchronously answered

kvbc/include/kvbc_app_filter/kvbc_key_types.h
vrancurel: We persist the type of deletion (e.g. `delete_files_in_range`), in case of crash and restart.

kvbc/src/v4blockchain/detail/blockchain.cpp
vrancurel: Apparently it does not fail unit tests.
It will cause deviation across replicas about keys which are not supposed to exist. The logical keys leaking are not supposed to be referenced any more, so it will normally not impact state transfer, block hashes, merkle trees, etc.

yontyon (approved):
LGTM!

glevkovich:
@vrancurel,
The call in BlocksDeleterAdapter::deleteBlocksUntil to: 
`auto upTo = kvbc_->deleteBlocksUntil(until, delete_files_in_range)`
Must first report to State transfer, else it might fail to validate the RVB group during cycle. see an example of the call to IStateTransfer::reportLastAgreedPrunableBlockId in Replica::deleteBlocksUntil:
```
BlockId Replica::deleteBlocksUntil(BlockId until, bool delete_files_in_range) {
  // Inform State Transfer about pruning. We must do it in this thread context for persistency considerations, and in
  // this layer to lower the chance for bugs (there are multiple callers to this function), in which pruning is not
  // notified to ST. In that case, ST state will be corrupted.
  if (m_stateTransfer && m_stateTransfer->isRunning()) {
    // We assume until > 0, see check above
    m_stateTransfer->reportLastAgreedPrunableBlockId(until - 1);
  }
  ISystemResourceEntity::scopedDurMeasurment mes(replicaResources_, ISystemResourceEntity::type::pruning_utilization);
  return m_kvBlockchain->deleteBlocksUntil(until, delete_files_in_range);
}

```

kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -823,6 +823,23 @@ bool ReconfigurationHandler::handle(const concord::messages::PruneRequest& comma
   return true;
 }
 
+// This is the only way to trigger the compaction outside of a snapshot if the system is idle
+bool ReconfigurationHandler::handle(const concord::messages::PruneCompactRequest& command,
```
vrancurel, line 827: @cloudnoize if I don't create this handler, the `KeyValueBlockchain::onNewBFTSequenceNumber()` method is not called and therefore the check for compaction is not done.

Is there another less intrusive way to trigger this method ?

cloudnoize: ```onNewBFTSequenceNumber``` currently is called only when a new block is added with a new BFT sequence number.
We can probably add a call to this method as a callback from the BFT engine after it has finished the execution of requests. It will remove the dependency on a new subsequence request to release the snapshot  and will support the required behavior I think

kvbc/include/v4blockchain/detail/blockchain.h
```diff
@@ -119,6 +122,8 @@ class Blockchain {
   std::shared_ptr<concord::storage::rocksdb::NativeClient> native_client_;
   util::ThreadPool thread_pool_{1};
   std::optional<std::future<BlockDigest>> future_digest_;
+  bool need_compaction_{false};
+  std::mutex compaction_mutex_;
```
cloudnoize, line 126: I don't think we need to protect ```need_compaction_```
It's either we deleteblocks i.e. pruning or adding a block, both operations can't happen in parallel.

kvbc/src/kvbc_adapter/replica_adapter.cpp
```diff
@@ -108,10 +109,15 @@ ReplicaBlockchain::ReplicaBlockchain(
...
 }
 
-BlockId ReplicaBlockchain::deleteBlocksUntil(BlockId until) {
+BlockId ReplicaBlockchain::deleteBlocksUntil(BlockId until, bool delete_files_in_range) {
+  LOG_INFO(logger_, "deleteBlocksUntil:" << KVLOG(until));
   auto prev_dur = delete_blocks_until_duration.Get().Get();
   auto start = std::chrono::steady_clock::now();
-  auto id = deleter_->deleteBlocksUntil(until);
+  if (delete_blocks_state_transfer_cb_) {
+    delete_blocks_state_transfer_cb_(until);
```
cloudnoize, line 117: @glevkovich, the existing code calls this method from ```Replica::deleteBlocksUntil``` which is a wrapper to the replica adapter.
Why do we also need to add a call here? Wouldn't it be called twice?

kvbc/src/v4blockchain/detail/blockchain.cpp
```diff
@@ -92,17 +93,49 @@ BlockId Blockchain::deleteBlocksUntil(BlockId until) {
...
-  LOG_INFO(V4_BLOCK_LOG, "Deleted " << blocks_deleted << " blocks, new genesis is " << genesis_block_id_);
+  LOG_INFO(V4_BLOCK_LOG,
+           "Deleted " << blocks_deleted << " blocks, new genesis is " << genesis_block_id_
+                      << " delete_files_in_range is " << delete_files_in_range);
   return last_deleted_block;
 }
 
+// atomically get and reset the need_compaction_ flag
+// to ensure that compaction is not started unnecessarily
+bool Blockchain::needCompaction() {
```
cloudnoize, line 123: As per my comment above, I don't think two threads can access ```need_compaction_``` in parallel.

kvbc/include/v4blockchain/detail/blockchain.h
vrancurel: It is to avoid a theoretical race between the compaction async function resetting the flag and another block being created.

kvbc/src/reconfiguration_kvbc_handler.cpp
vrancurel: Yes, this is an optimization: we could call the compaction immediately after the execution of PruneRequest once the reconfiguration block is persisted.

rastogiasr (approved):

PR #2775, opened 2022-08-17
Thin replica improve metrics and logs

glevkovich:
Problem Overview
This PR aims to improve visibility and debuggability in thin replica client-server and some parts of client service (e.g events service). This is done by fixing, adding improving logs, metrics, and histograms.
In addition, minimal (opportunistic) refactoring is done.
Commits:

Thin-Replica, event_service: add additional logs to improve debugging
thin_replica_client,event_service,concord_client: rename, add and modify metrics
concord_client: fix wrong variable name
thin_replica_impl: fix bug in ThinReplicaImpl::sendHash
clientservice, trc: refactor (shorten) use of std::chrono members
event_service: refactor metrics, and had histograms (snapshots)
Testing Done
Maestro manual test to check for stability, and that all logs apear as expected.

PR #2797, opened 2022-09-08
Remove incorrectly obtained TRC metric

shgandhi:
This PR - https://github.com/vmware/concord-bft/pull/2775, doesn't consider updates can be of type EventGroup when calculating metrics in pushUpdateToUpdateQueue method. This is causing all runs with event groups enabled to fail. This commit fixes this issue.

Testing done using unit tests and end-to-end test.


bsenthilr (approved):
was the hit caused by trace log?

shgandhi:
> was the hit caused by trace log?

No. The update pushed to the update queue can be of type `Update` or `EventGroup`, the following two lines assumed the EventVariant type `update` is always of type `Update`, and doesn't correctly handle for `EventGroups`.

```
const auto& block_id = std::get<Update>(*update).block_id;
auto num_events = std::get<Update>(*update).kv_pairs.size();
```

eyalrund:
Cant we find a different way to collect the metrics instead of just removing the code?

arc-vmware (approved):
LGTM!!!

shgandhi:
> Cant we find a different way to collect the metrics instead of just removing the code?

@eyalrund, I will create a separate MR for refactoring the changes. Will need Gil's input on that.

PR #2799, opened 2022-09-09
Thread pool support for thin replica client GRPC connection

bsenthilr:
Think replica clients GRPC connections leverage std::async with async launch policy that spins new threads as required. This change adds support to execute these task asynchronously on a thread pool. The thread pool currently only comprise of single thread. But we can tune this in future as required to delegate more worker threads. 

The current suite of tests were regressed. In addition this branch was performance tested with event group feature toggled to validate TPS and other memory resources.


client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -212,6 +225,11 @@ class GrpcConnection {
   std::shared_mutex rss_streams_mutex_;
   std::atomic_uint64_t current_req_id_{1};
 
+  // Reader threads
+  std::unique_ptr<concord::util::ThreadPool> worker_thread_pool_;
```
shgandhi, line 229: Nit: can we rename `worker_thread_pool_` to `subscription_pool_` for better context?

client/thin-replica-client/include/client/thin-replica-client/grpc_connection.hpp
```diff
@@ -77,17 +78,29 @@ class GrpcConnection {
   // Possible results of RPC operations a GrpcConnection may report.
   enum class Result { kUnknown, kSuccess, kFailure, kTimeout, kOutOfRange, kNotFound, kEndOfStream };
 
+  static const size_t maxWorkerThreadPoolSize = 8;
```
shgandhi, line 81: Why did we choose size 8 for max pool size?

bsenthilr: The default size is 1 thread. The max is to guard against any misuse. No specific reason for 8 whoever tunes it in future can bump it as required.

shgandhi:
@bsenthilr thanks for the change. LGTM! I just have a couple of minor comments and a request.

Gentle request: can we clean the commit history in this PR? currently it has 10 commits, of which only the top commit is relevant. In https://github.com/vmware/concord-bft/blob/master/CONTRIBUTING.md we specify the following commands to keep our feature-branch up-to-date (instead of merging remote into our feature branch, which results in a merge commit.) -

```
git checkout my-new-feature
git fetch -a
git pull --rebase upstream master
git push --force-with-lease origin my-new-feature
```
It would be great if you could look into it, and squash/remove the merge and irrelevant commits.

PR #2805, opened 2022-09-13
Add de/serialization to the new utt api

yontyon:
* **Problem Overview**  
  Add de/serialization to the new UTT API objects
* **Testing Done**  
  unittest + CI


utt/libutt/src/api/burn.cpp
```diff
@@ -6,7 +6,20 @@
...
 #include <utt/RegAuth.h>
+#include <ostream>
 
+std::ostream& operator<<(std::ostream& out, const libutt::api::operations::Burn& burn) {
+  out << *(burn.burn_) << std::endl;
+  out << burn.c_ << std::endl;
+  return out;
+}
+std::istream& operator>>(std::istream& in, libutt::api::operations::Burn& burn) {
+  in >> *(burn.burn_);
```
cloudnoize, line 17: why do you handle pointers differently ?  on paramters you perform ```params.params.reset(new libutt::Params());``` and here just override the values?

utt/libutt/src/api/commitment.cpp
```diff
@@ -12,6 +12,20 @@ libutt::api::Commitment operator+(libutt::api::Commitment lhs, const libutt::api
   return lhs;
 }
 
+std::ostream& operator<<(std::ostream& out, const libutt::api::Commitment& comm) {
+  out << *(comm.comm_);
+  return out;
+}
+std::istream& operator>>(std::istream& in, libutt::api::Commitment& comm) {
+  in >> *(comm.comm_);
```
cloudnoize, line 20: Same question on pointers

utt/libutt/src/api/mint.cpp
```diff
@@ -6,10 +6,30 @@
...
+  in >> *(mint.op_);
+  return in;
+}
+
 namespace libutt::api::operations {
 Mint::Mint(const std::string& uniqueHash, size_t value, const std::string& recipPID) {
   op_.reset(new libutt::MintOp(uniqueHash, value, recipPID));
 }
+Mint::Mint() { op_.reset(new libutt::MintOp()); }
+Mint::Mint(const Mint& other) {
```
cloudnoize, line 23: When do we copy the classes that you added copy operator for them?

cloudnoize (approved):
LGTM!! thanks @yontyon

utt/libutt/src/api/burn.cpp
yontyon: In burn, the default constructor already initiates the pointer. With UTTParams we are unable to create globalParams object before initiating the underlying libraries. This is why we don't have a default constructor there

utt/libutt/src/api/commitment.cpp
yontyon: same answer :)

utt/libutt/src/api/mint.cpp
yontyon: Currently, we don't. I have added a copy construct as part of the paradigm that assignment operator and copy constructor always come together

evdzhurov (commented):
In classes where you explicitly declared a copy constructor/assignment operator move operations are disabled. Any reason these classes can't support moves?

yontyon:
> In classes where you explicitly declared a copy constructor/assignment operator move operations are disabled. Any reason these classes can't support moves?

@evdzhurov , the only reason we need an explicit cctor/assignment operator is that we use unique_ptrs for the internal utt interface.
However, moving unique_ptr (or shared_ptr) is trivial, hence the trivial move ctor/move assignment operator are enough.

utt/tests/TestSerialization.cpp
```diff
@@ -0,0 +1,203 @@
...
+                       libutt::api::Coin::Type::Budget,
+                       Fr::random_element().to_words());
+  std::vector<uint8_t> c4_serialized = libutt::api::serialize<libutt::api::Coin>(c4);
+  libutt::api::Coin c4_deserialized = libutt::api::deserialize<libutt::api::Coin>(c4_serialized);
+  assertTrue(c4.getNullifier() == c4_deserialized.getNullifier());
+  assertTrue(c4.hasSig() == c4_deserialized.hasSig());
+  assertTrue(c4.getType() == c4_deserialized.getType());
+  assertTrue(c4.getVal() == c4_deserialized.getVal());
+  assertTrue(c4.getSN() == c4_deserialized.getSN());
+  assertTrue(c4.getExpDateAsCurvePoint() == c4_deserialized.getExpDateAsCurvePoint());
```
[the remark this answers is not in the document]
yontyon: done

evdzhurov (commented):
You aren't doing anything special in the move constructors and the move assignment operators, so the compiler generated versions of these functions (which do member-wise moves) can do the job.

You can simply default the move operations as such:
  Burn(Burn&& other) = default;
  Burn& operator=(Burn&& other) = default;

Even though you're asking for the defaults explicitly declaring the move operations marks them as "user defined" and they will not be omitted by the compiler.

yontyon:
> You aren't doing anything special in the move constructors and the move assignment operators, so the compiler generated versions of these functions (which do member-wise moves) can do the job.
> 
> You can simply default the move operations as such: Burn(Burn&& other) = default; Burn& operator=(Burn&& other) = default;
> 
> Even though you're asking for the defaults explicitly declaring the move operations marks them as "user defined" and they will not be omitted by the compiler.

fixed, please review and approve

evdzhurov (approved):
Thank you.

PR #2808, opened 2022-09-16
Thin Replica Server Metric to show history updates

mdarade:
Thin Replica Server Metric to show history updates

* **Problem Overview**  
Thin replica server metric were not showing history updates.

* **Testing Done**  
Executed all available tests.


thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -755,8 +764,13 @@ class ThinReplicaImpl {
           auto num_events = kvb_update.kv_pairs.size();
           auto client_id = getClientId(context);
           LOG_DEBUG(logger_, "Sending updates (history, data)" << KVLOG(client_id, block_id, num_events));
-
           sendData(stream, kvb_update);
+          metrics.last_sent_block_id.Get().Set(block_id);
+          last_sent_block_id_ = block_id;
+          if (++update_aggregator_counter == config_->update_metrics_aggregator_thresh) {
```
glevkovich, line 770: no need to make a change on this but why not just use a module instead of ceqoing the counter?

glevkovich (approved):
LGTM, i wrote small comment which doesn't require a change.
Thanks!

arc-vmware (approved):
LGTM!!!

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -755,8 +764,13 @@ class ThinReplicaImpl {
           auto num_events = kvb_update.kv_pairs.size();
           auto client_id = getClientId(context);
           LOG_DEBUG(logger_, "Sending updates (history, data)" << KVLOG(client_id, block_id, num_events));
-
           sendData(stream, kvb_update);
+          metrics.last_sent_block_id.Get().Set(block_id);
```
shgandhi, line 768: Instead of updating this metric here and in [SubscribeToUpdates #L441](https://github.com/vmware/concord-bft/blob/72cfb4761fd6a7236b5a9a21b354c1c4f4775ce4/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp#L441), why don't we update the metric in `SendData`/`SendHash` methods? If for some reason stream is closed when writing data over the stream, then we would still persist the correct value of last_sent_block_id, and not be off by 1. WDYT?

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -728,10 +736,11 @@ class ThinReplicaImpl {
                               ServerWriterT* stream,
                               kvbc::BlockId start,
                               kvbc::BlockId end,
-                              std::shared_ptr<kvbc::KvbAppFilter> kvb_filter) {
+                              std::shared_ptr<kvbc::KvbAppFilter> kvb_filter,
+                              ThinReplicaServerMetrics& metrics) {
     boost::lockfree::spsc_queue<kvbc::KvbFilteredUpdate> queue{10};
     std::atomic_bool close_stream = false;
-
+    uint16_t update_aggregator_counter = 0;
```
shgandhi, line 743: Why are we creating a new counter here? We already have a counter defined [here](https://github.com/vmware/concord-bft/blob/72cfb4761fd6a7236b5a9a21b354c1c4f4775ce4/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp#L287). We must use a single counter for a single component (TRS) to track the frequency with which we want to update the metrics aggregator. The idea is to update the metrics after every 100 blocks, whether we get them from storage or live update queue.

shgandhi (changes requested):
@mdarade thanks for the fix. I have requested a few changes.
Additionally, can you please update your commit message to be more descriptive. I am not sure the term `history updates` is known by everyone.
Can you add probably something along the lines below -

> Update TRS metric `last_sent_block_id` when reading from storage
Earlier `last_sent_block_id` was only updated after an update was read and sent from a live update queue. This commit fixes the gap by also updating the metric every time an update is read and sent from storage.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -286,6 +291,7 @@ class ThinReplicaImpl {
     metrics.setAggregator(aggregator_);
     uint16_t update_aggregator_counter = 0;
     metrics.subscriber_list_size.Get().Set(config_->subscriber_list.Size());
+    metrics.last_sent_block_id.Get().Set(last_sent_block_id_);
```
shgandhi, line 294: Note, the metric in this case may not monotonically increase. Consider a scenario where TRS sent 100 blocks to a client. The client only received 70 blocks (perhaps there was no agreement on the 71st block), the client would resubscribe for 71st block, so the metric would change from 100 -> 71 when the 71st block is sent by the replica. Just an fyi.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
```diff
@@ -120,6 +120,9 @@ class ThinReplicaImpl {
   // last timestamp when subscription status for live updates was not ok
   std::optional<std::chrono::steady_clock::time_point> last_failed_subscribe_status_time;
 
+  // TBD Persist to resume
+  uint64_t last_sent_block_id_{0};
```
shgandhi, line 124: We will need to keep a map of this metric for a client ID and stream_type. How would we know otherwise if we are correctly updating the metric? 
Client1 subscribed, last sent block ID was 100. Client2 subscribed to block 0, hasn't received any blocks, we are initializing the metric to 100 incorrectly!

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
mdarade: We throw exception when we stream is found to be closed. In such case, remaining code incrementing metric won't be executed.

mdarade (commented):
Thanks Shruti.

mdarade:
> @mdarade thanks for the fix. I have requested a few changes. Additionally, can you please update your commit message to be more descriptive. I am not sure the term `history updates` is known by everyone. Can you probably add something along the lines below -
> 
> > Update TRS metric `last_sent_block_id` when reading from storage
> > Earlier `last_sent_block_id` was only updated after an update was read and sent from a live update queue. This commit fixes the gap by also updating the metric every time an update is read and sent from storage.
> > Additionally, whenever a new subscription gets started, the metric is updated to reflect the last block sent in the previous subscription round.

I'll update the commit message. Thanks.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: With this change we will have two counters to identify when to update aggregator. Is that the purpose? We use a separate counter for storage updates and live updates?
Consider the scenario, TRS streams 70 updates from storage, and then 30 updates from live update queue. In this case we will not update the aggregator until 100 updates are sent from the live update queue, no? Am I missing anything? For such scenarios, I would suggest we use the counter declared [here](https://github.com/vmware/concord-bft/blob/3b66c30c521980f85f39058c2cdbb64a960b50d2/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp#L292), pass it to `readFromKvbAndSendData` via `syncAndSend`. WDYT?

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: I am not comparing b/w live updates and storage updates. My main concern is at TRS serves multiple clients at a time, and when we look at `last_sent_block_id` metric, we should know for which client and for which stream `last_sent_block_id` corresponds to, when it is initially set [here](https://github.com/vmware/concord-bft/blob/3b66c30c521980f85f39058c2cdbb64a960b50d2/thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp#L294) on a new subscription. Otherwise, we will be incorrectly setting the metric if the same TRS serves multiple clients and/or multiple streams.

> Tags will help us conclude that TRS served client 1 up to block id 100 and then it started serving history updates to client 2 starting from 0.

if the private variable `last_sent_block_id_` has been set to 100 for client1, and then client2 starts a new subscription for block 0, with this change, you will be setting the last_sent_block_id metric to 100 for client2, even though TRS has not yet streamed any blocks to client2.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: Yes, thanks, I missed that.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
mdarade: Addressed but in different way. Thanks.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
mdarade: Addressed. Thanks.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: Thanks for the fix.

thin-replica-server/include/thin-replica-server/thin_replica_impl.hpp
shgandhi: It looks like now we are not persisting last_sent_block_id metric between subscriptions? For every new subscription, TRS metric `last_sent_block_id` would always be 0, even if TRS had sent x number of blocks to a client in a previous subscription round? In that case, please remove the following last line of the commit message -

```
Additionally, whenever a new subscription gets started, the metric is
updated to reflect the last block sent in the previous subscription round.
```

mdarade: Done. Thanks.

shgandhi (approved):

PR #2814, opened 2022-09-20
Read Only Replica to export last-executed-seq-num starting with 0

mdarade:
Read Only Replica to export last-executed-seq-num starting with 0

* **Problem Overview** 
After the deployment, Sequence Number is not 0 for the FCC nodes. It happened as last-executed-seq-num was being used without initialization.
* **Testing Done**  
Did few runs in docker compose environment.


cjn:
LGTM. Thanks @mdarade

cjn (approved):

eyalrund (approved):

PR #2815, opened 2022-09-21
Adding compilation flags to control enabling/disabling low priority metrics in concord.

snehal-das:
The compilation flags added below can be turned ON/OFF to enable or disable metrics at compile time.


CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ option(BUILD_THIRDPARTY "Wheter to build third party librarie or use preinstalle
 option(CODECOVERAGE "Enable Code Coverage Metrics in Clang" OFF)
 option(ENABLE_RESTART_RECOVERY_TESTS "Enable tests for restart recovery" OFF)
 option(BUILD_UTT "Build UTT library" OFF)
+option(CONCORD_ENABLE_ALL_METRICS "Enable all metrics in concord" ON)
```
toly-kournik, line 50: How is this option reflects in code?

snehal-das: This option has to be removed as it is redundant. It was added when I was exploring ways to add compile time flags. I went with adding the flag in Makefile in this commit.

I need to evaluate if there are other ways to build concord, in which case I **might** need this variable.

arc-vmware (approved):
LGTM !!!

cjn (approved):
LGTM. Thanks @snehal-das !

snehal-das:
5 day LRT run was run successfully with these changes added; no issues seen.

PR #2816, opened 2022-09-21
Disabling test until stabilized.

HristoStaykov:
Test test_stop_primary_during_initiated_view_change is failing. We need to stabilize it before re-enabling back.

* **Problem Overview**  
Test test_stop_primary_during_initiated_view_change is failing.
We need to stabilize it before re-enabling back.
* **Testing Done**  
N/A


teoparvanov (approved):
LGTM, thanks Hristo!

PR #2818, opened 2022-09-23
Fix metric infra to enable tags in Thin Replica Server

mdarade:
Fix metric infra to enable tags in Thin Replica Server

* **Problem Overview**  
TRS can start serving live/history updates to any newly subscribed TRC. Wavefront needs to have a mapping from where this information can be derived. This commit would fix bug in metric infra.
* **Testing Done**  
Executed all TRC, TRS tests.

### DIFF
--- a/util/src/Metrics.cpp
+++ b/util/src/Metrics.cpp
@@ -39,15 +39,7 @@ T FindValue(const char* const val_type, const string& val_name, const vector<str
 
 /******************************** Class Component ********************************/
 
-Component::Handle<Gauge> Component::RegisterGauge(const string& name, const uint64_t val) {
-  names_.gauge_names_.emplace_back(name);
-  values_.gauges_.emplace_back(Gauge(val));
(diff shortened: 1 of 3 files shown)


util/src/Metrics.cpp
```diff
@@ -92,7 +76,7 @@ std::list<Metric> Component::CollectGauges() {
   if (!metricsEnabled_) return list<Metric>();
   std::list<Metric> ret;
   for (size_t i = 0; i < names_.gauge_names_.size(); i++) {
-    if (tags_.gauge_tags_.size() > i) {
+    if (tags_.gauge_tags_[i].size()) {
```
shgandhi, line 79: Nice catch!

shgandhi:
Thanks for the fix and updating the test @mdarade. LGTM!
One request, can you please describe the issue and what the fix does in the commit message.

shgandhi (approved):

arc-vmware (approved):
LGTM!!!

PR #2819, opened 2022-09-27
Add custom health service for clientservice

shgandhi:
This PR adds a custom health service, specifically custom Check/Watch methods for request service. This PR also adds the necessary code that would be easy to extend for event and stat-snapshot services' health. Please see individual commits for more detail.

Testing is done via grpcurl -
`grpcurl -plaintext -d '{"service":"vmware.concord.client.request.v1.RequestService"}' --proto health.proto <clientservice-ip>:50505 grpc.health.v1.Health/Check` and 
`grpcurl -plaintext -d '{"service":"vmware.concord.client.request.v1.RequestService"}' --proto health.proto <clientservice-ip>:50505 grpc.health.v1.Health/Watch`

Will add end-to-end tests once this PR gets merged.

### DIFF
--- a/client/clientservice/include/client/clientservice/health_service.hpp
+++ b/client/clientservice/include/client/clientservice/health_service.hpp
@@ -45,6 +45,10 @@ class HealthCheckServiceImpl : public grpc::health::v1::Health::Service {
   // atleast one client in the client pool is ready to serve requests
   HealthCheckResponse_ServingStatus getRequeserviceHealthStatus();
 
+  // updates the overall clientservice health, i.e., reports healthy
+  // iff all the services in clientservice are healthy
+  void updateAggregateHealth();
+
  public:
   const std::string kRequestService{"vmware.concord.client.request.v1.RequestService"};
   const std::string kEventService{"vmware.concord.client.event.v1.EventService"};
--- a/client/clientservice/include/client/clientservice/health_service.hpp
+++ b/client/clientservice/include/client/clientservice/health_service.hpp
@@ -1,0 +1,79 @@
+// Concord
+//
+// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the "License"). You may not use this product except in
+// compliance with the Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright notices and license terms. Your use of
+// these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#include <map>
+#include <mutex>
+
+#include <grpcpp/server_context.h>
+#include <grpcpp/support/status.h>
+#include "client/concordclient/concord_client.hpp"
+
+#include <health.grpc.pb.h>
+
+namespace concord::client::clientservice {
+
+using grpc::health::v1::HealthCheckRequest;
+using grpc::health::v1::HealthCheckResponse;
+using grpc::health::v1::HealthCheckResponse_ServingStatus;
+
+// A sync implementation of the health checking service.
+class HealthCheckServiceImpl : public grpc::health::v1::Health::Service {
+ public:
+  HealthCheckServiceImpl(std::shared_ptr<concord::client::concordclient::ConcordClient> client)
+      : logger_(logging::getLogger("concord.client.clientservice.event")), client_(client){};
+  grpc::Status Check(grpc::ServerContext* context,
+                     const HealthCheckRequest* request,
+                     HealthCheckResponse* response) override;
+  grpc::Status Watch(grpc::ServerContext* context,
+                     const HealthCheckRequest* request,
+                     grpc::ServerWriter<HealthCheckResponse>* writer) override;
+  void SetStatus(const std::string& service_name, HealthCheckResponse::ServingStatus status);
+  void SetAll(HealthCheckResponse::ServingStatus status);
+
+  void Shutdown();
+
+ private:
+  // Queries the health status of the client_pool via the concord client and returns true if
+  // atleast one client in the client pool is ready to serve requests
+  HealthCheckResponse_ServingStatus getRequeserviceHealthStatus();
+
+ public:
+  const std::string kRequestService{"vmware.concord.client.request.v1.RequestService"};
+  const std::string kEventService{"vmware.concord.client.event.v1.EventService"};
+  const std::string kStateSnapshotService{"vmware.concord.client.statesnapshot.v1.StateSnapshotService"};
+
+ private:
+  logging::Logger logger_;
+  std::mutex mtx_;
+  bool shutdown_ = false;
+  std::shared_ptr<concord::client::concordclient::ConcordClient> client_;
+  std::map<const std::string, HealthCheckResponse::ServingStatus> status_map_;
+};
+
+class HealthCheckService : public grpc::HealthCheckServiceInterface {
+ public:
+  explicit HealthCheckService(HealthCheckServiceImpl* impl) : impl_(impl) {
+    impl_->SetStatus("", HealthCheckResponse::SERVING);
+  }
+  void SetServingStatus(const std::string& service_name, bool serving) override {
+    impl_->SetStatus(service_name, serving ? HealthCheckResponse::SERVING : HealthCheckResponse::NOT_SERVING);
+  }
+
+  void SetServingStatus(bool serving) override {
+    impl_->SetAll(serving ? HealthCheckResponse::SERVING : HealthCheckResponse::NOT_SERVING);
+  }
+
+  void Shutdown() override { impl_->Shutdown(); }
+
+ private:
+  std::shared_ptr<HealthCheckServiceImpl> impl_;  // not owned
+};
+}  // namespace concord::client::clientservice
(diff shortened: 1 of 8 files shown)


client/concordclient/src/concord_client.cpp
```diff
@@ -232,4 +232,10 @@ void ConcordClient::getSnapshot(const StateSnapshotRequest& request, std::shared
   rss_->readSnapshotStream(rss_request, remote_queue);
 }
 
+bool ConcordClient::isClientPoolHealthy() {
+  if (client_pool_ and client_pool_->HealthStatus() == concord::concord_client_pool::PoolStatus::NotServing)
```
upshaw-alex, line 236: With this condition and return value placement, it looks like `isClientPoolHealthy` will return `true` if `client_pool_ is null. Is this intentional? If so, can you comment on the reasoning for this behavior?

(Sorry if I am missing or overlooking any relevant context for this behavior).

client/clientservice/include/client/clientservice/health_service.hpp
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  void SetServingStatus(bool serving) override {
+    impl_->SetAll(serving ? HealthCheckResponse::SERVING : HealthCheckResponse::NOT_SERVING);
+  }
+
+  void Shutdown() override { impl_->Shutdown(); }
+
+ private:
+  HealthCheckServiceImpl* impl_;  // not owned
```
arc-vmware, line 77: As it's not owned, we should use shared_ptr, so that the owner will not delete this ptr unless its out of scope from this class.

client/clientservice/include/client/clientservice/health_service.hpp
```diff
@@ -0,0 +1,79 @@
+// Concord
+//
+// Copyright (c) 2022-23 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
shgandhi: Thanks for the catch, updated.

client/clientservice/src/health_service.cpp
```diff
@@ -0,0 +1,92 @@
+// Concord
+//
+// Copyright (c) 2022-23 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
shgandhi: Updated.

client/clientservice/src/health_service.cpp
```diff
@@ -0,0 +1,92 @@
...
+        response.set_status(getRequeserviceHealthStatus());
+      } else {
+        response.set_status(iter->second);
+      }
+      if (response.status() != last_state) {
+        writer->Write(response, grpc::WriteOptions());
+        last_state = response.status();
+      }
+    }
+    gpr_sleep_until(gpr_time_add(gpr_now(GPR_CLOCK_MONOTONIC), gpr_time_from_millis(1000, GPR_TIMESPAN)));
```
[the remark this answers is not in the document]
shgandhi: This is for simplicity. For lower performance impact, the user of the health check has the alternative `Check` method to get the health status of a service. In any case, I have now increased the sleep time to 1 minute for this functionality.

client/clientservice/include/client/clientservice/health_service.hpp
```diff
@@ -0,0 +1,79 @@
...
+  const std::string kRequestService{"vmware.concord.client.request.v1.RequestService"};
+  const std::string kEventService{"vmware.concord.client.event.v1.EventService"};
+  const std::string kStateSnapshotService{"vmware.concord.client.statesnapshot.v1.StateSnapshotService"};
+
+ private:
+  logging::Logger logger_;
+  std::mutex mtx_;
+  bool shutdown_ = false;
+  std::shared_ptr<concord::client::concordclient::ConcordClient> client_;
+  std::map<const std::string, HealthCheckResponse::ServingStatus> status_map_;
```
arc-vmware, line 58: If we know the service names and numbers in advance, then we can do the following:
public:
enum SERVICE_ID : size_t {
ReqService = 0,
EvtService = 1,
SnpService = 2,
MaxService=3
};

private:
std::vector<std::atomic<HealthCheckResponse::ServingStatus>> status_vec_(SERVICE_ID.MaxService, HealthCheckResponse::NOT_SERVING)

I am not sure if the above syntax is correct or not, so if it's not syntactically fine, then please ignore the comment.
If something like above is possible, then we can get rid of the mutexes and locks and directly use the status_vec_ instead of status_map_

arc-vmware (commented):
@shgandhi good code, I have very few comments. 
Thanks

client/concordclient/src/concord_client.cpp
upshaw-alex: Thank you for this revision.

client/clientservice/src/health_service.cpp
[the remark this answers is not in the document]
upshaw-alex: Thank you for confirming the polling is just for simplicity, that is fine. Honestly, increasing the wait time to 1 minute, I would then be worried whether the latency for updating watchers about the status changes would be too high, though I would also be fine if we left the polling interval as it is at 1 minute as I am not sure there is much point in trying to optimize it unless we already have a specific use case for this in mind.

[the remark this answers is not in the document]
shgandhi: Yes, I am leaving it to 1 minute for now. Will update if there is a need for it.

client/clientservice/src/client_service.cpp
```diff
@@ -13,13 +13,17 @@
...
+#include <grpcpp/health_check_service_interface.h>
+#include <grpcpp/ext/proto_server_reflection_plugin.h>
+#include <grpcpp/ext/health_check_service_server_builder_option.h>
 
 using namespace std::literals::chrono_literals;
 
 namespace concord::client::clientservice {
 
 void ClientService::start(const std::string& addr, unsigned num_async_threads, uint64_t max_receive_msg_size) {
-  grpc::EnableDefaultHealthCheckService(true);
```
bsenthilr, line 22: did we review if there is a way to extend or register callbacks into the default health check service implementation? Seems their code is well structured but not sure the classes are accessible for external modules.

client/clientservice/src/health_service.cpp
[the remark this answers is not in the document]
bsenthilr: Reviewing grpc default health service implementation. Seems they are based on completion queues and async processing. If we follow similar we can eliminate this polling. Lets review that and sync up.

client/clientservice/include/client/clientservice/health_service.hpp
shgandhi: @arc-vmware if we use shared_ptr, we indicate shared ownership to the pointer instead of no-ownership. Afaik T* is a known convention for non-owned pointers.
Referred here in the official c++ guidelines -> https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c32-if-a-class-has-a-raw-pointer-t-or-reference-t-consider-whether-it-might-be-owning

client/clientservice/include/client/clientservice/health_service.hpp
shgandhi: @arc-vmware can you pls elaborate as to why using a vector of enums instead of map would help us get rid of mutexes and locks? The idea is that with usage of locks and mutex we guarantee consistent responses to the caller and we trade-off performance to achieve that. Pls let me know if I have missed anything.

client/clientservice/src/client_service.cpp
shgandhi: @bsenthilr we are deriving from the default health check service, pls refer to https://github.com/vmware/concord-bft/pull/2819/files#diff-7ede078b0f0a07554b6fd9358bebe4c37b3dec7d492cd0a899e62e0d0acf3d5f -

```
// A sync implementation of the health checking service.
class HealthCheckServiceImpl : public grpc::health::v1::Health::Service {
 public:
...
}
```

We needed to be able to query client-pool -> replica connections everytime `Check`/`Watch` are called, hence these methods had to be overridden and custom implementations added.

upshaw-alex (approved):

client/clientservice/include/client/clientservice/health_service.hpp
shgandhi: Had a discussion with @bsenthilr, updated for now.

bsenthilr (approved):
LGTM - we need to evaluate and eventually adapt back to leverage default implementation adding support to backend pushing up health status hooks ,

client/proto/health/v1/health.proto
```diff
@@ -0,0 +1,63 @@
...
+  // If the requested service is unknown, the call will fail with status
+  // NOT_FOUND.
+  rpc Check(HealthCheckRequest) returns (HealthCheckResponse);
+
+  // Performs a watch for the serving status of the requested service.
+  // The server will immediately send back a message indicating the current
+  // serving status.  It will then subsequently send a new message whenever
+  // the service's serving status changes.
+  //
+  // If the requested service is unknown when the call is received, the
```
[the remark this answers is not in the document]
shgandhi: This health.proto is officially provided by grpc - https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto. What behavior are you exactly referring to?

arc-vmware (approved):
Changes look great. Thanks @shgandhi

PR #2820, opened 2022-09-28
Enable UTT compilation by default.

evdzhurov:
* **Problem Overview**  
  UTT compilation was previously enabled only through the github CI because some of libutt's dependencies could cause compilation issues in other project environments where concord-bft is used. Additional cmake flags have been added to fix this and compilation of UTT is now enabled by default.
* **Testing Done**  
  Tested compiling the project.


teoparvanov (approved):
Awesome, thank you @evdzhurov !

PR #2821, opened 2022-09-28
Increase max_reply_size

nkgeorgiev:
* **Problem Overview**  
We need larger replies in some ethereum workloads (UTT for example)
* **Testing Done**  
Maestro runs with the new size
Manual test with larger requests and replies.

### DIFF
--- a/client/bftclient/include/bftclient/config.h
+++ b/client/bftclient/include/bftclient/config.h
@@ -76,7 +76,7 @@ struct ClientConfig {
 struct RequestConfig {
   bool pre_execute = false;
   uint64_t sequence_number = 0;
-  uint32_t max_reply_size = 64 * 1024;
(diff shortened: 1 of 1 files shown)


eyalrund:
We are currently trying to reduce the memory foot print of concord. What are the implications of this change in terms of memory consumption? Is it effecting the replicas? is it multiplied by the number of clients and the size of the batch?

nkgeorgiev:
> We are currently trying to reduce the memory foot print of concord. What are the implications of this change in terms of memory consumption? Is it effecting the replicas? is it multiplied by the number of clients and the size of the batch?

This is just the default value. In most cases, clientservice sets the max_reply_size based on the requests it receives so there should be almost no change.

teoparvanov (approved):
LGTM, thanks @nkgeorgiev !

PR #2824, opened 2022-10-06
Change INFO log to DEBUG log

situ-s:
As isCollectingState function is called multiple times, the INFO log was spamming the concord logs. Hence changed it to debug logs.


teoparvanov (approved):
Thanks for addressing this quickly, @situ-s !

PR #2825, opened 2022-10-06
Improve libutt new api

yontyon:
* **Problem Overview**  
  The new libutt implementation used forward declarations to hide the original libutt usage. however, it means we need to include the original libutt headers on usage. In this PR we moved all the implementation to follow PImpl DP such that the whole implementation is completely separated from the interfaces themselves.
There are a few cases where a friend classed has to know the implementation details (and it is too much work, for now, to completely separate them), in these cases, we return void* that includes the necessary implementation details. This void* getter is not meant to be used outside the library (and its tests).
* **Testing Done**  
  CI


teoparvanov (changes requested):
Hi @yontyon , I have concerns with this refactoring (both content-wise, and timing-wise). 

Contrary to what the description suggests, it does other (unsafe) changes than just migrating towards a pimpl idiom. If my understanding is correct, those changes are needed because you cannot rely on friend access between classes anymore. So you're replacing one issue (the need to include libutt headers) with another, possibly worse one (unsafe memory access via `void*`).

My suggestion is to postpone this refactoring to when we have something working E2E. Also, it needs to be better thought out and maybe also agreed upon within the team.

yontyon:
@teoparvanov , it is not crucial by any means, it is only to start "cleaning" the code and make it more robust and easy to use.
As for the `void*` issue, it can be used only internally by the friend classes so there is no danger for anyone misusing it.
Anyway, I believe that completely separating the classes will require a significant amount of work (including changing dependencies in the original libutt implementation). So, my feeling is that this is the kind of stuff that if we delay them, they will either grow worse (as more and more things will be depended on them), or we won't have the time to handle it later...

PR #2827, opened 2022-10-06
Add support for deployment and registration of users through the sample wallet cli app

evdzhurov:
* **Problem Overview**  
  To achieve the first e2e ethereum privacy flows of deploying a privacy app and registering a user we need support from the UTT enabled sample wallet app. Usually, these operations should be performed by an administrator (operator) or with the assistance of an administrator, but initially (for simplicity) we allow the wallet app to do the operations directly.
* A few pre-generated RSA keys have been added for convenience in a test implementation of IUserPKInterface and used in the wallet-cli app until we have a more complete solution for key generation.
* **Testing Done**  
  Testing has been done using a wallet service and privacy contracts which are not available currently in the concord-bft repository.


teoparvanov (approved):
Looks great, thanks @evdzhurov !

PR #2828, opened 2022-10-11
Memory management dev1 pr1

glevkovich:
* **Problem Overview**  
A group of fixes for some memory leaks we've detected using heaptrack/ASAN.
* **Testing Done**  
All Apollo tests pass, and leaks do not reproduce after fix is applied.

### DIFF
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -475,6 +475,7 @@ void BCStateTran::stopRunningImpl() {
   ConcordAssertNE(replicaForStateTransfer_, nullptr);
   // This one should always be first!
   running_ = false;
+  clearIoContexts();
   if (postProcessingQ_) {
     postProcessingQ_->stop();
   }
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -273,8 +273,8 @@ BCStateTran::BCStateTran(const Config &config, IAppState *const stateApi, DataSt
       postponedSendFetchBlocksMsg_(false),
       ioPool_(
           config_.maxNumberOfChunksInBatch,
-          nullptr,                                     // alloc callback
-          [&](std::shared_ptr<BlockIOContext> &ctx) {  // free callback
+          nullptr,                                           // alloc callback
+          [&](const std::shared_ptr<BlockIOContext> &ctx) {  // free callback
             if (ctx->future.valid()) {
               try {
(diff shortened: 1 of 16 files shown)


eyalrund:
please run maestro run (5 trading days) with these changes

titusen (approved):
LGTM, my changes are included.

eyalrund (approved):

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -273,8 +273,8 @@ BCStateTran::BCStateTran(const Config &config, IAppState *const stateApi, DataSt
       postponedSendFetchBlocksMsg_(false),
       ioPool_(
           config_.maxNumberOfChunksInBatch,
-          nullptr,                                     // alloc callback
-          [&](std::shared_ptr<BlockIOContext> &ctx) {  // free callback
+          nullptr,                                           // alloc callback
+          [&](const std::shared_ptr<BlockIOContext> &ctx) {  // free callback
```
hshomroni, line 277: how does const affect mem leaks?

threshsign/include/threshsign/bls/relic/Library.h
```diff
@@ -69,14 +69,12 @@ class LibraryInitializer {
 
 class Library {
  public:
-  static Library* GetPtr() {
+  static const Library& Get() {
     // C++11 guarantees this is thread-safe (but all threads will share this variable!)
-    static Library* lib = new Library();
+    static const Library lib;
```
hshomroni, line 74: js moving the Library object from the heap to the stack done on purpose?

glevkovich: Yes, it's a small object, its size is known at compile time, and there is no real reason to allocate it on the heap at first.

util/include/SimpleMemoryPool.hpp
```diff
@@ -51,6 +51,17 @@ class SimpleMemoryPool {
     }
   }
 
+  virtual ~SimpleMemoryPool() {
+    if (freeCallback_) {
```
hshomroni, line 55: what happens if there is no freeCallback_? is it even possible?

glevkovich: yes, it's possible if the user decided not to define it (and not define allocCallback). We don't check if one is defined and another one is not, it's the user's responsibility.
You can see examples where it happen in concord-bft/util/test/simple_memory_pool_test.cpp
It can happen if the pool manages some kind of native types or fixed-size objects. It never happens in ST, which is currently the only user of this util (as far as I know) and needs allocation and free callbacks.

hshomroni: looking forward isn't it bug prone? allowing the user to leak memory if they don't correctly define the callback?

hshomroni (approved):

PR #2829, opened 2022-10-11
Remove unnecessary namespace alias

StefanPalashev:
* **Problem Overview** 
The filesystem is now part of the standard library. It should no longer be referred as an experimental feature.
* **Testing Done**
Tested with local runs, where the alias was not used anymore.


nkgeorgiev (approved):

teoparvanov (approved):
LGTM!

PR #2830, opened 2022-10-11
Add an option to build wallet-cli in utt CMakeLists.txt

shgandhi:
This commit adds an option build wallet-cli, since it may only be built for special-purpose.

### DIFF
--- a/utt/CMakeLists.txt
+++ b/utt/CMakeLists.txt
@@ -35,6 +35,12 @@ if(${CURVE} STREQUAL "BN128")
   )
 endif()
 
+option(
+  BUILD_WALLET_CLI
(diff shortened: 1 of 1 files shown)


evdzhurov:
> Hi @shgandhi, is the build of the wallet significantly slowing down the overall concord-bft build process? Maybe it's a good idea to discuss with @evdzhurov to better understand the issue at hand.

I approve it. There's a compilation issue not present in the open source. The wallet cli app can be built on demand and in a way that does not hit the aforementioned compilation issue, so this change will not impact our work.

evdzhurov (approved):

rastogiasr (approved):

PR #2831, opened 2022-10-12
Move log to INFO.

HristoStaykov:
Move Unsubscribe event to INFO. We don't expect this event to be very frequent.

* **Problem Overview**  
  The unsubscription event can be a valuable information and should not be too frequent.
* **Testing Done**  
  N/A


yuliasherman:
Current TRC behavior is to connect/disconnect every 3 min in case there are no new blocks. I am afraid that this log will fill our log files.

glevkovich (approved):
LGTM!

PR #2832, opened 2022-10-12
BC-23627 : TLS Client Key Rotation Fix 

anshubit5516:
Problem Description :
BC-23627 : [TLS Client Key Rotation] Daml ledger api restarts after Client key exchange
When client key exchange request is fired via operator, its observed that daml ledge api restarts after client key exchange.

Fix: 
  After reproducing the issue its observed that client service is not able to connect to Thin replica server. After key exchange, client and replica nodes changes certificates, but thin replica server is still running on old certificates, as no notification is sent. Current grpc implementation doesn't allow dynamic loading of certificates, hence servers needs to shutdown and then restart again.
To fix this, we have registered a callback to restart TRS if any key-exchange event occurs.

* **Testing Done**  
This fix is tested on castor deployment via IMAGE_TAG: 0.0.1.0.1177. we simulated the scenario and found that after key exchange, TRS is successfully restarted. Also ran DAML sanity.
Replica Logs attached for reference. 
(2022-10-12T12:22:58,832,+0000|INFO ||concord.thin_replica||||||main.cpp:670|Calling Shutdown to Thin Replica Server)
[tls_rotation_client_fix_image_1177.txt](https://github.com/vmware/concord-bft/files/9766693/tls_rotation_client_fix_image_1177.txt)

### DIFF
--- a/communication/include/communication/StateControl.hpp
+++ b/communication/include/communication/StateControl.hpp
@@ -58,12 +58,16 @@ class StateControl {
   }
 
   const std::string getEventTypeAsString(const EventType& et) {
+    std::string str;
     switch (et) {
       case EventType::TLS_COMM:
-        return "TLS_COMM";
+        str = "TLS_COMM";
+        break;
       case EventType::THIN_REPLICA_SERVER:
-        return "THIN_REPLICA_SERVER";
+        str = "THIN_REPLICA_SERVER";
+        break;
     }
+    return str;
   }
   void setGetPeerPubKeyMethod(std::function<std::string(uint32_t)> m) { get_peer_pub_key_ = std::move(m); }
 
--- a/communication/include/communication/StateControl.hpp
+++ b/communication/include/communication/StateControl.hpp
@@ -28,7 +28,7 @@ struct EnumHash {
 
 class StateControl {
  public:
- using CallbackRegistry = concord::util::CallbackRegistry<uint32_t>;
+  using CallbackRegistry = concord::util::CallbackRegistry<uint32_t>;
   enum class EventType { TLS_COMM, THIN_REPLICA_SERVER };
 
   static StateControl& instance() {
@@ -81,9 +81,8 @@ class StateControl {
   }
 
   std::mutex lock_comm_;
-  //keeping function template to be void(uint32_t) for uniformity
-  std::unordered_map<const EventType, std::unique_ptr<CallbackRegistry>, EnumHash>
-      event_registry_;
+  // keeping function template to be void(uint32_t) for uniformity
+  std::unordered_map<const EventType, std::unique_ptr<CallbackRegistry>, EnumHash> event_registry_;
   std::function<std::string(uint32_t)> get_peer_pub_key_;
 };
 }  // namespace bft::communication
--- a/communication/include/communication/StateControl.hpp
+++ b/communication/include/communication/StateControl.hpp
@@ -21,6 +21,7 @@ class StateControl {
     static StateControl instance_;
     return instance_;
   }
+
   void lockComm() { lock_comm_.lock(); }
   bool tryLockComm() { return lock_comm_.try_lock(); }
   void unlockComm() { lock_comm_.unlock(); }
@@ -30,11 +31,15 @@ class StateControl {
   }
 
   void restartComm(uint32_t id) { comm_restart_cb_registry_.invokeAll(id); }
+
+  void restartThinReplicaServer(uint32_t id) {restartComm(id);}
+
   void setGetPeerPubKeyMethod(std::function<std::string(uint32_t)> m) { get_peer_pub_key_ = std::move(m); }
   std::string getPeerPubKey(uint32_t id) {
     if (get_peer_pub_key_) return get_peer_pub_key_(id);
     return std::string();
   }
+
 
  private:
   std::mutex lock_comm_;
--- a/communication/include/communication/StateControl.hpp
+++ b/communication/include/communication/StateControl.hpp
@@ -17,6 +17,7 @@
 #include <unordered_map>
 #include <variant>
 #include <memory>
+#include "Logger.hpp"
 
 namespace bft::communication {
 struct EnumHash {
@@ -55,6 +56,16 @@ class StateControl {
     invokeCallback(et, id);
   }
 
+  const std::string getEventTypeAsString(const EventType& et) {
+    switch (et) {
+      case EventType::TLS_COMM:
+        return "TLS_COMM";
+      case EventType::THIN_REPLICA_SERVER:
+        return "THIN_REPLICA_SERVER";
+      default:
+        return "Invalid Type";
+    }
+  }
   void setGetPeerPubKeyMethod(std::function<std::string(uint32_t)> m) { get_peer_pub_key_ = std::move(m); }
 
   std::string getPeerPubKey(uint32_t id) {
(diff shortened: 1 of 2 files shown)


communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,89 @@
...
+  }
   void setGetPeerPubKeyMethod(std::function<std::string(uint32_t)> m) { get_peer_pub_key_ = std::move(m); }
+
   std::string getPeerPubKey(uint32_t id) {
     if (get_peer_pub_key_) return get_peer_pub_key_(id);
     return std::string();
   }
 
  private:
+  StateControl() : logger_(logging::getLogger("concord-bft.comm.StetControl")) {}
```
glevkovich, line 77: logger names should be lowcase in my opinion, and there is typo here as well:

`concord-bft.comm.state_control`

communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,89 @@
...
+
+  void registerCallback(const EventType& et, std::function<void(uint32_t)> cb) {
+    if (cb != nullptr) {
```
glevkovich, line 80: what is the use case for a null callback passed?
If there is a case can you please shortly explain?
If not use-case, you should assert here instead.

communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,89 @@
...
+      if (event_registry_.find(et) == event_registry_.end()) {
```
glevkovich, line 81: Trying to follow logic here, callbacks under the EnumHash  are gathered into the same callback registry, right?
So you are maintaining multiple callback registries under an unordered map?

communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,89 @@
...
+      }
+      event_registry_[et]->add(cb);
+    }
+  }
+
+  void invokeCallback(const EventType& et, uint32_t id) {
+    if (event_registry_.find(et) != event_registry_.end()) {
+      event_registry_[et]->invokeAll(id);
+    } else {
+      LOG_WARN(logger_, "No callback(s) registered for eventType " << getEventTypeAsString(et));
```
glevkovich, line 92: Can this happen? and if yes (even in rare configurations?) you should think how often this log will print, might it produce too many entries?

communication/src/AsyncTlsConnection.cpp
```diff
@@ -469,6 +469,7 @@ std::pair<bool, NodeNum> AsyncTlsConnection::checkCertificate(X509* received_cer
   out.close();
   BIO_free(outbio);
   LOG_INFO(logger_, "new certificate has been updated on local storage, peer: " << peerId);
+  bft::communication::StateControl::instance().restartThinReplicaServer(0);
```
glevkovich, line 472: How does '0' translates into invoking all callbacks for TRS?

glevkovich (changes requested):

communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,89 @@
 #include <functional>
 #include <utility>
 #include "callback_registry.hpp"
+#include <unordered_map>
```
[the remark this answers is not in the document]
anshubit5516: Will definitely keep in mind next time.

communication/include/communication/StateControl.hpp
anshubit5516: In live scenario, else case won't run. I had to add it for the current apollo test case. In that test case, client key exchange is triggered, hence tjis callback is called, since the callback is registered in ThinReplicaServer which is not in scope of this test case. Hence, I have added as WARN. but if u suggest, i can remove the else case itself.

communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,89 @@
...
-  void setCommRestartCallBack(std::function<void(uint32_t)> cb) {
-    if (cb != nullptr) comm_restart_cb_registry_.add(cb);
+  // Added/retained these methods to act as facade
+  void setCommRestartCallBack(std::function<void(uint32_t)> cb, const EventType& et = EventType::TLS_COMM) {
+    registerCallback(et, cb);
+  }
+
+  void restartComm(uint32_t id, const EventType& et = EventType::TLS_COMM) { invokeCallback(et, id); }
+
+  void setTlsRestartCallBack(std::function<void(uint32_t)> cb, const EventType& et = EventType::THIN_REPLICA_SERVER) {
```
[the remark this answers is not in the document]
anshubit5516: This call (setTlsRestartCallBack) is called by thin replica server.

communication/include/communication/StateControl.hpp
anshubit5516: What i am trying is creating new object of CallbackRegistry for each type of event. So currently we know two types, hence two objects will be there. Now each object will maintain a list of callbacks, hence we successfully segregate events and its callbacks.

communication/src/AsyncTlsConnection.cpp
anshubit5516: Its just added for maintaining the function signature. if we want multiple signatures, we can use variant and remove this unnecessary 0. Adding variant will unnecessary add complexity, hence didn't add. Please suggest your oipnion.

communication/include/communication/StateControl.hpp
anshubit5516: added assert

communication/include/communication/StateControl.hpp
anshubit5516: updated as suggested.

communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,89 @@
...
+  }
+
+  const std::string getEventTypeAsString(const EventType& et) {
+    switch (et) {
+      case EventType::TLS_COMM:
+        return "TLS_COMM";
+      case EventType::THIN_REPLICA_SERVER:
+        return "THIN_REPLICA_SERVER";
+      default:
+        return "Invalid Type";
```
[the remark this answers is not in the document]
anshubit5516: removed the default case as it will never occur.

communication/include/communication/StateControl.hpp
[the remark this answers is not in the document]
anshubit5516: My bad, i will update the methods name.

yuliasherman (approved):
LGTM

communication/include/communication/StateControl.hpp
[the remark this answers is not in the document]
glevkovich: @anshubit5516  I see that you did delete it....it was not used by anyone..

communication/include/communication/StateControl.hpp
```diff
@@ -14,31 +14,91 @@
...
   std::string getPeerPubKey(uint32_t id) {
     if (get_peer_pub_key_) return get_peer_pub_key_(id);
     return std::string();
   }
 
  private:
+  StateControl() : logger_(logging::getLogger("concord-bft.comm.state_control")) {}
+
+  void registerCallback(const EventType& et, std::function<void(uint32_t)>&& cb) {
+    ConcordAssert(cb != nullptr);
```
glevkovich, line 83: nit: just for the future, better use:
`ConcordAssertNE(cb, nullptr);`

glevkovich (approved):
LGTM, left small comment but you can fix it only if you fell like, not really mandatory.

cjn (commented):
LGTM. Thanks @anshubit5516

cjn (approved):

PR #2833, opened 2022-10-17
Revert "Replace thread_safe_queue with spsc_queue b/w TRC and Event S…

shgandhi:
During client rehydration we are seeing events being lost at the queue between the TRC and event service. The queue between the TRC and event service is the last mile stop before the updates are served to a secondary client consuming the gRPC stream from the event service. If for some reason, this secondary client is slow in reading updates from the gRPC stream, the queue will get backlogged, and may overflow.
Note, in happy conditions, the queue typically doesn't observe more than 17 events per seconds, i.e., why the queue size was set to 10k, i.e. 588X more than the number of updates it is actually meant to hold per second. Increasing the queue size further means postponing the problem further. This revert is a shortcut solution that will relieve the problem of lost events, but may manifest as a performance issue, in some other form.

This change will require end-to-end testing to reproduce the issue, this is currently being done in parallel.


teoparvanov:
Hi @shgandhi , can you add a more meaningful description, explaining why you're reverting this, and how it has been/will be tested?

shgandhi:
> Hi @shgandhi , can you add a more meaningful description, explaining why you're reverting this, and how it has been/will be tested?
> 
> Let's use the format the team has agreed upon:
> 
> * **Problem Overview**
>   ...
> * **Testing Done**
>   ...

@teoparvanov thanks, I have updated my PR description.

teoparvanov:
Thanks @shgandhi , a good description helps a lot when reviewing the code!

yuliasherman:
@shgandhi The memory amount used by TRC should be limited in any case. Some kind of a back pressure mechanism has to be implemented here.

HristoStaykov:
Based on the LRT with interruptions we performed, the changes introduced show significant improvement of the stability of the TPS in the system. I think until introducing the back pressure handling mechanism we should apply this change.

HristoStaykov (approved):

PR #2834, opened 2022-10-17
IStateTransfe change to unqiue_ptr

titusen:
Replica.cpp is leaking memory due to undeleted memory.
Change from "*" pointer to "unique_ptr"

apollo tests passed

### DIFF
--- a/kvbc/include/Replica.h
+++ b/kvbc/include/Replica.h
@@ -243,7 +243,7 @@ class Replica : public IReplica,
   const bftEngine::ReplicaConfig &replicaConfig_;
   bftEngine::IReplica::IReplicaPtr m_replicaPtr = nullptr;
   std::shared_ptr<ICommandsHandler> m_cmdHandler = nullptr;
-  bftEngine::IStateTransfer *m_stateTransfer = nullptr;
+  std::unique_ptr<bftEngine::IStateTransfer> m_stateTransfer;
   std::unique_ptr<concord::storage::DBMetadataStorage> m_metadataStorage;
   std::unique_ptr<ReplicaStateSync> replicaStateSync_;
   std::shared_ptr<concordMetrics::Aggregator> aggregator_;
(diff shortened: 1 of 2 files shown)


kvbc/include/Replica.h
```diff
@@ -243,7 +243,7 @@ class Replica : public IReplica,
   const bftEngine::ReplicaConfig &replicaConfig_;
   bftEngine::IReplica::IReplicaPtr m_replicaPtr = nullptr;
   std::shared_ptr<ICommandsHandler> m_cmdHandler = nullptr;
-  bftEngine::IStateTransfer *m_stateTransfer = nullptr;
+  std::unique_ptr<bftEngine::IStateTransfer> m_stateTransfer;
```
glevkovich, line 246: - The header of the PR looks wrong - you are replacing State transfer pointer with a raw pointer.
- Why are you merging into master? we still have a dev branch.

kvbc/src/Replica.cpp
```diff
@@ -600,7 +600,7 @@ Replica::Replica(ICommunication *comm,
   m_dbSet.dataDBClient->setAggregator(aggregator_);
   m_dbSet.metadataDBClient->setAggregator(aggregator_);
   auto stKeyManipulator = std::shared_ptr<storage::ISTKeyManipulator>{storageFactory->newSTKeyManipulator()};
-  m_stateTransfer = bftEngine::bcst::create(stConfig, this, m_metadataDBClient, stKeyManipulator, aggregator_);
+  m_stateTransfer.reset(bftEngine::bcst::create(stConfig, this, m_metadataDBClient, stKeyManipulator, aggregator_));
```
glevkovich, line 603: So it looks like ST was nver called since we never called it's ctor, right?

glevkovich (changes requested):

kvbc/src/Replica.cpp
titusen: it is called later with "*m_stateTransfer", so the syntax persists.

kvbc/include/Replica.h
glevkovich: > No, other way round "change **to** unique_ptr"

@titusen 
Suggested header for the PR:
"Convert m_stateTransfer to a unqiue_ptr"

PR #2835, opened 2022-10-20
Fix a race in PreProcessor between release of expired requests and preeparation of the PreProcessReplyMsg message

yuliasherman:
* **Problem Overview**  
There was recently observed a new type of core dump on a non-primary replica. It was caused by a race condition between two async processes: preparation of the PreProcessReplyMsg to be sent back to the primary replica and cancellation of the expired request. As the memory used for the pre-executed result hash creation is allocated from the raw memory pool, it gets released after request cancellation and we have a crash trying to access it. The solution is to avoid the cancellation of requests that are in the middle of sending replies.
* **Testing Done**  
Regression test using Maestro.


HristoStaykov (approved):
LGTM, thanks, Yulia!

teoparvanov (approved):
LGTM, thanks Yulia! The test coverage looks good, thanks to the thorough Apollo tests for the pre-processor, as well as the Maestro runs you've completed.

PR #2838, opened 2022-10-21
Measure the number of storage reads made via TRS and KvbAppFilter

shgandhi:
This PR adds metrics in TRS and a counter in KvbAppFilter to calculate the number of storage reads made by the components. These metrics are required to monitor storage read activity for performance analysis.


kvbc/src/kvbc_app_filter/kvbc_app_filter.cpp
```diff
@@ -263,8 +264,12 @@ void KvbAppFilter::readBlockRange(BlockId block_id_start,
                                   spsc_queue<KvbFilteredUpdate> &queue_out,
                                   const std::atomic_bool &stop_execution) {
   if (block_id_start > block_id_end || block_id_end > rostorage_->getLastBlockId()) {
+    if (block_id_start <= block_id_end) {
+      num_storage_reads++;
+    }
     throw InvalidBlockRange(block_id_start, block_id_end);
   }
+  num_storage_reads++;
```
HristoStaykov, line 273: Why do we increment twice only? Shouldn't we increment with `block_id_end - block_id_start` ?

HristoStaykov (approved):
LGTM, thanks, @shgandhi ! I only have 1 question regarding the update of the counter in readBlockRangeHash.

PR #2840, opened 2022-10-28
Linking crypto and opentracing statically in util and threshsign library

bandatarunkumar:
* **Problem Overview**  
Due to dynamic linking of openssl(libcrypto.so), the components which use these modules(threhsign and util) will also need a dynamic linking to libcrypto.so. We have encountered a case where our internal standalone tool is depending on threhsign and util library, which make the tool to dependant on libcrypto.so dynamically, due to which the standalone behaviour of the tool is no more standalone. So to avoid this behaviour on internal tool, we are linking crypto as a static linking, which make the internal tool standalone.
* **Testing Done**  
  Tested internally by generating configs and doing deployment


pkthapa (approved):
LGTM!

cjn (approved):
LGTM. Please update the description.

threshsign/src/CMakeLists.txt
```diff
@@ -10,13 +10,16 @@ set(common_source_files
 add_library(threshsign STATIC ${common_source_files})
 target_include_directories(threshsign PUBLIC ../include)
 target_link_libraries(threshsign PUBLIC Threads::Threads)
+find_library(CRYPTO_STATIC_LIBRARY NAMES libcrypto.a PATHS ${OPENSSL_ROOT_DIR}/lib/ REQUIRED)
```
toly-kournik, line 13: use `find_package(OpenSSL)` there should be a variable populated for static library

util/CMakeLists.txt
```diff
@@ -22,8 +22,10 @@ if(USE_OPENSSL)
     if(NOT BUILD_THIRDPARTY)
         find_package(OpenSSL REQUIRED)
     endif()
+    find_library(CRYPTO_STATIC_LIBRARY NAMES libcrypto.a HINTS ${OPENSSL_ROOT_DIR}/lib/  REQUIRED)
```
toly-kournik, line 25: same

toly-kournik:
Please describe a problem

toly-kournik (changes requested):

toly-kournik:
https://cmake.org/cmake/help/latest/module/FindOpenSSL.html

toly-kournik:
Something like: 
```
set (OPENSSL_USE_STATIC_LIBS TRUE)
find_package(OpenSSL COMPONENTS Crypto  REQUIRED)
...
target_link_libraries(<target_name> OpenSSL::Crypto)
```

util/CMakeLists.txt
bandatarunkumar: @toly-kournik , thanks for sharing "OPENSSL_USE_STATIC_LIBS", this makes us to link static even it UTT.

threshsign/src/CMakeLists.txt
bandatarunkumar: done

toly-kournik (approved):
LGTM!

PR #2841, opened 2022-11-02
Memory management dev1 pr2

glevkovich:
[don’t auto-bump]

*Problem Overview*
This is the second group of commits related to memory management improvements done in the background.
Here we:
1) Fix two more memory leaks.
2) Convert some raw pointers to smart pointers.
3) Revert a wrong commit that converted a raw pointer (in a module that is not the owner of the pointer) to a smart pointer. This caused double-free on replica shutdown.
4) Add debug logs while shutting down, which help to understand which modules were destroyed.
5) Add names to a few types of memory pools. This helps to understand replica creation, shutdown, and resource consumption. Every pool created must have a name, which is printed on the log.

*Testing is done*
Ran on Maestro, leaks are gone. All apollo tests pass. No particular new tests have been added since no functionality has changed.


glevkovich:
don't auto-bump

Efrat1 (approved):
lgtm, my changes are in

eyalrund (approved):

yuliasherman (approved):
LGTM

PR #2844, opened 2022-11-08
Rvt reports invalid tree bug fix

glevkovich:
* **Problem Overview**  
Found CRITICAL bug in angeValidationTree::validate. All validations returned always true. In addition, the behavior of the new openSSL integer is slightly different, had to adapt by converting a NodeVal to positive on some operations due to failures on validation.
in addition, I've improved the way values are printed in this module by overloading operator<<  and more.

* **Testing Done**  
all ST unittests pass after change, all Apollo tests pass. Added a new NodeValBasicOperators test in RVT_tests to check all new Integer operations.


mdarade (approved):
Looks good.
Thanks Gil.

eyalrund (approved):

PR #2846, opened 2022-11-09
Allow tx without budget

yontyon:
* **Problem Overview**  
  Allow a non-budgeted transaction. 
The servers validate the budget coin's presence according to a configuration. The client should create the transaction with budget coin also according to the same configuration.
* **Testing Done**  
  unitests in libutt and client api

### DIFF
--- a/utt/admin-cli/include/admin.hpp
+++ b/utt/admin-cli/include/admin.hpp
@@ -33,7 +33,7 @@ class Admin {
   /// [TODO-UTT] Should be performed by an admin app
   /// @brief Deploy a privacy application
   /// @return The public configuration of the deployed application
-  static bool deployApp(Channel& chan, bool budget_policy = true);
+  static bool deployApp(Channel& chan, bool budget_policy);
 
   /// @brief Request the creation of a privacy budget. The amount of the budget is predetermined by the deployed app.
   /// This operation could be performed entirely by an administrator, but we add it in the admin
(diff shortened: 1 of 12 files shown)


utt/admin-cli/include/admin.hpp
```diff
@@ -33,7 +33,7 @@ class Admin {
   /// [TODO-UTT] Should be performed by an admin app
   /// @brief Deploy a privacy application
   /// @return The public configuration of the deployed application
-  static bool deployApp(Channel& chan, bool budget_policy = true);
+  static bool deployApp(Channel& chan, bool budget_policy);
```
cloudnoize, line 36: Should this PR also include a change to the wallet app?
as it checks 
```
const size_t budget = getPrivacyBudget();
  if (budget < amount) throw std::runtime_error("User has insufficient privacy budget!");```

cloudnoize: Isn't it more elegant to define the non-budget configuration as a budget coin with infinite value?
It will eliminate the need for `if - else` on the transaction and will reduce the configuration changes

evdzhurov (commented):
I don't think the option to require a budget should be called 'budget_policy', as a budget policy may include more logic than simply turning the budget requirement on or off. Maybe something like 'require_budget' is a bit more appropriate. Also, I don't think we're ready yet to expose this option in the admin when creating a deployment, there needs to be more discussion how this should actually work and what changes will be needed in the wallets as well. For now we can limit ourselves to add a unit test that validates turning the budget requirement off.

utt/admin-cli/include/admin.hpp
evdzhurov: It's more elegant for the code but will it be more elegant for the system if we still need to create and manage a 'hidden infinite' budget coin? And how does it reduce the configuration changes?

PR #2847, opened 2022-11-10
Replica imp and state transfer  bug fixes

glevkovich:
* **Problem Overview**  
Fixing a critical and minor bug  here. The critical bug is caused by a race condition between consensus thread and St thread, leading to a corruption of State Transfer state and for replica to stay out of the network forever.
Here we introduce state transfer "external state" in order to solve this issue once and for all.

* **Testing Done**  
All regular tests pass.
Running State Transfer for 48 hours every 2 minutes on high end deployment over 700 TPS. checking that State Transfer always works and there are no race conditions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -297,6 +297,7 @@ test-range: ## Run all tests in the range [START,END], inclusive: `make test-ran
 .PHONY: test-single-suite
 test-single-suite: SHELL:=/bin/bash
 test-single-suite: ## Run a single test `make test-single-suite TEST_NAME=<test name> NUM_REPEATS=<number of repeats,default=1,optional> BREAK_ON_FAILURE=<TRUE|FALSE,optional>`. Example: `make test-single-suite TEST_NAME=timers_tests BREAK_ON_FAILURE=TRUE NUM_REPEATS=3`
+	@if [ -z ${TEST_NAME} ]; then echo "Error: please set the TEST_NAME environment variable!"; exit 1; fi
 	num_failures=0; \
 	for (( i=1; i<=${NUM_REPEATS__}; i++ )); do \
 		echo "=== Starting iteration $${i}/${NUM_REPEATS__}"; \
@@ -313,8 +314,8 @@ test-single-suite: ## Run a single test `make test-single-suite TEST_NAME=<test 
 
 .PHONY: test-single-apollo-case
 test-single-apollo-case: ## Run a single Apollo test case: `make test-single-apollo-case TEST_FILE_NAME=<test file name> TEST_CASE_NAME=<test case name> NUM_REPEATS=<number of repeats,default=1,optional> BREAK_ON_FAILURE=<TRUE|FALSE,optional>`. Test suite file name should come without *.py. Test case is expected without a class name, and must be unique. Example: `make test-single-apollo-case BREAK_ON_FAILURE=TRUE NUM_REPEATS=100 TEST_FILE_NAME=test_skvbc_reconfiguration TEST_CASE_NAME=test_tls_exchange_client_replica_with_st`
-	@if [ -z ${TEST_FILE_NAME} ]; then echo "Error: TEST_FILE_NAME is mandatory"; exit 1; fi
-	@if [ -z ${TEST_CASE_NAME} ]; then echo "Error: TEST_CASE_NAME is mandatory"; exit 1; fi
+	@if [ -z ${TEST_FILE_NAME} ]; then echo "Error: please set the TEST_FILE_NAME environment variable!"; exit 1; fi
+	@if [ -z ${TEST_CASE_NAME} ]; then echo "Error: please set the TEST_CASE_NAME environment variable!"; exit 1; fi
 	$(eval PREFIX := $(shell docker run ${BASIC_RUN_PARAMS} \
 		${CONCORD_BFT_CONTAINER_SHELL} -c \
 		"mkdir -p ${CONCORD_BFT_CORE_DIR} && \
@@ -325,8 +326,8 @@ test-single-apollo-case: ## Run a single Apollo test case: `make test-single-apo
 		"python3 scripts/apollo_list_tests.py \
 		${CONCORD_BFT_TARGET_SOURCE_PATH}/tests/apollo/ f | grep ${TEST_FILE_NAME} | grep -w ${TEST_CASE_NAME}"))
 	@if [ -z "${PREFIX}" ] || [ -z "${POSTFIX}" ]; then \
-		echo "Error: Failed to start test, check if TEST_FILE_NAME=${TEST_FILE_NAME}" \
-			"or TEST_CASE_NAME=${TEST_CASE_NAME} exist."; exit 1;\
+		echo "Error: Failed to start test, please check if TEST_FILE_NAME=${TEST_FILE_NAME}" \
+			"or TEST_CASE_NAME=${TEST_CASE_NAME} environment variables exist!"; exit 1;\
 	fi
 	@docker run ${BASIC_RUN_PARAMS} \
 		${CONCORD_BFT_CONTAINER_SHELL} -c "cd tests/apollo/; \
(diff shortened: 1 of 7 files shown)


bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -917,12 +922,25 @@ void BCStateTran::startCollectingStateInternal() {
   startCollectingStateImpl();
 }
 
-void BCStateTran::startCollectingStateImpl() {
+void BCStateTran::setExternalCollectingState(bool newVal) {
+  if (newVal == externalCollectingState_) {
```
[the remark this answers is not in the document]
glevkovich: The logic changed few times so I removed it to prevent too much noise in log, then forgot to add.
With current logic I won't expect it to happen at all. 
Adding, thanks.

eyalrund (approved):

mdarade (approved):
Minor comment.
Thanks Gil.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -917,12 +922,26 @@ void BCStateTran::startCollectingStateInternal() {
...
```
yuliasherman, line 926: This function is not thread-safe. The fact that externalCollectingState_ is atomic does not help here, because there are multiple operations on it: 'if' and then replace. There could be a race in between.

Makefile
```diff
@@ -297,6 +297,7 @@ test-range: ## Run all tests in the range [START,END], inclusive: `make test-ran
 .PHONY: test-single-suite
 test-single-suite: SHELL:=/bin/bash
 test-single-suite: ## Run a single test `make test-single-suite TEST_NAME=<test name> NUM_REPEATS=<number of repeats,default=1,optional> BREAK_ON_FAILURE=<TRUE|FALSE,optional>`. Example: `make test-single-suite TEST_NAME=timers_tests BREAK_ON_FAILURE=TRUE NUM_REPEATS=3`
+	@if [ -z ${TEST_NAME} ]; then echo "Error: TEST_NAME is mandatory!"; exit 1; fi
```
WildFireFlum, line 300: A better message would be "Please set the TEST_NAME environment variable"

Makefile
```diff
@@ -313,8 +314,8 @@ test-single-suite: ## Run a single test `make test-single-suite TEST_NAME=<test
 
 .PHONY: test-single-apollo-case
 test-single-apollo-case: ## Run a single Apollo test case: `make test-single-apollo-case TEST_FILE_NAME=<test file name> TEST_CASE_NAME=<test case name> NUM_REPEATS=<number of repeats,default=1,optional> BREAK_ON_FAILURE=<TRUE|FALSE,optional>`. Test suite file name should come without *.py. Test case is expected without a class name, and must be unique. Example: `make test-single-apollo-case BREAK_ON_FAILURE=TRUE NUM_REPEATS=100 TEST_FILE_NAME=test_skvbc_reconfiguration TEST_CASE_NAME=test_tls_exchange_client_replica_with_st`
-	@if [ -z ${TEST_FILE_NAME} ]; then echo "Error: TEST_FILE_NAME is mandatory"; exit 1; fi
-	@if [ -z ${TEST_CASE_NAME} ]; then echo "Error: TEST_CASE_NAME is mandatory"; exit 1; fi
+	@if [ -z ${TEST_FILE_NAME} ]; then echo "Error: TEST_FILE_NAME is mandatory!"; exit 1; fi
```
WildFireFlum, line 317: same comment (for all of the "mandatory argument" prints).

bftengine/src/bcstatetransfer/BCStateTran.cpp
WildFireFlum: you can read the content externalCollectingState_ and use compare_exchange_strong.
Not sure if this should be done only for the prints though.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -917,12 +922,26 @@ void BCStateTran::startCollectingStateInternal() {
...
+  }
+  externalCollectingState_ = newVal;
+  LOG_INFO(logger_, std::boolalpha << KVLOG(externalCollectingState_));
+}
+
+void BCStateTran::startCollectingStateExternal() {
   ConcordAssert(running_);
-  if (psd_->getIsFetchingState()) {
+  if (isCollectingState()) {
     LOG_WARN(logger_, "Already in State Transfer, ignore call...");
```
WildFireFlum, line 937: What flows should cause multiple state transfer instances?
If there are none, shouldn't this throw?

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -917,12 +922,26 @@ void BCStateTran::startCollectingStateInternal() {
...
     return;
   }
+  startCollectingStateImpl();
+}
+
+void BCStateTran::startCollectingStateImpl() {
+  setExternalCollectingState(true);
```
WildFireFlum, line 944: This seems like a classic case for using RAII, have some StateCollectionGuard whose end of scope sets the flag to false.

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -167,6 +167,7 @@ class BCStateTran : public IStateTransfer {
                                  char* outStateDigest,
                                  char* outResPagesDigest,
                                  char* outRVBDataDigest);
+  void startCollectingStateExternal();
```
WildFireFlum, line 170: What does collectingstateexternal means?

this function should either be renamed or documented, especially because one line after we have a similarly named function.

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -766,6 +767,12 @@ class BCStateTran : public IStateTransfer {
 
   FetchingState lastFetchingState_;
 
+  // For speed, we choose to make the API call isCollectingState() non-blocked, and it doesn't pass through handoff
+  // queue. This flag represents ST basic state to the outside: is it currently collecting/not collecting. It is set in
+  // slightly different time points to ensure thread safety towards external modules and threads.
+  std::atomic_bool externalCollectingState_{false};
```
WildFireFlum, line 773: Why not simply name it isCollecting_?
The addition of "external" does not give additional information as there is no "internal" collection state.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4481,6 +4476,9 @@ ReplicaImp::ReplicaImp(bool firstTime,
         KeyExchangeManager::instance().sendInitialKey(this, lastExecutedSeqNum);
         LOG_INFO(GL, "Send initial key exchange after completing state transfer " << KVLOG(lastExecutedSeqNum));
       }
+    } else {
+      LOG_WARN(GL,
```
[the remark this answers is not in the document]
WildFireFlum: It can be written in a better way and maybe not start with the negated condition.
A replica should publish (one of) its public key(s) before other replicas can communicate with it, but this code suggests that it might have triggered state transfer before publishing its key. Once the replica returns to the quorum (meaning after it is done with state transfer) its key needs to be known to the other replicas.

WildFireFlum:
where exactly was the race?

WildFireFlum (changes requested):

situ-s (approved):
LGTM! Thanks for fixing it.

Makefile
```diff
@@ -325,17 +326,17 @@ test-single-apollo-case: ## Run a single Apollo test case: `make test-single-apo
 		"python3 scripts/apollo_list_tests.py \
 		${CONCORD_BFT_TARGET_SOURCE_PATH}/tests/apollo/ f | grep ${TEST_FILE_NAME} | grep -w ${TEST_CASE_NAME}"))
 	@if [ -z "${PREFIX}" ] || [ -z "${POSTFIX}" ]; then \
-		echo "Error: Failed to start test, check if TEST_FILE_NAME=${TEST_FILE_NAME}" \
-			"or TEST_CASE_NAME=${TEST_CASE_NAME} exist."; exit 1;\
+		echo "Error: Failed to start test, please check if TEST_FILE_NAME=${TEST_FILE_NAME}" \
+			"or TEST_CASE_NAME=${TEST_CASE_NAME} enviorment variables exist!"; exit 1;\
```
WildFireFlum, line 330: typo: environment

WildFireFlum (approved):

PR #2848, opened 2022-11-10
Draft: Pass messages as smart pointers E2E: phase one

yuliasherman:
* **Problem Overview**  
BFT messages are used as raw pointers in the concord-bft code, as a result, we have memory leaks. This code has to be improved by starting to use managed pointers instead. This PR changes the onMessage API to pass messages as smart pointers and have initial support for them in the PreProcessor module.
* **Testing Done**  
Runs on a reserved system with interruptions.


eyalrund:
we need to test on high end environment
need to add multiple reviewers

yuliasherman:
> we need to test on high end environment need to add multiple reviewers

It is not ready yet. I just want to verify it passes the CI.

PR #2851, opened 2022-11-11
Review comments addressed.

pkthapa:
* **Problem Overview**  
  Review comments resolution for https://github.com/vmware/concord-bft/pull/2806
* **Testing Done**  
  Local `make test`


kvbc/tools/object_store_utility/integrity_checker.cpp
```diff
@@ -75,15 +75,14 @@ std::pair<BlockId, Digest> IntegrityChecker::getLatestsCheckpointDescriptor() co
   LOG_INFO(logger_, "Last block: " << lastBlock);
   auto it = dynamic_cast<s3::Client*>(s3_dbset_.metadataDBClient.get())
                 ->getIterator<s3::Client::SortByModifiedDescIterator>();
-  auto [key, val] = it->seekAtLeast(Sliver::copy(checkpoints_prefix_.data(), checkpoints_prefix_.length()));
-  UNUSED(val);
+  const auto& kvdata = it->seekAtLeast(Sliver::copy(checkpoints_prefix_.data(), checkpoints_prefix_.length()));
```
glevkovich, line 78: not sure why did you removed the parameter pack?
Why not:
`const auto &[key, _] = it->seekAtLeast(Sliver::copy(checkpoints_prefix_.data(), checkpoints_prefix_.length()));`

I use '_' for the value since it is unused. Both are const references, and this form is more redadble.

kvbc/tools/object_store_utility/integrity_checker.cpp
```diff
@@ -75,15 +75,14 @@ std::pair<BlockId, Digest> IntegrityChecker::getLatestsCheckpointDescriptor() co
...
   if (it->isEnd()) throw std::runtime_error("no checkpoints information in S3 storage");
 
-  std::string suff = key.toString().substr(checkpoints_prefix_.length());
+  std::string suff = kvdata.first.toString().substr(checkpoints_prefix_.length());
```
glevkovich, line 81: Why creating a new string (allocations, copies)?
 use std::string_view

kvbc/tools/object_store_utility/integrity_checker.cpp
```diff
@@ -75,15 +75,14 @@ std::pair<BlockId, Digest> IntegrityChecker::getLatestsCheckpointDescriptor() co
...
   BlockId block = util::to<BlockId>(suff.substr(0, suff.find_last_of('/')));
-  LOG_INFO(logger_, "Latest checkpoint descriptor: " << key.toString() << " for block: " << block);
+  LOG_INFO(logger_, "Latest checkpoint descriptor: " << kvdata.first.toString() << " for block: " << block);
```
glevkovich, line 83: nit:
Use KVLOG macro wherever possible:
`LOG_INFO(logger_, "Latest checkpoint descriptor: " << kvdata.first.toString() << " for" << KVLOG(block);`

kvbc/tools/object_store_utility/integrity_checker.cpp
```diff
@@ -216,17 +215,14 @@ void IntegrityChecker::validateRange(const BlockId& until_block) const {
 void IntegrityChecker::validateKey(const std::string& key) const {
   // Retrieve the latest block number for specified key
   Sliver sKey = Sliver::copy(key.data(), key.length());
-  auto [containing_block, _] = s3_dbset_.dbAdapter->getValue(sKey, 0);
-  UNUSED(_);
-  auto containing_block_id = util::to<BlockId>(containing_block.data());
+  auto containing_block_id = util::to<BlockId>(s3_dbset_.dbAdapter->getValue(sKey, 0).first.data());
   LOG_INFO(logger_, "containing_block_id: " << containing_block_id);
```
glevkovich, line 219: `LOG_INFO(logger_, KVLOG(containing_block_id));`

kvbc/tools/object_store_utility/integrity_checker.cpp
```diff
@@ -216,17 +215,14 @@ void IntegrityChecker::validateRange(const BlockId& until_block) const {
 void IntegrityChecker::validateKey(const std::string& key) const {
   // Retrieve the latest block number for specified key
   Sliver sKey = Sliver::copy(key.data(), key.length());
-  auto [containing_block, _] = s3_dbset_.dbAdapter->getValue(sKey, 0);
-  UNUSED(_);
-  auto containing_block_id = util::to<BlockId>(containing_block.data());
+  auto containing_block_id = util::to<BlockId>(s3_dbset_.dbAdapter->getValue(sKey, 0).first.data());
```
glevkovich, line 218: It is a valid code, but parameter unpack is more appropriate in my opinion (more "mdoren C++" and makes code more readble). To handle the "ugly" UNUSED, do:
`[[maybe_unused]] auto [containing_block, _] = s3_dbset_.dbAdapter->getValue(sKey, 0);`
same for the rest of the code.

bottom line: i'm not sure why you removed all parameter pack and took it back to c++11 style, (although its a valid code).

threshsign/src/ThresholdSignaturesTypes.cpp
```diff
@@ -230,7 +230,7 @@ void Cryptosystem::validatePrivateKey(const std::string& key) const {
 }
 
 bool Cryptosystem::isValidCryptosystemSelection(const std::string& type, const std::string& subtype) {
-  UNUSED(subtype);
+  (void)subtype;
```
glevkovich, line 233: remove UNUSED(subtype);
In general, I would remoe this macro compeltely with C++17. maybe we should open a task if you see it being used so many times?
`bool Cryptosystem::isValidCryptosystemSelection(const std::string& type, [[maybe_unused]] const std::string& subtype) {
`

I won't comment more about UNUSED, please fix all occurences by using   [[maybe_unused]]

threshsign/src/eddsa/EdDSAMultisigVerifier.cpp
```diff
@@ -48,9 +48,8 @@ size_t EdDSASignatureAccumulator::getFullSignedData(char *outThreshSig, int thre
   ConcordAssertGE(static_cast<uint64_t>(threshSigLen),
                   static_cast<unsigned long>(signatures_.size()) * sizeof(SingleEdDSASignature));
   size_t offset = 0;
-  for (auto &[id, sig] : signatures_) {
-    UNUSED(id);
-    std::memcpy(outThreshSig + offset, &sig, sizeof(SingleEdDSASignature));
+  for (const auto &signature : signatures_) {
```
glevkovich, line 51: I wont comment more about this repeating issue, but I think parameter pack is the right choice, and you know already the way how to handle the UNUSED from my other comments. I would bring it back to how it was (with the const ref).

glevkovich (changes requested):
I can't go all over my prev. comments on another PR. but I remember one - What are we going to do about the very old openSSL? is there a task to upgrade?

kvbc/tools/object_store_utility/integrity_checker.cpp
pkthapa: gcc/g++ compilation fails for `[[maybe_unused]]`. Hence, not retaining parameter pack.

kvbc/tools/object_store_utility/integrity_checker.cpp
pkthapa: Done.

kvbc/tools/object_store_utility/integrity_checker.cpp
pkthapa: Done.

kvbc/tools/object_store_utility/integrity_checker.cpp
pkthapa: gcc/g++ compilation fails for `[[maybe_unused]]`. Hence, not retaining parameter pack.

threshsign/src/ThresholdSignaturesTypes.cpp
pkthapa: Done.

threshsign/src/eddsa/EdDSAMultisigVerifier.cpp
pkthapa: gcc/g++ compilation fails for `[[maybe_unused]]`. Hence, not retaining parameter pack.

pkthapa: Retaining UNUSED macro for below reason:
- `[[maybe_unused]]` could have been used instead of `UNUSED()`, but compilation with gcc/g++ fails (however clang++ compilation passes).
The g++ compiler version needs an upgrade to atleast 10 or 11. Current g++ version is v7.5.0. In current g++ version, `[[maybe_unused]]` is not recognised in cases such as range-based for loop (e.g. `for ([[maybe_unused]] const auto &[id, sig] : signatures_) {}`). In this example, if `id` is unused, g++ compiler throws an compilation error: "`error: unused variable 'id' [-Werror=unused-variable]`"

kvbc/tools/object_store_utility/integrity_checker.cpp
pkthapa: Not recommended to use `string_view`. Changing to `string_view` requires specialized templated function's parameter to be changed to `string_view` from `string`.
https://github.com/vmware/concord-bft/blob/d56ae0e6d8c488a1086105e24f7916fb32c86ec9/util/include/string.hpp#L57 

Inside the function, standard string conversion function is used (`std::stoul(s);`) which again asks for string as parameter. I see no point of doing conversion from `string_view` to `string` again at this point.

secretsmanager/src/aes.cpp
```diff
@@ -31,11 +30,11 @@ vector<uint8_t> AES_CBC::encrypt(const string& input) {
...
     }
 
-    unique_ptr<unsigned char[]> ciphertext(new unsigned char[input.size() + AES_BLOCK_SIZE]);
-    unique_ptr<unsigned char[]> plaintext(new unsigned char[input.size() + 1]);
+    auto ciphertext = std::make_unique<unsigned char[]>(input.size() + AES_BLOCK_SIZE);
+    auto plaintext = std::make_unique<unsigned char[]>(input.size() + 1);
 
     for (size_t i{0UL}; i < input.size(); ++i) {
-      plaintext.get()[i] = (unsigned char)input[i];
+      plaintext.get()[i] = static_cast<unsigned char>(input[i]);
```
glevkovich, line 37: 1) Not sure why plaintext is +1 ? std::string has no null terminator.
2) I suggest to change 36 to 38 to:
`std::copy(input.begin(), input.end(), plaintext.get());`

glevkovich (approved):
@pkthapa , approved with 1 comment.

I have 2 questions/suggestions:

1. Inability to fullky use [[ maybe unused ]] is just one example of many why we must upgrade our toolchains.
2. Do you know if anything was done to upgrade openssl version in container? It is very old, I've commented about it on previous PR.

secretsmanager/src/aes.cpp
pkthapa: Addressed in https://github.com/vmware/concord-bft/pull/2851/commits/d4b7969f8e0646cedea6cc7a8f9b76c8913ffee4

pkthapa:
> 

**OpenSSL upgrade**:
There is no any immediate action plan to upgrade openssl version. Also, we need to upgrade gcc/g++ version to atleast `10` or `11`. Current version is `v7.5`. gcc v7.5 does not work for cases like range-based for loop with `[[maybe_unused]]` .

pkthapa:
> @pkthapa , approved with 1 comment.
> 
> I have 2 questions/suggestions:
> 
> 1. Inability to fullky use [[ maybe unused ]] is just one example of many why we must upgrade our toolchains.
> 2. Do you know if anything was done to upgrade openssl version in container? It is very old, I've commented about it on previous PR.

**OpenSSL upgrade**:
There is no any immediate action plan to upgrade openssl version. Also, we need to upgrade gcc/g++ version to atleast `10` or `11`. Current version is `v7.5`. gcc v7.5 does not work for cases like range-based for loop with `[[maybe_unused]]` .

pkthapa:
> I can't go all over my prev. comments on another PR. but I remember one - What are we going to do about the very old openSSL? is there a task to upgrade?

https://github.com/vmware/concord-bft/pull/2851#issuecomment-1323047445

cjn:
LGTM. Thanks @pkthapa

cjn (approved):

secretsmanager/src/aes.cpp
```diff
@@ -21,22 +21,19 @@
...
+vector<uint8_t> AES_CBC::encrypt(std::string_view input) {
   if (algo_ == concord::crypto::SignatureAlgorithm::EdDSA) {
     if (input.empty()) {
       return {};
     }
 
-    unique_ptr<unsigned char[]> ciphertext(new unsigned char[input.size() + AES_BLOCK_SIZE]);
-    unique_ptr<unsigned char[]> plaintext(new unsigned char[input.size() + 1]);
+    auto ciphertext = std::make_unique<unsigned char[]>(input.size() + AES_BLOCK_SIZE);
+    auto plaintext = std::make_unique<unsigned char[]>(input.size());
```
WildFireFlum, line 34: You don't need this copy of the plaintext, just use the argument

secretsmanager/src/aes.cpp
```diff
@@ -21,22 +21,19 @@
 namespace concord::secretsmanager {
 using std::vector;
 using std::string;
-using std::unique_ptr;
 using concord::crypto::openssl::OPENSSL_SUCCESS;
 using concord::crypto::openssl::UniqueCipherContext;
 
-vector<uint8_t> AES_CBC::encrypt(const string& input) {
+vector<uint8_t> AES_CBC::encrypt(std::string_view input) {
   if (algo_ == concord::crypto::SignatureAlgorithm::EdDSA) {
```
WildFireFlum, line 28: No need for this check in this class

secretsmanager/src/aes.cpp
```diff
@@ -21,22 +21,19 @@
...
     if (input.empty()) {
       return {};
     }
 
-    unique_ptr<unsigned char[]> ciphertext(new unsigned char[input.size() + AES_BLOCK_SIZE]);
-    unique_ptr<unsigned char[]> plaintext(new unsigned char[input.size() + 1]);
+    auto ciphertext = std::make_unique<unsigned char[]>(input.size() + AES_BLOCK_SIZE);
```
WildFireFlum, line 33: use concord::Byte instead of unsigned char (whenever referring to bytes)

WildFireFlum, line 33: Instead of allocating both an array and a vector, allocate a vector here and return it.

secretsmanager/src/base64.cpp
```diff
@@ -36,7 +36,7 @@ string base64Enc(const vector<uint8_t>& msgBytes) {
```
WildFireFlum, line 36: Wrap every variable which holds a resource with a smart pointer.
This is an example implementation:
https://stackoverflow.com/a/63711162

threshsign/src/ThresholdSignaturesTypes.cpp
```diff
@@ -229,8 +226,7 @@ void Cryptosystem::validatePrivateKey(const std::string& key) const {
   isValidKey(keyType, key, expectedKeyLength);
 }
 
-bool Cryptosystem::isValidCryptosystemSelection(const std::string& type, const std::string& subtype) {
-  UNUSED(subtype);
+bool Cryptosystem::isValidCryptosystemSelection(const std::string& type, [[maybe_unused]] const std::string& subtype) {
```
WildFireFlum, line 229: Since we are not currently using [[maybe_unused]] in the code, either keep the UNUSED macro or refactor each usage of it.
Since these are places which we would potentially like to refactor (because their interface is not fully used), it will be easier to do so by searching a single term.
This comment applies to all of the other places you used this attribute.

PR #2856, opened 2022-11-16
Stability fix for test test_view_change_with_non_primary_replica_in_s…

HristoStaykov:
…tate_transfer.

While we are advancing the system we need to narrow down the selection of replicas to ones we are sure have the latest state. We therefore chose the expected next primary.

* **Problem Overview**  
 Test instability was observed during nightly runs.
* **Testing Done**  
  Tested locally with multiple runs to verify stability.


teoparvanov (approved):
Thanks @HristoStaykov !

PR #2862, opened 2022-11-20
DbCheckpointManager:fix forced kill of concord (sleeping monitor thread)

hshomroni:
DB checkpoint manager monitor thread was sleeping for 60 secs (default) intervals, preventing concord from shutting down gracefully on a relevant signal (such as SIGTERM). It is now using a cond var and the monitor thread is woken up upon destruction of DbCheckpointManager obj


glevkovich (approved):
LGTM!

PR #2867, opened 2022-11-21
Fix the memory leaks in sigProcessorTests

yontyon:
* **Problem Overview**  
  ASAN reported leaks in sigProcessorTests. 
  After digging into it, it appears that due to some circular ownership of smart pointers, the constructors have not been called. 
This PR is top fix the memory leaks reported by ASAN
To clarify, the leaks were related to the test only and were reported at its end as the smart pointers were not released. No leaks in the actual SigProcessor component.
* **Testing Done**  
  Running SigProcessorTests under ASAN

### DIFF
--- a/utt-replica/signature-processor/tests/sigProcessorTests.cpp
+++ b/utt-replica/signature-processor/tests/sigProcessorTests.cpp
@@ -293,8 +293,8 @@ class test_utt_instance : public ::testing::Test {
       message_comms[i] = msc;
       sig_processors[i] = std::make_shared<utt::SigProcessor>(
           i, n, t, 1000, msc, msr, ((IncomingMsgsStorageImp*)(ims.get()))->timers());
-      msr->registerMsgHandler(MsgCode::ClientRequest, [&, i](bftEngine::impl::MessageBase* message) {
-        ClientRequestMsg* msg = (ClientRequestMsg*)message;
+      msr->registerMsgHandler(MsgCode::ClientRequest, [&, i](std::unique_ptr<bftEngine::impl::MessageBase> message) {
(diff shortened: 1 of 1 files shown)


utt-replica/signature-processor/tests/sigProcessorTests.cpp
```diff
@@ -293,9 +293,8 @@ class test_utt_instance : public ::testing::Test {
       message_comms[i] = msc;
       sig_processors[i] = std::make_shared<utt::SigProcessor>(
           i, n, t, 1000, msc, msr, ((IncomingMsgsStorageImp*)(ims.get()))->timers());
-      auto sp = sig_processors[i];
-      msr->registerMsgHandler(MsgCode::ClientRequest, [&, sp](std::unique_ptr<bftEngine::impl::MessageBase> message) {
-        auto msg = std::make_unique<ClientRequestMsg>(message.release());
+      msr->registerMsgHandler(MsgCode::ClientRequest, [&, i](bftEngine::impl::MessageBase* message) {
+        ClientRequestMsg* msg = (ClientRequestMsg*)message;
```
eyalrund, line 297: where do we use this handler?

eyalrund (approved):

teoparvanov (approved):
Thanks for fixing this Yoni!

utt-replica/signature-processor/tests/sigProcessorTests.cpp
yontyon: In a real system, replicaImp implements this method. In the test, we need a way to handle the "callback" from the signature collector. This is why we have this handler here

PR #2869, opened 2022-11-21
[Draft] Working on review comments.

pkthapa:
* **Problem Overview**  
  * Move `concord::crypto::openssl::SHA3_256` to `concord::crypto::SHA3_256`.
* **Testing Done**  
  Local `make test`


util/include/crypto/openssl/digest_creator.hpp
```diff
@@ -12,13 +12,14 @@
...
 #include "hash.hpp"
+#include "crypto/crypto.hpp"
 #include "crypto/digest_creator.hpp"
 
 namespace concord::crypto::openssl {
 
 template <typename SHACTX,
-          typename = std::enable_if_t<std::is_same_v<SHACTX, concord::crypto::openssl::SHA2_256> ||
-                                      std::is_same_v<SHACTX, concord::crypto::openssl::SHA3_256>>>
+          typename = std::enable_if_t<std::is_same_v<SHACTX, concord::crypto::SHA2_256> ||
```
WildFireFlum, line 21: You should use the internal name here (openssl::EVPHash<EVP_sha3_256, 32>) because this code is in the openssl directory.

util/include/crypto/openssl/digest_creator.hpp
```diff
@@ -73,5 +74,5 @@ class OpenSSLDigestCreator : public concord::crypto::DigestCreator {
 
 namespace concord::crypto {
 
-using DigestGenerator = concord::crypto::openssl::OpenSSLDigestCreator<concord::crypto::openssl::SHA2_256>;
+using DigestGenerator = concord::crypto::openssl::OpenSSLDigestCreator<concord::crypto::SHA2_256>;
```
WildFireFlum, line 77: same here, the implementation of openssl should be explicit with the types it uses (not concord::crypto but rather concord::crypto::openssl)

util/include/crypto/openssl/digest_creator.hpp
pkthapa: From my point of view this is correct since we wanted to use hashing algorithm wrappers: `SHA2_256` & `SHA3_256`.

The alternate way would be to write as below:
`typename = std::enable_if_t<std::is_same_v<SHACTX, openssl::EVPHash<EVP_sha256, 32>>`

util/include/crypto/openssl/digest_creator.hpp
pkthapa: `concord::crypto` is used instead of `concord::crypto::openssl` as a wrapper. This is more cleaner way of writing. Wdyt?

util/include/crypto/openssl/digest_creator.hpp
WildFireFlum: Your code is correct, because it passes the tests, but this file is the internal openssl implementation of the digestcreator interface, and thus should use openssl types directly and not aliases.
If we decide to add the option of using another library for hashing, this code will no longer compile (assuming that we also change the default concord::crypto::SHA2_256).

util/include/crypto/openssl/digest_creator.hpp
WildFireFlum: Same argument as before. This code is **internal** to the specific openssl implementation of the interface, thus it can't use the alias which might change.

pkthapa: Addressed in https://github.com/vmware/concord-bft/pull/2869/commits/4a231052ce407519d8d8e3b592965876607f3a23

util/include/crypto/crypto.hpp
```diff
@@ -17,8 +17,75 @@
 #include <string>
 #include <cstddef>
 
+#include "crypto/openssl/hash.hpp"
```
toly-kournik, line 20: Why do we need this include here?

pkthapa: Removed.

PR #2880, opened 2022-12-02
Add return job status to test-single-suite in Makefile

titusen:
Github action restart recovery suite is reporting success even if test fails.

Addition to make job exits with number of failed jobs which in case any failed tests which would be seen by github action engine as failure,

Testing on branch. I have added temporary trigger for the job on every push. This change will not be merged.

### DIFF
--- a/.github/workflows/restart_recovery_suite.yml
+++ b/.github/workflows/restart_recovery_suite.yml
@@ -2,7 +2,6 @@ name: Restart recovery suite
 on:
   schedule:
     - cron: '0 1 * * *'
-on: push
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/restart_recovery_suite.yml
+++ b/.github/workflows/restart_recovery_suite.yml
@@ -2,6 +2,7 @@ name: Restart recovery suite
 on:
   schedule:
     - cron: '0 1 * * *'
+on: push
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -310,7 +310,8 @@ test-single-suite: ## Run a single test `make test-single-suite TEST_NAME=<test
 			if [[ '${BREAK_ON_FAILURE__}' = 'TRUE' ]];then echo "Breaking on first failure! (iteration $$i)"; exit $${RESULT}; fi; fi; \
 	done; \
 	echo "Test ${TEST_NAME} completed ${NUM_REPEATS__} iterations" \
-		"($$((${NUM_REPEATS__}-num_failures)) succeed, $${num_failures} failed)";
+		"($$((${NUM_REPEATS__}-num_failures)) succeed, $${num_failures} failed)"; \
+	exit $${num_failures}
```
teoparvanov, line 314: AFAIK, the issue is only seen with the Restart Recovery job.
On the other hand, it looks like this fix would affect all Apollo jobs that run as part of CI. Why aren't we seeing this issue with other jobs then?

.github/workflows/restart_recovery_suite.yml
```diff
@@ -2,6 +2,7 @@ name: Restart recovery suite
 on:
   schedule:
     - cron: '0 1 * * *'
+on: push
```
HristoStaykov, line 5: Question for me to get a better understanding of the git workflows - why do we need the "on: push" option? I understand the above one schedules the job periodically, but am not sure what the second does?

titusen: Just for testing purposes. It will not be merged.

Makefile
titusen: I've checked this and the return code which script returns by default is of the last executed command which in this case is "echo ...". Following script returns 0. If u comment last "ls" it will return code 2.
`#! /bin/bash
ls a_file_which_doesnt_exists
ls
`

teoparvanov: I think retries are expected in some cases (for flaky tests for instance).
With your fix, in case of retries, the suite will be considered failed, which we don't want.

teoparvanov (approved):
Thanks for fixing this, @titusen !

PR #2883, opened 2022-12-05
Utt user storage

yontyon:
* **Problem Overview**  
  Currently, a utt user is unrecoverable due to a lack of persistent storage. In this PR we introduce a minimal storage API for the utt user that allows the user to recover. In addition, we introduce a basic file-based persistent storage implementation  
* **Testing Done**  
  manual e2e test against a docker compose deployment


utt/utt-client-api/include/utt-client-api/IStorage.hpp
```diff
@@ -0,0 +1,48 @@
...
+  virtual void removeCoin(const libutt::api::Coin&) = 0;
+  virtual void startTransaction() = 0;
+  virtual void commit() = 0;
+
+  virtual uint64_t getLastExecutedSn() = 0;
+  virtual libutt::api::types::CurvePoint getClientSideSecret() = 0;
+  virtual libutt::api::types::CurvePoint getSystemSideSecret() = 0;
+  virtual libutt::api::types::Signature getRcmSignature() = 0;
+  virtual std::vector<libutt::api::Coin> getCoins() = 0;
+  virtual std::pair<std::string, std::string> getKeyPair() = 0;
```
teoparvanov, line 46: I'd suggest you document each individual method. Doing this might even help you find better names for some of them!

yontyon: Do you have any specific suggestions? IMO the names are pretty self-explained (you should be familiar with the client design and with utt in any case)

teoparvanov: Well, for example, what does `isNewStorage()` mean?

utt/wallet-cli/include/wallet.hpp
```diff
@@ -81,9 +81,7 @@ class Wallet {
   /// @brief Updates the ERC20 balance of the wallet.
   void updatePublicBalance(Channel& chan);
 
-  struct DummyUserStorage : public utt::client::IUserStorage {};
-
-  DummyUserStorage storage_;
+  std::unique_ptr<utt::client::IStorage> storage_;
```
teoparvanov, line 84: Also, as discussed offline - let's run the automated E2E test against a build containing these changes. Please add the outcome of that run in the "Testing done" section.

utt/wallet-cli/include/fileBasedUserStorage.hpp
```diff
@@ -0,0 +1,50 @@
...
+  bool isNewStorage() override;
+  void setKeyPair(const std::pair<std::string, std::string>&) override;
+  void setLastExecutedSn(uint64_t) override;
+  void setClientSideSecret(const libutt::api::types::CurvePoint&) override;
+  void setSystemSideSecret(const libutt::api::types::CurvePoint&) override;
+  void setRcmSignature(const libutt::api::types::Signature&) override;
+  void setCoin(const libutt::api::Coin&) override;
+  void removeCoin(const libutt::api::Coin&) override;
+  void startTransaction() override;
+  void commit() override;
```
teoparvanov, line 35: IMHO, non-business logic methods, such as `isNewStorage`, `startTransaction`, `commit` should be in a separate interface.

utt/utt-client-api/include/storage/ITransactionalStorage.hpp
```diff
@@ -0,0 +1,62 @@
...
+   * @brief Starts a new atomic transaction
+   *
+   */
+  virtual void startTransaction() = 0;
+
+  /**
+   * @brief Atomically commits a transaction
+   *
+   */
+  virtual void commit() = 0;
```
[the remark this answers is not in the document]
yontyon: IMO it's much more complicated to do so.
Consider that the transaction guard has to be familiar with ITransactionalStorage while the actual getters/setters are defined in the IStorage interface. 
So if we simply inherit from both of them, the question is which of the interfaces will be exposed to the User component? 
If it is the IStorage interface, then we wouldn't be able to have a transactional semantic. If it is the ITransacionalStorage, then we won't be able to get/set.
So, what we can do is either directly inherit from IStorage or define a new interface that is aware of both interfaces which will effectively be the same as not separating them at all.
The way it is implemented actually allows the user to implement completely two different interfaces.
I guess there are other solutions, but unless I miss something basic, I think the current approach is the simpler and cleaner one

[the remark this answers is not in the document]
yontyon: Notice, that there is no need to implement the IStorage method in the ITransactionalStorage implementation. It gets an IStorage implementation on its constructor and uses it.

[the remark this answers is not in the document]
teoparvanov: As discussed, let's revert to having a single `IStorage` interface with all the methods in it.

utt/wallet-cli/include/fileBasedUserStorage.hpp
teoparvanov: Let's keep it a single interface - ignore this comment for now.

utt/utt-client-api/include/storage/IStorage.hpp
```diff
@@ -0,0 +1,143 @@
...
+// file.
+
+#pragma once
+#include <vector>
+#include "coin.hpp"
+
+namespace utt::client {
+class IStorage {
+ public:
+  class guard {
```
teoparvanov, line 21: `tx_guard` maybe?

teoparvanov (commented):
Looks great, thanks Yoni! I'll approve once we get the following:
* a review from either @bsenthilr or @cloudnoize 
* a run of the automated E2E test with these changes (please update the `Testing done` section`)

utt/utt-client-api/src/User.cpp
```diff
@@ -96,6 +96,8 @@ struct User::Impl {
   std::vector<libutt::api::Coin> coins_;         // User's unspent UTT coins (tokens)
   std::optional<libutt::api::Coin> budgetCoin_;  // User's current UTT budget coin (token)
   std::set<std::string> budgetNullifiers_;
```
cloudnoize, line 98: this set ```budgetNullifiers_``` should also be persisted no?

utt/wallet-cli/src/fileBasedUserStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+    fs::remove(pending_path_);
+  }
+  if (fs::exists(state_path_)) {
+    std::ifstream f(state_path_);
+    state_->state_ = json::parse(f);
+  }
+}
+
+void FileBasedUserStorage::startTransaction() {
+  if (!state_->state_.contains("initialized")) state_->state_["initialized"] = true;
```
cloudnoize, line 78: what is the meaning of ```startTransaction``` ? it just set the state to initialized ?

utt/wallet-cli/src/fileBasedUserStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+  for (unsigned int i = 0; i < hex.length(); i += 2) {
+    std::string byteString = hex.substr(i, 2);
+    uint8_t byte = (uint8_t)strtol(byteString.c_str(), NULL, 16);
+    bytes.push_back(byte);
+  }
+
+  return bytes;
+}
+
+FileBasedUserStorage::FileBasedUserStorage(const std::string& path)
```
cloudnoize, line 61: please add unit tests to this class

utt/wallet-cli/src/fileBasedUserStorage.cpp
teoparvanov: Shouldn't this be clear from the method's documentation in the header file?

utt/utt-client-api/src/User.cpp
yontyon: no, it is stored from the current budget coin storage. We don't really care about the old budget coins (once we have a persistent storage that saves the latest one)

yontyon:
The target branch changed to master. 
See https://github.com/vmware/concord-bft/pull/2893

PR #2885, opened 2022-12-06
Readability improvements.

HristoStaykov:
Rename for each step reasoning lemma from IsEnabled to StepWasTaken. Move the Inv(c, v) to be a requires instead of part of the conjunction.

* **Problem Overview**  
  Readability improvements in step reasoning.
* **Testing Done**  
  dafny verifier used to check correctness of the changes.


teoparvanov (approved):
LGTM!

PR #2887, opened 2022-12-07
Remove the cyclic deps between reconfiguration and bftengine

yontyon:
* **Problem Overview**  
  Currently, there is a cyclic dependency between the reconfiguration library and core-bft.  This PR is to remove this dependency.
It is done by moving all the bft-related logic into the core-bft library and leaving the reconfiguration library clean from any core-bft functionality 
* **Testing Done**  
  CI


toly-kournik (approved):
LGTM

PR #2891, opened 2022-12-11
Makefile: merge both build-image commands into one

glevkovich:
When we build release image, we must buidld (and push to docker hub) a matching release image, since it is dependent on release. Here we merge both commands into one, and update the README.md file.

I've added an option to run each build command independently.

* **Problem Overview**  
This fix is supposed t simplify the way we create new concord-bft debug/release images.
* **Testing Done**  
NA


eyalrund (approved):

yuliasherman (commented):
LGTM

PR #2892, opened 2022-12-11
Add operator key to SigManager

yontyon:
* **Problem Overview**  
  Public keys are managed in the SigManager component in the replicas. In this PR we add the operator public key to SigManager such that we will be able to implement in the future a key exchange procedure for both, the operator tls certificate and main key 
* **Testing Done**  
  CI + manual test for the operator tls exchange in the product level

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -314,14 +314,13 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
   }
 
   std::string getOperatorPublicKey() {
-    std::string operator_key;
     std::ifstream op_key_file(pathToOperatorPublicKey_);
     if (!op_key_file.fail()) {
       std::stringstream buffer;
       buffer << op_key_file.rdbuf();
-      operator_key = buffer.str();
+      return buffer.str();
     }
-    return operator_key;
+    return std::string();
   }
 
  protected:
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -18,6 +18,7 @@
 #include <vector>
 #include <unordered_map>
 #include <chrono>
+#include <optional>
 #include "string.hpp"
 #include "kvstream.h"
 #include "crypto/factory.hpp"
@@ -310,6 +311,17 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
   }
   inline std::set<std::pair<const std::string, std::set<uint16_t>>>* getPublicKeysOfClients() {
     return (clientTransactionSigningEnabled || !clientsKeysPrefix.empty()) ? &publicKeysOfClients : nullptr;
+  }
+
+  std::string getOperatorPublicKey() {
+    std::string operator_key;
+    std::ifstream op_key_file(pathToOperatorPublicKey_);
+    if (!op_key_file.fail()) {
+      std::stringstream buffer;
+      buffer << op_key_file.rdbuf();
+      operator_key = buffer.str();
+    }
+    return operator_key;
   }
 
  protected:
--- a/bftengine/src/bftengine/ReplicaLoader.cpp
+++ b/bftengine/src/bftengine/ReplicaLoader.cpp
@@ -48,14 +48,22 @@ namespace {
 ReplicaLoader::ErrorCode loadConfig(LoadedReplicaData &ld) {
   ld.repsInfo = new ReplicasInfo(ld.repConfig, dynamicCollectorForPartialProofs, dynamicCollectorForExecutionProofs);
   auto &config = ld.repConfig;
-
-  ld.sigManager = SigManager::init(config.replicaId,
-                                   config.replicaPrivateKey,
-                                   config.publicKeysOfReplicas,
-                                   concord::crypto::KeyFormat::HexaDecimalStrippedFormat,
-                                   ReplicaConfig::instance().getPublicKeysOfClients(),
-                                   concord::crypto::KeyFormat::PemFormat,
-                                   *ld.repsInfo);
+  std::string operator_key;
+  std::ifstream op_key_file(bftEngine::ReplicaConfig::instance().pathToOperatorPublicKey_);
+  if (!op_key_file.fail()) {
+    std::stringstream buffer;
+    buffer << op_key_file.rdbuf();
+    operator_key = buffer.str();
+  }
+  ld.sigManager =
+      SigManager::init(config.replicaId,
+                       config.replicaPrivateKey,
+                       config.publicKeysOfReplicas,
+                       concord::crypto::KeyFormat::HexaDecimalStrippedFormat,
+                       ReplicaConfig::instance().getPublicKeysOfClients(),
+                       concord::crypto::KeyFormat::PemFormat,
+                       {ld.repsInfo->getIdOfOperator(), operator_key, concord::crypto::KeyFormat::PemFormat},
+                       *ld.repsInfo);
 
   std::unique_ptr<Cryptosystem> cryptoSys = std::make_unique<Cryptosystem>(ld.repConfig.thresholdSystemType_,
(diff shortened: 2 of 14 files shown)


bftengine/src/bftengine/ReplicaLoader.cpp
```diff
@@ -48,14 +48,22 @@ namespace {
...
-
-  ld.sigManager = SigManager::init(config.replicaId,
-                                   config.replicaPrivateKey,
-                                   config.publicKeysOfReplicas,
-                                   concord::crypto::KeyFormat::HexaDecimalStrippedFormat,
-                                   ReplicaConfig::instance().getPublicKeysOfClients(),
-                                   concord::crypto::KeyFormat::PemFormat,
-                                   *ld.repsInfo);
+  std::string operator_key;
+  std::ifstream op_key_file(bftEngine::ReplicaConfig::instance().pathToOperatorPublicKey_);
```
WildFireFlum, line 52: since this code is duplicated, add a sigmanager ctor which receives the path, parses the key and adds an operator. The operator should be an optional parameter, maybe even with std::optional.

bftengine/src/bftengine/ReplicasInfo.hpp
```diff
@@ -107,6 +107,7 @@ class ReplicasInfo {
   const std::set<PrincipalId> _idsOfExternalClients;
   const std::set<PrincipalId> _idsOfClientServices;
   const std::set<PrincipalId> _idsOfInternalClients;
+  PrincipalId _operator_id;
```
WildFireFlum, line 110: Document that there can only be a single operator in the network

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -162,6 +167,11 @@ SigManager::SigManager(PrincipalId myId,
       LOG_DEBUG(KEY_EX_LOG, "Adding key of client " << p.first << " key size " << key.size());
     }
   }
+  if (std::get<0>(operatorKey) > 0 && !std::get<1>(operatorKey).empty()) {
```
WildFireFlum, line 170: Use structured bindings to extract the fields out of the tuple. 
Example:
https://www.geeksforgeeks.org/structured-binding-c/

bftengine/src/bftengine/SigManager.hpp
```diff
@@ -55,6 +55,7 @@ class SigManager {
                           concord::crypto::KeyFormat replicasKeysFormat,
                           const std::set<std::pair<const std::string, std::set<uint16_t>>>* publicKeysOfClients,
                           concord::crypto::KeyFormat clientsKeysFormat,
+                          const std::tuple<PrincipalId, Key, concord::crypto::KeyFormat>& operatorKey,
```
WildFireFlum, line 58: Since clients also use sigmanager, the interface should allow to add an operator if wanted, but have a null default (std::optional supports this behavior).

bftengine/src/preprocessor/tests/messages/PreProcessResultMsg_test.cpp
```diff
@@ -49,6 +49,7 @@ bftEngine::impl::SigManager* createSigManagerWithSigning(
                           replicasKeysFormat,
                           publicKeysOfClients,
                           concord::crypto::KeyFormat::HexaDecimalStrippedFormat,
+                          {0, "", concord::crypto::KeyFormat::PemFormat},
```
WildFireFlum, line 52: you don't need this change if the parameter is optional.

WildFireFlum:
Can you please share your design for the wider change and how KeyExchangeManager, SigManager and the storage(or file-based storage) are going to interact when tls and main operator key are exchanged? This applies both to the operator whose keys are exchanges and the other participants who monitor his changed key.

bftengine/src/bftengine/ReplicaLoader.cpp
yontyon: done

bftengine/src/bftengine/ReplicasInfo.hpp
yontyon: done

bftengine/src/bftengine/SigManager.hpp
yontyon: done

bftengine/src/preprocessor/tests/messages/PreProcessResultMsg_test.cpp
yontyon: done

yontyon:
> Can you please share your design for the wider change and how KeyExchangeManager, SigManager and the storage(or file-based storage) are going to interact when tls and main operator key are exchanged? This applies both to the operator whose keys are exchanges and the other participants who monitor his changed key.

This PR is only for adding the operator public key to sigManager. The result of this change is (among others) that we can implement the operator tls key exchange in the very same way we do with the other bft clients. The actual implementation will be part of the product code and it happens in the very same way that every other TLS exchange happens.

In a nutshell:
1. a new certificate signed by the bft client main key is generated. 
2. the client restarts its communication channels
3. on the TLS handshake, the receiver gets the new certificate and validates it against the known public key of the peer
4. if it is valid, it saves the new certificate in the storage

bftengine/src/bftengine/ReplicasInfo.hpp
```diff
@@ -48,7 +48,7 @@ class ReplicasInfo {
   const std::set<PrincipalId>& idsOfExternalClients() const { return _idsOfExternalClients; }
   const std::set<PrincipalId>& idsOfInternalClients() const { return _idsOfInternalClients; }
   const std::set<PrincipalId>& idsOfIClientServices() const { return _idsOfClientServices; }
-
+  PrincipalId getIdOfOperator() { return _operator_id; }
```
WildFireFlum, line 51: should be a const member

WildFireFlum (approved):

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -312,6 +313,16 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
     return (clientTransactionSigningEnabled || !clientsKeysPrefix.empty()) ? &publicKeysOfClients : nullptr;
   }
 
+  std::string getOperatorPublicKey() {
```
toly-kournik, line 316: How will we distinguish between the situation when it's not an operator and a one when the should be a key, but it wasn't provided?

bftengine/src/bftengine/ReplicasInfo.cpp
```diff
@@ -144,6 +144,10 @@ ReplicasInfo::ReplicasInfo(const ReplicaConfig& config,
         if (start != end) LOG_INFO(GL, "Principal ids in _idsOfInternalClients: " << start << " to " << end - 1);
         return ret;
       }()} {
+  _operator_id = config.operatorEnabled_
+                     ? static_cast<PrincipalId>(numberOfReplicas() + _idsOfClientProxies.size() +
+                                                _idsOfExternalClients.size() + _idsOfClientServices.size())
```
eyalrund, line 149: Do we have some centralized place for this calculation? I assume we calculate it in different places. and it might change sometime along the way...

eyalrund (approved):

bftengine/src/bftengine/ReplicasInfo.cpp
yontyon: The calculation is listed here https://github.com/vmware/concord-bft/blob/master/bftengine/src/bftengine/ReplicasInfo.cpp#L19

yontyon: If it is changed, then we need to change the whole calculation of in this component

bftengine/include/bftengine/ReplicaConfig.hpp
yontyon: We don't do it at this point. The responsibility of ReplicaConfig is only to parse the operator key if it exists, not to assert if it wasn't provided and needed to be.
So, it is some higher level responsibility to assert this scenario. I mean, if it happens, then it should affect many more components, this problem is not a problem specifically related to this change.

PR #2895, opened 2022-12-12
bft.py: avoid input paths assertion when deployment already exist

glevkovich:
An existing deployment might pass some of the folders paths as empty so we skip the next assertion (see BftTestNetwork.existing()).

* **Problem Overview**  
An existing deployment (e.g) doesn't pass all paths non-empty to Apollo infra. Assertions in assert_dirs_exist should done only for Apollo tests
* **Testing Done**  
Ran Relevant Hermes tests and check that they pass.


eyalrund (approved):

toly-kournik (approved):
LGTM

PR #2896, opened 2022-12-13
Remove cre-corebft cyclic deps

yontyon:
* **Problem Overview**  
  CRE is linked to corebft in order to use ReplicaConfig. On the other side, replicas are using cre for running the internal state transfer cre.
Also, using ReplicaConfig is incorrect in the client context because no one initializes this component on the client side.
* **Testing Done**  
  CI


toly-kournik (approved):
LGTM!

PR #2902, opened 2022-12-13
bft.py: fix debug_tools access when running external tests

glevkovich:
For now, we do not support such scenario. All tools in Apollo infra should be used for internal Apollo concord-bft tests suites.

Problem Overview
External tests (Hermes) are failing since debug-tool wasn't defined correctly. Here we apply a fix.
Testing Done
Regular tests on a side job on Gitlab CI.


eyalrund (approved):

samcramer (approved):
Looks good.

PR #2904, opened 2022-12-16
Adding AES_GCM encryption mode to overcome AES_CBC limitations

anshubit5516:
* **Problem Overview**  
  Currently Concord supports AES_CBC as mode of encryption, which is vulnerable to oracle padding attack. More information can be found here(https://en.wikipedia.org/wiki/Padding_oracle_attack).
So added AES_GCM mode for all encryption/decryption functionalities. GCM is more secured and stable.
* **Testing Done**  
  Unit testing is done and added in the PR.

### DIFF
--- a/libs/secrets/src/aes.cpp
+++ b/libs/secrets/src/aes.cpp
@@ -12,7 +12,6 @@
 // file.
 //
 // This convenience header combines different block implementations.
-
 #include "aes.h"
 #include "crypto/openssl/crypto.hpp"
 #include "assertUtils.hpp"
@@ -25,7 +24,7 @@ using concord::crypto::openssl::OPENSSL_SUCCESS;
 using concord::crypto::openssl::UniqueCipherContext;
 
 vector<uint8_t> AES_CBC::encrypt(std::string_view input) {
-  if (algo_ == concord::crypto::SignatureAlgorithm::EdDSA) {
+  if (getAlgorithm() == concord::crypto::SignatureAlgorithm::EdDSA) {
     if (input.empty()) {
       return {};
     }
@@ -41,8 +40,9 @@ vector<uint8_t> AES_CBC::encrypt(std::string_view input) {
     int c_len{0};
     int f_len{0};
 
-    ConcordAssert(OPENSSL_SUCCESS ==
-                  EVP_EncryptInit_ex(ctx.get(), EVP_aes_256_cbc(), nullptr, params_.key.data(), params_.iv.data()));
+    ConcordAssert(
+        OPENSSL_SUCCESS ==
+        EVP_EncryptInit_ex(ctx.get(), EVP_aes_256_cbc(), nullptr, getKeyParams().key.data(), getKeyParams().iv.data()));
     ConcordAssert(OPENSSL_SUCCESS ==
                   EVP_EncryptUpdate(ctx.get(), ciphertext.get(), &c_len, plaintext.get(), input.size()));
     ConcordAssert(OPENSSL_SUCCESS == EVP_EncryptFinal_ex(ctx.get(), ciphertext.get() + c_len, &f_len));
@@ -56,7 +56,7 @@ vector<uint8_t> AES_CBC::encrypt(std::string_view input) {
 }
 
 string AES_CBC::decrypt(const vector<uint8_t>& cipher) {
-  if (algo_ == concord::crypto::SignatureAlgorithm::EdDSA) {
+  if (getAlgorithm() == concord::crypto::SignatureAlgorithm::EdDSA) {
     if (cipher.capacity() == 0) {
       return {};
     }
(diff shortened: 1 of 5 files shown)


libs/secrets/src/aes.cpp
```diff
@@ -41,8 +40,9 @@ vector<uint8_t> AES_CBC::encrypt(std::string_view input) {
     int c_len{0};
     int f_len{0};
 
-    ConcordAssert(OPENSSL_SUCCESS ==
-                  EVP_EncryptInit_ex(ctx.get(), EVP_aes_256_cbc(), nullptr, params_.key.data(), params_.iv.data()));
+    ConcordAssert(
+        OPENSSL_SUCCESS ==
+        EVP_EncryptInit_ex(ctx.get(), EVP_aes_256_cbc(), nullptr, getKeyParams().key.data(), getKeyParams().iv.data()));
```
eyalrund, line 45: why is it using aes_256_cbc?

libs/secrets/src/secrets_manager_enc.cpp
```diff
@@ -31,7 +30,13 @@ SecretsManagerEnc::SecretsManagerEnc(const SecretData& secrets)
     throw std::runtime_error("Unsupported encryption algorithm " + secrets.algo + "Supported encryptions: " + encs);
   }
 }
-
+std::unique_ptr<iAES_Mode> SecretsManagerEnc::getEncryptionMode() {
```
eyalrund, line 33: maybe call it getAESEncryptionMode()

libs/secrets/src/aes.cpp
```diff
@@ -25,7 +24,7 @@ using concord::crypto::openssl::OPENSSL_SUCCESS;
 using concord::crypto::openssl::UniqueCipherContext;
 
 vector<uint8_t> AES_CBC::encrypt(std::string_view input) {
-  if (algo_ == concord::crypto::SignatureAlgorithm::EdDSA) {
+  if (getAlgorithm() == concord::crypto::SignatureAlgorithm::EdDSA) {
```
eyalrund, line 27: Why is edDSA part of this class?

eyalrund (changes requested):

eyalrund:
Testing must be done also with real deployment

libs/secrets/src/aes.h
```diff
@@ -24,17 +24,41 @@
...
 
  private:
   KeyParams params_;
   concord::crypto::SignatureAlgorithm algo_;
 };
 
+class AES_CBC : public iAES_Mode {
+ public:
+  AES_CBC(const KeyParams& params,
+          concord::crypto::SignatureAlgorithm algo = concord::crypto::SignatureAlgorithm::EdDSA)
```
cjn, line 47: Do we need algo string ? If so is this the correct default value for the interface?

libs/secrets/src/aes.h
```diff
@@ -24,17 +24,41 @@
...
+  const concord::crypto::SignatureAlgorithm getAlgorithm() { return algo_; }
+  // Not adding setters as these algo/modes must be set at construction time
+  virtual ~iAES_Mode() = default;
 
  private:
   KeyParams params_;
   concord::crypto::SignatureAlgorithm algo_;
 };
 
+class AES_CBC : public iAES_Mode {
```
cjn, line 44: iAES_Mode - Can you check if we use this naming convention for other abstract classes?

toly-kournik (changes requested):
@anshubit5516,
this change should enter our `concord-crypto` library which will shortly spin off the `util` library.
In any case there shouldn't be any direct usage of `OpenSSL` library except from there.
This allows for seamless switching of underlying crypto implementations.
My suggestion is to wait a bit until we merge the `concord-crypto` separation merge this week

libs/secrets/src/aes.cpp
anshubit5516: Since we are supporting both (CBC and GCM) and this method is CBC implementation.

libs/secrets/src/secrets_manager_enc.cpp
anshubit5516: done

libs/secrets/src/aes.h
anshubit5516: What i have seen in the code, generally interface names are like IVerifier,ISigner..
I can rename the interface as IAesMode just to be aligned.

libs/secrets/src/aes.cpp
anshubit5516: removed

libs/secrets/src/aes.h
anshubit5516: i confirmed with code owners and removed it.

anshubit5516:
> @anshubit5516, this change should enter our `concord-crypto` library which will shortly spin off the `util` library. In any case there shouldn't be any direct usage of `OpenSSL` library except from there. This allows for seamless switching of underlying crypto implementations. My suggestion is to wait a bit until we merge the `concord-crypto` separation merge this week

As discussed, i will get this PR merged and then after crypto folder is organised, will move the code

eyalrund (approved):

cjn:
LGTM. Thanks @anshubit5516

cjn (approved):

toly-kournik (approved):
LGTM

PR #2906, opened 2022-12-19
Add metrics in clientservice

Efrat1:
* **Problem Overview**  
  Added metrics to clientservice:
  1. num_completed_requests
  2. num_incoming_requests
  3. num_failed_requests
* **Testing Done**  
  Apollo


client/clientservice/src/request_service.cpp
```diff
@@ -203,7 +211,28 @@ void RequestServiceCallData::sendToConcordClient() {
...
+concordMetrics::Component RequestServiceCallData::metrics_component_{
+    concordMetrics::Component("RequestService", std::make_shared<concordMetrics::Aggregator>())};
+
+RequestServiceCallData::Metrics RequestServiceCallData::metrics_{
+    metrics_component_.RegisterAtomicCounter("num_completed_requests", 0),
+    metrics_component_.RegisterAtomicCounter("num_incoming_requests", 0),
+    metrics_component_.RegisterAtomicCounter("num_failed_requests", 0)};
+
+void RequestServiceCallData::updateAggregator() {
+  const auto metrics_update_duration_microsec{std::chrono::microseconds(1s).count()};
```
yuliasherman, line 227: How often have we updated this metric before? 1sec feels too frequent.

yuliasherman (commented):
LGTM. Did you test all the metrics though Wavefront?

client/clientservice/src/request_service.cpp
Efrat1: What do you mean before? The metric that is used in **Trades/s** comes from DAML so I can't tell.
In replica, for example, the default update interval is 600ms (but it's configurable).

What other frequency do you think would be more reasonable?

Efrat1:
Yes, I tested the metrics though Wavefront.

client/clientservice/src/request_service.cpp
yuliasherman: I see that each component has its own update interval and 1 sec is even too much. Let's set it to 500ms and make it configurable.

Efrat1: Actually they all have the same update interval.
I made it configurable (config name: metrics_dump_interval_ms). By default it is set to 1000, other suggestion?

yuliasherman: 500ms?

yuliasherman (approved):
LGTM

eyalrund (approved):

PR #2910, opened 2023-01-01
Fix leak caused by a non-deleted original message

yuliasherman:
* **Problem Overview**  
In order to call the correct overloaded 'validate' function, incoming PreProcessor messages should be created through 'new' or 'make_unique' operators, which should be followed by the removal of the original message. The removal was missing and this caused a memory leak.
* **Testing Done**  
Tested through Maestro 5 days blockchain run.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -177,7 +177,7 @@ void ReplicaImp::registerMsgHandlers() {
 template <typename T>
 void ReplicaImp::messageHandler(std::unique_ptr<MessageBase> msg) {
   auto trueTypeObj = std::make_unique<T>(msg.get());
-  delete msg.release();
+  msg.reset();
   if (isCollectingState()) {
     // Extract only required information and handover to ST thread
     if (validateMessage(trueTypeObj.get())) {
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -177,7 +177,7 @@ void ReplicaImp::registerMsgHandlers() {
 template <typename T>
 void ReplicaImp::messageHandler(std::unique_ptr<MessageBase> msg) {
   auto trueTypeObj = std::make_unique<T>(msg.get());
-  delete msg.release();
+  msg.reset();
```
glevkovich, line 180: A leftover from previous PR I assume?

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -1299,8 +1299,8 @@ void PreProcessor::msgProcessingLoop() {
         }
         default:
           LOG_ERROR(logger(), "Unknown message" << KVLOG(msg->type()));
-          delete msg;
       }
+      delete msg;
```
glevkovich, line 1303: nit:
I suggest to add comments at the end of long scopes. I'm not sure, but that's what have probably caused rthw mistake?
So here you can probably prevent future mistakes .
e.g add 
`} // switch`
and add
`} // while`

glevkovich (approved):
@yuliasherman Thanks.
I'm going to add a task to convert this loop to smart pointer usage (was thinking to offer doing it here, but it will be more work).
Added one small comment.
Approved.

bftengine/src/bftengine/ReplicaImp.cpp
yuliasherman: Yep

eyalrund (approved):

PR #2914, opened 2023-01-12
Dafny opaque all invariants

HristoStaykov:
* **Problem Overview**  
 Dafny verification times have increased since we introduced the logic regarding View Change ans State Transfer. In this branch we restructure the proof to only reveal relevant predicates to the step that is being taken in order to better manage the triggering automation and not allow the verifier to trigger on irrelevant facts.
* **Testing Done**  
  The current state of the proof (without View Change invariants) passes correctly after the refactoring.

### DIFF
--- a/docs/sbft-formal-model/messages.dfy
+++ b/docs/sbft-formal-model/messages.dfy
@@ -80,7 +80,7 @@ module Messages {
     }
   }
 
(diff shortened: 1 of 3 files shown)


teoparvanov (approved):
LGTM, thanks @HristoStaykov !

PR #2915, opened 2023-01-13
remove separate asio lib usage

situ-s:
This PR removes separate asio installation and usage from concord-bft.
Currently we have upgraded to the recent boost version, so a separate asio lib shouldn’t be used anymore.
Current boost has asio included as well.


toly-kournik (approved):
LGTM!

PR #2918, opened 2023-01-19
Accomodating new fields added got GCM mode

anshubit5516:
Problem Overview
Currently Concord supports AES_CBC as mode of encryption, which is vulnerable to oracle padding attack. More information can be found here(https://en.wikipedia.org/wiki/Padding_oracle_attack).
So added AES_GCM mode for all encryption/decryption functionalities. GCM is more secured and stable.

Previous PR([https://github.com/vmware/concord-bft/pull/2904]) is merged in which tag length was part of additional info. But after collective discussion from agent DA and concord, it was decided that new key value pair (tag_length : <val_in_bits > will be added. 
So, raising this PR for concord to handle this new parameter.

Testing Done
Unit testing is done and added in the PR.
A private build (IMAGE_TAG: 0.0.1.0.1597) is created and tested. Able to validate behaviour between agent and concord for GCM mode.. But transactions can't be done without DA changes.

### DIFF
--- a/libs/secrets/src/aes.h
+++ b/libs/secrets/src/aes.h
@@ -26,22 +26,23 @@ namespace concord::secretsmanager {
 
 class IAesMode {
  public:
-  IAesMode(const KeyParams& params, const std::string& a_info = "") : params_{params}, additional_info(a_info) {}
+  IAesMode(const KeyParams& params, const uint32_t tagLengthBits = 128)
+      : params_{params}, tag_length_in_bits(tagLengthBits) {}
   virtual std::vector<uint8_t> encrypt(std::string_view input) = 0;
   virtual std::string decrypt(const std::vector<uint8_t>& cipher) = 0;
   const KeyParams getKeyParams() { return params_; }
-  const std::string getAdditionalInfo() { return additional_info; }
+  const uint32_t getTagLengthInBits() { return tag_length_in_bits; }
   // Not adding setters as these algo/modes must be set at construction time
   virtual ~IAesMode() = default;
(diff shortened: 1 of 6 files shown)


libs/secrets/src/aes.h
```diff
@@ -26,30 +26,31 @@ namespace concord::secretsmanager {
 
 class IAesMode {
  public:
-  IAesMode(const KeyParams& params, const std::string& a_info = "") : params_{params}, additional_info(a_info) {}
+  IAesMode(const KeyParams& params, const uint32_t tagLengthBits = 128)
+      : params_{params}, tag_length_in_bits(tagLengthBits) {}
   virtual std::vector<uint8_t> encrypt(std::string_view input) = 0;
   virtual std::string decrypt(const std::vector<uint8_t>& cipher) = 0;
   const KeyParams getKeyParams() { return params_; }
-  const std::string getAdditionalInfo() { return additional_info; }
```
cjn, line 33: Should we consider keeping this in the interface even though we may not be using it since it is meant for sending additional information in the original response.

cjn:
LGTM @anshubit5516  apart from a single comment.

libs/secrets/src/aes.h
anshubit5516: Since i removed the additional_info from the interface, hence removed the method as well. As discussed with Galya, we keep this as null and mostly used in case of any error which is dealt in other class.. We can still discuss and add this field for future purpose.

cjn (approved):

PR #2923, opened 2023-01-26
Fix the PreProcessor crash on graceful shutdown

yuliasherman:
* **Problem Overview**  
On a graceful shutdown, replica crashes in the PreProcessor destructor with a core dump.
1. Ignore incoming messages that arrive at the PreProcessor when the message processing loop has been stopped.
2. Delete non-processed messages from the PreProcessor queue in the context of the dispatcher thread.
* **Testing Done**  
Tested on a reserved system + 5 days Maestro regression run.

### DIFF
--- a/bftengine/src/preprocessor/PreProcessor.cpp
+++ b/bftengine/src/preprocessor/PreProcessor.cpp
@@ -315,21 +315,6 @@ bool PreProcessor::validateMessage(MessageBase *msg) const {
   }
 }
 
-void PreProcessor::stop() {
-  if (!msgLoopDone_) {
-    msgLoopDone_ = true;
-    msgLoopSignal_.notify_all();
-    cancelTimers();
-  }
-
-  if (!threadPool_.isStopped()) {
-    threadPool_.stop();
-    if (msgLoopThread_.joinable()) {
-      msgLoopThread_.join();
-    }
-  }
-}
-
 PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
                            shared_ptr<IncomingMsgsStorage> &incomingMsgsStorage,
                            shared_ptr<MsgHandlersRegistrator> &msgHandlersRegistrator,
@@ -424,18 +409,33 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
   addTimers();
 }
 
+void PreProcessor::stop() {
+  LOG_DEBUG(logger(), "Going to stop PreProcessor");
+  if (!msgLoopDone_) {
+    {
+      std::unique_lock<std::mutex> l(msgLock_);
+      msgLoopDone_ = true;
+    }
+    msgLoopSignal_.notify_all();
+    cancelTimers();
+  }
+  if (!threadPool_.isStopped()) {
+    threadPool_.stop();
+    if (msgLoopThread_.joinable()) {
+      msgLoopThread_.join();
+    }
+  }
+  LOG_DEBUG(logger(), "PreProcessor stopped");
+}
+
 PreProcessor::~PreProcessor() {
-  LOG_TRACE(logger(), "~PreProcessor start");
-
-  stop();
-
+  LOG_DEBUG(logger(), "~PreProcessor start");
   if (!memoryPoolEnabled_) {
     for (const auto &result : preProcessResultBuffers_) {
       delete[] result->buffer;
     }
   }
-  while (msgs_.read_available()) delete msgs_.front();
-  LOG_TRACE(logger(), "~PreProcessor done");
+  LOG_DEBUG(logger(), "~PreProcessor done");
 }
 
 void PreProcessor::addTimers() {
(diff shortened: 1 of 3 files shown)


bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -424,18 +409,32 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
   addTimers();
 }
 
-PreProcessor::~PreProcessor() {
-  LOG_TRACE(logger(), "~PreProcessor start");
+void PreProcessor::stop() {
+  LOG_DEBUG(logger(), "Going to stop PreProcessor");
+  lock_guard<mutex> lock(stopGuard_);
```
hshomroni, line 414: If I understand correctly, stop() func is only called from the dtor, hence it is guaranteed to be called only from one context (assuming no double destruction of the preprocessor). Thus, the lock here is redundant. what actually solved the problem was adding  "if (!msgLoopDone_)" in PreProcessor::messageHandler, which prevents writing to the preprocessor msg queue during the stop function, which deletes msgs from the same queue. Am I missing something?

yuliasherman: That's the point. It is also called when the replica gets destructed (as a callback).

hshomroni: So maybe the correct solution is to have the stop func called only from a single context? 
We do have a mix in our code base of RAII and "start-stop" patterns, but it seems like in the preprocessor case we use both approaches together...

yuliasherman: Actually, Tytus has already pushed the code that calls the stop() callback, I can remove the call from the dtor.

yuliasherman: I'll do that, thanks.

hshomroni (approved):
LGTM

bftengine/src/preprocessor/PreProcessor.cpp
```diff
@@ -424,18 +409,30 @@ PreProcessor::PreProcessor(shared_ptr<MsgsCommunicator> &msgsCommunicator,
   addTimers();
 }
 
-PreProcessor::~PreProcessor() {
-  LOG_TRACE(logger(), "~PreProcessor start");
-
-  stop();
+void PreProcessor::stop() {
+  LOG_DEBUG(logger(), "Going to stop PreProcessor");
+  if (!msgLoopDone_) {
```
glevkovich, line 414: === same comment to all below cases when you trigger notify_one/all on this cond variable. ===

Are you catching a lock somewhere?  If not, I think that this will never be 100% thread safe. It will be ok in most cases but fail in others (e.g once in 100 or once in 1000) , and lead to undefined behaviors in shutdown (crashes, etc).
You are changing a shared state and notifying without taking a lock. This kind of usage isn't according to the way required:
```
The thread that intends to modify the shared variable must:

1) Acquire a std::mutex (typically via [std::lock_guard](https://en.cppreference.com/w/cpp/thread/lock_guard))
2) Modify the shared variable while the lock is owned
3) Call [notify_one](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one) or [notify_all](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all) on the std::condition_variable (can be done after releasing the lock)
```

more info in:
https://en.cppreference.com/w/cpp/thread/condition_variable

bftengine/src/preprocessor/tests/preprocessor_test.cpp
```diff
@@ -769,7 +771,7 @@ TEST(requestPreprocessingState_test, batchCancelledNoConsensusReached) {
   }
   auto clientBatchReqMsg = make_unique<ClientBatchRequestMsg>(clientId, reqBatch, batchSize, cid);
   // Call the PreProcessor callback to handle ClientBatchRequest message
-  reqMsgHandlerCallback(move(clientBatchReqMsg));
+  reqMsgHandlerCallback(std::move(clientBatchReqMsg));
```
glevkovich, line 774: This is good for future, in practice it does nothing - compiler will delete the move constructor since we already define custome copy constructir (rule of 5).
We have this issue on all messages actually which do the same.

same for all below new std::move

glevkovich (changes requested):

bftengine/src/preprocessor/PreProcessor.cpp
yuliasherman: I still think that the mutex lock is unneeded in case you work with a lock-free underlined queue. You actually have no object to guard. In terms of the guide, there is no "shared variable". The conditional variable is used in my case just to avoid constant polling of the queue in case there is no data there. notify_one could be called without locking.

bftengine/src/preprocessor/tests/preprocessor_test.cpp
yuliasherman: I did not add 'move' now, just put 'std::' for readability.

bftengine/src/preprocessor/PreProcessor.cpp
glevkovich: `Even if the shared variable is atomic, it must be modified while owning the mutex to correctly publish the modification to the waiting thread.`

yuliasherman: The condition variable is not waiting for any variable to be modified, please look again at the code.

glevkovich (approved):
LGTM

eyalrund (approved):

PR #2925, opened 2023-01-31
Make state transfer use the actual number of reserved pages

Efrat1:
Instead of maxNumOfReservedPages configuration, state transfer uses the actual number of reserved pages to make
maxNumOfReservedPages config more stable.

* **Problem Overview**  
  maxNumOfReservedPages was used by ST to calculate max Vblock size and ST item size, and allocations were done accordingly. With that approach, maxNumOfReservedPages had to be kept as closed as possible to the actual number of reserved pages. But the number of reserved pages is affected by many configurations, so it is not a stable permanent number.
  With this PR, ST now uses the actual registered reserved pages instead, and that allows us setting the maxNumOfReservedPages to be higher and as a result more stable.

* **Testing Done**  
  Apollo

### DIFF
--- a/bftengine/include/bftengine/ReservedPagesClient.hpp
+++ b/bftengine/include/bftengine/ReservedPagesClient.hpp
@@ -77,6 +77,7 @@ class ResPagesClient : public ReservedPagesClientBase, public IReservedPages {
   }
   /**
    * Should be called at initialization if number of pages is not known at compile time
+   * This function must be called only from the ReplicaForStateTransfer constructor before stateTransfer->init().
    */
   static void setNumResPages(const uint32_t numPages) {
     Registry& reg = registry();
--- a/bftengine/src/bcstatetransfer/BCStateTran.cpp
+++ b/bftengine/src/bcstatetransfer/BCStateTran.cpp
@@ -249,18 +249,11 @@ BCStateTran::BCStateTran(const Config &config, IAppState *const stateApi, DataSt
       maxPostprocessedBlockId_{0},
       config_{config},
       replicas_{generateSetOfReplicas(config_.numReplicas)},
-      maxVBlockSize_{calcMaxVBlockSize(config_.maxNumOfReservedPages, config_.sizeOfReservedPage)},
-      maxItemSize_{calcMaxItemSize(config_.maxBlockSize, config_.maxNumOfReservedPages, config_.sizeOfReservedPage)},
-      maxNumOfChunksInAppBlock_{
-          calcMaxNumOfChunksInBlock(maxItemSize_, config_.maxBlockSize, config_.maxChunkSize, false)},
-      maxNumOfChunksInVBlock_{
-          calcMaxNumOfChunksInBlock(maxItemSize_, config_.maxBlockSize, config_.maxChunkSize, true)},
       maxNumOfStoredCheckpoints_{0},
       numberOfReservedPages_{0},
       cycleCounter_{0},
       running_{false},
       replicaForStateTransfer_{nullptr},
-      buffer_(new char[maxItemSize_]),
       randomGen_{randomDevice_()},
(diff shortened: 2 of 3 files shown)


bftengine/include/bftengine/ReservedPagesClient.hpp
```diff
@@ -77,6 +77,7 @@ class ResPagesClient : public ReservedPagesClientBase, public IReservedPages {
   }
   /**
    * Should be called at initialization if number of pages is not known at compile time
+   * Call this function in ReplicaForStateTransfer constructor.
```
yuliasherman, line 80: --> "This function must be called only from the ReplicaForStateTransfer constructor before it sets an overall number of reserved pages in the system."

bftengine/src/bcstatetransfer/BCStateTran.hpp
```diff
@@ -273,10 +273,10 @@ class BCStateTran : public IStateTransfer {
...
   const Config config_;
   const set<uint16_t> replicas_;
-  const uint32_t maxVBlockSize_;
-  const uint32_t maxItemSize_;
-  const uint32_t maxNumOfChunksInAppBlock_;
-  const uint32_t maxNumOfChunksInVBlock_;
+  uint32_t maxVBlockSize_;
+  uint32_t maxItemSize_;
+  uint32_t maxNumOfChunksInAppBlock_;
+  uint32_t maxNumOfChunksInVBlock_;
```
yuliasherman, line 279: Initialize those data members here or in the ctor.

bftengine/src/bcstatetransfer/BCStateTran.cpp
```diff
@@ -370,6 +362,13 @@ void BCStateTran::initImpl(uint64_t maxNumOfRequiredStoredCheckpoints,
     ConcordAssertEQ(replicaForStateTransfer_, nullptr);
     ConcordAssertEQ(sizeOfReservedPage, config_.sizeOfReservedPage);
 
+    maxVBlockSize_ = calcMaxVBlockSize(numberOfRequiredReservedPages, config_.sizeOfReservedPage);
+    maxItemSize_ = calcMaxItemSize(config_.maxBlockSize, numberOfRequiredReservedPages, config_.sizeOfReservedPage);
+    maxNumOfChunksInAppBlock_ =
+        calcMaxNumOfChunksInBlock(maxItemSize_, config_.maxBlockSize, config_.maxChunkSize, false);
+    maxNumOfChunksInVBlock_ = calcMaxNumOfChunksInBlock(maxItemSize_, config_.maxBlockSize, config_.maxChunkSize, true);
+    buffer_.reset(new char[maxItemSize_]);
+
```
yuliasherman, line 371: Please add a log printing those values and numberOfRequiredReservedPages

bftengine/include/bftengine/ReservedPagesClient.hpp
yuliasherman: The second one variant OK.

bftengine/src/bcstatetransfer/BCStateTran.cpp
Efrat1: numberOfRequiredReservedPages is assigned to numberOfReservedPages_ which is printed to log in line 382.

bftengine/src/bcstatetransfer/BCStateTran.hpp
Efrat1: Done

yuliasherman (approved):
LGTM

yuliasherman:
Please deploy the system with the configuration described in a bug and verify that it works.

eyalrund (approved):

Efrat1:
Will be verified on the system with the configuration described in a bug after we have the fix in concord. The fix is safe to merge

PR #2931, opened 2023-02-07
Install gmp library as utt dependency after it was removed from concord-crypto as relic library dependency

toly-kournik:
Install gmp library as utt dependency after it was removed from concord-crypto as relic library dependency


teoparvanov (approved):
LGTM, thanks Toly!

yontyon (approved):

eyalrund (approved):

PR #2932, opened 2023-02-10
Fix key exchange crash with f+1 replicas down

situ-s:
* **Problem Overview**  
  Replica crashes with assert 
  2023-01-27T09:53:52,064,+0000|FATAL|1|concord.bft|message-processing||17|0|slow|KeyExchangeManager.cpp:104| Assert: expression 'private_keys_.key_data().generated.pub == kemsg.pubkey' is false in function onKeyExchange (/concord/submodules/concord-bft/bftengine/src/bftengine/KeyExchangeManager.cpp 104) | [SQ:998]

The source replica(where key-exchange is performed) generates candidate key for sequence numbers on receiving key-exchange command. As f+1 replicas are down, the multiple key-exchange commands are sent from operator, candidate key gets generated for multiple seq numbers but not executed. When other replicas are up, the flow reaches to onKeyExchange, the first sequence number gets executed and candidate keys are cleared. The next sequence number does not have candidate keys, causing the crash.

Fix:
Whenever new candidate keys are generated for a sequence number, we are storing seqnum->candidate key pair in map.
For each execution of sequence number, we are clearing it from the map. In this way we always have a candidate key for the sequence number.

* **Testing Done**  
  Manually tested on build 0.0.1.0.1662.
  Hermes CI passed.
  Verified by ST Team

### DIFF
--- a/bftengine/src/bftengine/KeyExchangeManager.cpp
+++ b/bftengine/src/bftengine/KeyExchangeManager.cpp
@@ -96,9 +96,12 @@ std::string KeyExchangeManager::onKeyExchange(const KeyExchangeMsg& kemsg,
   }
   if (publicKeys_.keyExists(kemsg.repID, sn)) return "ok";
   publicKeys_.push(kemsg, sn);
-  if (kemsg.repID == repID_) {  // initiated by me
-    private_keys_.key_data().generated = candidate_private_keys_.generated;
+  if (kemsg.repID == repID_ && seq_candidate_map_.count(sn)) {  // initiated by me
+    private_keys_.key_data().generated = seq_candidate_map_[sn].generated;
+    seq_candidate_map_[sn].generated.clear();
     candidate_private_keys_.generated.clear();
+    // erasing seqnum from the map
+    seq_candidate_map_.erase(sn);
     ConcordAssert(private_keys_.key_data().generated.pub == kemsg.pubkey);
     private_keys_.onKeyExchange(cid, sn);
     for (auto e : registryToExchange_) e->onPrivateKeyExchange(private_keys_.key_data().keys[sn], kemsg.pubkey, sn);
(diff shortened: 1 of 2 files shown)


bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -259,8 +262,14 @@ void KeyExchangeManager::generateConsensusKeyAndSendInternalClientMsg(const SeqN
     candidate.pub = pub;
     candidate.cid = cid;
     candidate.sn = sn;
+    seq_candidate_map_[sn] = candidate_private_keys_;
```
yontyon, line 265: you should not generate a new candidate also when you already have the sn in the candidate's map.
(here https://github.com/vmware/concord-bft/blob/23ed4ae7f84571a12801adf0595f2b48d16e38ba/bftengine/src/bftengine/KeyExchangeManager.cpp#L251). 
In that case, we simply want to use the already-generated candidate

yontyon (approved):
Only one comment, please fix it before merging

bftengine/src/bftengine/KeyExchangeManager.cpp
situ-s: made the change.

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -245,7 +248,7 @@ void KeyExchangeManager::generateConsensusKeyAndSendInternalClientMsg(const SeqN
   }
 
   auto& candidate = candidate_private_keys_.generated;
-  bool generateNewPair = (sn != candidate.sn) || candidate.cid.empty();
+  bool generateNewPair = ((sn != candidate.sn) || candidate.cid.empty()) && (seq_candidate_map_.count(sn) == 0);
```
yontyon, line 251: once you `seq_candidate_map_.count(sn) == 0`, you don't need to check `sn != candidate.sn`, right?

situ-s: yes correct, if candidate is not there in map, then sn will never be equal to candidate.sn. I will remove this check.

cjn (approved):

PR #2936, opened 2023-02-14
Optimize digest calculation in pruning callback

mdarade:
<placeholder>
* **Problem Overview**  

Pruning can happen anytime resulting into removal of blocks from storage. RVBManager maintains Range Validation Tree (RVT) which it has to persist during every checkpoint by adding/removing set of Range Validation Block (RVB) nodes. Each such RVB node represents single block at interval (configurable through FRS).

Once pruning is done, digests for pruned blocks would be unavailable forever making it impossible for RVBManager to checkpoint RVT. To mitigate this RVBManager expects a callback whenever pruning is about to begin.

RVBManager was spending enormous time in this callback for calculating and storing digests of to-be-pruned blocks. Existing basic implementation was fetching and calculating digest one block at a time. This fix would parallelize the digest calculation by fetching blocks concurrently from the storage.

* **Testing Done**  
 Executed all tests locally.

### DIFF
--- a/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
+++ b/bftengine/include/bcstatetransfer/SimpleBCStateTransfer.hpp
@@ -38,7 +38,6 @@ namespace bftEngine {
 // can also handle a limited amount of arbitrary mutable state (which is
 // represented as a small set of fixed size pages).
 namespace bcst {
-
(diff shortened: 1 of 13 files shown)


mdarade:
This was a backup PR for 2926.
As 2926 has been merged, closing this PR.

PR #2938, opened 2023-02-16
Add rocksdb column family of keys history to get fast access to snapshots

Efrat1:
Add rocksdb column family of keys history to get fast access to snapshots

Keys: category_name|user_key|big_endian_block_id
Values: empty string

Add unit tests

* **Problem Overview**  
  We need a way to perform a fast read for a version of a value that is closest to a given version.
* **Testing Done**  
  Unit tests, Apollo.

### DIFF
--- a/bftengine/include/bftengine/ReplicaConfig.hpp
+++ b/bftengine/include/bftengine/ReplicaConfig.hpp
@@ -283,6 +283,15 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
                concord::crypto::SignatureAlgorithm::EdDSA,
                "A flag to specify the operator message signing algorithm. It is defaulted to use EDDSA algo.");
 
+  CONFIG_PARAM(keysHistoryMaxBlocksNum,
+               std::uint64_t,
+               10000u,
+               "threshold for pruning from keys history, versions below latest_block_id - this_value can be pruned."
+               "If set to 0, versions below the genesis block will be pruned, i.e. the whole available history will be "
+               "kept (archival).");
+
+  CONFIG_PARAM(enableKeysHistoryCF, bool, true, "Feature flag for keys history column family");
+
   // Parameter to enable/disable waiting for transaction data to be persisted.
   // Not predefined configuration parameters
   // Example of usage:
@@ -418,6 +427,8 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
     serialize(outStream, kvBlockchainVersion);
     serialize(outStream, operatorMsgSigningAlgo);
     serialize(outStream, replicaMsgSigningAlgo);
+    serialize(outStream, keysHistoryMaxBlocksNum);
+    serialize(outStream, enableKeysHistoryCF);
   }
   void deserializeDataMembers(std::istream& inStream) {
     deserialize(inStream, isReadOnly);
@@ -518,6 +529,8 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
     deserialize(inStream, kvBlockchainVersion);
     deserialize(inStream, operatorMsgSigningAlgo);
     deserialize(inStream, replicaMsgSigningAlgo);
+    deserialize(inStream, keysHistoryMaxBlocksNum);
+    deserialize(inStream, enableKeysHistoryCF);
   }
 
  private:
--- a/kvbc/include/Replica.h
+++ b/kvbc/include/Replica.h
@@ -93,6 +93,10 @@ class Replica : public IReplica,
 
   // Get the last block ID in the system.
   BlockId getLastBlockId() const override final;
+
+  std::optional<categorization::Value> getFromSnapshot(const std::string &category_id,
+                                                       const std::string &key,
+                                                       const BlockId snapshot_version) const override final;
   /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
   // checkpoint
   void checkpointInProcess(bool flag, kvbc::BlockId);
--- a/kvbc/include/db_interfaces.h
+++ b/kvbc/include/db_interfaces.h
@@ -81,6 +81,12 @@ class IReader {
   // Get the last block ID in the system.
   virtual BlockId getLastBlockId() const = 0;
 
+  // Get the version of 'key' that is the closest from below or equal to 'snapshot_version'
+  // Return std::nullopt if the key doesn't exist or is deleted at given version
+  virtual std::optional<categorization::Value> getFromSnapshot(const std::string &category_id,
+                                                               const std::string &key,
+                                                               const BlockId snapshot_version) const = 0;
+
   virtual ~IReader() = default;
 };
 
--- a/kvbc/include/kvbc_adapter/v4blockchain/blocks_reader_adapter.hpp
+++ b/kvbc/include/kvbc_adapter/v4blockchain/blocks_reader_adapter.hpp
@@ -72,6 +72,13 @@ class BlocksReaderAdapter : public IReader {
 
   // Get the last block ID in the system.
   BlockId getLastBlockId() const override final { return kvbc_->getLastReachableBlockId(); }
+
+  // Get the version of 'key' that is the closest from below or equal to 'snapshot_version'
+  // Return std::nullopt if the key doesn't exist or is deleted at given version
+  std::optional<concord::kvbc::categorization::Value> getFromSnapshot(
+      const std::string &category_id, const std::string &key, const BlockId snapshot_version) const override final {
+    return kvbc_->getFromSnapshot(category_id, key, snapshot_version);
+  }
   /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
  private:
--- a/kvbc/include/v4blockchain/detail/keys_history.h
+++ b/kvbc/include/v4blockchain/detail/keys_history.h
@@ -1,0 +1,117 @@
+// Concord
+//
+// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+
+#pragma once
+
+#include "rocksdb/native_client.h"
+#include "categorization/updates.h"
+#include "v4blockchain/detail/categories.h"
+#include <rocksdb/compaction_filter.h>
+
+namespace concord::kvbc::v4blockchain::detail {
+
+class KeysHistory {
+ public:
+  KeysHistory(const std::shared_ptr<concord::storage::rocksdb::NativeClient>&,
+              const std::optional<std::map<std::string, concord::kvbc::categorization::CATEGORY_TYPE>>&);
+
+  // Add the keys of the block to the keys history column family
+  void addBlockKeys(const concord::kvbc::categorization::Updates&, const BlockId, storage::rocksdb::NativeWriteBatch&);
+
+  // Return the id of the block where the key was last updated to its value at given version
+  std::optional<BlockId> getVersionFromHistory(const std::string& category_id,
+                                               const std::string& key,
+                                               const BlockId version) const;
+
+  // Delete the last added block keys
+  void revertLastBlockKeys(const concord::kvbc::categorization::Updates&,
+                           const BlockId,
+                           storage::rocksdb::NativeWriteBatch&);
+
+  struct KHCompactionFilter : ::rocksdb::CompactionFilter {
+    static ::rocksdb::CompactionFilter* getFilter() {
+      static KHCompactionFilter instance;
+      return &instance;
+    }
+    KHCompactionFilter() {}
+    const char* Name() const override { return "KeysHistoryCompactionFilter"; }
+    bool Filter(int /*level*/,
+                const ::rocksdb::Slice& key,
+                const ::rocksdb::Slice& /*val*/,
+                std::string* /*new_value*/,
+                bool* /*value_changed*/) const override;
+  };
+
+ private:
+  void handleCategoryUpdates(const std::string& block_version,
+                             const std::string& category_id,
+                             const concord::kvbc::categorization::BlockMerkleInput&,
+                             concord::storage::rocksdb::NativeWriteBatch&);
+  void handleCategoryUpdates(const std::string& block_version,
+                             const std::string& category_id,
+                             const concord::kvbc::categorization::VersionedInput&,
+                             concord::storage::rocksdb::NativeWriteBatch&);
+  void handleCategoryUpdates(const std::string& block_version,
+                             const std::string& category_id,
+                             const concord::kvbc::categorization::ImmutableInput&,
+                             concord::storage::rocksdb::NativeWriteBatch&);
+
+  template <typename UPDATES>
+  void handleUpdatesImp(const std::string& block_version,
+                        const std::string& category_id,
+                        const UPDATES& updates_kv,
+                        concord::storage::rocksdb::NativeWriteBatch& write_batch);
+
+  template <typename DELETES>
+  void handleDeletesUpdatesImp(const std::string& block_version,
+                               const std::string& category_id,
+                               const DELETES& deletes,
+                               concord::storage::rocksdb::NativeWriteBatch& write_batch);
+
+  void revertCategoryKeys(const std::string& category_id,
+                          const categorization::BlockMerkleInput& updates,
+                          const std::string& block_version,
+                          concord::storage::rocksdb::NativeWriteBatch& write_batch);
+  void revertCategoryKeys(const std::string& category_id,
+                          const categorization::VersionedInput& updates,
+                          const std::string& block_version,
+                          concord::storage::rocksdb::NativeWriteBatch& write_batch);
+  void revertCategoryKeys(const std::string& category_id,
+                          const categorization::ImmutableInput& updates,
+                          const std::string& block_version,
+                          concord::storage::rocksdb::NativeWriteBatch& write_batch);
+
+  template <typename UPDATES>
+  void revertKeysImp(const std::string& category_id,
+                     const UPDATES& updates_kv,
+                     const std::string& block_version,
+                     concord::storage::rocksdb::NativeWriteBatch& write_batch);
+
+  template <typename DELETES>
+  void revertDeletedKeysImp(const std::string& category_id,
+                            const DELETES& deletes,
+                            const std::string& block_version,
+                            concord::storage::rocksdb::NativeWriteBatch& write_batch);
+
+  void revertOneKeyAtVersion(const std::string& category_id,
+                             const std::string& block_id,
+                             const std::string& prefix,
+                             const std::string& key,
+                             concord::storage::rocksdb::NativeWriteBatch& write_batch);
+
+ private:
+  std::shared_ptr<concord::storage::rocksdb::NativeClient> native_client_;
+  v4blockchain::detail::Categories category_mapping_;
+};
+
+}  // namespace concord::kvbc::v4blockchain::detail
--- a/kvbc/src/v4blockchain/detail/keys_history.cpp
+++ b/kvbc/src/v4blockchain/detail/keys_history.cpp
@@ -1,0 +1,278 @@
+// Concord
+//
+// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+//
+// This product is licensed to you under the Apache 2.0 license (the
+// "License").  You may not use this product except in compliance with the
+// Apache 2.0 License.
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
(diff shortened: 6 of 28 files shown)


kvbc/src/v4blockchain/detail/latest_keys.cpp
```diff
@@ -226,7 +226,7 @@ void LatestKeys::revertCategoryKeysImp(const std::string& cFamily,
         data = val.GetSelf()->data();
         size = val.GetSelf()->size();
       }
-      // add the previous value , already contains currect version as postfix
```
cloudnoize, line 229: I think I meant to write "correct" :)

kvbc/src/v4blockchain/detail/keys_history.cpp
```diff
@@ -0,0 +1,226 @@
...
+                               storage::rocksdb::NativeWriteBatch& write_batch) {
+  LOG_DEBUG(V4_BLOCK_LOG, "Adding keys of block [" << block_id << "] to the keys history CF");
+
+  last_block_ = block_id;
+  for (const auto& [category_id, category_updates] : updates.categoryUpdates().kv) {
+    std::visit(overloaded{[category_id = category_id, &write_batch, &block_id, this](const auto& updates) {
+                            std::vector<std::string> updates_keys(updates.kv.size());
+                            for (const auto& [k, _] : updates.kv) {
+                              (void)_;  // unused
+                              updates_keys.push_back(k);
```
cloudnoize, line 56: The code is good, and you did a nice trick here, but I think it's better to do as done in the latest_keys handling i.e. overload each category to accept its update.
First, there is redundant copying of data e.g. push back on the vector, and in the handleUpdateImp as well, you can use getSliceArray from the latest_keys to compose your key minimal copies.
and also it will enable us to filter by configuration categories that we don't want to store

kvbc/src/v4blockchain/detail/keys_history.cpp
```diff
@@ -0,0 +1,226 @@
...
+void KeysHistory::handleUpdatesImp(const BlockId block_id,
+                                   const std::string& category_id,
+                                   const std::vector<std::string>& updates_keys,
+                                   concord::storage::rocksdb::NativeWriteBatch& write_batch) {
+  std::string block_key = v4blockchain::detail::Blockchain::generateKey(block_id);
+  ConcordAssertEQ(block_key.size(), sizeof(BlockId));
+  const auto& prefix = category_mapping_.categoryPrefix(category_id);
+  const std::string empty_val = "";
+  for (const auto& k : updates_keys) {
+    auto cf_key = prefix;
```
cloudnoize, line 83: You can use getSliceArray as done in the latst_keys to reduce copies

kvbc/src/v4blockchain/detail/keys_history.cpp
```diff
@@ -0,0 +1,226 @@
...
+              "Reading key " << std::hash<std::string>{}(key) << " not found at version " << version << " category_id "
+                             << category_id << " prefix " << prefix << " key is hex "
+                             << concordUtils::bufferToHex(key.data(), key.size()) << " raw key " << key);
+    return std::nullopt;
+  }
+  auto found_key = itr.key();
+  const size_t total_key_size = found_key.size();
+  auto user_key_with_prefix = found_key.substr(0, total_key_size - sizeof(BlockId));
+
+  if (user_key_with_prefix != prefix + key) {
```
cloudnoize, line 181: can it happen? should it be an assert?

kvbc/include/v4blockchain/detail/keys_history.h
```diff
@@ -0,0 +1,85 @@
...
+                const ::rocksdb::Slice& val,
+                std::string* /*new_value*/,
+                bool* /*value_changed*/) const override;
+  };
+
+ private:
+  std::shared_ptr<concord::storage::rocksdb::NativeClient> native_client_;
+  v4blockchain::detail::Categories category_mapping_;
+  static BlockId last_block_;                            // TODO (efrat): should be atomic?
+  static constexpr int MAX_BLOCKS_NUM_SAVE_KEYS = 1024;  // TODO (efrat): exact number TBD;
```
cloudnoize, line 82: eventually it should be configureable

kvbc/include/v4blockchain/detail/keys_history.h
```diff
@@ -0,0 +1,85 @@
...
+                const ::rocksdb::Slice& key,
+                const ::rocksdb::Slice& val,
+                std::string* /*new_value*/,
+                bool* /*value_changed*/) const override;
+  };
+
+ private:
+  std::shared_ptr<concord::storage::rocksdb::NativeClient> native_client_;
+  v4blockchain::detail::Categories category_mapping_;
+  static BlockId last_block_;                            // TODO (efrat): should be atomic?
```
cloudnoize, line 81: It should be atomic since the compaction filter operates on a different thread.
Do you initialize it on start?
I recall having issues with data members and the compaction filter registration (you need to register it on RocksDB init); I suggest for symmetry defining  std::atomic<BlockId> Blockchain::global_latest_block_id = INVALID_BLOCK_ID;
and handling it as the global_genesis_block_id.

kvbc/src/v4blockchain/detail/latest_keys.cpp
Efrat1: sorry, fixed

kvbc/src/v4blockchain/detail/keys_history.cpp
Efrat1: Thanks, done.
I moved getSliceArray definition to blockchain.h, let me know if you think of better place.

kvbc/src/v4blockchain/detail/keys_history.cpp
Efrat1: In case where the asked key has not defined yet at the given version we will get the closest key (prev key in lexical order). I guess it's a valid scenario so I didn't assert

kvbc/include/v4blockchain/detail/keys_history.h
Efrat1: ok makes sense, I'll do it as a second step.

kvbc/include/v4blockchain/detail/keys_history.h
Efrat1: Done

kvbc/src/v4blockchain/detail/keys_history.cpp
Efrat1: Ok, thanks

kvbc/src/v4blockchain/detail/keys_history.cpp
cloudnoize: oh, got it, cool

kvbc/include/v4blockchain/detail/keys_history.h
```diff
@@ -0,0 +1,124 @@
...
+namespace concord::kvbc::v4blockchain::detail {
+
+class KeysHistory {
+ public:
+  KeysHistory(const std::shared_ptr<concord::storage::rocksdb::NativeClient>&,
+              const std::optional<std::map<std::string, concord::kvbc::categorization::CATEGORY_TYPE>>&);
+  void addBlockKeys(const concord::kvbc::categorization::Updates&, const BlockId, storage::rocksdb::NativeWriteBatch&);
+
+  // get the value and return deserialized value if needed.
+  std::optional<BlockId> getUpdateVersionFromSnapshot(const std::string& category_id,
```
cloudnoize, line 37: Let's rename this method and not use snapshot in the key history class, as it's confusing.

kvbc/src/v4blockchain/v4_blockchain.cpp
```diff
@@ -581,6 +584,16 @@ std::optional<categorization::Value> KeyValueBlockchain::getLatest(const std::st
...
 }
 
+std::optional<categorization::Value> KeyValueBlockchain::getFromSnapshot(const std::string &category_id,
+                                                                         const std::string &key,
+                                                                         const BlockId version) const {
+  auto block_id = keys_history_.getUpdateVersionFromSnapshot(category_id, key, version);
+  if (block_id) {
+    return get(category_id, key, block_id.value());
+  }
+  return std::nullopt;
```
cloudnoize, line 594: nit: I would swap the order of returns, return nullopt if !block_id and the happy path is the end of function return

kvbc/src/v4blockchain/detail/keys_history.cpp
```diff
@@ -0,0 +1,293 @@
...
+  if (native_client_->createColumnFamilyIfNotExisting(v4blockchain::detail::KEYS_HISTORY_CF,
+                                                      KHCompactionFilter::getFilter())) {
+    LOG_INFO(V4_BLOCK_LOG, "Created [" << v4blockchain::detail::KEYS_HISTORY_CF << "] column family for keys history");
+  }
+}
+
+void KeysHistory::addBlockKeys(const concord::kvbc::categorization::Updates& updates,
+                               const BlockId block_id,
+                               storage::rocksdb::NativeWriteBatch& write_batch) {
+  LOG_DEBUG(V4_BLOCK_LOG, "Adding keys of block [" << block_id << "] to the keys history CF");
```
cloudnoize, line 37: let's also add a feature flag, that, when disabled returns without updating the keys

kvbc/include/v4blockchain/detail/keys_history.h
Efrat1: ok, changed to getVersionFromHistory, is that a clear name in your opinion?
maybe getUpdateVersionFromHistory or getChangeVersionFromHistory are better?

kvbc/src/v4blockchain/v4_blockchain.cpp
Efrat1: ok done

kvbc/src/v4blockchain/detail/keys_history.cpp
Efrat1: added enableKeysHistoryCF

kvbc/test/v4blockchain/v4_keys_history_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+  const std::string immutable_category = "immutable";
+};
+
+TEST_F(v4_kvbc, creation) {
+  v4blockchain::detail::KeysHistory keys_history{db, categories};
+  ASSERT_TRUE(db->hasColumnFamily(v4blockchain::detail::KEYS_HISTORY_CF));
+  auto options = db->columnFamilyOptions(v4blockchain::detail::KEYS_HISTORY_CF);
+  ASSERT_NE(options.compaction_filter, nullptr);
+  ASSERT_EQ(std::string(options.compaction_filter->Name()), std::string("KeysHistoryCompactionFilter"));
+}
```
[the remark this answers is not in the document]
Efrat1: Added, thanks!

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -283,6 +283,13 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
                concord::crypto::SignatureAlgorithm::EdDSA,
                "A flag to specify the operator message signing algorithm. It is defaulted to use EDDSA algo.");
 
+  CONFIG_PARAM(keysHistoryMaxBlocksNum,
+               std::uint64_t,
+               1024u,  // TODO (efrat): default value TBD
+               "Max number of blocks to be saved in keys history column family");
```
cloudnoize, line 289: the description is inaccurate, should be something like, "threshold for pruning from keys history, versions below latest_block_id - this_value can be pruned"

kvbc/include/Replica.h
```diff
@@ -93,6 +93,12 @@ class Replica : public IReplica,
 
   // Get the last block ID in the system.
   BlockId getLastBlockId() const override final;
+
+  // Get the value of 'key' in `category_id` as it is in 'snapshot_version'
```
cloudnoize, line 97: "Get the version of `key` that is the closest from below or equal to `snapshot_version`"

kvbc/include/db_interfaces.h
```diff
@@ -81,6 +81,12 @@ class IReader {
   // Get the last block ID in the system.
   virtual BlockId getLastBlockId() const = 0;
 
+  // Get the value of 'key' in `category_id` as it is in 'snapshot_version'
```
cloudnoize, line 84: same as above

kvbc/include/kvbc_adapter/v4blockchain/blocks_reader_adapter.hpp
```diff
@@ -72,6 +72,13 @@ class BlocksReaderAdapter : public IReader {
 
   // Get the last block ID in the system.
   BlockId getLastBlockId() const override final { return kvbc_->getLastReachableBlockId(); }
+
+  // Get the value of 'key' in `category_id` as it is in 'snapshot_version'
```
cloudnoize, line 76: I think it's enough to have the description only in the db_interfaces and v4_blockchain files

kvbc/test/v4blockchain/v4_keys_history_test.cpp
```diff
@@ -0,0 +1,319 @@
...
+#include "bftengine/ReplicaConfig.hpp"
+#include <utility>
+#include <random>
+
+using concord::storage::rocksdb::NativeClient;
+using namespace concord::kvbc;
+using namespace ::testing;
+
+namespace {
+
```
cloudnoize, line 26: Great tests and effort!
Can you please add a test that test shows a flow of a key, for example, if we define key="K":
1 - try to get the version of the K before it was inserted.
2 - insert K at version X
3 - try to get the version of K with version < X
4 - try to get the version of K with version == x 
5 - try to get the version of K with version > x
6 - update K at version Y
7 - try to get the version of K with version < Y and > X
8 - try to get the version of K with version > Y
etc etc

Efrat1: Added, thanks!

kvbc/include/kvbc_adapter/v4blockchain/blocks_reader_adapter.hpp
Efrat1: OK np.

kvbc/include/Replica.h
Efrat1: ok, thanks

bftengine/include/bftengine/ReplicaConfig.hpp
Efrat1: ok, thanks

kvbc/src/v4blockchain/v4_blockchain.cpp
```diff
@@ -581,6 +585,31 @@ std::optional<categorization::Value> KeyValueBlockchain::getLatest(const std::st
...
+                                                                         const BlockId version) const {
+  if (bftEngine::ReplicaConfig::instance().enableKeysHistoryCF) {
+    auto block_id = keys_history_.getVersionFromHistory(category_id, key, version);
+    if (!block_id) {
+      return std::nullopt;
+    }
+    return get(category_id, key, block_id.value());
+  } else {
+    BlockId limit = 0;
+    BlockId max_blocks_to_look = bftEngine::ReplicaConfig::instance().keysHistoryMaxBlocksNum;
```
cloudnoize, line 599: First we need to do getLatest from the latest CF to know the latest version of the key.
If it's higher than "version" then we should perform the iteration.

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -283,6 +283,14 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
                concord::crypto::SignatureAlgorithm::EdDSA,
                "A flag to specify the operator message signing algorithm. It is defaulted to use EDDSA algo.");
 
+  CONFIG_PARAM(keysHistoryMaxBlocksNum,
+               std::uint64_t,
+               1024u,  // TODO (efrat): default value TBD
+               "threshold for pruning from keys history, versions below latest_block_id - this_value can be pruned."
+               "If set to 0, versions below the genesis block will be pruned");
```
cloudnoize, line 290: Can you add to the description:
"If set to 0, versions below the genesis block will be pruned, i.e. the whole available history will be kept (archival)."

kvbc/test/v4blockchain/v4_keys_history_test.cpp
```diff
@@ -0,0 +1,446 @@
...
+  for (BlockId i = 1; i <= total_blocks_number; i++) {
+    opt_val = keys_history.getVersionFromHistory(merkle_category, key1, i);
+    if (i <= num_blocks) {
+      ASSERT_EQ(opt_val, std::nullopt);
+    } else if (i <= additional_blocks) {
+      ASSERT_NE(opt_val, std::nullopt);
+      ASSERT_EQ(opt_val.value(), num_blocks + 1);
+    } else if (i <= more_blocks) {
+      ASSERT_NE(opt_val, std::nullopt);
+      ASSERT_EQ(opt_val.value(), additional_blocks + 1);
```
cloudnoize, line 436: To complete the sequence can you add also a deletion of key1 at the end with deletion version assertion as well, thanks

kvbc/test/v4blockchain/v4_keys_history_test.cpp
```diff
@@ -0,0 +1,446 @@
...
+TEST_F(v4_kvbc, follow_key_history) {
+  v4blockchain::detail::KeysHistory keys_history{db, categories};
+  std::random_device seed;
+  std::mt19937 gen{seed()};                       // seed the generator
+  std::uniform_int_distribution dist{10, 40};     // set min and max
+  const BlockId num_blocks = (BlockId)dist(gen);  // generate number
+  const int num_keys = dist(gen);
+  std::optional<BlockId> opt_val;
+  const auto key1 = "key1";
+
```
cloudnoize, line 363: And also query for key1 when the CF is empty and when it's populated with some keys but not with key1

kvbc/src/v4blockchain/v4_blockchain.cpp
Efrat1: Added the `getLatest` check.
but not exactly as it was in concord - I check for snapshot value also if `getLatest` returns nullopt, for the cases where we look for a key at snapshot x, and that key was deleted at block x+1 (or after).
btw, when key history CF is disabled, if bc returns nullopt for a key at a block, we can't know whether that key was deleted or just wasn't updated at the certain block. So our results are not accurate when we have a deleted key.
[that is not the case for the categorized storage, because its `getLatestVersion()` remember also deletes]

bftengine/include/bftengine/ReplicaConfig.hpp
Efrat1: added thanks

kvbc/test/v4blockchain/v4_keys_history_test.cpp
Efrat1: great, done

kvbc/src/v4blockchain/detail/keys_history.cpp
```diff
@@ -0,0 +1,272 @@
...
+    return std::nullopt;
+  }
+  const auto& prefix = category_mapping_.categoryPrefix(category_id);
+  const auto& version_key = v4blockchain::detail::Blockchain::generateKey(version);
+  std::string get_key = prefix;
+  get_key.append(key);
+  get_key.append(version_key);
+
+  auto itr = native_client_->getIterator(v4blockchain::detail::KEYS_HISTORY_CF);
+  itr.seekAtMost(get_key);
```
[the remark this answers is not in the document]
Efrat1: Right. done

kvbc/test/v4blockchain/v4_keys_history_test.cpp
Efrat1: Added query when CF is empty. already checked when CF contains keys but not key1 (the first for loop)

kvbc/src/v4blockchain/v4_blockchain.cpp
cloudnoize: Nice catch, add it as a comment, and we'll think whether it's worth fixing it

kvbc/test/v4blockchain/v4_blockchain_test.cpp
```diff
@@ -2503,6 +2506,147 @@ TEST_F(v4_kvbc, digest_checks) {
...
+  ASSERT_EQ(std::get<categorization::MerkleValue>(*opt_val).block_id, 2);
+  opt_val = blockchain->getFromSnapshot("merkle", "merkle_key1", 3);
+  ASSERT_TRUE(opt_val);
+  ASSERT_EQ(std::get<categorization::MerkleValue>(*opt_val).block_id, 2);
+  opt_val = blockchain->getFromSnapshot("merkle", "merkle_key3", 3);
+  ASSERT_TRUE(opt_val);
+  ASSERT_EQ(std::get<categorization::MerkleValue>(*opt_val).block_id, 3);
+  opt_val = blockchain->getFromSnapshot("merkle", "merkle_key1", 4);
+  ASSERT_TRUE(opt_val);
+  ASSERT_EQ(std::get<categorization::MerkleValue>(*opt_val).block_id, 2);
```
cloudnoize, line 2556: As discussed, let's add a test case where the key history is enabled, but the version is read from the latest CF

Efrat1: added

PR #2943, opened 2023-02-22
Fix the pyclient path 

shgandhi:


tests/apollo/util/bft.py
```diff
@@ -33,7 +33,7 @@
 
 from util.test_base import repeat_test
 
-sys.path.append(os.path.abspath("util/pyclient"))
+sys.path.append(os.path.abspath("pyclient"))
```
toly-kournik, line 36: @shgandhi , probably the more reliable way, i.e. the one non-dependent on where the test is invoked from is to do something like this:
`sys.path.append(os.path.join(os.path.abspath(__file__), "pyclient")) `

teoparvanov (approved):
LGTM, thanks @shgandhi !

toly-kournik (approved):
LGTM!

PR #2944, opened 2023-02-22
Memory management dev1 final

glevkovich:
* **Problem Overview**  
Final merge for the memory management epic, from development branch. Incliudes different issues:
1) README: improve instruction on how to push images to Docker Hub
2) Change in requestsOfNonPrimary field from raw ptr ClientRequestMsg to unique
3) add incoming messages buffers allocations and releases monitor:
* **Testing Done**  
No new tests added, some of the commits had Maestro runs for stability.

### DIFF
--- a/bftengine/src/bcstatetransfer/Messages.hpp
+++ b/bftengine/src/bcstatetransfer/Messages.hpp
@@ -40,8 +40,12 @@ class MsgType {
 };
 
 struct BCStateTranBaseMsg {
-  BCStateTranBaseMsg(uint16_t type) : type(type) {}
+  BCStateTranBaseMsg(uint16_t type, bool isIncomingMsg = false) : type(type), isIncomingMsg_(isIncomingMsg) {}
+
   uint16_t type;
+  // this flag is for monitoring messages buffer allocs and releases so it's mutable even though incoming ST msgs objs
+  // are const during handling
+  mutable uint8_t isIncomingMsg_;
   // This struct and its derived structs are used to de/serialize message buffers sent over communication channels.
   // Since virtual methods modify the memory layout of a struct, there cannot be any for both this base struct and its
   // inheritors.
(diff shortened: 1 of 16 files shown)


bftengine/src/bcstatetransfer/Messages.hpp
```diff
@@ -152,7 +156,7 @@ struct CheckpointSummaryMsg : public BCStateTranBaseMsg {
   }
 
   static bool equivalent(const CheckpointSummaryMsg* a, uint16_t a_id, const CheckpointSummaryMsg* b, uint16_t b_id) {
-    static_assert((sizeof(CheckpointSummaryMsg) - sizeof(requestMsgSeqNum) == 87),
+    static_assert((sizeof(CheckpointSummaryMsg) - sizeof(requestMsgSeqNum) == 88),
```
eyalrund, line 159: why use a number in code? define a CONST

bftengine/src/bftengine/messages/MessageBase.cpp
```diff
@@ -46,11 +46,66 @@ void MessageBase::printLiveMessages() {
 namespace bftEngine {
 namespace impl {
 
+// static class members for diagnostics server
+std::array<std::atomic<size_t>, MsgCode::LastMsgCodeVal> MessageBase::Statistics::AliveIncomingExtrnMsgsBufs{};
+static_assert(MsgCode::LastMsgCodeVal < 2000,
```
eyalrund, line 51: where 2000 comes from?

eyalrund (approved):

bftengine/src/bcstatetransfer/Messages.hpp
glevkovich: That's how code was written, @hshomroni  nnly changed the number. I'm not sure we need it as a constexpr, since it is used only here. What's the point to define it one line above if it is used only here? I dont see a reason.

bftengine/src/bftengine/messages/MessageBase.cpp
glevkovich: This is an approximation @hshomroni  decided to do after my comment - since he is using some std::array of size MsgCode::LastMsgCodeVal, he must be sure the array wont way too big if someone decides to jump the message codes into a very big number.

PR #2947, opened 2023-02-28
Merkle management layer

HristoStaykov:
* **Problem Overview**  
  An initial infrastructure for building data structures that will serve for verification of the correctness of parts of, or the entire state.
* **Testing Done**  
N/A


kvbc/include/merkle_builder/BaseMerkleBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+
+namespace concord {
+namespace kvbc {
+namespace sparse_merkle {
+
+using Hasher = concord::kvbc::categorization::Hasher;
+
+class BaseMerkleBuilder : public IMerkleBuilder {
+ public:
+  bool needProcessing(char t) {
```
eyalrund, line 27: why t?
const function

HristoStaykov: I will change `t` to `type`. We need to process the keys that are related to Ethereum accounts (balance, nonce, code hash and storage) and we determine the type by the prefix of the key. I have made corrections for const correctness for the helper functions and changed the argument to the function that extracts the address from the key to const reference. Thanks!

kvbc/include/proof_builder/IProofBuilder.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace concord {
+namespace kvbc {
+
+// This is an interface of a general proof builder.
+class IProofBuilder {
+ public:
+  virtual ~IProofBuilder() = default;
+
+  virtual void Init(uint numVersionsStored) = 0;
```
cloudnoize, line 29: can we have a brief description for each method

kvbc/include/proof_builder/IProofBuilder.h
```diff
@@ -0,0 +1,39 @@
...
+
+  virtual void ProcessUpdates(const categorization::Updates& updates) = 0;
+
+  virtual std::string GetProof(const std::string& key) = 0;
+
+  virtual void WaitForScheduledTasks() = 0;
```
cloudnoize, line 35: Is this method for blocking till async ProcessUpdates finishes?

kvbc/include/v4blockchain/detail/verifiability_management_layer.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace concord::kvbc::v4blockchain::detail {
+
+using namespace concord::kvbc;
+
+class VerifiabilityManagementLayer {
+ public:
+  VerifiabilityManagementLayer();
+
+  IProofBuilder* getBuilder() { return proofBuilder_.get(); }
```
cloudnoize, line 33: nice!

cloudnoize: But builder name is confusing :)

kvbc/src/v4blockchain/v4_blockchain.cpp
```diff
@@ -105,6 +105,7 @@ BlockId KeyValueBlockchain::add(categorization::Updates &&updates) {
 BlockId KeyValueBlockchain::add(const categorization::Updates &updates,
                                 v4blockchain::detail::Block &block,
                                 storage::rocksdb::NativeWriteBatch &write_batch) {
+  getVerifiabilityManagementLayer().getBuilder()->ProcessUpdates(updates);
```
cloudnoize, line 108: Doesn't it need the block id as well to know the version?

kvbc/include/merkle_builder/BaseMerkleBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+namespace concord {
+namespace kvbc {
+namespace sparse_merkle {
+
+using Hasher = concord::kvbc::categorization::Hasher;
+
+class BaseMerkleBuilder : public IMerkleBuilder {
+ public:
+  bool needProcessing(char type) const {
+    if (type == kKvbKeyEthBalance || type == kKvbKeyEthCode || type == kKvbKeyEthStorage || type == kKvbKeyEthNonce) {
```
cloudnoize, line 28: You assume here that it's an Ethereum deployment, can we do it without this assumption? maybe have the set of chars/strings to save as input?

kvbc/include/merkle_builder/BaseMerkleBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  address getAddress(const std::string& key) const {
+    ConcordAssert(key.size() > IMerkleBuilder::address_size);  // Include 1 byte indicating key type
```
cloudnoize, line 35: You assume that it's an Ethereum deployment, it will crash if the key starts with with some need processing char but it's not Ethereum

kvbc/include/merkle_builder/BaseMerkleBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+    return address(&key[1], IMerkleBuilder::address_size);
+  }
+
+  virtual void ProcessUpdates(const categorization::Updates& updates) override {
+    for (const auto& k : updates.categoryUpdates().kv) {
+      if (const categorization::BlockMerkleInput* pval = std::get_if<categorization::BlockMerkleInput>(&k.second)) {
```
cloudnoize, line 41: in Ethereum you'll need to update only the Merkle category, maybe in general we can define that only the Merkle category updates the proof module.

kvbc/include/proof_builder/IProofBuilder.h
```diff
@@ -0,0 +1,39 @@
...
+#include <string>
+#include <vector>
+#include "categorization/updates.h"
+#include "util/sliver.hpp"
+
+namespace concord {
+namespace kvbc {
+
+// This is an interface of a general proof builder.
+class IProofBuilder {
```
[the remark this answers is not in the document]
HristoStaykov: Renamed Builder to Processor.

kvbc/include/proof_builder/IProofBuilder.h
HristoStaykov: Yes, in case of async processing of the Merkle updates we need a mechanism to sync and wait for all scheduled jobs to be completed.

kvbc/src/v4blockchain/v4_blockchain.cpp
HristoStaykov: Our implementation of the MPT is not strictly coupled with the block height. Every set of keys/values added to a MPT bumps the version by 1.

kvbc/include/merkle_processor/BaseMerkleProcessor.h
```diff
@@ -0,0 +1,77 @@
...
+#include "kvbc_app_filter/kvbc_key_types.h"
+#include "categorization/base_types.h"
+
+namespace concord {
+namespace kvbc {
+namespace sparse_merkle {
+
+using Hasher = concord::kvbc::categorization::Hasher;
+
+class BaseMerkleProcessor : public IMerkleProcessor {
```
WildFireFlum, line 25: please move implementations to cpp files.
Moreover, the convention is that cpp header files have a hpp extension, no?

kvbc/include/merkle_processor/SyncMerkleProcessor.h
```diff
@@ -0,0 +1,43 @@
...
+  virtual ~SyncMerkleProcessor() = default;
+
+  virtual void Init(uint numVersionsStored) override;
+
+  virtual void BeginVersionUpdateBatch() override;
+  virtual void UpdateAccountTree(const address& addr, const std::string& key, const value& data) override;
+
+  virtual void CommitVersionUpdateBatch() override;
+
+  virtual std::vector<std::string> GetAccountMerkleRootPath(const address& addr) override;
```
WildFireFlum, line 34: what does root path mean? full path from root?

kvbc/include/merkle_processor/SyncMerkleProcessor.h
```diff
@@ -0,0 +1,43 @@
...
+  virtual std::vector<std::string> GetAccountStorageKeyMerklePath(const address& addr, const std::string& key) override;
```
WildFireFlum, line 35: Since the return type here is the same as the previous function, there should be some documentation explaining what the difference between the paths is.

Moreover, the dynamic allocations for the path vectors will be expensive (assuming proofs are requested in the hot path), do we have an upper bound for the path length?

After our last discussion I thought that path parts are 4-bit symbols, shouldn't the type reflect that somehow?

kvbc/include/merkle_processor/SyncMerkleProcessor.h
```diff
@@ -0,0 +1,43 @@
...
+  virtual std::string GetProof(const std::string& key) override;
```
WildFireFlum, line 36: I personally find proofs being of type std::string to be very confusing. If you think of them as byte arrays, use the appropriate type (concord::Byte).

kvbc/include/merkle_processor/IMerkleProcessor.h
```diff
@@ -0,0 +1,52 @@
...
+#include <vector>
+#include "categorization/updates.h"
+#include "proof_processor/IProofProcessor.h"
+#include "util/sliver.hpp"
+
+namespace concord {
+namespace kvbc {
+namespace sparse_merkle {
+
+// This is an interface used for managing a set of Merle trees' storage.
```
WildFireFlum, line 26: So shouldn't its name reflect that?

kvbc/include/proof_processor/EmptyProofProcessor.h
```diff
@@ -0,0 +1,34 @@
...
+
+#pragma once
+
+#include "IProofProcessor.h"
+#include "kvbc_app_filter/kvbc_key_types.h"
+
+namespace concord {
+namespace kvbc {
+
+class EmptyProofProcessor : public IProofProcessor {
```
WildFireFlum, line 21: Is this a test class? If so, shouldn't it be in some test directory? If not, why is it needed?

kvbc/include/v4blockchain/detail/verifiability_management_layer.h
```diff
@@ -0,0 +1,39 @@
...
+#include "v4blockchain/detail/categories.h"
+#include <rocksdb/compaction_filter.h>
+#include "rocksdb/snapshot.h"
+#include "proof_processor/IProofProcessor.h"
+
+namespace concord::kvbc::v4blockchain::detail {
+
+using namespace concord::kvbc;
+
+class VerifiabilityManagementLayer {
```
WildFireFlum, line 29: With such a general name, please add some documentation as to what it verifies and where it gets it from.

Also, currently all it does is encapsulating another object, shouldn't it have additional functions?

kvbc/include/v4blockchain/v4_blockchain.h
```diff
@@ -86,6 +87,9 @@ class KeyValueBlockchain {
   const v4blockchain::detail::Blockchain &getBlockchain() const { return block_chain_; };
   const v4blockchain::detail::StChain &getStChain() const { return state_transfer_chain_; };
   const v4blockchain::detail::LatestKeys &getLatestKeys() const { return latest_keys_; };
+  v4blockchain::detail::VerifiabilityManagementLayer &getVerifiabilityManagementLayer() {
```
WildFireFlum, line 90: Shouldn't this be optional? (Ethereum only?)

kvbc/include/merkle_processor/SyncMerkleProcessor.h
HristoStaykov: This call returns the Merkle path in the Global State MPT from the GS Merkle root to the address's tree Merkle root. The next one - GetAccountStorageKeyMerklePath returns the path from the key, the value of which we want to verify, to the address's tree Merkle root.

kvbc/include/merkle_processor/SyncMerkleProcessor.h
HristoStaykov: Those are proof paths which essentially are a list of hashes. Our last discussion was focused on the Nibbles (4 bit symbols), which are a primitive used to address the sub-tree records in the database and will also be used for the generation of the list of hashes here.

kvbc/include/proof_processor/EmptyProofProcessor.h
HristoStaykov: We want to have the possibility to not build the data structures related to verifiability. In deployments where they will not be required we don't want to pay the performance penalty of building and maintaining them. The current idea for the future is to instantiate the actual Proof processor based on run time configuration.

kvbc/include/merkle_builder/BaseMerkleBuilder.h
HristoStaykov: The general low level part of the MPT is in [sparse_merkle](https://github.com/vmware/concord-bft/tree/master/kvbc/include/sparse_merkle).

kvbc/include/merkle_processor/BaseMerkleProcessor.h
HristoStaykov: We have both .h and .hpp. I have moved the implementation to a .cpp file.

kvbc/include/merkle_processor/SyncMerkleProcessor.h
HristoStaykov: In the response to GetProof for Ethereum we will have a JSON object as defined in [eth-getproof](https://docs.alchemy.com/reference/eth-getproof). What would be the benefit of replacing the std::string with an std::vector<concord::Byte>?

WildFireFlum: The proof itself is defined by the hash bytes. The answer to the client request is these bytes encoded as hex strings. From a design perspective, types should match the held data. From a practical perspective - if we use raw byte arrays, when we need to reuse the proof data, we won't need to decode it from its hex string representation.

kvbc/include/v4blockchain/detail/verifiability_management_layer.h
HristoStaykov: Yes, in the near future it will check the runtime configuration to instantiate the correct ProofProcessor. This is not part of this initial effort though.

kvbc/include/merkle_processor/BaseMerkleProcessor.h
HristoStaykov: Moved implementations to cpp file.

kvbc/include/merkle_processor/IMerkleProcessor.h
HristoStaykov: It is named IMerkleProcessor because the proofs it will maintain and provide are based on Merkle paths.

kvbc/include/v4blockchain/v4_blockchain.h
HristoStaykov: The actual processor of the state will be instantiated based on configuration. See SyncMerkleProcessor.h and EmptyProofProcessor.h

kvbc/include/proof_builder/IProofBuilder.h
```diff
@@ -0,0 +1,39 @@
...
+ public:
+  virtual ~IProofBuilder() = default;
+
+  virtual void Init(uint numVersionsStored) = 0;
+
+  virtual void ProcessUpdates(const categorization::Updates& updates) = 0;
+
+  virtual std::string GetProof(const std::string& key) = 0;
```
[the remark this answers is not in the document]
HristoStaykov: For extracting range proofs we could pass the subpart of the keys that is common. The extraction of the proof will then not reach a LeafChild, but return only a path up to the specified common part of the key. The proof validator can be completely separate form this code.

teoparvanov (approved):
LGTM as a foundation for the upcoming synchronous Merkle tree processor.

PR #2948, opened 2023-02-28
Replicas main key rotation implementation

WildFireFlum:
* **Problem Overview**  
This PR add support for main key rotations on demand (for example, when compromised).
The implementation uses the replica's consensus key as its main key, thus when an operator sends a consensus key rotation request, the resulting state is that both the main key and the consensus key are changed.

* **Testing Done**  
Unit tests and apollo tests (local and remote)


Makefile
```diff
@@ -168,6 +168,10 @@ CONCORD_BFT_ADDITIONAL_RUN_PARAMS+=\
 	--network host
 endif
 
+CONCORD_BFT_ADDITIONAL_CTEST_RUN_PARAMS="--output-on-failure"
+FAILED_CASES_FILE=/concord-bft/build/apollogs/latest/failed_cases.txt
```
yontyon, line 172: please document the new tests output files in the PR

bftengine/include/bftengine/IStateTransfer.hpp
```diff
@@ -89,6 +89,8 @@ class IStateTransfer : public IReservedPages {
   // Accepts the checkpoint number as a parameter.
   // Callbacks must not throw.
   // Multiple callbacks can be added.
+  // TODO: the callbacks should be stored in a queue/priority queue and run in a
+  // predetermined order to prevent repeating logic such as loading main keys from reserved pages
```
yontyon, line 93: If you load the callback only once, how would it executed more than once?

bftengine/include/bftengine/ReplicaConfig.hpp
```diff
@@ -170,6 +170,12 @@ class ReplicaConfig : public concord::serialize::SerializableFactory<ReplicaConf
 
   // Keys Management
   CONFIG_PARAM(keyExchangeOnStart, bool, false, "whether to perform initial key exchange");
+  CONFIG_PARAM(
+      singleSignatureScheme,
+      bool,
+      true,
+      "whether to use a single signature scheme to identify the replica in the network."
+      "When set to false, a replica may use a different signature scheme to sign consensus protocol messages.");
```
yontyon, line 178: A different than what?

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
```diff
@@ -71,15 +72,7 @@ class Communication : public ICommunication {
...
-class InternalSigner : public concord::crypto::ISigner {
- public:
-  size_t signBuffer(const concord::Byte* dataIn, size_t dataLen, concord::Byte* sigOutBuffer) override {
-    return bftEngine::impl::SigManager::instance()->sign(dataIn, dataLen, sigOutBuffer);
-  }
-
-  size_t signatureLength() const override { return bftEngine::impl::SigManager::instance()->getMySigLength(); }
-  std::string getPrivKey() const override { return ""; }
-};
+// Implement cryptomanager update handler
```
yontyon, line 75: I guess you mean to the MainKeyUpdateHandler, if so, please remove the comment. If not, then what does the comment means?

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
```diff
@@ -136,8 +129,41 @@ class ScalingReplicaHandler : public IStateHandler {
     return logger_;
   }
 };
-std::shared_ptr<ClientReconfigurationEngine> CreFactory::create(std::shared_ptr<MsgsCommunicator> msgsCommunicator,
-                                                                std::shared_ptr<MsgHandlersRegistrator> msgHandlers) {
+
+// TODO(yf): remove
```
yontyon, line 133: remove the TODO, this handler is used by the read-only replica

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
```diff
@@ -136,8 +129,41 @@ class ScalingReplicaHandler : public IStateHandler {
...
+    }
+    return false;
+  }
+
+  bool execute(const State& state, WriteState&) override {
+    LOG_INFO(getLogger(), "execute MainKeyUpdateHandler");
+    concord::messages::ClientStateReply crep;
+    concord::messages::deserialize(state.data, crep);
+    concord::messages::ReplicaMainKeyUpdate update = std::get<concord::messages::ReplicaMainKeyUpdate>(crep.response);
+    bftEngine::CryptoManager::instance().onPublicKeyExchange(update.key, update.sender_id, update.seq_num);
```
yontyon, line 152: Is there a way to get the latest sender_id key's sequence number?
I wonder if it is possible already now to handle the following scenario:
The ROR has the main key of replica X for sequence number S1. However, as it reads the state only from f+1 replicas, then f Byzantine replicas + 1 late replica may give it X's key for sequence number S2 < S1.
If it is possible to get from Cryptomanager X's latest key sequence number, then it can be solved at the handler level.

On the other hand, we may need to handle globally, for all AsyncStateTransferCRE handlers...
WDYT?

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -341,16 +341,16 @@ void DbCheckpointManager::sendInternalCreateDbCheckpointMsg(const SeqNum& seqNum
...
   req.noop = noop;
-  std::vector<uint8_t> data_vec;
+  std::vector<concord::Byte> data_vec;
   concord::messages::db_checkpoint_msg::serialize(data_vec, req);
-  std::string sig(SigManager::instance()->getMySigLength(), '\0');
-  SigManager::instance()->sign(reinterpret_cast<char*>(data_vec.data()), data_vec.size(), sig.data());
-  req.signature = std::vector<uint8_t>(sig.begin(), sig.end());
+  // TODO: this signature does not appear to be validated anywhere in the codebase
+  req.signature.resize(SigManager::instance()->getMySigLength());
+  SigManager::instance()->sign(seqNum, data_vec.data(), data_vec.size(), req.signature.data());
```
yontyon, line 348: As far as I understand, the method will be called at every checkpoint. 
This method is being called as part of the stable checkpoint callbacks (see https://github.com/vmware/concord-bft/blob/c769f077da98a4ef6f5ae8fa859d4618c70552d5/bftengine/src/bftengine/ReplicaImp.cpp#L5356).
Now, changing the cryptosystem also happens on a (stable ?) checkpoint, right? if so, does the order of the callbacks matter? 
Could it be an issue if we first sign the dbcheckpoint message with the old cryptosystem and only then update the cryptosystem? 
Your comment says that there is no validation of this message anywhere in the code, but if it were, will the other replicas try to validate it with the new or with the old cryptosystem (and are the meanings)?

bftengine/src/bftengine/KeyStore.cpp
```diff
@@ -38,7 +39,19 @@ std::optional<ClusterKeyStore::PublicKeys> ClusterKeyStore::loadReplicaKeyStoreF
     std::istringstream iss(buffer_);
     PublicKeys ks;
     PublicKeys::deserialize(iss, ks);
-    for (auto [sn, pk] : ks.keys) LOG_DEBUG(KEY_EX_LOG, "rid: " << repID << " seqnum: " << sn << " pubkey: " << pk);
+    // std::vector<SeqNum> keysToRemove;
+    for (auto [sn, pk] : ks.keys) {
+      /*if (ks.keys.size() - keysToRemove.size() > 2) {
```
yontyon, line 44: remove the comment

bftengine/src/bftengine/KeyStore.cpp
```diff
@@ -38,7 +39,19 @@ std::optional<ClusterKeyStore::PublicKeys> ClusterKeyStore::loadReplicaKeyStoreF
...
+        keysToRemove.push_back(sn);
+      }*/
+      LOG_INFO(KEY_EX_LOG, "Deserialized public key from reserved pages: " << KVLOG(repID, sn, pk));
+    }
+
+    /*for (auto sn : keysToRemove) {
```
yontyon, line 50: remove the comment

bftengine/src/bftengine/KeyStore.h
```diff
@@ -35,6 +35,7 @@ class ClusterKeyStore : public ResPagesClient<ClusterKeyStore> {
       auto res = keys.insert(std::make_pair(sn, pub));
       if (!res.second) ConcordAssert(pub == res.first->second)  // if existed expect same key
     }
+    void remove(const SeqNum sn) { ConcordAssertEQ(keys.erase(sn), 1); }
```
yontyon, line 38: Do we use this method in practice?

bftengine/src/bftengine/ReadOnlyReplica.cpp
```diff
@@ -31,6 +31,7 @@
 #include "util/json_output.hpp"
```
yontyon, line 31: nit: for what is the permissions change?

bftengine/src/bftengine/Reconfiguration.cpp
```diff
@@ -33,7 +33,8 @@ bool ReconfigurationHandler::handle(const WedgeCommand& cmd,
                                     uint32_t,
                                     const std::optional<bftEngine::Timestamp>&,
                                     concord::messages::ReconfigurationResponse&) {
-  LOG_INFO(getLogger(), "Wedge command instructs replica to stop at sequence number " << bft_seq_num);
+  LOG_INFO(getLogger(),
+           "Wedge command instructs replica to stop at next checkpoint after sequence number " << bft_seq_num);
```
yontyon, line 37: nit: its the next next checkpoint (for example, if bft_seq_num is 10, then the wedpoint is 300)

bftengine/src/bftengine/ReplicaBase.hpp
```diff
@@ -80,11 +81,17 @@ class ReplicaBase {
...
+  template <typename MessageType>
+  bool validateMessage(MessageType* msg) {
+    if (config_.debugStatisticsEnabled) {
+      DebugStatistics::onReceivedExMessage(msg->type());
+    }
     try {
-      if (config_.debugStatisticsEnabled) DebugStatistics::onReceivedExMessage(msg->type());
-
-      msg->validate(*repsInfo);
+      if constexpr (std::is_same_v<MessageType, CheckpointMsg>) {
```
yontyon, line 90: It is a good idea to document why the checkpoint message is handled differently.

bftengine/src/bftengine/ReplicaBase.hpp
```diff
@@ -80,11 +81,17 @@ class ReplicaBase {
...
+        msg->validate(*repsInfo, false);
```
yontyon, line 91: Another question:
We have this mechanism for async validation (see https://github.com/vmware/concord-bft/blob/c769f077da98a4ef6f5ae8fa859d4618c70552d5/bftengine/src/bftengine/ReplicaImp.cpp#L273.
As far as I understand from the code, if async validation is enabled, then the checkpoint message is still being validated. If this is the case, then we need either (1) to fix this for the async validation also or (2) to make sure that checkpoint messages are never async validated (by making sure https://github.com/vmware/concord-bft/blob/c769f077da98a4ef6f5ae8fa859d4618c70552d5/bftengine/src/bftengine/messages/CheckpointMsg.hpp#L73 returns false and asserting on it).
Still, I think I'm missing something, because currently, https://github.com/vmware/concord-bft/blob/c769f077da98a4ef6f5ae8fa859d4618c70552d5/bftengine/src/bftengine/messages/CheckpointMsg.hpp#L73 returns true, and the relevant configuration variable is true by default https://github.com/vmware/concord-bft/blob/c769f077da98a4ef6f5ae8fa859d4618c70552d5/bftengine/include/bftengine/ReplicaConfig.hpp#L209.
Can you clarify?

bftengine/src/bftengine/ReplicaForStateTransfer.cpp
```diff
@@ -70,18 +73,44 @@ ReplicaForStateTransfer::ReplicaForStateTransfer(const ReplicaConfig &config,
 }
 
 void ReplicaForStateTransfer::start() {
-  cre_ = bftEngine::bcst::asyncCRE::CreFactory::create(msgsCommunicator_, msgHandlers_);
+  /* TODO: Transaction signing needs to be disabled for replicas, as state transfer may change
```
yontyon, line 76: https://github.com/vmware/concord-bft/pull/2948/files?diff=split&w=0#r1122745971

bftengine/src/bftengine/ReplicaForStateTransfer.cpp
```diff
@@ -70,18 +73,44 @@ ReplicaForStateTransfer::ReplicaForStateTransfer(const ReplicaConfig &config,
...
+
+          // Make sure to sign the reconfiguration client messages using the key
+          // other replicas expect
+          SigManager::instance()->setReplicaLastExecutedSeq(checkpoint * checkpointWindowSize);
+
+          // Need to update private key to match the loaded public key in case they differ (key exchange was executed
+          // on other replicas but not on this one, finishing ST does not mean that missed key exchanges are executed)
+          // This can be done by iterating the saved cryptosystems and updating their private key if their
+          // public key matches the candidate saved in KeyExchangeManager
+          // TODO: persist the candidate
```
yontyon, line 100: This TODO belongs to key exchange manager where we actually produce the candidate

bftengine/src/bftengine/ReplicaForStateTransfer.cpp
```diff
@@ -70,18 +73,44 @@ ReplicaForStateTransfer::ReplicaForStateTransfer(const ReplicaConfig &config,
...
+          CryptoManager::instance().onCheckpoint(checkpoint);
+          auto [priv, pub] = KeyExchangeManager::instance().getCandidateKeyPair();
```
yontyon, line 102: Notice that this PR https://github.com/vmware/concord-bft/pull/2932 enables multiple candidates. 
So (1) the getCandidateKeyPair should be changed accordingly (maybe return all the existing candidates) and (2) you need to try and match the correct candidate with the one that we have on the reserved pages.

Another general comment:
https://github.com/vmware/concord-bft/blob/c769f077da98a4ef6f5ae8fa859d4618c70552d5/bftengine/src/bftengine/KeyExchangeManager.cpp#L98
does multiple things (like clearing the candidate), persisting the private key, and more. Maybe, it is worth simply calling it directly here such that we won't miss anything.

bftengine/src/bftengine/ReplicaForStateTransfer.cpp
```diff
@@ -94,12 +123,11 @@ void ReplicaForStateTransfer::start() {
                 succ = false;
                 break;
               }  // else if (!isGettingBlocks)
+              // 2. Now we can safely halt cre. We know for sure that there are no update in the state transferred
+              // blocks that haven't been handled yet
+              cre_->halt();
```
yontyon, line 128: I think you need to move this outside of the while loop. 
Here you simply halt the cre on the very first loop.

bftengine/src/bftengine/ReplicasInfo.cpp
```diff
@@ -55,6 +55,7 @@ ReplicasInfo::ReplicasInfo(const ReplicaConfig& config,
                            bool dynamicCollectorForPartialProofs,
                            bool dynamicCollectorForExecutionProofs)
     : _myId{config.replicaId},
+      _isRoReplica{(_myId >= config.numReplicas) && (_myId < config.numReplicas + config.numRoReplicas)},
```
yontyon, line 58: There is a configuration variable in config that specifies if the replica is a read-only, please use it instead

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -56,6 +56,14 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
                              concordUtils::SpanWrapper& parent_span) {
   bool has_pruning_request = false;
   for (auto& req : requests) {
+    LOG_INFO(GL,
```
yontyon, line 59: I suggest moving this to DEBUG (this will be print on every request execution, in a real work load we may have dozens - hundreds requests in a batch

bftengine/src/bftengine/RequestHandler.cpp
```diff
@@ -72,6 +80,9 @@ void RequestHandler::execute(IRequestsHandler::ExecutionRequestsQueue& requests,
     } else if (req.flags & MsgFlag::RECONFIG_FLAG) {
       ReconfigurationRequest rreq;
       deserialize(std::vector<std::uint8_t>(req.request, req.request + req.requestSize), rreq);
+      LOG_INFO(
```
yontyon, line 83: same

WildFireFlum: Shouldn't reconfiguration requests be rare?

bftengine/src/bftengine/RequestHandler.cpp
WildFireFlum: Done

bftengine/src/bftengine/ReplicaForStateTransfer.cpp
WildFireFlum: moved

bftengine/src/bftengine/ReplicaBase.hpp
WildFireFlum: I think that you missed that asyncValidateMessage was already changed to delegate to validateMessage in this branch.
Please resolve this conversation if you are satisfied with this answer.

bftengine/src/bftengine/ReplicaBase.hpp
WildFireFlum: done

bftengine/src/bftengine/Reconfiguration.cpp
WildFireFlum: done

bftengine/src/bftengine/ReadOnlyReplica.cpp
WildFireFlum: These are git's default permissions, why should this file have execution permissions?

bftengine/src/bftengine/KeyStore.h
WildFireFlum: No, I removed it.

bftengine/src/bftengine/KeyStore.cpp
WildFireFlum: done

bftengine/src/bftengine/KeyStore.cpp
WildFireFlum: done

bftengine/src/bftengine/DbCheckpointManager.cpp
WildFireFlum: There is no "changing the cryptosystem".
A new cryptosystem is created when the execution handler for the key exchange is called, say in sequence s, and say the new key pair is k.
Once checkpoint (s+(150*2)) /150 is reached, the old cryptosystem which has k is removed (in CryptoManager::OnCheckpoint).
Once a stable checkpoint is reached, we use its sequence number to decide which cryptosystem to use.
If the stable checkpoint is too far behind, this message can be signed using some different new key (k' with s' > s).

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.hpp
```diff
@@ -17,11 +18,30 @@
...
  public:
   static std::shared_ptr<concord::client::reconfiguration::ClientReconfigurationEngine> create(
-      std::shared_ptr<MsgsCommunicator> msgsCommunicator, std::shared_ptr<MsgHandlersRegistrator> msgHandlers);
+      std::shared_ptr<MsgsCommunicator> msgsCommunicator,
+      std::shared_ptr<MsgHandlersRegistrator> msgHandlers,
+      std::unique_ptr<concord::crypto::ISigner> transactionSigner);
 };
+
+/*
+ * TODO: Transaction signing needs to be disabled for replicas, as state transfer may change
```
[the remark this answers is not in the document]
WildFireFlum: Guy said that state transfer should not include/rely on any signature validation, only on a valid communication channel.

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
WildFireFlum: Done

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
WildFireFlum: removed

bftengine/include/bftengine/ReplicaConfig.hpp
WildFireFlum: changed the documentation

bftengine/include/bftengine/IStateTransfer.hpp
WildFireFlum: I repeated the logic which parses the reserved pages because I needed it to happen prior to the async CRE fetching the state. This could have been avoided had I the option to guarantee the execution order of the callbacks.

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
```diff
@@ -162,7 +162,9 @@ void ViewChangeMsg::finalizeMessage() {
   // |               Message Body               |
   // +------------------------------------------+
 
-  sigManager->sign(body(), bodySize, body() + bodySize);
+  // TODO: since there are multiple message with different sequences, make sure that lastStable()
```
yontyon, line 165: AFAIU if a replica is signing with its latest known sequence number of the view, then it is the correct cryptosystem for sure.
Assuming the last stable is X for replica R. so the R's working window is in between [X, X + 300]. Also, we do know that R is synced up to X. 
The only gap is if the other replicas have committed R's new key between [X, X + 300], but even so, the key won't be effective before X + 600.
If X is too far from the current working window, then R will first perform state transfer, sync with the state (and the correct key), and then the above also holds.
Hence, I think the code is correct and the TODO can be removed.

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
```diff
@@ -81,9 +81,8 @@ std::optional<std::string> PreProcessResultMsg::validatePreProcessResultSignatur
   for (const auto& sig : sigs) {
     bool verificationResult = false;
     if (myReplicaId == sig.sender_replica) {
-      std::vector<uint8_t> mySignature(sigManager_->getMySigLength(), '\0');
-      sigManager_->sign(hash.data(), hash.size(), mySignature.data());
-      verificationResult = mySignature == sig.signature;
+      // TODO: why didn't we just validate this signature normally?
```
yontyon, line 84: Is this todo still relevant?

bftengine/src/preprocessor/tests/preprocessor_test.cpp
```diff
@@ -806,18 +671,18 @@ TEST(requestPreprocessingState_test, batchCancelledNoConsensusReached) {
         std::make_unique<PreProcessBatchReplyMsg>(clientId, senderId, repliesList, cid, overallRepliesSize, viewNum));
   }
 
-  usleep(reqTimeoutMilli * 1000);
+  usleep(reqTimeoutMilli * 10000);
```
yontyon, line 674: Is this timeout increase necessary?

client/bftclient/src/bft_client.cpp
```diff
@@ -70,9 +72,11 @@ Client::Client(SharedCommPtr comm, const ClientConfig& config, std::shared_ptr<c
   communication_->start();
   if (config_.replicas_master_key_folder_path.has_value()) {
     bft::communication::StateControl::instance().setGetPeerPubKeyMethod([&](uint32_t rid) {
+      // The client reads a public key file which the cre client which runs on the same machine maintains
+      // the cre client updates the file after polling updates about replica key exchanges
       concord::secretsmanager::SecretsManagerPlain psm_;
       std::string key_path = config_.replicas_master_key_folder_path.value() + "/" + std::to_string(rid) + "/pub_key";
-      return psm_.decryptFile(key_path).value_or("");
+      return std::array<std::string, replicaIdentityHistoryCount>{psm_.decryptFile(key_path).value_or(""), ""};
```
yontyon, line 79: If we keep only the latest key (which makes sense of course) why do we need to return an array of size 2 where the second element is always an empty one?

client/reconfiguration/src/poll_based_state_client.cpp
```diff
@@ -29,7 +29,14 @@ concord::messages::ReconfigurationResponse PollBasedStateClient::sendReconfigura
   concord::messages::ReconfigurationResponse rres;
   try {
     if (read_request) {
-      bft::client::ReadConfig read_config{request_config, bft::client::LinearizableQuorum{}};
+      // TODO: State transfer can work with f + 1 as long as there are no byzantine replicas
```
yontyon, line 32: Most readers won't understand this comment.
Can you add that the poll_based_state_client is also used by state transfer as an async reconfiguration update mechanism?

kvbc/include/reconfiguration_kvbc_handler.hpp
```diff
@@ -314,7 +314,8 @@ class InternalPostKvReconfigurationHandler : public concord::reconfiguration::IR
   InternalPostKvReconfigurationHandler(kvbc::IBlockAdder& block_adder, kvbc::IReader& ro_storage)
       : ReconfigurationBlockTools{block_adder, ro_storage} {}
   bool verifySignature(uint32_t sender_id, const std::string& data, const std::string& signature) const override {
-    return true;
+    if (!bftEngine::impl::SigManager::instance()->hasVerifier(sender_id)) return false;
```
yontyon, line 317: As an internal handler, this handler should serve only messages which are coming from replicas. Can you here another validation about sender_id is a replica id?

kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -441,6 +443,7 @@ concord::messages::ClientStateReply KvbcClientReconfigurationHandler::buildRepli
           } else if (command_type == std::string{kvbc::keyTypes::reconfiguration_rep_main_key}) {
             concord::messages::ReplicaMainKeyUpdate cmd;
             concord::messages::deserialize(data_buf, cmd);
+            cmd.seq_num -= (cmd.seq_num < 2 * checkpointWindowSize ? 0 : 2 * checkpointWindowSize);
```
yontyon, line 446: I think the cmd seq_num should reflect the same in both, the original message and in the state reply.
The checkpoint calculation from the given sequence number can be done on the receiver side (or the original message creator should put there a checkpoint message and not a sequence number, and the filed name should be changed accordingly)

bftengine/src/preprocessor/tests/preprocessor_test.cpp
WildFireFlum: yes, this test failed on my machine, and as it doesn't test performance, it shouldn't be so sensitive.

bftengine/src/preprocessor/messages/PreProcessResultMsg.cpp
WildFireFlum: Yes

tools/KeyfileIOUtils.cpp
```diff
@@ -130,9 +135,20 @@ Cryptosystem* inputReplicaKeyfileMultisig(const std::string& filename, ReplicaCo
 
   if (config.isReadOnly) return nullptr;
 
+  // TODO(yf): protect this code with a feature flag
```
yontyon, line 138: Is this todo still relevant?

Makefile
yontyon: In the PR description, sorry

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -31,81 +36,72 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
-  std::unique_ptr<Cryptosystem>& getLatestCryptoSystem() const { return cryptoSystems_.rbegin()->second->cryptosys_; }
+
+  static std::shared_ptr<CryptoManager> s_cm;
+
+  static std::shared_ptr<CryptoManager> init(std::unique_ptr<Cryptosystem>&& cryptoSys);
+  static CryptoManager& instance();
+  static void reset(std::shared_ptr<CryptoManager> other);
+
+  std::shared_ptr<IThresholdSigner> thresholdSignerForSlowPathCommit(const SeqNum sn) const;
+  std::shared_ptr<IThresholdVerifier> thresholdVerifierForSlowPathCommit(const SeqNum sn) const;
```
eyalrund, line 47: Do we still use the slowPath signer? does it use the same key?

bftengine/include/bftengine/KeyExchangeManager.hpp
```diff
@@ -31,27 +32,34 @@ class KeyExchangeManager {
...
-  void sendInitialKey(const ReplicaImp* repImpInstance, const SeqNum& = 0);
-  // The execution handler implementation that is called when a key exchange msg has passed consensus.
+
+  void waitForQuorum(const ReplicaImp* repImpInstance);
+
+  // Waits for a quorum and calls generateConsensusKeyAndSendInternalClientMsg
+  void waitForQuorumAndTriggerConsensusExchange(const ReplicaImp* repImpInstance, const SeqNum s = 0);
+  // The execution handler implementation that is called after a key exchange msg passed consensus.
+  // The new key pair will be used from two checkpoints after kemsg.generated_sn
   std::string onKeyExchange(const KeyExchangeMsg& kemsg, const SeqNum& req_sn, const std::string& cid);
```
eyalrund, line 46: I think keyRotatation is the correct term

bftengine/include/bftengine/KeyExchangeManager.hpp
```diff
@@ -31,27 +32,34 @@ class KeyExchangeManager {
...
   // whether initial key exchange has occurred
 
-  bool exchanged() const {
-    uint32_t liveClusterSize = ReplicaConfig::instance().waitForFullCommOnStartup ? clusterSize_ : quorumSize_;
-    bool exchange_self_keys = publicKeys_.keyExists(ReplicaConfig::instance().replicaId);
-    return ReplicaConfig::instance().getkeyExchangeOnStart()
-               ? (publicKeys_.numOfExchangedReplicas() + 1 >= liveClusterSize) && exchange_self_keys
-               : true;
-  }
+  bool exchanged() const;
```
eyalrund, line 62: please replace with more informative name

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
```diff
@@ -154,14 +177,31 @@ std::shared_ptr<ClientReconfigurationEngine> CreFactory::create(std::shared_ptr<
   bftClientConf.replicas_master_key_folder_path = std::nullopt;
   std::unique_ptr<ICommunication> comm = std::make_unique<Communication>(msgsCommunicator, msgHandlers);
   bft::client::Client* bftClient = new bft::client::Client(std::move(comm), bftClientConf);
```
eyalrund, line 179: can we use a smart pointer?

bftengine/src/bftengine/CryptoManager.cpp
```diff
@@ -0,0 +1,245 @@
...
+                                                   checkpointUpperBound,
+                                                   checkpointCandidate,
+                                                   cryptosystem,
+                                                   checkpointToSystem().size(),
+                                                   checkpointWindowSize));
+      return checkpointCandidate;
+    }
+  }
+
+  // Replicas might encounter old messages, crashing is not
```
yontyon, line 187: crashing is not what?

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
```diff
@@ -396,7 +396,7 @@ class CollectorOfThresholdSignatures {
           // if verification failed again
           // signer index starts with 1, therefore shareId-1
```
eyalrund, line 397: if this is not true anymore remove the comment

bftengine/src/bftengine/CryptoManager.cpp
```diff
@@ -0,0 +1,245 @@
+// Concord
+//
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
```
eyalrund, line 3: change to 2023

bftengine/src/bftengine/CryptoManager.cpp
```diff
@@ -0,0 +1,245 @@
...
+    auto currentKey = cryptoSystem->cryptosys_->getMyVerificationKey();
+    LOG_INFO(logger(), "checking keys after ST" << KVLOG(checkpoint, currentKey, secretKey, verificationKey));
+    if (currentKey == verificationKey) {
+      cryptoSystem->cryptosys_->updateKeys(secretKey, verificationKey);
+      LOG_INFO(logger(), "Updated private key for cryptosystem with checkpoint:" << checkpoint);
+    }
+  }
+}
+
+void CryptoManager::onPrivateKeyExchange(const std::string& secretKey,
```
eyalrund, line 84: onPrivateKeyRotation

bftengine/src/bftengine/CryptoManager.cpp
```diff
@@ -0,0 +1,245 @@
...
+                                         const std::string& verificationKey,
+                                         const SeqNum& sn) {
+  LOG_INFO(logger(), "Private key exchange:" << KVLOG(sn, verificationKey));
+  auto sys_wrapper = create(sn);
+  sys_wrapper->cryptosys_->updateKeys(secretKey, verificationKey);
+  sys_wrapper->init();
+}
+
+void CryptoManager::onPublicKeyExchange(const std::string& verificationKey,
```
eyalrund, line 93: same

bftengine/src/bftengine/CryptoManager.cpp
```diff
@@ -0,0 +1,245 @@
...
+    result[1] = {
+        riter->first,
+        std::reinterpret_pointer_cast<EdDSAMultisigVerifier>(riter->second->thresholdVerifierForOptimisticCommit_)};
+  }
+  return result;
+}
+
+std::array<std::shared_ptr<EdDSAMultisigSigner>, 2> CryptoManager::getLatestSigners() const {
+  std::lock_guard<std::mutex> guard(mutex_);
+  auto riter = checkpointToSystem().rbegin();
```
eyalrund, line 147: do you need to check if riter is valid?

bftengine/src/bftengine/CryptoManager.cpp
eyalrund: an option :-)

Makefile
WildFireFlum: Done

bftengine/src/bftengine/CryptoManager.cpp
yontyon, line 187: Assuming we couldn't find a cryptosystem that matches the given sequence number (for example, we got a message from a lagged replica).
Why does it make sense to take the newest one (rbegin) on top of the oldest one?

bftengine/src/bftengine/CryptoManager.cpp
```diff
@@ -0,0 +1,245 @@
...
+  // Cryptosystem for this sn will be activated upon reaching a second checkpoint from now
+  uint64_t chckp = (sn / checkpointWindowSize) + 2;
+  if (auto it = checkpointToSystem().find(chckp); it != checkpointToSystem().end()) {
+    LOG_WARN(logger(), "Cryptosystem already exists for checkpoint " << chckp);
+    return it->second;
+  }
+
+  // copy construct new Cryptosystem from a last one as we want it to include all the existing keys
+  std::unique_ptr<Cryptosystem> cs =
+      std::make_unique<Cryptosystem>(*checkpointToSystem().rbegin()->second->cryptosys_.get());
```
yontyon, line 213: Will we always create the cryptosystem in ascending order?
If not, wouldn't it make sense to take the cryptosystem of the largest sequence number that is still smaller than the given sequence number?

bftengine/src/bftengine/CryptoManager.cpp
WildFireFlum: checkpointToSystem ensures that the size is valid, the mutex ensures that this is the only access - the iterator is expected to be valid.
As for checking that the cryptosystem is not null, I can add this check.

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -90,36 +117,33 @@ std::string KeyExchangeManager::onKeyExchange(const KeyExchangeMsg& kemsg,
...
     initial_exchange_ = true;
     if (ReplicaConfig::instance().getkeyExchangeOnStart() &&
-        ReplicaConfig::instance().publishReplicasMasterKeyOnStartup)
+        ReplicaConfig::instance().publishReplicasMasterKeyOnStartup) {
       sendMainPublicKey();
+      sentKeyUpdateToExecution = true;
+    }
+  }
+
+  // The key was exchanged successfully, write a block containing the new key to the chain
```
yontyon, line 142: The preferred way to do it would be to have both functions in the same message (in case this feature is enabled).
I see the complications of doing so (because of multiple mechanisms (such as key exchange, reconfiguration), and so on).
The potential problem I can see with two different mechanisms is that we have a gap between the consensus exchange and the main key exchange.
For example. assuming the system is highly overloaded and the main key message is executed after X >>> 300 sequence numbers, will it cause any issues?
IMO, it is safe (for now), because:
1. For the replicas, the actual exchange happens already now, the mainKeyPublicKey message is meant to clients (as they fetch the key state using CRE)
2. Clients, however, don't care about consensus keys, so they shouldn't care about this gap.

Do you think it worth a comment here, only to say there might be an uncontrollable gap between the two executions?

bftengine/src/bftengine/KeyExchangeManager.cpp
```diff
@@ -214,29 +241,31 @@ void KeyExchangeManager::exchangeTlsKeys(const SeqNum& bft_sn) {
   exchangeTlsKeys("server", bft_sn);
   exchangeTlsKeys("client", bft_sn);
   metrics_->tls_key_exchange_requests_++;
-  bft::communication::StateControl::instance().restartComm(repID_);
   LOG_INFO(KEY_EX_LOG, "Replica has generated a new tls keys");
+  bft::communication::StateControl::instance().restartComm(repID_);
+  LOG_INFO(KEY_EX_LOG, "Replica communication restarted after tls exchange");
 }
 void KeyExchangeManager::sendMainPublicKey() {
+  auto [seq, latestPublicKey] = SigManager::instance()->getMyLatestPublicKey();
```
yontyon, line 249: If the feature is disabled, don't we need to generate a new main key first?

bftengine/src/bftengine/ReplicaBase.hpp
yontyon: thanks

bftengine/src/bftengine/KeyExchangeManager.cpp
WildFireFlum: There are no rotations when the feature is disabled, but this is a bug because we this code will crash when the feature is turned off but the publish-main-key-on-startup flag is on. Nice catch.

bftengine/src/bftengine/KeyExchangeManager.cpp
WildFireFlum: I think that the issue is the lack of atomicity in this operation, which affects both clients and the current replica state (key exchange reflected in reserved pages but not in block storage).
I mentioned this issue in the following lines of this comment.

bftengine/src/bftengine/CryptoManager.cpp
WildFireFlum: Both will probably not be able to validate, the code for getting the most recent one was the simplest.
I don't mind changing it either way.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -2808,6 +2800,10 @@ void ReplicaImp::onMessage<ReplicaStatusMsg>(std::unique_ptr<ReplicaStatusMsg> m
     }
   }
 
+  /*if (msgLastStable > lastStableSeqNum + (checkpointWindowSize * 2) && !isCollectingState()) {
```
yontyon, line 2803: remove this comment

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5702,6 +5685,7 @@ void ReplicaImp::sendResponses(PrePrepareMsg *ppMsg, IRequestsHandler::Execution
                                                                         req.outReplicaSpecificInfoSize,
                                                                         executionResult);
         if (replyMsg) {
+          LOG_INFO(GL, "Sending reply for req " << req.requestSequenceNum);
```
yontyon, line 5688: move to debug (or remove). This will be sent for every client request

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5798,66 +5782,9 @@ void ReplicaImp::executeNextCommittedRequests(concordUtils::SpanWrapper &parent_
...
-                                    RECONFIG_FLAG,
-                                    current_time,
-                                    strMsg.size(),
-                                    strMsg.c_str(),
-                                    60000,
-                                    "wedge-noop-command-" + std::to_string(lastExecutedSeqNum));
-    // Now, try to send a new PrePrepare message immediately, without waiting to a new batch
-    onMessage(std::unique_ptr<ClientRequestMsg>(crm));
-    tryToSendPrePrepareMsg(false);
+    primaryPushNoOpIfWedgePending(lastExecutedSeqNum + 1);
```
yontyon, line 5785: shouldn't it be lastExecutedSeqNum (without the +1)?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5798,66 +5782,9 @@ void ReplicaImp::executeNextCommittedRequests(concordUtils::SpanWrapper &parent_
...
   }
+  primaryPushNoOpIfWedgePending(lastExecutedSeqNum);
```
yontyon, line 5787: isn't this a duplicate of the above (and also out of scope)?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5324,34 +5334,7 @@ void ReplicaImp::onExecutionFinish() {
...
-                                    requestSeqNum,
-                                    strMsg.size(),
-                                    strMsg.c_str(),
-                                    60000,
-                                    "wedge-noop-command-" + std::to_string(lastExecutedSeqNum + 1));
-    // Now, try to send a new PrePrepare message immediately, without waiting to a new batch
-    onMessage(std::unique_ptr<ClientRequestMsg>(crm));
-    tryToSendPrePrepareMsg(false);
-  }
+  primaryPushNoOpIfWedgePending(lastExecutedSeqNum + 1);
```
yontyon, line 5337: shouldn't it be without the +1?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -192,7 +170,7 @@ SigManager::SigManager(PrincipalId myId,
     }
   }
 
-  clientsPublicKeys_.version = 1;  // version `1` suggests RSAVerifier.
+  clientsPublicKeys_.version = 2;  // version `1` suggests RSAVerifier.
```
yontyon, line 173: specify what version 2 suggests

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -225,19 +201,23 @@ uint16_t SigManager::getSigLength(PrincipalId pid) const {
...
+      return result;
     } else {
       LOG_ERROR(GL, "Unrecognized pid " << pid);
       return 0;
     }
   }
 }
 
-bool SigManager::verifySig(
+bool SigManager::verifyNonReplicaSig(
```
yontyon, line 213: verifyNonReplicaSig -> verifyClientSig ?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -225,19 +201,23 @@ uint16_t SigManager::getSigLength(PrincipalId pid) const {
...
     PrincipalId pid, const concord::Byte* data, size_t dataLength, const concord::Byte* sig, uint16_t sigLength) const {
+  LOG_DEBUG(GL, "Validating NON-replica with: " << KVLOG(myId_, pid, sigLength));
   bool result = false;
   {
     std::shared_lock lock(mutex_);
+    ConcordAssert(!replicasInfo_.isIdOfReplica(myId_) || !replicasInfo_.isIdOfReplica(pid));
```
yontyon, line 219: I think sigManager is used by replicas only, so `replicasInfo_.isIdOfReplica(myId_)` always return true

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -271,18 +251,103 @@ bool SigManager::verifySig(
...
+bool SigManager::verifyReplicaSig(PrincipalId replicaID,
+                                  const concord::Byte* data,
+                                  size_t dataLength,
+                                  const concord::Byte* sig,
+                                  uint16_t sigLength) const {
+  LOG_DEBUG(GL, KVLOG(myId_));
+  ConcordAssert(replicasInfo_.isIdOfReplica(myId_) || replicasInfo_.isRoReplica());
+  ConcordAssert(replicasInfo_.isIdOfReplica(replicaID));
+  int i = 0;
+  for (auto [seq, multisigVerifier] : CryptoManager::instance().getLatestVerifiers()) {
```
yontyon, line 288: multisigVerifier -> verifier?
Why multisig?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -271,18 +251,103 @@ bool SigManager::verifySig(
...
+bool SigManager::verifySig(
+    PrincipalId pid, const concord::Byte* data, size_t dataLength, const concord::Byte* sig, uint16_t sigLength) const {
+  if (replicasInfo_.isIdOfReplica(pid) && ReplicaConfig::instance().singleSignatureScheme) {
+    return verifyReplicaSig(pid, data, dataLength, sig, sigLength);
+  }
+
+  return verifyNonReplicaSig(pid, data, dataLength, sig, sigLength);
+}
+
+bool SigManager::verifyOwnSignature(const concord::Byte* data,
```
yontyon, line 321: why don't we validate the signature with a validator?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -304,10 +369,71 @@ void SigManager::setClientPublicKey(const std::string& key, PrincipalId id, KeyF
...
+  }
+  return mySigner_->getPrivKey();
+}
+
+std::array<std::string, replicaIdentityHistoryCount> SigManager::getPublicKeyOfVerifier(uint32_t id) const {
+  std::array<std::string, replicaIdentityHistoryCount> publicKeys;
+
+  if (ReplicaConfig::instance().singleSignatureScheme && replicasInfo_.isIdOfReplica(id)) {
+    int i = 0;
+    for (auto [seq, multisigVerifier] : CryptoManager::instance().getLatestVerifiers()) {
```
yontyon, line 408: why multisigVerifier and not simply verifier?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -304,10 +369,71 @@ void SigManager::setClientPublicKey(const std::string& key, PrincipalId id, KeyF
...
+  }
+
+  return publicKeys;
+}
+
 const concord::crypto::IVerifier& SigManager::getVerifier(PrincipalId otherPrincipal) const {
   return *verifiers_.at(otherPrincipal);
 }
 
+void SigManager::setReplicaLastExecutedSeq(SeqNum seq) {
```
yontyon, line 428: Do you use this method in the RO replica?
is a ROR even capable of exchanging its keys?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -304,10 +369,71 @@ void SigManager::setClientPublicKey(const std::string& key, PrincipalId id, KeyF
...
+  ConcordAssert(replicasInfo_.isIdOfReplica(myId_) || replicasInfo_.isRoReplica());
```
yontyon, line 429: Do you use this method in the RO replica?
is a ROR even capable of exchanging its keys?

bftengine/src/bftengine/SigManager.cpp
```diff
@@ -304,10 +369,71 @@ void SigManager::setClientPublicKey(const std::string& key, PrincipalId id, KeyF
...
+  replicaLastExecutedSeq_ = seq;
+}
+
+SeqNum SigManager::getReplicaLastExecutedSeq() const {
+  ConcordAssert(replicasInfo_.isIdOfReplica(myId_) || replicasInfo_.isRoReplica());
```
yontyon, line 434: same

bftengine/src/bftengine/SigManager.hpp
```diff
@@ -26,71 +28,69 @@
...
   // returns 0 if pid is invalid - caller might consider throwing an exception
   uint16_t getSigLength(PrincipalId pid) const;
+
   // returns false if actual verification failed, or if pid is invalid
-  bool verifySig(PrincipalId pid,
-                 const concord::Byte* data,
-                 size_t dataLength,
-                 const concord::Byte* sig,
-                 uint16_t sigLength) const;
+  virtual bool verifySig(PrincipalId replicaID,
```
yontyon, line 64: why virtual? do we have any subclass?

bftengine/src/bftengine/ValidationOnlyIdentityManager.cpp
```diff
@@ -0,0 +1,53 @@
...
+                 {"", concord::crypto::KeyFormat::HexaDecimalStrippedFormat},
+                 publickeys,
+                 publicKeysMapping,
+                 false,
+                 {},
+                 replicasInfo) {}
+
+bool ValidationOnlyIdentityManager::verifySig(
+    PrincipalId pid, const concord::Byte* data, size_t dataLength, const concord::Byte* sig, uint16_t sigLength) const {
+  return verifyNonReplicaSig(pid, data, dataLength, sig, sigLength);
```
yontyon, line 32: If it validates only NonReplicaSigs, please specify it in the method name

bftengine/src/bftengine/ValidationOnlyIdentityManager.hpp
```diff
@@ -0,0 +1,41 @@
...
+//
+// This product may include a number of subcomponents with separate copyright
+// notices and license terms. Your use of these subcomponents is subject to the
+// terms and conditions of the subcomponent's license, as noted in the LICENSE
+// file.
+#include "SigManager.hpp"
+#pragma once
+
+namespace bftEngine::impl {
+/* This class is a hack to enable the validation of replica signatures
```
yontyon, line 17: can you point me to where it is used?

bftengine/src/bftengine/ValidationOnlyIdentityManager.hpp
```diff
@@ -0,0 +1,41 @@
...
+   using fixed keys without initializing a CryptoManager object.
+   Messages such as CheckpointMsg use a global singleton to expose their validation logic,
```
yontyon, line 19: The verify method eventually calls to `verifyNonReplicaSig`. Maybe I'm missing something, but checkpoint messages are signed by replicas, right?

libs/util/assertUtils.hpp
```diff
@@ -60,10 +60,11 @@ inline void printCallStack() {
           free(ret);
         }
       }
-      LOG_FATAL(GL, "\n" << os.str());
+      LOG_INFO(GL, "\n" << os.str());
```
yontyon, line 63: Why did you change from FATAL to INFO?

tests/apollo/test_skvbc_backup_restore.py
```diff
@@ -59,16 +59,19 @@ class SkvbcBackupRestoreTest(ApolloTest):
...
-        3) Stop all replicas in a random order (with 5s delay in between)
-        4) Start all replicas in a random order (with 5s delay in between)
+        3) Stop all replicas in a random order
+        4) Start all replicas in a random order
         5) Make sure the initial view is stable
         6) Send sufficient number of client requests to trigger another checkpoint
         7) Make sure checkpoint propagates to all the replicas
+
+        Note: UDP configuration waits for 5 seconds until it assumes network communication is established.
+        A replica can thus trigger a view change if it
```
yontyon, line 74: `A replica can thus trigger a view change if it`, if it what?

tests/apollo/test_skvbc_history_tracker.py
```diff
@@ -537,7 +538,7 @@ def test_num_blocks_to_linearize_over(self):
         req_index = 14
         last_known_block = 6
         last_consecutive_block = 3
-        missing_intermediate_blocks = 2
+        missing_intermediate_blocks = {1, 2}
```
yontyon, line 541: can you explain this change?

tests/apollo/test_skvbc_persistence.py
```diff
@@ -345,7 +345,7 @@ async def test_st_when_fetcher_and_sender_crash(self, bft_network, tracker):
 
         client, known_key, known_val = \
             await skvbc.prime_for_state_transfer(stale_nodes={stale_node},
-                                                           checkpoints_num=random.randint(10, 13))
+                                                 checkpoints_num=10)
```
yontyon, line 348: can you explain this change?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -544,13 +546,13 @@ async def test_tls_exchange_replicas_replicas_with_ror(self, bft_network):
         for i in range(301): # Produce 301 new blocks
             await skvbc.send_write_kv_set()
         # Now, lets switch keys to all replicas
-        exchanged_replicas = bft_network.all_replicas()
+        exchanged_replicas = bft_network.all_replicas(without={1,2,3,4,5,6})
```
yontyon, line 549: can you explain this change? why do we have to exchange the TLS keys only for replica 0 now?

tests/apollo/test_skvbc_reconfiguration.py
```diff
@@ -1610,6 +1615,57 @@ async def test_add_nodes(self, bft_network):
             nb_fast_path = await bft_network.get_metric(r, bft_network, "Counters", "totalFastPaths")
             self.assertGreater(nb_fast_path, 0)
 
+
+    @with_trio
+    @with_bft_network(start_replica_cmd, bft_configs=[BFTConfig(n=4, f=1, c=0)], publish_master_keys=True)
+    async def test_key_exchange_after_add_nodes_with_failure(self, bft_network):
```
yontyon, line 1621: cant we add it as part of the add_nodes_with_failure (and save the test time)?

yontyon: Also, maybe I'm missing something, but I don't see that there are failed replicas in this test

tests/apollo/test_skvbc_state_transfer.py
```diff
@@ -7,7 +7,7 @@
 #
 # This product may include a number of subcomponents with separate copyright
 # notices and license terms. Your use of these subcomponents is subject to the
-# terms and conditions of the subcomponent's license, as noted in the LICENSE
+# terms and conditions of the subcomponent's license, as noted in the LICENStE
```
yontyon, line 10: LICENStE - >LICENSE

tests/apollo/test_skvbc_state_transfer.py
```diff
@@ -293,6 +292,7 @@ async def test_state_transfer_rvt_validity_after_pruning(self, bft_network):
         # Wait for the RVT root values to be in sync before the pruning
         await bft_network.wait_for_replicas_rvt_root_values_to_be_in_sync(bft_network.all_replicas())
 
+        return
```
yontyon, line 295: ?

tests/apollo/test_skvbc_view_change.py
```diff
@@ -459,28 +461,17 @@ async def test_multiple_vc_slow_path(self, bft_network, tracker):
...
-        await bft_network.wait_for_view(
-            replica_id=current_primary,
-            err_msg="Make sure all ongoing view changes have completed."
-        )
-
+        await bft_network.wait_for_view_with_threshold(view)
         await skvbc.read_your_writes()
 
-        #check after test is fixed
-        await bft_network.assert_slow_path_prevalent(0, 0, stable_replica)
```
yontyon, line 483: Is it removed on purpose?

tests/apollo/util/bft.py
```diff
@@ -223,11 +236,14 @@ async def test_with_bft_network():
...
                             await test_with_bft_network()
+                    ApolloTest.FAILED_CASES.pop(subtest_id)
+                    test_instance.register_errors()
+
         return wrapper
     return decorator
 
 MAX_MSG_SIZE = 64*1024 # 64k
-REQ_TIMEOUT_MILLI = 5000
+REQ_TIMEOUT_MILLI = 5000 * 2
```
yontyon, line 246: is it doubled in purpose?
It means that in all the tests that are waiting for a client request to timeout, it will take ~10 seconds

tests/apollo/util/bft.py
```diff
@@ -1109,6 +1180,31 @@ async def get_metric(self, replica_id, bft_network, mtype, mname, component='rep
...
+        Waits for at least threshold replicas to reach view view_number
+        :param view_number: The view number to reach
+        :param threshold: The minimal amount of replicas which are expected to reach view view_number,
+                          Defaults to 2f + 2c + 1
+        :param timeout_seconds: Maximal amount of seconds to wait
+        :param sleep_seconds: Time to sleep between samples, in seconds
+        :param error_msg: Error to display in case of timeout
+        :return: The number of replicas who reached view view_number at the time of the last sample
+        """
+        threshold = threshold if threshold else 2 * self.config.f + 2 * self.config.c + 1
```
yontyon, line 1195: I think it is enough to have a regular bft quorum (2f + c +1) replicas in the new view.
Indeed, in order to complete the view change, the new primary has to get at least 2f + 2c +1 viewchange messages from different replicas, but it is enough to consider a view as valid view if the replicas are able to commit in it, and for that, it is enough to have only 2f + c + 1 replicas.
I suggest to also ask @guyg7

tests/apollo/util/bft.py
```diff
@@ -1130,7 +1226,9 @@ async def wait_for_view(self, replica_id, expected=None,
                 matching_view = await self._wait_for_matching_agreed_view(replica_id, expected)
                 action.log(message_type=f'Matching view #{matching_view} has been agreed among replicas.')
 
-                nb_replicas_in_matching_view = await self._wait_for_active_view(matching_view)
+                nb_replicas_in_matching_view = \
+                    await self.wait_for_view_with_threshold(view_number=matching_view,
+                                                            threshold=2 * self.config.f + 2 * self.config.c + 1)
```
yontyon, line 1231: same here

tests/apollo/util/bft.py
```diff
@@ -1314,24 +1407,30 @@ async def wait_for_state_transfer_to_stop(self, up_to_date_node, stale_node, sto
...
                                     action.log(message_type="Not complete yet",
-                                        seq_num=n, last_stored_checkpoint=last_stored_checkpoint, on_transferring_complete=on_transferring_complete)
-
-                                # Exit condition
-                                if n >= expected_seq_num:
-                                    action.add_success_fields(n=n, expected_seq_num=expected_seq_num)
+                                               seq_num=stale_node_metric, expected_seq_num=expected_seq_num,
+                                               last_stored_checkpoint=last_stored_checkpoint,
+                                               on_transferring_complete=on_transferring_complete)
+                                    log.log_message(message_type="Not complete yet",
```
yontyon, line 1426: what is the difference between `log.log_message` and `action.log`?
Do we need them both?

tests/apollo/util/pyclient/bft_client.py
```diff
@@ -47,6 +67,9 @@ def next(self):
         Calculate the next req_seq_num.
         Return the calculated value as an int sized for 64 bits
         """
+
+        globals()['__COUNT'].inc()
+        return self.val()
         milli = int(time.time() * 1000)
```
yontyon, line 73: If the old use to get a sequence number is not going to be used anymore, please either remove it or implement them both (in two different ReqSeqNum implementations) and let the user choose which one he wants with a flag

tests/apollo/util/pyclient/bft_client.py
```diff
@@ -60,6 +83,7 @@ def next(self):
         return self.val()
 
     def val(self):
+        return globals()['__COUNT'].val()
```
yontyon, line 86: same

tests/apollo/util/skvbc.py
```diff
@@ -344,17 +345,20 @@ async def assert_successful_put_get(self):
...
-            assert last_block + 1 == reply.last_block_id
+            last_block_after_write = reply.last_block_id
+            assert last_block_after_write > last_block, f'last_block_after_write: {last_block_after_write},' \
+                                                        f'last_block: {last_block}'
 
             # Retrieve the last block and ensure that it matches what's expected
             read_reply = await client.read(self.get_last_block_req())
             newest_block = self.parse_reply(read_reply)
-            assert last_block + 1 == newest_block
+            assert newest_block >= last_block_after_write, f'newest_block: {newest_block}, ' \
```
yontyon, line 355: shouldn't it be `newest_block > last_block_after_write` without the `=`?

tests/simpleKVBC/TesterReplica/internalCommandsHandler.cpp
```diff
@@ -263,6 +316,23 @@ std::optional<std::map<std::string, std::string>> InternalCommandsHandler::getBl
     }
   }
 
+  // Handle main key updates which should be ignored by the tests
```
yontyon, line 319: `  // Handle main key updates which should be ignored by the tests` ->
`   // Handle main key updates which should be ignored by the linzeriability checker of apollo tests` (or something similar).
It is ignored only by the skvbc tracker, right?

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
yontyon: Lets handle it globally in a different PR

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.hpp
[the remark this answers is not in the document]
yontyon: ok, but why is it related to transaction signatures?
At the moment, we can start state transfer without any signature validation.
To leave state transfer, we do need that at least f+1 replicas will be able to validate the CRE requests, but this is done at the reconfiguration level and not at the transaction signing level (and as we discussed offline, we maybe need to start thinking on a different solution than Async CRE).
I still don't understand how transaction signing related to all of this

bftengine/src/bftengine/RequestHandler.cpp
yontyon: it should, we can leave it in INFO

bftengine/src/bftengine/CryptoManager.cpp
yontyon: up to you :) 
I only wanted to understand

tests/apollo/util/bft.py
WildFireFlum: Yes, this was a change proposed by you because of the 5 seconds we wait for replicas with UDP connections

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: Both names are inaccurate, this is the logic which uses the keys stored in signmanager (unlike keys which are stored in cryptomanager)

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: This is not true, it is used by replicas, read only replicas, and previously by the integritychecker.
Someone might try and use it for an incompatible component in the future, this assertion will prevent it.

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: There were casts inside this function in a previous version.

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: IDK, I just extracted a logic which was used in the preprocessor.

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: There were casts inside this function in a previous version. This name reminds you that the source of this object is CryptoManager.

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -31,81 +36,72 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
+
+  // Important note:
+  // CryptoManager's cryptosystems are currently implemented using a naive eddsa multisig scheme
+  // The following methods break the abstraction of the threshsign library in order
+  // to extract ISigner and IVerifier objects.
+  // This abstraction is broken to allow using the consensus key as the replica's main key (In SigManager), thus
+  // enabling an operator to change it (key rotation).
+  // This code will need to be refactored if a different implementation is used.
+  std::shared_ptr<EdDSAMultisigSigner> getSigner(SeqNum seq) const;
+  std::shared_ptr<EdDSAMultisigVerifier> getMultisigVerifier(SeqNum seq) const;
```
toly-kournik, line 95: Please remove if it's unused

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -31,81 +36,72 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
+  void onCheckpoint(uint64_t newCheckpoint);
+
+  // Important note:
+  // CryptoManager's cryptosystems are currently implemented using a naive eddsa multisig scheme
+  // The following methods break the abstraction of the threshsign library in order
+  // to extract ISigner and IVerifier objects.
+  // This abstraction is broken to allow using the consensus key as the replica's main key (In SigManager), thus
+  // enabling an operator to change it (key rotation).
+  // This code will need to be refactored if a different implementation is used.
+  std::shared_ptr<EdDSAMultisigSigner> getSigner(SeqNum seq) const;
```
toly-kournik, line 94: I'm still convinced we need to return an `IThresholdSigner` interface

bftengine/include/bftengine/CryptoManager.hpp
```diff
@@ -31,81 +36,72 @@ class CryptoManager : public IKeyExchanger, public IMultiSigKeyGenerator {
...
+  std::shared_ptr<EdDSAMultisigVerifier> getMultisigVerifier(SeqNum seq) const;
+  std::array<std::pair<SeqNum, std::shared_ptr<EdDSAMultisigVerifier>>, 2> getLatestVerifiers() const;
```
toly-kournik, line 96: same here and below for `IThresholdVerifier`

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.hpp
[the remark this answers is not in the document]
WildFireFlum: The reconfiguration client uses the same signature setup as a regular client:
`bftClient->setTransactionSigner(transactionSigner.release());`
(in CreFactory::create)
Guy's requirement was that a replica can rely on ST to return to the working window without relying on anything but a valid communication channel.

tools/KeyfileIOUtils.cpp
WildFireFlum: yes, done

bftengine/src/bftengine/CryptoManager.cpp
WildFireFlum: true, changed the comment

bftengine/include/bftengine/KeyExchangeManager.hpp
WildFireFlum: Renaming will create inconsistencies with messages (KeyExchangeMsg)

bftengine/include/bftengine/KeyExchangeManager.hpp
WildFireFlum: done

bftengine/src/bcstatetransfer/AsyncStateTransferCRE.cpp
WildFireFlum: I changed this function, but the interfaces should also reflect the ownership model.
Since the interface changes also affect clients, I think that they are out of scope.

bftengine/src/bftengine/CollectorOfThresholdSignatures.hpp
WildFireFlum: done

bftengine/src/bftengine/CryptoManager.cpp
WildFireFlum: done

bftengine/src/bftengine/CryptoManager.cpp
WildFireFlum: This is an implementation of the IKeyExchanger interface.
Since there are a lot of renames to do, I think it should be done in a separate PR.

bftengine/src/bftengine/ReplicaImp.cpp
WildFireFlum: done

bftengine/src/bftengine/ReplicaImp.cpp
WildFireFlum: done

bftengine/src/bftengine/ReplicaImp.cpp
WildFireFlum: idk, just extracted a function for huge duplication

bftengine/src/bftengine/ReplicaImp.cpp
WildFireFlum: the original version had the +1

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: done

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: Once ROR finishes state transfer, its last executed sequence number is updated.

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: same answer

bftengine/src/bftengine/ValidationOnlyIdentityManager.cpp
WildFireFlum: Can't, need to override the parent's method.

bftengine/src/bftengine/ValidationOnlyIdentityManager.hpp
WildFireFlum: Indeed. As I explained in a previous discussion, the naming is poor, but the meaning is that Sigmanager's internal map is used instead of redirecting to CryptoManager.

libs/util/assertUtils.hpp
WildFireFlum: Because I added call stack prints while debugging and didn't want my tests to fail.

tests/apollo/test_skvbc_backup_restore.py
WildFireFlum: done

tests/apollo/test_skvbc_history_tracker.py
WildFireFlum: block numbers instead of block count

tests/apollo/test_skvbc_persistence.py
WildFireFlum: wanted to be able to replicate the test's behavior and compare logs

tests/apollo/test_skvbc_reconfiguration.py
WildFireFlum: 1. add_nodes_with_failure needs to be shortened if anything.
2. This is a simpler (and faster) scenario where the network is scaled up, two replicas are behind, and they are expected to complete ST successfully one by one and exchange their keys.

tests/apollo/test_skvbc_view_change.py
WildFireFlum: I re-added this check

tests/apollo/test_skvbc_reconfiguration.py
WildFireFlum: Reverted

bftengine/src/bftengine/messages/ViewChangeMsg.cpp
WildFireFlum: removed

bftengine/src/bftengine/RequestHandler.cpp
WildFireFlum: nevermind, already changed to debug

bftengine/src/bftengine/ReplicasInfo.cpp
WildFireFlum: Done

tests/apollo/util/bft.py
WildFireFlum: I think that this change is out of scope (affects too many cases)

tests/apollo/util/bft.py
WildFireFlum: same answer

tests/apollo/util/bft.py
WildFireFlum: We don't need both.

tests/apollo/util/bft.py
yontyon: If its only for UDP setups, maybe it is worth doubling the timeout only for those cases.

tests/apollo/util/pyclient/bft_client.py
WildFireFlum: done

tests/apollo/util/pyclient/bft_client.py
WildFireFlum: done

bftengine/src/bftengine/ReplicaForStateTransfer.cpp
WildFireFlum: Since the candidate is not persisted, I thought that the persistency should have been handled in a separate PR.
Nevertheless, I've implemented the logic you requested.

bftengine/src/bftengine/ValidationOnlyIdentityManager.hpp
WildFireFlum: renamed.

libs/util/assertUtils.hpp
WildFireFlum: reverted

tests/apollo/util/bft.py
WildFireFlum: Makes it more difficult to debug timeouts that way

kvbc/src/reconfiguration_kvbc_handler.cpp
WildFireFlum: Done

bftengine/src/bftengine/ReplicaImp.cpp
WildFireFlum: A duplicated code section was extracted to a function..
It is out of scope.

bftengine/src/bftengine/SigManager.cpp
WildFireFlum: Renamed

PR #2950, opened 2023-03-08
Preprepare msg timestamp include in signature

titusen:
Timestamp of preprepare message is now cryptographically verified.  

Tests are passing

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -738,6 +738,10 @@ PrePrepareMsgCreationResult ReplicaImp::finishAddingRequestsToPrePrepareMsg(PreP
                                                                             uint32_t maxSpaceForReqs,
                                                                             uint32_t requiredRequestsSize,
                                                                             uint32_t requiredRequestsNum) {
+  if (getReplicaConfig().timeServiceEnabled) {
+    prePrepareMsg->setTime(time_service_manager_->getClockTimePoint());
+  }
+
   if (prePrepareMsg->numberOfRequests() == 0) {
     LOG_INFO(GL, "No client requests added to the PrePrepare batch, delete the message");
     return std::make_pair(nullptr, false);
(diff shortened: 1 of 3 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -738,6 +738,10 @@ PrePrepareMsgCreationResult ReplicaImp::finishAddingRequestsToPrePrepareMsg(PreP
                                                                             uint32_t maxSpaceForReqs,
                                                                             uint32_t requiredRequestsSize,
                                                                             uint32_t requiredRequestsNum) {
+  if (getReplicaConfig().timeServiceEnabled) {
+    prePrepareMsg->setTime(time_service_manager_->getClockTimePoint());
+  }
+
```
HristoStaykov, line 744: I am not sure I understand why we add the time here and conditionally later add it ReplicaImp.cpp:879?

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -173,6 +175,7 @@ PrePrepareMsg::PrePrepareMsg(ReplicaId sender,
   b()->viewNum = v;
   b()->time = 0;
 
+  // b()->time = ready ? *((int64_t*)(spanContext.data().c_str() + 6 + 8 + 8 + 8 + 2)) : 0;
```
HristoStaykov, line 178: Is this comment intentionally added here? If this is the case can we add some explanation as well?

HristoStaykov (commented):
Good work, @titusen ! I have some questions to get a better understanding of the fix before I can approve.

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
```diff
@@ -76,6 +76,8 @@ void PrePrepareMsg::calculateDigestOfRequests(Digest& digest) const {
     for (const auto& sod : sigOrDigestOfRequest) {
       sigOrDig.append(sod.first, sod.second);
     }
+    LOG_INFO(GL, "Block timestamp: " << b()->time << "sigOrDig");
```
teoparvanov, line 79: I suggest to keep this at the DEBUG level, otherwise it may spam the logs when the system is under load.

titusen: Removed

bftengine/src/bftengine/messages/PrePrepareMsg.cpp
titusen: Removed

bftengine/src/bftengine/ReplicaImp.cpp
titusen: Cause later in finishAddingRequestsToPrePrepareMsg the diggest is computed, so the timestamp needs to be set before that.

HristoStaykov (approved):
LGTM, thanks, @titusen !

PR #2954, opened 2023-03-11
Clientservice and replica handling for primary-only logic for Eth Filter APIs

harsht1737:
* **Problem Overview**  
 This handling involves introducing PRIMARY_ONLY flag and setting it when we receive request from ethrpc and propagate this flag appropriately to Concord where it can, send replies from primary node only and do not process request on any other node. Replies are tagged with same flag so that we don't participate in reply consensus as reply is served only from primary node.

### DIFF
--- a/client/client_pool/src/concord_client_pool.cpp
+++ b/client/client_pool/src/concord_client_pool.cpp
@@ -480,6 +480,9 @@ void SingleRequestProcessingJob::execute() {
       read_config_.request.max_reply_size = max_reply_size_;
     }
     read_config_.request.reconfiguration = flags_ & RECONFIG_FLAG_REQ;
+    if ((flags_ & PRIMARY_ONLY_REQ) == PRIMARY_ONLY_REQ) {
+      read_config_.request.primary_only = true;
+    }
     res = processing_client_->SendRequest(read_config_, std::move(request_));
   } else {
     write_config_.request.timeout = timeout_ms_;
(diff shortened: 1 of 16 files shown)


yuliasherman (commented):
What is the use case for this approach? We cannot serve write requests this way.

yuliasherman:
What is the use case for this approach? We cannot serve write requests this way.

client/client_pool/src/concord_client_pool.cpp
```diff
@@ -473,6 +480,9 @@ void SingleRequestProcessingJob::execute() {
       read_config_.request.max_reply_size = max_reply_size_;
     }
     read_config_.request.reconfiguration = flags_ & RECONFIG_FLAG_REQ;
```
rachitchadha, line 482: is reconfiguration and primary_only mutually exclusive for a single read request?
If yes, then better to place them in if else block like it has been done in above code.

client/proto/request/v1/request.proto
```diff
@@ -65,6 +65,12 @@ message Request {
   // `correlation_id` to a sequence number which can be used to track the request
   // in log messages in the blockchain network.
   optional string correlation_id = 5;
+
+  // Optional flag to mark the request as primary-only.
+  // A primary-only request is a read-only request intended to be served by 
+  // primary node only and this request is ignored by other non-primary nodes.
```
rachitchadha, line 71: To avoid the confusion please use validator node instead of node.

PR #2955, opened 2023-03-12
Clientservice and replica handling for primary-only logic for Eth Filter APIs

harsht1737:
* **Problem Overview**  

This handling involves introducing PRIMARY_ONLY flag and setting it when we receive request from ethrpc and propagate this flag appropriately to Concord where it can, send replies from primary node only and do not process request on any other node. We tag the corresponding match config with same flag in reply certificates so that we don't participate in reply consensus as reply is served only from primary node.

The use case for this approach is to support ETH Filter APIs (eth_newFilter, eth_getFilterLogs, eth_uninstallFilter etc). These APIs are designed to be read requests which are intended to served by primary node only and not served by any other non-primary nodes.

### DIFF
--- a/bftengine/include/bftengine/SimpleClient.hpp
+++ b/bftengine/include/bftengine/SimpleClient.hpp
@@ -40,7 +40,7 @@ struct SimpleClientParams {
 };
 
 // Possible values for 'flags' parameter
-enum ClientMsgFlag : uint8_t {
+enum ClientMsgFlag : uint64_t {
   EMPTY_FLAGS_REQ = 0x0,
   READ_ONLY_REQ = 0x1,
   PRE_PROCESS_REQ = 0x2,
@@ -49,6 +49,7 @@ enum ClientMsgFlag : uint8_t {
   EMPTY_CLIENT_REQ = 0x10,
   RECONFIG_FLAG_REQ = 0x20,
   RECONFIG_READ_ONLY_REQ = 0x21,  // Same as READ_ONLY_REQ | RECONFIG_FLAG_REQ
+  PRIMARY_ONLY_REQ = 0x401,
 };
 
 // Call back for request - at this point we know for sure that a client is handling the request, so we can assure that
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -4687,6 +4687,14 @@ void ReplicaImp::executeReadOnlyRequest(concordUtils::SpanWrapper &parent_span, 
   ClientReplyMsg reply(currentPrimary(), request->requestSeqNum(), config_.getreplicaId());
   uint16_t clientId = request->clientProxyId();
   int status = 0;
+
+  // Send dummy reply to clients if this is not primary replica, as primary replica will only process the request
+  if (request->isPrimaryOnly() && !isCurrentPrimary()) {
+    LOG_DEBUG(GL, "PrimaryOnly & ReadOnly request received and node not current primary, sending dummy reply back.");
+    send(&reply, clientId);
+    return;
+  }
+
   bftEngine::IRequestsHandler::ExecutionRequestsQueue accumulatedRequests;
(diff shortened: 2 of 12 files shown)


KTeofanidis (approved):
LGTM!

bftengine/include/bftengine/SimpleClient.hpp
```diff
@@ -40,7 +40,7 @@ struct SimpleClientParams {
 };
 
 // Possible values for 'flags' parameter
-enum ClientMsgFlag : uint8_t {
+enum ClientMsgFlag : uint64_t {
```
teoparvanov, line 43: I think the `SimpleClient.hpp` is not used at all. If so, let's not change it at all.

teoparvanov (approved):
LGTM, after discussing in detail the reasoning behind the `ClientMsgFlag` change. Thanks @harsht1737 for taking the time to explain!

bftengine/include/bftengine/SimpleClient.hpp
harsht1737: As discussed offline, `SimpleClient.hpp` is still being used at various places in the code (for eg `external_client.hpp` and `concord_client_pool.hpp` import it) and utilise this enum. 
Ideally (as discussed), `SimpleClient.hpp` and its related code should be removed from repo (because its mostly  dead code) and such cases (like enum usage) should be handled gracefully. Will add a tech debt item for it upon discussion with Celine.

Also, reason for upgrading from 8b to 64b for this flag, is to maintain consistency with enum `MsgFlag : uint64_t` in `Replica.hpp` which already is on 64b . We need to have flag consistency across all 3 places changed.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -4687,6 +4687,13 @@ void ReplicaImp::executeReadOnlyRequest(concordUtils::SpanWrapper &parent_span,
   ClientReplyMsg reply(currentPrimary(), request->requestSeqNum(), config_.getreplicaId());
   uint16_t clientId = request->clientProxyId();
   int status = 0;
+
+  // Send dummy reply to clients if this is not primary replica, as primary replica will only process the request
+  if (request->isPrimaryOnly() && !isCurrentPrimary()) {
+    send(&reply, clientId);
```
rachitchadha, line 4693: Please add a debug log here.

client/bftclient/src/matcher.h
```diff
@@ -27,6 +27,7 @@ struct MatchConfig {
   MofN quorum;
   uint64_t sequence_number;
   bool include_primary_ = true;  // by default part of the match is the current primary
+  bool isPrimaryOnly = false;    // set only if the request is primary-only request
```
rachitchadha, line 30: can you change it to is_primary_only ?

client/client_pool/src/concord_client_pool.cpp
```diff
@@ -474,6 +481,9 @@ void SingleRequestProcessingJob::execute() {
       read_config_.request.max_reply_size = max_reply_size_;
     }
     read_config_.request.reconfiguration = flags_ & RECONFIG_FLAG_REQ;
+    if ((flags_ & PRIMARY_ONLY_REQ) == PRIMARY_ONLY_REQ) {
```
rachitchadha, line 484: read_config_.request.primary_only = (flags_ & PRIMARY_ONLY_REQ) == PRIMARY_ONLY_REQ;

rachitchadha (approved):
LGTM

bftengine/src/bftengine/ReplicaImp.cpp
harsht1737: Fixed

client/client_pool/src/concord_client_pool.cpp
harsht1737: Fixed

client/bftclient/src/matcher.h
harsht1737: Fixed

PR #2960, opened 2023-03-14
MultiSizeBufferPool: remove invalid thrown exception, avoid overflow

glevkovich:
* **Problem Overview**  
Due to the thread model of the pool, which avoid as much as possible from locking, I have to remove one of the exception thrown. The exception logic is invalid in some rare cases.
Furthermore, due to the chance for overflow, we convert some of the statistics class members into signed integers - overflow on unsigned integers is undefined according to the C++ standard.
* **Testing Done**  
Tested current Apollo tests. There is no need to add any new ones.

### DIFF
--- a/libs/util/MultiSizeBufferPool.hpp
+++ b/libs/util/MultiSizeBufferPool.hpp
@@ -253,39 +253,39 @@ class MultiSizeBufferPool {
       // Subpool level:
       // numAllocatedBytes_ + numNonAllocatedBytes_ = SubpoolConfig::numMaxBuffers * (SubpoolConfig::bufferSize +
       // chunkMetadataSize())
-      std::atomic_uint64_t numAllocatedBytes_{};
-      std::atomic_uint64_t numNonAllocatedBytes_{};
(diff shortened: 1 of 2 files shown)


eyalrund:
what are the rare cases?

eyalrund (approved):

glevkovich:
> what are the rare cases?

I didn't think that explaining the details of the rare cases will help.
We have a case here of the main subpool (64K) need to allocate all all 1400 chunks. Why did it happen? I have no explenation, since there are no logs. But it can happen - iys a valid case. I assume there was some issue in the execution engine or in consenssus which caused that. In that case, we reached the max quota but due to the non-locking pool all statistics class members are updated one by one. Multiple threads enter the pop/push logic and the exception  if statment is invalid, due to the speific timing in which the statistics members are updated (we have some temprary integer overflow).
I had to remove that sanity check,  since I do not want to lock the queue. No other choice. Of course, we can lock and keep the sanity check ,but this will slow down the operation..

yuliasherman (approved):
LGTM

PR #2962, opened 2023-03-16
Move reconfiguration lib to libs

yontyon:
* **Problem Overview**  
  As part of the open source strategy, this PR moves the reconfiguration library to libs
* **Testing Done**  
  CI


eyalrund (approved):

libs/reconfiguration/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+project (libreconfiguration LANGUAGES CXX)
```
toly-kournik, line 1: can you `git mv` this file for history tracking

teoparvanov:
@yontyon, IMHO this change looks risky at this late stage in the release (and not functionally needed for 1.9 either). Also, wouldn't there be any related changes in GitLab (f.e. due to the moving of header files to different directories)?

teoparvanov (changes requested):
In addition, I'd suggest to talk to @harsht1737 to make sure all the reconfiguration tests have been re-enabled in post-master before proceeding with this change.

toly-kournik:
> @yontyon, IMHO this PR looks risky at such a late stage in the release (and not functionally needed for 1.9 either). Also, wouldn't there be any related changes in GitLab (f.e. because header files are moving to different directories)?

@teoparvanov , there's no risk as no new functionality is added, we are just arranging the code in a new way.

toly-kournik:
> @yontyon, IMHO this PR looks risky at such a late stage in the release (and not functionally needed for 1.9 either). Also, wouldn't there be any related changes in GitLab (f.e. because header files are moving to different directories)?

The includes were already prefixed by the containing folder name, i.e `#include "reconfiguration/*`

toly-kournik:
> Thanks for clarifying, Toly! Ideally though we would wait for the 1.9 branch to be pulled (mid next week I expect). I'd also suggest to manually run all Hermes reconfiguration tests on a private build - even the ones that are currently disabled temporarily.

Basically it can wait for the next week

teoparvanov (approved):
Approving this PR, as the 1.9/Beta2 release has been postponed. Thanks for the patience @yontyon and @toly-kournik !

PR #2963, opened 2023-03-17
Add signature field to response proto object

nkgeorgiev:
* **Problem Overview**  
  We need a new signatures field in the response so that any user can verify that the response is valid.
  Currently this is done only for ethereum requests.
  This change collects the signatures from the RSI field and stores them in the new signatures field.
  Whoever made the request will handle the signatures as needed.
* **Testing Done**  
  Tested with signing ethereum requests and then verifying them on the user side


HristoStaykov (approved):
LGTM, thanks, @nkgeorgiev !

teoparvanov (approved):
LGTM!

PR #2965, opened 2023-03-21
Move storage to libs

yontyon:
* **Problem Overview**  
  Move storage library to libs
* **Testing Done**  
  CI


libs/storage/CMakeLists.txt
toly-kournik, line 1: Please `git mv` this file

PR #2976, opened 2023-03-30
[Readme] Add a link to the SBFT formal model

teoparvanov:
* **Problem Overview**  
  The Readme does not advertise the detailed and high value formal model that we have developed for the SBFT protocol.
* **Testing Done**  
  Checked the Markdown preview, to make sure the new entry is rendered correctly, and that the link works.


HristoStaykov (approved):
Great to have this hint in the README. Thanks, @teoparvanov !

PR #2977, opened 2023-03-30
Application specific data storage in privacy wallet service

mgrychow:
* **Problem Overview**  
  For privacy user web app syncing purpose a storage for last known transaction number is needed in privacy wallet service
* **Testing Done**  
  UT in test_privacy_wallet_grpc_service suite + manual E2E testing

### DIFF
--- a/utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
+++ b/utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
@@ -101,6 +101,23 @@ message GetStateResponse {
     map<string, uint64> coins = 4;
 }
 
+message SetAppDataRequest {
+    string key = 1;
+    string value = 2;
+}
+
+message SetAppDataResponse {
+    bool succ = 1;
+}
+
+message GetAppDataRequest {
+    string key = 1;
+}
+
+message GetAppDataResponse {
+    string value= 1;
+}
+
 message PrivacyWalletRequest {
     oneof req {
         PrivacyAppConfigRequest privacy_app_config = 1;
--- a/utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
+++ b/utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
@@ -129,6 +129,8 @@ message PrivacyWalletRequest {
         GenerateBurnTx generate_burn_tx_request = 7;
         GenerateTransferTx generate_transfer_tx_request = 8;
         GetStateRequest get_state_request = 9;
+        SetAppDataRequest set_app_data_request = 11;
+        GetAppDataRequest get_app_data_request = 12;
 
     }
     // Required timeout which defines the maximum amount of time the caller is
@@ -148,5 +150,7 @@ message PrivacyWalletResponse {
         ClaimCoinsReponse claim_coins_response = 6;
         GenerateTxResponse generate_tx_response = 7;
         GetStateResponse get_state_response = 8;
+        SetAppDataResponse set_app_data_response = 9;
+        GetAppDataResponse get_app_data_response = 10;
     }
 }
--- a/utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
+++ b/utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
@@ -102,8 +102,8 @@ message GetStateResponse {
 }
 
 message SetAppDataRequest {
-    string key = 1;
-    string value = 2;
+    repeated string keys = 1;
+    repeated string values = 2;
 }
 
 message SetAppDataResponse {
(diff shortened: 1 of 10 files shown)


teoparvanov (approved):
LGTM, thanks @mgrychow !

utt/privacy-wallet-service/src/PrivacyService.cpp
```diff
@@ -389,4 +395,30 @@ ::grpc::Status PrivacyWalletServiceImpl::handleGetStateRequest(
...
+  }
+  return grpc::Status::OK;
+}
+
+::grpc::Status PrivacyWalletServiceImpl::handleGetAppDataRequest(
+    ::grpc::ServerContext*,
+    const ::vmware::concord::privacy::wallet::api::v1::PrivacyWalletRequest* request,
+    ::vmware::concord::privacy::wallet::api::v1::PrivacyWalletResponse* response) {
+  auto& get_app_data_req = request->get_app_data_request();
+  auto get_app_data_resp = response->mutable_get_app_data_response();
```
StefanPalashev, line 417: Nit: move those two lines under the check for `response` as they are only used in that scope.

utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
```diff
@@ -101,6 +101,23 @@ message GetStateResponse {
     map<string, uint64> coins = 4;
 }
 
+message SetAppDataRequest {
+    string key = 1;
```
bsenthilr, line 105: Idea of having flexible or transparent KV is good. Few comments:
 - It is better to make this repeated KV pair (or) map. Since single key update would not be flexible.
 - value could be blob but it really depends on what we want to store so not keen on this at this point of time.

utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
```diff
@@ -112,6 +129,8 @@ message PrivacyWalletRequest {
         GenerateBurnTx generate_burn_tx_request = 7;
         GenerateTransferTx generate_transfer_tx_request = 8;
         GetStateRequest get_state_request = 9;
+        SetAppDataRequest set_app_data_request = 11;
```
bsenthilr, line 132: set tag to 10

utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
bsenthilr: we could also just force this to be single key/value pair where value is serialized app data (blob) which is specific to application.

utt/privacy-wallet-service/src/Wallet.cpp
```diff
@@ -22,11 +22,10 @@ using namespace utt::client::utils::crypto;
 Wallet::Wallet(std::string userId,
                const std::string& private_key,
                const std::string& public_key,
-               const std::string& storage_path,
+               const std::shared_ptr<utt::client::IStorage> storage,
```
bsenthilr, line 25: why did u need this change?

mgrychow: IStorage object (FileBasedUserStorage) is constructed in PrivacyService and injected to Wallet to enable calling storage functions from PrivacyService

utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
mgrychow: 10 is already used by `google.protobuf.Duration timeout = 10;`

utt/privacy-wallet-service/tests/grpc-service/grpc-server-test.cpp
```diff
@@ -481,6 +481,32 @@ TEST_F(test_privacy_wallet_grpc_service, test_merge_and_transfer_cycles) {
   ASSERT_EQ(cycles, 3);
 }
 
+TEST_F(test_privacy_wallet_grpc_service, test_set_application_data) {
```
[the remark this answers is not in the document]
mgrychow: changed to `test_set_and_get_application_data` and added test for getting not set data

utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
mgrychow: extedned it to support multiple KV in single message

utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
bsenthilr: ah didn't notice it. We could move the timeout to tag 1 and keep others consecutive to be more clear.

utt/privacy-wallet-service/proto/api/v1/wallet-api.proto
```diff
@@ -101,6 +101,23 @@ message GetStateResponse {
     map<string, uint64> coins = 4;
 }
 
+message SetAppDataRequest {
+    repeated string keys = 1;
```
bsenthilr, line 105: minor comment: keep kv as single message proto and make that repeated to avoid misuse. You could also consider map<string, string>

PR #2983, opened 2023-04-03
implicit copy constructor call removal

titusen:
Memory allocated during buildPreparedMsg was leaking due to an implicit call of the copy constructor in the msg flow instead of smart pointer constructor.

Testing to be done in the maestro environment.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1520,8 +1520,8 @@ void ReplicaImp::onInternalMsg(InternalMessage &&msg) {
     return onCarrierMessage(*vldMsg);
   }
 
-  // Handle a PrePrepare sent by self
   if (auto *ppm = std::get_if<PrePrepareMsg *>(&msg)) {
+    std::shared_ptr<PrePrepareMsg> p(*ppm);
     // This condition will always be true but keeping this to ensure that we never decrement this variable when it is
     // at its lower bound.
     // Why we do not use ConcordAssert here?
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1536,7 +1536,7 @@ void ReplicaImp::onInternalMsg(InternalMessage &&msg) {
       ppm = nullptr;
       return;
     } else {
-      auto p = std::make_shared<PrePrepareMsg>(*ppm);
+      std::shared_ptr<PrePrepareMsg> p(*ppm);
```
glevkovich, line 1539: What's the difference?
Both implicitly trigger copy constructor (although move should have been the right choice in that case).
Why do we delete ppm only in 1st condition and not the second?
I think the right solution should be to delete ppm always. Also, ive refactored the code a bit. It is written in a strange way, unclear and misleading.
I was trying to catch you for a chat, couldn't.
Let's discuss it 1:1 when you have time. 
Here is the right fix, in my opinion:

```
// Handle a PrePrepare sent by self
  if (PrePrepareMsg **ppm = std::get_if<PrePrepareMsg *>(&msg)) {
    // This condition will always be true but keeping this to ensure that we never decrement this variable when it is
    // at its lower bound.
    // Why we do not use ConcordAssert here?
    // Ans: 0 is a correct value for this variable, as this is unsigned, it will never become negative. So check for
    // lower bound and then decrement is the correct idiom for this case.
    if (numOfTransientPreprepareMsgs_ > 0) {
      numOfTransientPreprepareMsgs_--;
    }
    if (isCollectingState() || bftEngine::ControlStateManager::instance().getPruningProcessStatus()) {
      LOG_INFO(GL, "Received PrePrepareMsg while pruning or state transfer, so ignoring the message");
    } else {
      auto p = std::make_shared<PrePrepareMsg>(*ppm);
      startConsensusProcess(p, true);
    }
    delete *ppm;
    return;
  }
```

titusen: Done

PR #2988, opened 2023-04-07
Fix recovery issue due to budget coins

bsenthilr:
* **Problem Overview**  
During restart and recovery post transfers a change budget coin gets created. The recovery does not account this and fails during recovery.
This considers multiple budget coins and uses the one with lowest budget value for now. This is because budget coin pool with expiry date scope is not yet supported. For current release we only consider single outstanding budget coin.
* **Testing Done**  
- The google test was extended to validate restore after restart
- VMBC kubernetes setups was validated for restores


teoparvanov (approved):

utt/privacy-wallet-service/src/PrivacyService.cpp
```diff
@@ -54,6 +54,9 @@ void PrivacyWalletService::Shutdown() {
 PrivacyWalletServiceImpl::PrivacyWalletServiceImpl() {
   try {
     wallet_ = Wallet::recoverFromStorage(PrivacyWalletServiceImpl::wallet_db_path);
+  } catch (std::exception& e) {
```
yontyon, line 57: if you catch here, you can remove the catch(...) from below

utt/utt-client-api/src/User.cpp
```diff
@@ -316,11 +316,25 @@ std::unique_ptr<User> User::recoverFromStorage(std::shared_ptr<IStorage> storage
     if (c.getType() == libutt::api::Coin::Normal) {
       user->pImpl_->coins_.emplace(c.getNullifier(), c);
     } else if (c.getType() == libutt::api::Coin::Budget) {
+      // Currently the original budget coin and change budget coins would
+      // be stored in assets.
+      // Until budget coin pool is supported, use the budget with lowest value
+      // to stick to current single outstanding budget workflow model
       if (user->pImpl_->budgetNullifiers_.size() >= 1) {
-        throw std::runtime_error("Currently multiple budget coins are not supported");
+        if (c.getVal() < user->pImpl_->budgetCoin_.value().getVal()) {
```
yontyon, line 324: what if the former budget coin value is 0 (which means that we have to mint a new budget coin), will it not be supported at the moment?

yontyon (approved):

utt/utt-client-api/src/User.cpp
bsenthilr: assumed the size would be zero if there is no budget coin associated is that not the case?

bsenthilr: okay the scenario is if user spends all his budget then the value is 0

PR #2989, opened 2023-04-07
Added logic to extract/validate MPT proof paths.

HristoStaykov:
We have added logic to extract proof paths from our MPT implementation and to validate that a previously extracted proof path reaches a giver root hash. A unit test is added that fills an MPT and then validates the paths extracted for every key/value inserted.

* **Problem Overview**  
  Completed implementation of extraction/verification of proof paths from our MPT implementation.
* **Testing Done**  
 A new unit test was added that fills a MPT and then every inserted key/value is checked.


kvbc/include/sparse_merkle/internal_node.h
```diff
@@ -166,6 +166,7 @@ std::ostream& operator<<(std::ostream& s, std::optional<T> const& v) {
 class BatchedInternalNode {
  public:
   static constexpr size_t MAX_CHILDREN = 31;
+  static constexpr size_t MAX_HEIGHT = 5;
```
vrancurel, line 169: why so ?

kvbc/include/sparse_merkle/base_types.h
```diff
@@ -66,7 +66,7 @@ class CustomPrefix {
   CustomPrefix() = default;
   CustomPrefix(size_t size) {
     ConcordAssert(size <= SIZE_IN_BYTES);
-    prefix_.assign('\0', size);
+    prefix_.assign(size, '\0');
```
vrancurel, line 69: how could it pass before ?

kvbc/src/sparse_merkle/internal_node.cpp
```diff
@@ -386,12 +386,12 @@ size_t BatchedInternalNode::numLevel0Children() const {
...
 }
 
-size_t BatchedInternalNode::nibbleToIndex(Nibble nibble) const {
+size_t BatchedInternalNode::nibbleToIndex(Nibble nibble) {
   constexpr size_t level_0_start = MAX_CHILDREN / 2;
   return level_0_start + nibble.data();
 }
 
-size_t BatchedInternalNode::height(size_t index) const {
+size_t BatchedInternalNode::height(size_t index) {
```
vrancurel, line 394: Is it possible to document this strange function ?

kvbc/src/sparse_merkle/proof_path_processor.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  return next == rootHash;
+}
+
+std::vector<Hash> getProofPath(Sliver key, std::shared_ptr<IDBReader> db, const std::string& custom_prefix) {
+  std::vector<Hash> retVal;
+  Hasher hasher;
+  Hash valueHash{};
+  Hash globalLeaveHash{};
+  auto key_hash = hasher.hash(key.data(), key.length());
```
vrancurel, line 73: I assume we continue hashing the keys for now, but we will change this later right ?

kvbc/src/sparse_merkle/proof_path_processor.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "sparse_merkle/keys.h"
+
+namespace concord {
+namespace kvbc {
+namespace sparse_merkle {
+namespace proof_path_processor {
+
+using Sliver = concordUtils::Sliver;
+
+bool verifyProofPath(Sliver key, Sliver value, const std::vector<Hash>& proofPath, const Hash& rootHash) {
```
vrancurel, line 24: can you document the logic of the function ?

vrancurel (changes requested):

kvbc/include/sparse_merkle/internal_node.h
HristoStaykov: The height in a single record (sub-tree) is 5, because we have a binary tree and 16 elements in the the bottom of the record. It is shown in the following documentation: [internal_node.h:114](https://github.com/vmware/concord-bft/blob/66b249a1e709ed4d1baf5f4e5e73bae19bbadd58/kvbc/include/sparse_merkle/internal_node.h#L114)

kvbc/src/sparse_merkle/internal_node.cpp
HristoStaykov: It is documented in the .h file, basically we pass the index of a child from the array **children_**, containing the 31 elements, and the return value is the height in the sub-tree. [internal_node.h:328](https://github.com/vmware/concord-bft/blob/66b249a1e709ed4d1baf5f4e5e73bae19bbadd58/kvbc/include/sparse_merkle/internal_node.h#L328)

kvbc/src/sparse_merkle/proof_path_processor.cpp
HristoStaykov: Yes, we have not yet changed the internals of our MPT. It still hashes the keys. This is a task scheduled for the near future.

kvbc/include/sparse_merkle/base_types.h
HristoStaykov: The unit test `tree_test.cpp` stores its database in RAM and does not serialize the elements. The elements of the tree are:
`  map<std::string, Version> latest_version_;
  map<LeafKey, LeafNode> leaf_nodes_;
  map<InternalNodeKey, BatchedInternalNode> internal_nodes_;`
[test_db.h:43](https://github.com/vmware/concord-bft/blob/66b249a1e709ed4d1baf5f4e5e73bae19bbadd58/kvbc/test/sparse_merkle/test_db.h#L43)
In the more advanced test we are planning we will store the test DB's elements serialized in a way similar to:
`map<std::string, Version> latest_version_;
map<std::string, std::string> nodes_;`

kvbc/src/sparse_merkle/internal_node.cpp
HristoStaykov: It is one of the functions that I changed to static, because they define the structure of the batched internal node and don't need to access the state of a particular instance.

kvbc/src/sparse_merkle/proof_path_processor.cpp
HristoStaykov: Added comment with description.

kvbc/test/sparse_merkle/tree_test.cpp
```diff
@@ -1156,6 +1159,46 @@ TEST(tree_tests, add_and_remove_of_same_existing_key_in_single_update_add_wins)
...
+}
+
+TEST(tree_tests, fill_tree_then_extract_proof_paths_and_verify_them) {
+  srand(1);
+  std::shared_ptr<TestDB> db(new TestDB);
+  Tree tree(db, "MY_ETH_ADDR");
+
+  std::map<std::string, std::string> all;
+
+  for (int i = 1; i <= 1500; i++) {
```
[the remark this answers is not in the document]
vrancurel: ok

kvbc/include/sparse_merkle/internal_node.h
vrancurel: It would be good to put a comment to explain this.

vrancurel (approved):

kvbc/include/sparse_merkle/internal_node.h
WildFireFlum: Even better would be to add a constexpr function which deduces this number from the parameters (max key length, etc...)

HristoStaykov: Yes, will add the constexpr function, I think it will be better to calculate the `MAX_HEIGHT` based on the `MAX_CHILDREN` considering that out MPT is binary.

teoparvanov (approved):

kvbc/include/sparse_merkle/internal_node.h
```diff
@@ -166,6 +166,11 @@ std::ostream& operator<<(std::ostream& s, std::optional<T> const& v) {
 class BatchedInternalNode {
  public:
   static constexpr size_t MAX_CHILDREN = 31;
```
WildFireFlum, line 168: Isn't the radix 16?

kvbc/src/sparse_merkle/proof_path_processor.cpp
```diff
@@ -0,0 +1,146 @@
...
+  Hasher hasher{};
+  auto keyHash = hasher.hash(key.data(), key.length());
+  auto valueHash = hasher.hash(value.data(), value.length());
+
+  enum class Direction : std::uint8_t { Left, Right };
+  const auto BatchedInternalNodeLevelsIterated = BatchedInternalNode::MAX_HEIGHT - 1;
```
WildFireFlum, line 30: Can be made constexpr

kvbc/src/sparse_merkle/proof_path_processor.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  std::vector<Direction> pathOrdering;
+  for (int nibble_index = 0; nibble_index < Hash::MAX_NIBBLES; nibble_index++) {
+    auto current_nibble = keyHash.getNibble(nibble_index).data();
+    auto node_index = BatchedInternalNode::nibbleToIndex(current_nibble);
+    std::vector<Direction> subPath;
```
WildFireFlum, line 36: Can be made an array and save up the allocation

kvbc/src/sparse_merkle/proof_path_processor.cpp
```diff
@@ -0,0 +1,146 @@
...
+      } else {
+        break;
+      }
+    }
+
+    auto current_index = node.nibbleToIndex(key_hash.getNibble(nibble_index));
+    auto child = node.children()[current_index];
+    nextNodetype = Nodetype::None;
+    bool firstInternalNodeFount = false;
+    std::vector<Hash> hashesCollectedFromInternalNode;
```
WildFireFlum, line 109: Change to std::array?

kvbc/include/sparse_merkle/internal_node.h
HristoStaykov: We store 16 children at the lowest level of the record then 8, 4, 2 and 1 in the layers above - in total 31 children.

kvbc/src/sparse_merkle/proof_path_processor.cpp
HristoStaykov: Agree, will change to constexpr.

kvbc/src/sparse_merkle/proof_path_processor.cpp
HristoStaykov: Yes, here we will benefit from using an array and the additional complexity will remain hidden. Will make the change.

kvbc/src/sparse_merkle/proof_path_processor.cpp
HristoStaykov: Agree, will change accordingly.

kvbc/include/sparse_merkle/proof_path_processor.h
```diff
@@ -0,0 +1,57 @@
...
+namespace kvbc {
+namespace sparse_merkle {
+namespace proof_path_processor {
+
+// Verify that a given list of hashes (proofPath) will
+// reach the given Merkle root (rootHash) for the
+// key/value that we need to check.
+bool verifyProofPath(concordUtils::Sliver key,
+                     concordUtils::Sliver value,
+                     const std::vector<Hash>& proofPath,
```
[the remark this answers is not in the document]
WildFireFlum: You can use boost::static_vector with the capacity limitation.

kvbc/src/sparse_merkle/proof_path_processor.cpp
```diff
@@ -20,37 +20,39 @@ namespace sparse_merkle {
...
 
-  std::vector<Direction> pathOrdering;
-  std::array<Direction, BatchedInternalNodeLevelsIterated> subPath;
+  static_vector<Direction, Hash::SIZE_IN_BITS> pathOrdering;
+  static_vector<Direction, BatchedInternalNodeLevelsIterated> subPath;
   for (size_t nibble_index = 0; nibble_index < Hash::MAX_NIBBLES; nibble_index++) {
     auto current_nibble = keyHash.getNibble(nibble_index).data();
     auto node_index = BatchedInternalNode::nibbleToIndex(current_nibble);
-    size_t foundElemens = 0;
+    subPath.clear();
```
WildFireFlum, line 41: Why is this needed?

WildFireFlum (approved):

kvbc/src/sparse_merkle/proof_path_processor.cpp
HristoStaykov: This clearing is needed, because we reuse the same container when we iterate to extract the directions based on the nibbles of the key for each sub-record of the tree.

kvbc/src/sparse_merkle/proof_path_processor.cpp
```diff
@@ -109,7 +113,7 @@ std::vector<Hash> getProofPath(Sliver key, std::shared_ptr<IDBReader> db, const
     auto child = node.children()[current_index];
     nextNodetype = Nodetype::None;
     bool firstInternalNodeFount = false;
-    size_t foundElemens = 0;
+    hashesCollectedFromInternalNode.clear();
```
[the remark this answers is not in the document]
HristoStaykov: Same as above, but here we actually go through multiple `BatchedInternalNodes` to collect the necessary hashes from each one and for all iterations we reuse `hashesCollectedFromInternalNode`.

kvbc/include/sparse_merkle/internal_node.h
```diff
@@ -97,6 +97,11 @@ std::ostream& operator<<(std::ostream& s, std::optional<T> const& v) {
   return s;
 }
 
+static constexpr size_t BatchedInternalNodeGetMaxHeight(const size_t children) {
+  if (children == 0) return 0;
+  return 1 + BatchedInternalNodeGetMaxHeight(children / 2);
```
nkgeorgiev, line 102: You do know that this is just a log2, right?

HristoStaykov: Yes, this is only called during compile time and the max number of children in 1 record is currently 31. We don't expect it to go to values that can actually cause slow down in compilation.

PR #2990, opened 2023-04-09
Apollo stability improvements

WildFireFlum:
* **Problem Overview**  
Many apollo tests are flaky due to invalid assumptions such as a single random replica representing the whole network's state.
Moreover, many tests assume that blocks can only be created by client requests, an assumption which is not valid when replicas publish their main key on startup.
This branch addresses these issues, and was written as part of the single signature scheme feature (https://github.com/vmware/concord-bft/pull/2948).
Big thanks to @yontyon with his help making these changes.
* **Testing Done**  
Passed internal CI (Prior to Ubuntu 22 merge).


teoparvanov (approved):
Looks great, thanks @WildFireFlum for all the test fixes, and also the debuggability improvements!

WildFireFlum:
@HristoStaykov Most of the changes were made along the way when I encountered failures in my single signature scheme branch, and at the time I didn't expect them to accumulate to this amount. Lesson learned.

PR #2992, opened 2023-04-11
Fix test command generation.

HristoStaykov:
* **Problem Overview**  
The omitted specifier SUITE was causing the restart recovery suite to run all Apollo tests.
* **Testing Done**  
Tested locally that the correct test command is generated.


teoparvanov (approved):
Great catch, thanks @HristoStaykov !

PR #2993, opened 2023-04-12
Let nightly runs determine the needed timeout.

HristoStaykov:
* **Problem Overview**  
This is a fix for the restart recovery nightly run. It takes ~6200s for it to pass. Setting the timeout in the CMakelist as a property of the test disables the timeout specified when running the test.
* **Testing Done**  
Tested locally that the correct timeout is set for the restart recovery suite and the correct timeouts are set for the CI Apollo test suites.


teoparvanov (approved):
Awesome, thanks @HristoStaykov !

PR #2998, opened 2023-04-17
[WIP] fix db checkpoint async bug

yontyon:
* **Problem Overview**  
  The current implementation of the db checkpoint feature has a synchronization bug:
  While we take the db checkpoint in the background, we don't align anything with the checkpoint sequence number, i.e. the block number, bft metadata, pending reserved pages, and more. 
  Once we put more client requests in a different resolution than 150 we start to see a wide set of issues:
  For example:
  1. The recovered replica won't start from a stable checkpoint, instead, it starts from the point where the db checkpoint was taken.
  2. In a case where the checkpoint was taken in the middle of another execution phase, we won't have the pending reserved pages to recover correctly.
  3. We trim the block at the point where the db checkpoint was taken, but we don't update the bft metadata accordingly. 
Below is an example of part of these issues:
On replica 0, block 302 was created on sequence number 304
```
0|03-04-2023 10:48:24.125|INFO |skvbctest.replica|post-execution-thread|||304||executeWriteCommand|L:482|ConditionalWrite message handled; writesCounter=302 currBlock=302 | [SQ:1215]
```
However, on recovery, the recovered replica has the same block that was created on sequence number 305:
```
5|03-04-2023 10:48:34.293|INFO |skvbctest.replica|post-execution-thread|||305||executeWriteCommand|L:482|ConditionalWrite message handled; writesCounter=1 currBlock=302 | [SQ:3]
```

In this PR we propose a wide change that fixes the problem.
Till now, the decision of whether to create a db checkpoint was the primary only. If the primary decided that it's time to create the db checkpoint, it sends a bft command whose execution is, creating a DB checkpoint. 
(Note that this approach, regardless of the above bugs, is not safe in terms of  DOS attacks, a malicious primary can order the replicas to continuously create db checkpoints).
Here we propose a different solution: the decision to create a db checkpoint is based on a deterministic local event (such as how much time has passed since the last created db checkpoint). 
This way, once decided a db checkpoint creation callback is registered to the stable sequence number event. Once the replica reaches this stable sequence number, it starts to create the db checkpoint asynchronously, but now the block number is aligned with the sequence number because it was taken right after the sequence number execution.
To make the above feasible, (1) we cannot rely on local timeouts (instead, we consider only the time being received by consensus), (2) the db checkpoint metadata (such as sequence number and timestamp) has to be shared between all replicas (via reserved pages).
* **Testing Done**  
  CI +  Changing an existing test to verify the changes


kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -966,13 +983,13 @@ struct ResetMetadata {
...
-    concord::serialize::Serializable::serialize(outStream, dbchkpt_mdt);
-    auto data = outStream.str();
-    std::vector<uint8_t> v(data.begin(), data.end());
-    p->setDbCheckpointMetadata(v);
+    // // remove db checkpoint metadata after updating it with empty dbcheckpoint metadata
+    // DbCheckpointMetadata dbchkpt_mdt;
+    // std::ostringstream outStream;
+    // concord::serialize::Serializable::serialize(outStream, dbchkpt_mdt);
+    // auto data = outStream.str();
+    // std::vector<uint8_t> v(data.begin(), data.end());
```
eyalrund, line 991: ?

tests/apollo/test_skvbc_dbsnapshot.py
```diff
@@ -931,7 +931,7 @@ async def test_restore_from_snapshot_of_other(self, bft_network, tracker):
 
         crashed_replica = list(bft_network.random_set_of_replicas(1, {initial_prim}))[0]
         bft_network.stop_replica(crashed_replica)
-        await skvbc.send_n_kvs_sequentially(DB_CHECKPOINT_WIN_SIZE)  # run till the next checkpoint
+        await skvbc.send_n_kvs_sequentially(DB_CHECKPOINT_WIN_SIZE + 100)  # run till the next checkpoint
```
eyalrund, line 934: Do you want to pass the next checkpoint?

yontyon: yes, in order to test that we don't have state deviation (as we had before this fix)

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
yontyon: need to be removed, we do want to save the recovered knowledge about the old db checkpoints

tests/apollo/test_skvbc_dbsnapshot.py
WildFireFlum: Maybe multiply instead of adding a constant for the day when we make DB_CHECKPOINT_WIN_SIZE configurable.

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5375,28 +5382,22 @@ void ReplicaImp::onExecutionFinish() {
...
-               "sendInternalCreateDbCheckpointMsg: " << KVLOG(
-                   lastExecutedSeqNum, activeExecutions_, seq_num_to_create_dbcheckpoint.value()));
     }
   }
 
   if (isCurrentPrimary() && requestsQueueOfPrimary.size() > 0) tryToSendPrePrepareMsg(true);
 }
 
+void ReplicaImp::sendInternalNoopCommand(const std::string &cid) {
+  std::string strMsg;
```
WildFireFlum, line 5396: Wasn't this code already implemented for wedges?
If not, why do these noops include a new flag?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5201,7 +5184,31 @@ void ReplicaImp::finishExecutePrePrepareMsg(PrePrepareMsg *ppMsg,
...
+  const auto &seq_num_to_create_dbcheckpoint = DbCheckpointManager::instance().getNextStableSeqNumToCreateSnapshot();
+  if ((timeSinceLastSnapshot >= getReplicaConfig().dbSnapshotIntervalSeconds.count() &&
+       sn - DbCheckpointManager::instance().getLastDbCheckpointSeqNum() >= getReplicaConfig().dbCheckPointWindowSize) ||
+      (seq_num_to_create_dbcheckpoint.has_value() && *seq_num_to_create_dbcheckpoint == sn)) {
+    std::optional blockId(DbCheckpointManager::instance().getLastReachableBlock());
+    LOG_INFO(GL,
+             "creating db checkpoint in the next stable sequence number" << KVLOG(timeSinceLastSnapshot, sn, *blockId));
+    DbCheckpointManager::instance().saveDbMetadataToReservedPages(
+        sn, std::chrono::duration_cast<std::chrono::seconds>(consensus_time_ms));
+    // The above should be happen determenistically among all replicas, regardless if they actually create the
```
WildFireFlum, line 5201: Since readers usually read from top to bottom, can you move the comment so that it precedes the line it comments?

bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -214,6 +201,8 @@ class DbCheckpointManager {
   concordMetrics::GaugeHandle lastDbCheckpointSizeInMb_;
   concordMetrics::GaugeHandle lastDbCheckpointBlockId_;
   concordMetrics::CounterHandle numOfDbCheckpointsCreated_;
+  std::string page;
```
WildFireFlum, line 204: 1. Why do you need this member? (please document)
2. Please conform to the naming rule (_ suffix)

bftengine/include/bftengine/DbCheckpointManager.hpp
```diff
@@ -214,6 +201,8 @@ class DbCheckpointManager {
...
+  DbLastCheckpointMetadata shared_metadata_;
```
WildFireFlum, line 205: Please document why this is needed

kvbc/src/reconfiguration_kvbc_handler.cpp
```diff
@@ -234,8 +234,9 @@ bool StateSnapshotReconfigurationHandler::handle(const concord::messages::StateS
              "StateSnapshotRequest(participant ID = " << cmd.participant_id << "): using existing last checkpoint ID: "
                                                       << last_checkpoint_desc->checkPointId_);
   } else {
-    const auto checkpoint_id =
-        DbCheckpointManager::instance().createDbCheckpointAsync(sequence_number, timestamp, std::nullopt);
+    // Here we don't save the shared data about the db checkpoint creation. This is because this snapshot is not meant
```
WildFireFlum, line 237: So what should it be used for?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -3524,7 +3525,8 @@ void ReplicaImp::onSeqNumIsStable(SeqNum newStableSeqNum, bool hasStateInformati
     IControlHandler::instance()->onStableCheckpoint();
     ControlStateManager::instance().wedge();
   }
-  onSeqNumIsStableCallbacks_.invokeAll(newStableSeqNum);
+  onSeqNumIsStableOneTimeCallbacks_.invokeAll(newStableSeqNum);
+  onSeqNumIsStableOneTimeCallbacks_.clear();
```
cloudnoize, line 3529: adding clear changes the behavior of the callback mechanism ,but I guess that the name OneTime reflects that the change is desired

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5201,7 +5184,31 @@ void ReplicaImp::finishExecutePrePrepareMsg(PrePrepareMsg *ppMsg,
...
+    LOG_INFO(GL,
+             "creating db checkpoint in the next stable sequence number" << KVLOG(timeSinceLastSnapshot, sn, *blockId));
+    DbCheckpointManager::instance().saveDbMetadataToReservedPages(
+        sn, std::chrono::duration_cast<std::chrono::seconds>(consensus_time_ms));
+    // The above should be happen determenistically among all replicas, regardless if they actually create the
+    // checkpoint or not.
+    if (getReplicaConfig().dbCheckpointFeatureEnabled) {
+      DbCheckpointManager::instance().setCheckpointInProcess(true, *blockId);
+      onSeqNumIsStableOneTimeCallbacks_.add(
+          [blockId](SeqNum seqNum) { DbCheckpointManager::instance().createDbCheckpointAsync(seqNum, blockId); });
```
cloudnoize, line 5206: I don't fully understand the relationship between the blocked id, the sequence number that is saved in the reserved pages and the sequence number that is passed to the callback on execution.
- the block id is the latest block id at the time of executing this method.
- the sn that is saved in the reserved pages, is also the sn at the time of executing this method and matches the block id.
- the sn that is passed to the CB does not necessarily match the reserved pages sn, no?

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -5201,7 +5184,31 @@ void ReplicaImp::finishExecutePrePrepareMsg(PrePrepareMsg *ppMsg,
...
+                                   .count();
+  const auto &seq_num_to_create_dbcheckpoint = DbCheckpointManager::instance().getNextStableSeqNumToCreateSnapshot();
+  if ((timeSinceLastSnapshot >= getReplicaConfig().dbSnapshotIntervalSeconds.count() &&
+       sn - DbCheckpointManager::instance().getLastDbCheckpointSeqNum() >= getReplicaConfig().dbCheckPointWindowSize) ||
+      (seq_num_to_create_dbcheckpoint.has_value() && *seq_num_to_create_dbcheckpoint == sn)) {
+    std::optional blockId(DbCheckpointManager::instance().getLastReachableBlock());
+    LOG_INFO(GL,
+             "creating db checkpoint in the next stable sequence number" << KVLOG(timeSinceLastSnapshot, sn, *blockId));
+    DbCheckpointManager::instance().saveDbMetadataToReservedPages(
+        sn, std::chrono::duration_cast<std::chrono::seconds>(consensus_time_ms));
```
cloudnoize, line 5200: Where do we use the sn that is saved in the reserved pages?

bftengine/src/bftengine/DbCheckpointManager.cpp
```diff
@@ -177,6 +175,26 @@ uint64_t DbCheckpointManager::directorySize(const _fs::path& directory, const bo
...
+
+void DbCheckpointManager::loadDbMetadataFromReservedPages() {
+  if (!loadReservedPage(0, sizeOfReservedPage(), page.data())) {
+    LOG_WARN(getLogger(), "unable to read db checkpoint shared metadata from reserved pages");
+    return;
+  }
+  std::istringstream inStream;
+  inStream.str(page);
+  concord::serialize::Serializable::deserialize(inStream, shared_metadata_);
+  lastCheckpointCreationTime_ = shared_metadata_.lastCmdTimestamp_;
```
cloudnoize, line 196: you don't load the sn from the reserved pages

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: Wedge noops are going through a different code path (for example, they are not completely ignored and create new blocks). 
I agree that eventually, to use this new NOOP flag in wedge also, but I prefer to make it in a different PR

bftengine/include/bftengine/DbCheckpointManager.hpp
yontyon: 1. a reserved page is in size of 4K, it makes sense to allocate it only once instead of allocating it every time we want to write to the reserved page
2.  fixed

kvbc/src/reconfiguration_kvbc_handler.cpp
yontyon: It is used for retrieving the blockchain state for a client node. A client node doesn't care about the bft data, only about the blockchain itself

kvbc/tools/db_editor/include/kv_blockchain_db_editor.hpp
```diff
@@ -943,19 +943,36 @@ struct ResetMetadata {
     uint16_t numOfObjects = 0;
     auto objectDescriptors = ((PersistentStorageImp *)p.get())->getDefaultMetadataObjectDescriptors(numOfObjects);
     bool isNewStorage = mdtStorage->initMaxSizeOfObjects(objectDescriptors, numOfObjects);
+
+    // set manually the last executed sequence number and the last primary used sequence number to bypass some asserts
```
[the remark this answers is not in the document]
yontyon: no, the persistent storage API is aware of the fact that it runs under some protocol and it asserts if the data is incorrect w.r.t the protocol. For example, we cannot add a sequence number that is smaller than the one we have. However, in this specific scenario, this is exactly what we are trying to do. We are trying to align the metadata to a previous point in the past. This is why we use here the metadata interface directly.
We don't explicitly check the checkpoint file here, it is the user responsibility not to use this method without knowing what he is doing

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: we cannot "skip" a stable checkpoint, if the replica didn't get to stable checkpoint x, it won't be able to get x + 150

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: fixed

bftengine/src/bftengine/DbCheckpointManager.cpp
yontyon: fixed

eyalrund (approved):

bftengine/src/bftengine/ReplicaImp.cpp
cloudnoize: Doesn't the window allow 2 checkpoints i.e. it won't be able to get to x+300 ?

cloudnoize (approved):
LGTM!

bftengine/src/bftengine/ReplicaImp.cpp
yontyon: It is in a size of 300, but 0 indexed - for example, 0 - 299, or 150 - 449

PR #3002, opened 2023-04-25
Draft: Initial changes to handle invalid coin(s) in transfer

snehal-das:
* **Problem Overview**  
Current privacy library does not handle invalid coins gracefully. The logic of claiming coins is modified to handle this scenario.

* **Testing Done**  
Testcase: Create a scenario where an invalid coin is transferred to a user. Observe how the invalid coin is handled at the recipient's end.

### DIFF
--- a/utt/privacy-wallet-service/src/PrivacyService.cpp
+++ b/utt/privacy-wallet-service/src/PrivacyService.cpp
@@ -275,14 +275,16 @@ std::pair<utt::Transaction, utt::TxOutputSigs> PrivacyWalletServiceImpl::buildCl
       response->set_err(err_msg);
       return grpc::Status(grpc::StatusCode::ABORTED, err_msg);
     }
+  } catch (const libutt::api::operations::InvalidCoinsInTransfer& e) {
+    std::cout << e.what() << std::endl;
+    response->mutable_claim_coins_response()->set_warning(e.what());
   } catch (const std::exception& e) {
     std::cout << e.what() << std::endl;
     response->set_err(e.what());
     return grpc::Status(grpc::StatusCode::ABORTED, e.what());
   }
   if (response) {
-    auto resp = response->mutable_claim_coins_response();
-    resp->set_succ(true);
+    response->mutable_claim_coins_response()->set_succ(true);
   }
   return grpc::Status::OK;
 }
--- a/utt/utt-client-api/src/User.cpp
+++ b/utt/utt-client-api/src/User.cpp
@@ -447,9 +447,12 @@ void User::updateTransferTx(const Transaction& tx, const TxOutputSigs& sigs) {
 
     // Claim coins
     auto claimedCoins = pImpl_->client_->claimCoins(uttTx, pImpl_->params_, sigs);
-
+    bool invalidCoinsInTransfer(false);
     for (auto& coin : claimedCoins) {
-      if (!pImpl_->client_->validate(coin)) throw std::runtime_error("Invalid normal coin in transfer!");
+      if (!pImpl_->client_->validate(coin)) {
+        invalidCoinsInTransfer = true;
+        continue;
+      }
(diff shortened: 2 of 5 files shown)


snehal-das:
Changes to the privacy-wallet-service is complete and tested to work as expected. Changing status from draft to ready for merge.

utt/privacy-wallet-service/src/PrivacyService.cpp
```diff
@@ -275,14 +275,16 @@ ::grpc::Status PrivacyWalletServiceImpl::handleUserClaimCoinsRequest(::grpc::Ser
       response->set_err(err_msg);
       return grpc::Status(grpc::StatusCode::ABORTED, err_msg);
     }
+  } catch (const libutt::api::operations::InvalidCoinsInTransfer& e) {
+    std::cout << e.what() << std::endl;
+    response->mutable_claim_coins_response()->set_warning(e.what());
```
yontyon, line 280: In this case, shouldn't we return false in the succ?
Also, I would suggest returning the invalid nullifiers and the reply.
Finally, we need to expose a way for the user to remove these invalid coins, otherwise it won't be able to proceed (the invalid coins will be chose over and over again)

utt/utt-client-api/src/User.cpp
```diff
@@ -447,9 +447,13 @@ void User::updateTransferTx(const Transaction& tx, const TxOutputSigs& sigs) {
 
     // Claim coins
     auto claimedCoins = pImpl_->client_->claimCoins(uttTx, pImpl_->params_, sigs);
-
+    bool invalidCoinsInTransfer(false);
```
yontyon, line 450: use the regular convention here 
either `invalidCoinsInTransfer = false` or `invalidCoinsInTransfer{false}`

utt/utt-client-api/src/User.cpp
```diff
@@ -447,9 +447,13 @@ void User::updateTransferTx(const Transaction& tx, const TxOutputSigs& sigs) {
...
     for (auto& coin : claimedCoins) {
-      if (!pImpl_->client_->validate(coin)) throw std::runtime_error("Invalid normal coin in transfer!");
+      if (!pImpl_->client_->validate(coin)) {
+        logdbg_user << "Invalid coin found; coin details: " << dbgPrintCoins({coin}) << endl;
```
yontyon, line 453: why not throw the exception here?
(and then you don't need  `invalidCoinsInTransfer`)

utt/utt-client-api/src/User.cpp
snehal-das: I'll update in the next commit.

utt/utt-client-api/src/User.cpp
snehal-das: The idea is to prevent prematurely exiting from the function. The recipient should be able to claim coins received after the invalid coin, so I am letting it process all coins before raising the exception.

utt/privacy-wallet-service/src/PrivacyService.cpp
snehal-das: > In this case, shouldn't we return false in the succ?
Technically, the recipent was able to claim coins received. Reporting back false after claiming all valid coins seems incorrect to me. The more correct approach seems to be to return true with a warning.

>  Also, I would suggest returning the invalid nullifiers and the reply.
I didn't want the dapp to handle the invalid coins, but this is something that can be implemented. I'll discuss with you in more detail.

snehal-das:
I am not able to edit the reviewers list, the only option I see is to mention the person here. 

@bsenthilr, @yontyon Please take a look and add your thoughts.

Some items pending are:
1. Add a proper error response instead of sending back an error string.
2. Correct the initialisation of invalidCoinsInTransfer.
3. Look for better options for error reporting.

PR #3003, opened 2023-04-25
Implicit copy constructor call removal

titusen:
An implicit copy constructor call is replaced within direct shred_ptr cstr call in entire file.

Correctness is to be revealed by maestro tests.

### DIFF
--- a/bftengine/src/bftengine/ReplicaImp.cpp
+++ b/bftengine/src/bftengine/ReplicaImp.cpp
@@ -1059,7 +1059,7 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsgUPtr msg) {
              "etc...)");
     return;
   }
-  PrePrepareMsgShPtr message = std::make_shared<PrePrepareMsg>(msg.release());
+  PrePrepareMsgShPtr message(msg.release());
   if (!getReplicaConfig().prePrepareFinalizeAsyncEnabled) {
     if (!validatePreProcessedResults(message.get(), getCurrentView())) {
       // trigger view change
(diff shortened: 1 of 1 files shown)


bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1059,7 +1059,7 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsgUPtr msg) {
              "etc...)");
     return;
   }
-  PrePrepareMsgShPtr message(msg.release());
+  PrePrepareMsgShPtr message(std::move(msg));
```
glevkovich, line 1062: A more efficient way to do this, is to bind straight into a shared pointer and remove this line:
```
template <>
void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsgShPtr msg) {
```

glevkovich (changes requested):
@titusen , approved, but I suggest to changed according to the comment.

glevkovich (approved):

bftengine/src/bftengine/ReplicaImp.cpp
titusen: That would change the template method onMessage signature

eyalrund (approved):

bftengine/src/bftengine/ReplicaImp.cpp
```diff
@@ -1059,7 +1059,7 @@ void ReplicaImp::onMessage<PrePrepareMsg>(PrePrepareMsgUPtr msg) {
              "etc...)");
     return;
   }
-  PrePrepareMsgShPtr message = std::make_shared<PrePrepareMsg>(msg.release());
```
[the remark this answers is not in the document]
titusen: Because it was calling inproper cstr.

